PR #28, opened 2020-08-02
Add URL matching algorithm

### LINKED ISSUE #24: Enhance URL matching algorithm
https://github.com/ScratchAddons/extension/blob/95fa8f3f2f4857b5e49684afc3728e2de1d3fa6c/background/get-userscripts.js#L36
`https://scratch.mit.edu/users/kaj` should match `https://scratch.mit.edu/users/*/`. A slash at the end should be implicit.
I should also remove the function from the window object. I did that to debug something and forgot about it.

BoomerScratch:
Resolves #24 
This uses [a library](https://www.npmjs.com/package/wildcard) for URL matching. Tell me if there is anything wrong or I forgot something.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -42,5 +42,6 @@ function userscriptMatches(data, matches) {
 
 function urlMatchesPattern(_pattern, _url) {
   _url = ((_pattern.endsWith("/")) && (!_url.endsWith("/"))) ? `${_url}/` : _url;
+  _url = _url.split("#")[0].split("?")[0];
   return !!(wildcard(_pattern,_url));
 }
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -1,3 +1,4 @@
+import wildcard from "./wildcard.js";
 if (scratchAddons.localState.allReady) getUsercripts();
 else window.addEventListener("scratchaddonsready", getUsercripts);
 
@@ -40,13 +41,6 @@ function userscriptMatches(data, matches) {
 }
 
 function urlMatchesPattern(_pattern, _url) {
-  const pattern = _pattern.split("/");
-  const url = _url.split("/");
-  while (pattern.length) {
-    const p = pattern.shift();
-    const q = url.shift();
-    if (p !== q && p !== "*") return false;
-  }
-  return true;
+  _url = ((_pattern.endsWith("/")) && (!_url.endsWith("/"))) ? `${_url}/` : _url;
+  return !!(wildcard(_pattern,_url));
 }
-window.urlMatchesPattern = urlMatchesPattern;
--- a/background/wildcard.js
+++ b/background/wildcard.js
@@ -1,0 +1,114 @@
+/* jshint node: true */
+'use strict';
+
+var REGEXP_PARTS = /(\*|\?)/g;
+
+/**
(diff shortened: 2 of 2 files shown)


WorldLanguages:
Ohh nice!

background/get-userscripts.js
```diff
@@ -40,13 +41,7 @@ function userscriptMatches(data, matches) {
...
-  while (pattern.length) {
-    const p = pattern.shift();
-    const q = url.shift();
-    if (p !== q && p !== "*") return false;
-  }
-  return true;
-}
-window.urlMatchesPattern = urlMatchesPattern;
+  _url = ((_pattern.endsWith("/")) && (!_url.endsWith("/"))) ? `${_url}/` : _url;
+  _url = _url.split("#")[0].split("?")[0];
```
WorldLanguages, line 45: Seems pretty weird that the library wouldn't handle hashes...?
If the library is kinda "dumb" like this one, wouldn't we better off using the built-in URL constructor, add an implicit "/" at the end of the path, and compare the paths with the previous code?

WorldLanguages:
@tb148 Btw wanna check why Prettier seems to be failing? I'm still not totally sure how this should be working.

tb148:
> 
> 
> @tb148 Btw wanna check why Prettier seems to be failing? I'm still not totally sure how this should be working.

Because the code have a different style. The action would **fail** if the code does not use prettier's style. Format the code using prettier and it will pass.

tb148:
@WorldLanguages There seems to be a bug in my workflow, ignore it first. Gonna fix it in another PR.

tb148:
> 
> 
> @WorldLanguages There seems to be a bug in my workflow, ignore it first. Gonna fix it in another PR.

Fixed!

background/get-userscripts.js
BoomerScratch: Okay. So, can you give some example code?

tb148:
> 
> 
> Btw wanna check why Prettier seems to be failing? I'm still not totally sure how this should be working.

This is the expected behaviour.

WorldLanguages:
@tb148 Want to write a guide on how to set up Prettier on our forks? Or is it automatic?

background/get-userscripts.js
WorldLanguages: See https://developer.mozilla.org/en-US/docs/Web/API/URL
Try getting the pathname for the URLs, add an "/" to the URL if it doesn't have one, then compare them with the previous code I written (or write your own, if you'd like)

tb148:
> 
> 
> @tb148 Want to write a guide on how to set up Prettier on our forks? Or is it automatic?

There are two steps:
1. [Sync the fork](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) so the newest version of the action is used.
2. Go to the "Actions" page of the fork and enable it.

WorldLanguages:
@BoomerScratch Can I take this? This probably can be simplified by using the URL constructor like I said in a comment. A library is definitely not needed.

BoomerScratch:
@WorldLanguages Sure!

WorldLanguages:
Okay. Closing, then (I'll make another pull request soon)

PR #55, opened 2020-08-10
Add lint workflow

### LINKED ISSUE #54: [Feature] Automated linting
We can use [ESLint](https://eslint.org) for linting code automatically on pull requests and commits. ~~Note: this requires a `package.json`.~~ EDIT: noticed it isn't needed 😉 It is only required for generating a `.eslintrc.json` file.

BoomerScratch:
**Resolves**

Resolves #54 

**Changes**

Add `.eslintrc.json` and a new workflow.

**Reason for changes**

To lint code without having to run it manually.

**Tests**

Nothing really to test.

(I'm still new to GitHub actions so tell me if there is anything wrong :stuck_out_tongue_winking_eye:)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,12 @@
+{
+    "env": {
+        "browser": true,
+        "es2020": true
+    },
+    "parserOptions": {
+        "ecmaVersion": 11,
+        "sourceType": "module"
+    },
+    "rules": {
+    }
+}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,14 @@
+name: Lint code
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: Run lint
+        run: npx eslint .


WorldLanguages:
I think we're already linting automatically with Prettier?

BoomerScratch:
@WorldLanguages Hmm... it seems like you are correct...

WorldLanguages:
I believe you need to enable it manually when forking, by going into "actions" then click the green button. But after that, it's automatic.
Keep in mind however, that because of this, if you push too often you might get git conflicts. Make sure to wait until the Prettier workflow stops before pushing again.

BoomerScratch:
@WorldLanguages I have found out the linter of Prettier does not lint everything. For example, try this:
```javascript
/(*)/;
```
This should raise a parsing error. ESLint raises indeed a parsing error, but the linter of Prettier doesn't.

Also, about the git conflicts: if that happens, you should force-push it using the `--force` flag.

WorldLanguages:
> ESLint raises indeed a parsing error, but the linter of Prettier doesn't.
Well, in that case the script has a SyntaxError, which would probably be caught while testing, right?
We could use ESLint, but imo we should have as few rules as possible, unless there's a very obvious error. Stuff like "this function parameter isn't used" and similar should only warn, not throw, what do you think?

`Also, about the git conflicts: if that happens, you should force-push it using the --force flag.`
Lol yeah, but I'd rather wait 15 seconds for it to lint and avoid force pushing, since I'm still a newbie :P

BoomerScratch:
> Stuff like "this function parameter isn't used" and similar should only warn, not throw, what do you think?

I'm not sure if that is possible, but a warning for unused variables and functions _is_ possible.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,14 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: Run lint
+        run: npx eslint .
```
tb148, line 14: Uhh.. Is ESlint installed by default?

.eslintrc.json
```diff
@@ -0,0 +1,12 @@
...
+        "browser": true,
+        "es2020": true
+    },
+    "parserOptions": {
+        "ecmaVersion": 11,
+        "sourceType": "module"
+    },
+    "rules": {
+    }
+}
```
tb148, line 12: Should we also add [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) and add it to the workflow?
We also need to add `extends`:
```
{
  "extends": [
    "some-other-config-you-use",
    "prettier"
  ]
}
```

.github/workflows/lint.yml
BoomerScratch: Not sure.

.eslintrc.json
BoomerScratch: What does it do? Linting _and_ formatting?

tb148: > 
> 
> What does it do? Linting _and_ formatting?

According to official documentation, it:
> Turns off all rules that are unnecessary or might conflict with Prettier.

tb148:
Also it seems like the linting and formatting workflow can be merged by [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier), but I don't know whether or not to do so.

Hans5958:
Just use [DeepScan](https://deepscan.io/) and just fix the ESLint errors manually, duh.

WorldLanguages:
> duh
@Hans5958 Use proper language please lol, we're all learning here

Hans5958:
Oh, whoops. Apologies.

BoomerScratch:
This PR has been inactive for a while, and I don't think we need this. Closing the pull request, but keeping the `lint-action` branch, if someone changes their mind.

PR #57, opened 2020-08-12
Remove content security policy

### LINKED ISSUE #56: Iconify Won't Work On Extensions Page
**Describe the bug**

**To Reproduce**

Open the options page. See Not Working icons
Open console
See 
not available
﻿
index.html:121 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-rHlxGRx2sBxFirVgDiDAJcH8G0KjhzqrwHXB4CNXtak='), or a nonce ('nonce-...') is required to enable inline execution.
index.html:1 Refused to load the script 'https://code.iconify.design/1/1.0.7/iconify.min.js' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

**Expected behavior**

Icons should work

ChromeOS

BoomerScratch:
**Resolves**

Resolves #56

**Changes**

Remove content_security_policy from the manifest.

**Reason for changes**

Why do we need a content security policy? It's useless.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -30,7 +30,6 @@
     "storage",
     "notifications"
   ],
-  "content_security_policy": "script-src 'self' 'unsafe-eval'; object-src 'self'",
   "web_accessible_resources": [
     "content-scripts/inject/*",
     "addon-api/*",


WorldLanguages:
It's needed because the default CSP doesn't include `unsafe-eval` and eval is used to execute addons.

PR #101, opened 2020-08-23
Add addon.tab.waitForElement

### LINKED ISSUE #96: [Feature] On react page load

there is no way (i think) to detect when a react page (all scratch-www pages) such as https://scratch.mit.edu/projects/80 is fully loaded. this makes it really annoying and difficult to make a script that modifies the page after the load. (or maybe i'm stupid)

an event part of the addon.* api that tells me when the page is actually done loading. it's probably possible because userscript managers can do it with `// @run-at document-start`. 

for an addon i just made it wait 1 one second but this is a terrible solution for many reasons

**Additional context**

if i try something like `document.getElementById('title').innerText = 'helo'`, it doesn't exist yet so it doesn't work. `window.onload` or whatever doesn't work either since react is wack

easrng:
**Resolves**

Resolves #96

**Changes**

Adds addon.tab.waitForElement

**Reason for changes**

This function allows addons to wait for elements to appear in the DOM. For example, `await addon.tab.waitForElement("#navigation span.message-count")` will wait for the message-count element to load. If the element already exists, it will resolve instantly.

**Tests**

I haven't tested it with the rest of the extention, but it works on it's own.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -9,4 +9,20 @@ export default class Tab {
   getScratchVM() {
     return scratchAddons.methods.getScratchVM();
   }
+  waitForElement(selector) {
+    if (!document.querySelector(selector)) {
+        return new Promise((resolve) => new MutationObserver(function(mutationsList, observer) {
+            if (document.querySelector(selector)) {
+                observer.disconnect();
+                resolve();
+            }
+        }).observe(document.body, {
+            attributes: true,
+            childList: true,
+            subtree: true
+        }))
+    } else {
+        return Promise.resolve()
+    }
+  }
 }


WorldLanguages:
> I haven't tested it with the rest of the extention, but it works on it's own.

Yeah ATM you can't enable addons with the UI so if you want to test it, you can enable the discuss button addon by inspecting the background page inside `chrome://extensions` and then run `changeEnabledState("discuss-button", true)`. Then, you can modify `addons/discuss-button/userscript.js` to test it, and then recover its original state before committing.

WorldLanguages:
Please enable GH actions in your fork, so that it gets linted there and not here. For next time :) Just go to the "actions" tab and click the big green button.

PR #107, opened 2020-08-25
Remove addon.auth.sessionId

### LINKED ISSUE #100: Remove addon.auth.sessionId
`scratchsessionsid` is an HTTP only cookie, so I don't see the point why it should be ever stored by the extension, or accessed by addons. I only see security concerns.

BoomerScratch:
**Resolves**

Resolves #100

**Changes**

Remove all uses of sessionId.

**Reason for changes**

#100

**Tests**

Not tested, so we'll see if it works lol


WorldLanguages:
Let me test it...

Issue #70, opened 2020-08-17
[New addons] Cat+ userscripts


WorldLanguages:
**Describe the new addon**

Userscripts that improve Scratch 3. (currently unmaintained, many scripts are broken or work inconsistently)

**Link to the new addon**

https://github.com/forkphorus/cat-plus#readme

GarboMuffin:
Hi, I was the developer of the Cat+ userscripts. I noticed I never added a license so I've done that now.

The only scripts that I think might be worth porting to ScratchAddons would be:

 - colored-editor-context-menu.user.js
 - extra-metadata.user.js
 - make-projects-shared.user.js
 - searchable-dropdowns.user.js (has some weird quirks but it works)

The other ones are probably broken or griffpatch's developer tools do it better.

I'd be willing to submit PRs for some of these, if you'd like.

WorldLanguages:
Thanks for commenting on the issue, @GarboMuffin!
Thanks for offering to help, but keep in mind there are still many upcoming wiki pages to write. If you feel lost, that's expected at this stage. Only if you feel like it, try to make a  PR. There's a small guide on how to make a new addon here: https://github.com/ScratchAddons/ScratchAddons/issues/74#issuecomment-678488280
Make sure to refresh the extension from `chrome://extensions` after making changes to your addon.
Feel free to tag me if you need help with a specific issue.

TheColaber:
Are there any other things we might want from Cat+?

Ani-Xan:
@WorldLanguages @TheColaber 
I think we should add an option to toggle the unrounded stage within the editor menus of Scratch (and add the unrounded stage userscript). We should have it enabled by default, but people who really want it have to go into the editor of a project and enable it. 👍🏻 
😸

GrahamSH-LLK:
We've ported 2 of these, and the others have their own issues.

Ani-Xan:
@GrahamSH-LLK We should find a fix.

GrahamSH-LLK:
> @GrahamSH-LLK We should find a fix. (or alter the code)

What? Oh, I meant issues as in GitHub Issues, like this one.

Ani-Xan:
@GrahamSH-LLK Oh, lol. 😖

PR #116, opened 2020-08-27
Add colored context menus addon

GarboMuffin:
**Resolves**

Part of #70

**Changes**

Adds an addon that makes the color of context menus in the editor match the block that was clicked:

image

**Reason for changes**

It looks cool and Scratch 2 did it. That's about the only reason.

**Tests**

It works in Chrome and Firefox.

It might break if the addon runs before the body is setup but I'm not sure if that's possible in this extension.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -1,1 +1,1 @@
-["msg-count-badge", "scratch-notifier", "discuss-button", "editor-devtools", "editor-stepping", "image-uploader"]
+["msg-count-badge", "scratch-notifier", "discuss-button", "editor-devtools", "editor-stepping", "image-uploader", "editor-colored-context-menus"]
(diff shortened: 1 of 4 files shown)


WorldLanguages:
> It might break if the addon runs before the body is setup but I'm not sure if that's possible in this extension.
All userscripts run as if they were script tags with "defer", so yep, the body is always available. That will soon be in the docs.

WorldLanguages:
Btw about prettier: you can enable GitHub Actions in your fork. Or you can run "npx prettier --write ." locally in the repo folder (be careful where you run that command).
Or just don't care about it at all. If I merge this PR, it will be linted through another commit on master anyway.

GarboMuffin:
I fixed all the issues that prettier was complaining about. This is ready to be reviewed.

PR #135, opened 2020-08-31
Prototype handler

### LINKED ISSUE #123: [Discussion] Prototype Injector
I am working on Prototype Injector. 

## Once and Many
"Once" refers to objects that are singleton (like VM) - because you only need to receive "once". "Many" is for non-singleton, like Thread. References to "Once" objects are kept in a variable, but you need to keep references to "Many" yourself if needed.

## API
Here are the APIs:
- `addon.tab.onceValues` - reference storage for "Once", See below for keys.
- `addon.tab.addOnceListener/addManyListener` - adds listener, like `addEventListener`. When an object is trapped, the handlers will fire. Event name may be chosen from ones listed below if you just want to listen to one object trapped - or if `*` is specified as event name, it will listen to all trapped objects for Once/Many, and you can check `event.prototypeName` to see what's trapped. In both cases `event.value` is the trapped object itself.
- `addon.tab.removeOnceListener/removeManyListener` - removes listener.

`vmready` event will be deprecated in favor of this. use
```js
addon.tab.addOnceListener("*", ev => {
    if (ev.prototypeName.startsWith("vm")) doSomething(ev.value)
})
```
instead. `addon.tab.getScratchVM` will still be alive but with more accuracy.

## Injected Attribute
For all trapped objects, `__scratchAddonsReadyOnce/Many` attribute is injected to count how many times it was trapped (starting with 1). Useful for initializing further pollutions.

## Available objects for now
Currently, only `thread` is Many. Others are all Once.

- `vm`: VM trapped using current method
- `vm.propsVMBind`: VM trapped by checking `props.vm` during `bind`
- `vm.propsVMAssign`: VM trapped by checking `props.vm` during `assign`
- `thread`: `Thread` trapped during `push`.
- `ScratchBlocks`: `ScratchBlocks` trapped during `setLocale` trapped during `bind`.

apple502j:
**Resolves**
Resolves #123 

**Changes**
Adds prototype handler. There are a lot of changes since the draft linked. Wiki coming soon, but for now you can read JSDoc

**Reason for changes**
Because useful

**Tests**
Tested myself


JeffaloBob:
i dont know what this means but it looks cool! 🎉

WorldLanguages:
@BoomerScratch In case you didn't notice, this API allows your "highlight the current executing block" to work without an interval!

addon-api/content-script/Tab.js
```diff
@@ -27,4 +40,16 @@ export default class Tab {
       return Promise.resolve();
     }
   }
+  /**
+   * @type {string} editor mode (or empty string for non-editors).
+   */
+  get editorMode() {
```
WorldLanguages, line 46: Looks good.
Self TODO: add this to wiki addon.* APIs

addon-api/content-script/Tab.js
```diff
@@ -27,4 +40,16 @@ export default class Tab {
...
+    const pathname = location.pathname.toLowerCase();
+    const split = pathname.split("/").filter(Boolean);
+    if (!split[0] || split[0] !== "projects") return "";
+    if (split.includes("editor")) return "editor";
+    if (split.includes("fullscreen")) return "fullscreen";
+    if (split.includes("embed")) return "embed";
```
WorldLanguages, line 52: Note that Scratch Addons won't run on iframes! But users can still go to the embed URL in a new tab, and Scratch Addons will run there.

addon-api/content-script/Tab.js
```diff
@@ -27,4 +40,16 @@ export default class Tab {
       return Promise.resolve();
     }
   }
+  /**
+   * @type {string} editor mode (or empty string for non-editors).
+   */
+  get editorMode() {
+    const pathname = location.pathname.toLowerCase();
+    const split = pathname.split("/").filter(Boolean);
+    if (!split[0] || split[0] !== "projects") return "";
```
WorldLanguages, line 49: Might be better to return `null`, maybe?

addon-api/content-script/Tab.js
WorldLanguages: Added to wiki

addon-api/content-script/Tab.js
WorldLanguages: Note: I know I'm saying that after merging (lol), but no addon is using this yet (checking against empty string, or null) so it's fine.

PR #175, opened 2020-09-02
More features for Scratch Messaging addon

### LINKED ISSUE #108: Port Griffpatch's Messaging Extension
Which means I'll need to add support for popups: #16 
This will also likely require additions to the `addon.account` API to avoid duplicated work.

**For everyone reading this: there's a chance griffpatch will agree to rename the Messaging Extension to Scratch Addons, and upload this extension in its place. This would mean immediate install to 3000+ users if this addon works as supposed! This is epic, so this is a pretty important issue. Please add a tada reaction to pressure me into doing this right lol**

WorldLanguages:
Kinda resolves #108 

- Replying to comments works as expected
- Comments are loaded when clicking on the popup, for up to date information. They load from top to bottom.
- Fix links to Scratch URLs by adding <base> element in head
- Scratch Notifier and Scratch Messaging: fix unlikely, but possible timing bug (auth change event happening while ajax promises pending, like message count and message content)
- Add studio activity
- Remove useless console logs
- Remove chrome onMessage event that had an async function


WorldLanguages:
Hmm getting conflicts on my own addon, guess I shouldn't have used the same branch :P

WorldLanguages:
I have some missing TODOs, will undraft after then.

Issue #176, opened 2020-09-02
Use CDNs for libraries used in userscripts?


WorldLanguages:
Modules are nice because they don't pollute the global scope... However, it's not really great to have libraries inside the extension source code, that don't run in extension context at all. Extension reviewers might think there's something sketchy there, because minified code can hide malicious code. Of course, Vue and Redux are exception to this rule, since there's no way to load them through a CDN without requiring a less strict content security policy (or in the case of Redux, the traps wouldn't work at all, I think)

WorldLanguages:
At the moment, these libraries would be `text-field-edit.js`, `moment.js`, and `md5.js`, as well as some in addons still in PR (like @Hans5958's - but don't worry, we'll accept these PRs then make the changes).
I propose having `addon.tab.loadScript(url)`, returning a promise that resolves when the script has executed (onload), or rejecting if it somehow failed (onerror).

JeffaloBob:
what about how some libraries had to be modified slightly to work?

Hans5958:
> 
> 
> what about how some libraries had to be modified slightly to work?

I think it will be a really rare case for that, but well, edge cases.

WorldLanguages:
> what about how some libraries had to be modified slightly to work?

Those were modified to be converted into modules that can be imported directly to addon script modules - however, what I'm saying here is to just pollute the global scope (so any addon will be able to use `md5` if only one addon imports it)

BoomerScratch:
> I propose having `addon.tab.loadScript(url)`, returning a promise that resolves when the script has executed (onload), or rejecting if it somehow failed (onerror).

But that is the same as `import()`, right?

Hans5958:
I don't think it does.

BoomerScratch:
@Hans5958 Use it like this:
```javascript
import("URL to script here")
```
For example:
```javascript
import("https://code.jquery.com/jquery-3.5.1.min.js"); // Imports JQuery
```

Hans5958:
I never import modules on my day-to-day basis so I don't know.

WorldLanguages:
> But that is the same as `import()`, right?

Kinda - if 2 addons want to load the same URL, and it has already been loaded by another addon, an `addon.*` API could figure that out. Keep in mind this is only for typical libraries that go in the global scope (like jQuery on `window.$`), modules are different.

PR #178, opened 2020-09-03
addon.json changes

### LINKED ISSUE #159: [Question] Can you make the scripts run faster and earlier?
The addon is completely awesome, no question. But a little thing that annoys me is that the discuss button only loads after a second. Same thing goes for my forkphorus button. Is it possible to make that faster?

### LINKED ISSUE #160: [Question] Should you add your name to the description?
Currently, only the deleted comments thing includes "by abc" (in this case @JeffaloBob) in the description, so I wonder if you should always add that, if we should add a "creator" tag in the JSON that would be placed next to the name of the addon or just let it be.

apple502j:
**Resolves**
Resolves #160 
Resolves #159 

**Changes**
addon.json can now have these fields:
- `userscripts` section can now have `runAtComplete` (bool, defaults to true) - if true, scripts run at onload (like it did before), and if false, at DOMContentLoaded. Userstyles load at DOMContentLoaded.
- `credits` is an array of `{"name": "string", "link": "link URL (optional)"}`
- `libraries` is an array of library name, e.g. `["md5", "text-field-edit"]`
**Reason for changes**
Because it's nice

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*]
+indent_size = 2
--- a/addons/discuss-button/addon.json
+++ b/addons/discuss-button/addon.json
@@ -4,7 +4,8 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["https://scratch.mit.edu/*"]
+      "matches": ["https://scratch.mit.edu/*"],
+      "runAtComplete": false
     }
   ],
   "options": [
--- a/addons/forum-time-zones/addon.json
+++ b/addons/forum-time-zones/addon.json
@@ -9,5 +9,6 @@
     }
   ],
   "tags": ["forums", "community"],
-  "enabled_by_default": false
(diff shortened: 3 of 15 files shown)


JeffaloBob:
>   **Reason for changes**
>   Because it's nice
agreed

Hans5958 (commented):
Don't forget to handle external libraries for #176, probably?

NoobTracker:
Ah, thank you a lot! That faster loading is cool! I already thought about making that somehow. Yeah, that also nicely keeps old things compatible.

NoobTracker:
So, why isn't that already merged?

NoobTracker:
Okay, little question: How do the credits look like?

apple502j:
Credits are just `Credits to: A, B, C`

NoobTracker:
Ah, okay. It would be really nice if you could make a screenshot.

PR #181, opened 2020-09-03
Ctrl + F Fix

### LINKED ISSUE #177: Options page: focus search box when pressing Ctrl+F
...instead of browser search. If the user really wants to use browser search, they can focus the address bar, and then press Ctrl+F.
I'm creating this issue bc it should be easy to implement, and I've seen myself pressing Ctrl+F trying to focus the search box, when it's already autofocused (silly me...), and browser search doesn't work as well.

GrahamSH-LLK:
**Resolves**
Resolves #177 
**Changes**

Focuses the search box when ctrl + f is clicked.

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -132,4 +132,4 @@ window.addEventListener("keydown",function (e) {
         $('#searchBox').focus();
     }
 })
- 
+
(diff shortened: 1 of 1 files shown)


NoobTracker:
That sounds like a good idea!

webpages/settings/index.js
```diff
@@ -126,3 +126,9 @@ chrome.runtime.sendMessage("getSettingsInfo", ({ manifests, addonsEnabled, addon
 vue.$watch("selectedTab", function (newSelectedTab) {
   this.selectedTag = null;
 });
+window.addEventListener("keydown", function (e) {
+  if (e.keyCode === 114 || (e.ctrlKey && e.keyCode === 70)) {
+    e.preventDefault();
+    $("#searchBox").focus();
```
WorldLanguages, line 132: I don't think there's jQuery in this context. This might have worked in the console because `$` is a shortcut for `document.querySelector`, but I don't think this would actually work.

Issue #185, opened 2020-09-03
[New addon] Replacing /discuss/youtube links with true youtube links


NoobTracker:
**Describe the new addon**

It would just replace links for those annoying embedded videos with real videos. It would have two parts:
The first one replaces the links for /discuss/youtube with true youtube links so when you actually click somewhere where someone else wants to rickroll you without having the extension you get to the YT page instead. It may just be only triggered when you visit /discuss/youtube so you get redirected from there in case it isn't that easy to do that straight with the link. 
The second one replaces the y in youtube with %79 so the censor has no clue that you just fooled it. That could just be done for every URL since it changes more or less nothing. It happens when you create a link in the forums. This part effects everyone so you can finally link something with `?t=whatever`.
The second part might be removed and is optional in case that causes issues.

**Link to the new addon**

Doesn't exist so far.

Hans5958:
@WorldLanguages close pls

Issue #172, opened 2020-09-02
[Discussion] Multiple options for things being enabled by default


NoobTracker:
My idea is that you can select which extensions should be enabled, so you can have different categories, such as

- None
- Few
- Recommended
- All

So I asked a while ago what exactly can be enabled by default and it's kinda the second category so I think it would be neat to have "Recommended" stuff like the discuss button and the image uploader. What do you think?

YOYITsM3M8:
Yes, this would be amazing

YOYITsM3M8:
But how would we decide what is in what

NoobTracker:
Common sense. As @WorldLanguages said about being enabled by default. You recommend stuff like the image uploader, but not a dark theme. Small but useful things are recommended.

Explosion-Scratch:
Sure! This sounds great!

Explosion-Scratch:
Also, maybe merge all addons even if they're experimental but put them in the 'beta' section?

NoobTracker:
@Explosion-Scratch Every single addon? Some work quite well, that's not beta.

YOYITsM3M8:
We can test which are good and not put them in beta

NoobTracker:
I guess we can

WorldLanguages:
Closed in favor of #318

PR #198, opened 2020-09-04
True YouTube links

NoobTracker:
The extension replaces every https://scratch.mit.edu/discuss/youtube/ link with a youtu.be link.
I think that it's okay that it's enabled by default since it is more or less a fix.
I guess seeing Rick's face in a YT tab is nicer then doing it in a Scratch tab. 

I can remove the default tag but I see no reason why.
Also, should I add the tag "forum"?

### DIFF
--- a/addons/true-youtube-links/addon.json
+++ b/addons/true-youtube-links/addon.json
@@ -14,6 +14,6 @@
     }
   ],
   "options": [],
-  "tags": ["community"],
-  "enabled_by_default": true
+  "tags": ["community", "forums"],
+  "enabled_by_default": false
 }
--- a/addons/true-youtube-links/addon.json
+++ b/addons/true-youtube-links/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "True YT links",
+  "name": "True YouTube links",
   "description": "Replaces links to the embedded site with true YT links",
   "credits": [
     {
--- a/addons/true-youtube-links/addon.json
+++ b/addons/true-youtube-links/addon.json
@@ -1,0 +1,20 @@
+{
+  "name": "True YT links",
+  "description": "Replaces links to the embedded site with true YT links",
+  "credits": [
+    {
+      "name": "HTML-Fan/NT"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/*"],
+      "runAtComplete": false
+    }
+  ],
+  "options": [
+  ],
+  "tags": ["community"],
+  "enabled_by_default": true
+}
--- a/addons/true-youtube-links/userscript.js
+++ b/addons/true-youtube-links/userscript.js
@@ -1,0 +1,11 @@
+﻿export default async function ({ addon, global, console }) {
+  addon.settings.addEventListener("change", () => console.log("changed!"));
+  while(true){
+    await addon.tab.waitForElement("a:not(.trueYTLinksViewed)");
+    var element = document.querySelector("a:not(.trueYTLinksViewed)");
+    if(element.href.indexOf("https://scratch.mit.edu/discuss/youtube/") == 0){
+      element.href = element.href.replace("https://scratch.mit.edu/discuss/youtube/", "https://youtu.be/");
+    }
+    element.classList.add("trueYTLinksViewed");
+  }
+}
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -13,5 +13,6 @@
   "move-your-cat",
   "external-player-button",
   "remove-sprite-confirm",
-  "editor-searchable-dropdowns"
+  "editor-searchable-dropdowns",
+  "true-youtube-links"
 ]


NoobTracker:
The second part I mentioned in #185 could be added later.

NoobTracker:
@JeffaloBob I think you messed with the inserting things in the text box stuff when you worked on the image uploader - can you help me to code the second part that just converts `youtu` to `%79outu`? Why `youtu`? ~~Because we only have 8 KB ROM and we want to waste as little as possible~~ Because that way we don't have to code a special case for `youtu.be`.

addons/true-youtube-links/userscript.js
```diff
@@ -0,0 +1,11 @@
+﻿export default async function ({ addon, global, console }) {
+  addon.settings.addEventListener("change", () => console.log("changed!"));
+  while (true) {
```
WorldLanguages, line 3: Why a loop? Posts can't be dynamically added without the page refreshing. You should use document.querySelectorAll then loop through the elements.

addons/true-youtube-links/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/*"],
+      "runAtComplete": false
+    }
+  ],
+  "options": [],
+  "tags": ["community"],
+  "enabled_by_default": true
```
WorldLanguages, line 18: This should be false

addons/true-youtube-links/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/*"],
+      "runAtComplete": false
+    }
+  ],
+  "options": [],
+  "tags": ["community"],
```
WorldLanguages, line 17: This should also include "forums".

addons/true-youtube-links/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "Replaces links to the embedded site with true YT links",
+  "credits": [
+    {
+      "name": "HTML-Fan/NT"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/*"],
```
WorldLanguages, line 12: This should be /discuss/*

WorldLanguages (changes requested):

JeffaloBob:
don't automatically edit YouTube URLs people are posting please, that would be against the CGs for everyone who uses it.

WorldLanguages:
> don't automatically edit YouTube URLs people are posting please

I don't think this does that?

addons/true-youtube-links/addon.json
NoobTracker: Okay .... if you say so ...

addons/true-youtube-links/addon.json
NoobTracker: Okay

addons/true-youtube-links/addon.json
NoobTracker: Why? There might be a YT-Video linked in the comments or ... well, anywhere. It can happen and in that case it should still work.

addons/true-youtube-links/userscript.js
NoobTracker: Comments can.

WorldLanguages: Oh, ok. You should use document.querySelectorAll anyway. You can keep the while true loop (and remove the second line of the file that makes a useless console log)

WorldLanguages: If when you first load the page (the first time the while loop runs) there's more than one YouTube link, not using querySelectorAll will only match the first link.

addons/true-youtube-links/addon.json
NoobTracker: But why exactly? It is kinda like a fix.

NoobTracker:
> 
> 
> don't automatically edit YouTube URLs people are posting please, that would be against the CGs for everyone who uses it.

Against which CG?

addons/true-youtube-links/addon.json
WorldLanguages: Please stop trying to discuss everything I say :( It's clearly not a fix. Links to YouTube do not go to YouTube with the intention of them not going to YouTube.

NoobTracker: Okay ........

addons/true-youtube-links/userscript.js
NoobTracker: But one loop later it also replaces the next one.
It really replaces every link. I checked that.

WorldLanguages: Hmm, you're right

JeffaloBob:
> Against which CG?

well, i assumed that getting around a filter designed to keep scratch safe was against the CG.

like auto redirecting for users is fine, but replacing user's links seems a little too far. (yes i know it doesn't do that yet)

NoobTracker:
> 
> 
> > Against which CG?
> 
> well, i assumed that getting around a filter designed to keep scratch safe was against the CG.
> 
> like auto redirecting for users is fine, but replacing user's links seems a little too far. (yes i know it doesn't do that yet)

Well whenever I want to link a certain part of a YT video I manually use the exploit.
And it's kinda ironic that exactly you complain since you added the %2E hack to your extension I guess.

WorldLanguages:
> And it's kinda ironic that exactly you complain since you added the %2E hack to your extension I guess.

It's just a bit ironic; the %2E hack can only be used for `*.scratch.mit.edu` subdomains that aren't whitelisted. While we are 100% sure they don't want direct links to YouTube to be enabled by default to thousands of extension users, we aren't 100% sure there would be a problem with using Scratch's own domains for image hosting.

NoobTracker:
Well they want you to use a separate image host to upload stuff - otherwise they would have that uploader by default, right?
And also - if they really don't want you to get to YT, why do these embed pages still have the YT link although it would be pretty easy to remove it?

addons/true-youtube-links/addon.json
NoobTracker: I think that we really need recommended stuff, #172

PR #201, opened 2020-09-04
Bitmap clipboard

### LINKED ISSUE #204: [Bug] Tab.js:17 Cannot assign to read only property 'path' of object '#<CustomEvent>'
**To Reproduce**

Open a project in Chrome. 

**Expected behavior**

No errors

**Screenshots**

image

Chrome only bug

apple502j:
Resolves #204 

This adds a small addon which allows you to copy the paint editor (bitmap only for now) content. Also a lot of bugfixes. This will add `clipboardWrite` to the permissions for Firefox.


WorldLanguages:
> This will add clipboardWrite to the permissions for Firefox.
... and Chrome as well? In Chrome(ium) (where we will have the most users) we can't add any more permissions (except `tabs`, currently unused, probably won't ever be) because updating from Better3.0 or Scratch Messaging Extension wouldn't work, since Scratch Addons would require more permissions than those. Instead, we should use optional permissions, like the Scratch Notifier addon does - it requests the notifications permission to be enabled for it to be enabled through the `permissions` array in the manifest.

Explosion-Scratch:
Wait, why isn't this merged?

Explosion-Scratch:
Oh, nvm, it's awaiting WorldLanguages' review.

apple502j:
@WorldLanguages We cannot use optional permission because we can't ask for it inside background task or redux handler.

background/handle-clipboard.js
```diff
@@ -0,0 +1,26 @@
...
+  for (let i = 0; i < byteString.length; i++) {
+    uintArray[i] = byteString.charCodeAt(i);
+  }
+  return arrayBuffer;
+};
+
+if (typeof browser !== "undefined") {
+  // Firefox
+  browser.runtime.onMessage.addListener(function (request) {
+    if (request.clipboardDataURL && browser && browser.clipboard && browser.clipboard.setImageData) {
```
WorldLanguages, line 15: Well, `browser` will be truthy here unless it's 0 :P

WorldLanguages (approved):

WorldLanguages:
List of changes:
- Added `clipboardWrite` optional permission.
- Makes API changes to Redux API to fix #204 
- Adds clipboard handler on background
- Removes global state from DOM after it's been read - no more visible tokens there

PR #206, opened 2020-09-05
Update "True YouTube links" addon (05/09/2020)

Hans5958:
**Changes**

This pull request introduces this request.

- Change CSS selector so tagging with class is not needed
- Use the direct YouTube link (https://www.youtube.com/watch?v=) instead of the shortened format (https://youtu.be/)

**Reason for changes**

Why not? Is that acceptable?

**Tests**

In theory, it should have worked. I'm not sure how to test it because I don't know any page that has a YouTube link, so please help with testing it.

### DIFF
--- a/addons/true-youtube-links/userscript.js
+++ b/addons/true-youtube-links/userscript.js
@@ -1,13 +1,14 @@
 ﻿export default async function ({ addon, console }) {
-
   addon.settings.addEventListener("change", () => console.log("changed!"));
 
   while (true) {
     await addon.tab.waitForElement("a[href^=https://scratch.mit.edu/discuss/youtube/");
     var elements = document.querySelectorAll("a[href^=https://scratch.mit.edu/discuss/youtube/]");
-    elements.forEach(element => {
-      element.href = element.href.replace("https://scratch.mit.edu/discuss/youtube/", "https://www.youtube.com/watch?v=");
-    })
+    elements.forEach((element) => {
+      element.href = element.href.replace(
+        "https://scratch.mit.edu/discuss/youtube/",
+        "https://www.youtube.com/watch?v="
+      );
+    });
   }
-
 }
--- a/addons/true-youtube-links/userscript.js
+++ b/addons/true-youtube-links/userscript.js
@@ -1,11 +1,13 @@
-﻿export default async function ({ addon, global, console }) {
+﻿export default async function ({ addon, console }) {
+
   addon.settings.addEventListener("change", () => console.log("changed!"));
+
   while (true) {
-    await addon.tab.waitForElement("a:not(.trueYTLinksViewed)");
-    var element = document.querySelector("a:not(.trueYTLinksViewed)");
-    if (element.href.indexOf("https://scratch.mit.edu/discuss/youtube/") == 0) {
-      element.href = element.href.replace("https://scratch.mit.edu/discuss/youtube/", "https://youtu.be/");
-    }
-    element.classList.add("trueYTLinksViewed");
+    await addon.tab.waitForElement("a[href^=https://scratch.mit.edu/discuss/youtube/");
+    var elements = document.querySelectorAll("a[href^=https://scratch.mit.edu/discuss/youtube/]");
+    elements.forEach(element => {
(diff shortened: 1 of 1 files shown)


addons/true-youtube-links/userscript.js
```diff
@@ -1,11 +1,14 @@
-﻿export default async function ({ addon, global, console }) {
+﻿export default async function ({ addon, console }) {
```
BoomerScratch, line 1: I think this will break. Not sure though.
```suggestion
﻿export default async function ({ addon, global, console }) {
```

Hans5958: It is okay to remove unused variables declared within a function.

BoomerScratch: Yes, but `console` will now be `global`.

addons/true-youtube-links/userscript.js
```diff
@@ -1,11 +1,14 @@
...
   addon.settings.addEventListener("change", () => console.log("changed!"));
+
   while (true) {
-    await addon.tab.waitForElement("a:not(.trueYTLinksViewed)");
-    var element = document.querySelector("a:not(.trueYTLinksViewed)");
-    if (element.href.indexOf("https://scratch.mit.edu/discuss/youtube/") == 0) {
-      element.href = element.href.replace("https://scratch.mit.edu/discuss/youtube/", "https://youtu.be/");
-    }
-    element.classList.add("trueYTLinksViewed");
+    await addon.tab.waitForElement("a[href^=https://scratch.mit.edu/discuss/youtube/");
```
apple502j, line 5: Missing `]`? Also it should be `""`ed

addons/true-youtube-links/userscript.js
```diff
@@ -1,11 +1,14 @@
...
+    var elements = document.querySelectorAll("a[href^=https://scratch.mit.edu/discuss/youtube/]");
```
apple502j, line 6: "" here too

addons/true-youtube-links/userscript.js
apple502j: No, it won't.

addons/true-youtube-links/userscript.js
Hans5958: Not sure if it's needed. Added it nonetheless.

addons/true-youtube-links/userscript.js
```diff
@@ -1,11 +1,14 @@
...
   addon.settings.addEventListener("change", () => console.log("changed!"));
+
   while (true) {
-    await addon.tab.waitForElement("a:not(.trueYTLinksViewed)");
-    var element = document.querySelector("a:not(.trueYTLinksViewed)");
-    if (element.href.indexOf("https://scratch.mit.edu/discuss/youtube/") == 0) {
-      element.href = element.href.replace("https://scratch.mit.edu/discuss/youtube/", "https://youtu.be/");
-    }
-    element.classList.add("trueYTLinksViewed");
+    await addon.tab.waitForElement('a[href^="https://scratch.mit.edu/discuss/youtube/"]');
```
WorldLanguages, line 5: Clever.

WorldLanguages (changes requested):
In some situations, the href attribute isn't a full URL, but it looks like this: `/discuss/youtube/dQw4w9WgXcQ/`. Those aren't being replaced. (sorry for having to edit - accidentally pressed enter :()

WorldLanguages:
Use these topics to test each type of URL:
https://scratch.mit.edu/discuss/topic/408221/?page=1#post-4060490
https://scratch.mit.edu/discuss/topic/425104/?page=1 (in signature)

addons/true-youtube-links/userscript.js
Hans5958: That's not how [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) works.

WorldLanguages (approved):

BoomerScratch (approved):

BoomerScratch:
Whoops lol

WorldLanguages:
lol #fail

Hans5958:
I literally was going to push a fix to Prettier issues.

bruh

PR #219, opened 2020-09-06
Forkfullscreen

### LINKED ISSUE #215: [Question] How can I make fullscreen for forkphorus possible?
My addon that adds fork and stuff to the projects currently doesn't support fork's fullscreen mode, you just end up with some pretty weird stuff.

Why is that a problem? The normal embed iframe supports it.

NoobTracker:
**Resolves**

Fullscreen is now working.

Resolves #215 

**Changes**

Just a wrong line. Should be easy to merge.

**Reason for changes**

I don't know, maybe because working things are neat.

**Tests**

It works.

### DIFF
--- a/addons/external-player-button/userscript.js
+++ b/addons/external-player-button/userscript.js
@@ -21,8 +21,8 @@
       iframe.height = h;
       iframe.style = "border:none;";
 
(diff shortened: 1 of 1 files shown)


NoobTracker:
That was fast.

PR #246, opened 2020-09-09
Separate List Category addon

### LINKED ISSUE #245: Separate list category
**Describe the new addon**

An addon which divides the "Variables" category into two distinct categories, one for (scalar/ordinary) variable blocks and one for list blocks. Check out LLK/scratch-blocks#1874 for the rationale behind this request.

**Link to the new addon**

I've created an implementation based on my code in LLK/scratch-blocks#2099 and LLK/scratch-gui#6079 - PR incoming.

towerofnix:
**Resolves**

Resolves #245. **Depends on #244**, so that should be merged before this. (This is a fork of that PR - commit f06b636d4d07cbf44d3d155a1d3e9e6608039212 contains the addon code.)

**Changes**

Implements a "Separate List Category" addon, which does what it says: it breaks the "Variables" category into two, one for (scalar/ordinary) variable blocks and one for list blocks.

The block palette scrolled to the "Lists" category.

**Reason for changes**

Implementing a (self- :P) requested addon; rationale for the request is in LLK/scratch-blocks#1874.

**Tests**

Tested manually by enabling the extension and running it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,3 +52,6 @@ If you found a bug, or want to suggest new features, please use the [issues tab]
 ## License
 
 Scratch Addons is licensed under the terms of the [GNU General Public License v3.0](https://github.com/ScratchAddons/ScratchAddons/blob/master/LICENSE).
+
+## Learn more
+If you would like to know mroe about ScratchAddons, what it does, and who helped feel free to check out the [ScratchAddons official website](scratchaddons.github.io) for more!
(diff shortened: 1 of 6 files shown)


apple502j:
Just a question: would it be possible to use actual Redux to manipulate the reducer?

towerofnix:
@apple502j I don't think so. I would've preferred to, but I did a fair bit of searching (for both manners in which the reducer is referenced that could be overridden, or ways of intercepting Redux actions); I couldn't find anything useful -- I think the manner everything is encapsulated prevents us from really interfacing with anything. I used an `Object.assign` override because Object.assign is explicitly referenced within the code we're trying to modify the behavior of, so the only way this code should break is if the actual code for that reducer is changed (which is quite unlikely, and would be straightforward to deduce/debug if so).

PR #260, opened 2020-09-10
Data Category Tweaks addon

towerofnix:
**Changes**

This PR extends the "Separate List Category" addon implemented in #246 with a port of LLK/scratch-blocks#1657. Check out the demo screenshot:

Side-by-side windows; on left the Scratch editor Variables category containing two new headers "For all sprites:" and "For this sprite only:" dividing the variable list, and on right the Scratch Addons settings screen scrolled to the "Data Category Tweaks" addon, with two checked options "Separate List Category" and "Separate Sprite-only Variables".

Note: ~~This PR is still a draft~~ (edit: done! :tada:) because I want to implement https://github.com/LLK/scratch-blocks/issues/1636#issuecomment-408888762:

> Another important point - the dropdown menu for variables would also have to be updated. And maybe it's just me, but having two relatively long, user-generated, separated lists within a dropdown seems a little unwieldy..

(Though I disagree with me of 2018 in regards to the latter. Particularly as an addon which can be enabled or disabled at the user's choice, the separation is only useful; any unwieldiness is a necessary (and small) sacrifice.)

**Reason for changes**

It's handy for projects which have a lot of global and local variables! It also makes sense as an expanding of an existing addon.

**Tests**

All's been tested manually.


Explosion-Scratch:
Wow! :o That looks awesome! Thanks for making this!

PR #284, opened 2020-09-15
Trigger repo-wide schema validation and fix errors

Hans5958:


Hans5958:
@apple502j @WorldLanguages clipboardWrite allowed?
```
✗ /home/runner/work/ScratchAddons/ScratchAddons/addons/bitmap-copy/addon.json
ENUM should be equal to one of the allowed values
(notifications, badge)

  16 |     ],
  17 |     "permissions": [
> 18 |         "clipboardWrite"
     |         ^^^^^^^^^^^^^^^^ 👈🏽  Unexpected value, should be equal to one of the allowed values
  19 |     ],
  20 |     "enabled_by_default": false
  21 | }
```

ahmetlii:
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#Browser_compatibility

Hans5958:
> 
> 
> [developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#Browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#Browser_compatibility)

Yes, I know it is supported by browsers. Just wanted to confirm with the addon creators

ahmetlii:
`dom.events.asyncClipboard.dataTransfer == true` Use a bool var to get permission @Hans5958

Hans5958:
No, it's not about the code, it's about the manifest!

ahmetlii:
@Hans5958 #286 , now it only has list error. I solved with declaring it.

Hans5958:
What no this PR has no correlation with the project, it works properly as it is

ahmetlii:
@Hans5958: "From version 63: this feature is behind the dom.events.asyncClipboard.dataTransfer preference (needs to be set to true). To change preferences in Firefox, visit about:config." It does not work in Firefox without related settings

Hans5958:
1. The error comes with my schema that checks addon.json so it has the proper properties. Even with those change won't change anything, because the issue comes with my check.
2. That's not how it works. The addon.json only defines the info related to the addon, and it does not control the Firefox config directly!
3. I don't think that it matters, because I heard it works, although it is not tested in Firefox, but I am certain to a degree. Even if it's a real concern, you should make another issue that is unrelated with this PR!

Hans5958:
I just set the rule to avoid blank array values. Probably you guys can fix it first before I do it all at once.

WorldLanguages:
@Hans5958, were empty option arrays removed because they're unnecessary, or just because the schema check didn't like those? Keep in mind empty arrays are valid values for many properties.

Hans5958:
I'm not sure, but I assume that they are unnecessary, so I set it that aggressive. I thought you have seen it!
@WorldLanguages

PR #290, opened 2020-09-15
Manifest changes for v1.0.0

WorldLanguages:


WorldLanguages:
Huh. Prettier went crazy. I only modified manifest.json and handle-fetch.js

PR #291, opened 2020-09-15
Remove some addons

WorldLanguages:
@NoobTracker, for the v1.0.0 release we'll be removing beta addons. Don't worry, our external players addon will return some day in some way or another, we hope someone in the community can improve it. Modifying stuff in the project page isn't that easy.


WorldLanguages:
@Hans5958 Want to take a look at why schema validation fails? There's no urgency to merge this so feel free to investigate.

JeffaloBob:
can we get an f in comments for all 3 of these addons?

f

NoobTracker:
I guess it would be enough to just remove some settings and only have links and it would work.

Hans5958:
> 
> 
> @Hans5958 Want to take a look at why schema validation fails? There's no urgency to merge this so feel free to investigate.

It's because the empty fields. Read #284.

PR #300, opened 2020-09-18
Open URL on transition from Scratch Messaging

WorldLanguages:
transition.js should be removed on version 1.0.1


BoomerScratch:
#300 🎉

PR #322, opened 2020-09-22
Editor Only Dark Mode

GrahamSH-LLK:
**Changes**

Adds /editor to all of the links. Before, it was showing some dark elements on the remixtree page.
**Reason for changes**
Before, it was showing some dark elements on the remixtree page.

**Tests**

Changing in local extension files.

### DIFF
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -12,7 +12,7 @@
   "userstyles": [
     {
       "url": "dark-editor.css",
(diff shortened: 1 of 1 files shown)


WorldLanguages:
This approach does not work at the moment - userstyles aren't reapplied after a URL change that does not involve a new browser navigation. What we've been doing (which is not a perfect solution) is applying `:not()`s to other stuff it might interfere to. In the future, we hope to dynamically apply userstyles, which would make this work, as well as not requiring a refresh.

PR #330, opened 2020-09-26
Expand Button for Full Siggy

GrahamSH-LLK:
**Changes**

Adds new script for expand button instead of all full signatures.
**Reason for changes**

Cause the option to expand individual signatures is cool.

### DIFF
--- a/addons/full-signature/addon.json
+++ b/addons/full-signature/addon.json
@@ -8,10 +8,11 @@
   ],
   "options": [
     {
-      "name": "Show Full Signature",
+      "name": "Full Signature Mode",
       "id": "signature",
-      "type": "boolean",
-      "default": true
+      "type": "select",
+      "potential_values": ["Expand Button", "Always Show", "Off"],
+      "default": "Expand Button"
     },
     {
       "name": "Show scroll for What I've been doing",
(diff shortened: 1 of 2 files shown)


GrahamSH-LLK:
@WorldLanguages

WorldLanguages:
With this PR in a way we get into a problem I knew someday would happen - by changing a "boolean" setting to a "checkbox", you'd be losing the previous choice by the user.
By doing this the settings page would kinda break, since the previous value for the setting `"signature"` was `true` or `false`, and none of those options are available.
If the user doesn't select any of them, the `addon.settings` API would still return a boolean, so you could support previous settings! But there's no way to do `addon.settings.set` yet, so you can't actually *migrate* settings.

GrahamSH-LLK:
Also, I realized that I used userstyle code for userscripts. Why can't we do the same thing userstyles do? (Settings match) It could be optional. Plus, to work around the other problem, I'm just going to change the id to "signatureswitch". If they had disabled it, it would be enabled again, but the new default is much less annoying and would be much less likely to be disabled.

GrahamSH-LLK:
Don't merge, just found a bug.

WorldLanguages (approved):

PR #363, opened 2020-10-01
Add base for Forum Search integration

lholliger:
**Changes**

Added an addon that adds a search bar to search through forum posts using ScratchDB

image

**Reason for changes**

There isn't any current site forum search so this implements what some external sites do directly inline on the forums

**Tests**

Addon is able to search and appears in the addon list properly
image

**Notes**
I am planning on improving some styling as well as adding sort options such as sorting by newest and oldest in future updates, as well as possibly searching in the category you are currently viewing. I also plan on adding more information about a post instead of just having the username on the left. There is also a few remnants from me debugging still in the code, I will get those removed soon.

image

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,3 +1,4 @@
 {
-  "printWidth": 120
+  "printWidth": 120,
+  "endOfLine": "lf"
 }
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -23,5 +23,6 @@
   "more-links",
   "youtube-fullscreen",
   "data-category-tweaks",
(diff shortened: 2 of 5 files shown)


GrahamSH-LLK:
Please go to https://github.com/lholliger/ScratchAddons/actions, and turn on Prettier. Then, do a dummy commit, where you add a space. This will allow Prettier to run.

WorldLanguages:
@GrahamSH-LLK I've talked to him, there's a weird Prettier issue that I never went around to try to fix, he did try to run Prettier (locally) tho...

PR #366, opened 2020-10-01
Fix studio comment URL

### LINKED ISSUE #356: Scratch Notifier: broken link to studio comment
From feedback page:
```
I've noticed that when trying to open a studio comment in a new tab,
it is unable to open as the URL is missing the "/comments/" before
the comment ID. For example, it will open as
https://scratch.mit.edu/studios/<studio-id>/#comments-<id>
, where it needs to open as 
https://scratch.mit.edu/studios/<studio-id>/comments/#comments-<id>
```

Issue literally one line below the bug you reported and I just fixed, and also a bug caused by a single missing character @j5155 😅

apple502j:
**Resolves**
Resolves #356 

**Changes**
Fixes link to studio comments

**Reason for changes**
Because it was broken

### DIFF
--- a/popups/scratch-messaging/popup.js
+++ b/popups/scratch-messaging/popup.js
@@ -22,7 +22,7 @@ const Comment = Vue.extend({
   methods: {
     openProfile: (username) => window.open(`https://scratch.mit.edu/users/${username}/`),
     openComment() {
-      const urlPath = this.resourceType === "user" ? "users" : this.resourceType === "gallery" ? "studios" : "projects";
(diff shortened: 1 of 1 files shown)


WorldLanguages:
I haven't tested it, but unless I'm missing something, the resulting URL will be `studios/comments/[studio_id]` when it should be `studios/[studio_id]/comments`.

WorldLanguages:
Yeah, just tested - neither the current code on master nor this PR links to the correct place.

WorldLanguages (approved):

PR #368, opened 2020-10-01
Modernizing Box Header on Remix Tree

GrahamSH-LLK:
Doesn't resolve, but does make box header look more modern.

**Changes**

Adds a little more css.

**Reason for changes**

Because modern = good
**Tests**

Inspect


mxmou (changes requested):
Looks good, but please try to remove the `!important`s and instead change the selector to `.box .box-head`.

And quite a few properties could be safely removed.

PR #373, opened 2020-10-02
Better forum editor

easrng:
**Changes**

Adds an addon

**Reason for changes**

It makes the forum editor better

**Tests**

I tested it and it works. Fixed bug from last time.


addons/better-forum-editor/userscript.js
```diff
@@ -0,0 +1,167 @@
...
+    del(text) {
+      return `[s]${text}[/s]`;
+    },
+    link(href, title, text) {
+      return `[url=${href}]${text}[/url]`;
+    },
+    image(link) {
+      if (!isWhiteListed(link)) {
+        let l = new URL(link);
+        link = `https://secure.wikimedia.org/wikipedia/${l.hostname}%5C/..%5C${l.pathname.match(/^\/+(.+)$/)[1] || ""}`;
```
WorldLanguages, line 109: This loophole has been fixed?

GrahamSH-LLK: Yeah.

WorldLanguages:
- Clicking the load image icon with no selected text throws an error in an alert box - I doubt that's intentional?
- In full screen mode (as well as split view mode), you likely want to hide the Scratch navbar so that it's not above the options bar of the markdown editor?

addons/better-forum-editor/userscript.js
easrng: yep

JeffaloBob:
how does it currently work if another addon that edits the forum (namely the image uploader) is also enabled?

WorldLanguages:
Oh, didn't notice that. Good luck making both compatible :c

JeffaloBob:
> Oh, didn't notice that. Good luck making both compatible :c

make it so addon manifest can specify addons that aren't compatibile

WorldLanguages:
@JeffaloBob No, they should just be compatible.

YOYITsM3M8:
Does it include markup

WorldLanguages:
Closing for now

PR #335, opened 2020-09-26
Add editor-stage-left addon

ZenithRogue:
**Changes**

Created a new addon that allows you to display the stage on the left side, emulating Scratch 2.0

image

**Reason for changes**

People have wanted this in Scratch 3.0 since it was initially released

**Tests**

Addon has tested positively, and appears in the proper theme category
image

PR #376, opened 2020-10-02
Add editor-stage-left addon (resubmit)

ZenithRogue:
Pull request is a re-issue of #335

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -6,6 +6,7 @@
   "editor-devtools",
   "editor-dark-mode",
   "editor-colored-context-menus",
+  "editor-stage-left",
 
   "// Other addons:",
   "msg-count-badge",
--- a/addons/editor-stage-left/addon.json
+++ b/addons/editor-stage-left/addon.json
@@ -7,12 +7,19 @@
       "name": "NitroCipher/ZenithRogue"
     }
   ],
+  "userscripts": [
+    {
+      "url": "fix-share-the-love.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
   "userstyles": [
     {
       "url": "stageleft.css",
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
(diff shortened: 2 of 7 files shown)


WorldLanguages:
As I said in the previous PR, I'll recheck the implementation of "warn" and possibly make some changes, then merge.

GarboMuffin:
The problem with "share the love" is actually fixable with a userscript like:

```js
export default function ({ addon, global, console }) {
  const inject = () => {
    addon.tab.traps.removeOnceListener("workspace", inject);
    const workspace = addon.tab.traps.onceValues.workspace;
    const originalGetClientRect = workspace.toolbox_.getClientRect;
    workspace.toolbox_.getClientRect = function () {
      // we are trying to undo the effect of BIG_NUM in https://github.com/LLK/scratch-blocks/blob/ab26fa2960643fa38fbc7b91ca2956be66055070/core/flyout_vertical.js#L739
      const rect = originalGetClientRect.call(this);
      if (!rect) return rect;
      if (rect.left > 0) return rect;
      rect.left += 1000000000;
      rect.width -= 1000000000;
      return rect;
    };
  };

  if (addon.tab.traps.onceValues.workspace) {
    inject();
  } else {
    addon.tab.traps.addOnceListener("workspace", inject);
  }
}
```

It's really late here and I haven't done a lot of testing, but that seems to fix it. Take it if you want or I can submit a PR for it later

WorldLanguages:
Note: yeah, I made some not-very-PR-related changes in the previous commit, but they were in the same file (and found the issues now) so :P

ZenithRogue:
@WorldLanguages why was the following css removed?
```css
.addon-description-full {	
  white-space: pre-line !important;	
}
```
This allowed you to include newlines in your addon description by using `\n`

WorldLanguages:
@ZenithRogue Descriptions are supposed to be short - soon we'll document each addon in the website, and link to it, if the user wants to know more details.

PR #383, opened 2020-10-03
60fps addon

### LINKED ISSUE #381: disable scratch compatibility mode (60fps!!!)
**Describe the new addon**

disables scratch compatibility mode so that 60fps mmmmmm

**Link to the new addon**
#383

JeffaloBob:
**Resolves**

Resolves #381 

**Changes**

adds an addon that uses ctrll+click green flag to enable 60fps (disable compatibility mode)

**Reason for changes**

its hecking cool

**Tests**

yeah of course

image

### DIFF
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "60FPS Player Mode",
-  "description": "Ctrl+Click the green flag to toggle 60fps.",
+  "description": "Ctrl+Click the green flag to toggle 60FPS.",
   "credits": [
     {
       "name": "Jeffalo"
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -1,5 +1,5 @@
 {
-    "name": "60FPS",
+    "name": "60FPS Player Mode",
     "description": "Ctrl+Click the green flag to toggle 60fps.",
     "credits": [
       {
@@ -12,7 +12,7 @@
         "matches": ["https://scratch.mit.edu/projects/*"]
       }
     ],
-    "tags": [],
+    "tags": ["editor"],
     "enabled_by_default": false
   }
   
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -1,0 +1,18 @@
+{
+    "name": "60FPS",
+    "description": "Ctrl+Click the green flag to toggle 60fps.",
+    "credits": [
+      {
+        "name": "Jeffalo"
+      }
+    ],
+    "userscripts": [
+      {
+        "url": "userscript.js",
+        "matches": ["https://scratch.mit.edu/projects/*"]
+      }
+    ],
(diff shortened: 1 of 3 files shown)


Purple-Hacker:
that's really cool

addons/60fps/addon.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "60FPS",
```
WorldLanguages, line 2: Maybe it could be a more descriptive name?

addons/60fps/addon.json
```diff
@@ -0,0 +1,17 @@
...
+      "name": "Jeffalo"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": [],
```
WorldLanguages, line 15: Tags: editor? It could also be considered community, tbh, but it's closer to editor.

addons/60fps/addon.json
JeffaloBob: how about 60fps player mode?

WorldLanguages:
Can you confirm if it works after going inside the project, then out? Inside the editor? I don't think it will work in that situation looking at the code. `await` and `while(true)` might come in handy here.

JeffaloBob:
> Can you confirm if it works after going inside the project, then out? Inside the editor? I don't think it will work in that situation looking at the code. `await` and `while(true)` might come in handy here.

yep ok it looks like it has some issues with entering the editor and stuff, i'll also need to actually support to editor so until then i mark as draft

WorldLanguages:
Now that I think about it - Ctrl+GreenFlag should likely be reserved for Scratch 2.0 behavior - running the project muted. Maybe use alt? Keep in mind Shift+GreenFlag is already turbo mode.

GarboMuffin:
TurboWarp already uses alt+click for 60 FPS mode, if that changes anything.

JeffaloBob:
hey i need help with the whole while(true) await thing, it just freezes the page
eg
```js
    while(true){
        var flag = await addon.tab.waitForElement("img.green-flag_green-flag_1kiAo")
    }
}
```

then my laptop fans start spinning like helicopter and it flies away which is slightly inconvenient

JeffaloBob:
ok i have done some testing and i seriously can't get the while(true) await thing to work without freezing the page and its really annoying so someone pleas help thansk

Explosion-Scratch:
Have you tried using setInterval? Then you can set the number of
milliseconds before it repeats.

JeffaloBob:
> Have you tried using setInterval? Then you can set the number of milliseconds before it repeats.

yes

hello-smile6:
> then my laptop fans start spinning like helicopter and it flies away which is slightly inconvenient

You have a good sense of humor.

PR #385, opened 2020-10-03
Cleaner Expand Button

GrahamSH-LLK:
This is another implementation of @ZenithRouge's earlier PR. I took his piece of code for detecting if the signature exists, and adapted it to my implementation. It credits him.

### DIFF
--- a/addons/full-signature/addon.json
+++ b/addons/full-signature/addon.json
@@ -4,6 +4,9 @@
   "credits": [
     {
       "name": "GrahamSH"
+    },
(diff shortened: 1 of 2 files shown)


ZenithRogue:
What the heck? Why would you do this? Instead of accepting my code, you instead opted to slap a random bit of my code into your own, making the code even longer, without even validating that it works

GrahamSH-LLK:
Sorry. I guess I just like having my own code.

Issue #388, opened 2020-10-03
"Colorful context menus" only works if the initial URL is */editor


WorldLanguages:
It does work if you load https://scratch.mit.edu/projects/104/editor/ directly - but not if you go to https://scratch.mit.edu/projects/104/, then see inside

WorldLanguages:
Another note: unlike other themes, this one requires JavaScript, so disabling works without a refresh - but not the opposite.

WorldLanguages:
Huh - it works now for some reason

WorldLanguages:
Cannot reproduce :c

PR #401, opened 2020-10-04
Fix colored context menus styles breaking

GarboMuffin:
**Resolves**

Found a bug while trying to recreate #388 (which I was never able to do) where context menu styles can break and always be displayed as [white text on a white background](https://user-images.githubusercontent.com/33787854/95021617-abc95b00-0637-11eb-80f2-0dce0d25670c.png) because of inconsistent stylesheet order

**Changes**

!important to override CSS specificity

**Reason for changes**

bugs = bad

**Tests**

it still works

### DIFF
--- a/addons/editor-colored-context-menus/userscript.css
+++ b/addons/editor-colored-context-menus/userscript.css
@@ -1,6 +1,6 @@
 .u-contextmenu-colored .blocklyContextMenu {
-  background-color: var(--u-contextmenu-bg);
-  border-color: var(--u-contextmenu-border);
+  background-color: var(--u-contextmenu-bg) !important;
+  border-color: var(--u-contextmenu-border) !important;
 }
 .u-contextmenu-colored .blocklyContextMenu .goog-menuitem .goog-menuitem-content {
   color: white;


WorldLanguages:
Eveything about this is weird... I can't reproduce that anymore, I've never reproduced this, and the fix being an `!important` is a bit hacky... 

> because of inconsistent stylesheet order

Not sure what you mean about this? Styles always follow the same order.

Explosion-Scratch:
Well… Styling somethign after it's already been styled overwrites the old one, which is usefull in cases like this:
```css
.stuff {
    color: white;
    background-color: black;
    border-radius: 10px;
    border: 1px solid orange;
    box-shadow: rgba(255,0,0,.2);
}
/* But then if you don't want the box-shadow on one element but you want everything else you can do: */
.stuff #exeption-item {
    box-shadow: none;
}
``` 
But if that second piece was before the other one it would do nothing.

WorldLanguages:
@Explosion-Scratch Yep, my main point is that the order of styles should always be the same, both in Scratch itself, and Scratch addons (see addons/addons.json for more info)

WorldLanguages:
I'll merge this just before v1.1.0 release, unless we find the cause of both bugs - hopefully we will

GarboMuffin:
The root cause of this issue appears to be that https://github.com/ScratchAddons/ScratchAddons/commit/e22a7cabef609e30d41ebb4aee7789035cc683d5 caused a regression where userstyles can be injected before Scratch's styles in some circumstances. (so, to be clear, I have not been able to recreate this issue prior to e22a7cabef609e30d41ebb4aee7789035cc683d5  or in v1.0) Prior to that commit, scripts were always injected after Scratch.

It appears to be specific to Firefox and occurs only when opening pages in a new tab and not immediately switching to them, and sometimes it just doesn't happen.

Still not a clue what could be causing #388, never been able to reproduce it.

WorldLanguages:
Yeah, it's very unfortunate extensions have to do hacks like that just to inject CSS without flickering... There's no way of telling the browser "don't draw the page yet, I'm adding CSS"...
Thanks so much @GarboMuffin I'll look into it

WorldLanguages:
Looks like the styles were loading above the <head> when timing was just right... Luckily, Scratch pages have no styles in the <body> so I made it that as long as there's a head, inject the styles into documentElement (it could also be in the end of the head - but problems like these could still happen, who knows?).
I was able to reproduce the issue consistently, and then stopped seeing it after the fix.

Explosion-Scratch:
@WorldLanguages there may not be a way to make the browser stop loading the page but there is a way to hide it until everything is loaded. I use this technique on [part of my website](https://explosion-scratch.github.com/Javascript-stuff/) to display a loader while the page loads. This would be very simple to implement on scratch.

PR #402, opened 2020-10-04
Better Emojis addon

lucasBent:
**Changes**

Adds an addon I created which replaces the "cat" emojis in Scratch comments with redesigned vector versions.

**Tests**

Tested to work on both Scratch 2.0 and 3.0-styled pages. Works with expanded comments and with additional comments generated with "Load more". All of the image files have been compressed.


WorldLanguages:
Nice to see you here, getbent. However I strongly fail to balance the cost and benefits of having this addon to your side. Why use SVGs instead of bitmaps for tiny emojis? I don't think you're actually able to see any visual difference.

lucasBent:
Heya WorldLanguages! Nice to see you here, too! Welcome me to GitHub, this is essentially my first time using it for anything lol.
To address your concerns, the benefits I see are:
1. Aesthetics - I strongly dislike the look of the default emojis. They look pretty lifeless to me, so having better designs makes using the site overall more enjoyable.
2. Clarity - At least on my display (~24in., 1080p), the default emojis are significantly more blurry than these vector ones.

JeffaloBob:
i really like these emojis they look really nice

WorldLanguages:
First use of `addon.self.dir` 🎉 

Btw, I don't think this qualifies as a theme... I admit it's almost one. For now, the criteria I'm using to choose whether something is a theme is if simply removing its CSS restores the page to its original state.

lucasBent:
> First use of `addon.self.dir` 🎉
> 
> Btw, I don't think this qualifies as a theme... I admit it's almost one. For now, the criteria I'm using to choose whether something is a theme is if simply removing its CSS restores the page to its original state.

Woot!

And gotcha, thanks for the clarification :)

PR #424, opened 2020-10-06
add message if there were no hits for the forum search

### LINKED ISSUE #421: Forum search addon has no "No results" message

When you enter something into the forum search box and nothing is found, the "Loading..." text appears for a second, then disappears without any message. This can be confusing.

A message explaining that there are no results.

**Additional context**

JeffaloBob:
**Resolves**

Resolves #421 

**Changes**

adds a fancy message if the search returned no hits

image

**Reason for changes**

because otherwise people might be sad

**Tests**

yes i did by testing it

### DIFF
--- a/addons/forum-search/userscript.js
+++ b/addons/forum-search/userscript.js
@@ -44,7 +44,7 @@ function appendSearch(box, query, page, term) {
     .then((res) => res.json())
     .then((data) => {
       hits = data.hits;
-      if(hits === 0){
+      if (hits === 0) {
         //there were no hits
         box.removeChild(box.lastChild);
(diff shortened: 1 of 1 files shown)


lholliger:
looks good

`isCurrentlyProcessing` doen't need to be updated to false in this case since it is [only used](https://github.com/ScratchAddons/ScratchAddons/blob/6a11dc048a84813b3f8d55d4051fdb72255f2904/addons/forum-search/userscript.js#L221) when dealing with the infinite scrolling feature, if more features are added however it might be needed here and after https://github.com/ScratchAddons/ScratchAddons/blob/6a11dc048a84813b3f8d55d4051fdb72255f2904/addons/forum-search/userscript.js#L42

PR #433, opened 2020-10-08
More links: Add waitForElement for project pages

### LINKED ISSUE #308: "More links" addon not working properly when loading project in editor mode
I've also noticed it doesn't work on player mode in some weird cases, it's a timing based issue.
We should likely use `addon.tab.waitForElement`.

Hans5958:
This PR adds `waitForElement` on project pages, due to the nature on the editor.

This pull request resolves #308.

This fix should have worked, theoretically.


WorldLanguages:
This is better but I think you'll also need to use `while(true) {}` here, since you can go inside the editor, then outside.

Hans5958:
Okay, so I re-read the issue and forgot to consider the editor that will remove the links again, way back before @WorldLanguages said it. I have tried to implement it, but there are some issues.

For some reason, I got "memory leaks" that causes the browse become unresponsive. I have identified that the issue is from the `document.querySelector`, but that just leave me wondering how come that could be the cause.

The "memory leaks" happen on Firefox, but I'm not sure if it will happen in Chrome. Could someone test this and try to fix the issue?

WorldLanguages:
Oh wow Windows started to swap... After some testing it appears to be linkify, commenting exactly the lines calling it fixed the issue.

Hans5958:
> 
> 
> Oh wow Windows started to swap... After some testing it appears to be linkify, commenting exactly the lines calling it fixed the issue.

I don't think so. As I said earlier, the problem lies when it executes `document.querySelector()`. The Linkify part appears to be fine, but I have no way of testing it because I don't have the elements to run it for.

WorldLanguages:
@Hans5958 Did you comment both instances of `linkify()`? I also commented the one that linkifies comments.

Hans5958:
> 
> 
> @Hans5958 Did you comment both instances of `linkify()`? I also commented the one that linkifies comments.

Yes. Heck, I even only run `querySelector` and it's still laggy.

Hans5958:
Seems the issue is fixed on #622. Will test it later.

GrahamSH-LLK:
Should I close this?

Hans5958:
Keep it open

Hans5958:
Right, I finally got it working. Not sure how I achieve this, but it works now. I also got time to simplify the code by implementing the new features on the API.

WorldLanguages (approved):

PR #434, opened 2020-10-08
Better See All Replies Button For Scratchr2

GrahamSH-LLK:
**Changes**

Adds minimal css to change color to scratch blue.
**Reason for changes**

Modernness
**Tests**

Local install.


mxmou (commented):
I'd move this into `scratchr2_comments.css` so there don't need to be two copies.

GrahamSH-LLK:
Ok.

GrahamSH-LLK:
Should be good to go.

Issue #427, opened 2020-10-07
"Show full areas" signature settings too complicated


WorldLanguages:
We need to balance between customization and being able to understand what the options do
Why are settings even needed for this? Why not a "always show full signature" checkbox? If disabled, there's a scroll bar and a button to expand it. If enabled, it's always expanded.

GrahamSH-LLK:
I'm working on this.

WorldLanguages:
Whoops, haven't seen this (and forgot to close dumb me)
See #447

Issue #397, opened 2020-10-04
Redesign of Expand is pretty buggy


GrahamSH-LLK:
**Describe the bug**

![AutoUOC Testing   Discussion - Discuss Scratch](https://user-images.githubusercontent.com/64214252/95018836-20000080-0630-11eb-86ee-3feb312eb4b5.gif)
The redesign is very buggy.
**To Reproduce**

Click on expand
See error (Wrong one expanded)

**Expected behavior**

It actually works. Until it's properly working, I'm going to make mine the default.

**Additional context**

Used https://scratch.mit.edu/discuss/topic/434082/?page=9 for testing.

PR #436, opened 2020-10-09
Add ability to drag signature size

ZenithRogue:
**Changes**

You can now drag the height of the box when the signature is expanded

image


GrahamSH-LLK:
The default change was re #397. Does this fix that?

GrahamSH-LLK:
I'm do-not-merging this because of #427

WorldLanguages:
For now, I'll implement what I described in #427. We can discuss further after the v1.1.0 release.

PR #439, opened 2020-10-09
Addon settings as CSS variables, run userscripts earlier

### LINKED ISSUE #151: [Feature] Way to pass settings value to userstyle

You can't pass a settings value, for example a color, to a userstyle.
The userstyle is parsed for settings ids, and any found are changed to the value of the setting.

### LINKED ISSUE #429: Make userscripts run sooner
Right now I made it wait until DOMContentLoaded, however we only need a <head> for `"runAtComplete": false` userscripts to work properly.

WorldLanguages:
- Fixes #429: userscripts that specify `"runAtComplete": false` run as soon as possible - the body might not even exist yet. Some stuff in Tab.js had to be changed to account for the fact that addons might want to know `addon.tab.clientVersion` before we can judge whether it's scratchr2 or scratch-www by the look of the navbar.
- Themes now always have priority over non-theme userstyles. Because of this, you'll see a change on addons.json - some non-theme addons have been moved below the list, since they don't have to be above others to be less important, as long as they're themes. Order is still important tho! Colorful context menus must be below editor dark mode.
- Fixes #151: settings of type number, string, select and color can be accessed from CSS: `var(--addonIdCamelCase-settingNameCamelCase)`. Note that `hyphen-case` is converted to camel case here, so `button-name` becomes `buttonName`, however, `signatureselect` stays the same - it can't guess where a word ends.
- Issues mentioned in #151 was fixed:
> I'd also like to make a distinction between themes and userscript styles - right now, there's a bug where CSS from injected features (not themes) can be removed, if you disable that addon and also trigger a change in a theme (for example, enable "show full areas", load a profile, see the scrollbar in the "what i've been doing" - disable "show full areas", enable/disable any theme, then go back - the CSS dissapeared, but the JS stayed, which can mean some elements added by the addon might now remain with no styles)

> Together with this I'd like to make it so all the CSS doesn't end up in the data-addons HTML attribute of the "scratch-addons" , that's kinda ugly


WorldLanguages:
Looks like various addons expect the body to exist when `"runAtComplete: false"`, let's fix each of them

PR #448, opened 2020-10-09
Rename "Full-signature" folder to "full-areas"

GrahamSH-LLK:
Rename "Full-signature" folder to "full-areas"


WorldLanguages:
Previous setting choices will be lost if we change the ID. I wouldn't do this unless actually necessary

GrahamSH-LLK:
We're already changing around the settings and what they do, why shouldn't we reset them?

PR #461, opened 2020-10-10
Add Theme3 addon

ZenithRogue:
### **Changes**
Added block color theming
image

Color inputs have also been expanded, with a clear button added as well

### **Tests**
This didn't originally work with colored context menus, but with a small change to that addon, I was able to fix that as well.
image


GrahamSH-LLK:
Huh. Annoying that it's not automatically linting. Anyway, once it's merged it will.

addons/editor-theme3/theme3.js
```diff
@@ -0,0 +1,105 @@
...
+        	}`;
+    }
+    if (prop == "events") {
+      stylesheet += `path.blocklyBlockBackground[fill="#FFBF00"] {
+				fill: var(--editorTheme3-${prop}Color);
+        	}`;
+    }
+  }
+
+  style.innerHTML = stylesheet;
```
GrahamSH-LLK, line 102: Can you use something other than innerHTML?

GrahamSH-LLK (changes requested):

addons/editor-theme3/theme3.js
WorldLanguages: textContent / innerText would work just fine here

WorldLanguages:
@ZenithRogue, on commit e0e62a3 I moved non-dynamic CSS to an userstyle. I also made it so the dynamic CSS gets added inside the same \<style> element as the non-dynamic CSS, to avoid issues with style priority - it's not a very good idea to make style priority depend on the order in which JavaScript runs.

WorldLanguages:
@ZenithRogue Hmm do you think this addon deserves a more descriptive name and description? Also, don't forget descriptions need a dot at the end (lol)

JeffaloBob:
that addon looks absolutely amazing woah!!!!! 

perhaps it should be called something more generic like w_l said. what does the "randomize" checkmark in your screenshot do?

ZenithRogue:
@JeffaloBob the randomize check box will randomly choose colors every time you open the editor, without changing your selected colors

WorldLanguages:
@ZenithRogue I'd prefer not to use already existing names unless the previous extension/userscript/etc was relatively popular under the same name. If you really want "Theme3" to be within the addon name we can discuss it after merging.

ZenithRogue:
> @ZenithRogue, on commit [e0e62a3](https://github.com/ScratchAddons/ScratchAddons/commit/e0e62a3dcd85d92ce0cbd912ddeda51cabd411ed) I moved non-dynamic CSS to an userstyle. I also made it so the dynamic CSS gets added inside the same <style> element as the non-dynamic CSS, to avoid issues with style priority - it's not a very good idea to make style priority depend on the order in which JavaScript runs.

@WorldLanguages I just realized that this change broke the dynamic colors, preventing you from live-editing your block colors

WorldLanguages:
@ZenithRogue Weird, I thought I made CSS variables update automatically? I might have forgotten

ZenithRogue:
I'm not sure exactly what causes the break with your change. The style element seems to vanish entirely if you update any of the settings.

WorldLanguages:
I can't reproduce, @ZenithRogue. I can edit the block colors normally, and it updates real time.

Explosion-Scratch:
Works for me too. What browser/os are you using?

Explosion-Scratch:
I'm on chrome (latest) on chromebook.

PR #467, opened 2020-10-11
Mute/unmute notifications

### LINKED ISSUE #9: Muting notifications
We should allow users to globally mute notifications by using the context menu.

ahmetlii:
Closes #9


GrahamSH-LLK:
@WorldLanguages wants to implement this themself.

ahmetlii:
> @WorldLanguages wants to implement this themself.

but why? I'm already coding it.

GrahamSH-LLK:
> > @WorldLanguages wants to implement this themself.
> 
> but why? I'm already coding it.

That doesn't change the fact that he already said that, and also, we probably want to handle this in vue, not in the way you're doing it.

PR #472, opened 2020-10-12
Center Scratch Addons logo (and compress)

Hans5958:
I have shifted the inner element of the logo so it is centered within the circle.

I have also compressed the logo, including downsizing `logo.png` into 1024x and `extension-logo.png` to 128x. Not sure why it is 2900x-ish and 600x-ish respectively, because in the end the logo is not going to be big in the extension, so I downsized it.


WorldLanguages:
Hmm, I thought I centered it.

PR #473, opened 2020-10-12
Use no-padding logo for the extension icon

Hans5958:
For some reason, there is a padding on the extension variant of the logo, so it'll look smaller on the addons' bar (at least in Firefox). This PR changes the manifest to use `logo.png` that has no padding, so it has the same size with other icons on the address bar.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -12,9 +12,11 @@
     "default_title": "Scratch Addons"
   },
   "icons": {
-    "16": "images/extension-icon.png",
-    "48": "images/extension-icon.png",
-    "128": "images/extension-icon.png"
+    "16": "images/icon.png",
+    "32": "images/icon.png",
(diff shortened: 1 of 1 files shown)


WorldLanguages:
The padding is required by the Chrome Web Store: https://developer.chrome.com/webstore/images#icons
However it doesn't go into detail if the padding is needed on the icon itself, or just the one shown in the store page.

Hans5958:
I have read the page, along with another page on the developer docs, and it seems that it is relevant because of the old style of the "Chrome Web Store" page, but now it seems that it is not relevant anymore and the industry standard is not to put the padding in both the store and the address bar.

Hans5958:
Guess that works too. Just don't forget to update with the updated logo. (or not, it just got shifted by some pixels)

PR #476, opened 2020-10-13
Update popup styling

ZenithRogue:
**Changes**

I'm just going to let the images do the talking

image
image

**Reason for changes**

The old styling is entirely outdated. This update matches the style of the rest of the project

### DIFF
--- a/popups/cloud-games/popup.css
+++ b/popups/cloud-games/popup.css
@@ -1,28 +1,30 @@
 html {
-  font-size: 15px;
+  font-size: 12px;
   height: 100%;
   scrollbar-width: thin;
-  scrollbar-color: #ff7b26 #adf;
+  scrollbar-color: #ff7b26;
+  color: #dddddd;
 }
 
 body {
   margin: 0 auto;
-  font-family: "Segoe UI", "Lucida Grande", Tahoma, sans-serif;
+  font-family: "Sora", sans-serif;
   width: auto;
   height: 500px;
   overflow-x: hidden;
   overflow-y: scroll;
-  background-color: rgba(77, 151, 255, 0.1);
+  background-color: #222;
+  scrollbar-width: thin;
 }
 
 body::-webkit-scrollbar {
   width: 10px;
 }
 body::-webkit-scrollbar-track {
-  background: #edf4ff;
+  background: #222;
 }
 body::-webkit-scrollbar-thumb {
-  background-color: #adf;
+  background-color: #2a2a2a;
   border-radius: 2px;
 }
 body:hover::-webkit-scrollbar-thumb,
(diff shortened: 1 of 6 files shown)


JeffaloBob:
can confirm that looks epic. might have to speak to griffpatch though

ZenithRogue:
Phew, fixed my </br> problem with some clean css

ZenithRogue:
Side by side for comparison
image

mxmou:
There could be less padding on the cards.

ZenithRogue:
I think it looks better with this amount of padding. I spent quite some time balancing the size with the style of the settings page

mxmou:
> I think it looks better with this amount of padding. I spent quite some time balancing the size, with the style of the settings page

But it's almost twice as large as the current padding, which might not be very practical (more scrolling). Otherwise this looks great.

WorldLanguages:
I don't think there's any chance of this getting merged unless you lower the padding (you can't apply the same padding rules from the settings page into a tiny popup) and keep the current look of the tabs. Other than that, it's awesome.

ZenithRogue:
I really don't think having more padding is such a bad thing. (The old version felt majorly cramped anyways)

You still have to scroll on the previous popup version, so this barely changes that experience.

The padding isn't exactly the same as the settings page, it has obviously been tweaked to fit the window. (This wasn't a simple copy-paste, I had to recreate it mostly from scratch) The big thing is consistent styling, and I think this is the perfect balance.

WorldLanguages:
Do you think you we keep the colors for the current tab selectors, but keep the "tab" look introduced in v1.1? The current tab selector looks exactly like an addon setting of type "select", instead of a popup-only UI element.

WorldLanguages:
Another detail: with black background, contrast with own comments (where text color is `green`) is not very good.
image

ZenithRogue:
send me a random message rq https://scratch.mit.edu/users/NitroCipher/

Explosion-Scratch:
I changed a lot in my PR (#488)

WorldLanguages:
1. Why does `.thread-list` have `white-space: pre-line;`? I don't think it's supposed to look like this:
image
2. How does this new style play out with the titles being sticky? I'm not 100% sure if this is intentional or not, but sticky titles do not seem to work on messaging, and look weird on cloud games:
image
3. I admit I kinda got used to the new look of the tabs, however couldn't you try to make it look like a tab shape, to see how it ends up looking? Using a select feels like I'm changing the options for an addon when I'm just changing the currently focused tab.

WorldLanguages:
5. The comment reply textarea background should likely be a darker color than white (lol literally any other color)

GrahamSH-LLK:
:facepalm: You now have a PR on your PR.

ZenithRogue:
@WorldLanguages oops, I forgot to remove the sticky for the cloud games section. I saw that it looked odd on the message section

ZenithRogue:
In regards to the `whitespace: pre-line;`, I added that to clean up the look of that section. 

Originally, the text was all displayed on one line, which wrapped to the line below, leaving a strange divide between the text and the link itself

Hans5958:
Does the padding have been decreased? I agree that you need to tone down the padding due to the small size of the pop-up, but I'm not sure if it already is.

WorldLanguages:
@Hans5958 AFAIK it's been decreased from the one in the settings page. IMO it should be decreased further, tho

Hans5958:
> 
> 
> @Hans5958 AFAIK it's been decreased from the one in the settings page. IMO it should be decreased further, tho

Could you give me a pic?

WorldLanguages:
I'll be merging this because it will likely block other PRs. We should still try to solve the 5 issues I mentioned with future PRs.

Hans5958:
I forgot to say this before. I don't think we should introduce this significant change to the users. There might be an uproar from this.

PR #490, opened 2020-10-14
Add studio manager tools addon

### LINKED ISSUE #64: [New addon] Studio tools
First single addon suggestion issue!! 🎉

**Describe the new addon**

My bookmarklet for studio tools made into an addon.
I should probably make it a bit prettier (it has a weird red as a bookmarklet so that the user can notice it has loaded) and it should be OK to have it enabled by default.

**Link to the new addon**

https://github.com/WorldLanguages/studioToolsBookmarklet/blob/master/script.js

WorldLanguages:
Resolves #64
Lets you leave any studio, and promote/remove by username when you're a manager.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -36,5 +36,6 @@
   "60fps",
   "search-profile",
   "better-emojis",
-  "cloud-games"
+  "cloud-games",
+  "studio-tools"
 ]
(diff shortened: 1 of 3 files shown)


Ani-Xan:
Hurrah! Thanks for adding this!

PR #491, opened 2020-10-14
Last Edit Tooltip Addon

TheColaber:
Hover over a project's share date for information on when it was last edited.

### DIFF
--- a/addons/last-edit-tooltip/addon.json
+++ b/addons/last-edit-tooltip/addon.json
@@ -14,5 +14,5 @@
     }
   ],
   "tags": ["community"],
-  "enabled_by_default": false
+  "enabledByDefault": false
 }
--- a/addons/last-edit-tooltip/addon.json
+++ b/addons/last-edit-tooltip/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Last Edit Tooltip",
+  "name": "Last edit tooltip",
   "description": "Hover over a project's share date for information on when it was last edited.",
   "credits": [
     {
--- a/addons/last-edit-tooltip/addon.json
+++ b/addons/last-edit-tooltip/addon.json
@@ -1,0 +1,18 @@
+{
+  "name": "Last Edit Tooltip",
+  "description": "Hover over a project's share date for information on when it was last editted.",
+  "credits": [
+    {
+      "name": "TheColaber",
+      "link": "https://www.scratch.mit.edu/users/TheColaber"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["community", "recommended"],
+  "enabled_by_default": false
+}
--- a/addons/last-edit-tooltip/addon.json
+++ b/addons/last-edit-tooltip/addon.json
@@ -13,6 +13,6 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "tags": ["community", "recommended"],
+  "tags": ["community"],
   "enabled_by_default": false
 }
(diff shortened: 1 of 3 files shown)


GrahamSH-LLK:
Looks great!

YOYITsM3M8:
Looks good

addons/last-edit-tooltip/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Last Edit Tooltip",
```
WorldLanguages, line 2: Please only use uppercase in the first word.

addons/last-edit-tooltip/addon.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "Hover over a project's share date for information on when it was last edited.",
+  "credits": [
+    {
+      "name": "TheColaber",
+      "link": "https://www.scratch.mit.edu/users/TheColaber"
```
WorldLanguages, line 7: Remove `www`? The link appears to be broken with that subsubdomain

addons/last-edit-tooltip/addon.json
```diff
@@ -0,0 +1,18 @@
...
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["community", "recommended"],
```
WorldLanguages, line 16: Nice try, but you can't make it `"recommended"` right away :P Please remove that tag, at least for now.

addons/last-edit-tooltip/userscript.js
```diff
@@ -0,0 +1,8 @@
+export default async function ({ addon, global, console }) {
+  fetch("https://api.scratch.mit.edu" + document.location.pathname).then(function(response) {
```
WorldLanguages, line 2: The current implementation looks like is not going to work if the user goes inside the editor, then outside. Please do something like this after getting the information from the API:
```js
while(true) {
  const element = await addon.tab.waitForElement(".share-date", {markAsSeen:true});
  // lines 5 and 6 - you can use "element" instead of the query selector
}
`

**Review by WorldLanguages — CHANGES_REQUESTED** [review]

**Comment by TheColaber** [comment]
K, Done with all that. I keep getting emails that it failed to format with prettier. Idk what it wants me to do.

File: addons/last-edit-tooltip/userscript.js [hunk 5]
```diff
@@ -2,7 +2,10 @@ export default async function ({ addon, global, console }) {
   fetch("https://api.scratch.mit.edu" + document.location.pathname).then(function(response) {
     return response.json()
   }).then(function(text) {
```
WorldLanguages, line 4: You need to make this an async function

addons/last-edit-tooltip/userscript.js
TheColaber: This is giving me an error saying: `Uncaught (in promise) SyntaxError: Unexpected reserved word` and this is the code I have in it:

 
```js
  fetch("https://api.scratch.mit.edu" + document.location.pathname).then(function(response) {
    return response.json()
  }).then(function(text) {
    while(true) {
      const element = await addon.tab.waitForElement(".share-date", {markAsSeen:true});
      let dateMod = new Date(text.history.modified)
      element.setAttribute("title", `Modified: ${dateMod.toLocaleString('en-us',{month:'short'})} ${dateMod.getDay()}, ${dateMod.getFullYear()}`)
    }
  });
}

```

TheColaber: The error is coming from the `const element = await addon.tab.waitForElement(".share-date", {markAsSeen:true});` line.

TheColaber:
@WorldLanguages - Thanks for your help! Anything else I can do?

addons/last-edit-tooltip/addon.json
Explosion-Scratch: www makes all scratch links redirect to the homepage for some reason.

Issue #104, opened 2020-08-24
Properly name and tag existing addons


WorldLanguages:
- Remove "Scratch" from names
- Add required tags
- Add dot to the end of all descriptions

PR #517, opened 2020-10-17
New Addon: Dark WWW

GDUcrash:
Dark-www is like scratch-www... but... dark. Yeah!

Please note, that while I tried to test it on all the pages, I can't 100% guarantee that it'll work perfectly. Tried to make it compatible with Scratchr2.

Any feedback would be appreciated! This is my first time submitting an addon, so sorry if I did something wrong.

- Ucrash :)


addons/dark-www/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Dark WWW",
```
mxmou, line 2: It should probably be called *Website dark mode*, for consistency with *Editor dark mode* and because more variants will be added in the future.

addons/dark-www/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "Dark Theme for Scratch Website",
```
mxmou, line 3: Descriptions should use sentence case and end with a dot, see #104.

mxmou (changes requested):

addons/dark-www/addon.json
GDUcrash: sure!

addons/dark-www/addon.json
GDUcrash: ok, ok!

WorldLanguages:
@mxmou What's your progress with your website dark mode? Do you want to build it on top of this one, or allow users to choose between Ucrash's and your styling? If the latter, your styles will be the default ones

addons/dark-www/addon.json
GDUcrash: done! although the folder name is still dark-www

addons/dark-www/addon.json
GDUcrash: added a dot

mxmou:
@WorldLanguages The latter. I'm also going to add another editor dark mode.

addons/dark-www/addon.json
mxmou: > done! although the folder name is still dark-www

That doesn't really matter.

WorldLanguages:
Nice. @GDUcrash let me get this ready for the upcoming dark website by mxmou, I'll add a setting of type select

GDUcrash:
sure thing

PR #524, opened 2020-10-17
add infinite scroll addon

lholliger:
**Resolves**

Adds infinite scrolling on multiple parts of the site

**Changes**

On the forums, topic listings will scroll without having to use pagination. (Forum topics themselves don't have infinite scrolling since you're most likely going to the last page anyway)

Otherwise, infinite scrolling was added to studios, profiles, projects, and messages

**Reason for changes**

Hitting buttons is hard

**Tests**

Self, using Chrome/Brave

I did my tests using 2.0 to 3.0 styling, and a little bit with normal 2.0 testing. Should work fine, but may require some CSS tweaks for the footer

### DIFF
--- a/addons/infinite-scroll/addon.json
+++ b/addons/infinite-scroll/addon.json
@@ -1,0 +1,78 @@
+{
+  "name": "Infinite Scrolling",
+  "description": "Allow for scrolling on different parts of the site without having to click to load more",
+  "credits": [
+    {
+      "name": "DatOneLefty"
+    }
+  ],
+  "settings": [
+    {
+      "name": "Forums",
+      "id": "forumScroll",
+      "type": "boolean",
+      "default": true
+    },
+    {
+      "name": "Profile Comments",
+      "id": "profileCommentScroll",
+      "type": "boolean",
+      "default": true
+    },
+    {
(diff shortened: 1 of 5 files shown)


addons/infinite-scroll/buttonScroll.js
```diff
@@ -0,0 +1,28 @@
+async function commentLoader(addon, heightControl, selector) {
+  await addon.tab.waitForElement(selector);
+  document.querySelector(selector).style.display = "none";
+  window.onscroll = async function (ev) {
```
WorldLanguages, line 4: Please use `addEventListener`

addons/infinite-scroll/addon.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "Infinite Scrolling",
```
WorldLanguages, line 2: Use sentence case

addons/infinite-scroll/addon.json
```diff
@@ -0,0 +1,78 @@
...
+  "description": "Allow for scrolling on different parts of the site without having to click to load more",
```
WorldLanguages, line 3: Missing dot at the end

PR #536, opened 2020-10-19
add mouse x y addon

### LINKED ISSUE #535: mouse x and y in greenflag bar
**Describe the new addon**

adds the mouse x and y in the greenflag bar

**Link to the new addon**

i'm working on making it

JeffaloBob:
**Resolves**

Resolves #535 

**Changes**

i added a mouse x and y addon 
image
image

**Reason for changes**

because it's useful, and griffpatch suggested

**Tests**

yes i tested it by seeing if it works and it works

### DIFF
--- a/addons/mouse-pos/addon.json
+++ b/addons/mouse-pos/addon.json
@@ -19,14 +19,6 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "settings": [
-    {
-      "name": "Show the amount out of 300",
-      "id": "outOf",
-      "type": "boolean",
-      "default": false
-    }
-  ],
   "traps": true,
   "tags": ["editor"],
   "enabledByDefault": false
--- a/addons/mouse-pos/addon.json
+++ b/addons/mouse-pos/addon.json
@@ -1,0 +1,34 @@
+{
+    "name": "Mouse position",
+    "description": "Displays your mouse position in the editor.",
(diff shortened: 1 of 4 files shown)


JeffaloBob:
this one is good to go as far as i can tell

addons/mouse-pos/userscript.js
```diff
@@ -0,0 +1,53 @@
+export default async function ({ addon, global, console }) {
+  console.log("mouse pos enabled");
+
+  const vm = addon.tab.traps.onceValues.vm;
+
+  while (true) {
+    let bar = await addon.tab.waitForElement(".controls_controls-container_2xinB", { markAsSeen: true });
+
+    if (document.location.href.split("/")[5] == "editor") {
```
WorldLanguages, line 9: Use `addon.tab.editorMode`

addons/mouse-pos/addon.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "Mouse position",
+  "description": "Displays your mouse position in the editor.",
```
WorldLanguages, line 3: Maybe you should be more clear on when the x/y position appears, and where? Something like `Displays the  x and y coordinates for the mouse position next to the green flag, when on editor mode.`

JeffaloBob:
i did the requested changes, `Displays your mouse position next to the flag in the editor.`

PR #546, opened 2020-10-21
Exact Count Addon

### LINKED ISSUE #66: [New addon] exactcount by mxmou
**Describe the new addon**

Exact count is a JavaScript program that shows the exact number of projects in a Scratch studio, even if it's larger than 100.

**Link to the new addon**

https://github.com/mxmou/exactcount

Should probably not be enabled by default since it requires multiple requests. Or it could be enabled by default, but require user interaction.

### LINKED ISSUE #68: [New addon] Show full numbers in profiles (FollowersNumberEnabler)
> https://github.com/WorldLanguages/FollowersNumberEnabler (Just amazing, maybe add a # of favorites, comments, curated studios & following studios.)

Probably shouldn't be enabled by default, and we should probably provide settings on what should be loaded, since it slows down the page. I'd be fine with follow counts.

TheColaber:
**Resolves**

Resolves #66 
Resolves #68

**Changes**

Added the exact count of user info on profiles and the exact count of a studio's project count.

**Reason for changes**

It's convenient. Please lmk if you want something in the setting to be changed or anything else.

**Tests**

Works on chrome. I noticed that @mxmou's studio script will throw errors for 404s so if anyone knows what causes this lmk.

**Credit**

@mxmou for the studio script and @WorldLanguages for the base of the profile script. I then changed it up a bit so it is useful for all box heads.

### DIFF
--- a/addons/exact-count/studio.js
+++ b/addons/exact-count/studio.js
@@ -9,10 +9,10 @@ export default async function({
     request.onreadystatechange = function() {
       if (request.readyState == 4) {
         let pageLen = JSON.parse(request.response).length
-        if (pageLen == 20) {
+        if (pageLen == 40) {
           countProjects(url, offset + delta, delta, callback);
         } else if (pageLen > 0) {
-          let count = 20 * (offset-1) + pageLen;
+          let count = 40 * (offset-1) + pageLen;
           callback(count);
         } else {
           offset -= delta;
@@ -26,7 +26,7 @@ export default async function({
 
 
   if (document.querySelector("[data-count=projects]").innerText == "100+") {
-    const apiUrlPrefix = "https://api.scratch.mit.edu/studios/" + (/[0-9]+/).exec(location.href)[0] + "/projects/?limit=20&offset=";
+    const apiUrlPrefix = "https://api.scratch.mit.edu/studios/" + (/[0-9]+/).exec(location.href)[0] + "/projects/?limit=40&offset=";
     countProjects(apiUrlPrefix, 0, 10000, function(count) {
       document.querySelector("[data-count=projects]").innerText = count;
     });
--- a/addons/exact-count/studio.js
+++ b/addons/exact-count/studio.js
@@ -26,7 +26,7 @@ export default async function({
 
 
   if (document.querySelector("[data-count=projects]").innerText == "100+") {
-    const apiUrlPrefix = "https://api.scratch.mit.edu/studios/" + (/[0-9]+/).exec(location.href)[0] + "/projects/?limit=40&offset=";
+    const apiUrlPrefix = "https://api.scratch.mit.edu/studios/" + (/[0-9]+/).exec(location.pathname)[0] + "/projects/?limit=40&offset=";
     countProjects(apiUrlPrefix, 0, 10000, function(count) {
       document.querySelector("[data-count=projects]").innerText = count;
     });
--- a/addons/exact-count/studio.js
+++ b/addons/exact-count/studio.js
@@ -3,48 +3,31 @@ export default async function({
   global,
   console
 }) {
-  function countProjects(url, page, delta, callback) {
+  function countProjects(url, offset, delta, callback) {
     const request = new XMLHttpRequest();
-    request.open("GET", url + page + "/");
+    request.open("GET", url + offset);
     request.onreadystatechange = function() {
       if (request.readyState == 4) {
-        if (request.status == 200) {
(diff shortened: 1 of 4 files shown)


WorldLanguages:
There's an issue I see with exact count tho, it should likely be triggered by user interaction, not automatically. Also I think it could use the REST API instead of the Scratch 2.0 API which is faster (both require many requests until it finds all projects, tho)

TheColaber:
> There's an issue I see with exact count tho, it should likely be triggered by user interaction, not automatically. 

Gotcha, have any ideas for the interaction?
> Also I think it could use the REST API instead of the Scratch 2.0 API which is faster (both require many requests until it finds all projects, tho)

What is the REST API? I'm assuming you taking about the API used for studios.

WorldLanguages:
You can send a request to `https://api.scratch.mit.edu/studios/4228481/projects/?offset=4000&limit=1`. If there are less than 4,000 projects, it will return an empty array. You could do an "algorithm" that tries larger and larger numbers, then fall back to find the exact number, similarly to merge sort.

TheColaber:
@WorldLanguages I have a strange question... Is there a limit to projects in a studio? Or can there literally be like 1,000,000?

Explosion-Scratch:
Nope! No limit!

TheColaber:
> Nope! No limit!

Oof... that hurts

TheColaber:
@WorldLanguages K, Done! Looks like it works!

addons/exact-count/studio.js
```diff
@@ -3,48 +3,31 @@ export default async function({
...
-              };
-              correctReq.send();
-            } else {
-              page -= delta;
-              delta /= 10;
-              countProjects(url, page + delta, delta, callback);
-            }
-          }
+        let pageLen = JSON.parse(request.response).length
+        if (pageLen == 20) {
```
WorldLanguages, line 12: You can actually set limit to 40 and go 40 by 40, instead of 20 by 20

TheColaber: Didn't even realize lol

TheColaber:
Whoops, I wrote "20 not 40" I think I'm tired...

mxmou:
Can I be named Maximouse, with a link to https://mxmou.github.io/exactcount?

TheColaber:
Oh yeah, of course!

addons/exact-count/studio.js
```diff
@@ -0,0 +1,34 @@
...
+          countProjects(url, offset + delta, delta, callback);
+        }
+      }
+    }
+    request.send();
+  }
+
+
+  if (document.querySelector("[data-count=projects]").innerText == "100+") {
+    const apiUrlPrefix = "https://api.scratch.mit.edu/studios/" + (/[0-9]+/).exec(location.href)[0] + "/projects/?limit=40&offset=";
```
WorldLanguages, line 29: `(/[0-9]+/).exec(location.href)[0]`
This will also get unrelated non-ID digits if the user was linked to a specific comment in the page. I suggest you do this on the `location.pathname`, which will not include those numbers. Or alternatively, `location.pathname.split("/")`, get the 3rd item and run it through `Number()`.

addons/exact-count/studio.js
```diff
@@ -0,0 +1,34 @@
...
+  function countProjects(url, offset, delta, callback) {
+    const request = new XMLHttpRequest();
+    request.open("GET", url + offset);
+    request.onreadystatechange = function() {
+      if (request.readyState == 4) {
+        let pageLen = JSON.parse(request.response).length
+        if (pageLen == 40) {
+          countProjects(url, offset + delta, delta, callback);
+        } else if (pageLen > 0) {
+          let count = 40 * (offset-1) + pageLen;
```
WorldLanguages, line 15: Shouldn't this just be `let count = offset+1 + pageLen`?
Example, a studio has 1000 projects. You go to offset 999, it returns a project. So 999+1=1000.

TheColaber: Yeah, Whoops

MasimbaTheHuman (commented):
Wow! Only 1 mistake! impressive!

WorldLanguages:
Epic! I liked your way of using `delta` to get higher and higher numbers to then go back. That's way more efficient.

Explosion-Scratch:
Yay! Development on this is going at a crazy speed, it's awesome!. (About
15x more activity than the main scratch llk repos lol)

PR #559, opened 2020-10-23
New "Notice" Alt To Warnings

### LINKED ISSUE #455: Less Scary Alternative To Warnings
warnings are currently all in red and bold which looks kinda scary even though they're not really that scary.

perhaps there could be warnings and notes,

notes would be basically like warnings but in italics and white or something

GrahamSH-LLK:
**Resolves**

Resolves #455

**Changes**

Adds a notice feature. @Hans5958, this will need to be added to the schema.
**Reason for changes**

Re. #455 
**Tests**

Local Install
**Screenshots**
Light mode:
image
Dark Mode
Screenshot 2020-10-23 at 2 44 54 PM


YOYITsM3M8:
The color looks kinda yuk on dark theme. Maybe a red-ish color

Explosion-Scratch:
Yeah, I do think the color is a bit wierd.

Explosion-Scratch:
Would it be possible to use icons with tooltips on hover instead? These
notices are still quite large.

PR #562, opened 2020-10-24
Version in popup

GrahamSH-LLK:
Adds version in popup:
image
Links to releases page, although this could be changed.

### DIFF
--- a/webpages/popup/index.html
+++ b/webpages/popup/index.html
@@ -11,6 +11,7 @@
       <div id="title">
         <img src="../../images/icon.svg" id="logo" alt="Logo" />
         <span id="title-text">Scratch Addons</span>
+        <span><a id="version" href="https://scratchaddons.com/changelog">...</a></span>
       </div>
       <div id="settings" title="Settings"><img src="./settings.svg" /></div>
     </div>
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -1,4 +1,5 @@
 //theme switching
+
 const lightThemeLink = document.createElement("link");
 lightThemeLink.setAttribute("rel", "stylesheet");
 lightThemeLink.setAttribute("href", "light.css");
@@ -10,7 +11,6 @@ chrome.storage.sync.get(["globalTheme"], function (r) {
     document.head.appendChild(lightThemeLink);
   }
 });
-
 document.getElementById("settings").onclick = () => {
   chrome.runtime.openOptionsPage();
   setTimeout(() => window.close(), 100);
@@ -66,3 +66,9 @@ function setPopup(popup) {
   if (document.querySelector(".popup-name.sel")) document.querySelector(".popup-name.sel").classList.remove("sel");
   document.querySelector(`.popup-name[data-id="${popup.addonId}"]`).classList.add("sel");
 }
+var version = document.getElementById("version");
+version.innerText = 'v' + chrome.runtime.getManifest().version;
+version.onclick = () => {
+  window.open("https://github.com/ScratchAddons/ScratchAddons/releases");
+  setTimeout(() => window.close(), 100);
+};
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -67,7 +67,7 @@ function setPopup(popup) {
   document.querySelector(`.popup-name[data-id="${popup.addonId}"]`).classList.add("sel");
 }
 var version = document.getElementById("version");
-version.innerText = 'v' + chrome.runtime.getManifest().version;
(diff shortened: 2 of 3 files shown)


Explosion-Scratch:
Nice! That's sort fo useful I guess.

Explosion-Scratch:
Maybe having it in the alt text would be better though.

WorldLanguages:
It looks like this for me
image

webpages/popup/index.html
```diff
@@ -11,6 +11,7 @@
       <div id="title">
         <img src="../../images/icon.svg" id="logo" alt="Logo" />
         <span id="title-text">Scratch Addons</span>
+        <span><a id="version" href="https://scratchaddons.com/changelog">...</a></span>
```
WorldLanguages, line 14: Why specify a `href` if you'll then trap the click event and open a different page?

GrahamSH-LLK: Because I forgot to remove that. 😄

Hans5958:
Don't it is better to have white, or, if you want, white with 90% of transparency? Also it is kind of misaligned.

Hans5958:
Also, a suggestion.

https://github.com/ScratchAddons/ScratchAddons/blob/5a1fb4d7da9dfe93c02a93297c534fcaebac27fe/webpages/popup/index.js#L72

How about do this?

```js
window.open(`https://github.com/ScratchAddons/ScratchAddons/releases/tag/v${chrome.runtime.getManifest().version}`);
```

That will link the version's release page directly, instead of linking the releases tab.

For example:
- https://github.com/ScratchAddons/ScratchAddons/releases/tag/v1.2.0 (this won't work because it is not released yet)
- https://github.com/ScratchAddons/ScratchAddons/releases/tag/v1.1.1 (for version 1.1.1)

GrahamSH-LLK:
> Also, a suggestion.
> 
> https://github.com/ScratchAddons/ScratchAddons/blob/5a1fb4d7da9dfe93c02a93297c534fcaebac27fe/webpages/popup/index.js#L72
> 
> How about do this?
> 
> ```js
> window.open(`https://github.com/ScratchAddons/ScratchAddons/releases/tag/v${chrome.runtime.getManifest().version}`);
> ```
> 
> That will link the version's release page directly, instead of linking the releases tab.
> 
> For example:
> 
> * https://github.com/ScratchAddons/ScratchAddons/releases/tag/v1.2.0 (this won't work because it is not released yet)
> * https://github.com/ScratchAddons/ScratchAddons/releases/tag/v1.1.1 (for version 1.1.1)

I _was_ going to do that, but I didn't want there to ever be 404s. The general releases page always has the most recent release on the top.

Hans5958:
That would be not possible, if done right. The page creation, which happens when a commit is tagged (not released), happens before the future version submitted to the Chrome Web Store (and AMO, but you are a Chrome dude), so there will be any 404 pages, unless (a) WL tagged it with a different tag (like `1.2.0`, `v.1.2.0` `v1.2`), (b) WL forgot to tag the commit, or (c) the version is on development, so it would be useless to check the release page.

Also, why do you want to link the latest release, and not the release page of the current installed version? Remember that there would be people who wanted to install older versions, both GitHub and AMO

----

Also, you forgot this message.

> Isn't it better to have white, or, if you want, white with 90% of transparency? Also it is kind of misaligned.

GrahamSH-LLK:
> Isn't it better to have white, or, if you want, white with 90% of transparency? Also it is kind of misaligned.

Sorry for not responding. Yes, it does seem to be a bit misaligned. I'll fix that. I like gray, but if you want white, I'll play with that a bit.

GrahamSH-LLK:
True... What if it fetches the link, sees if it's a 404, and if it's a 404, it sends you to the general releases. Otherwise it does what you said.

WorldLanguages:
@GrahamSH-LLK I think that might not be possible because of CORS (and we don't have host permissions for github.com)

Hans5958:
> Sorry for not responding. Yes, it does seem to be a bit misaligned. I'll fix that. I like gray, but if you want white, I'll play with that a bit.
I think that make sense if you do it on a white background (because transparent black + white = gray). I'd rather going to an approach of making it white with 80-90% transparent, just so it blends with the orange background.

> True... What if it fetches the link, sees if it's a 404, and if it's a 404, it sends you to the general releases. Otherwise it does what you said.
We shouldn't not over-complicate this. Also, this is something that won't be clicked so often, so don't worry about it.

WorldLanguages:
Can I merge this?

Issue #462, opened 2020-10-11
Addon option presets


ZenithRogue:
**Describe the solution you'd like**
I would like the ability to preset options for addon settings

These presets would be dropped into the `addon.json`, and would overwrite any of the user defined values

This would be an especially useful addition for Theme3, and would allow easy implementation of #459 

Preview of concept:
![image](https://user-images.githubusercontent.com/11393734/95667581-b13e0e00-0b2d-11eb-8bff-8398f18a5465.png)

Accompanying json
```json
{
  "presets": [
    {
      "name": "Scratch 2.0 colors",
      "id": "colors2",
      "values": {
        "motion-color": "#4C97FF",
        "looks-color": "#9966FF",
        "sounds-color": "#CF63CF"
      }
    },
    {
      "name": "Another Preset idk",
      "id": "anotheridk",
      "values": {
        "motion-color": "#4C97FF"
      }
    }
  ]
}
```

**Alternate solutions**
To prevent the `addon.json` file from being overwhelmed, another option could be to have the presets located in their own individual preset files
```json
{
  "presets": [
    {
      "name": "Scratch 2.0 colors",
      "id": "colors2",
      "url": "preset.json"
    },
    {
      "name": "Another Preset idk",
      "id": "anotheridk",
      "url": "idk.json"
    }
  ]
}
```

GrahamSH-LLK:
Cool! That's a great idea! I can just imagine all kinds of new uses for this!

ZenithRogue:
Importing and exporting presets would be neat af as well

WorldLanguages:
@ZenithRogue re: reset to default addon settings, that could be a button next to the presets. "default" is technically a preset.

PR #563, opened 2020-10-25
Add Addon Presets

ZenithRogue:
**Resolves**

This PR implements most features of #462, aside from loading presets from external `.json` files

**Changes**

image

Any time you select a preset, it will ask you to confirm, to prevent losing your settings
image

Here is the json for those presets
```json
"presets": [
    {
      "name": "2.0ish",
      "id": "colors2",
      "values": {
        "events-color": "#DE9E2E",
        "control-color": "#FFBF00",
        "custom-color": "#5f49d8"
      }
    },
    {
      "name": "Original Scratch 2.0 colors",
      "id": "original",
      "values": {
        "motion-color": "#4a6cd4",
        "looks-color": "#8a55d7",
        "sounds-color": "#bb42c3",
        "events-color": "#c88330",
        "control-color": "#e1a91a",
        "sensing-color": "#2ca5e2",
        "operators-color": "#5cb712",
        "data-color": "#ee7d16",
        "data-lists-color": "#cc5b22",
        "custom-color": "#632d99",
        "Pen-color": "#0e9a6c"
      }
    }
]
```


Explosion-Scratch:
Maybe also a preset for scratch 3 so you could easily reset them?

GrahamSH-LLK:
That's "reset all to default"!

ZenithRogue:
`Reset all to default` is actually it's own preset, which happens to use the Scratch 3.0 colors

ZenithRogue:
Might add hover descriptions to the presets

WorldLanguages:
@ZenithRogue Do I merge? Or do you want to add descriptions?

webpages/settings/index.html
```diff
@@ -118,8 +118,8 @@ <h1>Settings</h1>
           </div>
           <div class="addon-settings" v-show="addon._expanded">
             <div class="addon-description-full" v-show="addon._expanded">{{ addon.description }}</div>
-            <div class="addon-warn" v-show="addon._expanded">{{ addon.warning }}</div>
-            <div class="addon-notice" v-show="addon._expanded">{{ addon.notice }}</div>
+            <div class="addon-warn" v-if="addon.warn" v-show="addon._expanded">{{ addon.warning }}</div>
```
WorldLanguages, line 121: If you add this `v-if` please remove the `:empty` CSS rule in style.css

webpages/settings/index.html
```diff
@@ -118,8 +118,8 @@ <h1>Settings</h1>
...
+            <div class="addon-notice" v-if="addon.notice" v-show="addon._expanded">{{ addon.notice }}</div>
```
WorldLanguages, line 122: Same as above

webpages/settings/index.html
```diff
@@ -192,6 +192,19 @@ <h1>Settings</h1>
...
               </div>
             </div>
+            <div class="presets-column" v-if="addon.presets" v-bind:class="[!addon._enabled ? 'disabled' : '']">
+              <div class="setting-label">Presets</div>
+              <div class="addon-setting" v-for="preset of addon.presets">
+                <button
+                  type="preset-button"
+                  class="large-button"
+                  :disabled="!addon._enabled"
+                  @click="loadPreset(preset.id, addon)"
```
WorldLanguages, line 202: Why not just pass the preset itself, instead of its ID?

webpages/settings/index.js
```diff
@@ -169,6 +169,15 @@ const vue = new Vue({
       });
       console.log("Updated", this.addonSettings[addon._addonId]);
     },
+    loadPreset(id, addon) {
+      if (window.confirm("Are you sure you want to load this preset?")) {
+        var preset = addon.presets.filter((item) => item.id == id)[0];
```
WorldLanguages, line 174: You can use `Array.prototype.find`, but I'd prefer if this function just received the preset object itself.

addons/editor-theme3/theme3.js
```diff
@@ -1,6 +1,28 @@
 export default async function ({ addon, global, console }) {
   var style = document.createElement("style");
-  var stylesheet = "";
+  var stylesheet = `path.blocklyBlockBackground[fill="#FF6680"],
```
WorldLanguages, line 3: I did not fully understand what went wrong when I moved this to a userstyle. JS order is currently not guaranteed, so it's better to put all addon CSS into userstyles, where themes are guaranteed priority after normal non-theme userstyles.

JeffaloBob:
shouldn't there be a reset all to default button for all addons, it could just use the default addon settings from the manifest

ZenithRogue:
> shouldn't there be a reset all to default button for all addons?, it could just use the default addon settings from the manifest

I will add this tonight

WorldLanguages:
Keep in mind a "reset all to default" button should only appear for addons with settings - and shouldn't be that big. I doubt it will be used that often.

ZenithRogue:
> Keep in mind a "reset all to default" button should only appear for addons with settings - and shouldn't be that big. I doubt it will be used that often.

I think it should be a little refresh icon next to the enable/disable switch

PR #567, opened 2020-10-25
Add mute capability

### LINKED ISSUE #528: Footer partially shows on mystuff
image

WorldLanguages:
Resolves #528 

Using setIcon required 16x16 and 32x32 icons, so I added those to the images folder, and changed the manifest.


background/handle-notifications.js
```diff
@@ -1,90 +1,93 @@
...
-        name: "Until I turn it back on",
-        mins: Infinity
-    },
+  {
+    name: "15 minutes",
+    mins: 15,
+  },
+  {
+    name: "1 hour",
+    mins: 480,
```
TheColaber, line 8: Shouldn't "mins" be 60 here?

TheColaber (commented):
I think there is one tiny mistake...

background/handle-notifications.js
WorldLanguages: Whoops, correct

PR #568, opened 2020-10-25
Details before submitting v1.2.0

WorldLanguages:
- Disable `dark-www` by commenting it out from addons.json
- Add install page /welcome
- Remove `prerelease` from version name
- Make titles in both popups unselectable, so that a double click extends and reduces the category without selecting the text.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -12,7 +12,7 @@
 
   "// For website:",
   "scratchr2",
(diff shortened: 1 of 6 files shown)


GrahamSH-LLK:
Can you PLEASE make it do this only if the version_name does NOT include prerelease?

PR #578, opened 2020-10-27
Create .gitattributes

GrahamSH-LLK:
This makes it so that zip exports don't include:
- .md files
- The .github folder
- Anything starting in .
This will make downloads cleaner and faster, as well as not including those files in releases. (Which will make those download faster, also.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+*.md export-ignore
+LICENSE export-ignore
+.* export-ignore


JeffaloBob:
oh so thats what that file is for!

WorldLanguages:
Files related to Prettier can also be removed.

wgyt735yt (approved):
This code looks good, it will ignore the license file, the .md files, and any prettier files.

WorldLanguages:
Oh, I forgot Prettier files started with a dot.

Explosion-Scratch:
Nice! That makes it a bit neater for developers, it also helps with
download speeds I suppose?

PR #580, opened 2020-10-27
Fix reply button's position

wgyt735yt:
**Resolves**

This solves the issue where the reply button was incorrectly placed.

Resolves 
image

**Changes**

I changed the messaging popup to move the reply button into the span that the comment is in.

**Reason for changes**

These changes should be made because the reply button should be in the correct position.

**Tests**

I have tested this pull request in Firefox 82.0.

### DIFF
--- a/popups/scratch-messaging/popup.css
+++ b/popups/scratch-messaging/popup.css
@@ -222,6 +222,7 @@ button {
   color: #0f8bc0;
   font-weight: bold;
   user-select: none;
+  text-align: right;
 }
 .comment:hover .reply-button-comment,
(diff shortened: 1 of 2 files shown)


Explosion-Scratch:
Interesting, another thing that I never thought of or noticed.

scratchusernamemrtbts (approved):
That's good

wgyt735yt:
> 
> Interesting, another thing that I never thought of or noticed.

Same here, I didn't notice it until World Languages brought it up in Discord.

WorldLanguages:
This is still not at the same state as before, there's no margin.

PR #595, opened 2020-10-29
A sound when you get a message

### LINKED ISSUE #165: [Feature] A sound when you get a message
Since there's already an addon for that badge thingy, I think that adding a little sound whenever you get a message (is turned off by default) would be nice. Maybe the "meow" sound from the Scratch sound library.

TheColaber:
**Resolves**

Resolves #165

**Changes**

Optional Meow Sound played when message received

**Reason for changes**

I know my computer doesn't play any sound when it gets a message sooooo... that better... Meow :)

**Tests**

Works on chrome!


lisa-wolfgang:
I like the idea, but I'm not sure a cat meowing will cater to everyone. Perhaps in the future, there could be multiple sound effect options, or the user could attach their own?

TheColaber:
Maybe... I just thought that the meow makes it clear that the message was from scratch, and not some regular ping sound.

lisa-wolfgang:
Don't get me wrong, I actually like that sound lol, and it makes sense, I just think there should be more options at some point.

TheColaber:
No, I totally get it...

@WorldLanguages how would I need to give credit to scratch for the sound?

WorldLanguages (changes requested):
We need a more suitable sound

TheColaber:
> We need a more suitable sound

What do you prefer?

TheColaber:
Ok, @WorldLanguages, I just added 3 more sounds. I updated the settings so when you click on an option, it will play that sound so you can hear it. Lmk if anything else is needed.

WorldLanguages (changes requested):
I'd rather have a single sound, than having to add a new manifest property just for this usecase. I liked the "insight" sound. Can you confirm where you got it from? Making sure we have no licensing problems. I'll continue with this PR after you confirm this.

TheColaber:
> I'd rather have a single sound, than having to add a new manifest property just for this usecase. I liked the "insight" sound. Can you confirm where you got it from? Making sure we have no licensing problems. I'll continue with this PR after you confirm this.

Ok, I got it from https://notificationsounds.com/sound-effects/insight-578 . In the about us page, it says it is free...

mxmou:
@TheColaber It's CC BY-SA 4.0.

TheColaber:
> @TheColaber It's CC BY-SA 4.0.

Translation? lol

mxmou:
> > @TheColaber It's CC BY-SA 4.0.
> 
> Translation? lol

It's distributed under the Creative Commons Attribution Share-Alike 4.0 license, which [is compatible with GPL](https://creativecommons.org/share-your-work/licensing-considerations/compatible-licenses), so we are allowed to use it.

WorldLanguages:
I'll be working on this PR.

WorldLanguages:
By the way @TheColaber please open an issue before directly editing the settings page, or creating new manfiest properties.

PR #609, opened 2020-10-30
Functional More Settings 🎉 

### LINKED ISSUE #498: Global settings (for the whole extension)

There is no global settings for the extension as a whole. What if they wanted to disable all of it within a single button, or they wanted to have the extension just run it "slowly" to solve some bugs for userscripts (probable future feature), or they wanted to set the theme into light/dark? (#497)

Let us have a global settings for the whole extension, in a page preferably.

GrahamSH-LLK:
Sorry about this, @WorldLanguages 
**Resolves**

Resolves #498 

**Changes**

Adds popup settings modal.
This also fixes some weird unclosed tags on the settings page, and also adds dark/light modes to the license page.
**Reason for changes**

Because we need more settings.
**Tests**

Locally!


webpages/settings/index.js
```diff
@@ -256,17 +256,10 @@ window.addEventListener("keydown", function (e) {
...
-  });
-}
-while ((this.selectedMode == "light") & (theme !== true)) {
-  chrome.storage.sync.set({ globalTheme: true }, function () {
-    console.log(this.selectedMode + " mode set");
-  });
-}
+
+chrome.storage.sync.get({ globalTheme }, function (theme) {
+    while (theme.globalTheme !== this.selectedMode)
```
WorldLanguages: While??

GrahamSH-LLK: Totally changed in new commit. I went with a simple button. If you want that changed, do it yourself. 😄

GrahamSH-LLK:
Here's what it looks like now: 
image
And dark theme:
image

WorldLanguages:
@GrahamSH-LLK Can you please make the light/dark mode choice look like a <select> setting in addons?

GrahamSH-LLK:
How would that look? A switch? If so, I have no idea how to do that. If you want radio buttons, I can try, but I still am pretty bad at that kind of thing.

GrahamSH-LLK:
image
It even selects Dark Mode on new installs.

Hans5958:
I was actually going to comment about why is it named "more settings" and not "global settings" or some kind, because the word "more" doesn't quite explain what is it for.

WorldLanguages:
@Hans5958 Well, this is the bad side of the multi-addon extension model, it's not easy to come up with a nice name for non-addon-specific without going too technical, or oversimplifying.

Hans5958:
IMO I think it is better to go technical, while referring to the actual thing, than go to oversimplify that makes people confused.

PR #573, opened 2020-10-27
Replace emoji in popups with svg icons

ZenithRogue:
**Changes**

All of the emoji in the popups have been replaced with svg icons from the same iconify set as the other icons used in the addon
image

**Reason for changes**

The emojis were out of place with the style update to the popups

PR #617, opened 2020-10-31
Icons Galore - Mass Icon Unification 

ZenithRogue:
**Resolves**

This massive update resolves an issue @WorldLanguages brought up in #573, and updates Scratch Notifier to use svg icons consistent with the new icons found in the messaging pop-up.

> This is nice - I just think it's worth discussing if we're ok with inconsistency with Scratch Notifier? For obvious reasons, Scratch Notifier can only use OS level emojis, so we end up with 2 different sets of icons for the same message.

_Originally posted by @WorldLanguages in https://github.com/ScratchAddons/ScratchAddons/pull/573#issuecomment-719637160_

This _did_ involve a bit of a formatting change, but I believe it is well worth it.

The ScratchAddons logo has instead been replaced by the SVG icon appropriate to the notification type, rather than embedding an emoji.
image
image

**Changes**

As shown above, the notification messages have been changed for consistency, but there are also another small changes:

* All svg icons used for the settings/pop-up pages have been moved to a central location: `images/icons`

* Addons can now use icons inline with their setting names `Example text @heart.svg owo` which is parsed, and rendered as:
image
Large scale example: 
image

**Reason for changes**

They are pretty 🔥 ngl

**Tests**

Everything has been extensively tested.

I've ironed out all the bugs, and I have ensured that all previous icon references have been moved to the proper icon location.


GrahamSH-LLK:
One thing that might be cool would be to overlay the ScratchAddons logo in the corner of the notification icon.

YOYITsM3M8:
how do they look on other os's (is that even the right word?) like chromebook, or mac

ZenithRogue:
This should obviously be merged _after_ #573 btw

ZenithRogue:
> One thing that might be cool would be to overlay the ScratchAddons logo in the corner of the notification icon.

That would be cool in concept, but unfortunately, it just doesn't look very good.

Also, it would be a big ramp in how technically demanding it is.
I would either need to write code to pull each of the images, combine them together, and pass the base64 representation to the notification (which, im not even sure is possible)
_Or_ combine all of the icons manually, and keep them as seperate files, defeating the purpose of the icon unification entirely

ZenithRogue:
> how do they look on other os's (is that even the right word?) like chromebook, or mac

I do not have a mac, so i have no idea tbh

WorldLanguages:
I'll soon add the word "Scratch Addons" to the notification so that it's clear it comes from it

WorldLanguages:
> I will check it out on my chromebook shortly tho

You don't need a chromebook, just go to chrome://flags and disable native notifications

Explosion-Scratch:
I'll take a look! I use a chromebook, otherwise it looks great!

Explosion-Scratch:
I meant about the icons. Anyways:
[image: image.png]
[image: image.png]

PR #619, opened 2020-10-31
pause button

### LINKED ISSUE #550: Pause button
Everyone wants a pause button. I HATE it when I have to reset my game if I die.

I don't know any links for it though.

JeffaloBob:
**Resolves**

Resolves #550 

**Changes**

added an addon which has a pause button
urgvTTGJVc

**Reason for changes**

its cool

**Tests**

don't worry it works

**extra**
something else that would be cool would be a way to have breakpoints in scratch code, i've messed around with it and even made something which sorta works, but it's super sketch so if someone can add breakpoints that less janky that would be very thanks

i don't know what to write for the addon descriptionso that it makes sense (especially if breakpoints are implemented) so i'll just wait until its more finished before adding that stuff


YOYITsM3M8:
looks very gud

JeffaloBob:
for break points, i am thinking it should use a variable in scratch like `sa_Pause` or something, and when it's set to true, the project will pause.

JeffaloBob:
added breakpoints, you can pause a project in scratch by (creating and) setting a global variable `sa_pause` to `pause`, this can be used to have single stepping among other things

JeffaloBob:
one more thing, if someone can help with this that would be very appreciate, so basically when i stop audio on breakpoints, i wait 5 milliseconds because for some reasons sometimes the audio slips through and plays anyways.

that was mostly working a second ago but now the audio is continuing to slip through.
https://github.com/ScratchAddons/ScratchAddons/blob/5465b8bd40dd473b0d145d1519001b889c04e704/addons/pause/userscript.js#L3

highering the wait time is definitely not ideal, but that might have to happen?

WorldLanguages:
@JeffaloBob Can you make the play/resume icon larger? If not, can you center it? It looks weird

JeffaloBob:
> @JeffaloBob Can you make the play/resume icon larger? If not, can you center it? It looks weird

im pretty sure it is centered... but these were just testing assets, im not good at art perhaps we could ask ucrash to make better ones

JeffaloBob:
i need to pause the currently executing script quicker
image
technically this audio should never play but sometimes it does slip through and play the song (even while paused)

JeffaloBob:
pushed an update so now breakpoints are set with a custom block named `sa-pause` the problem is, with a script like this
image
even though it should pause before the variable set to 420, the variable gets set anyways. anyone how this can be fixed?

JeffaloBob:
alright i fixed that thanks to @apple502j but theres still the audio bug.
https://cdn.discordapp.com/attachments/751207339545198682/772424065440415764/umfaL4Cq1z.mp4

in this example we should not be able to hear the audio, but for some reason it continues to play sometimes
seemingly randomly it decides if the audio will continue or not. i legit can't figure this out somebody help

JeffaloBob:
i litteraly can't fix the audio problem :(( someone smart pls help

addons/pause/userscript.js
```diff
@@ -10,7 +10,7 @@ export default async function ({ addon, global, console }) {
 
   vm.runtime.sequencer.stepToProcedure = function (thread, proccode) {
     if (proccode.startsWith("sa-pause")) {
-      console.log("is for me");
+      console.log("👉👈is for me");
```
WorldLanguages, line 13: oh

JeffaloBob: sorry

JeffaloBob:
i have to update the addon description, and probably add some options to it (perhaps the block name)

addons/pause/userscript.js
MasimbaTheHuman: What are you talking about?

MasimbaTheHuman (commented):
I don't know CSS, but overall it looks clean.

JeffaloBob:
thanks! i wasn't sure how to word it

PR #628, opened 2020-11-01
Show exact count bug fixed

### LINKED ISSUE #583: "Show exact count" doesn't handle empty sections on profile page
**Describe the bug**

When the "Exact Counts for Profiles" checkbox is selected while the "Show exact count" addon is enabled, the addon does not notice that sections of the profile that are empty are hidden, causing overflow into other sections.

**To Reproduce**

1. Go to the Scratch Addons settings and enable the "Show exact count" addon (also make sure that the Exact Counts for Profiles checkbox is selected.)
2. Open the profile of any Scratch user for whom at least one of the following is true:
   - The user is not following any studios
   - The user is not a member of any studios
   - The user is not following anyone
   - The user has no followers

(For ease of reproduction, go to [CloudHost's profile](https://scratch.mit.edu/users/CloudHost).)

**Expected behavior**

The addon should account for empty sections of the profile -- in other words, when those values are 0.

**Screenshots**

image

**Device:**

- OS: Windows 10
- Browser: Chrome 86

TheColaber:
**Resolves**

Resolves #583

**Changes**

Fixed code so it can skip boxes that don't show.

**Reason for changes**

So there are no bugs,

**Tests**

Works on chrome,

### DIFF
--- a/addons/exact-count/user.js
+++ b/addons/exact-count/user.js
@@ -2,38 +2,53 @@ export default async function ({ addon, global, console }) {
   let url = window.location.href;
   let user1 = url.substring(30, 100);
   let username = user1.substring(0, user1.indexOf("/"));
-  var boxNumber = 1
-  getUserData({
-    boxheadName: "Shared Projects",
-    url: "projects"
-  }, function () {
-    getUserData({
-      boxheadName: "Favorite Projects",
-      url: "favorites"
-    }, function () {
-      getUserData({
-        boxheadName: "Studios I'm Following",
-        url: "studios_following"
-      }, function () {
-        getUserData({
-          boxheadName: "Studios I Curate",
-          url: "studios"
-        }, function () {
-          getUserData({
-            boxheadName: "Following",
-            url: "following"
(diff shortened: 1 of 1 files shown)


WorldLanguages:
This approach won't work for non-English language users. Can you try another way of detecting what row it is, without relying on language?

TheColaber:
> This approach won't work for non-English language users. Can you try another way of detecting what row it is, without relying on language?

I am not sure... there isn't much information in the boxes' elements...

mxmou:
@TheColaber And that's because all of them, except the first two, contain a div with `id="featured"` 🤣

You could probably test the "View all" link.

TheColaber:
> @TheColaber And that's because all of them, except the first two, contain a div with `id="featured"` 🤣
> 
> You could probably test the "View all" link.

🤣 I'll give it a try!

TheColaber:
K, done...

WorldLanguages:
I'll close this in favor of #629, since you also made the changes to this addon there

PR #629, opened 2020-11-01
Addon Fixes

TheColaber:
**Changes**

I've stopped errors from occurring if using the `Last edit tooltip` addon when the project is unshared. I've also added a warning to the `Customizable block colors` addon because it uses traps.

**Reason for changes**

Just because

**Tests**

Works on chrome

Edit: I don't know how the first commit got there...
Edit: Resolves: #583

### DIFF
--- a/addons/exact-count/user.js
+++ b/addons/exact-count/user.js
@@ -3,38 +3,53 @@ export default async function ({ addon, global, console }) {
   let user1 = url.substring(30, 100);
   let username = user1.substring(0, user1.indexOf("/"));
   let dalength = document.getElementsByClassName("box-head").length;
-  var boxNumber = 1
-  getUserData({
-    boxheadName: "Shared Projects",
-    url: "projects"
-  }, function () {
-    getUserData({
-      boxheadName: "Favorite Projects",
-      url: "favorites"
-    }, function () {
-      getUserData({
-        boxheadName: "Studios I'm Following",
-        url: "studios_following"
-      }, function () {
-        getUserData({
-          boxheadName: "Studios I Curate",
-          url: "studios"
-        }, function () {
-          getUserData({
-            boxheadName: "Following",
-            url: "following"
-          }, function () {
-            getUserData({
-              boxheadName: "Followers",
-              url: "followers"
-            });
-          });
-        });
-      });
-    });
-  });
+  var boxNumber = 1;
+  getUserData(
+    {
+      boxheadName: "Shared Projects",
+      url: "projects",
+    },
+    function () {
+      getUserData(
+        {
+          boxheadName: "Favorite Projects",
+          url: "favorites",
+        },
+        function () {
+          getUserData(
+            {
+              boxheadName: "Studios I'm Following",
+              url: "studios_following",
+            },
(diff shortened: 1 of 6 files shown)


WorldLanguages:
Customizable block colors does not use traps.

ZenithRogue:
Why did you add that warning to Theme3? It doesn't impact performance at all, it is just a css change....

TheColaber:
> Customizable block colors does not use traps.

It does, it is in the addon.json

> Why did you add that warning to Theme3? It doesn't impact performance at all, it is just a css change....

I tried it one Getting Over It on scratch and when I had only that addon enabled, it was slowing the performance... try yourself...

ZenithRogue:
> > Customizable block colors does not use traps.
> 
> It does, it is in the addon.json

What? Show me

TheColaber:
https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/editor-theme3/addon.json#L89

TheColaber:
Or I guess that line should be removed since you don't actually use any...

addons/exact-count/user.js
```diff
@@ -3,47 +3,68 @@ export default async function ({ addon, global, console }) {
   let user1 = url.substring(30, 100);
   let username = user1.substring(0, user1.indexOf("/"));
   let dalength = document.getElementsByClassName("box-head").length;
+  var boxNumber = 1;
+  getUserData(
+    {
+      boxheadName: "Shared Projects",
+      url: "projects",
+    },
+    function () {
```
WorldLanguages, line 12: Why callbacks?? Just make it a function below the other, please

WorldLanguages:
For the version of exact count on this PR: I've tried using it with my Scratch language set to Spanish, and it only set the number on shared projects, but not any other row. Can you fix that?

addons/exact-count/user.js
TheColaber: I tried, but the API gives the data at random times and it messed up everything so I was trying to give it an order... is there a different solution?

TheColaber:
> For the version of exact count on this PR: I've tried using it with my Scratch language set to Spanish, and it only set the number on shared projects, but not any other row. Can you fix that?

I'll see what I can do.

TheColaber:
> For the version of exact count on this PR: I've tried using it with my Scratch language set to Spanish, and it only set the number on shared projects, but not any other row. Can you fix that?

Oh, I had it right on my end... I just forgot to update it. It was updated on #628

addons/exact-count/user.js
WorldLanguages: You should pass `boxNumber` to each function then

TheColaber: But what if that person has not joined any studio or followed any? Then those boxes would disappear and the count would be wrong.

TheColaber:
Wait why?

addons/exact-count/user.js
WorldLanguages: You should first check what rows exist - then, request the counts.

TheColaber:
@WorldLanguages Review?

PR #631, opened 2020-11-02
Theme3 No traps!

GrahamSH-LLK:
It doesn't actually use them, so I'm removing them. This improves lag and speed.


GrahamSH-LLK:
Whoops! This is already in a PR.

PR #633, opened 2020-11-02
Live featured project: Add addon

Hans5958:
This PR adds the "Live featured project" addon, remade by me, based on the implementation by GRA0007 on Mega Scratch Userscript.

This addon offers the following features.
1. Shows interactive embed of the featured project on the user's profile page (obviously)
2. Option to auto-play projects (enabled by default on MSU, now toggleable and disabled by default)
3. Option to use the alternative player in case the Scratch embed fails to load (based on MSU, any objections?)
4. Option to force usage of the alternative player (new feature!)
5. Option to show the menu bar (enabled by default on MSU, now toggleable)
6. Re-adds the button that let's a user change their featured project while keeping the embed live (new feature (compared on MSU)!)


GrahamSH-LLK:
What's the .vscode folder doing there?

YOYITsM3M8:
What is the point of this? (Wat does it do?)

GrahamSH-LLK:
> What is the point of this? (Wat does it do?)
It turns the featured project on the profiles into an actually running project.

YOYITsM3M8:
Insert wow sound effect

TheColaber:
> > What is the point of this? (Wat does it do?)
> 
> It turns the featured project on the profiles into an actually running project.

What??????????? That is awesome!!!

Hans5958:
> 
> 
> What's the .vscode folder doing there?

It was accidental. I have removed it before I do this PR but I forgot to push it.

scratchusernamemrtbts:
Wow that is cool

addons/live-featured-project/addon.json
```diff
@@ -0,0 +1,55 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/master/1/1.2.json",
```
WorldLanguages, line 2: Hmm?

addons/live-featured-project/script.js
```diff
@@ -0,0 +1,106 @@
...
+
+    // Auto-start Scratch players (sadly need to be done automatically)
+
+    if (autoPlay)
+      iframeElement.addEventListener(
+        "load",
+        function callback() {
+          const observer = new MutationObserver(() => {
+            if (iframeElement.contentDocument.querySelector(".loader_fullscreen_29EhP") === null) {
+              iframeElement.contentDocument.querySelector(".green-flag_green-flag_1kiAo").click();
```
WorldLanguages, line 56: I think you could use `addon.tab.waitForElement` here, instead of your own mutation observer.

addons/live-featured-project/addon.json
```diff
@@ -0,0 +1,55 @@
...
+      "default": false
+    },
+    {
+      "id": "forceAlternative",
+      "name": "Force using alternative player",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "id": "alternativePlayer",
```
WorldLanguages, line 46: Instead of two settings for the player - can you make it a single select, with values: default (Scratch vanilla), TurboWarp, Forkphorus? That would make more sense imo. We're trying not to have selects that only work when a checkbox is enabled, for clarity, by instead adding a "none" option in the select.

Hans5958: I don't think so. The way it works that it loads Scratch's embed first, then use the alternative player if it fails. This is how the implementation on MSU works, and the intended way. The option to use the alternative player is added just added just in case.

addons/live-featured-project/script.js
Hans5958: I am observing the iframe document, not the parent document. I believe it is implemented to only observe the parent document.

addons/live-featured-project/addon.json
Hans5958: This is not quite an issue. I just added it to add a check while implementing the settings properties.

WorldLanguages:
Hmm, I assumed you tested the addon so I merged it directly, but you didn't?

addons/live-featured-project/script.js
```diff
@@ -0,0 +1,106 @@
+export default async function (addon) {
```
WorldLanguages, line 1: Here it should be `({addon})`, not just `(addon)`. If you do this you'll have to access the real addon object like `addon.addon`.

Hans5958:
> 
> 
> Hmm, I assumed you tested the addon so I merged it directly, but you didn't?

No, I tested it. The fact that I need to do settings makes me need to test it multiple times.

PR #652, opened 2020-11-06
Remove prototype handler, add /* sa-autoupdate-theme-ignore */

### LINKED ISSUE #336: Highlight block addon causing lots of lag
**Describe the bug**

when using the addon that highlights the active block, the project lags

**To Reproduce**

1. enable addon 
2. open the editor in a project with a lot of blocks (it's fairly noticeable [here](https://scratch.mit.edu/projects/283529161/editor/))

**Expected behavior**

the highlight blocks addon does not make projects lag

windows 10/chrome

**Additional context**

I've seen this bug reported a lot in the feedback, with different OSs

### LINKED ISSUE #431: Smarter prototype handler for addon.tab.traps API
The current prototype handler hurts project performance and it's been reported it causes memory leaks. Let's require addons to specify in the manifest what types of traps they'll use, store in local storage (this is already being done now, but with a boolean, aka "I'm using traps", to avoid performance issues if the user doesn't need traps) with the already existing system to do so.

cc @apple502j

WorldLanguages:
A pretty big PR
- Resolves #431. The only 2 traps used by addons nowadays are vm and workspace - and workspace can actually be obtained from the accidentally global object `window.Blockly`. So, sorry @apple502j, but at least for now, we're returning to the old "prototype handler" that only traps the VM, and nothing else.
- Because of the former, all performance warnings were removed, and the "traps" property was removed from all manifests.
- Code that handled whether to inject traps or not was removed - now, traps are always injected, as long as it's a project.
- Resolves #336, returns to previous non-trap code for editor-stepping, using a setInterval of 500ms. Beta tag removed - no addons have it as of now.
- Adds `/* sa-autoupdate-theme-ignore */`, a way for themes to ask for their userstyles to work like non-theme userstyles. This is specially important for the editor-stage-left addon, which needs a userscript to work properly (to enable sharing scripts between sprites, aka "share the love"). So now, a full refresh is needed to enable/disable it. Another addon benefits from this: Customizable block colors can now move its CSS that was previously in a JS string, to the userstyle CSS itself, and it's guaranteed not to be removed.
- Fixes bug where "share the love" would not work on editor-stage-left after going inside the editor, then outside, then inside again.

### DIFF
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -12,7 +12,6 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  ,
   "tags": ["editor"],
   "enabledByDefault": false
 }
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -1,7 +1,6 @@
 {
   "name": "60FPS player mode",
   "description": "Alt+Click the green flag to toggle 60FPS.",
-  "warning": "This feature might make Scratch projects lag, especially if you're not using a powerful computer. Performance will be enhanced in future updates.",
   "credits": [
     {
       "name": "Jeffalo"
@@ -13,7 +12,7 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "traps": true,
+  ,
   "tags": ["editor"],
   "enabledByDefault": false
 }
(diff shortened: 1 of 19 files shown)


WorldLanguages:
@ZenithRogue Remember I moved the CSS to the theme userstyle and it broke? I think with this change, it should work just fine, since the userstyle can't be replaced.

PR #654, opened 2020-11-06
Update addon tags, make some addons recommended

### LINKED ISSUE #487: Recommended addons should be enabled by default
v1.2.0 will be released 4 weeks after the transition from Scratch Messaging Extension. As another step following this linear transition, all (or most?) recommended addons should be enabled by default to new users.
As well, we should remove "scratch messaging" and "message count on extension icon" as hardcoded first 2 listed addons in the settings page.

WorldLanguages:
Resolves #487 

- With the removal of the prototype handler, `60fps` is now recommended and turned on by default.
- The new `resizable-comments` addon will be turned on for all users.
- The new `animated-thumb` addon will be turned on for all users.
- Made `exact-count` recommended.
- `full-signature` is now recommended and will be turned on for new users.
- `forum-time-zones` and `infinite-scroll` are now beta.
- `studio-tools` will be turned on for new users.
- The new `pause` addon is beta.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -14,20 +14,18 @@ export default class Tab extends EventTarget {
       : document.querySelector("script[type='text/javascript']")
       ? "scratchr2"
       : null;
-    if (info.traps) {
-      this.traps = new Trap();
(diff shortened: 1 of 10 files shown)


TheColaber:
Sounds Awesome!! I'm glad that 60fps is going to be faster!

PR #655, opened 2020-11-06
Enable some old addons for all users on v1.3.0

WorldLanguages:
This enables 60fps, studio tools and full signature for all users. Since it's been a while since Scratch Addons was Scratch Messaging, it seems to be a good time to do it.
It's possible some users did try out these new addons and then disabled them, but that should be a minority. Sadly, because of the way I coded this, it's not possible to know if a user disabled an addon themselves, or if they never enabled a non-enabled-by-default addon.


TheColaber:
Maybe something in the settings that lets users know we now have some default addons?

PR #668, opened 2020-11-07
Forum search: Better styling on the search bar

Hans5958:
I'm not sure why the search bar elements doesn't have a margin on them. This PR adds it, along with a border on the top to separate it and changing it so it uses flex.

Keep in mind that this is quite subjective, and comments are appreciated.

image

### DIFF
--- a/addons/forum-search/addon.json
+++ b/addons/forum-search/addon.json
@@ -4,6 +4,9 @@
   "credits": [
     {
       "name": "DatOneLefty"
+    },
+    {
+      "name": "Hans5958"
     }
   ],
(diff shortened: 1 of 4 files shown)


mxmou:
Is it compatible with 2.0 → 3.0?

Hans5958:
I forgot to test that, but I have plans of what to do, such as canceling most of the effects. Will do later.

Hans5958:
Okay, it is done. The new styles are cancelled on 2.0 → 3.0

mxmou (approved):

PR #680, opened 2020-11-08
Making Icons More Visible and Cute

samuraininja360:
I am still working on more

This style makes it easier to identify what the icon means because it has a fill colour that scratch users are more familiar with.


GrahamSH-LLK:
We use a single iconset. Please use icons from there.

PR #681, opened 2020-11-08
Update heart.svg

samuraininja360:
Same Reason as last pull request

Is there a way where I can edit multiple files and create a pull request?


GrahamSH-LLK:
We use a single iconset. Please use icons from there.

PR #684, opened 2020-11-08
Live featured project: Disable addon by default

Hans5958:
This PR set the `enabledByDefault` value of the "Live featured project" addon to false.

I thought a post-merge PR is enough. Turns out it isn't, and the damage was done. On the bright side, at least I got some inquiries about it.

P/S: I am really pissed off rn

### DIFF
--- a/addons/live-featured-project/addon.json
+++ b/addons/live-featured-project/addon.json
@@ -50,5 +50,5 @@
     }
   ],
   "tags": ["community"],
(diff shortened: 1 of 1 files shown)


WorldLanguages:
This is not really a great solution. This only makes the addon stay disabled for new users (or users that haven't updated yet).

I'll make a discussion issue on what to do.

PR #685, opened 2020-11-08
It's version bumping time again

GrahamSH-LLK:
This adds -prerelease and bumps the version to 1.4.0.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,8 +2,8 @@
   "manifest_version": 2,
   "name": "Scratch Addons",
   "description": "Scratch Addons provides new features and themes for the scratch.mit.edu website and project editor.",
-  "version": "1.3.0",
-  "version_name": "1.3.0",
+  "version": "1.4.0",
+  "version_name": "1.4.0-prerelease",
   "background": {
     "page": "background/background.html"
   },


WorldLanguages:
Lol no, we will have to publish v1.3.1 because "live featured project" wasn't supposed to be turned on for everyone.

Hans5958:
@Hans5958 why are you still using MSU as a base

Base, as in inspiration. Ofc it won't depend in MSU. I build it from the ground up based on the original code

Hans5958:
Bold of you to assume that the next version is 1.4.0 and not 1.3.2

wgyt735yt:
> 
> 
> Bold of you to assume that the next version is 1.4.0 and not 1.3.2

I think that 1.3.2 should be used for a smaller update and 1.4.0 should be used for a major update.

Hans5958:
> 
> 
> > Bold of you to assume that the next version is 1.4.0 and not 1.3.2
> 
> I think that 1.3.2 should be used for a smaller update and 1.4.0 should be used for a major update.

thanks captain obvious /s  
We have discussed it earlier about this, and it seems WL has it's own versioning scheme. I just wanted to point out that we don't know whether if we need to bump the second number or the third number.

wgyt735yt:
> 
> 
> > > Bold of you to assume that the next version is 1.4.0 and not 1.3.2
> > 
> > 
> > I think that 1.3.2 should be used for a smaller update and 1.4.0 should be used for a major update.
> 
> thanks captain obvious /s
> 
> We have discussed it earlier about this, and it seems WL has its own versioning scheme. I just wanted to point out that we don't know whether if we need to bump the second number or the third number.

/nick Captain Obvious
Ok, so we don't know which number to bump, if WL has a versioning scheme, can't we use that? Or am I missing something?

WorldLanguages:
@wgyt735yt The first number likely won't ever be bumped. The second number is bumped if there are any updates other than changes to an addon (exception: hotfixes). The third number is bumped if the changes were only to addons, not the addon loader itself.

PR #690, opened 2020-11-09
Add v1.3.1 changelog

GrahamSH-LLK:


wgyt735yt (approved):
Looks good, you added the one change from 1.3.1.

Hans5958:
It's all good, I did some changes for things I forgot to do last time, and then some adjustments on your changes. The only thing that I wanted to point out is about giving a new line after the start tag. This does not matter, however, because it will genererate the same view, but just wanted to point it out, for tidiness.

Hans5958 (approved):

PR #699, opened 2020-11-10
Add Experimental Dark

### LINKED ISSUE #526: Make dark mode... Darker?
## Old version:
image
## New version:
Screenshot 2020-10-18 at 11 50 29 AM

mxmou:
**Resolves**

Resolves #526

**Changes**

Added my dark mode, which I decided to name "Experimental Dark" because it was called *Dark mode (experimental!)* in my userstyle for such a long time.

**Reason for changes**

Because I said I'm going to make it "soon": https://github.com/ScratchAddons/ScratchAddons/issues/526#issuecomment-711300578

**Tests**

Still has some issues for which changes to the addon loader will be needed - this is why it's a draft. I will also need to test compatibility with other addons.


WorldLanguages:
Nevermind, I saw #700.

addons/scratchr2/experimental_dark_compatibility.css
```diff
@@ -0,0 +1,6 @@
+:root {
+  --scratchr2-dropdown-image: url(https://mxmou.github.io/customize-scratch/assets/caret.svg);
```
WorldLanguages, line 2: Maybe you could save this SVG in the addon folder? I think `url()` functions in CSS are relative to the CSS file itself, not the page they're injected to, so something like `url(../caret.svg)` should work?

WorldLanguages (approved):

addons/scratchr2/experimental_dark_compatibility.css
mxmou: But theme userstyles are injected as `<style>` elements, so the browser doesn't know which file they came from.

WorldLanguages: Nice catch. In the future, I might add something like `%addon.self.dir%` to the userstyle injector.

WorldLanguages: Moved discussion to #741

Hans5958:
Dude just merged without knowing that all PRs are on hold

PR #704, opened 2020-11-10
Make theme userstyles ordered

### LINKED ISSUE #700: Userstyles should be loaded in a predictable order

I noticed the order in which userstyles are loaded is not the same as they appear in `addon.json`. This is a problem for Experimental Dark, which expects page-specific styles to override the generic ones.

Userstyles (at least theme ones) should be loaded in the same order as they are listed in `addon.json`.

mxmou:
**Resolves**

Resolves #700

**Changes**

Theme userstyles now load in the same order as they are listed in `addon.json`.

**Reason for changes**

This makes them predictable.

**Tests**

I opened the editor with editor dark mode enabled and the styles were in correct order.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -45,7 +45,9 @@ async function getContentScriptInfo(url) {
           fetchThemeStylesPromises.push(
             fetch(chrome.runtime.getURL(`/addons/${addonId}/${styleUrl}`))
               .then((res) => res.text())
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Nice catch. Because this is a `fetch` to the extension itself, I forgot order isn't guaranteed.

WorldLanguages (approved):

PR #710, opened 2020-11-10
Update forum.svg

GrahamSH-LLK:
Switches out: Ugly icon with Less ugly icon!


samuraininja360 (approved):
Nice it actually suits it more

PR #723, opened 2020-11-11
Add favicon on the web pages

Hans5958:
There is no icon on the extension web pages. This PR adds it. Oh, and I also moved the "Scratch Addons" portion of text to the right, since it is already described on the icon.

image


YOYITsM3M8:
i already had icons. 
image

Hans5958:
I think that's a Chrome thing. We don't have that on Firefox.

YOYITsM3M8:
oh

WorldLanguages:
Thanks, Firefox! 🙄

PR #724, opened 2020-11-11
Remove unnecessary word "theme"

Hans5958:
In this PR, I remove the word "theme" from the "Dark Theme" and the "Light Theme" phrase as it is unnecessary because it is already described that the settings are for themes.

This might be too minor and subjective, so it is open for discussion.

If this got rejected, how about we talk about the capitalization?


GrahamSH-LLK:
Sounds good. I thought that that was a little redundant when I was making it, but I wanted to be clear.

GrahamSH-LLK (approved):

PR #725, opened 2020-11-11
Curator link: Add addon

Hans5958:
This is a migration of my userscript, based on what I made on https://github.com/Hans5958/userscripts/blob/master/Scratch/Curator%20Links.user.js. 

It creates a clickable link for the curator of the front page.

image


TheColaber:
If it only does that, maybe it is too small to be an addon...

WorldLanguages:
Eh, maybe this should be added within "more links"

TheColaber:
> Eh, maybe this should be added within "more links"

Good Idea, probably doesn't need a setting, just run automatically.

Hans5958:
> 
> 
> Eh, maybe this should be added within "more links"

I don't think so. The purpose of that addon is quite different than this one. I'm also not a fan of merging things if it's not the same.

WorldLanguages (approved):

PR #729, opened 2020-11-11
Another Load more fix

mxmou:
A Load more fix is needed again, for search results this time: LLK/scratch-www#4624

I thinks it's OK to have it enabled by default.


WorldLanguages:
Oh.

PR #746, opened 2020-11-13
All frames

### LINKED ISSUE #702: Addons don't run in iframes

The contact form and the preview in the forums are iframes. Addons don't affect them.

Userscripts and userstyles should be able to run in iframes.

GrahamSH-LLK:
Adds the all_frames permission to our content scripts. This will allow user scripts and styles to access these. This resolves #702.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -21,12 +21,14 @@
     {
       "matches": ["https://scratch.mit.edu/*"],
       "run_at": "document_start",
-      "js": ["content-scripts/cs.js"]
+      "js": ["content-scripts/cs.js"],
+      "all_frames": true
     },
     {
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Eh, this requires testing before merging something like this. We want to make sure this only works when scratch.mit.edu embeds itself. We don't want to mess with embeds by other websites.

Hans5958:
I think it would work considering that [this page](https://developer.chrome.com/content_scripts) and [that page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts) mentions that it is injected "matching the specified URL requirements"

PR #748, opened 2020-11-14
New addon: fix broken SVG uploads

WorldLanguages:
It's been reported through feedback that SVG uploads from Affinity Designer don't work very well in Scratch. This can be fixed by removing `width="100%" and height="100%"` from the SVG's XML. This addon automates this for the user.
This is not enabled by default because this doesn't affect most SVG editors.

Demo: https://scratch.mit.edu/projects/447085841/


GrahamSH-LLK:
This resolves https://github.com/LLK/scratch-svg-renderer/issues/124, as an addon.

PR #752, opened 2020-11-15
Make Thumbnail Setter a Scratch Editor Addon

### LINKED ISSUE #695: "Thumbnails setter" should be classified as an editor addon
Technically it does have to do with the website, but it's only usable within the editor for individual projects, not the website itself.

Mast3rGenius:
**Resolves**

Resolves #695 

**Changes**

 Moves the thumbnail setter addon to the scratch editor section.

**Reason for changes**

The thumbnail setter is a editor feature,

**Tests**

 N/A


TheColaber:
I don't think we should remove "community" from here. It shows up on your project so the community will see it.

lisa-wolfgang:
The problem is that it's not listed as an editor addon at all. Since it appears in the editor (and is enabled by default) I thought of it as an editor addon, and I'm sure other users will have the same line of reasoning.

TheColaber:
> The problem is that it's not listed as an editor addon at all. Since it appears in the editor (and is enabled by default) I thought of it as an editor addon, and I'm sure other users will have the same line of reasoning.
> 
> Is this perhaps part of a bigger issue, that addons can't have multiple labels?

WDYM it is an editor addon. Look in the file: https://github.com/ScratchAddons/ScratchAddons/blob/2cd9e7dab3f9e3f50395b2c69c72ad3369c9fbf5/addons/animated-thumb/addon.json#L20

TheColaber:
It needs no change

lisa-wolfgang:
It doesn't matter what it says in the code if that code doesn't translate to function:

image
image

TheColaber:
Oh, I see. Because the community tag is here, the editor tag has no effect.

TheColaber (approved):

WorldLanguages:
> Oh, I see. Because the community tag is here, the editor tag has no effect.

Yeah, some tags are actually categories, and you're supposed to only use one of them.

PR #753, opened 2020-11-15
Localize everything

### LINKED ISSUE #480: Localize addons (NOTE: All PRs are on hold!)
It's likely too early to focus on this, but I want to create the issue. It would be nice if addon titles, descriptions, warnings, setting names, etc. could be localized. There could also be an `addon.i18n` API or similar, in order to get localized strings from userscripts or popups.

### LINKED ISSUE #703: Console log will be flooded on Firefox
**Describe the bug**
Console log gets flooded by 297 SA logs describing "messages from background", at least on Firefox.

**To Reproduce**
Open new tab (for example, front page) then open console.

**Expected behavior**
No (or a few) debug messages

win10, firefox82.0.3

**Additional context**
Log:
[flooded log.txt](https://github.com/ScratchAddons/ScratchAddons/files/5516947/flooded.log.txt)

### LINKED ISSUE #727: Fix wrong HTML on DevTools
image

### LINKED ISSUE #750: Clicking version number opens two tabs
Open popup
Find the version number
Click
It opens two tabs
Because both href and event handler does the thing

apple502j:
**Resolves**
Resolves #480
Resolves #727 
Resolves #750 

**Changes**
Many changes. Ask me for details.

**Reason for changes**
Obvious


pufferfish101007:
Are the addon tags localised?

Hans5958:
Isn't it bad to change the JSON so it almost empty, when you can do, you know, put the English strings there as the baseline/default and then use the other language translations when it is needed/existed?

YOYITsM3M8:
Umm lol

GrahamSH-LLK:
On the subject of translation, should we use something like Crowdin? It would lower the annoyance of having a million "localization prs." Just a thought...

apple502j:
@pufferfish101007 They are, on Settings page (addon JSON is untouched)

apple502j:
@Hans5958 Old ones still work, I just replaced all because this is what the PR's purpose is.

WorldLanguages:
> Isn't it bad to change the manifest so it is almost empty, when you can do, you know, put the English strings there as the baseline/default and then replace those with the translation of another language when it is needed/existed?
Just telling so it is still easy to implement new addons without an hassle of adding the translation files, but still able to be translated in a later time.

I got confused by this, but what you mentioned is actually what the PR does. If you don't specify `null` in the manifest for a property, it uses what you specified.

Hans5958:
> 
> 
> @Hans5958 Old ones still work, I just replaced all because this is what the PR's purpose is.

No, what I meant is to keep the English names and descriptions on the manifest, whilst the other languages can go to the addons-l10n pages

Hans5958:
@WorldLanguages @apple502j please respond

apple502j:
In case we use the platforms like Transifex or Crowdin, having the source translation as a separate file helps when setting up the stuff.

Hans5958:
I'd still rather having the strings also on the manifest.json, and have the "en" only as a reference for the translation tool only.

WorldLanguages:
@Hans5958 So you're saying you want all English localizations to be duplicated?

WorldLanguages:
> I don't want to have the bars higher because people needed to consider another thing.

People only need to consider this if they consider localization as well. We won't reject addons because they aren't ready to be localized. We will do the work ourselves, or keep it unlocalized.

Hans5958:
Yes. I think this makes it easier for outsiders to develop without minding the business of the l10n, and it's just an easy duplication* to set it up for translation. I could go up and make a CI of this so if we need it

Except if there are conflicts, probably also modify the translation pages so the language tool knows if that needs to be updated, if that''s a thing.

Hans5958:
Also, opt- instead of settings-. Come on, another refactor?

oslA, it's interesting when you do dashes to separate "level1-level2-level3", instead of "level1/level2/level3" (I don't know, and hope this is intended)

WorldLanguages (changes requested):
My final say for the current state of this PR
1) Instead of `@opt-`, I'd prefer something like `@settings`. So `discuss-button/@opt-name-removeIdeasBtn` would now be `discuss-button/@settings/removeIdeasBtn`. (BTW: why is the addon ID at the beginning even necessary? It's on the name of the json file)
2) I agree with @Hans5958 on the fact English localization should be in `addon.json`. For a potential GUI based translation service, we could make a script that automates a pseudo-localization-file for English, so that we can get working localization files for other languages. This would mean there would be no `null`s over manifests.
3) I'm not sure about the current state of giving addon scripts both a `msg` and `safeMsg` function. IMO, we should pass an `l10n` object with both functions on it.
4) My current browser language is `en-US`, so I get lots of console errors in the background because it's attempting to get localizations for `en-US`, and we only have them for `en`. IMO it should stop looking for translations if `_general.json` isn't found. Alternatively, we could have a language-codes.json file of some sort. (This is not a big deal, since it's normal to see errors raise from the "inject-css-asap" code often, but the lesser errors the better)

WorldLanguages:
Regarding 2, just an idea: we should probably have another repo for translations, then "merge" those into this one before each release. That repo could also handle the translation of strings present on the extension web stores, like descriptions, and maybe screenshots, so that translators can help with those the same way they translate extension strings.

Hans5958:
I don't think a second repo is needed as if we used a translation tool we can set up an integration for it. But, you can do it just in case we need it IG.

GrahamSH-LLK:
> I don't think a second repo is needed as if we used a translation tool we can set up an integration for it. But, you can do it just in case we need it IG.

Agreed.

WorldLanguages (approved):
1) Was partially solved, but fine, I guess you have programming reasons to keep it like this
2) Solved
3) Wasn't solved but fine
4) Wasn't solved, but can be fixed later normally since it's not a addon developer facing change

WorldLanguages:
@Hans5958 I'll merge this tomorrow, please comment now if you have something to say after the new changes.

Hans5958:
Another one. Just wanted to point out that there are some setting id that uses underscore and ones that uses dashes? Probably equalize them (or maybe use camelCase)?

WorldLanguages:
IMO we should just make the syntax `addon-id/@settings-names/settingId_or_id`

Hans5958:
BTW, @mxmou just merged #699 without the permission. Probably also make changes for them?

apple502j:
@Hans5958 @WorldLanguages  Messages are snake-case, unless it's a setting message (which requires changing setting ID - no thanks)

Hans5958:
I have did necessary changes with the manifest. Just merge master and it should be green.

Hans5958:
* not counting the library property

PR #757, opened 2020-11-15
Feature unshared projects

TheColaber:
**Changes**

New Addon... Adds unshared projects to be featured on your profile.

**Tests**

Works on chrome.


Explosion-Scratch:
Cool!!!

WorldLanguages:
@Explosion-Scratch A few more offtopic comments and I'm banning you from the org. To see if you've learnt, please do not reply to this message - instead, react with a thumbs up :P

WorldLanguages (approved):
We will try to find a better way to share strings like "load more" between addons before merging this (localization)

WorldLanguages (changes requested):
Instead of adding the string `"Load more"` (which will not work very well in other languages), please use Scratch's string. This code would work: `document.querySelector("[data-control='load-more'] > span").textContent`

TheColaber:
Fixed @WorldLanguages

WorldLanguages (approved):

PR #768, opened 2020-11-17
New addon: Context menu block switching

### LINKED ISSUE #593: Switch blocks via dropdown (operators, conditionals)
**Describe the new addon**

When right-clicking a +, -, *, or / block, the dropdown would allow the block to be switched out for any of the others a la 2.0.

image

Would also work for the AND/OR/NOT blocks and the '>'  '<'  '=' blocks.

GarboMuffin:
**Resolves**

Resolves #593 

**Changes**

Adds item to the context menu that can change the block type, similar to Scratch 2.

image

It supports pretty much every combination of blocks you can think of. These are currently hardcoded.

LkpmxOf84E

This PR also changes urlChange to also fire in response to history.pushState in addition to history.replaceState because Scratch uses both of them at different times. This makes this addon actually run consistently when navigating from Scratch homepage to project page to editor (that first project > editor transition uses pushState not replaceState)

**Reason for changes**

People clearly want it in #593 

**Tests**

It seems to work fine from my testing in Chrome and Firefox.

Works with the Scratch 3 dev tools and colored context menus. Undo/redo works properly. Comments work. Converting if else to if will leave the else part of the chain in a separate script. Switching between blocks with different input names should work (hardcoded mappings)

This definitely needs more extensive testing and feedback.


TheColaber:
Wow! I know how hard it must have been to create this as I have tried to make this before. Would it be possible to make it not hard coded?

WorldLanguages:
WOW!!!!!!!!!!!!!!!!!!!!!!!!!!

GarboMuffin:
> Would it be possible to make it not hard coded?

What would be the use case for full customizability? Every useful combination should just be part of the addon. There's non-trivial complexity involved in mapping opcodes, sometimes mapping input names, sometimes splitting substacks, and never mind the interface that would have to be created (current options API is no where near flexible enough) I might look into automating most of the mapping logic, then maybe adding a fully customizable interface might actually be viable.

I intend to, at the very least, add options for each category of blocks, like motion, looks, variables, operators, etc.

Hoping to have this in a mergeable state sometime this weekend.

griffpatch:
That's real nice - real nice! I didn't notice if you could switch the:
- turn left / turn right block? I think you might
- broadcast <-> broadcast and wait?
- forever <-> repeat until

This is making Scratch so much easier to use!

GarboMuffin:
This PR is ready to be reviewed

addons/block-switching/userscript.js
```diff
@@ -0,0 +1,691 @@
+import blockToDom from "./blockToDom.js";
```
WorldLanguages, line 1: Glad to see an addon use modules :D

content-scripts/inject/module.js
```diff
@@ -48,6 +48,17 @@ history.replaceState = function () {
   return returnValue;
 };
 
+const originalPushState = history.pushState;
```
WorldLanguages, line 51: This is OK, but from my experience, going inside/outside the editor uses `replaceState`, not `pushState`. Any idea why you had to add this? Maybe Scratch changes behavior depending on the browser?

webpages/settings/style.css
```diff
@@ -6,7 +6,7 @@ body {
   background: #222;
   font-family: "Sora", sans-serif;
   margin: 0;
-  height: 100vh;
+  min-height: 100vh;
```
WorldLanguages, line 9: This change appears to cause styling issues when the addon container doesn't overflow:

image

WorldLanguages:
Variable switching has no border (I have borders on, and they work for other types of blocks, such as operators). Is this intentional?

image

addons/block-switching/addon.json
```diff
@@ -0,0 +1,87 @@
...
+      "name": "GarboMuffin"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["editor"],
```
WorldLanguages, line 15: This deserves the `"recommended"` tag as well, since it was a 2.0 feature.

WorldLanguages (changes requested):
Asked for a few changes, see above

content-scripts/inject/module.js
GarboMuffin: 1. add `console.log('replaceState', arguments);` and `console.log('pushState', arguments);` in module.js to log calls to those
2. open this in a fresh new tab: https://scratch.mit.edu/projects/422112922/
3. click see inside
4. observe that a pushState is logged 
image
but every subsequent see inside/see project page will use replaceState for some reason
Happens in Chrome and Firefox

GarboMuffin:
I believe everything should be fixed now.

WorldLanguages (approved):
Thanks <3

Explosion-Scratch:
Yasssss

Explosion-Scratch:
This is amazing!

lisa-wolfgang:
If-else switching, too? This addon is going to be a lifesaver!

PR #405, opened 2020-10-04
Search Profiles Addon

Geotale:
**Changes**

Adds the "Search Profiles" add on that I've created. I've also essentially "unminified" it, and commented on it.

**Reason for changes**

It's simply a convenience so you don't have to go through a lengthy process to get to someone's profile.

**Tests**

I have tested my addon as a JavaScript URL, and the JSON should be correct, but I can't guarantee it.

PR #770, opened 2020-11-17
Search Profile Addon: Don't go to user if they don't exist

### LINKED ISSUE #468: Only show link to user profile if user exists

When using the user search addon, the profile tab still gets added even if the user doesn't exist. This could be confusing when it goes to a 404 page.

Fetch the user page, if it doesn't have a `200` status code then don't add the tab.

Rather than not adding the tab, we could just disable it or add a message when it's clicked saying that this user doesn't exist or something.

**Additional context**

I won't make a pr yet, as I don't know whether we want this or not, or want to do an alternative, but you can see the changes at https://github.com/pufferfish101007/ScratchAddons/blob/user-search-enhancement/addons/search-profile/userscript.js

TheColaber:
**Resolves**

Resolves #468

**Changes**

fetches the user's profile. If it doesn't exist, our tab will turn red and there will be a tooltip saying the user doesn't exist and the link to the user will not be added to the element

**Reason for changes**

So users know.

**Tests**

I do use a CSS filter that turns the image red so I am not 100% if this effects things on other browsers but this does work on chrome.


MasimbaTheHuman:
I use chrome, so I could test it.

YOYITsM3M8:
I could test too

YOYITsM3M8:
> I could test too

Confirmed: Profile tab turns red when the user is not found
Feedback: Add credit in the `addon.json`

TheColaber:
Ok, btw who else worked on this addon, right now there are no credits.

YOYITsM3M8:
> Ok, btw who else worked on this addon, right now there are no credits.

idk but also is this supposed to happen? 
image at first i thought i had another version of SA with user search on but i dont :C

TheColaber:
> > Ok, btw who else worked on this addon, right now there are no credits.
> 
> idk but also is this supposed to happen?
> image at first i thought i had another version of SA with user search on but i dont :C

There is no way that could have happened

YOYITsM3M8:
found the original pr: #405

TheColaber:
> > > I could test too
> > 
> > 
> > @TheColaber
> > Confirmed: Profile tab turns red when the user is not found
> > Feedback: Add credit in the `addon.json`
> 
> Actually, could you send me the addon?

Hmm?

YOYITsM3M8:
> > > I could test too
> > 
> > 
> > @TheColaber
> > Confirmed: Profile tab turns red when the user is not found
> > Feedback: Add credit in the `addon.json`
> 
> Actually, could you send me the addon?

https://github.com/TheColaber/ScratchAddons/tree/user-exists

pufferfish101007:
The reason no pr was ever made for this was because scratchr2's response times are really slow and we want to keep everything as lag-free as possible - see https://github.com/ScratchAddons/ScratchAddons/issues/468#issuecomment-706721646.
And while we're adding credits, I fixed a bug. (#466)

TheColaber:
I am using the scratch website and it's pretty fast...

GrahamSH-LLK:
> I am using the scratch website, not scratchr2 and it's pretty fast...

Scratch Website = Scratchr2

TheColaber:
> > I am using the scratch website, not scratchr2 and it's pretty fast...
> 
> Scratch Website = Scratchr2

Ok but is it really that slow? I mean, it's only one call.

mxmou:
> Ok but is it really that slow? I mean, it's only one call.

Compared to api.scratch.mit.edu or ScratchDB, it's slow. But if this doesn't cause a delay before the tab appears, it's OK.

MasimbaTheHuman:
> > > > I could test too
> > > 
> > > 
> > > @TheColaber
> > > Confirmed: Profile tab turns red when the user is not found
> > > Feedback: Add credit in the `addon.json`
> > 
> > 
> > Actually, could you send me the addon?
> 
> Hmm?

Nvm it doesn't work for me

WorldLanguages (changes requested):
I don't think this is a good idea. Fetching a scratchr2 page can take seconds sometimes. I'd rather see someone add a regex check to see if the searched string could potentially be a username.

TheColaber:
> I don't think this is a good idea. Fetching a scratchr2 page can take seconds sometimes. I'd rather see someone add a regex check to see if the searched string could potentially be a username.

Where did you want to fetch from?

WorldLanguages:
> Where did you want to fetch from?

I'd like to fetch the API, but there's no way to fetch the API that will with 100% certainty tell you a user won't 404. Only way is fetching old and slow scratchr2. So I'd rather not fetch. I'd like to see the "user" button not appear if the search is an invalid username, for example if it includes spaces.

wgyt735yt:
> > Where did you want to fetch from?
> 
> I'd like to fetch the API, but there's no way to fetch the API that will with 100% certainty tell you a user won't 404. Only way is fetching old and slow scratchr2. So I'd rather not fetch. I'd like to see the "user" button not appear if the search is an invalid username, for example if it includes spaces.

Ok, here's my take:
The api: can fetch data about banned users (wouldn't work)
Fetching scratchr2: too slow

>I'd like to see the "user" button not appear if the search is an invalid username, for example if it includes spaces.

Fair point

TheColaber:
> > Where did you want to fetch from?
> 
> I'd like to fetch the API, but there's no way to fetch the API that will with 100% certainty tell you a user won't 404. Only way is fetching old and slow scratchr2. So I'd rather not fetch. I'd like to see the "user" button not appear if the search is an invalid username, for example if it includes spaces.

Alright.

WorldLanguages:
I'm closing this PR, please open another one if you want to add a regex check

PR #781, opened 2020-11-18
Fix white corners in predicate block when middle clicking on dev tools

### LINKED ISSUE #777: White corners in predicate block when middle clicking on dev tools
**Describe the bug**

White corners in predicate block when middle clicking on dev tools

**To Reproduce**

Turn on dark mode and dev tools
Middle click

**Expected behavior**

Corners should match background color

**Screenshots**

White corners in predicate block when middle clicking on dev tools

Windows 10, Cr-based Edge 86

TheColaber:
**Resolves**

Resolves #777

**Changes**

Matched the color with some CSS to get rid of this:

**Reason for changes**

Cuz it didn't look right before.


WorldLanguages:
Does this work with all 3 dark modes? I think you might end up with a gray-ish background above a deep black background if you go for 3.Darker

TheColaber:
> Does this work with all 3 dark modes? I think you might end up with a gray-ish background above a deep black background if you go for 3.Darker

It should as the file that changes the background of the menu is the global.css which runs on all 3 settings.

PR #782, opened 2020-11-18
SA DevTools

### LINKED ISSUE #726: Add option to show the addon ID

Sometimes I wanted to check the ID of an addon for developing purposes, but then I sometimes confused because the ID an addon is sometimes not the same as the name of the addon.

I propose to add an option/setting to show the ID of all addon. I think the best way is to place the ID besides the name, styled with a gray color. This option should be added on the more settings page, of course.

TheColaber:
**Resolves**

Resolves #726

**Changes**

Addon IDs in settings and no install and uninstall link for devs.

**Reason for changes**

Cuz devs want it


WorldLanguages:
> Also removed transition to 1.3.1 cuz why not.

Please remove that, you'll cause conflicts.

TheColaber:
> > Also removed transition to 1.3.1 cuz why not.
> 
> Please remove that, you'll cause conflicts.

Ok

GrahamSH-LLK (changes requested):
It would be good if you could make a "Dev Mode" setting, and depend on that instead of the prerelease part.

TheColaber:
> It would be good if you could make a "Dev Mode" setting, and depend on that instead of the prerelease part.

Why? You can only really be a dev if it is in prerelease, we insert that after we send it to the store. If it doesn't have prerelease, it's from the store.

GrahamSH-LLK:
> Why? You can only really be a dev if it is in prerelease, we insert that after we send it to the store. If it doesn't have prerelease, it's from the store.

True, but what if you want to use dev mode? So far, these are things that could be beneficial to people who get it from the store.

WorldLanguages:
> It would be good if you could make a "Dev Mode" setting, and depend on that instead of the prerelease part.

Why, tho?

> So far, these are things that could be beneficial to people who get it from the store.

I don't think so. Adding a dev mode seems to be overcomplicating things. No dev can test their own addon with the web store version...

MasimbaTheHuman:
> > > Why? You can only really be a dev if it is in prerelease, we insert that after we send it to the store. If it doesn't have prerelease, it's from the store.
> > 
> > 
> > True, but what if you want to use dev mode? So far, these are things that could be beneficial to people who get it from the store.
> 
> Ok... I'll see what I can do then.

Well... dev mode is only for devs, and we are devs, so we just get it from the source.

mxmou:
> > It would be good if you could make a "Dev Mode" setting, and depend on that instead of the prerelease part.
> 
> Why, tho?

One reason is to be able to turn off developer mode even if you're a developer (for example to test things).

WorldLanguages:
@mxmou A dev can easily remove `-prerelease` from the version name.

PR #791, opened 2020-11-20
Setting Warnings/Notice Icons

### LINKED ISSUE #743: Add appropriate icons next to the warning and notice texts
How about adding icons such as a circle with an exclamation mark for notices and a triangle with an exclamation mark for warnings, for accessibility purposes and make it more clear what are those text for?

TheColaber:
**Resolves**

Resolves #743

**Changes**

Added Icons to settings
image

**Reason for changes**

For accessibility reasons

**Tests**

Works on chrome


WorldLanguages (changes requested):
The icons are too tall, specially for notices. Most notices are simply notes, and this style makes them more important than needed. IMO the height should stay the same it's always been for notices. For warnings, it's fine, but I'd like to see it a bit smaller as well.

TheColaber:
I'm trying to understand... are you saying the icons are too big or the message is too big... The icon shouldn't change the size of the message...

WorldLanguages:
> The icon shouldn't change the size of the message...

It does if the message is less than 3 lines long.

TheColaber:
> > The icon shouldn't change the size of the message...
> 
> It does if the message is less than 3 lines long.

Hmm seems that the icon is bigger with fewer lines and smaller with more lines... why, how?

WorldLanguages:
> Hmm seems that the icon is bigger with fewer lines and smaller with more lines... why, how?

Lol no that's likely your perception, the icon is the same size, but it looks bigger or smaller depending on the amount of lines.

TheColaber:
I wasn't kidding about the size changing:

mxmou:
> I wasn't kidding about the size changing:

That's how flexbox works. If you don't want it to change, make sure it has both `flex-shrink` and `flex-grow` set to 0 and that the text has `flex-grow: 1`.

TheColaber:
> > I wasn't kidding about the size changing:
> 
> That's how flexbox works. If you don't want it to change, make sure it has both `flex-shrink` and `flex-grow` set to 0 and that the text has `flex-grow: 1`.

Thanks @mxmou you're the best!

mxmou:
You don't need to put `display: flex` on the img.

WorldLanguages (approved):

PR #793, opened 2020-11-20
New settings "key"

### LINKED ISSUE #759: Settings Capability - Support Hotkeys

I'm am creating the Editor Hotkeys Addon #705.

I would like the user to be able to, not type their preferred hotkey, but actually perform it.
This is good number one: So syntax is not a problem
Number two: more convient

Make it hard without this feature.

**Additional context**

I'm doing this because @WorldLanguages said to here: https://github.com/ScratchAddons/ScratchAddons/pull/705#issuecomment-726247036

TheColaber:
**Resolves**

Resolves #759

**Changes**

I have added it in the settings and added some code for it in the index.js

**Reason for changes**

So it can support hotkeys for new addons so we can create addons for the editor, for devtools, and for the website.

**Tests**

Works on chrome. Report bugs, please.


webpages/settings/index.js
```diff
@@ -248,6 +248,41 @@ chrome.runtime.sendMessage("getSettingsInfo", ({ manifests, addonsEnabled, addon
...
+            }
+          }
+        }
+      });
+      keyInputs[i].addEventListener("keyup", function (e) {
+        e.target.value =
+          e.target.value == "Ctrl" || objectVals.filter((v) => v === e.target.value).length > 1 ? "" : e.target.value;
+      });
+    }
+  }, 2000);
```
WorldLanguages, line 285: This could be a timeout of 0

webpages/settings/index.html
```diff
@@ -174,6 +174,17 @@ <h1>{{ msg("settings") }}</h1>
...
+                  <input
+                    type="text"
+                    class="setting-input key"
+                    v-model="addonSettings[addon._addonId][setting.id]"
+                    @input="updateSettings(addon)"
+                    :disabled="!addon._enabled"
+                    v-if="setting.type === 'key'"
+                    :placeholder="setting.default"
+                    maxlength="100"
+                    :id="setting.id"
```
WorldLanguages, line 186: This could break if 2 addons have the same setting ID

WorldLanguages (changes requested):
I'll fix them myself, brb

webpages/settings/index.js
```diff
@@ -248,6 +248,41 @@ chrome.runtime.sendMessage("getSettingsInfo", ({ manifests, addonsEnabled, addon
...
+                return;
+              }
+            }
+          }
+        }
+      });
+      keyInputs[i].addEventListener("keyup", function (e) {
+        e.target.value =
+          e.target.value == "Ctrl" || objectVals.filter((v) => v === e.target.value).length > 1 ? "" : e.target.value;
+      });
```
WorldLanguages, line 283: What does this event do?

webpages/settings/index.html
TheColaber: Who would make an addon that has the same as another?

webpages/settings/index.js
TheColaber: e.target.value == "Ctrl" || objectVals.filter((v) => v === e.target.value).length > 1 ? "" : e.target.value;
user only pressed control key OR another setting uses that hotkey then make input empty. Otherwise keep as is.

webpages/settings/index.html
WorldLanguages: This is for the *setting* ID, not the addon ID. Setting IDs must be unique for each addon, but might not be unique for all addons.

webpages/settings/index.js
WorldLanguages: Please let's keep talking on the Discord server, I have no idea what `objectVals` means on here.

WorldLanguages (approved):

PR #797, opened 2020-11-21
Confirm actions addon

### LINKED ISSUE #223: New addon: confirm project share
I removed that feature from Developer Tools because it's too unrelated to the purpose of the extension. Instead, it should be a separate addon.

TheColaber:
**Resolves**

Resolves #223

**Changes**

Confirms you are ok with sharing, following, and joining studios.

**Reason for changes**

Was taken out of dev tools and griffpatch says he shares projects on accident a lot.

**Tests**

Works on chrome.


WorldLanguages (changes requested):
1. You forgot while/true and markAsSeen in your code, so this will only protect one button of each kind each page load, specially important in projects, where going outside then inside the editor creates another button.
2. Why does this have the forum tag? It asks if you're sure about following a forum? If that's the case, I don't think it's enough to warrant a forums tag.
3. You did not consider the "share" button inside the project player (not inside the editor)
4. You're not supposed to mark addons recommended without telling, but this one is fine.
5. To make it easy to localize, there's no need to include the username you're following/unfollowing in the alert. Something like `Are you sure you want to follow/unfollow?` should be enough.

TheColaber:
Sorry about the tags, I duplicate another addon folder to work

TheColaber:
@WorldLanguages Fixed

WorldLanguages (approved):

PR #800, opened 2020-11-21
Add search profile username validator

### LINKED ISSUE #468: Only show link to user profile if user exists

pufferfish101007:
**Resolves**

Resolves #468

**Changes**

Adds a regex and length check for the search term to see if it's a valid username; it tests the pattern `/[^a-z1-9_-]/i/`, and if it matches that, it's shorter than 1 character, or more than 20 characters, it turns the tab to go to the profile red, and adds a tooltip saying `"username" is an invalid username!`. The link to the profile page is only added if the username is valid.

**Reason for changes**

@WorldLanguages suggested it in #770

**Tests**

Works on chrome 86

**Additional information (because there is some)**

Please ignore the fact that there are 50 commits, and that none of them have meaningful names - the latest 2 commits (excluding code formatting) are the important ones that add changes.

Edit: #800!!!! 🎉


TheColaber:
Rather than hiding the button to confuse the user why the addon wasn't working, it's probably better to do this:
```js
  const nav = await addon.tab.waitForElement(".sub-nav.tabs");
  //Create elements for tab
  const tab = nav.appendChild(document.createElement("a")),
    li = tab.appendChild(document.createElement("li")),
    img = li.appendChild(document.createElement("img")),
    span = li.appendChild(document.createElement("span")),
    user = document.querySelector('[name="q"]').value.trim(),
    valid = user.length <= 20 && user.length >= 3 && !user.includes(" ") && user.match(/([\x30-\x39]|[\x41-\x5a]|[\x61-\x7a]|\x2d|\x5f)+/g)[0] == user;
  //Set up elements
  img.src = addon.self.dir + "/user.svg";
  img.className = "tab-icon";
  span.innerText = "Profile";
  tab.href = "/users/" + user + "/";
  //Check if whats entered is a valid username
  if (!valid) {
    img.style.filter = "grayscale(100%) brightness(100%) sepia(100%) hue-rotate(-50deg) saturate(600%) contrast(1)";
    span.style.color = "red";
    li.title = user + " is an invalid username!"
  }
}

```
However, this needs a bit of localization.

TheColaber:
lol I was litterly going to PR that code a second ago

pufferfish101007:
@TheColaber I'll go do something like that, I think I'll show the error message on hover so as to not clutter up the page/make things look too angry.

TheColaber:
> @TheColaber I'll go do something like that, I think I'll show the error message on hover so as to not clutter up the page/make things look too angry.

No one will see the error message, and it's not angry... it just red. I did add a tooltip but I don't expect a lot of people to see it.

pufferfish101007:
Updated - it now goes red and has a tooltip saying that it's an invalid username - essentially what @TheColaber put above, but with a couple of changes, including localisation.

WorldLanguages (approved):

PR #833, opened 2020-11-23
No errors in console!

### LINKED ISSUE #814: Get rid of all the annoying errors in console.
**Describe the bug**

The console is getting flooded with errors from addons that don't need localization.

**To Reproduce**

Enable addons that don't need localization and look in the console.

**Expected behavior**

I'd prefer no errors

**Screenshots**

image

TheColaber:
**Resolves**

Resolves #814

**Changes**

Updated script in l10n.js and got rid of unnecessary addon.json l10n's set to true.

These should only be set to true if used.

**Reason for changes**

There are no more of these annoying errors:

**Tests**

It should work.


WorldLanguages:
@apple502j told me this won't actually work. All `"l10n": true` does is determine if the strings for that addon should be submitted to Transifex. We could change that behavior, tho.

TheColaber:
> @apple502j told me this won't actually work. All `"l10n": true` does is determine if the strings for that addon should be submitted to Transifex. We could change that behavior, tho.

Hmm, so there is no alternative?

apple502j:
Closing as this is not a proper fix for the problem. The easiest fix is turning on "Ignore Network".

PR #836, opened 2020-11-23
Merge v1.3.3 into v1.4.0 on master

### LINKED ISSUE #835: Do not use CDNs in addons
Both Chrome and Firefox stores dislike or disallow the use of "remote code", even if it's in non-extension context and from trustworthy CDNs.
This was fixed in v1.3.3 (note from apple: v1.3.2 is the XSS fix release) but hasn't been merged into v1.4.0 (master) yet
This also adds `addon.self.lib`, which returns the URL for the "libraries" folder of the extension.

WorldLanguages:
v1.3.3 resolves #835


Hans5958:
Suggestion: How about merge master to v1.3.x instead, and set v1.3.x as a new master? That would make more sense, I guess.

apple502j:
@Hans5958 Does not make sense because we're developing v1.4.x

Hans5958:
That doesn't make sense either becaue 1.3.x is released before 1.4.0 and it is earlier based on semver, so it should come earlier on the commit history.

Maybe we should have a way to handle this next time.

PR #851, opened 2020-11-24
Smooth Project Borders on 2 to 3.0 pages

Daniel4-Scratch:
**Changes**

I have smoothed the borders for all project and profile preview
Screen Shot 2020-11-25 at 7 49 59 am

using:
```css
.thumb img {
    margin-bottom: 2px;
    border-radius: 5px;
}
```
and 
```json
{
   "url": "all.css",
   "matches": ["https://scratch.mit.edu/*"]
}
```
**Reason for changes**
Just making the project borders smooth

**Tests**
All test passed.

<sup> not 100% if the will be pulled

### DIFF
--- a/addons/scratchr2/addon.json
+++ b/addons/scratchr2/addon.json
@@ -100,10 +100,6 @@
     {
       "url": "experimental_dark_compatibility.css",
       "matches": ["https://scratch.mit.edu/*"]
-    },
-    {
-      "url": "all.css",
-      "matches": ["https://scratch.mit.edu/*"]
     }
   ],
   "l10n": true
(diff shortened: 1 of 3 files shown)


WorldLanguages (changes requested):
@mxmou will do the final call, but what I am sure about is that this new CSS should not belong in a new file. Likely in scratchr2.css but not sure.

mxmou (changes requested):
To match the Scratch 3.0 design as much as possible, this should not apply to profile pages (because they are similar to the 3.0 home page, where thumbnails are not rounded). Please move it to `studio.css` (for project and curator tumbnails). And yes, as @WorldLanguages said, `scratchr2.css` is the file that applies to all 2.0-styled pages.

Also, the border radius should be 4px and the margin 8px (to match the Explore page).

Daniel4-Scratch:
Ok, I will do this soon :)

Daniel4-Scratch:
Ok, updated the files

Daniel4-Scratch:
Done

WorldLanguages (approved):

mxmou (approved):

PR #854, opened 2020-11-25
Fix exact count l10n

TheColaber:
Fixes this: 
image


WorldLanguages:
This addon shouldn't require these strings **at all.**

PR #857, opened 2020-11-25
Fix project-notes-tabs addon for remixes

TheColaber:
**Changes**

Checks if there are any remix sections and reinserts the tabs so the tabs don't appear on the remix


addons/project-notes-tabs/userscript.js
```diff
@@ -40,5 +40,11 @@ export default async function ({ addon, global, console }) {
       document.querySelectorAll(".description-block")[tab].style.display = "block";
       document.querySelectorAll(".description-block")[!tab + 0].style.display = "none";
     }
+    (async function remixHandeler() {
```
WorldLanguages, line 43: I think you meant `handler`

PR #873, opened 2020-11-27
Update README

Explosion-Scratch:
I did a lot on it.

**Resolves**

Resolves #

**Changes**

**Reason for changes**

**Tests**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <img src="https://raw.githubusercontent.com/ScratchAddons/ScratchAddons/master/images/icon.svg" alt="Scratch Addons logo" align="right" width="128px"></img>
-# Welcome to Scratch Addons' repository!
+<h1 align='center'>Welcome to the Scratch Addons repository!</h1>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 <img src="https://raw.githubusercontent.com/ScratchAddons/ScratchAddons/master/images/icon.svg" alt="Scratch Addons logo" align="right" width="128px"></img>
-# Welcome to Scratch Addons' repository!
+<h1 align='center'>Welcome to the Scratch Addons repository!</h1>
```
WorldLanguages, line 2: Hmm, HTML in markdown?

WorldLanguages:
The readme is fine as-is

Hans5958:
We have a right-alinged logo so a middle one doesn't makes sense to me. Also, we already got a long sentence, so it doesn't looked good. It would make sense to be centered if it is just a short text like "Scratch Addon," but that might be a story for another time.

README.md
Explosion-Scratch: Yup. Some limited tags are supported (That's also what Hans did to right align the image). I fixed some grammar and I also aligned the text center. I also added a paragraph at the top because I thought that just having the sheilds images and then the about header seemed really empty and disorganized.

PR #887, opened 2020-11-29
Fix title case

### LINKED ISSUE #884: Some addon names are still Title Case
Forum Search and Curator Link

pufferfish101007:
**Resolves**

Resolves #884 

**Changes**

Removed Title Case from Forum Search, Scratch Messaging, Scratch Notifier and Curator Link. I noticed that in the issue, @WorldLanguages didn't mention messages or notifier - if we want to keep them in title case I can revert them.

**Reason for changes**

To be consistent with the rest of the addons.

**Tests**

None. It shouldn't need testing.


GrahamSH-LLK:
A. You should probably use different branches so that you don't get the 3 million commits
B. Scratch Messaging and Scratch Notifier are names of pre existing services, so we probably want to keep those uppercase.

pufferfish101007:
A. I know - they're all from ages ago when I didn't realise that branches were any use - after this pr I might just make a new fork to get rid of all the wxtra commits.
B. Ok, thanks for letting me know.

WorldLanguages (approved):

PR #895, opened 2020-12-01
Created Profile Page Banner addon

Ezlambry:
**Resolves**
None

**Changes**
Adds an addon that changes the featured project into a nice-looking banner.
profile page banner

**Reason for changes**
I showed this off in the Discord server, and everyone thought it was pretty cool!

**Tests**
I have tested this with the 2.0 to 3.0 style, the dark mode style, and the Live Featured Project addon (which I had to move the embedded project that it creates with some very hacky CSS). :P I have also tested this with Firefox and Brave, which both seem to work exactly the same.

### DIFF
--- a/addons/better-featured-project/addon.json
+++ b/addons/better-featured-project/addon.json
@@ -10,14 +10,20 @@
     "userscripts": [
         {
             "url": "userscript.js",
-            "matches": ["https://scratch.mit.edu/users/*"]
+            "matches": [
+                "https://scratch.mit.edu/users/*"
+            ]
         }
     ],
     "userstyles": [
         {
             "url": "userstyle.css",
-            "matches": ["https://scratch.mit.edu/users/*"]
+            "matches": [
+                "https://scratch.mit.edu/users/*"
+            ]
         }
     ],
-    "tags": ["community"]
+    "tags": [
+        "community"
+    ]
 }
--- a/addons/better-featured-project/addon.json
+++ b/addons/better-featured-project/addon.json
@@ -1,0 +1,23 @@
+{
+    "name": "Profile page banner",
+    "description": "Makes your featured project into a nice-looking banner for your profile.",
+    "credits": [
+        {
+            "name": "Ezlambry",
+            "link": "https://ezlambry.github.io"
+        }
+    ],
+    "userscripts": [
+        {
+            "url": "userscript.js",
+            "matches": ["https://scratch.mit.edu/users/*"]
+        }
+    ],
+    "userstyles": [
+        {
+            "url": "userstyle.css",
+            "matches": ["https://scratch.mit.edu/users/*"]
+        }
+    ],
+    "tags": ["community"]
+}
--- a/addons/better-featured-project/addon.json
+++ b/addons/better-featured-project/addon.json
@@ -1,29 +1,23 @@
 {
-    "name": "Profile page banner",
-    "description": "Makes your featured project into a nice-looking banner for your profile.",
-    "credits": [
-        {
-            "name": "Ezlambry",
-            "link": "https://ezlambry.github.io"
-        }
-    ],
-    "userscripts": [
-        {
-            "url": "userscript.js",
-            "matches": [
-                "https://scratch.mit.edu/users/*"
-            ]
-        }
-    ],
-    "userstyles": [
-        {
-            "url": "userstyle.css",
-            "matches": [
-                "https://scratch.mit.edu/users/*"
-            ]
-        }
-    ],
-    "tags": [
-        "community"
-    ]
+  "name": "Profile page banner",
+  "description": "Makes your featured project into a nice-looking banner for your profile.",
+  "credits": [
+    {
+      "name": "Ezlambry",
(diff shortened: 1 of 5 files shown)


YOYITsM3M8:
looks good from a first glance

GrahamSH-LLK:
Can you make the About, WIWO, and Activity taller?

Ezlambry:
> Can you make the About, WIWO, and Activity taller?

Well the About and WIWO are actually the same size that they originally were, they just look shorter because I moved them around, but yeah, I had to make the Activity smaller so that it would fit in with the others. I don't think it would be a good idea to make them taller because the banner already takes up a bunch of space how it is right now, and on some screens it could even push the Shared Projects box off of the screen

addons/better-featured-project/addon.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Profile page banner",
+  "description": "Makes your featured project into a nice-looking banner for your profile.",
```
WorldLanguages, line 3: The description appears to imply the addon changes what other users see, instead of what the extension user sees.

addons/better-featured-project/addon.json
```diff
@@ -0,0 +1,23 @@
...
+  "credits": [
+    {
+      "name": "Ezlambry",
+      "link": "https://ezlambry.github.io/"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/users/*"]
```
WorldLanguages, line 13: This should be `https://scratch.mit.edu/users/*/` to avoid matching other pages like followers, etc.

addons/better-featured-project/addon.json
```diff
@@ -0,0 +1,23 @@
...
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["https://scratch.mit.edu/users/*"]
```
WorldLanguages, line 19: Same as above

addons/better-featured-project/userscript.js
```diff
@@ -0,0 +1,55 @@
+export default async function ({ addon, global, console, msg }) {
+  function createBetterProfilePage(featuredThumb, featuredLink, featuredHeading, featuredTitle) {
+    document.documentElement.style.setProperty("--featured-thumb", `url("${featuredThumb}")`);
+    document.documentElement.style.setProperty("--refresh-notice", `\"${msg("refresh-notice")}\"`);
```
WorldLanguages, line 4: This text is not visible with the "website dark mode" addon enabled. Maybe you could open a `prompt()` that asks the user if they want to refresh to see the changes?

addons/better-featured-project/userstyle.css
```diff
@@ -0,0 +1,121 @@
...
+  height: 140px !important;
+}
+.doing {
+  height: 130px !important;
+  overflow: hidden;
+}
+.profile-details span {
+  display: inline !important;
+}
+.profile-details #better-follower-count {
```
WorldLanguages, line 109: `#better-follower-count` (?)

WorldLanguages (changes requested):

WorldLanguages:
The WIWO placeholder is offcenter
image

addons/better-featured-project/userstyle.css
Ezlambry: oh, in my old version it also contained a follower count that got it from scratchdb, but because another addon already has that feature i just got rid of it in the scratchaddons version

WorldLanguages:
Not needed, but a better way to have a link to the featured project would allow new scratchers to be promoted
image

jeffalo:
Where does it get the banner image from?

TheColaber:
> Where does it get the banner image from?

The featured project.

addons/better-featured-project/userscript.js
Ezlambry: I figured out how to make it so that it automatically refreshes, which I think would be easier and quicker for the user, but I've also noticed that most of the text isn't even visible with the "website dark mode" addon enabled and "Scratch 2.0 -> 3.0" disabled, which should probably be reported as a separate issue.
image

addons/better-featured-project/addon.json
Ezlambry: > Makes the featured project on any user's profile page into a nice-looking banner.
Hopefully this is better?

WorldLanguages (approved):

Explosion-Scratch:
> Not needed, but a better way to have a link to the featured project would allow new scratchers to be promoted
> image

Unrelated but it really ticks me off that the 'i' is not capitol XD

Explosion-Scratch:
Also what about people that both want a banner and want a different featured project? Could we add upload an image? (Using jeffalo's system)

Explosion-Scratch:
Oh wait, we'd have to store the image urls =|

YOYITsM3M8:
> Oh wait, we'd have to store the image urls =|

So we would need a external server/ a external db
Right?

PR #899, opened 2020-12-01
Load More in What's happening?

TheColaber:
**Resolves**

Resolves @WorldLanguages 

**Changes**

New Setting for to Load More in What's happening?

**Reason for changes**

Cuz WL wants it and I don't mind doing it.

**Tests**

I'm 99% sure this works.


WorldLanguages (approved):
Keep in mind this feature will be turned on for most users on the v1.5.0 update

TheColaber:
As long as the button shows and works and disappears when no more activity left, I think we should be fine.

PR #901, opened 2020-12-02
Fix forum search

TheColaber:
**Resolves**

Resolves ME

**Changes**

I've changed `html_first_checked` to `html_last_checked` because idk why it was `html_first_checked` that's not even a thing...
This somehow got changed in between localization
image

**Reason for changes**

Because the forum search is not working! Try it!


j5155:
I was wondering why it always gave the same results whatever I searched! Thanks for fixing it!

PR #934, opened 2020-12-03
Add forum warning as an addon

apple502j:
Forum warning is now an addon, not Content Script feature.

Changes:
- It's an addon, so it's toggleable (on by default)
- It blocks posts from being sent, if it contains `scratch[ -]?(?:addons|messaging[ -]extension)`(case-insensitive). Can be turned off via addon settings.
- Works on all forums, on topic addition as well as replies.


mxmou:
Could something like this be reconsidered as an extension of #2089? (not as an addon for consistency)

WorldLanguages:
@mxmou What would the reasoning be behind that?

mxmou:
@WorldLanguages To let people new to the forums know about the browser extension policy before they break it and likely get an alert.

WorldLanguages:
Ohh, I see what you mean now. I thought you were talking specifically about moving the logic to an addon.
We should probably create a different issue for discussing that. But initially I'm not convinced about applying our own censor. "Scratch Addons" in comments is different, because 1. you get muted 2. there's not enough context for the user to figure out that "Scratch Addons" was what triggered the filter. In the case you mentioned, you don't get muted (just warned) and the alert links you to the extension policy.

mxmou:
> Ohh, I see what you mean now. I thought you were talking specifically about moving the logic to an addon.
> We should probably create a different issue for discussing that. But initially I'm not convinced about applying our own censor. "Scratch Addons" in comments is different, because 1. you get muted 2. there's not enough context for the user to figure out that "Scratch Addons" was what triggered the filter. In the case you mentioned, you don't get muted (just warned) and the alert links you to the extension policy.

People don't want to get warnings from the moderators, and some forumers post vague responses such as "please don't name extensions because you might get banned" to posts mentioning Scratch Addons, which can be confusing.

WorldLanguages:
@mxmou Can you create a new issue?

WorldLanguages:
I kinda changed my mind after I learnt the bad word filter in the forums is a giant regex... A PR that adds a notice above the post editor when the user typed in "scratch addons" in cs.js would be a good addition.

Issue #951, opened 2020-12-05
Custom Navbar


retronbv:
**Describe the new addon**

Adds the ability to change the url and text on the Ideas and About buttons 

**IMPORTANT: As W_L requested ANY `javascript:` url will make the addon not run**
The user is warned about this when changing settings

**Link to the new addon**

https://github.com/YOYITsM3M8/ScratchAddons/tree/custom-navbar
(I will be making a pull request if people like this)

GrahamSH-LLK:
I think this is awesome!

apple502j:
Fix syntax error before submitting...

YOYITsM3M8:
> Fix syntax error before submitting...

ok done sry

TheColaber:
Yay!

Marc92020:
this is a cool idea!

YOYITsM3M8:
> this is a cool idea!

Thank you!

retronbv:
~~took long enough~~

Issue #742, opened 2020-11-13
Change the color of the warning and notice texts


Hans5958:
The current color for the warning and notice texts are red and yellow, which I think people may be unable to differentiate the purpose of the texts straight away. How about change them to yellow and light blue respectively, just like how browser logs work?

GrahamSH-LLK:
What? Red = error/warning Yellow = notice It aligns pretty well with logs and makes sense.

Hans5958:
Errors are red and warnings are yellow on console logs. Info is none, but it is quite common that they give blue on that.

![image](https://user-images.githubusercontent.com/11584103/99084230-46369a00-25f9-11eb-8c29-ba8d798383a9.png)

![image](https://user-images.githubusercontent.com/11584103/99084261-4df63e80-25f9-11eb-82a0-1df2f8c7a5ee.png)

https://css-tricks.com/a-guide-to-console-commands/

mxmou:
The point of notices and warnings is that we want people to read them. Making their colors less prominent is probably not a good idea.

WorldLanguages:
Styling changed since this suggestion was made. Another issue must be made if appropriate

PR #957, opened 2020-12-05
New warnings and notices system

### LINKED ISSUE #744: Change how warning and notice texts are added
Instead of creating two properties for them, how about creating an array of objects that cover both properties? The object includes the type and the content of the texts.

This may give multiple benefits.
- Easier implementation because you don't need to check additional properties (or other reasons perhaps?)
- Cleaner manifest because they are stored in a single array.
- Ability to add multiple warning and notice texts (probably?)
- Ability (in the source code) to add additional, separate warning text for some cases (like when it uses addon traps)

TheColaber:
**Resolves**

Resolves #744

**Changes**

New addon.json format...
In order to add notices or warnings it will look like this:
```json
{
  "extra": [
    {
      "type": "notice",
      "text": "Example notice.",
      "id": "exmapleID"
    },
    {
      "type": "warning",
      "text": "Example warning.",
      "id": "exmapleID"
    },
  ],
}
```

**Reason for changes**

This allows for multiple warnings and notices like this:
image

**Tests**

I need to confirm I added the notices for all the addons.
Also, someone would need to update the Wiki when this gets merged


YOYITsM3M8:
If this gets accepted will I need to change the notice in my new addon? #951

TheColaber:
Yeah

Hans5958:
Nice, but can you put a descriptive PR title?

WorldLanguages:
I mean, fine, but looks unnecessary? The localization code will need to be tweaked a bit for this as well.

TheColaber:
> I mean, fine, but looks unnecessary? The localization code will need to be tweaked a bit for this as well.

How exactly?

apple502j:
Waiting Hans for schema update. I can do l10n stuff myself

Hans5958:
> 
> 
> Waiting Hans for schema update. I can do l10n stuff myself

I said you don't need to wait for me, smh

WorldLanguages:
@Hans5958 Lol this is partially why I dislike PR tags, you keep adding and removing "do not merge" :P

Hans5958:
Okay, here's something: Why is it called `extra`? Would it be more informative to do it as `info` or `extraInfo`?

GrahamSH-LLK:
> Okay, here's something: Why is it called `extra`? Would it be more informative to do it as `info` or `extraInfo`?

Agreed. Extra is weird. Info is better.

WorldLanguages:
Info sounds good.

Hans5958:
Another idea: Maybe how about making it `notice` instead (since both types of it are all notices) and changing the `notice` type into something like `info` (like how `console.info()` works)

Speaking of which, I still don't know why we didn't have a less scary colors such as light blue and yellow instead of yellow and red, but that's #742.

TheColaber:
I need a final confirmation before I spend a bit of time changing stuff..
- I should change `extra` to `info`

or

- I should change `extra` to `notice` and change the current `notice` to info

~~and maybe change the colors~~?

apple502j:
Can't we make notice/warning an array instead?

Hans5958:
> 
> 
> Can't we make notice/warning an array instead?

It is an array?

Hans5958:
Another one: ID for every `info`, for translation purposes

apple502j:
Yeah. How about:
```json
{
"warnings": [ { "id": "xxx", "content": "Lorem Ipsum!" } ]
}
```

TheColaber:
Comment Updated, Check Above

Hans5958:
> 
> 
> Yeah. How about:
> 
> ```json
> {
> "warnings": [ { "id": "xxx", "content": "Lorem Ipsum!" } ]
> }
> ```

Note that the PR and the issue calls for merging two of the properties into one single array, not just having both of them change into bunch of arrays.

Hans5958:
@TheColaber - Yeah, you should add IDs. The l10n system is going to need it anyways.

PR #960, opened 2020-12-06
Address feedback on dark-www

apple502j:
- Dark WWW stylesheets will no longer load at all on editor pages. On those pages, a media query `color and not color` is added. Since this always returns false, stylesheets are not loaded.
- Fixed background color on cloudmonitor


mxmou (approved):
It would be better if the rows on cloudmonitor were still striped (something like `.table-striped tbody > tr:nth-child(odd) > td {background: rgba(0, 0, 0, 0.1);}`) but it isn't necessary.

Issue #657, opened 2020-11-06
"Editor dark mode": convert modes to presets, allow for custom colors


WorldLanguages:
Now that presets exists, it would be nice to convert the current "3 userstyles" approach from `editor-dark-mode` to a single CSS file that uses CSS variables, then allow users to choose a preset, or customize it themselves as they wish. Hopefully no setting will need a userscript to function correctly? If not, it's not a big deal anyway.

mxmou:
Then that addon would be "customize everything", not a dark mode.

WorldLanguages:
@mxmou Indeed, I am not sure how we could name an addon like that. It would be a custom colors addon with dark mode presets.

WorldLanguages:
This would be a nice opportunity to clear up the editor context menu mess - dark modes usually mess with borders, heights, and don't make it easy to distinguish the separations between different buttons added by different addons:

![image](https://user-images.githubusercontent.com/17484114/99882724-61a93280-2c01-11eb-8c9d-72d8d4ad9e43.png)

mxmou:
Would "Editor customized colors" and "Website customized colors" be good names for the addons after this change?

WorldLanguages:
@mxmou Not sure, they'd effectively be those, but I bet most people would just use them for dark mode.

mxmou:
@WorldLanguages I agree, but I think it's better to cover all features in the name and instead mention in the description that it can be used for dark mode.

RedGuy12:
@MaxiMouse well, 60fps was changed to custom fps, but the title stayed the same since "most people would keep it at 60".
________________________________

mxmou:
@RedGuy12 @WorldLanguages I think it depends on what the defaults are. If the default colors are dark, then dark mode is the correct name. If not, it's customized colors.

WorldLanguages:
Dark mode will continue to be the default

mxmou:
> Dark mode will continue to be the default

This is very unintuitive to use. It means that the default dark mode version can't have a name, that resetting the addon to defaults would make it a dark mode instead of the default colors like in custom block colors, and that there would have to be a preset with Scratch's default colors (for people who just want to tweak them slightly instead of using dark mode).

WorldLanguages:
Oh right, haven't considered that.
What I meant was that by default, colors would be black. But if the user clicks "reset" in a color, it would return to the Scratch default.
[...]

PR #997, opened 2020-12-10
Toggle dots in editor dark mode

### LINKED ISSUE #815: editor dark modes should have option to toggle the dots
the dots
image

should be toggleable for each theme (only one has it now and its hardcoded in) like the drop shadows

TheColaber:
**Resolves**

Resolves #815

**Changes**

Added a new file called toggle-dots.css that will toggle the dots only when the setting is on

**Tests**

This needs to be tested. Something seems weird about the colors


addons/editor-dark-mode/addon.json
```diff
@@ -116,6 +129,12 @@
       "id": "textShadow",
       "type": "boolean",
       "default": false
+    },
+    {
+      "name": "Toggle dots",
```
mxmou, line 134: The setting name should be something like "Show dots" or "Show grid", not "Toggle dots".

WorldLanguages:
This PR is ok but I'd prefer to go all in #657. Adding more and more options will increase the complexity of migraring current users to the new preset system.

WorldLanguages:
I'm closing this, I believe we should do #657 all at once

PR #1011, opened 2020-12-11
 Show update notification for existing users & "new" tag in settings page

WorldLanguages:
Related: #991 

I believe the new onion skinning addon is important so it makes sense to notify existing users about it.
I've also added a "New!" tag on that addon and made it always go first in the settings page (for now)
I also added this "New!" tag to the "help translate" button in the settings page.

Graphic design is my passion:
image


TheColaber:
Maybe for each update, every new addon (or feature I guess) gets a "New!" tag? I guess there are now too many addons to tell!

apple502j:
Closing, working on my own. The code looks too buggy to merge.

PR #1026, opened 2020-12-13
add profile option to my-ocular

jeffalo:
changes:

makes my-ocular addon more epic
image
image


addons/my-ocular/addon.json
```diff
@@ -1,15 +1,41 @@
 {
-  "name": "Show my-ocular status",
-  "description": "Shows my-ocular statuses under usernames on the forums. Uses my-ocular.jeffalo.net to obtain status data. Statuses are moderated and filtered.",
+  "name": "Show ocular status",
+  "description": "Shows ocular statuses and favourite colours. Requests my-ocular.jeffalo.net to obtain status data. Statuses are moderated and filtered.",
```
WorldLanguages, line 3: Favorite.

WorldLanguages: Also, why did you change "uses" to "requests"?

addons/my-ocular/profile.js
```diff
@@ -0,0 +1,42 @@
...
+
+  var container = document.querySelector(".location");
+
+  var response = await fetch(`https://my-ocular.jeffalo.net/api/user/${username}`);
+  var data = await response.json();
+
+  var statusText = data.status;
+  var color = data.color;
+  if (statusText) {
+    var statusSpan = document.createElement("i"); // for whatever reason, chrome turns variable named status into text. why the heck. aaaaaaaaaaaaaaaaaa
```
WorldLanguages, line 12: lol, because of `window.status`

apple502j: Another reason to use let/const :P

addons/my-ocular/addon.json
jeffalo: it flowed a big better, should I change it back?

WorldLanguages: Yes, keep it as before

addons/my-ocular/profile.js
```diff
@@ -0,0 +1,42 @@
...
+    statusSpan.title = `This is ${username}'s ocular status, displayed with Scratch Addons. You can set one at https://my-ocular.jeffalo.net.`;
+    statusSpan.innerText = statusText;
+
+    var dot = document.createElement("span");
+    dot.title = `This is ${username}'s ocular favourite colour, displayed with Scratch Addons. You can set one at https://my-ocular.jeffalo.net.`;
```
WorldLanguages, line 17: This will require localization. I suggest you also remove the username placeholder for easier translation, changing it to "this user's" or something like that.

WorldLanguages: I'd also merge both strings (favorite color and ocular status) into a single one.

jeffalo: 👍

jeffalo:
how about now?

addons/my-ocular/addon.json
jeffalo: done

addons/my-ocular/addon.json
```diff
@@ -1,15 +1,41 @@
...
+      "url": "discuss.js",
+      "matches": ["https://scratch.mit.edu/discuss/topic/*"],
+      "settingMatch": {
+        "id": "discuss",
+        "value": true
+      }
+    },
+    {
+      "url": "profile.js",
+      "matches": ["https://scratch.mit.edu/users/*"],
```
TheColaber, line 20: Might want to add a "/" at the end... otherwise you can go to https://scratch.mit.edu/users/Scratchteam/following/ for example and you will get an error in the console.

TheColaber (changes requested):

addons/my-ocular/addon.json
jeffalo: fix

WorldLanguages (changes requested):
1. Keep same addon name
2. Localize span title

jeffalo:
> 1. Keep same addon name
the status that one sets on my-ocular is called an ocular status, so that's why i was thinking the addon name would be show ocular status instead of show my-ocular status

WorldLanguages:
@jeffalo Still, the user needs to set the status on my-ocular status. Scratch Addons is not connected with Ocular so it doesn't make sense to have consistent naming in expense of users not knowing where they can set an "ocular status" (that is, in my-ocular)

WorldLanguages (approved):

PR #1006, opened 2020-12-11
Non hardcoded popups & toggle addons through popup (responsive iframe)

TheColaber:
**Resolves**

Resolves #773
 - completes number one
 - completes number two
 - 3 and 4 yet to be done if even necessary
 
Resolves #1054
Soon to resolve #521

**Changes**

added a new addon.json property: "popup"
removed cloud-games notice
change popups system
added vue to popup
removed the now unnecessary error code for scratch-messaging
added links to users in cloud-games
made settings even more responsive for iframes
made a new popup for quick settings

**Tests**
:O WE ARE READY TO MERGE :O

Definitely, definitely, definitely, we will need a lot of testing on this PR after this last todo is done. I have a feeling this PR will be open for a while...

PR #1067, opened 2020-12-17
Links to users in cloud games popup now work

### LINKED ISSUE #1054: Cloud games popups: links to users don't work
They're \<a>'s without href's :P

pufferfish101007:
**Resolves**

Resolves #1054 

**Changes**

adds a href attribute to the user `<a>` so it actually links to the user

**Reason for changes**

because it's annoying

**Tests**

Works on chrome


pufferfish101007:
Oops sorry I didn't realise that #1006 fixes it 
*facepalm*

PR #1071, opened 2020-12-18
Add setting to enable/disable developer mode

### LINKED ISSUE #921: Add an option to turn on developer mode instead of detecting the version name
I'm not sure why we went the way of detecting the `-prerelease` prefix instead of having a simple checkbox on the more/global settings page. I wanted it for my usage with the release versions, just like me using my browser's developer tools on a release version. Also, one wrong typo and it won't work, instead of a single checkbox that will always work, as long you don't wipe your storage area. Maybe also implement so a `-prerelease` prefix triggers the developer tools, if the settings is not defined on the browser storage.

apple502j:
Resolves #921 

Also, uninstall URL no longer fires in developer mode, and added ability to switch back to addon name display when it is showing addon id.


webpages/settings/index.html
```diff
@@ -306,6 +306,41 @@ <h1 class="setting-heading">{{ msg("moreSettings") }}</h1>
               </div>
             </div>
           </div>
+          <div class="setting-body">
+            <div class="addon-topbar">
+              <span class="addon-name"
+                ><img src="../../images/icons/wrench.svg" style="vertical-align: middle; margin-right: 5px" /> {{
```
TheColaber, line 312: image
orange icon... you might want to add some style...

background/transition.js
```diff
@@ -1,19 +1,23 @@
...
-    });
+    if (details.reason === "install") {
+      chrome.storage.local.set({
+        bannerSettings: { lastShown: `${major}.${minor}` },
+      });
+    }
+  });
+  if (!isDevVersion) {
+    chrome.runtime.setUninstallURL("https://scratchaddons.com/farewell");
   }
```
TheColaber, line 22: My current version when running this is 1.6.0-prerelease (so devMode is already enabled). When I turned off dev mode in the settings, there wasn't an un install link. So you need a listener for new more settings enabled to check if that setting was changed. But it also goes the opposite way too. If I were to install this extension as version 1.5.0 and in the settings turn on devMode setting, when I uninstall, I would still get an uninstall link.

TheColaber (changes requested):

background/transition.js
TheColaber: But (with my new saved settings) if I were to uninstall and reinstall again, then this wouldn't be a problem since it only decides these things when the extension is installed.

WorldLanguages:
I honestly believe a developer is clever enough to add `-prerelease` on the manifest. With the current features we provide to developers (none at all, I can only think of one or two features that depend on prerelease) I don't see any benefits, this requires more translations and code. If you're not cloning the git repo locally, the welcome and uninstall pages might be an inconvenience, sure. But if we let developers bypass that limitation they'll have less incentives to actually use git properly. IMO having a developer mode is just a fancy addition that we don't need at all.

Touchcreator (commented):
Nice addition! But 1.6.0 will have dev mode already

PR #1083, opened 2020-12-22
New addon: Record Project Video

apple502j:


TheColaber:
So there is no way to add sound?

WorldLanguages (approved):
This is OK as a beta addon

WorldLanguages:
Some things I'd like to see in the future, hopefully:
1. Sound? There must be some hacky way? No need to support the music extension, sounds alone would be just fine. Can we somehow record the `AudioContext`?
2. Better UI - showing how many seconds are left
3. The .webm file has no length property (Chrome stable), which kinda sucks. Is this something we can control?
4. Can we name the file properly? (instead of just "video")
5. Option to show the cursor? (likely not possible)

apple502j:
1. Probably possible, but not sure if the inconsistency is acceptable.
2. Yeah, maybe.
3. What do you mean?
4. Filename should be controlled via user agent, not us.
5. Yeah, that's impossible.

WorldLanguages:
By 3 I mean, you can open the webm file in a browser and it will not show how long it is until it reaches the end. And if you see the Windows file properties for the file, it doesn't have a set length property.
I don't understand what you mean by 4, since the current name `video.webm` is hardcoded in the code

apple502j:
3: This is browser limitation. All we do is make a blob and download it.
4: When saving, browsers should ask users what to name the downloaded file.

Marc92020:
what the button could look like:
record video in 3 0

Issue #783, opened 2020-11-18
Userscript autorefresh


WorldLanguages:
In the same way theme userstyles (most times) autorefresh, it would be cool if userscripts could do that too, specially theme userscripts. This would mean users could instantly try themes like "colored context menus" without refreshing the page.
Together with this, it would be nice to allow any userscript to be aware if they were disabled (for example, the "discuss button" addon could remove it from the navbar after being disabled, and the "editor stage left" could remove its "share the love" fix dynamically, meaning toggling the addon wouldn't need a refresh)

TheColaber:
So part of the userscript would need to tell it how to undo its changes?

mxmou:
This would be useful for #415, so it could automatically determine if `onprimary` needs to be light or dark based on the primary color.

pufferfish101007:
I think rather than autorefreshing - as then the addon api would need to figure out how to undo all the changes, I think it would be better to encourage the use of the settings change event, and add enable/disable events, and encourage the use of those.

WorldLanguages:
> as then the addon api would need to figure out how to undo all the changes

This is not what I suggested. Addons are the one in charge of undoing their own changes.

TheColaber:
Many addons can try to do this, but some addons remove elements from the page so that might be a bit tricky.

mxmou:
> Many addons can try to do this, but some addons remove elements from the page so that might be a bit tricky.

They could save the removed elements.

mxmou:
Anyway, enable/disable events @pufferfish101007 suggested would be functionally equivalent so maybe also consider those?

WorldLanguages:
@mxmou Kind of - disable events would work just fine, something like `addon.self.addEventListener("disabled", () => {})`. However, I think userscripts should opt in into being injected after the page has loaded, not opt out. This is specially important for addons with `"runAtComplete": false`.

mxmou:
> However, I think userscripts should opt in into being injected after the page has loaded, not opt out. This is specially important for addons with `"runAtComplete": false`.

Isn't listening for an "enable" event opting in?

WorldLanguages:
> Isn't listening for an "enable" event opting in?

If a userscript registered an event, then it is already running, so not really. By enable, I mean running userscripts that weren't run on page load. If a user disables an addon, then reenables it, the userscript will simply be ran again - no need to have an "enable" event for those cases.

PR #1096, opened 2020-12-24
Add color settings to Scratch 2.0 → 3.0

### LINKED ISSUE #415: Add color selection settings to Scratch 2.0 -> 3.0
This awesome addon by Maximouse can easily alter the colors of the header and the hover color and normal color of links. Please add this function into settings!

mxmou:
**Resolves**

Resolves #415 

**Changes**

Adds color settings to the 2.0 → 3.0 addon.
image

These settings will be moved to website dark mode if #657 is implemented. I added the notice because enabling the addon, then changing the primary color immediately doesn't update the text color (see #783).

The added files have nothing to do with compatibility, but they're named like that for a reason.


WorldLanguages:
This won't be merged in time for v1.6.0 because of localization (and well, Christmas!) 
Can you explain what you mean by this?
>These settings will be moved to website dark mode if #657 is implemented.

I do not understand, what stops you from moving these to website dark mode, where it would likely make the most sense?

mxmou:
> I do not understand, what stops you from moving these to website dark mode, where it would likely make the most sense?

2.0 → 3.0 currently overrides website dark mode, and some parts of it rely on that. And website dark mode doesn't change the primary bar color on 2.0-styled pages, so there are currently two different colors. I'm not yet sure how this should be implemented -- as two separate settings?

WorldLanguages:
@mxmou That makes sense, I simply did not understand why you said the sentence I quoted.

WorldLanguages:
My navbar looks like this on default settings, what am I missing?
image

mxmou:
@WorldLanguages Fixed.

WorldLanguages (approved):

Issue #396, opened 2020-10-03
Bugs on "data category tweaks"


WorldLanguages:
> Sometimes, a list got duplicated with the same name, and the exact same content. When i deletes the real list, the "ghost list" still exist, but the list section is empty.. I reloaded but its still the same. The next day, I found out that the ghost list suddenly changed from a global list to a local list, plus its now showing in the for this sprite only list section... 

> Sorting "this sprite only" variables is broken. 

> The separation for lists and local variables do not work properly after programming.

> For the variables and list ( for this sprite and all sprite ), sometimes it works, sometimes it shows the previous version and the separated version, so basically 2 variables group.

Explosion-Scratch:
@towerofnix ^^^

RokCoder:
I'm encountering this issue frequently. I haven't noticed the list reappearing as local though. This is particularly gnarly as I tend to notice it when a list is being displayed. When the bug occurs there's no way to hide the list (other than to manually attack the json) but a duplicate list can be toggled on and off.

mxmou:
I've actually seen posts in Scratch's Bugs and Glitches forum about this, so I think it's a serious problem.

https://scratch.mit.edu/discuss/topic/444244/
https://scratch.mit.edu/discuss/topic/445060/

WorldLanguages:
Oh wow, this got into the Bugs and Glitches forum

towerofnix:
Just putting some notes here so they're public (and in case I can't figure out where to go from here) - I'm working on this ATM:

> Sometimes, a list got duplicated with the same name, and the exact same content. When i deletes the real list, the "ghost list" still exist, but the list section is empty.. I reloaded but its still the same. The next day, I found out that the ghost list suddenly changed from a global list to a local list, plus its now showing in the for this sprite only list section...

(Terminology note: list "monitors" and "variables" are different. A **list variable** is the internally stored VM/Blockly object which contains the items in the list, its name, ID, etc. A **list monitor** is the on-screen display for that list. It references a list variable by name (not ID), and there is normally exactly one monitor for every list variable.)

@WorldLanguages and I found a few things while investigating [this project](https://scratch.mit.edu/projects/432567841/editor/) (which includes a few cases of the bug):

**On list behavior:**

[...]

PR #1097, opened 2020-12-24
Bring back data category tweaks

### LINKED ISSUE #751: Slowly add back "Data category tweaks"
I understand that this addon was too buggy to keep in the extension, but I personally found it very useful. I never experienced any problems with it, actually.

I'd suggest to take one of two courses of action:
1. Re-add the complete addon, but label it as Beta and/or add a warning about potential buggy behavior in the addon description. As bugs are reported and fixed, the Beta status could eventually be removed. (Or possibly not lol)
2. Re-add one component at a time. Whichever of the two (list category separation or local variable separation) was the least buggy could be added back first. If all goes well with that, add the second component on top of that. This could potentially help with finding bugs caused by both extensions being enabled at once.

GarboMuffin:
**Resolves**

Resolves #751 

**Changes**

image

This is a reimplementation of the data category tweaks addon that apparently caused various issues (#396). This PR uses a very different approach than the old version (#260) that is less invasive (no prototype hacks and doesn't reimplement the entire data category) so it might not have the same problems.

It supports both options that were supported by the old version (separate category for lists and separating global and local variables)

The old version changed the sorting logic for things like dropdowns and context menus when separating local variables is enabled, but this version does not. This is intentional - I don't want to touch prototypes or anything invasive until it's clear that what's already written doesn't cause issues.

This should be marked as beta until real users have had some time to try to break it because of all the issues with the old version.

**Tests**

This hasn't been thoroughly tested, but it seems to work fine for me. It needs more testing.

I did try out the old version and could consistently break it by making a new custom block, but this version is not affected by that particular issue, so it has at least 1 fewer bugs.


scratchusernamemrtbts:
How about seperating cloud var and non-cloud var because some non-cloud var also have cloud stumble

WorldLanguages:
TODO: if we keep the old addon ID, anyone that had this addon enabled when it was available will now have it enabled again. Is this intended?

addons/data-category-tweaks/userscript.js
```diff
@@ -0,0 +1,178 @@
...
+  };
+
+  const separateLocalVariables = (workspace, toolboxXML) => {
+    const { variables, lists } = separateVariablesByType(toolboxXML);
+
+    const SMALL_GAP = 8;
+    const BIG_GAP = 24;
+
+    const makeLabel = (l10n) => {
+      const label = new DOMParser().parseFromString("<label></label>", "text/xml").firstChild;
```
TheColaber, line 23: Why do you need to do this to make a label?
Can's you just do
```js
const label = document.createElement("label")
```

addons/data-category-tweaks/userscript.js
```diff
@@ -0,0 +1,178 @@
...
+    const originalGetBlocksXML = vm.runtime.getBlocksXML;
+    vm.runtime.getBlocksXML = function (target) {
+      const result = originalGetBlocksXML.call(this, target);
+      hasSeparateListCategory = addon.settings.get("separateListCategory");
+      if (hasSeparateListCategory) {
+        result.push({
+          id: "data",
+          xml: `
+          <category
+            name="%{BKY_CATEGORY_VARIABLES}"
```
TheColaber, line 143: Whoops looks like this needs to be a "$"

GarboMuffin: % is correct. Blockly will parse and convert that to a translated string (https://github.com/LLK/scratch-gui/blob/ddd2fa06f2afa140a46ec03be91796ded861e65c/src/lib/make-toolbox-xml.js#L694-L704)

TheColaber: Ah... I see

GD-Distortion (approved):

WorldLanguages:
I'll rename the addon ID so that no user gets a beta addon enabled by default after the update. We could keep the old addon ID and add some logic to the update routine but it's not worth it.

WorldLanguages (approved):

PR #1117, opened 2020-12-28
Fix some bugs in project-notes-tabs

GarboMuffin:
**Changes**

Here's the bugs in project-notes-tabs that this fixes:

 - the tab buttons did not work at all in Firefox
 - instructions would not fill empty space in Firefox
 - it would throw errors and stop working after the initial page load when there is only one section
 - if the project only had instructions, the tab would be called "Notes and credits"
 - the tab buttons could occasionally appear above the remix credit box
 - there was a brief period of time (before onload) where the unmodified instructions and n&c are shown without tabs.

**Tests**

Tested on these projects (just found from trending) in Chrome and Firefox:

 - https://scratch.mit.edu/projects/467543815/
 - https://scratch.mit.edu/projects/455720538/
 - https://scratch.mit.edu/projects/455488994/

Editing descriptions still works. Doesn't break if you go in/out of editor. Remixing a project works fine

### DIFF
--- a/addons/project-notes-tabs/addon.json
+++ b/addons/project-notes-tabs/addon.json
@@ -10,7 +10,8 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["https://scratch.mit.edu/projects/*"]
+      "matches": ["https://scratch.mit.edu/projects/*"],
+      "runAtComplete": false
     }
   ],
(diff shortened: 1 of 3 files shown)


TheColaber:
Thx. I never realized these bugs existed. Nor were they reported.

PR #1121, opened 2020-12-29
Block counter in editor

### LINKED ISSUE #1115: Block counter
counts the blocks used in the project

TheColaber:
**Resolves**

Resolves #1115

**Changes**

Add a setting for the project-info addon to enable a live block count in the editor.

**Tests**

This is tested and everything works and is way faster for that matter... 

Order Fixed my mxmou thx!
VM help from GarboMuffin


addons/project-info/userscript.js
```diff
@@ -1,30 +1,50 @@
...
-          scriptCount: scriptCount,
-        };
-      } else if ("info" in projectData) {
-        projectInfo = projectData.info;
-      } else {
-        projectInfo = { spriteCount: "?", scriptCount: "?" };
-      }
+  const vm = addon.tab.traps.onceValues.vm;
+
+  const getBlockCount = async () => {
```
GarboMuffin, line 4: This doesn't need to be a async, infact being async will make it run extremely slow (relatively) in Firefox

addons/project-info/addon.json
```diff
@@ -14,5 +20,14 @@
...
   ],
+  "settings": [
+    {
+      "id": "editorCount",
+      "name": "Live block count in editor",
+      "type": "boolean",
+      "default": false
+    }
+  ],
+  "traps": true,
```
apple502j, line 31: This is no longer used

apple502j (changes requested):

mxmou:
> I'm just dealing with this order...
> image

You can add `order: 1` and it will always be on the end. It would also be a good idea to add some padding between the last button and the label.

TheColaber:
> order: 1

Where should I put that?

TheColaber:
Ah, You mean css, thx

j5155:
nice but could it also show by the sprite and script count without editing the project?

TheColaber:
> nice but could it also show by the sprite and script count without editing the project?

it already does on the project page

j5155:
> > nice but could it also show by the sprite and script count without editing the project?
> 
> it already does on the project page

no? 
image
I mean it would be cool if block count was shown here, unless that's what script count is and I'm dumb

TheColaber:
> > > nice but could it also show by the sprite and script count without editing the project?
> > 
> > 
> > it already does on the project page
> 
> no?
> image
> I mean it would be cool if block count was shown here, unless that's what script count is and I'm dumb

Oh sorry, I read your question wrong the first time. I mean that would be cool but 1) It doesn't fit, and if we make the font even smaller it will be too hard to read 2) I would think that because this value edits so frequently, that it's the only one that should go in the editor.

WorldLanguages (approved):

PR #1127, opened 2020-12-29
Better Quoter Addon

### LINKED ISSUE #124: [New addon] Better quoter
**Describe the new addon**

Make quoting easier by adding a quote button when someone selects
3piO1GSXVK

**Link to the new addon**

https://github.com/JeffaloBob/ScratchAddons/tree/quote

but i messed up alot of my code and i think it would be better if someone smarter made something similar instead of merging this half functional piece of terrible code

TheColaber:
**Resolves**

Resolves #124

**Changes**

Select text in a post then hit the quote button. Just like Github!

**Tests**

Yeah it works


jeffalo:
how about a post with some bbcode markup in it?

mxmou:
> how about a post with some bbcode markup in it?

That would be hard. The only way I can think of is fecthing the source of the post, then using a modified(?) version of postmarkup to generate HTML with data attributes describing the position of elements with the source, then checking if it matches with the HTML of the post. If yes, quote the selection, otherwise quote the whole post because it was likely edited before the user pressed Quote.

jeffalo:
> > how about a post with some bbcode markup in it?
> 
> That would be hard. The only way I can think of is fecthing the source of the post, then using a modified(?) version of postmarkup to generate HTML with data attributes describing the position of elements with the source, then checking if it matches with the HTML of the post. If yes, quote the selection, otherwise quote the whole post because it was likely edited before the user pressed Quote.
> 
> _Edit:_ It would probably be easier to try to convert the HTML back to BBCode. Another advantage of this is that the resulting BBCode is always well-formed.

yeah, that's where i got stuck with my quoting addon, but i guess this works as it is too

TheColaber:
This can definitely be improved for the future, but I think the basic part is down for this.

WorldLanguages (approved):

PR #1131, opened 2020-12-29
Custom FPS

### LINKED ISSUE #679: Framerate Setter
**Describe the new addon**

An option on projects to set the framerate.

**Link to the new addon**

None, however, @GarboMuffin made https://garbomuffin.github.io/p3/ which has an option for the framerate the project runs at. They stated that they: 
> Replaced Scratch's "compatibility mode" with a more generic "framerate" option

@GarboMuffin Would it be possible to bring this to Scratch Addons?

TheColaber:
**Resolves**

Resolves #679

**Changes**

Replaced Jeffalo's 60FPS addon with my Custom FPS addon. The default FPS is 60 however. I still gave them credit ;)

**Reason for changes**

Much better and cooler.

**Tests**

Works, but should be tested more.
Also @apple502j how are my files? B)


WorldLanguages (changes requested):
Most users will keep it at 60 - please let's keep the old addon name.

WorldLanguages (approved):

addons/60fps/userscript.js
```diff
@@ -15,25 +13,48 @@ export default async function ({ addon, global, console }) {
...
+    const setFPS = (fps) => {
+      gloabal_fps = fps;
+      clearInterval(vm.runtime._steppingInterval);
+      vm.runtime._steppingInterval = null;
+      vm.runtime.start();
+    };
+    const checkFPS = (fps) => {
+      let valid = 30;
+      // if framerate is a number and is not negitive or 0
+      if (Number(fps) == Number(fps) && fps > 0) {
```
WorldLanguages, line 40: No need to check, it's guaranteed to be a positive integer. If you don't want 0 to be a possible value, change the setting to "integer" and set a min of 1.

TheColaber: Yet, here we are... 
image

PR #1143, opened 2020-12-31
Fix website dark mode bugs

### LINKED ISSUE #1015: Website Dark Mode: Change or Remove Studio Frame
**The Studios Section on a User's Profile Has a Border that Doesn't Change After Switching to Dark Mode**

This issue makes the user's profile look a bit ugly especially when you see different colours that don't fit in place. I also realised that the transparent blue on the focused comment or reply chain has a weird double-layer transparency instead of having a uniform colour

mxmou:
**Resolves**

Fixes problems mentioned in #792, except the border between comments.

**Changes**

- #1090 but the proper way
- The comments disabled message is now white
- Fixed the featured project and report profile/studio modal when Scratch 2.0 → 3.0 is disabled
- Made scrollbars dark. On Edge the ideal way to do this would be `color-scheme: dark` but this would need different CSS depending on the browser.

@apple502j Why were both PRs that fixed issue #1015 closed?


apple502j:
@mxmou Probably because I accidentally deleted a branch.

PR #1145, opened 2020-12-31
Fix double scrollbar in popup

### LINKED ISSUE #600: Double scrollbar in popup
**Describe the bug**

The extension popup has two scrollbars.

**To Reproduce**

I don't know what causes this.
**Edit:** Now I do. In your browser's settings, set the default zoom to 110% (or more) and open the popup.

**Expected behavior**

There should only be the orange scrollbar.

**Screenshots**

image

- OS: Windows
- Browser: Chrome 86

mxmou:
**Resolves**

Resolves #600 (finally)

**Changes**

Removes the double scrollbar in the popup when browser zoom is set to more than 100%. I don't like the solution because it uses JavaScript, but I don't know how else I could fix it. I also changed the height of the messaging tab so there is no scrollbar unless there are many messages.

**Reason for changes**

#600

**Tests**

I was able to zoom up to 175% before the second scrollbar appeared.

### DIFF
--- a/popups/scratch-messaging/popup.css
+++ b/popups/scratch-messaging/popup.css
@@ -13,7 +13,7 @@ body {
   margin: 0 auto;
   font-family: "Sora", sans-serif;
   width: auto;
-  height: 500px;
+  height: 491px;
   overflow-x: hidden;
   overflow-y: scroll;
   background-color: #222;
--- a/popups/scratch-messaging/popup.css
+++ b/popups/scratch-messaging/popup.css
@@ -13,7 +13,7 @@ body {
   margin: 0 auto;
   font-family: "Sora", sans-serif;
   width: auto;
-  height: 491px;
+  height: 490px;
   overflow-x: hidden;
   overflow-y: scroll;
   background-color: #222;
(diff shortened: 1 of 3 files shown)


WorldLanguages:
Huh, it works - care to explain why?

WorldLanguages (approved):

WorldLanguages:
> Huh, it works - care to explain why?

If I wasn't clear enough - it would be cool to understand why this works?

mxmou:
> Huh, it works - care to explain why?

`100vh` doesn't work in popups the way one would expect, so I had to use JavaScript. `window.innerHeight` returns the height of the browser window or, in this case, the popup, but it has to be called after it loads, otherwise it would be 0.

I have no idea why the magic numbers I used have these values -- I just got them by testing.

WorldLanguages:
lol fine, it works, let's not touch it

WorldLanguages:
Unfortunately this does not work Firefox. There are scrollbars in both axis at 100% zoom level :(

WorldLanguages (changes requested):
Firefox support pending

mxmou:
@WorldLanguages Fixed it by making the popup 1px smaller.

WorldLanguages:
Lol, this PR is a joke, but thanks for taking the time to fix this

PR #1177, opened 2021-01-03
Implement easter egg addons

WorldLanguages:
This PR adds a new valid tag to addons, `easterEgg`. Addons with this tag will be considered part of the "easter egg" category, so there's no need to specify editor, website or theme.
Disabled easter egg addons are not shown unless users type in the konami code in the settings page. This key combination selects the invisible easterEgg category. The user can then select any other categories back if they wish.
This also fixes a settings page bug where `@click.ctrl` would not work properly, and still trigger even if ctrl isn't pressed.


RedGuy12:
So how do you see the easter eggs?

TheColaber:
look here: https://github.com/ScratchAddons/ScratchAddons/pull/1177/files#diff-716aae76bfb4eaa095810abaddc214bb23abc4a9ddc232f27cb3203233c1c0daR417-R428

PR #1193, opened 2021-01-04
better-quoter fixes

apple502j:
- fixes a crash
- fixes scrolling to wrong location
- fixes selection check (half of #1195)

### DIFF
--- a/addons/better-quoter/userscript.js
+++ b/addons/better-quoter/userscript.js
@@ -14,9 +14,7 @@ export default async function ({ addon, global, console }) {
         selection.focusNode &&
         blockpost.contains(selection.focusNode)
       )
-        textarea.value += `[quote=${
-          blockpost.querySelector(".black.username").innerText
-        }]${selectionStr}[/quote]`;
+        textarea.value += `[quote=${blockpost.querySelector(".black.username").innerText}]${selectionStr}[/quote]`;
       else copy_paste(blockpost.id);
       textarea.scrollIntoView(false);
       textarea.focus();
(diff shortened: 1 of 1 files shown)


WorldLanguages (approved):

addons/better-quoter/userscript.js
```diff
@@ -4,13 +4,19 @@ export default async function ({ addon, global, console }) {
     let quoteButton = await addon.tab.waitForElement(".postquote a", { markAsSeen: true });
     quoteButton.setAttribute("onclick", "return false");
     quoteButton.addEventListener("mouseup", (e) => {
-      let blockpost = e.path.find((e) => e.classList.contains("blockpost"));
-      if (window.getSelection().toString().length)
-        textarea.value += `[quote=${
-          blockpost.querySelector(".black.username").innerText
-        }]${window.getSelection().toString()}[/quote]`;
+      let blockpost = quoteButton.closest(".blockpost");
```
pufferfish101007, line 7: Can you add a `.querySelector(".post_body_html")` (should work) on the end of here so edit timestamps , siggies and user info can't be quoted? Thanks.

pufferfish101007 (changes requested):

RedGuy12:
I've wanted to quote siggies a few times tho....

pufferfish101007:
> I've wanted to quote siggies a few times tho....

That's understandable, but it shouldn't be possible to quote half siggies and half posts as it is currently. When I make my pr for correct bbcode quoting I'll add that in as an optional setting unless apple does it in this one first.

PR #1200, opened 2021-01-04
Enable/Disable All Addons

### LINKED ISSUE #318: Enable All Addons
It takes a while to enable every single extension, and it would be more helpful to add a button to enable them all at once.

Add a button to do that.
**Mockup**

image

### LINKED ISSUE #1211: "Collapse/expand all" button in settings page
Moved from #1207, suggestion 2

It would be great to add this button where the search bar is currently placed. If clicked, it will collapse all addons (enabled addons are expanded by default). After being clicked once, it will now be a "expand all" button.

RedGuy12:
**Resolves**

Resolves #318 
Resolves #1211

**Changes**

added enabling/disabling all addons
added expand/collapse all addons

**Reason for changes**

people want them

**Tests**

On Microsoft Edge Version 87.0.664.66 (Official Build) (64-bit)

Needs Firefox tests

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -550,16 +550,16 @@ document.addEventListener("keydown", (e) => {
 
 chrome.runtime.sendMessage("checkPermissions");
 
-function enableAll(){
+document.getElementById("enableAll").addEventListener("click", () => {
   Object.values(document.querySelectorAll('.switch[state=off]')).forEach(s=>s.click())
   Object.values(document.querySelectorAll('.reverted')).forEach(s=>s.click())
-}
+});
 
-function disableAll(){
+document.getElementById("disableAll").addEventListener("click", () => {
   Object.values(document.querySelectorAll('.switch[state=on]')).forEach(s=>s.click())
-}
+});
 
-function invertSelection(){
+document.getElementById("invertSelection").addEventListener("click", () => {
   Object.values(document.querySelectorAll('.switch')).forEach(s=>s.click())
   Object.values(document.querySelectorAll('.reverted')).forEach(s=>s.click())
-}
+});
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -549,3 +549,17 @@ document.addEventListener("keydown", (e) => {
 });
 
 chrome.runtime.sendMessage("checkPermissions");
+
+function enableAll(){
+  Object.values(document.querySelectorAll('.switch[state=off]')).forEach(s=>s.click())
+  Object.values(document.querySelectorAll('.reverted')).forEach(s=>s.click())
+}
+
+function disableAll(){
+  Object.values(document.querySelectorAll('.switch[state=on]')).forEach(s=>s.click())
+}
+
+function invertSelection(){
+  Object.values(document.querySelectorAll('.switch')).forEach(s=>s.click())
+  Object.values(document.querySelectorAll('.reverted')).forEach(s=>s.click())
+}
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -552,7 +552,6 @@ chrome.runtime.sendMessage("checkPermissions");
 
 document.getElementById("enableAll").addEventListener("click", () => {
   Object.values(document.querySelectorAll('.switch[state=off]')).forEach(s=>s.click())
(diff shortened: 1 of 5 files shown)


RedGuy12:
Wait #1200 #1200 #1200 #1200 #1200 Yayyyy

RedGuy12:
How do I run Prettier?

apple502j:
It doesn't seem like this PR is tested at all. Closing for now.

RedGuy12:
I did test it...

RedGuy12:
Oh wait I never looked at light mode

RedGuy12:
reopen this?

RedGuy12:
repoen this pls? i can fix the light mode bugs

RedGuy12:
Please reopen this, I can fix it. If you don't what to, please tell me _why_.

RedGuy12:
I fixed it!!

WorldLanguages:
This will likely conflict with #1006 

Moving to v1.8 milestone

webpages/settings/index.js
```diff
@@ -650,6 +650,41 @@ document.addEventListener("keydown", (e) => {
...
+    document.querySelectorAll('.addon-body:not([style="display: none;"]) .switch[state=off]')
+  ).forEach((s) => s.click());
+});
+
+document.getElementById("disableAll").addEventListener("click", () => {
+  Object.values(document.querySelectorAll('.addon-body:not([style="display: none;"]) .switch[state=on]')).forEach((s) =>
+    s.click()
+  );
+});
+
```
GrahamSH-LLK, line 664: Use Vue @click, not event listeners for all of these, not just the ones I have selected

RedGuy12:
Ok I dont know vue but I'll figure it out

WorldLanguages:
This works, but is not executed very well.
1. Ideally we'd directly use Vue data and methods to enable addons instead of triggering events on the DOM
2. These buttons should not appear in "popup mode" (addons tab in the extension popup)
3. "Enable all" is just a terrible idea - some addons require permission prompts, some are beta, some are very preference based. At most, "enable all recommended addons" would be fine, if the UI/UX is done right.
4. I don't see any use for "disable addons", I'd rather see users disable the whole extension so that they can keep their settings.
5. "Invert selected addons" is just bizarre
6. Expand/collapse all are fine, but I'd prefer a single button, to the left of the search bar. By default enabled addons are expanded, so this button should say "collapse all" unless all addons are collapsed (a computed property could be used to check the former)

RedGuy12:
> * Ideally we'd directly use Vue data and methods to enable addons instead of triggering events on the DOM

done, needs to be uploaded

> * These buttons should not appear in "popup mode" (addons tab in the extension popup)

done, needs to be uploaded

> * "Enable all" is just a terrible idea - some addons require permission prompts, some are beta, some are very preference based. At most, "enable all recommended addons" would be fine, if the UI/UX is done right.

uhhhhh see https://github.com/ScratchAddons/ScratchAddons/issues/318#issue-705827136 that's the main purpose of the issue...
plus I only have like 2 not enabled

> * I don't see any use for "disable addons", I'd rather see users disable the whole extension so that they can keep their settings.

bug testing - someone has a ton of addons enabled, but doesn't want any enabled for a minute to see which addons cause a certain bug

> * "Invert selected addons" is just bizarre

ok

> * Expand/collapse all are fine, but I'd prefer a single button, to the left of the search bar. By default enabled addons are expanded, so this button should say "collapse all" unless all addons are collapsed (a computed property could be used to check the former)

ok

WorldLanguages:
> uhhhhh see #318 (comment) that's the main purpose of the issue...
plus I only have like 2 not enabled

I wrote on https://github.com/ScratchAddons/ScratchAddons/issues/318#issuecomment-729225302 that it needed further discussion. I added it to the backlog because I considered it something to explore, not because the implementation in the first comment of the issue was ideal.
You're trying to tackle too many issues with this PR (mass enable functionality, its UI and collapse/expand all) when the first 2 haven't been discussed enough. Doing this with addons is usually fine (since 90% of addons are disabled by default), but changing the settings page affects all users.
Can you first try to work on the former first? I detailed how it should work in the previous comment (see 6)

RedGuy12:
Ok, I moved it to @click already and removed it from the popup. Lemme work on the others

RedGuy12:
will open new pr when i have time

PR #1214, opened 2021-01-05
Add refreshRequired to manifest

apple502j:
This changes the manifest to add `refreshRequired` which replaces `sa-autoupdate-theme-ignore`. It also shows notice on settings. `refreshRequired` should only be used for themes.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -244,5 +244,8 @@
   },
   "confirmImport": {
     "message": "Confirm"
+  },
+  "refreshRequired": {
+    "message": "Refresh the editor after enabling/disabling this theme."
   }
 }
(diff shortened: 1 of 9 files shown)


apple502j:
Note: merge ScratchAddons/manifest-schema#22 first

WorldLanguages:
Needs discussion, moving to v1.8 milestone in order to focus in other stuff in the meanwhile

WorldLanguages:
Instead of working on this, I'd like to work on userscript autorefresh (#783) and "on disabled" listener (#1303) directly, if possible - this is a middle step that could be avoided, and we could make refreshing directly unnecessary.

WorldLanguages:
Added those issues to v1.9.0 milestone - we could still merge this, but it's not worth it.

TheColaber:
I'd actually really like for this to be merged... I'm working on #1303 right now and it's almost done. Until everything transitions to #783 most addons will need refreshRequired for the addons manifest.

PR #1217, opened 2021-01-05
Play project muted

### LINKED ISSUE #395: Scratch 2.0 behavior: ctrl+greenflag to run the project muted
We could also allow users to mute the project after it started, and vice versa.

TheColaber:
Resolves #395

### DIFF
--- a/addons/60fps/userscript.js
+++ b/addons/60fps/userscript.js
@@ -18,6 +18,7 @@ export default async function ({ addon, global, console }) {
     let mode = false;
     const flagListener = (e) => {
       if (altPressedRecently) {
+        e.cancelBubble = true;
         e.preventDefault();
         mode = !mode;
         if (mode) setFPS(addon.settings.get("framerate"));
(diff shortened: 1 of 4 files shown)


YOYITsM3M8:
> done... I expect not a single comment, just a quick merge...

hmmmm ok

addons/60fps/userscript.js
```diff
@@ -18,6 +18,7 @@ export default async function ({ addon, global, console }) {
     let mode = false;
     const flagListener = (e) => {
       if (altPressedRecently) {
+        e.cancelBubble = true;
```
RedGuy12, line 21: why do you need this line?

TheColaber: This line will not allow the project to restart when the flag is pressed.

RedGuy12: oh i was just confused as it doesn't look like it's related to this addon

PR #1220, opened 2021-01-06
Iconfont

GrahamSH-LLK:
I did the impossible - I turned most of the SVGs into an icon font.
----
In seriousness:
- I skipped the popup b/c I didn't want conflicts from the popup PR, I'm going to wait for it to be merged
- I haven't finished a few other items

### Why this is helpful:
Right now, many of our icons look warped because we are using ``filter: brightness()`` to color them. Using an icon font allows us to easily color them with `` color: color``. It also allows us to use any icon, without downloading it.


Hans5958:
By the way, have you guys considered using `fill: [color]`?

GrahamSH-LLK:
> By the way, have you guys considered using `fill: [color]` on the SVG?
It doesn't work on non inline svgs.

WorldLanguages:
Instead of using an icon font, let's use better CSS that does not make SVGs look pixelated.

GrahamSH-LLK:
> Instead of using an icon font, let's use better CSS that does not make SVGs look pixelated.

In progress.

PR #1228, opened 2021-01-07
Better better quoting

### LINKED ISSUE #1195: Better quoter addon can quote any and all text on page, should include bbcode as well.
**Describe the bug**

When using the better-quoter addon, you can select any text on the page and then press any quote button, and the selected text will be quoted. In addition (this should probably be a separate issue), this addon should include the correct bbcode of the quoted text, as scratch's native quoter does. There should also be an optional option to automatically close bbcode tags when quoting, but again, that's another issue.

**To Reproduce**

1. Go to any non-closed forum topic
2. select any amount of text from anywhere on the page
3. click any quote button on the page
4. you will see that all of the selected text will be quoted, and there will be no bbcode from any posts you selected.

**Expected behavior**

It should either not do anything if you try to quote things outside of a post, just quote the original post or just quote the part that is selected inside the post.
The selected post's bbcode should be included in the quote.

**Screenshots**

- OS: windows 10
- Browser: chrome 87
- Version: sa 1.7.0-prerelease

**Additional context**

I should be able to fix this (if someone could assign me that would be nice). But there are a couple of options, either one of those as detailed in expected behaviour, or we could do what discourse does and add a little quote button next to the end of the selected text if it's all in one post, and use that to quote small parts, and the usual button to quote as normal. This makes the most sense to me.
Regarding the bbcode, we can fetch https://scratch.mit.edu/post/POST_ID/source to get the bbcode of it, and then figure out where about the person has selected it and then go from there... tbh that's not really viable but oh well. 
This will require a complete revamp of this addon.

pufferfish101007:
**Resolves**

Resolves #1195

**Changes**

When using the better-quoting addon, quoting now also includes bbcode (other than scratchblocks), and adds me to the credits

**Reason for changes**

They are useful and without it this addon is pretty useless (for me anyway)

**Tests**

tested on chrome. However, there may be some tags/scenarios that I haven't thought of and therefore cause bugs.

**additional info**
- scratchblocks aren't supported
- it uses innerhtml once, but doesn't set it, and there are no workarounds for it. Hopefully that will be fine. It also uses outerhtml a couple of times - I could probably change that but I'm too lazy. Hopefully that will be fine as it actually replaces elements with plain strings.
- you can still quote across quote, siggie, author info and date. Again, I'm too lazy to fix it.
- I'll try and sort out the conflicts, I'm not sure where they're coming from.
- sorry for having loads of commits again, I'm not sure where they came from either.


TheColaber:
You have a conflict...
image

addons/better-quoter/addon.json
```diff
@@ -5,6 +5,10 @@
     {
       "name": "TheColaber",
       "link": "https://scratch.mit.edu/users/TheColaber"
+    },
+    {
+      "name": "pufferfish101007",
+      "link": "https://scratch.mit.edu"
```
TheColaber, line 11: maybe add your profile link?

pufferfish101007:
@theColaber it appears to be doing a weird thing where it compares it to previous versions of my fork rather than the current version of this fork. I'll delete my current fork, make a new one and try again - that usually sorts it out.

addons/better-quoter/userscript.js
```diff
@@ -1,16 +1,160 @@
...
+      range = document.selection.createRange();
+      return range.htmlText;
+    } else if (window.getSelection) {
+      var selection = window.getSelection();
+      if (selection.rangeCount > 0) {
+        range = selection.getRangeAt(0);
+        var clonedSelection = range.cloneContents();
+        var div = document.createElement("div");
+        div.appendChild(clonedSelection);
+        return div.innerHTML;
```
TheColaber, line 14: I'm not sure if this is actually true... but the way you are doing this can cause Cross Scripting if someone puts a script tag for example... I'm sure the chrome reviewers might now like that very much...

TheColaber:
These reviews are just for future reference...

pufferfish101007:
@theColaber I'm pretty sure scratch escapes all of those so it should be fine. And anyway, it never attatches these things to the document, so even if it was vulnerable, it should all be fine?...

TheColaber:
I belive you are correct, but even so, I don't think the chrome reviewers would like it just because they can't control scratch.mit.edu and could turn around at any time

pufferfish101007:
Fair enough. Well, we'll see what they say but hopefully they'll be intelligent enough to see that scripts will never run.

PR #1230, opened 2021-01-07
Better better quoting

### LINKED ISSUE #1195: Better quoter addon can quote any and all text on page, should include bbcode as well.

pufferfish101007:
**Resolves**

Resolves #1195

**Changes**

When using the better-quoting addon, quoting now also includes bbcode (other than scratchblocks), and adds me to the credits

**Reason for changes**

They are useful and without it this addon is pretty useless (for me anyway)

**Tests**

Tested on chrome. However, there may be some tags/scenarios that I haven't thought of and therefore cause bugs.

**additional info**
- scratchblocks aren't supported
- it doesn't use innerHTML or outerHTML anymore
- you can still quote across quote, siggie, author info and date. Again, I'm too lazy to fix it.
- pls merge soon because I forgot to make a new branch so I'll have to make another fork again :)

### DIFF
--- a/addons/better-quoter/addon.json
+++ b/addons/better-quoter/addon.json
@@ -8,7 +8,7 @@
     },
     {
       "name": "pufferfish101007",
-      "link": "https://scratch.mit.edu"
+      "link": "https://scratch.mit.edu/users/pufferfish101007"
     }
   ],
   "userscripts": [
@@ -19,6 +19,12 @@
   ],
   "tags": ["forums", "community"],
   "l10n": true,
-  "warning": "Scratchblocks tags are not supported, and the bbcode may not be 100% accurate. To get fully accurate quotes, quote as normal and delete any parts that you don't need."
+  "info": [
+     {
+        "type": "notice",
+        "text": "Scratchblocks tags are not supported, and the bbcode may not be 100% accurate. To get fully accurate quotes, quote as normal and delete any parts that you don't need.",
+        "id": "scratchblocks_notice"
+      }
+    ]
 }
 
--- a/addons/better-quoter/addon.json
+++ b/addons/better-quoter/addon.json
@@ -20,11 +20,10 @@
   "tags": ["forums", "community"],
   "l10n": true,
   "info": [
-     {
-        "type": "notice",
-        "text": "Scratchblocks tags are not supported, and the bbcode may not be 100% accurate. To get fully accurate quotes, quote as normal and delete any parts that you don't need.",
-        "id": "scratchblocks_notice"
-      }
-    ]
+    {
+      "type": "notice",
+      "text": "Scratchblocks tags are not supported, and the bbcode may not be 100% accurate. To get fully accurate quotes, quote as normal and delete any parts that you don't need.",
+      "id": "scratchblocks_notice"
+    }
+  ]
 }
-
--- a/addons/better-quoter/addon.json
+++ b/addons/better-quoter/addon.json
@@ -5,6 +5,10 @@
     {
       "name": "TheColaber",
       "link": "https://scratch.mit.edu/users/TheColaber"
+    },
+    {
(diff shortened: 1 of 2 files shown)


addons/better-quoter/addon.json
```diff
@@ -14,5 +18,7 @@
     }
   ],
   "tags": ["forums", "community"],
-  "l10n": true
+  "l10n": true,
+  "warning": "Scratchblocks tags are not supported, and the bbcode may not be 100% accurate. To get fully accurate quotes, quote as normal and delete any parts that you don't need."
```
TheColaber, line 22: this was updated, use the system according to https://github.com/ScratchAddons/ScratchAddons/pull/957#issue-533087250

addons/better-quoter/addon.json
```diff
@@ -5,6 +5,10 @@
     {
       "name": "TheColaber",
       "link": "https://scratch.mit.edu/users/TheColaber"
+    },
+    {
+      "name": "pufferfish101007",
+      "link": "https://scratch.mit.edu"
```
TheColaber, line 11: link to your profile

addons/better-quoter/addon.json
TheColaber: and don't do warning, do notice.

GrahamSH-LLK:
Instead of using innerhtml etc. to figure out the tags, can't you use the API (or scratchdb)?

RedGuy12:
@mxmou suggested this because it would be faster

addons/better-quoter/addon.json
pufferfish101007: Woops thanks for letting me know

pufferfish101007:
> Instead of using innerhtml etc. to figure out the tags, can't you use the API (or scratchdb)?

Yes as @RedGuy12 said, this is a lot faster, and also it would be practically impossible to actually tell where the selection is in comparison to the source. So to answer your question, no. Unless we want slowness.

pufferfish101007:
Actually @GrahamSH-LLK I just realised that the innerHTML's only there for browser compatibility- I'll try removing it and see what happens, and while I'm at it I'll get rid of the outerhtmls.

WorldLanguages:
Moving to v1.8 milestone so that translators can focus on the new extension (not addon) locals, such as the new "popups" category and the update notification.

WorldLanguages:
Is this OK to merge?

pufferfish101007:
> Is this ok to merge?

Yes, it's all good to go.

PR #1232, opened 2021-01-07
Disable Autosave Addon

### LINKED ISSUE #1221: Disable Autosave Feature
I think there should be a "disable autosave" feature in Scratch Editor Features. It can help when a project autosaves and you made a mistake and you can't retrieve it.

RustingRobot:
**Resolves**

Resolves #1221

**Changes**

Adds an addon that can disable the autosave feature of the online editor. Pressing "Save Now" will still work. 

**Reason for changes**

While regular and automatic saves are mostly useful, there are also downsides to this. Already shared projects can't be updated in the online editor without the risk of saving an unfinished version. 

**Tests**

Tested in Chrome and Firefox

### DIFF
--- a/addons/disable-auto-save/userscript.js
+++ b/addons/disable-auto-save/userscript.js
@@ -1,45 +1,51 @@
-export default async ({ addon,console }) => { 
-    var manualPress = false;
-    // shamelessly stolen from animated-thumb/persistent-thumb.js
-    const xhrOpen = XMLHttpRequest.prototype.open;
-    XMLHttpRequest.prototype.open = function (method, path, ...args) {
-        if (method === "put" && String(path).startsWith("https://projects.scratch.mit.edu/")){
-            if(!manualPress) method = "OPTIONS";
-        }
-        return xhrOpen.call(this, method, path, ...args);
-    };
-    //initial adding of the observers and refreshing
-    addMutationListeners();
-    addon.tab.addEventListener("urlChange", function(event) { addMutationListeners(); });
-
-    async function addMutationListeners(){
-        //add a function to the "Save Now" button to disable request interception if it was a manual save
-        var saveContainer = await addon.tab.waitForElement("[class^='menu-bar_account-info-group']");
-        saveContainer = saveContainer.childNodes[0];
-        const saveObserver = new MutationObserver(function (mutations) {
-            if(saveContainer.childNodes[0]){
-                if(saveContainer.childNodes[0].className.startsWith("save-status_save-now")){
-                    manualPress = false;
-                    saveContainer.childNodes[0].addEventListener("click", function(){manualPress = true;})
-                }
-            }
-        });
-        saveObserver.observe(saveContainer, { childList: true, });
-        var alertContainer = await addon.tab.waitForElement("[class^='alerts_alerts-inner-container']");
-        //edgecase: exit editor and enter editor
-        if(addon.tab.editorMode == "editor" && alertContainer.childNodes[0] !== undefined){
-            alertContainer.childNodes[0].style.visibility = "hidden";
-        }
-        //hide the "can't save" warnings scratch gives us (don't worry, there won't be any zombie warnings :P)
-        const alertObserver = new MutationObserver(function (mutations) {
-            if(alertContainer.childNodes[0]){
-                if(manualPress){
-                    alertContainer.childNodes[0].style.visibility = "";
-                    alertContainer.childNodes[0].childNodes[1].childNodes[0].addEventListener("click", function(){manualPress = true;})
-                    manualPress = false;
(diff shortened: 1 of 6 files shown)


addons/disable-auto-save/userscript.js
```diff
@@ -0,0 +1,45 @@
...
+        const saveObserver = new MutationObserver(function (mutations) {
+            if(saveContainer.childNodes[0]){
+                if(saveContainer.childNodes[0].className == "save-status_save-now_2shdk"){
+                    manualPress = false;
+                    saveContainer.childNodes[0].addEventListener("click", function(){manualPress = true;})
+                }
+            }
+        });
+        saveObserver.observe(saveContainer, { childList: true, });
+        var alertContainer = await addon.tab.waitForElement("[class^='alerts_alerts-inner-container_1KTuF box_box_2jjDp']", {markAsSeen: true,});
```
TheColaber, line 28: these elements you are selecting always end with an underscore and random letters. These change with every update of scratch. Try to not include those in your selectors

addons/disable-auto-save/userscript.js
```diff
@@ -0,0 +1,45 @@
...
+        const alertObserver = new MutationObserver(function (mutations) {
+            if(alertContainer.childNodes[0]){
+                if(manualPress){
+                    alertContainer.childNodes[0].style.visibility = "";
+                    alertContainer.childNodes[0].childNodes[1].childNodes[0].addEventListener("click", function(){manualPress = true;})
+                    manualPress = false;
+                }else alertContainer.childNodes[0].style.visibility = "hidden";
+            }
+        });
+        alertObserver.observe(alertContainer, { childList: true, });
```
TheColaber, line 43: instead of this you could use a `await addon.tab.waitForElement` to make it easier?

TheColaber:
Good Job on your first PR!

addons/disable-auto-save/userscript.js
RustingRobot: Oh, didn't know that. Would something like this work?
```javascript
var saveContainer = await addon.tab.waitForElement("[class^='menu-bar_account-info-group']", {markAsSeen: true,});
        console.log(saveContainer);
        saveContainer = saveContainer.childNodes[0];
        const saveObserver = new MutationObserver(function (mutations) {
            if(saveContainer.childNodes[0]){
                if(saveContainer.childNodes[0].className.startsWith("save-status_save-now")){
                    console.log(saveContainer.childNodes[0].className);
                    manualPress = false;
                    saveContainer.childNodes[0].addEventListener("click", function(){manualPress = true;})
                }
            }
        });
        saveObserver.observe(saveContainer, { childList: true, });
        var alertContainer = await addon.tab.waitForElement("[class^='alerts_alerts-inner-container']", {markAsSeen: true,});
```

TheColaber: looks like it will! test it to be sure!

addons/disable-auto-save/userscript.js
RustingRobot: I could use waitForElement in line 38 to get the button if that's what you are suggesting but if I should replace the whole segment how would I test if I need to add the event listener again?

addons/disable-auto-save/userscript.js
RustingRobot: yup, it returns the nodes I need

WorldLanguages: > these elements you are selecting always end with an underscore and random letters. These change with every update of scratch. 

Wrong, dark modes have been using the same random letters since Scratch 3.0 release.
It's not a bad idea to do this, but saying that they change on every update is a lie :P

WorldLanguages:
Moving to v1.8 milestone so that translators can focus on the new extension (not addon) locals, such as the new "popups" category and the update notification.

WorldLanguages:
This addon will likely stay "beta" forever, even if it works as expected, to make sure users pay attention when enabling this addon.

WorldLanguages (approved):

Hans5958:
> This addon will likely stay "beta" forever, even if it works as expected, to make sure users pay attention when enabling this addon.

That's nonsense. It doesn't make sense if it's beta because the point of beta is for experimental and incomplete addons. Users should have to know what they are doing if they turned it on in the first place, so the beta tag is not needed.  A warning/notice should replace the beta tag.

WorldLanguages:
@Hans5958 The main issue I saw is someone may enable this addon from the popup settings and might not read the warning - in the popup, if you enable an addon it doesn't get extended unless you manually do so.
I'll change that behavior to make sure warnings are visible when enabling the addon in popup mode (so that the addon item gets extended if it has warnings) and remove the beta tag. We should also consider showing the "beta" tag in popup mode, we currently hide all tags in that view.

PR #1265, opened 2021-01-10
Remove color setting from Remix Tree addon

### LINKED ISSUE #1262: Remove color setting from remix tree button

The remix tree button addon has a color setting. I don't think many people use it (other addons that add buttons don't have one), and it means that people who change the primary color in 2.0 → 3.0 will probably want to change the remix tree button color too.

The addon should not have a color setting, the color would only be controlled by 2.0 → 3.0.

TheColaber:
Resolves #1262


TheColaber:
huh... extra commits... let me try again

PR #1311, opened 2021-01-13
Remove search-load-more-fix addon

### LINKED ISSUE #907: Hide "Fix Load More in Search Results" in Firefox
This bug only applies to Chrome.

apple502j:
**Resolves**
Resolves #907 

**Changes**
Removes search-load-more-fix addon as the issue is fixed browser-side on Chrome (and presumably on Chromium Edge as well). The issue never existed on Firefox.

The issue was fixed on Scratch side as well, althought it only affects old Edge.


mxmou:
> fixed browser-side on Chrome (and presumably on Chromium Edge as well)

Tested on Edge, it's fixed.

PR #1312, opened 2021-01-13
Remove notice for editor stepping addon

TheColaber:
The addon was misnamed anyway so had no effect for a while... if you go to the current version or the released version of SA you won't see it anyway... The message is for 1.3.0 and we are on 1.8.0... it's time.


GrahamSH-LLK:
Quik merge?

TheColaber:
> Quik merge?

up to you?

PR #1319, opened 2021-01-14
Auto-hide block palette addon

TheColaber:
**Resolves**

Requested in #1156

**Changes**

I wrote the addon.

**Tests**

Tested and works


addons/hide-flyout/addon.json
```diff
@@ -0,0 +1,25 @@
...
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+
+  "tags": ["editor"],
```
WorldLanguages, line 23: Recommended

addons/hide-flyout/userscript.js
```diff
@@ -0,0 +1,26 @@
+export default async function ({ addon, global, console }) {
+  while (true) {
+    let flyOut = await addon.tab.waitForElement(".blocklyFlyout", { markAsSeen: true });
+    let state = true;
+    function mouseMove(e) {
+      let sidePos = document.querySelector(".blocklyToolboxDiv").getBoundingClientRect().x;
+      let scrollBar = document.querySelector(".blocklyFlyoutScrollbar");
+      if (e.clientX - sidePos < 310) {
```
WorldLanguages, line 8: What's 310? This feels like it will change screen to screen.

TheColaber: I think that it should be the same everywhere... This is the width of the flyout.

TheColaber:
Ready to merge

TheColaber:
Having a problem currently... Deleting blocks doesn't work! It used to... interesting...

WorldLanguages (approved):

CuriousG2:
I tested this, but it seems to have a lot of bugs:

1. The animation seems to go over the stage when the stage is on the left, and you are moving a block.

2. When the stage is on the left in small mode the lock displays in the wrong place.

3. You can see the lock on the project page.

4. Going to the project page and back into the project breaks it.

5. When you go outside the project then go back in you can see the lock under the sounds and costumes tabs.

6. For some reason it creates a scroll bar on the right side of the screen.

This happens on Firefox, Chrome, and Edge.

TheColaber:
Thanks for this @CuriousG2 I'll fix all these asap.

PR #1332, opened 2021-01-15
Add a11y addon

easrng:
**Changes**

Add a11y addon that adds a category to the toolbox with blocks to allow people to add WAI-ARIA roles and labels to their sprites.
Two scratch blocks. The first one has the text "set sprite aria role to" then an input with the text "button" in it. The second one has the text "set sprite label to" then an input with the text "click me" in it.

**Reason for changes**

Scratch projects aren't accessible to people using assistive technologies. I think that is an issue.

**Tests**

I have tested it and fixed all bugs I've found.

### DIFF
--- a/addons/a11y/addon.json
+++ b/addons/a11y/addon.json
@@ -1,7 +1,6 @@
 {
   "name": "Accessibile Projects",
   "description": "Adds support for WAI-ARIA and custom cursors to the Scratch player.",
-  "info": [],
   "credits": [
     {
       "name": "A-E-"
@@ -19,7 +18,7 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "tags": ["beta"],
+  "tags": ["beta","editor"],
   "enabledByDefault": false,
   "l10n": true
 }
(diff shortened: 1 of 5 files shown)


pufferfish101007:
The schema validation doesn't like the empty info array in the addon.jsom - you should delete that.

addons/a11y/addon.json
```diff
@@ -0,0 +1,25 @@
...
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["beta"],
```
pufferfish101007, line 22: This should include 'editor' as well.

pufferfish101007 (changes requested):

apple502j:
This is rejected, as this essentially adds new blocks.

WorldLanguages:
Reopening for now, might need discussion

jeffalo:
the custom block colour addon should show this

PR #1334, opened 2021-01-15
Better better better quoting

apple502j:
Fixes linter failure and potential xss


pufferfish101007:
Thanks, I didn't know insertadjacenttext existed
Also this means when scratchblocks support is added it will be better better better better quoting

PR #1345, opened 2021-01-17
Replace linkifyjs with a regex

### LINKED ISSUE #1304: "More links" addon lags the page
Linkify appears to block the main thread.

apple502j:
Resolves #1304 

Originally I tried to put everything in a Web Worker so that it doesn't block the main thread - but if we gave up some small functionalities that linkifyjs provides, it's actually easier to just make our own linkifier. I still kept the code for loading a worker, just in case it's useful in the future.

Note that there is a small chance that this can cause more lag (google "catastrophic backtracking") but I don't think it can be triggered easily.


WorldLanguages:
IIRC linkify detects "google.com" as a link, I doubt this version does. If we want to still provide this functionality we may need to have an array of valid TLDs.

apple502j:
@WorldLanguages This detects any `xxx.xx` - there will inevitably be some false positives but saves disk.

WorldLanguages (approved):

PR #1347, opened 2021-01-18
mediarecorder: improve UI and add options

apple502j:
- Bumped the max limit to 120 secs
- Added options UI
- Added "Include project sounds" option, enabled by default
- Added "Do not start recording until the Green Flag is clicked" option, enabled by default
- Fixed addon description
- Un-"beta"-ified

UI is a mix of editor-devtools and vanilla.

image


WorldLanguages (changes requested):
This is an improvement but I'd like to see some more things:
1. add a red dot icon next to "start recording" to make it more noticeable (?)
2. i'd prefer "click green flag after start recording" rather than the current behavior
3. can we add how many seconds have passed, and how many are left?
4. automatically stop recording on project stop, as an option?
5. "record duration" input type number does not look like an input unless you hover it, not amazing UX
6. "record the stage as a WebM file" should be ".webm" not "WebM", to be consistent with the addon description
7. "you can save it on your computer", use "to" instead of "on"

apple502j:
Finalized? UI

image

Secret-chest:
> This is an improvement but I'd like to see some more things:
> 
> 1. add a red dot icon next to "start recording" to make it more noticeable (?)
> 2. i'd prefer "click green flag after start recording" rather than the current behavior
> 3. can we add how many seconds have passed, and how many are left?
> 4. automatically stop recording on project stop, as an option?
> 5. "record duration" input type number does not look like an input unless you hover it, not amazing UX
> 6. "record the stage as a WebM file" should be ".webm" not "WebM", to be consistent with the addon description
> 7. "you can save it on your computer", use "to" instead of "on"

There should be an option to move Start Recording in File or Edit, it takes so much space

PR #1389, opened 2021-01-23
fix uploading image message

### LINKED ISSUE #1387: uploading images with dark mode
**Describe the bug**

with website dark mode and forum image uploader you can't read the "uploading" notice well
additionally, it looks weird that it is not the same height as the bbcode buttons

**To Reproduce**

enable dark mode and forum image uploader
upload an image to the forums
look at the uploading notice

**Expected behavior**

the text to be black

**Screenshots**

image

### LINKED ISSUE #1388: curator link dosen't work
**Describe the bug**

the curator link addon doesn't linkify the username

I get this error:

```
Uncaught (in promise) TypeError: Cannot read property 'splash.projectsCuratedBy' of undefined
    at EventTarget.scratchMessage (Tab.js:116)
    at Module.default (script.js:14)
scratchMessage @ Tab.js:116
default @ script.js:14
async function (async)
default @ script.js:5
loadUserscript @ run-userscript.js:20
async function (async)
loadUserscript @ run-userscript.js:16
window.addEventListener.once @ run-userscript.js:30
```

**To Reproduce**

go to the front page
see error

**Expected behavior**

the text to be linked

RedGuy12:
**Resolves**

Resolves #1387

**Changes**

added a class to the upload message
added css to the upload message dark mode

**Reason for changes**

because its a bug...
**Tests**

it works

### DIFF
--- a/addons/dark-www/experimental_forums.css
+++ b/addons/dark-www/experimental_forums.css
@@ -85,8 +85,8 @@
   background-color: #606060;
   color: transparent;
 }
-.markItUpHeader > ul > li:hover:not(.uploadStatus) {
-  background-color: #fff;
+.markItUpHeader > ul > li.uploadStatus {
+  background: transparent !important;
 }
 .markItUpPreviewFrame {
   border-color: #606060 !important;
--- a/addons/dark-www/experimental_forums.css
+++ b/addons/dark-www/experimental_forums.css
@@ -92,3 +92,9 @@
   border-color: #606060 !important;
   background: transparent;
 }
+
+.uploadStatus {
+  height: 19px;
+  color: black;
+  padding: 1.5px;
+}
--- a/addons/dark-www/experimental_forums.css
+++ b/addons/dark-www/experimental_forums.css
@@ -85,7 +85,7 @@
   background-color: #606060;
   color: transparent;
 }
-.markItUpHeader > ul > li:hover {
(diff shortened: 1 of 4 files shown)


addons/dark-www/experimental_forums.css
```diff
@@ -92,3 +92,9 @@
   border-color: #606060 !important;
   background: transparent;
 }
+
+.uploadStatus {
+  height: 19px;
+  color: black;
+  padding: 1.5px;
```
WorldLanguages, line 99: I don't think this is the right way to do it, why are you specifying the height and padding in dark mode css?

RedGuy12: because you don't notice it unless you are in dark mode. it only has a background in dark mode

mxmou: A better fix would be removing the background.

RedGuy12: @mxmou done

mxmou:
This didn't work correctly after the last change. It should be fixed now.

RedGuy12:
Huh, why did we even have that line if it was always overridden by line 64?

mxmou:
@RedGuy12 I removed it now. It worked in the original userstyle but when porting it to Scratch Addons I forgot to add `!important` to it.

WorldLanguages:
@mxmou I'm unfamiliar with the status of this PR, please decide yourself whether to merge or not

PR #1411, opened 2021-01-26
Add Show BBCode addon

### LINKED ISSUE #1219: Show BBCode
**Describe the new addon**

A button that shows the BBCode of a post.

mxmou:
**Resolves**

Resolves #1219

**Changes**

Adds a button that shows the BBCode of a post.

### DIFF
--- a/addons/show-bbcode/userscript.js
+++ b/addons/show-bbcode/userscript.js
@@ -1,0 +1,58 @@
+function viewSource(post, msg) {
+  return function(event) {
+    event.preventDefault();
+    const body = post.querySelector(".postmsg");
+    if (event.target.getAttribute("data-state") === "post") {
+      event.target.innerText = msg("source-button-active");
+      event.target.removeAttribute("title");
+      if (event.target.originalHTML === undefined) {
+        event.target.originalHTML = body.firstElementChild;
+      }
+      body.removeChild(body.firstElementChild);
+      const source = document.createElement("div");
+      body.insertBefore(source, body.firstElementChild);
+      source.className = "post_body_html";
+      if (event.target.sourceText !== undefined) {
+        event.target.setAttribute("data-state", "source");
+        source.innerText = event.target.sourceText;
+        return;
+      }
+      event.target.setAttribute("data-state", "loading");
+      source.innerText = msg("loading");
+      fetch("https://scratch.mit.edu/discuss/post/" + post.id.substring(1) + "/source/").then(function(res) {
+        res.text().then(function(text) {
+          event.target.setAttribute("data-state", "source");
+          source.innerText = event.target.sourceText = text;
+        });
+      });
+    } else if (event.target.getAttribute("data-state") === "source") {
+      event.target.innerText = msg("source-button");
+      event.target.title = msg("source-button-tooltip");
+      event.target.setAttribute("data-state", "post");
+      body.removeChild(body.firstElementChild);
+      body.insertBefore(event.target.originalHTML, body.firstElementChild);
+    }
+  };
+}
+
+export default async function({ addon, console, msg }) {
+  for (const post of document.querySelectorAll(".blockpost")) {
+    const actionRow = post.querySelector(".postfootright ul");
+    const sourceItem = document.createElement("li");
+    const quoteItem = actionRow.querySelector(".postquote");
+    if (quoteItem) {
+      actionRow.insertBefore(sourceItem, quoteItem);
+    } else {
+      actionRow.appendChild(sourceItem);
+    }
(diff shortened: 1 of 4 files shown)


addons/show-bbcode/userscript.js
```diff
@@ -0,0 +1,58 @@
...
+      event.target.title = msg("source-button-tooltip");
+      event.target.setAttribute("data-state", "post");
+      body.removeChild(body.firstElementChild);
+      body.insertBefore(event.target.originalHTML, body.firstElementChild);
+    }
+  };
+}
+
+export default async function ({ addon, console, msg }) {
+  for (const post of document.querySelectorAll(".blockpost")) {
```
pufferfish101007, line 39: You should use while (true) {const post = await addon.tab.waitForElement(".blockpost")...} like in better-quoter so that when load more is enabled the newly loaded posts get it added as well

addons/show-bbcode/userscript.js
```diff
@@ -0,0 +1,58 @@
...
+      body.insertBefore(source, body.firstElementChild);
+      source.className = "post_body_html";
+      if (event.target.sourceText !== undefined) {
+        event.target.setAttribute("data-state", "source");
+        source.innerText = event.target.sourceText;
+        return;
+      }
+      event.target.setAttribute("data-state", "loading");
+      source.innerText = msg("loading");
+      fetch("https://scratch.mit.edu/discuss/post/" + post.id.substring(1) + "/source/").then(function (res) {
```
pufferfish101007, line 22: Wouldn't it be better to store this somewhere, so you don't have to have a slow request every single time?

addons/show-bbcode/userscript.js
```diff
@@ -0,0 +1,58 @@
+function viewSource(post, msg) {
+  return function (event) {
+    event.preventDefault();
+    const body = post.querySelector(".postmsg");
+    if (event.target.getAttribute("data-state") === "post") {
```
pufferfish101007, line 5: Shouldn't you check to see if it's already loading so you don't get multiple requests/dom changes happening at once?

pufferfish101007 (changes requested):

addons/show-bbcode/userscript.js
mxmou: Thanks, I'll change that.

addons/show-bbcode/userscript.js
mxmou: If it's already loading then the state would be `loading`, not `post`.

addons/show-bbcode/userscript.js
mxmou: It's stored as `event.target.sourceText`. See lines 15 - 19.

addons/show-bbcode/userscript.js
mxmou: Infinite scrolling doesn't currently support topics, so this shouldn't be a problem, but I changed it anyway.

addons/show-bbcode/userscript.js
pufferfish101007: But there's nothing stopping you from clicking the button _while_ the state is loading.

addons/show-bbcode/userscript.js
pufferfish101007: Ah, ok, I understand now. I didn't see that earlier.

WorldLanguages (approved):
Looks good

WorldLanguages:
Just one question, cc @pufferfish101007, couldn't this method be used for "better forum quoter" in order to have completely accurate bbcode?

mxmou:
> Just one question, cc @pufferfish101007, couldn't this method be used for "better forum quoter" in order to have completely accurate bbcode?

No. There is no way to know which part of the BBCode corresponds to the selection.

pufferfish101007:
> Just one question, cc @pufferfish101007, couldn't this method be used for "better forum quoter" in order to have completely accurate bbcode?

Potentially, but people didn't want this (I think it was you, or maybe mxmou) because we didn't want to make requests, as it's just so slow. In addition, I don't really know how it would be possible to somehow match up the selected text with the correct part of the source. I'll think about it very deeply and see if I can think of a way to match them up, but for now the answer's no.

mxmou:
> I'll think about it very deeply and see if I can think of a way to match them up, but for now the answer's no.

We could make a BBCode parser that would remember where in the BBCode HTML elements were, compare its output to the actual post, then use the original BBCode if it matches or the current method if it doesn't (if the post was edited after the page was loaded). But this is slower than the current method, much more complex and not much better.

PR #1419, opened 2021-01-26
New mode for "Auto-hide block palette" addon: Hover in category

### LINKED ISSUE #1404: New mode for "Auto-hide block palette" addon: Hover in category
**Describe the bug**

Auto-hide block palette is currently useless (tested in Firefox), because an invisible element still obstructs the editor when the palette is hidden. This means you can't meaningfully use the screen space gained from the hidden sidebar:

- When in "hover" mode, the sidebar will appear on the screen as soon as you hover the space it would normally occupy
- When in "click" mode, there is a dead space on the space the sidebar is supposed to be.

**To Reproduce**

1. Enable the extension
2. Go to the editor

**Expected behavior**

I guess the sensing space of the sidebar should be update when it hidden

- OS: Linux Mint
- Browser Firefox
- Version 84

**Additional context**

This idea is really amazing since Scratch 3 made everything so big, this is vital for people with small screens like me :smile:

TheColaber:
**Resolves**

Resolves #1404

**Changes**

New setting option: Hover in category

**Reason for changes**

mxmou wants it and it's not to bad :)

**Tests**

It's epic as always probably ;)

### DIFF
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -56,6 +56,10 @@
         {
           "id": "category",
           "name": "Category click"
+        },
+        {
+          "id": "cathover",
+          "name": "Category hover"
         }
       ],
       "default": "hover"
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -7,6 +7,13 @@
       "link": "https://scratch.mit.edu/users/TheColaber"
     }
   ],
+  "info": [
+    {
+      "type": "notice",
+      "text": "\"Palette area hover\" mode only extends the viewing area. If you want to be able to drag blocks into that area without them getting trashed, use one of the other modes.",
+      "id": "hoverExplanation"
+    }
+  ],
   "userscripts": [
     {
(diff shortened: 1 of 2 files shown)


addons/hide-flyout/addon.json
```diff
@@ -56,6 +56,10 @@
         {
           "id": "category",
           "name": "Category click"
+        },
+        {
+          "id": "cathover",
+          "name": "Category hover"
```
mxmou, line 62: It would make more sense before Category click.

addons/hide-flyout/addon.json
```diff
@@ -53,6 +53,10 @@
           "id": "hover",
           "name": "Hover"
```
WorldLanguages, line 54: Maybe this could have a better name now that "category hover" is also an option? What about `Block palette area hover` or similar?
These options can be confusing, so maybe we should add notices that better explain what each option does?

TheColaber: I agree. I'm not so good with names, so I'll probably not have the best ideas.

PR #1447, opened 2021-01-28
Introduce addon.tab.scratchClass

WorldLanguages:
`addon.tab.scratchClass` is a handy function that should make it easier for us to be consistent with Scratch UI.
It accepts multiple arguments, for example 
```js
addon.tab.scratchClass("modal_header-item", "modal_header-item-title")
```
returns `modal_header-item_2zQTd modal_header-item-title_tLOU5`.

It also accepts an option object, that allows us to merge Scratch and non-Scratch classes automatically, for example: 
```js
addon.tab.scratchClass("modal_modal-content", { others: "mediaRecorderPopup" })
```
 returns `modal_modal-content_1h3ll mediaRecorderPopup`. (`others` can be either a string or array)

If an addon calls this function before the classes are available, they temporarily fallback to `scratchAddonsScratchClass/[classNameToFind]`, and after the classes load, all elements with these get their classes replaced accordingly.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -150,4 +150,29 @@ export default class Tab extends Listenable {
     const worker = new Worker(workerURL);
     return new Promise((resolve) => worker.addEventListener("message", () => resolve(worker), { once: true }));
   }
+
+  scratchClass(...args) {
+    let res = "";
+    args
+      .filter((arg) => typeof arg === "string")
+      .forEach((classNameToFind) => {
+        if (scratchAddons.classNames.loaded) {
+          res +=
+            scratchAddons.classNames.arr.find(
+              (className) =>
+                className.startsWith(classNameToFind + "_") && className.length === classNameToFind.length + 6
+            ) || "";
+        } else {
+          res += `scratchAddonsScratchClass/${classNameToFind}`;
+        }
+        res += " ";
+      });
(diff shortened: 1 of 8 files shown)


mxmou:
What would it return for `input_input-form`? Scratch has two classes with that name, `input_input-form_l9eYg` (in scratch-gui) and `input_input-form_1Y0wX` (in scratch-paint).

WorldLanguages:
Unfortunately `input_input-form_1Y0wX` comes first - let me think what to do

TheColaber:
> Unfortunately `input_input-form_1Y0wX` comes first - let me think what to do

maybe it should return an array of all findable classes 

instead of find

https://github.com/ScratchAddons/ScratchAddons/pull/1447/files#diff-e7de085f705ea26bf41689817c917b9e7d2ade4198bb9c8801b1b54cb7d78834R161

use filter

WorldLanguages:
Looks like there are 5 duplicates:
image

WorldLanguages:
I'll use the hashed class here, looks like TurboWarp has the exact same class name so it should be fine

WorldLanguages:
Looks like all the duplicates from scratch-paint are a single stylesheet, so we'll just ignore that stylesheet.

WorldLanguages:
No more duplicates in the editor

PR #1489, opened 2021-02-02
Icon.js

GrahamSH-LLK:
This uses my library (icon.js) for icons


libraries/icon.js
```diff
@@ -0,0 +1,21 @@
...
+      let set = item.getAttribute("data-set");
+      let icon = item.getAttribute("data-icon");
+      let json;
+      if (localStorage.getItem(`${set}-${icon}`) === null) {
+        let data = await fetch(`http://api.iconify.design/${set}.json?icons=${icon}&pretty=1`);
+        json = JSON.parse(await data.text());
+      } else {
+          json = JSON.parse(localStorage.getItem(`${set}-${icon}`));
+      }
+      item.outerHTML = `<svg class="${item.getAttribute("class")}" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" aria-hidden="true" focusable="false" width="1em" height="1em" style="-ms-transform: rotate(360deg); -webkit-transform: rotate(360deg); transform: rotate(360deg);" preserveAspectRatio="xMidYMid meet" viewBox="0 0 ${json.width} ${json.height}">${json.icons[icon].body}</svg>`
```
GarboMuffin, line 18: This is a massive security issue, especially because of the `http://` in the iconify URL which makes a man-in-the-middle attack trivial. What is there to prevent the API from returning an SVG with a <script> that steals user credentials? Even with https://, MITM is still possible or Iconify themselves could be hacked.

What's wrong with storing icons in the extension?

Vaguely on topic: does anyone know where Iconify's privacy policy is? I can't seem to find it so I can only assume that they're selling everything they gather to the Russian government.

WorldLanguages:
Yeah, nice on creating a library but Scratch Addons will keep storing its icons locally, there's really no reason not to. The kilobytes taken by the icons are insignificant compared to all the locale files. If we wanted to optimize on extension size, we should focus on those, not icons. And apart from that, I can't think of any other reasons to load icons dynamically from the internet each time.

libraries/icon.js
TheColaber: Well said garbo... I love every part of it. Especially this...
> I can't seem to find it so I can only assume that they're tracking everything possible and selling it to the Russian government.

apple502j:
Note that XSS is not an issue (because of strict CSP) and those who can MITM in https session should be able to sniff Scratch API request as well.

GrahamSH-LLK:
I upgraded this to https.
(This is mostly for testing a webhook)

GrahamSH-LLK:
Hear me out. I've made it so that icons are stored in a JS file instead of fetching them. There isn't a remote security vuln here. This will give us the benefits of 
A. Having cleaner icons (inline svgs can be easily colored, without warping)
B. Easier sizing of icons
C. Being able to use any icon without resizing, and using custom icons easily.

apple502j:
At least remove unnecessary icons before reopening.

PR #1494, opened 2021-02-03
Scratch Messaging: use v-if instead of v-show to fix CSS issue

WorldLanguages:
lol i keep tinkering with the lovers & favers tooltip


thecoder876:
lol

PR #1524, opened 2021-02-08
Bump to prerelease 1.10.0

GrahamSH-LLK:
What do you _think_ this is?
<sub><sup>(Rhetorical Question)</sup></sub>

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,8 +2,8 @@
   "manifest_version": 2,
   "name": "__MSG_extensionName__",
   "description": "__MSG_extensionDescription__",
-  "version": "1.9.0",
-  "version_name": "1.9.0",
+  "version": "1.10.0",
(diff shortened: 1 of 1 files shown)


TheColaber:
not evil code

WorldLanguages:
nuuu where's v2.0.0

TheColaber:
someone do something great for a 2 point o

GrahamSH-LLK:
> nuuu where's v2.0.0

@WorldLanguages I was tempted to make this PR change every single bit of orange to blue and call it v2 😛

TheColaber:
> > nuuu where's v2.0.0
> 
> @WorldLanguages I was tempted to make this PR change every single bit of orange to blue and call it v2 😛

now truly we could have an easter egg that does that

PR #1525, opened 2021-02-08
Does this fix it????

GrahamSH-LLK:
**Resolves**

Resolves #

**Changes**

**Reason for changes**

**Tests**


GrahamSH-LLK:
This wasn't spam. I was seeing if using an unminified version of intl-messageformat would fix our bug.

PR #1529, opened 2021-02-09
v1.9.x

apple502j:
Rename icon for manager notification


RedGuy12:
You just had to rename a file? Why was that a problem? Is `studio-add.svg` a problem too?

PR #1551, opened 2021-02-11
Improve message when thumb server is down

apple502j:
This also adds a way for msg() to get messages from `_general.json`.


WorldLanguages (changes requested):
I'd like to discuss why these messages are on _general.json? I don't understand

PR #1555, opened 2021-02-11
Improve pause addon

GarboMuffin:
**Changes**

 - remove beta tag
 - remove unnecessary console.log
 - fix pausing blocks like "wait 10 seconds"
 - keep glowing paused scripts (debatable whether this is good or not, I think it is)
 - fix pausing some [weird scripts](https://user-images.githubusercontent.com/33787854/107668115-ebdbc280-6c55-11eb-8d8c-5597a5676929.png)
 - fix pause not resuming old threads if the user starts a new thread while paused
 - "when sprite clicked" and "when key pressed' events should not be fired when paused
 - don't fall apart when sa-pause is run when already paused

### DIFF
--- a/addons/pause/addon.json
+++ b/addons/pause/addon.json
@@ -25,7 +25,7 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "tags": ["editor", "beta"],
+  "tags": ["editor"],
   "enabledByDefault": false,
   "l10n": true
 }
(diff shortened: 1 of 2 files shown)


addons/pause/addon.json
```diff
@@ -25,7 +25,7 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "tags": ["editor", "beta"],
+  "tags": ["editor"],
```
WorldLanguages, line 28: Recommended?

WorldLanguages: Scratch Addons only note: let's also mark this addon as new and feature it in the release.

PR #1563, opened 2021-02-12
Fix "custom Colors addon makes devtools highlighting not work"

### LINKED ISSUE #1545: Custom Colors addon makes devtools highlighting not work
Custom Colors addon makes devtools highlighting not work. It has some css somewhere that has !important or something that makes highlighting not appear.

RedGuy12:
**Resolves**

Resolves #1545

**Changes**

The Blockly method just edited the `fill` attribute, which is overridden by the stylesheet. This way edits the `style` attribute, which overrides everything.

**Reason for changes**

it's a bug

**Tests**

it works

### DIFF
--- a/addons/editor-devtools/blockly/BlockFlasher.js
+++ b/addons/editor-devtools/blockly/BlockFlasher.js
@@ -22,7 +22,7 @@ export default class BlockFlasher {
      * @private
      */
     function _flash() {
-      myFlash.block.setColour(flashOn ? "#ffff80" : myFlash.colour);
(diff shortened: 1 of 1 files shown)


WorldLanguages (approved):

RedGuy12:
oof srry I forgot to update the title

griffpatch:
I've added a fix that was causing previously highlighted blocks to stay highlighted (because only one of the colour changing scripts was addressed in the original fix)

PR #1567, opened 2021-02-13
Trim empty lines on replies to ST comments

### LINKED ISSUE #1444: Comments linebreaks spacing issues
- When you reply to an ST member, the first text node starts with an asterisk and the following whitespace is not trimmed.

- lf a comment contains a link, a space is ***not*** added after it, whether or not there was one.

- If a comment contains a link, a space is added before it, whether or not there was one.

- Any linebreaks before/after a link are not preserved.

apple502j:
Resolves some of #1444

Found while fixing lint errors.


RedGuy12:
None of the things I mentioned in #1444 were fully fixed. I'm going to experiment to figure out when it happens now.

PR #1570, opened 2021-02-13
fix unreadable crash message with dark-www

RedGuy12:
sometimes scratch crashes and it shows a error message

with dark-www, it looks like this: 
Screenshot 2021-01-29 103611
unreadable

**Changes**

update dark-www.css, now it looks like this:
Screenshot 2021-01-29 103638

**Reason for changes**

it's a bug

**Tests**

works

### DIFF
--- a/addons/dark-www/dark-www.css
+++ b/addons/dark-www/dark-www.css
@@ -652,3 +652,8 @@ img.tips-icon {
 .tips-divider {
   border-top-color: #000;
 }
+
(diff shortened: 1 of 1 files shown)


WorldLanguages:
@mxmou We did a whoopsie here. We added this to dark-www.css which is unused. Could you add this to the correct file?

RedGuy12:
Oops me

PR #1608, opened 2021-02-17
Fix matches for image-uploader 

### LINKED ISSUE #1605: image-uploader should only run on topics
Error Found - When at https://scratch.mit.edu/discuss/ from image-uploader addon
image

TheColaber:
Resolves #1605

### DIFF
--- a/addons/image-uploader/addon.json
+++ b/addons/image-uploader/addon.json
@@ -9,9 +9,7 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": [
-        "https://scratch.mit.edu/discuss/*"
-      ]
+      "matches": ["https://scratch.mit.edu/discuss/*"]
     }
   ],
   "tags": ["forums", "community"],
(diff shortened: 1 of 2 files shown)


pufferfish101007:
What about creating a,new topic, or editing? This should also include https://scratch.mit.edu/discuss/post/*/edit and  https://scratch.mit.edu/discuss/*/new (I think it's that for new topics)

jeffalo:
better solution would be to check if all of the right elements are there in the userscript. that way you dont need to worry about all of the url matches and it will work on every forum page

WorldLanguages:
> better solution would be to check if all of the right elements are there in the userscript. that way you dont need to worry about all of the url matches and it will work on every forum page

"better", kinda, but good enough

TheColaber:
Note to you @WorldLanguages I did change it already and I think jeffalo confirms.

WorldLanguages (approved):
Well, this is fine but it's also fine if it throws an error. Many addons already do this if a project is a 404 and in so many other situations.

TheColaber:
> Well, this is fine but it's also fine if it throws an error. Many addons already do this if a project is a 404 and in so many other situations.

We already have soooooooo many errors in the console... might as well reduce them.

PR #1614, opened 2021-02-17
Add profile statistics addon

### LINKED ISSUE #494: ScratchStats
**Describe the new addon**

This addon would add scratchstats info to profiles. Each item, like Followers stats, Followers over time (beta), etc., could be toggled on and off. This would be really cool.
**Link to the new addon**

scratchstats.com

mxmou:
**Resolves**

Resolves #494

**Changes**

Adds an addon I made a few weeks ago that adds data from ScratchDB to profile pages and displays it in a similar way to ScratchStats. It isn't customizable yet.

**Tests**

image


addons/scratchr2/profile.css
```diff
@@ -98,25 +99,25 @@
...
   border: none;
   background: rgba(77, 151, 255, 0.1);
   padding-top: 16px;
   padding-bottom: 16px;
   margin-bottom: 0;
 }
-#content > .box:not(#profile-data):not(.slider-carousel-container) .box-head,
-#content > .box:not(#profile-data):not(.slider-carousel-container) .box-content {
+#content > .box:not(#profile-data):not(.slider-carousel-container):not(.sa-stats) .box-head,
+#content > .box:not(#profile-data):not(.slider-carousel-container):not(.sa-stats) .box-content {
```
TheColaber, line 111: I think we are trying to avoid making our own classes (if possible) and use scratch's class to avoid having different selectors and also to work for dark mode.

mxmou: There is no Scratch's class that would make it possible to exclude the box from this selector without changing it (`slider-carousel-container` changes padding).

RedGuy12:
The graph looks a little big...any way to shorten it?

addons/scratchr2/profile.css
TheColaber: Yeah, I understand, just checking.

mxmou:
> The graph looks a little big...any way to shorten it?

Yes, I could make it smaller.

addons/scratchstats/userscript.js
```diff
@@ -0,0 +1,115 @@
...
+  const item = document.createElement("div");
+  const numberDiv = document.createElement("div");
+  item.appendChild(numberDiv);
+  numberDiv.className = "sa-stats-number";
+  numberDiv.innerText = number;
+  const labelDiv = document.createElement("div");
+  item.appendChild(labelDiv);
+  labelDiv.innerText = label;
+  return item;
+}
```
RedGuy12, line 11: i think this is supposed to go inside the `export default async function ({ addon, msg, console }) {`

jeffalo: it doesn't have to

retronbv: it doesn’t? I always thought it did.

jeffalo: i dont think it needs to be part of the exported module

mxmou: > i dont think it needs to be part of the exported module

It doesn't. Exported functions can always use functions that are not exported.

addons/scratchr2/profile.css
```diff
@@ -98,25 +99,25 @@
 .activity-stream .actor a {
   color: #575e75 !important;
 }
-#content > .box:not(#profile-data):not(.slider-carousel-container):not(#page-404) {
+#content > .box:not(#profile-data):not(.slider-carousel-container):not(.sa-stats):not(#page-404) {
```
WorldLanguages, line 102: Can't we add `.slide-carousel-container` to the `.sa-stats` element to avoid adding this?

libraries/license-info.json
```diff
@@ -6,5 +6,6 @@
   "scratch-blocks": "Apache-2.0",
   "intl-messageformat": "BSD-3-Clause",
   "vue": "MIT",
-  "tinycolor2": "MIT"
+  "tinycolor2": "MIT",
+  "chartjs": "MIT"
```
WorldLanguages, line 10: This should also be added to the MD file inside /libraries

addons-l10n/en/scratchstats.json
```diff
@@ -0,0 +1,13 @@
+{
+  "scratchstats/title": "Statistics",
+  "scratchstats/view-more": "View on ScratchStats",
```
WorldLanguages, line 3: I'd rather use "ScratchStats.com"

addons/scratchstats/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Profile statistics",
+  "description": "Shows statistics from ScratchDB on profile pages.",
```
WorldLanguages, line 3: Similarly to other addons that use ScratchDB (or that fetch any other external resource), we should specify this uses ScratchDB and ScratchStats.com (since the img icon is fetched from it)

addons/scratchr2/profile.css
mxmou: Possibly, but Scratch has some styles specific to slider carousels and I don't want them to affect this element.

addons/scratchstats/addon.json
mxmou: I'd rather include the icon in the addon directory then - the statistics themselves aren't from ScratchStats.com and I don't want the description to be misleading.

addons/scratchr2/profile.css
WorldLanguages: I think reverting each of those styles is OK, I've seen us do that previously. Or if you want to go hardcore mode add `all: initial;` to the element :P

mxmou: Apparently I was wrong, adding the class didn't change anything.

addons/scratchstats/addon.json
WorldLanguages: Sure include the icon

addons/scratchstats/userscript.js
WorldLanguages: @RedGuy12 The function that is exported by default is simply a function that runs when everything is ready, similarly to `int main() {}` in other programming languages. Only the default function gets references to `addon`, `msg` and other objects, but you can always pass those as arguments for other functions as well.

Hans5958 (approved):
LGTM? What's holding it? Next major version?

WorldLanguages:
@Hans5958 I prefer not to merge PRs that would add new strings near to release, to give translators some time.

GrahamSH-LLK:
Okay, so will we merge once you release?

Hans5958:
> @Hans5958 I prefer not to merge PRs that would add new strings near to release, to give translators some time.

Ah okay

AwesomeKalin:
I love the idea!

addons/scratchstats/userscript.js
```diff
@@ -0,0 +1,115 @@
...
+  fetch(`https://scratchdb.lefty.one/v2/user/info/${username}`).then(async function (response) {
+    stats.removeChild(stats.firstChild); // remove loading message
+    const followRow = document.createElement("div");
+    stats.appendChild(followRow);
+    followRow.className = "sa-stats-row";
+    const ranksRow = document.createElement("div");
+    stats.appendChild(ranksRow);
+    ranksRow.className = "sa-stats-row";
+    const data = await response.json();
+    followRow.appendChild(createItem(data.statistics.followers, msg("followers")));
```
WorldLanguages, line 48: Maybe we add a thousand comma separator (maybe in all numbers, but most importantly on followers?)

addons/scratchstats/userscript.js
```diff
@@ -0,0 +1,115 @@
...
+  const statsHeader = document.createElement("div");
+  statsBox.appendChild(statsHeader);
+  statsHeader.className = "box-head";
+  const statsTitle = document.createElement("h4");
+  statsHeader.appendChild(statsTitle);
+  statsTitle.innerText = msg("title");
+  const statsMoreLink = document.createElement("a");
+  statsHeader.appendChild(statsMoreLink);
+  statsMoreLink.innerText = msg("view-more");
+  statsMoreLink.href = "https://scratchstats.com/" + username;
```
WorldLanguages, line 30: target _blank?

addons/scratchstats/userscript.js
WorldLanguages: Probably worth it for ranking numbers as well, as some users get 6 digit numbers there

GrahamSH-LLK: We could alternatively go with `6K`, `7M`, etc

addons/scratchstats/userscript.js
mxmou: I don't like `target="_blank"` because I'm often confused by it when I encounter it on other sites. I expect to be able to use the back button to get back after opening a link, unless I middle clicked it. See [this article](https://css-tricks.com/use-target_blank/) for a more detailed explanation.

I don't want to add it unless there is a reason other than it being an external link (e.g. similar links on Scratch itself, but I don't think those exist).

PR #1631, opened 2021-02-19
Discuss button addon: Do not touch the conference pages

### LINKED ISSUE #1630: discuss button addon should not touch the scratch confrence page
https://scratch.mit.edu/conference/2020/
image

TheColaber:
Resolves #1630

### DIFF
--- a/addons/discuss-button/userscript.js
+++ b/addons/discuss-button/userscript.js
@@ -1,5 +1,6 @@
 export default async function ({ addon, global, console }) {
   if (!addon.tab.clientVersion) return;
+  if (document.querySelector(".mod-conference")) return;
 
   const link = document.createElement("li");
   link.className = "link discuss";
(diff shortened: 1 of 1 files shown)


addons/discuss-button/userscript.js
```diff
@@ -1,5 +1,6 @@
 export default async function ({ addon, global, console }) {
   if (!addon.tab.clientVersion) return;
+  if (document.querySelector(".mod-conference")) return;
```
WorldLanguages, line 3: Please look at the URL instead, this element is rendered by React and might not be ready by the time this addon runs

WorldLanguages (changes requested):

addons/discuss-button/userscript.js
Hans5958: That'll do?

TheColaber: Kinda feel like this isn't the optimal way however, it's fine for now until we find another navbar we shouldn't be appending to.

WorldLanguages (approved):

PR #1657, opened 2021-02-24
Change studio that cloud games come from

AwesomeKalin:
This is so that the games aren't just griffpatch games. P.S. WeRule123 is me.

**Resolves**

This resolves the issue that the cloud games were only from griffpatch
Resolves #

**Changes**

I changed the studio that the cloud games come from. The new studio is https://scratch.mit.edu/studios/28943839/
At the minute it's only 4 new games, although I want people to suggest more games that should be added
**Reason for changes**

So that the games aren't just griffpatch games
**Tests**
I have not tested this pull request, although testing shouldn't be necessary for this small change.

### DIFF
--- a/popups/cloud-games/popup.js
+++ b/popups/cloud-games/popup.js
@@ -48,7 +48,7 @@ import WebsiteLocalizationProvider from "../../libraries/website-l10n.js";
     },
     async created() {
       document.title = l10n.get("cloud-games/popup-title");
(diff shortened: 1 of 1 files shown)


TheColaber:
Changing this variable will need discussion. I think a better PR would be one that allows you to set to any studio... not just set studio.

AwesomeKalin:
This was supposed to be a temporary change until we change this, but OK!

PR #1667, opened 2021-02-25
cloud-games: add loading indicator, do not request data in parallel

WorldLanguages:
1. Adds loading indicator
2. Shows incomplete data after more than 50% of projects loaded, but keeps the loading indicator until 100% loads
3. Instead of starting each request in parallel, request every 125ms

### DIFF
--- a/popups/cloud-games/popup.html
+++ b/popups/cloud-games/popup.html
@@ -10,21 +10,28 @@
     <script src="./popup.js" type="module"></script>
   </head>
 
-  <body style="display: none" v-show="loaded">
-    <div class="games" v-for="project of projectsSorted">
-      <div class="title">
-        <a class="project-name" :href="'https://scratch.mit.edu/projects/' + project.id" target="_blank"
-          >{{ project.title }}</a
-        >
-        <span class="float-right"><img class="small-icon" src="../../images/icons/users.svg" /> {{ project.amt }}</span>
-      </div>
-      <div class="project-details" v-show="true || project.amt > 0 && project.extended">
-        <div class="username-list">
-          <a v-for="user of project.users" :href="'https://scratch.mit.edu/users/' + user" target="_blank"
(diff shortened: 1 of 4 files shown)


RedGuy12:
why not in parallel?

popups/cloud-games/popup.html
```diff
@@ -10,21 +10,31 @@
...
-        <a class="project-name" :href="'https://scratch.mit.edu/projects/' + project.id" target="_blank"
-          >{{ project.title }}</a
-        >
-        <span class="float-right"><img class="small-icon" src="../../images/icons/users.svg" /> {{ project.amt }}</span>
-      </div>
-      <div class="project-details" v-show="true || project.amt > 0 && project.extended">
-        <div class="username-list">
-          <a v-for="user of project.users" :href="'https://scratch.mit.edu/users/' + user" target="_blank"
-            >{{ user }}</a
+  <body style="display: none" v-show="true">
```
TheColaber, line 13: not sure why you need those 2 atterabutes

popups/cloud-games/popup.html
```diff
@@ -10,21 +10,31 @@
...
+    <div
+      v-show="projects.length === 0 || projects.length !== projectsChecked"
+      style="text-align: center; font-size: 0.9rem; padding-top: 1rem; font-weight: bold"
```
TheColaber, line 16: could we have this style in the CSS file plz

TheColaber (changes requested):

popups/cloud-games/popup.html
WorldLanguages: hide the body until vue loads

WorldLanguages:
> why not in parallel?

not a scientific explanation, but feels wrong to start more than 10 requests at the same time. scratch often suggests doing less than 10 req/s

PR #1674, opened 2021-02-26
New addon - Longer WIWO

### LINKED ISSUE #1521: 55 more chars in WIWO
**Describe the new addon**

Employs W_L's long wiwo hack. 

**Link to the new addon**

http://longwiwo.blogspot.com/

RedGuy12:
added new addon to resolve #1521


addons/longer-wiwo/userscript.js
```diff
@@ -0,0 +1,23 @@
+export default async function ({ msg }) {
+  const status = $('textarea[name="status"]');
+  const left = $($("#status-chars-left")[0].parentElement);
+
+  // remove scratch's event listeners
+  status.off("input");
+  status.off("focusin");
+  status.off("focusout");
```
WorldLanguages, line 8: This seems to work, but if you're removing the event listeners, why does it still upload the updated WIWO when its changed? What am I missing? I want to know what's going on lol

addons/longer-wiwo/userscript.js
```diff
@@ -0,0 +1,23 @@
+export default async function ({ msg }) {
+  const status = $('textarea[name="status"]');
+  const left = $($("#status-chars-left")[0].parentElement);
```
WorldLanguages, line 3: `$("#status-chars-left").parent()` ?

addons/longer-wiwo/userscript.js
WorldLanguages: Also, I noticed that when the WIWO does update, Scratch adds a green modal to show it saved, but for some reason that modal is empty (has no text). Any ideas why?

RedGuy12: 1: that event listener is on a different element, `div#status`
2: idk, i saw that before i even started working on this

addons/longer-wiwo/userscript.js
RedGuy12: ok

addons/longer-wiwo/userscript.js
mxmou: > Also, I noticed that when the WIWO does update, Scratch adds a green modal to show it saved, but for some reason that modal is empty (has no text). Any ideas why?

This is a Scratch bug (also happens without Scratch Addons). It's supposed to say "Changes saved".

addons/longer-wiwo/userscript.js
```diff
@@ -0,0 +1,23 @@
+export default async function ({ msg }) {
+  const status = $('textarea[name="status"]');
+  const left = $($("#status-chars-left").parent()[0]);
```
WorldLanguages, line 3: Huh, asking again, why not simply `const left = $("#status-chars-left").parent();`? Not sure why you're getting the real DOM element object to then convert it to a jQuery object again, seems unnecessary?

RedGuy12: *facepalm* good point, didn't see that

WorldLanguages: lol

WorldLanguages (approved):

PR #1691, opened 2021-02-28
Detect new elements with placeholder addon.tab.scratchClass classes

### LINKED ISSUE #1690: `addon.tab.scratchClass` should detect newly appended elements with "scratchAddonsScratchClass/" classes
Currently these elements are only checked once after the classes are loaded. If an addon called `addon.tab.scratchClass` when the classes weren't ready, and didn't append the element immediately, this placeholder class will remain.
We can fix this by adding a mutation observer. This may hurt performance slightly but we have many mutation observers already, one per each addon running `addon.tab.waitForElement`...

WorldLanguages:
Resolves #1690

### DIFF
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -144,19 +144,28 @@ function loadClasses() {
     ),
   ];
   scratchAddons.classNames.loaded = true;
-  document.querySelectorAll("[class*='scratchAddonsScratchClass/']").forEach((el) => {
-    [...el.classList]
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Note for @GarboMuffin: you do **not** need to change your TurboWarp implementation for `addon.tab.scratchClass` because of this PR. As long as you don't run addons before the stylesheets are available in the head, you can ignore Scratch Addons code handling these placeholder classes that start with `scratchAddonsScratchClass/`.

PR #1692, opened 2021-02-28
Clarify on Animated Thumbnail notice

RedGuy12:
Clarifies the notice on `animated-thumbnail` to explain *why* it'd be marked as NFE.

### DIFF
--- a/addons/animated-thumb/addon.json
+++ b/addons/animated-thumb/addon.json
@@ -9,7 +9,7 @@
     },
     {
       "type": "warning",
-      "text": "Your project may be marked as NFE (not for everyone) by the Scratch Team if the thumbnail is animated. NFE projects aren't able to get into the front page, the explore page, or search results.",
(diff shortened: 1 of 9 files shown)


WorldLanguages:
Create a new branch for each PR otherwise you'll be including the same commits again and cause conflicts if they were merged already

RedGuy12:
oops sorry thought i did

addons/animated-thumb/addon.json
```diff
@@ -9,7 +9,7 @@
     },
     {
       "type": "warning",
-      "text": "Your project may be marked as NFE (not for everyone) by the Scratch Team if the thumbnail is animated. NFE projects aren't able to get into the front page, the explore page, or search results.",
+      "text": "Your project may be marked as NFE (not for everyone) by the Scratch Team if the thumbnail is animated. NFE projects aren't able to get into the front page, the explore page, or search results. This is because animated GIFs can lag the page.",
```
WorldLanguages, line 12: This is correct (Scratch Team uses that justification) but I'm not sure why we need to tell addon users about it? Also, that answer was given by the ST when Internet Explorer was supported, and IE didn't like gifs very much.

PR #1712, opened 2021-03-02
Infinite scrolling: fix bugs, improvements

### LINKED ISSUE #749: Clicking on comment link does not show correct comment (Infinite Scrolling)
**Describe the bug**

Clicking on a message link with InfiniteScrolling/ProjectComments enabled does not take you to the correct comment.

**To Reproduce**

1. Go to https://scratch.mit.edu/projects/444598492/#comments-168272601

**Expected behavior**

Should take you to the correct comment (_Oof that's sad..._)

Screenshot 2020-11-14 at 12 25 25

**Screenshots**

This is the option in question (deselected in screenshot)

Screenshot 2020-11-14 at 12 22 08

Clicking on the message link takes you to the following screenshot which bears no correlation to the expected destination

Screenshot 2020-11-14 at 12 26 36

- macOS Catalina Version 10.15.7
- Chrome Version 86.0.4240.75 (Official Build) (x86_64)

**Additional context**

WorldLanguages:
- Resolves #749 
- Resolves infinite message scrolling not working if URL is `/messages` instead of `/messages/`
- Cleaner code
- "Fix footer": decrease height, not sure why it had to take so much space
- Add 1px border in top of footer (if fixed)

@mxmou Please review, since I think you originally increased the height of the footer. Also, I'd like to find a good way to make the border I added darker if website dark mode is on.

### DIFF
--- a/addons/infinite-scroll/userscript.css
+++ b/addons/infinite-scroll/userscript.css
@@ -15,7 +15,8 @@ body {
   top: initial;
   bottom: 0;
   width: 100%;
-  height: 46px;
+  border-top: 1px solid #b3b3b3;
+  height: 10px;
   padding: 0;
   z-index: 20;
   transition-duration: 0.5s;
--- a/addons/infinite-scroll/userscript.css
+++ b/addons/infinite-scroll/userscript.css
@@ -16,7 +16,7 @@ body {
   bottom: 0;
   width: 100%;
   border-top: 1px solid #b3b3b3;
-  height: 10px;
+  height: 32px;
   padding: 0;
   z-index: 20;
   transition-duration: 0.5s;
--- a/addons/infinite-scroll/buttonScroll.js
+++ b/addons/infinite-scroll/buttonScroll.js
@@ -1,28 +1,43 @@
 async function commentLoader(addon, heightControl, selector) {
-  await addon.tab.waitForElement(selector);
-  document.querySelector(selector).style.display = "none";
-  window.addEventListener("scroll", async () => {
-    if (window.scrollY + window.innerHeight >= document.getElementById(heightControl).offsetHeight - 500) {
-      if (document.querySelector(selector)) {
-        document.querySelector(selector).click();
-        await addon.tab.waitForElement(selector);
-        document.querySelector(selector).style.display = "none";
+  let func;
+  while (true) {
+    const el = await addon.tab.waitForElement(selector, { markAsSeen: true });
+    if (func) window.removeEventListener("scroll", func, { passive: true });
+    el.style.display = "none";
+    func = () => {
+      if (window.scrollY + window.innerHeight >= document.querySelector(heightControl).offsetHeight - 500) {
+        if (el) el.click();
       }
-    }
-  });
+    };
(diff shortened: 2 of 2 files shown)


WorldLanguages:
We should also consider removing "beta" from this addon's tag list

Barthdry:
> We should also consider removing "beta" from this addon's tag list

YES

addons/infinite-scroll/userscript.css
```diff
@@ -15,7 +15,8 @@ body {
   top: initial;
   bottom: 0;
   width: 100%;
-  height: 46px;
+  border-top: 1px solid #b3b3b3;
+  height: 10px;
```
mxmou, line 19: 10px is too small to easily access. scratch-www uses a height of 48px for most interactive elements, so please make it at least 32px.

For the border, I'd use `#d9d9d9` and add `#footer` to [this selector](https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/dark-www/experimental_scratchwww.css#L110-L138).

WorldLanguages: I know 10px is small, but it's still fixed at the bottom of the page? All you need to access it is move the cursor to the bottom of the viewport. Doesn't seem fair to compare it against interactive elements that do not have that behavior?
Also, are you OK adding a border to the footer on dark-www, even tho vanilla Scratch does not have a border there? It's only added by infinite scrolling.

mxmou: > I know 10px is small, but it's still fixed at the bottom of the page? All you need to access it is move the cursor to the bottom of the viewport. Doesn't seem fair to compare it against interactive elements that do not have that behavior?

I don't think being at the bottom of the page makes it easier to access with the cursor (except in full screen). It's close to the bottom of the screen but not at the bottom, which isn't very practical.

> Also, are you OK adding a border to the footer on dark-www, even tho vanilla Scratch does not have a border there? It's only added by infinite scrolling.

`border-color` doesn't add a border if it isn't already there, so adding `#footer` to the selector I linked in the previous comment should work.

WorldLanguages: > except in full screen

What do you mean by that? On windows OS at least, if you move your cursor to the taskbar, it also shows the footer. You don't have to put the cursor within those 10 CSS pixels, you can move the cursor further down with no issue

> border-color doesn't add a border if it isn't already there, so adding #footer to the selector I linked in the previous comment should work.

Clever

mxmou: > > except in full screen
> 
> What do you mean by that? On windows OS at least, if you move your cursor to the taskbar, it also shows the footer. You don't have to put the cursor within those 10 CSS pixels, you can move the cursor further down with no issue

Doesn't work for me. Even if it did, it would only be possible if the window was maximized.

WorldLanguages: Good point

WorldLanguages:
@mxmou Please PR adding a `border-color` to the footer on `dark-www`. Make sure to add a comment specifying that property is specifically to be compatible with this addon (for future reference)

PR #1721, opened 2021-03-03
Make sure malformed settings don't cause big problems

apple502j:
Since there are some people who want to load crafted settings, this PR makes sure they don't cause big problems (such as crashes).

### DIFF
--- a/addons/60fps/userscript.js
+++ b/addons/60fps/userscript.js
@@ -21,7 +21,7 @@ export default async function ({ addon, global, console }) {
         e.cancelBubble = true;
         e.preventDefault();
         mode = !mode;
-        if (mode) setFPS(addon.settings.get("framerate"));
+        if (mode) setFPS(Math.min(240, Math.max(31, addon.settings.get("framerate") | 31)));
         else setFPS(30);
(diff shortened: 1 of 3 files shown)


WorldLanguages:
Should we care about this? If users are able to change settings through the console then they likely know what they're doing?

retronbv:
> Should we care about this? If users are able to change settings through the console then they likely know what they're doing?

They could also change it by importing malformed settings json

WorldLanguages:
> They could also change it by importing malformed settings json

Good point, didn't think of that

addons/60fps/userscript.js
```diff
@@ -21,7 +21,7 @@ export default async function ({ addon, global, console }) {
         e.cancelBubble = true;
         e.preventDefault();
         mode = !mode;
-        if (mode) setFPS(addon.settings.get("framerate"));
+        if (mode) setFPS(Math.min(240, Math.max(31, addon.settings.get("framerate") | 31)));
```
GarboMuffin, line 24: Surely that's not supposed to be a bitwise or?

apple502j: whoops

WorldLanguages:
Addons shouldn't validate settings

PR #1727, opened 2021-03-04
v1.11: add update notification and change addons with "new" tag

WorldLanguages:
"New" addons are: folders, variable-manager and scratchstats


WorldLanguages:
Merging at 11 PM UTC+0
If someone wants to suggest changes do it before that

PR #1741, opened 2021-03-05
Simplify disable-auto-save

### LINKED ISSUE #1709: Disable autosave addon disables CTRL+S and File menu saving
**Describe the bug**

When the Disable autosave addon is enabled, the Scratch editor doesn't save at all. The saving animation plays briefly, but the project doesn't actually upload to Scratch's servers.

**To Reproduce**

1. Enable the "Disable autosave" addon
2. Go to any of your projects
3. Try to save it
4. Once the animation finishes, reload the page
5. Your change was probably not saved

**Expected behavior**

The addon should still allow manual attempts to save.

- OS: Windows
- Browser: Firefox 86
- Version: v1.10.0

**Additional context**

I think the addon used to work, but I can't remember for sure.

apple502j:
**Resolves**
Resolves #1709 

**Changes**
Simplify disable-auto-save to use redux. Too simple that I hate to test this.

NOTE: this does remove disabling "save when going to project page", but the cost of accidental saving is significantly smaller than the cost of losing projects. Note that this behavior is not "autosave" per se; maybe we can rename the button to "Save and See Project Page"?

### DIFF
--- a/addons/disable-auto-save/userscript.js
+++ b/addons/disable-auto-save/userscript.js
@@ -1,51 +1,8 @@
 export default async ({ addon, console }) => {
-  var manualPress = false;
-  // shamelessly stolen from animated-thumb/persistent-thumb.js
-  const xhrOpen = XMLHttpRequest.prototype.open;
-  XMLHttpRequest.prototype.open = function (method, path, ...args) {
-    if (method === "put" && String(path).startsWith("https://projects.scratch.mit.edu/")) {
-      if (!manualPress) method = "OPTIONS";
-    }
-    return xhrOpen.call(this, method, path, ...args);
-  };
-  //initial adding of the observers and refreshing
-  addMutationListeners();
-  addon.tab.addEventListener("urlChange", function (event) {
-    addMutationListeners();
+  addon.tab.redux.initialize();
+  addon.tab.redux.addEventListener("statechanged", ({ detail }) => {
+    if (detail.action.type !== "timeout/SET_AUTOSAVE_TIMEOUT_ID") return;
+    clearTimeout(detail.next.scratchGui.timeout.autoSaveTimeoutId);
+    console.log("Pending autosave prevented.");
   });
-
-  async function addMutationListeners() {
-    //add a function to the "Save Now" button to disable request interception if it was a manual save
-    var saveContainer = await addon.tab.waitForElement("[class^='menu-bar_account-info-group']");
-    saveContainer = saveContainer.childNodes[0];
-    const saveObserver = new MutationObserver(function (mutations) {
-      if (saveContainer.childNodes[0]) {
-        if (saveContainer.childNodes[0].className.startsWith("save-status_save-now")) {
-          manualPress = false;
-          saveContainer.childNodes[0].addEventListener("click", function () {
-            manualPress = true;
-          });
-        }
-      }
-    });
-    saveObserver.observe(saveContainer, { childList: true });
-    var alertContainer = await addon.tab.waitForElement("[class^='alerts_alerts-inner-container']");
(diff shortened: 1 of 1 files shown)


TheColaber:
I also did this code before, and World_Languages said it wasn't fully correct.

WorldLanguages:
> I also did this code before, and World_Languages said it wasn't fully correct.

Probably because it did save after going to the project page.
Maybe we should add a `prompt()` when clicking "save to project page" asking the user if they really want to, because the project would get saved?

WorldLanguages:
This changes behavior of an already existing addon. Could we try to resolve #1709 by instead checking if Ctrl+S or File>save are triggered, in the same way it's done with the "save now" button?

TheColaber:
@WorldLanguages Just so you know, me and some others have had conflicts with the project not saving, even when we do press save, and it says it has saved. This solution is good for now because the addon is a big problem right now.

mxmou:
@TheColaber And potentially saving when not intended is much better than not saving when you click Save.

apple502j:
Doing a coup on this one. It is likely some customers lost their data. Breaking change needs to occur.

WorldLanguages:
> Doing a coup on this one. It is likely some customers lost their data. Breaking change needs to occur.

Uhh, the addon is likely called "disable auto-save", and you're worried about users losing data? They opted in into this. No autosave means no saving unless you explicitly "save". Going into project player mode is not saving.

mxmou:
> Uhh, the addon is likely called "disable auto-save", and you're worried about users losing data? They opted in into this.

They probably expect it to save when they click "Save" in the file menu.

> Going into project player mode is not saving.

It isn't auto-save either - auto-save is saving without user interaction. It makes sense to save when going to the project page because the project page doesn't have a save button.

WorldLanguages:
@TheColaber told me a good point @apple502j did not mention for some reason: if you go into player mode, and close the tab, there's no call to `preventDefault()` on beforeunload which would trigger a "are you sure you want to close this tab?" modal.
So, options are:
1. Make that modal also work if you're on player mode, and there are pending changes (likely complex to do properly)
2. What this PR did (but some users may not expect it)
3. Add a `prompt()` when "see project page" is clicked, if there are pending changes. Prompt would make sure to tell the user that changes were not saved, and that closing the tab would make them lose the changes. (Likely means more complex addon)
4. Force the user to save the project before going into player mode, but add a `prompt()` so that they don't save the project accidentally. Easy to code.

I'd prefer not changing addon behavior, or going all in with option 4.

Hans5958:
I prefer to have this addon is fully disable auto-save. It's a scam to have an addon that can disable auto-save, but not fully. I suggest number 4, and maybe 3 down the line.

mxmou:
@Hans5958 I don't think saving when exiting the editor can be considered auto-save.

Hans5958:
> I don't think saving when exiting the editor can be considered auto-save.

I think it is an autosave. When I'm going out into the project view, I didn't explictly say to save the project. I just want to go into the project view. But it saves anyway, so it is an autosave.

If there is a techincal limitation that makes changes gone when going into the project view. I expect a window of "You have unsaved changes. Do you want to save?" not an autosave.

apple502j:
Anyway, this PR is merged, and you probably missed #1850.

Hans5958:
> 
> 
> Anyway, this PR is merged, and you probably missed #1850.

1. Personally, I don't like something is ended with the "oh no, anyway" theme. I think it is inappropriate because a discussion may not end after it's completion (especially if it is done one-sided coup, but that's not related with that)
2. I was going to edit my message but got carried away by @mxmou's discussion.

PR #1754, opened 2021-03-08
addon.tab.createBlockContextMenu API

### LINKED ISSUE #1749: addon.tab.editorContextMenu API
Some addons that use ( and plan to use ) the editor context menu should be able to do it the same way as each. The API can also choose the order. The method shouldn't use DOM like editor-devtools, but as the way working as [block-switching](https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/block-switching/userscript.js#L710). This should also hae more options like, "only on variables" or "on all blocks".

### LINKED ISSUE #1968: Using find functionality can leave found occurrences highlighted
**Describe the bug**

Search for something in a Scratch file using ctrl-f. Tap the next button to move to the next occurrence while the first occurrence is flashing. It will remain lit up if that's the state it was in when moving to the next one.

**To Reproduce**

Quickly move through occurrences of string using find functionality.

**Expected behavior**

No strings remain incorrectly coloured

**Screenshots**

image

- Windows/Mac
- Chrome

**Additional context**

You can revert to normal colour by changing to a different sprite and then back again.

TheColaber:
Resolves #1749
Resolves #1968


TheColaber:
Lemme explain the function:
```js
createBlockContextMenu(
    callback,
    { workspace = false, allblocks = false, flyout = false, catagories = [], beforeAppend } = {}
  )
```
The first param is the callback... the callback gets the param with all the items currently in the list. The function should return the array of items back. Then the second param is the options params. `workspace` decides if it should run the callback when the user right-clicks on the workspace. `allblocks` will run the callback when the user right-clicks on any block. `flyout` if true will also run the callback when the user right-clicks on a block in the flyout as well as the extra params (example below). `categories` decides if it should run the callback when the user right-clicks on the listed categories. And finally, `beforeAppend` is a function that runs before it runs your callback. If gets 2 params, the first being the `event` and the other being a blocky block. You must return true in this function in order to run your callback function.

Here is an example:
```js
  addon.tab.createBlockContextMenu((items) => {
    items.push({
      callback: (event) => {
        console.log("Clicked!!", event);
      },
      enabled: true,
      text: "Swap Variables in sprite"
    });
    return items;
  }, {
    workspace: false, // Not on the workspace
    allblocks: false, // Not on all blocks
    catagories: ["data"], // on the variables catagory
    flyout: true, // Also on the flyout
    beforeAppend: (event, block) => {
      return block.type == "data_variable"; // Only allow callback on the actually reporter variables
    }
  });
```

mxmou:
> // TODO: Create Separators somehow... help!

Separators are not a Scratch feature, you need to add them using CSS after the menu loads.

WorldLanguages:
@TheColaber If this resolves #1848 please add it as a linked issue.

addons/editor-devtools/blockly/BlockFlasher.js
```diff
@@ -9,20 +9,19 @@ export default class BlockFlasher {
   static flash(block) {
     if (myFlash.timerID > 0) {
       clearTimeout(myFlash.timerID);
-      myFlash.block.setColour(myFlash.colour);
+      myFlash.block.svgPath_.style.fill = "";
```
WorldLanguages, line 12: Please elaborate the bug that this change (and others below it) fixes, related to theme3. Make an issue number for it and link it to this PR

TheColaber: Problem is, devtools would set the blocks to their default color, and not it's custom. By setting its fill to "", it goes back to its custom color.

apple502j:
Can you make #1848 fix in separate PR to put it in 1.11.1?

content-scripts/inject/module.js
```diff
@@ -160,17 +161,13 @@ async function onDataReady() {
...
-    // To be done when userscripts can be dynamicly ran
-    // Will invlove a promice most likely...
-  }
+  const exampleAddon = new Addon({});
+  const exampleTab = exampleAddon.tab;
+  if (exampleAddon.editorMode === "editor") inject(await exampleTab.traps.getBlockly());
+  exampleTab.addEventListener(
+    "urlChange",
+    async () => exampleTab.editorMode === "editor" && inject(await exampleTab.traps.getBlockly())
+  );
```
WorldLanguages, line 170: You're in module.js, you can do better than this

addons/block-switching/userscript.js
```diff
@@ -669,111 +669,41 @@ export default async function ({ addon, global, console, msg }) {
...
-            continue;
-          }
-          const children = node.children;
-          const item = children[addBorderToContextMenuItem];
-          if (item) {
-            item.style.paddingTop = "2px";
-            item.style.borderTop = "1px solid hsla(0, 0%, 0%, 0.15)";
-          }
-          addBorderToContextMenuItem = -1;
+  const Blockly = await addon.tab.traps.getBlockly();
```
pufferfish101007, line 672: Wouldn't it make more sense to assign Blockly to the `await addon.tab.traps.getBlockly()` on line 4?

TheColaber: huh, yeah, didnt see that

PR #1770, opened 2021-03-09
solve #1763

jeffalo:
adds a filter to make sure broadcasts don't end up the variable tab


apple502j:
You can omit that for local variables (since all broadcasts are global), but I'm gonna merge.

jeffalo:
> You can omit that for local variables (since all broadcasts are global), but I'm gonna merge.

this will also make sure any other things scratch stores as variables will not be counted as a var in var-manager

PR #1799, opened 2021-03-10
Remove Vid Sensing from load-extensions addon

TheColaber:


apple502j:
This makes it enable then disable camera - but it would still e.g. ask for permissions when the user doesn't expect it to. Rejected.

TheColaber:
Updated. We decided to just remove the setting because of how much confusion it was causing and had an un-optimal fix.

apple502j:
The fix is "optimal", although merging.

PR #1606, opened 2021-02-17
Darkmode fixes

TheColaber:
Resolves #1599

EDIT: as of Feb 17 only fixes this:
image

PR #1805, opened 2021-03-10
Convert editor dark mode to presets

### LINKED ISSUE #657: "Editor dark mode": convert modes to presets, allow for custom colors

mxmou:
**Resolves**

Resolves #657

**Changes**

Add color settings to editor dark mode and replaces the different versions with presets.

Experimental Dark
Example of customized colors

There are a lot of different settings and I'd like to split some (e.g. separate monitor background from tab background and asset editor background from accent background) but there are already too many so I tried to minimize the number of options.

The addon changes the Scratch UI a bit, even with default settings, to avoid adding too many customizable colors. 3.Dark and 3.Darker used to make cancel buttons red, but I removed that because I don't think dark themes should make changes like that.

To do:
* [x] Add option to make scrollbars (except the workspace scrollbar) dark
* [x] Automatically select the appropriate preset depending on the current settings for <del>people who already have the addon enabled</del> everyone. This should not affect the text shadow setting.
* [x] Make the 3.Darker preset selected by default
* [x] Better settings UI: change "Reset" to a dropdown with "Default" and the different presets as options
* <del>Maybe use filters to recolor icons when the primary color is changed</del> I decided not to do this because I made the highlight color dark in Dark Editor and this would make icons hard to see with a dark primary color.
* Maybe move the text color functionality to the addon loader so users won't have to wait for an userscript to run to see the correct color. See #2099.


mxmou:
Note that I moved the dark loading screen from dark-www because it's a part of the project player, see https://github.com/ScratchAddons/ScratchAddons/pull/1606#issuecomment-780780303.

I renamed Experimental Dark to make it clear that it isn't experimental.

WorldLanguages:
> Note that I moved the dark loading screen from dark-www because it's a part of the project player, see [#1606 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/1606#issuecomment-780780303).
> 
> I renamed Experimental Dark to make it clear that it isn't experimental.

Can you rephrase? I forgot what theme makes loading screens dark at the moment.

WorldLanguages:
Merging this will probably take some time.
Some comments:
1. We need to make sure all presets look pretty much the same than they did when they were not presets (except making cancel buttons red, etc.)
2. This does not affect "join scratch" modal, epic!
3. I'd prefer having less colors. Honestly I don't have the time and patience to check what every color does, not sure if anyone else will. I guess some UI colors could be based on other ones automatically (like the color of the scrollbar). Not sure
4. It won't be very easy to get good UI/UX out of this, at the cost of personalization. Specially because now "default" can mean default Scratch colors, or the default dark mode (3.Darker)

WorldLanguages:
Can you remind me why 3.Darker couldn't be the default, and vanilla Scratch a preset? I'm thinking we maybe should make 3.Darker colors default, as well as making 3.Darker a preset (yes, duplicated), with vanilla Scratch also as a preset. Adding a way to reset a single color to any preset would allow this, I think?

mxmou:
> > Note that I moved the dark loading screen from dark-www because it's a part of the project player, see [#1606 (comment)]
> 
> Can you rephrase? I forgot what theme makes loading screens dark at the moment.

Currently loading screens are affected by website dark mode, but this doesn't make sense because otherwise the editor dark mode controls the project player.

> 1. We need to make sure all presets look pretty much the same than they did when they were not presets (except making cancel buttons red, etc.)

I'll try to make a list of all changes this makes to the existing dark mode versions (most of them are minor).

> 3. I'd prefer having less colors. Honestly I don't have the time and patience to check what every color does, not sure if anyone else will. I guess some UI colors could be based on other ones automatically (like the color of the scrollbar). Not sure

I think it's already obvious what most colors do, except primary, secondary and accent which change lots of things. Reducing the number of colors without changing the appearance of presets too much is hard, but I think the following changes could be made:
- The scrollbar could be a transparent overlay, I'd just have to make sure it still looks the same with the default Scratch colors and with Experimental Dark (where it needs to be the same color as borders;. 3.Darker and 3.Dark don't currently make the scrollbar dark.)
- The selected category background could be an overlay too: it already is in 3.Dark(er) and this wouldn't change the other dark modes too much.
- Instead of separate primary and secondary color settings, there could be just one named "highlight color" and the secondary color would be a darker variant of it.

> Can you remind me why 3.Darker couldn't be the default, and vanilla Scratch a preset? I'm thinking we maybe should make 3.Darker colors default, as well as making 3.Darker a preset (yes, duplicated), with vanilla Scratch also as a preset. Adding a way to reset a single color to any preset would allow this, I think?

It's more intuitive if clicking the reset icon, or choosing "default" in a reset menu, resets to a Scratch color and not a dark mode, but the way you suggested where the default settings are duplicated as a preset would work too, the only problem(?) would be that 3.Darker would be called "Default" in the reset menu.

WorldLanguages:
> The only problem(?) would be that 3.Darker would be called "Default" in the reset menu.

The addon will still be called "Editor dark mode [...]" so I don't think that's a problem, tbh.
We can work around this and make it so it says "3.Darker" instead of "Default", somehow

mxmou:
> > The only problem(?) would be that 3.Darker would be called "Default" in the reset menu.
> 
> The addon will still be called "Editor dark mode [...]" so I don't think that's a problem, tbh.
> We can work around this and make it so it says "3.Darker" instead of "Default", somehow

It depends on how the dropdown menu would work. In cases like this, a way to tell that default settings have a name in `addon.json` (maybe a property called `defaultPresetName`) so they would appear with that name in the reset menu would be useful.

mxmou:
> > 1. We need to make sure all presets look pretty much the same than they did when they were not presets (except making cancel buttons red, etc.)
> 
> I'll try to make a list of all changes this makes to the existing dark mode versions (most of them are minor).
> 

All presets:
* `scratch-blocks` inputs and zoom icons are now dark.
* Improved block context menu appearance: it now uses the same colors as the `scratch-gui` context menus and separators added by some other addons are now clearly visible.
* Input placeholders are now transparent instead of using the browser default (which is typically gray).
* Text on unselected tabs is now transparent white (it used to be gray in Dark Editor and white in other dark modes).

3.Darker:
* I changed the border color from transparent black to transparent white to make borders of some elements, such as backpack items, more visible.

Both 3.Darker and 3.Dark:
* Stage monitors are now dark.
* `scratch-blocks` scrollbars are now dark.
* Block comments are now dark.
* Icons on dark backgrounds are now white.
* Previously the background of a selector item used two different colors when not hovered, one behind the icon and the other behind the name of the item. Because Scratch doesn't do this, it would require an additional setting, so I removed it.
* The body of the stage selector now has the same background color as its header, for the same reason.
* The backpack header has a different color too for this reason.
* Borders in the paint editor are now transparent.
* Slightly different modal background to avoid adding a separate setting for it.
* Modal headers use the primary color as their background instead of the same color as the modal itself. I think I'm going to change that to the menu bar color.
* Buttons in the block palette now change background on hover.
* Cancel buttons are no longer red.

Dark Editor:
* All elements now have the same border color, `#111111`.
* Tabs now change color on hover.
* I removed the font weight change from tabs. It affects layout, which dark mode shouldn't. Instead I made the text color more visible (see "All presets").
* The background of the paint and sound editors is now the same as that of the sprite selector header, again to avoid adding too many settings.
* Removed the changed background color and font weight of paint editor buttons: another thing dark mode shouldn't do.
* Modals are now dark. As I mentioned already, I'm going to change their header to use the menu bar background which will otherwise make them almost the same as before.
* Dark Editor used to change the color of some blue buttons to gray to make using dark mode more comfortable - see LLK/scratch-gui#59 for interesting discussion about this. I removed this for now, but I want to add an option called something like "unobtrusive highlights" that would reduce the overall use of the primary color, probably replacing it with the either a transparent/darker version of it or the input background.

Experimental Dark:
* No changes other than bug fixes and the ones mentioned under "All presets". Experimental Dark is essentially just a dark version of Scratch's color scheme so converting it to a preset was easy.

mxmou:
@WorldLanguages I removed a few settings and changed the Dark Editor highlight color to a darker blue.

mxmou:
Waiting for #1832 for the reset dropdown.

addons/editor-dark-mode/text_color.js
```diff
@@ -0,0 +1,110 @@
...
+  textColor("input-transparentText", settings.get("input"), "rgba(87, 94, 117, 0.6)", "rgba(255, 255, 255, 0.4)");
+  textColor("input-filter", settings.get("input"), "none", "brightness(0) invert(1)");
+  textColor("input-codeZoomFilter", settings.get("input"), "none", "invert(1) hue-rotate(180deg)");
+  textColor(
+    "categoryMenu-selection",
+    settings.get("categoryMenu"),
+    "rgba(87, 124, 155, 0.13)",
+    "rgba(255, 255, 255, 0.05)"
+  );
+  transparentVariant("primary-transparent35", settings.get("primary"), "0.35");
```
towerofnix, line 79: I'm not sure this code (`transparentVariant` being passed `settings.get`) is working? I noticed this when I looked at the colors around a selected sprite:

a selected sprite

The proper contrast around the (thin) border of the element and the (thick) box shadow isn't there, even though the colors are set appropriately (`--editorDarkMode-primary` and `--editorDarkMode-primary-transparent35`).

The RGB value of `primary` is, as expected, `#47566b`. But the RGB of `primary-transparent35` is `#4d97ff`. (This is a much brighter color - gauging by the preview, possibly the default 3.0 primary? - so when it's reduced to 35% alpha, it looks pretty much the same as the correctly overridden primary color against a dark background.)

(PS, I'm testing with the Dark Editor theme. I'm the author of it, and I love seeing how it's been worked into this PR! It's a lot more refined than my original version :) )

RedGuy12: I think it needs to be `addon.settings.get`

addons/editor-dark-mode/experimental_editor.css
```diff
@@ -1,215 +1,548 @@
...
 }
 .stage-header_stage-button-icon_3zzFK,
 .sprite-info_icon-wrapper_3Wbqq:not(.sprite-info_radio_v-fgn),
+.sprite-info_radio_v-fgn:not(.sprite-info_is-active_ewF6e) img,
 img.tool-select-base_tool-select-icon_tJ-rr,
 .paint-editor_button-group-button-icon_10kVn,
 img[src="/static/assets/7bd7487b704797cb5ab3cb441486ea70.svg"],
 .library-item_featured-extension-metadata-detail_1M9BV img {
-  filter: brightness(0) invert(1);
+  filter: var(--editorDarkMode-accent-filter);
```
towerofnix, line 230: This mostly works, but there should be a different rule for toggle switches. For example, there isn't appropriate contrast here:

"show" toggle with the "shown" option selected

The option which isn't selected should be dimmer. (Probably a different filter should be applied.)

Rules that this applies to:
- `sprite-info_radio_v-fgn`
- `stage-header_stage-button-icon`
- `tool-select-base_tool-select-icon` (unselected paint tools should be dim)
- Possibly ` .paint-editor_button-group-button-icon`? I'm not sure where this shows up.

addons/editor-dark-mode/text_color.js
towerofnix: Another place this shows up is in the tag menu bar (also `primary-transparent35`):

Two stacked headers in the costume modal. Top one is primary, bottom is broken transparent color (slightly saturated blue).

And here's an example of the "action menu" colors not being overridden (`primary-variant` & `primary-transparent35`):

"Choose a Sprite" action menu. The slide-out portion is a medium-bright blue. The ring around the main button is the usual incorrect transparent color.

mxmou: The shadow of the main action menu button not being overridden is a bug, I'll fix it soon. Other instances of primary color variations work for me - does it work correctly when you reload?

> (PS, I'm testing with the Dark Editor theme. I'm the author of it, and I love seeing how it's been worked into this PR! It's a lot more refined than my original version :) )

I think the most noticeable changes are the use of `#111111` as the border color for all elements to avoid adding another setting and the darker primary color which I changed based on your comment [here](https://github.com/LLK/scratch-gui/issues/59#:~:text=I'd%20love%20to%20hear%20more%20as%20to%20why%20you%20made%20%28the%20add,similar%20to%20what%20I%20did%20for%20the%20menu%20bar.).

addons/editor-dark-mode/text_color.js
```diff
@@ -0,0 +1,110 @@
...
+    "input",
+    "categoryMenu",
+    "palette",
+  ]) {
+    textColor(`${setting}-text`, settings.get(setting));
+  }
+  textColor("page-tabHoverFilter", settings.get("page"), "grayscale(100%)", "brightness(0) invert(1)");
+  textColor("primary-filter", settings.get("primary"), "brightness(0.4)", "none");
+  textColor("primary-filter2", settings.get("primary"), "none", "brightness(0) invert(1)");
+  textColor("secondary-filter", settings.get("secondary"), "brightness(0.4)", "none");
```
towerofnix, line 61: This line isn't used for anything, and throws an error because the "secondary" setting doesn't exist. It should be removed - this fixes https://github.com/ScratchAddons/ScratchAddons/pull/1805#discussion_r593939501!

addons/editor-dark-mode/text_color.js
towerofnix: I found the code which was broken on my end :) https://github.com/ScratchAddons/ScratchAddons/pull/1805#discussion_r593962470

I think it was still working for you because you had the "secondary" setting internally stored, but from a fresh install (i.e. mine) that setting had been removed and so wasn't present, hence the error.

mxmou: Fixed.

addons/editor-dark-mode/experimental_editor.css
mxmou: I think adding `opacity(0.7)` to the filter for unselected toggle buttons would work well.

Scratch doesn't make unselected paint tools dim and I don't think dark mode should change that. `paint-editor_button-group-button-icon` are zoom button icons.

mxmou: Fixed.

mxmou:
This is now fully compatible with the variable manager and editor messages.

WorldLanguages:
Let's aim for v1.13.0

mxmou:
The border color setting now uses a color picker with transparency.

> Let's aim for v1.13.0

Is #1832 planned to be merged by then?

mxmou:
I made the reset button on the settings page a dropdown.
image
*(For this particular addon "Default" should probably be called "3.Darker (default)" in the menu and "Scratch's default colors" should be singular, but this would require additional `addon.json` properties so I think I will do that in a separate PR.)*

Only presets that change the color from its default value are shown:
image

If there are no such presets, it's just a button, like before:
image

WorldLanguages:
@mxmou Looks good, will take a deeper look later. 
I changed the date for the v1.13.0 milestone to April 16th in order to have more time, since v1.13.0 will include 2 refactors (CSS and removal of "theme tag") as well as this. I think we may also do partial rollout in Chrome to make sure we don't introduce any important bugs accidentally to all users.
After merging this I'd like to focus on #2031 which will also touch the settings page (doesn't matter if we resolve that issue within v1.13.0 or v1.14.0).

mxmou:
This is now finished, the only thing that is still not done is loading the appropriate preset automatically for existing users.

WorldLanguages:
> This is now finished, the only thing that is still not done is loading the appropriate preset automatically for existing users.

I'll handle that, will push into this PR soon ™️

TheColaber:
Not sure if this was mentioned, but I would make the "INPUT BACKGROUND" setting have the block inputs
image
be different from the dom inputs 
image

TheColaber:
<details>
<summary>Btw, this is the turbowarp settings, which I've had requests for...</summary>
<pre>
"accent": "#111111ff",
"activeTab": "#1e1e1eff",
"border": "#00000026",
"categoryMenu": "#111111ff",
"darkComments": false,
"darkScrollbars": false,
"dots": true,
"fullscreen": "#111111ff",
"input": "#fcfcfcff",
"menuBar": "#333333ff",
"page": "#111111ff",
"palette": "#111111cc",
"primary": "#ff4d4dff",
"selectedMode": "3-darker",
"selector": "#1e1e1eff",
"selector2": "#2e2e2eff",
"selectorSelection": "#111111ff",
"stageHeader": "#111111ff",
"tab": "#2e2e2eff",
"textShadow": false,
"workspace": "#1e1e1eff"
</pre>
</details>
Not sure if you want to include it in this PR but...

mxmou:
> Not sure if this was mentioned, but I would make the "INPUT BACKGROUND" setting have the block inputs be different from the dom inputs

I don't want to add more settings: that would make this addon harder to use. If you're trying to convert an existing theme to a preset set input background to the background of DOM inputs.

Secret-chest:
@mxmou It's only one more setting, what is so bad about that?

mxmou:
> @mxmou It's only one more setting, what is so bad about that?

True, but I don't know why anyone would need different colors for block and normal inputs.

towerofnix:
> True, but I don't know why anyone would need different colors for block and normal inputs.

I :+1:'d it because it'd personally be a nice addition. My own theme only customized the color of dom inputs, as I prefer having all familiar block colors, while keeping a clean, dark look in the rest of the editor.

TheColaber:
> I don't want to add more settings: that would make this addon harder to use. If you're trying to convert an existing theme to a preset set input background to the background of DOM inputs.

I'm not convinced I like 
image
or
image

mxmou:
Maybe a checkbox called "Change color of block inputs"? It could be disabled in 3.Dark because blue inputs look weird on most blocks.

towerofnix:
Adding on @TheColaber's illustrative screenshots—Scratch's block colors are suited for and designed around a generally light color scheme. When we make any userstyle, it's valuable to keep in mind the contrasts and appearance originally intended; dark inputs inside of blocks throw that balance off, IMO.

towerofnix:
> Maybe a checkbox called "Change color of block inputs"? It could be disabled in 3.Dark because blue inputs look weird on most blocks.

I'm curious how this is less trouble than just adding a separate color, which users can customize to their own intent?

towerofnix:
PS: Ooh, worth noting carljbowman's own prototype designs for a dark mode involved distinct dom and block input colors: https://github.com/LLK/scratch-gui/issues/59#issuecomment-456416739

mxmou:
> I'm curious how this is less trouble than just adding a separate color, which users can customize to their own intent?

It isn't less trouble (more trouble actually) but there already 16 colors. I don't think anyone who wants to change the color of a particular element would want to check more than that to find the correct one.

> Adding on @TheColaber's illustrative screenshots—Scratch's block colors are suited for and designed around a generally light color scheme. When we make any userstyle, it's valuable to keep in mind the contrasts and appearance originally intended; dark inputs inside of blocks throw that balance off, IMO.

I wanted to make the blocks themselves dark too but that would make it impossible to use dark mode and custom block colors at the same time. Maybe block input background could be a setting in the custom block colors addon?

TheColaber:
> Maybe block input background could be a setting in the custom block colors addon?

I think this is a good idea, it is about the block's input color after all.

mxmou:
The input background setting doesn't affect block inputs any more. I want to add a setting for those to the custom block colors addon.

WorldLanguages:
Pushed change to get-addon-settings.js that handles the migration
Migration only happens once, and the previous mode is not deleted, it's still stored "just in case"

An easy way to test the migration is running this in the background page, then reloading the extension:
```js
scratchAddons.globalState.addonSettings["editor-dark-mode"] = {selectedMode:"dark-editor", textShadow:true};
chrome.storage.sync.set({ addonSettings: scratchAddons.globalState.addonSettings });
```
Possible values for selectedMode are `3-darker`, `3-dark`, `dark-editor` and `experimental-dark`. Possible values for textShadow are `true` and `false`.
If it migrated it will print `Migrated editor-dark-mode to presets` to the console.
The migration does not affect new users of the extension (AKA users that have no setting called selectedMode in their storage).
If someone updates from pre-v1.4.0, 3.Darker without text shadows will be used (keep in mind Scratch Addons v1.4.0 remapped select IDs about 6 months ago).

WorldLanguages:
@mxmou 
1. I'd be OK moving the list of presets above the list of settings, so that they are more visible. Unless you have a reason not to do that, I'd recommend we switch the order.
2. "Highlight color" setting is only defined in the "dark editor" preset, what am I missing? Shouldn't it be defined in all presets?
3. Because of 2., I left a TODO here: https://github.com/ScratchAddons/ScratchAddons/pull/1805/files#diff-0d52c24886f528dc9e64b55bc525cbb4c91bf8a17fcde0f32520029797d5e770R25
4. I got lost on what the behavior should be for the reset dropdown and default preset. Right now, it says "default" instead of "3.darker", what lines of code are causing that (3.darker being listed as "default")?
5. After all of this, I guess we may want to try to find any tiny little differences we might have missed between the old modes and new presets (and possibly discuss whether some intentional changes for the migration should stay)

webpages/settings/index.html
```diff
@@ -307,24 +329,34 @@ <h1>{{ msg("settings") }}</h1>
...
-                  <button
-                    type="clear-button"
-                    class="large-button clear-button"
-                    :disabled="!addon._enabled"
-                    v-if="setting.type === 'color'"
-                    @click="updateOption(setting.id, setting.default, addon)"
-                  >
-                    {{ msg("reset") }}
-                  </button>
+                  <template v-if="setting.type === 'color'">
```
WorldLanguages, line 332: I'm confused on why there's a template element inside the body

mxmou: https://v3.vuejs.org/guide/conditional.html#conditional-groups-with-v-if-on-template

mxmou:
> I'd prefer adding the color to all presets, and then try to deal with that problem through the settings page UI/UX (possibly allowing users to change what specific settings to apply when setting a preset, and letting addons select the ones that are selected by default - in this case, we would make it so "highlight color" needs to be explicitly enabled when setting a preset, other settings like "text shadows" we might want to do the same).

This might discourage people from using presets the way they were designed (e.g. dark editor with the dark highlight color).

> Main motivation for me to mention that here is the fact 3.Darker and 3.Dark modes (previous to this PR) use a light gray as a highlight color (AFAIK), and this would change it to blue for many users, which is a pretty big change IMO. Tell me if I'm missing something.

3.Darker and 3.Dark always used blue as the highlight color (which is used borders of focused inputs and selected sprites, the new sprite button etc.)

> I see. I don't think we should add a new property yet, but I'd like to see the dropdown say `Default (3.Darker)` instead of just `Default` in that case, if possible?

If it checks if there is exactly one preset that uses the same color as the default, there's still the possibility of a theoretical future addon where this doesn't give the intended result.

WorldLanguages:
> 3.Darker and 3.Dark always used blue as the highlight color (which is used borders of focused inputs and selected sprites, the new sprite button etc.)

You know I'm no expert on this, I have no idea where most colors are used over the editor :P I was talking pretty much specifically about the change of color used when hovering over a context menu item (screenshots: 3.darker mode vs preset)
image
image

Tbh it's not a big deal, but as a user I'm not sure I would like this change at first lol

infinitytec:
I'm liking this. Seems like a practical evolution of theming the editor.

One thing that might be a good change though is to make the highlight color also be the be the context menu highlight color.

mxmou:
@WorldLanguages I didn't notice that 3.Dark(er) changes the color of hovered context menu items. I'm not sure if adding a new color to preserve that behavior is a very good idea.

@infinitytec If "context menu highlight color" is the color of hovered context menu items, then that's already how it works.

mxmou:
> > 2. "Highlight color" setting is only defined in the "dark editor" preset, what am I missing? Shouldn't it be defined in all presets?
> 
> To me it seems like something that users would change to their favorite color, so I wanted to preserve it when switching presets. A problem with this is that switching to Dark Editor and back changes it, so maybe I should just add the color to all presets.

I just noticed it's already defined in all presets. The reset dropdown only shows Dark Editor because presets that use the default color are intentionally omitted.

mxmou:
Text color is now a library. This means that other addons (custom block colors) and, if #2099 is implemented, the addon loader will be able to use it.

WorldLanguages:
> I just noticed it's already defined in all presets. The reset dropdown only shows Dark Editor because presets that use the default color are intentionally omitted.

Good

> Text color is now a library

Good idea! I moved the issue to the v1.14.0 milestone.

@mxmou Epic work, even tho I'll miss the original 3.Dark(er) context menu hover colors :P However if many people ask for a setting like that we might need to add it back anyway
Some little things and this is good to merge for v1.13.0 release in a week or so:
1. Rename the addon to `Editor dark mode and customizable colors` ("customizable" instead of "customized"), I've been told that's better English hehe
2. I've noticed the editor dark mode affects the background color of the fullscreen button, even if the project is in player mode. Probably not intended?
image
3. Fyi, we'll probably remove the `If any colors are incorrect or if text is hard to read, reload the page.` warning from both of your addons after merging #1961, which will allow for userscripts to be dynamically injected.

WorldLanguages:
4. I've also just noticed "Scratch's default color" preset does still make the scrollbars dark in the editor (and I think we'd expect that preset to keep the editor as if the addon was disabled)

mxmou:
> 1\. Rename the addon to `Editor dark mode and customizable colors` ("customizable" instead of "customized"), I've been told that's better English hehe

Done.

> 2\. I've noticed the editor dark mode affects the background color of the fullscreen button, even if the project is in player mode. Probably not intended?

This is intended, currently the project player is only affected by editor dark mode (except the loading screen, until this PR). But it doesn't look very good so changing the full screen button to be styled by website dark mode when not in full screen or in the editor probably makes sense.

> 4\. I've also just noticed "Scratch's default color" preset does still make the scrollbars dark in the editor (and I think we'd expect that preset to keep the editor as if the addon was disabled)

That was a bug with the text color functionality, I fixed it now.

Note that the addon changes the background of hovered Blockly context menu items to the highlight color (Scratch uses a lighter blue by default), even when the default colors preset is selected.

WorldLanguages:
> Note that the addon changes the background of hovered Blockly context menu items to the highlight color (Scratch uses a lighter blue by default), even when the default colors preset is selected.

Ohh... Are you sure we do not want just one more setting to kill 2 birds with one stone? Using the highlight color instead of a dedicated highlight context menu item color setting means 1. there's a noticeable difference between the addon disabled vs. default color preset 2. there's a noticeable difference between old 3.Darker and new 3.Darker (and other modes as well). I'll leave it for you to decide but for me it looks like a no brainer, yes it's one more setting and there are many already, but it makes it way harder to notice that we transitioned to presets and that the addon is enabled, when default colors should be as similar as possible to vanilla Scratch

> But it doesn't look very good so changing the full screen button to be styled by website dark mode when not in full screen or in the editor probably makes sense.

I agree, it messes up with where we draw the line on what gets styled by what addons, but it also looks too bad, since it's the only thing styled in player mode by editor dark mode (except anything sitting above the canvas)

mxmou:
> > Note that the addon changes the background of hovered Blockly context menu items to the highlight color (Scratch uses a lighter blue by default), even when the default colors preset is selected.
> 
> Ohh... Are you sure we do not want just one more setting to kill 2 birds with one stone? Using the highlight color instead of a dedicated highlight context menu item color setting means 1. there's a noticeable difference between the addon disabled vs. default color preset 2. there's a noticeable difference between old 3.Darker and new 3.Darker (and other modes as well). I'll leave it for you to decide but for me it looks like a no brainer, yes it's one more setting and there are many already, but it makes it way harder to notice that we transitioned to presets and that the addon is enabled, when default colors should be as similar as possible to vanilla Scratch

This is what I thought too, and I would do it if there wasn't another problem: Scratch has two different types of context menus. For Blockly context menus it uses the default Blockly style, and for sprite/asset context menus it uses a different, more Scratch-like appearance with the Scratch primary color as the hovered item background.

Unfortunately the only way to make both 3.Dark(er) and default context menus the same as before is to add *two* new settings.

WorldLanguages:
> Unfortunately the only way to make both 3.Dark(er) and default context menus the same as before is to add _two_ new settings.

Even better, a new full row of color settings! /s
For real tho, still worth it in my opinion, 16 color settings vs 18
A hacky alternative is checking whether the context menu color when not highlighted matches these presets (3.Darker, 3.Dark, Scratch default colors) from the userscript. If that's the case, style the items when highlighted using an internal CSS variable that's not exposed as a setting to the proper color, to achieve backwards compatibility / pure vanilla look. If that's not the case, fallback to the highlight color.

mxmou:
I actually think making both types of context menus consistent is a positive change. I'm still not sure if I should add a setting for context menu highlights or not, but it should affect both types of context menus.

The text color and border color of Blockly context menus will be changed anyway, even if the highlight color stays the same.

WorldLanguages:
Ok then. This should be good to merge, but I think there was just 1 thing pending, making the fullscreen button background color be darkened by dark-www instead of editor-dark-mode in player mode (I think we both agreed on doing that, even tho it messes up with our definition on what gets darkened by who)

mxmou:
I fixed the full screen button. The stage border didn't look very good either when website dark mode was disabled so I did the same thing to it.

mxmou:
Now that I made this change I'm no longer sure if making the project player always styled by editor dark mode is a good idea. I want to add a similar color customization system to website dark mode and I think it doesn't make sense if the player is controlled by one addon but everything else on the same page is controlled by a different addon.

WorldLanguages:
@mxmou IMO anything above the canvas (monitors etc.) should be styled by editor dark mode, even on player mode, for consistency

WorldLanguages:
Merge this when we're ready

TheColaber:
Btw `Blockly.Colours` exists and does a lot of the things that this PR is doing for editor-darkmode

mxmou:
> Btw `Blockly.Colours` exists and does a lot of the things that this PR is doing for editor-darkmode

You mean the workspace, scrollbar etc. colors? Those (at least some of them) are overridden by the scratch-blocks `css.js`.

RedGuy12:
Maybe styled by editor DM on the editor page and by website DM on the project, fullscreen, and embed pages?

mxmou:
> @mxmou IMO anything above the canvas (monitors etc.) should be styled by editor dark mode, even on player mode, for consistency

Depending on what we want, consistency across pages or consistency between different elements on the same page.

> Maybe styled by editor DM on the editor page and by website DM on the project, fullscreen, and embed pages?

Currently it works like this:
* project page is styled by website dark mode, except any elements on the stage (monitors, ask prompts), which are styled by editor dark mode
* editor and full screen modes are styled by editor dark mode - I don't think this should be changed
* embed is styled by website dark mode (#2042), elements that would normally be styled by editor dark mode are not dark there

I think it would be better if the project and embed pages were only affected by website dark mode and the other two only by editor dark mode.

WorldLanguages:
> I think it would be better if the project and embed pages were only affected by website dark mode and the other two only by editor dark mode.

Smh that was hard to read
That would mean the monitors would look different depending if the project is in player mode or editor/fullscreen mode right?

mxmou:
@WorldLanguages It would. But it would also mean that if one set the website highlight color to red and the editor highlight color to green, the monitor background would be red in player mode, not green.

WorldLanguages:
can we merge pls

mxmou:
> can we merge pls

Probably, other changes could be made when the website dark mode is rewritten.

WorldLanguages:
@mxmou I noticed editor dark mode + colored context menus means highlight color is not used. May be what we want but just making sure you're aware
image

mxmou:
@WorldLanguages This is intended.

WorldLanguages:
Feedback: 
> There isnt an option to change the variable background color, it always uses the page bg, I don't like the look of the dark variables when using the dark editor mode

Possible solution: a boolean settings on whether monitors (AKA anything above the canvas) should be affected (defaults to true)?

TheColaber:
> There isnt an option to change the variable background color, it always uses the page bg, I don't like the look of the dark variables when using the dark editor mode

I just made this issue from feedback we got from the fan server... #2239

WorldLanguages:
@TheColaber Oh is this a bug?

infinitytec:
image
image
image


What have I done?

WorldLanguages:
Feedback:
> You should add an "Active Tab Text (Color, Costumes, Sound, Variables)" color option to Costumizable Editor Theme settings because it can't be changed. It's always blue. If you make Active Tab Background color to blue the text becomes impossible to read. Or if you just want to make a different color themed editor it just looks bad.

mxmou:
@WorldLanguages I already thought about this. The setting could be called "highlight text" because there are a few other places that use blue text. I don't think each of them should have a separate setting, especially because no preset would need to change it.

PR #1818, opened 2021-03-12
Compress dem svgs

GrahamSH-LLK:
Compresses svgs with svgo


GrahamSH-LLK:
I literally went through every changed icon. They're all identical to the eye

PR #1825, opened 2021-03-12
Create alternative.css

### LINKED ISSUE #1123: Change Sora font to Roboto
### **I don't like the Sora font.**

I just don't like how it looks. Period.

My solution is **NOT** to replace it. Intsead, add an option in _More Settings_ to change the font.

This would be also good for accesibility, as Roboto is much easier to read.

Secret-chest:
Contains DM Sans, to improve readability, please add an option to use this CSS file over the other ones

**Resolves**

Resolves #1123

**Changes**

Adds a new CSS file that can be used over both light and dark. Please make it usable from More Settings.

**Reason for changes**

Sora is hard to read.

**Tests**

No tests, first we need a way to activate the CSS from More Settings

### DIFF
--- a/webpages/settings/alternative.css
+++ b/webpages/settings/alternative.css
@@ -1,0 +1,9 @@
+@import url("../styles/colors.css");
+@import url("../styles/components/tooltips.css");
+@import url("../styles/components/badges.css");
+@import url("../styles/components/categories.css");
+@import url("https://fonts.googleapis.com/css2?family=DM+Sans&display=swap");
+
+body {
+  font-family: "DM Sans", sans-serif;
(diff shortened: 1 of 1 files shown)


webpages/settings/alternative.css
```diff
@@ -0,0 +1,9 @@
+@import url("../styles/colors.css");
+@import url("../styles/components/tooltips.css");
+@import url("../styles/components/badges.css");
+@import url("../styles/components/categories.css");
```
GrahamSH-LLK, line 4: Those imports (except for the font) are useless because index.css is always there.

Issue #1222, opened 2021-01-06
Scratchblocks scroll


RedGuy12:
**Describe the new addon**
Add `overflow-x: scroll;` to scratch blocks to make them horizontally scrollable. Currently, they are cut off at the edge of the screen. This could be a part of the siggy scrolling addon.

PR #1831, opened 2021-03-12
scratchblocks scroll

RedGuy12:
**before**
image

**after**
image

#1222

### DIFF
--- a/addons/full-signature/addon.json
+++ b/addons/full-signature/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Show full areas",
-  "description": "Shows the full signature of users on the forums, stops Scratchblocks from being cut off at the edge of the screen, and shows full What I've been doing on profiles.",
+  "description": "Makes some elements with limited height resizable or scrollable, such as user signatures on the forums and What I've been doing on profiles.",
   "credits": [
     {
       "name": "GrahamSH"
--- a/addons/full-signature/addon.json
+++ b/addons/full-signature/addon.json
@@ -25,7 +25,7 @@
       "default": true
     },
     {
-      "name": "Show scroll for Scratchblocks",
+      "name": "Show scroll for forum Scratchblocks",
       "id": "blocks",
       "type": "boolean",
       "default": true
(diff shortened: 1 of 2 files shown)


GrahamSH-LLK:
Maybe a styled scrollbar? That one looks a little out of place

RedGuy12:
> Maybe a styled scrollbar? That one looks a little out of place

it is styled by dark mode
image
just the first screenshot was taken from the preview iframe which dark mode doesn't have access to

TheColaber:
This was already merged tho...

RedGuy12:
> This was already merged tho...

it was reverted #1237

and I was too lazy to make a new one

mxmou:
> just the first screenshot was taken from the preview iframe which dark mode doesn't have access to

It does have access to it (using a userscript). If its scrollbar CSS isn't injected into the preview then that's a bug.

RedGuy12:
> > just the first screenshot was taken from the preview iframe which dark mode doesn't have access to
> 
> It does have access to it (using a userscript). If its scrollbar CSS isn't injected into the preview then that's a bug.

looks like it's not working

mxmou:
> looks like it's not working

You're right. I'll make an issue about this.

addons/full-signature/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Show full areas",
-  "description": "Shows the full signature of users on the forums, and shows full What I've been doing on profiles.",
+  "description": "Shows the full signature of users on the forums, stops Scratchblocks from being cut off at the edge of the screen, and shows full What I've been doing on profiles.",
```
WorldLanguages, line 3: Not sure on what we should do with this description. We never updated it to reflect we support scroll for "What's happening?". Listing every feature in the description is not very useful

mxmou: Maybe something like *Makes some elements with limited height resizable or scrollable, such as user signatures on the forums and What I've been doing on profiles.*

Not really related, but should it be "on the forums" or "in the forums"? Different addons don't seem to agree about this.

RedGuy12: done

PR #1843, opened 2021-03-13
Fix crash when updating theme settings

### LINKED ISSUE #1841: Error when changing theme settings
**Describe the bug**

When changing the settings of a theme while it's running in a tab, it behaves as if the settings were reset and an error appears in that tab's console. This is because line 52 of `content-scripts/cs.js` calls `injectUserstylesAndThemes` without providing the `addonsWithUserstyles` argument (it should probably be an empty array), causing an error on line 68.

**To Reproduce**

1. Enable editor dark mode
2. Open the editor
3. Change the dark mode version setting
4. Go back to the editor and open the console

**Expected behavior**

The settings would change correctly.

apple502j:
Resolves #1841 

No changelog needed, regression in current prerelease

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -172,9 +172,13 @@ scratchAddons.localEvents.addEventListener("themesUpdated", () => {
       if (tab.url || (!tab.url && typeof browser !== "undefined")) {
         chrome.tabs.sendMessage(tab.id, "getInitialUrl", { frameId: 0 }, async (res) => {
           if (res) {
+            const csInfo = await getContentScriptInfo(res);
             chrome.tabs.sendMessage(
               tab.id,
-              { themesUpdated: (await getContentScriptInfo(res)).themes },
+              {
+                themesUpdated: csInfo.themes,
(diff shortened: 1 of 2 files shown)


background/get-userscripts.js
```diff
@@ -172,9 +172,13 @@ scratchAddons.localEvents.addEventListener("themesUpdated", () => {
       if (tab.url || (!tab.url && typeof browser !== "undefined")) {
         chrome.tabs.sendMessage(tab.id, "getInitialUrl", { frameId: 0 }, async (res) => {
           if (res) {
+            const csInfo = await getContentScriptInfo(res);
             chrome.tabs.sendMessage(
               tab.id,
-              { themesUpdated: (await getContentScriptInfo(res)).themes },
+              {
+                themesUpdated: csInfo.themes,
+                addonsWithUserstyles: csInfo.addonsWithUserstyles,
```
WorldLanguages, line 180: This is not a good idea, we never send two messages at once (they all have one key)

PR #1846, opened 2021-03-13
Add version on variable manager (1.11.1)

Hans5958:
This adds the version number for variable manager feedbacks, because I think Jefallo decides that he won't add one, and an unexpected influx of feedback that are coming.

I use a plus sign instead of a dash since a dash may be confused as a prerelease build. A plus sign is for build metadata, which I think it is more appropriate.

Maybe we need to update our feedback so we can also supply source parameter. We might also need to update notif so it also uses plus for now.

### DIFF
--- a/addons/variable-manager/userscript.js
+++ b/addons/variable-manager/userscript.js
@@ -37,7 +37,7 @@ export default async function ({ addon, global, console, msg }) {
   const feedbackButton = document.createElement("a");
   feedbackButton.className = "sa-var-manager-feedback-button";
   feedbackButton.innerText = msg("feedback");
-  feedbackButton.href = "https://scratchaddons.com/feedback?version=variable-manager";
+  feedbackButton.href = "https://scratchaddons.com/feedback?version=1.11.1+variable-manager";
   feedbackButton.target = "_blank";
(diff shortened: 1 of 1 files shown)


apple502j:
Isn't plus encoded form of space?

Hans5958:
No, that's %20. Unless you are talking Google.

apple502j:
@Hans5958 Both `+` and `%20` are accepted

WorldLanguages:
We're going to remove the feedback button completely from this addon in v1.11.1.

Hans5958:
> 
> 
> @Hans5958 Both `+` and `%20` are accepted

You sure?

> 
> 
> We're going to remove the feedback button completely from this addon in v1.11.1.

Around 5 days only? That's quite fast. Alrighty, then.

mxmou:
> Both `+` and `%20` are accepted

image

WorldLanguages:
Don't start discussing over this please :P

RedGuy12:
> > Both `+` and `%20` are accepted
> 
> image

When you submit a forum it encodes using uri encode except spaces are `+`s

https://stackoverflow.com/questions/2678551/when-to-encode-space-to-plus-or-20/2678602#2678602

PR #1884, opened 2021-03-15
change popup setting styling from @media to class

### LINKED ISSUE #1882: The settings page looks like the addons tab in the popup on small window sizes
**Describe the bug**

For some reason the removal of the top bar when the settings page is opened in the popup (#1006) is done using a media query. This means that on small window sizes (width < 540px) the settings page has no top bar.

**To Reproduce**

1. Open the settings page
2. Decrease the window width until the top bar disappears.

**Expected behavior**

The top bar should be there (removing it should be done by setting a class, not using a media query). Clicking the sidebar button in the top bar when the window is small is supposed to open the sidebar as a dropdown menu, which also appears to have been removed in #1006.

RedGuy12:
**Resolves**

Resolves #1882

**Changes**

changed it. the class was already added before this pr but is never used

**Reason for changes**

...

**Tests**

yes

### DIFF
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -647,52 +647,52 @@ input[type="color"]::-webkit-color-swatch {
   border-radius: 4px;
 }
 
-body.iframe  .badge:not(.beta-badge) {
-    display: none !important;
-  }
-body.iframe  .tooltiptext {
-    display: none;
-  }
-body.iframe  .addon-description {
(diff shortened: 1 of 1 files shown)


GrahamSH-LLK:
This is gonna conflict like heck with my PR. I'll handle it there. Thanks for the PR, tho

RedGuy12:
K

Issue #1234, opened 2021-01-07
"Addonify" popups


WorldLanguages:
Moved from #773

Right now, addon popups are pretty much part of the extension, instead of independent from it and simply addons.
Possible improvements are:

- [x] No need for popups to specify what strings they'll use, so that msg() calls work like in userscripts and persistent scripts
- [ ] No need for addons to set the HTML lang (could be set by the extension popup)
- [x] Shared CSS classes between settings page, extension popup and addon popups
- [ ] Way for addons to detect light mode or dark mode
- [ ] Way for popups to know if they're fullscreen through `addon.self`
- [ ] `addon.*` APIs available in popups, as well as `msg` and `safeMsg`
- [ ] Communicating to userscripts and persistent scripts
- [ ] `window.Vue` and Sora automatically available

RedGuy12:
#1234 oooooo

GrahamSH-LLK:
My future PR will handle 2.

TheColaber:
> No need for addons to set the HTML lang (could be set by the extension popup)

Well, what if the addon popup was opened in a new tab? The extension popup can't do anything about that.

WorldLanguages:
> Well, what if the addon popup was opened in a new tab? The extension popup can't do anything about that.

Good point

WorldLanguages:
Before doing most of these @apple502j we need your help to make popups load all strings of an addon by default, instead of the popup manually having to specify what strings it needs (userscripts and persistent scripts don't have this limitation). Otherwise `msg()` wouldn't work normally.

Hans5958:
updates?

GrahamSH-LLK:
Not on my side. I've been putting it off, cause the next part of the CSS rewrite'll take a while.

WorldLanguages:
> updates?

soon ™️

retronbv:
🎉 #1234

TheColaber:
I think this needs some discussion so let me bring up a few points. This has been moving up the milestones with not much prgress.

*  No need for addons to set the HTML lang (could be set by the extension popup)
   * Like I set before, this cannot always be the case if the addon popup was opened in a new tab.
*  `addon.*` APIs available in popups, as well as `msg` and `safeMsg`
   * This one is a bit tricky and will need the most discussion. Will it be available in the iframe's window? Will it be like addons `export default`? I think I should mention that we will need an `addon-api/popups` if that wasn't clear enough.
[...]

PR #1885, opened 2021-03-15
addon.storage

### LINKED ISSUE #1653: addon.storage API
Would be nice for future addons...

RedGuy12:
# New `addon.storage` API

## Resolves

Resolves #1653 

## Changes

Added new API. The new properties in `manifest.json` are required so 1. I can import from a file that has helper functions and 2. so I can send messages to the background page from the new class.

## Reason for changes

why not

## Tests

yes

## API Documentation

The new API features two methods: `addon.storage.get` and `addon.storage.set`. Both are available in userscripts and background scripts.

### `addon.storage.get(key, mode)`
`key`: Key set with `addon.storage.set` Type string. No need to specify the perfix used internally.
`mode`: Mode used with `addon.storage.set` Type string of value `sync`, `local`, or `cookie`.

### `addon.storage.set(key, value, set)`
`key`: Key for retrieving values with `addon.storage.get` Type string. Will be prefixed with `addon-name/` internally.
`value`: Value to save. Any type.
`mode`: Mode of saving. Type string of value `sync`, `local`, or `cookie`. `sync` will save in `chrome.storage.sync`, `local` in `chrome.storage.local`, and `cookie` in `chrome.cookie`.


TheColaber:
The whole point was not to use scratch's cookies...

RedGuy12:
> The whole point was not to use scratch's cookies...

i can remove the `cookie` mode, that's the only mode that uses cookies

i made it so sensitive info can be stored (like session ids for an account-switching addon I'm making)

TheColaber:
Why can't you just use `document.cookie` then?

TheColaber:
whoops

RedGuy12:
> Why can't you just use `document.cookie` then?

you could, but when you're communicating between userscripts and background scripts (and popups after #1234) it gets kinda complicated. sometimes use `document.cookie` which is a bit of a pain as it's just a long string of text, sometimes `chrome.cookie` which uses promises, etc, etc, etc.

TheColaber:
iirc the only reason we wanted addon.storage was to store something in the extensions so it's in all devices, not just one. We weren't planning on storing sensitive data. I cannot agree that it is hard to mess with document.cookies. See https://www.w3schools.com/js/js_cookies.asp.

RedGuy12:
> iirc the only reason we wanted addon.storage was to store something in the extensions so it's in all devices, not just one. We weren't planning on storing sensitive data. I cannot agree that it is a pain to mess with document.cookies. See [w3schools.com/js/js_cookies.asp](https://www.w3schools.com/js/js_cookies.asp).

so you're saying we should only have the `sync` mode?

TheColaber:
I am saying, that this is not the API that was asked... this one modifies document.cookies (which already means that this can't be done for persistent scripts and popups). Reread this post again for more details. https://github.com/ScratchAddons/ScratchAddons/issues/1653#issuecomment-784266011

RedGuy12:
> I am saying, that this is not the API that was asked... this one modifies document.cookies (which already means that this can't be done for persistent scripts and popups). Reread this post again for more details. [#1653 (comment)](https://github.com/ScratchAddons/ScratchAddons/issues/1653#issuecomment-784266011)

you:
> this could only work by interaction with the chrome.storage API, which is exactly what we need
so store it in `chrome.storage`, which `sync` and `local` do

also you:
> We weren't planning on storing sensitive data. I cannot agree that it is a pain to mess with document.cookies.
so no `cookie mode

and also you:
> the only reason we wanted addon.storage was to store something in the extensions so it's in all devices, not just one
so no `local` mode


so you are saying I should make it always use the `sync` mode, correct?

WorldLanguages:
@RedGuy12 Would be cool to talk on the discord server about this PR later

RedGuy12:
> which already means that this can't be done for persistent scripts and popups

There's a `chrome.cookies` api that let's you and which I used here

PR #1907, opened 2021-03-17
patch so it stops overwriting console

easrng:
**Resolves/changes**

makes it not overwrite console

**Reason for changes**

overwriting globals == bad

**Tests**

none


WorldLanguages:
Where were you overwriting globals?

Issue #10, opened 2020-07-28
Allow for multiple addons using badge


WorldLanguages:
And we should let users choose what addon can use the badge.
Also, what would happen if the addon the user configured to use the badge got disabled? :thinking:

Explosion-Scratch:
Bump?

TheColaber:
What addons would want to use the badge? If disabled, there could be a message in the popup that let's you know.

lisa-wolfgang:
When more functionality for the badge does come around, I think it would be more straightforward to have a `select` type option bar on the message count badge addon and just have that addon get access to the badge, unless we have mutually exclusive addons in the future.

WorldLanguages:
If another addon wants to use the badge, let's open another issue.

return-misaPuding:
Which badge do you mean?

RedGuy12:
The extension icon badge, currently reserved for message count


________________________________

CST1229:
Maybe possible allow 2 max badges, one being at the top of the extension icon (the message count overrides any other addon that tries to take this spot), the other being at the bottom.

PR #1914, opened 2021-03-19
Change Message Count Color

wgyt:
**Resolves**

image

**Changes**

Change the badge color to a dark orange

**Reason for changes**

Made so I can see it better

**Tests**

It works 
image


retronbv:
darn it i wanted to find it so bad lmao

retronbv:
how does it look in light mode?

wgyt:
> how does it look in light mode?

still the same 
image

retronbv:
> > how does it look in light mode?
> 
> still the same
> image

no, you dont understand..... how does it look in browser light mode???
image

Endermaster3690:
what about the scratch color too?

mxmou:
> image

This color is a bit ugly - could you use dark green instead?

wgyt:
> snipped
> no, you dont understand..... how does it look in browser light mode???
> image

image


> > image
> 
> This color is a bit ugly - could you use dark green instead?

I could, should I use a color like the Github comment button?

wgyt:
m

RedGuy12:
why did you merge other prs into this one

make a new pr


p.s. maybe a setting on the addon to choose a color?

WorldLanguages:
Note: this green was the same one used by Scratch Messaging Extension

wgyt:
> why did you merge other prs into this one
> 
> make a new pr
> 
> p.s. maybe a setting on the addon to choose a color?

I didn't. I was just keeping this branch up to date with master
p.s. I guess I could? I probally won't tho.

WorldLanguages:
PR is now ok, but why this color exactly?

mxmou:
> PR is now ok, but why this color exactly?

Maybe `#008000` could be used instead, because it's already used to display the user's own username in comments.

wgyt:
> PR is now ok, but why this color exactly?

I can read it.

> Maybe `#008000` could be used instead, because it's already used to display the user's own username in comments.

That's a good idea.

TheColaber:
Feel like this could be a setting... in more settings... Tbh, I don't mind the green, but I understand if it's hard to read for others.

wgyt:
> Feel like this could be a setting... in more settings... Tbh, I don't mind the current green, but I understand if it's hard to read for others.
> 
> ps... this branch is from the world-languages org??? lol

I should do that soon, but I'm in school right now.

ps... yes lol

wgyt:
I just added the option to change the color, can someone test it?

mxmou:
> if this is merged, can I be invited to the org?

See https://github.com/ScratchAddons/ScratchAddons/discussions/1183#discussioncomment-260417

TheColaber:
With the current way you are doing this... if we have any more addons that have badges, they would also need that setting... is that how it should be done?

wgyt:
> With the current way you are doing this... if we have any more addons that have badges, they would also need that setting... is that how it should be done?

I guess this could be a Scratch Addon wide setting, but no addons other then this one use it so 🤷🏻‍♂️

wgyt:
> With the current way you are doing this... if we have any more addons that have badges, they would also need that setting... is that how it should be done?

https://github.com/ScratchAddons/ScratchAddons/issues/10 would need this to be made part of core

RedGuy12:
I think that this we shouldn't need to bother about moving it to more settings until #10 is implemented.

TheColaber:
This is ok as a temporary solution to just having one single addon using the badge. But I see 2 problems with this PR:
1. The default color should be the current, just so current users don't complain and might not find the new setting. Honestly, I can read white on green.
2. You update the color every 2500 milliseconds. Please use a listener so it updates faster.

If we could first implement another addon that may as well want to use the badge, we can flesh out the details like if all addons should have the same badge color, and how should we decide which get prio, etc.

wgyt:
> This is ok as a temporary solution to just having one single addon using the badge. But I see 2 problems with this PR:
> 
> 1. The default color should be the current, just so current users don't complain and might not find the new setting. Honestly, I can read white on green.

I guess, but what is the default hex right now?

> 2. You update the color every 2500 milliseconds. Please use a listener so it updates faster.

Time to figure out what a listener is, while I'm in class :|

> If we could first implement another addon that may as well want to use the badge, we can flesh out the details like if all addons should have the same badge color, and how should we decide which get prio, etc.

I don't want to deal with this until #10 is done

TheColaber:
`#8ce68c`, `addon.settings.addEventListener("change", callback)`

wgyt:
> `#8ce68c`, `addon.settings.addEventListener("change", callback)`

Ok, all I need to do is figure out the callback, but I'll do that on my lunch break (1 hour from now)

wgyt:
@TheColaber I added what you asked, could you test it?

addons/msg-count-badge/background.js
```diff
@@ -5,5 +5,8 @@ export default async function ({ addon, global, console, setTimeout, setInterval
     addon.badge.color = addon.settings.get('color')
```
TheColaber, line 5: remove this

wgyt: Ok

wgyt: Ok

mxmou:
> The default color should be the current, just so current users don't complain and might not find the new setting. Honestly, I can read white on green.

You can, but not everyone can. Unlike Firefox, Chrome always uses white text on badges, and the current contrast doesn't meet any accessibility guidelines.

wgyt:
> > The default color should be the current, just so current users don't complain and might not find the new setting. Honestly, I can read white on green.
> 
> You can, but not everyone can. Unlike Firefox, Chrome always uses white text on badges, and the current contrast doesn't meet any accessibility guidelines.

Can we just use a darkish blue?

mxmou:
@wgyt I would use green because the badge is already green and most users wouldn't want a completely different color.

wgyt:
Now that we have something that works, can we merge this?

wgyt:
> > Now that we have something that works, can we merge this?
> 
> idk if format code with prettier is required but i think it is

it isn't. It fails ecause it doesn't have access to my fork

WorldLanguages:
> Maybe `#008000` could be used instead, because it's already used to display the user's own username in comments.

That green is now the default. And users can rollback easily:

image

Any suggestions for one or two more presets? Some blue shade, maybe? Pure black?

WorldLanguages:
@wgyt Whoops I can't push commits into this PR. Can you check "allow maintainers to idk what" on this PR?

RedGuy12:
Maybe orange?

PR #1916, opened 2021-03-19
Drag and drop files into the editor

### LINKED ISSUE #1227: Drag and Drop files to Editor
So what if you're on the image or sound editor and you wanted to drag and drop the files to the editor? This should be part of the developer tools.

SheepTester:
**Resolves**

Resolves #1227 

**Changes**

I added the ability to drag files into the costume/sound list, the sprites pane, and the stage sidebar thing. This inserts the dropped files into the same file input that is used by the "Upload Xyz" option in the slidey "Choose a Xyz" menu.

Dragging files into the costume and sound list

Dragging files into the sprites list and stage

**Reason for changes**

Some people, especially those not super versed with computers, prefer dragging files in from their file explorer instead of having to hunt through folders in a file selection window.

**Tests**

I dragged in various files into the drop areas I listed above. If you drag in a file that isn't supported, nothing happens. I'm not sure whether this will be incompatible with other addons, though.

### DIFF
--- a/addons/drag-drop/addon.json
+++ b/addons/drag-drop/addon.json
@@ -12,6 +12,12 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "tags": ["editor", "theme"],
+  "userstyles": [
+    {
+      "url": "dragged-over.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["editor"],
   "enabledByDefault": false
 }
--- a/addons/drag-drop/addon.json
+++ b/addons/drag-drop/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "Drag and drop files to upload",
-  "description": "TODO lol",
+  "name": "Drag and drop files",
+  "description": "Lets you drag files from your file explorer into the sprites pane or costume list.",
   "credits": [
     {
       "name": "Sheep_maker"
--- a/addons/drag-drop/addon.json
+++ b/addons/drag-drop/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Drag and drop files",
-  "description": "Lets you drag files from your file explorer into the sprites pane or costume list.",
+  "description": "Lets you drag images and sounds from your file explorer into the sprites pane or costume/sound list.",
   "credits": [
     {
       "name": "Sheep_maker"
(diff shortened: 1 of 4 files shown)


Explosion-Scratch (approved):
Wow! I'm impressed at how efficient your code is! =D This addon looks great!

addons/drag-drop/userscript.js
```diff
@@ -0,0 +1,42 @@
+export default async function ({ addon, global, console }) {
+  const DRAG_AREA_CLASS = "sa-drag-area";
+  const DRAG_OVER_CLASS = "sa-dragged-over";
```
RedGuy12, line 3: i think we're trying not to make our own classes

addons/drag-drop/userscript.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  droppable(
+    addon.tab.waitForElement('div[class*="sprite-selector_sprite-selector"]'),
+    addon.tab.waitForElement('div[class*="sprite-selector_sprite-selector"] input[class*="action-menu_file-input"]')
+  );
+  droppable(
+    addon.tab.waitForElement('div[class*="stage-selector_stage-selector"]'),
+    addon.tab.waitForElement('div[class*="stage-selector_stage-selector"] input[class*="action-menu_file-input"]')
+  );
```
TheColaber, line 32: does this still work when exitting and reentering the editor?

RedGuy12 (changes requested):

addons/drag-drop/userscript.js
TheColaber: See how it is used in the css... this is perfectly fine.

addons/drag-drop/addon.json
```diff
@@ -0,0 +1,23 @@
...
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "dragged-over.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["editor"],
+  "enabledByDefault": false
```
WorldLanguages, line 22: Add "l10n": true

addons/drag-drop/userscript.js
TheColaber: This does not work when exitting and reentering the editor. Please fix this.

retronbv (approved):
It’s good, just add `"l10n": true`

WorldLanguages:
Excellent, not sure how no one had done this until now - code is so simple and clever as well.

PR #1928, opened 2021-03-20
Use Comlink instead of `template#scratch-addons`, stop spamming contentScriptInfo

### LINKED ISSUE #703: Console log will be flooded on Firefox

### LINKED ISSUE #1659: Using comlink for content script <> page communication
Right now, we make our content script (which can access `chrome.*`) and the page itself (which runs addons) communicate through the `template#scratch-addons` element and its attributes. Instead, we could use the [Comlink](https://github.com/GoogleChromeLabs/comlink) library by Google which internally uses `window.postMessage` to make it so that you can "share" objects between contexts. So instead of having to update an HTML attribute after a global state change, the content script could simply run `_page_.setGlobalState(newGlobalState)`.
Since both contexts share the `window` object (kind of), I made two way communication work by using 2 about:blank iframes in the head of the document (see branch "comlink" on my fork for a demo). It might be possible to make this work with a single iframe, but I won't spend the time trying that, since this works and because the purpose of comlink is to abstract communication, it doesn't really matter at all. It might even be possible with no iframes at all, but for that I would need to look deeply into how the library works.

WorldLanguages:
Resolves #1659 
- Adds Comlink for CS <> page communication. For it to work properly I had to use 4 about:blank iframes, not ideal but not a big deal either. Good part of using 4 iframes is that they message each other, so no message goes through the main Scratch frame, which should help avoid conflicts with other extensions.
- Previously, neither module.js nor theme CSS were injected until there was a `<head>`. That limitation was now removed, but userscripts should still be executed after the `<head>` is created.
- Use `import.meta.url` instead of `template#scratch-addons[data-path]`
- Introduce `scratchAddons.methods.copyImage` for copying images on Firefox without needing Tab.js to use the template element
- I could list every change, but TL;DR is that everything that previously used HTML attributes in the `<template id="scratch-addons">` element now uses Comlink.
- Code is way cleaner now, we can magically access objects from "the other world" thanks to Comlink :)

Resolves #703 (not done, but good enough to test Comlink)
- After some testing, I concluded there's no need to spam contentScriptInfo - requiring the content script to ask for it when it executed works just fine (reminder: the purpose of all this "spam" was to make "website dark mode" loads fast enough so that user does not notice flicker)
- However, we still want to preload what addons and themes will be needed in that page to load them ASAP, so we'll have a `Map` object that caches contentScriptInfo by listening to `chrome.webRequest.beforeRequest`, which obviously gets fired before the content script has a chance to execute while the page is loading.

### DIFF
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -51,7 +51,7 @@ const page = {
       );
     }
   },
-  setMsgCount({count}) {
+  setMsgCount({ count }) {
     pendingPromises.msgCount.forEach((promiseResolver) => promiseResolver(count));
     pendingPromises.msgCount = [];
   },
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -71,8 +71,7 @@ function onDataReady() {
 
   scratchAddons.methods = {};
   scratchAddons.methods.getMsgCount = () => {
-    // TODO: not sure why we do this (pending promises arr), just transitioning to comlink for now
-    _cs_.requestMsgCount();
+    if (!pendingPromises.msgCount.length) _cs_.requestMsgCount();
     let promiseResolver;
     const promise = new Promise((resolve) => (promiseResolver = resolve));
     pendingPromises.msgCount.push(promiseResolver);
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -59,7 +59,6 @@ const page = {
 Comlink.expose(page, Comlink.windowEndpoint(comlinkIframe4.contentWindow, comlinkIframe3.contentWindow));
 
 function onDataReady() {
-  debugger;
   const addons = page.addonsWithUserscripts;
 
   scratchAddons.l10n = new Localization(page.l10njson);
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -54,12 +54,12 @@ const page = {
   setMsgCount(count) {
     pendingPromises.msgCount.forEach((promiseResolver) => promiseResolver(count));
     pendingPromises.msgCount = [];
-  }
+  },
 };
 Comlink.expose(page, Comlink.windowEndpoint(comlinkIframe4.contentWindow, comlinkIframe3.contentWindow));
 
 function onDataReady() {
-debugger;
+  debugger;
   const addons = page.addonsWithUserscripts;
 
   scratchAddons.l10n = new Localization(page.l10njson);
@@ -81,7 +81,7 @@ debugger;
   };
   scratchAddons.methods.copyImage = async (dataURL) => {
     return _cs_.copyImage(dataURL);
-  }
+  };
 
   const runUserscripts = () => {
     for (const addon of addons) {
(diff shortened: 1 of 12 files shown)


content-scripts/inject/module.js
```diff
@@ -1,43 +1,107 @@
...
+    return this._globalState;
+  },
+  set globalState(val) {
+    const set = (this._globalState = val);
+    scratchAddons.globalState = val;
+    return set;
+  },
+
+  l10njson: null, // Only set once
+  addonsWithUserscripts: null, // Onlu set once
```
WorldLanguages, line 28: Fix typo 😅

jeffalo:
i don't know what this means but if it means it won't spam the console with content script update then im happy

mxmou:
> Previously, neither module.js nor theme CSS were injected until there was a `<head>`. That limitation was now removed, but userscripts should still be executed after the `<head>` is created.

I probably don't understand this, but how do styles override Scratch's ones if they're injected before the head?

WorldLanguages:
> I probably don't understand this, but how do styles override Scratch's ones if they're injected before the head?

I haven't looked at that deeply yet. All I meant is that currently (on master) themes are injected

WorldLanguages:
> I probably don't understand this, but how do styles override Scratch's ones if they're injected before the head?

I haven't looked at that deeply yet. All I meant is that currently (on master) themes are guaranteed to be injected after the `<head>` is available, and in the current state of this PR that's not the case. It still looks like themes are still injected after `<head>` but timing is sketchy so that might not be the case in 100% of page loads.

GrahamSH-LLK:
How would this work with multiple SAs? Could we set the `<iframe>` ids to something unique to the extension install? ie. hashing the install time? This could add nice functionality.

apple502j:
Multiple SA has never worked, and should never do.

WorldLanguages:
@GrahamSH-LLK Extension still expects a single `window.scratchAddons` tho.

GrahamSH-LLK:
> @GrahamSH-LLK Extension still expects a single `window.scratchAddons` tho.

Hmm, could it be window[(unique id)]?

WorldLanguages:
> > @GrahamSH-LLK Extension still expects a single `window.scratchAddons` tho.
> 
> Hmm, could it be window[(unique id)]?

No, if we wanted to have multiple `scratchAddons` objects we should directly pass them to the `new Addon()` constructor.

GrahamSH-LLK:
> > > @GrahamSH-LLK Extension still expects a single `window.scratchAddons` tho.
> > 
> > 
> > Hmm, could it be window[(unique id)]?
> 
> No, if we wanted to have multiple `scratchAddons` objects we should directly pass them to the `new Addon()` constructor.

Can we do that?

PR #1937, opened 2021-03-22
Disable cat blocks scripts on april fools

### LINKED ISSUE #1896: disable cat blocks addon when it's april fools day
april fools day for 2021 currently seems to have catblocks planned. we should disable catblocks on our end to ensure ntohing wakcy happens

TheColaber:
**Resolves**

Resolves #1896

**Changes**

just `return`s

**Reason for changes**

So no conflicts with this year.

**Tests**

no but

### DIFF
--- a/addons/cat-blocks/userscript.js
+++ b/addons/cat-blocks/userscript.js
@@ -1,4 +1,7 @@
 export default async function ({ addon, global, console }) {
+  const now = new Date().getTime() / 1000;
+  // April fools
+  if (now < 1617364800 && now > 1617192000) return;
   // TODO: Switch completely to Blockly?
   // Well anyway future self, here you go: https://github.com/LLK/scratch-blocks/compare/hotfix/totally-normal-2020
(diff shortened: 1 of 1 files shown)


apple502j:
Can you make it so that it detects whether Scratch is running AF or not?

addons/cat-blocks/userscript.js
```diff
@@ -1,4 +1,7 @@
 export default async function ({ addon, global, console }) {
+  const now = new Date().getTime() / 1000;
+  // April fools
+  if (now < 1617364800 && now > 1617192000) return;
```
WorldLanguages, line 4: We don't know when the ST will decide to enable/disable april fools, might not match our schedule

WorldLanguages:
> Can you make it so that it detects whether Scratch is running AF or not?

@apple502j Can you suggest a way to do that?

TheColaber:
I was thinking of checking if one of the blockly constants were changed but /shrug

WorldLanguages:
After some testing, everything seems to work fine even if we run cat blocks above cat blocks.

Issue #1340, opened 2021-01-16
Automate merging localization updates


Hans5958:
@apple502j, isn't it easier to automate the merging of the update PRs? I could help with that. Close if it is not needed.

P.S. pls use rebase/squash?

apple502j:
We sometimes only want to push to Transifex, plus I'm usually available every day.

PR #1941, opened 2021-03-22
Danger PR checks

GrahamSH-LLK:
This adds WIP checks on PRs.

Currently, it tells you if you changed libraries but not Credits.md or licenses-info.json.
It also fails if you don't add a description, and warns if you don't have files prettified.


dangerfile.js
```diff
@@ -2,7 +2,7 @@ import { message, danger, warn } from "danger";
 const prettier = require("prettier");
 const fs = require("fs");
 var path = require("path");
-
+message("Thanks for PRing! Someone will be along to check it shortly!");
```
WorldLanguages, line 5: no pls

apple502j:
In most cases it doesn't require changes (AKA first-party libraries) - and it's still somewhat maintainable by human (or fruit).

Hans5958:
Eh, you just closed for a single feature for the whole thing. That's not how it works.

apple502j:
We don't need to worry about prettiers not running (they'll run when the commit goes to master, and doesn't break stuff) - and diffs appear on UI, library detection is flawed, it's useless

dangerfile.js
GrahamSH-LLK: How about I only show that if it's a first-time contributor?

GrahamSH-LLK:
> We don't need to worry about prettiers not running (they'll run when the commit goes to master, and doesn't break stuff) - and diffs appear on UI, library detection is flawed, it's useless

That's why they aren't fails, just warns. The point of this is to lower our workload. Someone always says "Turn on prettier", etc. As I add more checks it should become more useful for us.

RedGuy12:
why do we need this

GrahamSH-LLK:
> why do we need this

This lets us have a single place on PRs to look for information. It adds checks that we didn't have before, and overall improve the experience of maintainers.

Hans5958 (changes requested):
Standby.

GrahamSH-LLK:
> Standby.

Standby for what?

Hans5958:
Alright, here's a comment of what I think about it, while addressing all of these comments. 

> In most cases it doesn't require changes (AKA first-party libraries) - and it's still somewhat maintainable by human (or fruit).

> why do we need this

Here's an assessment that, based on my experience, you aren't likely to read it.

Within my investigation, I can conclude that Danger is a tool that has a potential to optimize code review. It checks for errors that can be checked by lines of code instead of a human, which makes code reviewing more effective, and could make PRs being merged faster.* 

How is it effective, you ask again? It checks for common errors that can be checked by code, especially errors that are missed just because someone didn't read the docs properly. These errors could be resolved faster, instead of waiting of someone to review in different time zones. This could also decrease human error, because these simple errors could be missed by a human eye, causing either a huge bug that needs a patch, or a small one that took ages to be found.

This could also let us present the checks in a better view, instead of being buried deep down on the CI logs. This is what I thought when developing the CI, because there are some that still asking why it went wrong, etc. The checks by Danger could solve this easily. 

I can see that the checks that are implemented are still minimal and looked kinda useless, but I am certain that as time goes on, it could be very useful on a day-to-day basis for someone as a reviewer.

> We don't need to worry about prettiers not running

We actually don't need to worry about all the checks either, because none of it are required.

> diffs appear on UI

Tell that to the novice who probably can't find the issue due because they are starting coding JS.

> library detection is flawed

So improve it? Don't see why we can't do it.

> and it's still somewhat maintainable by human (or fruit).

...and we are reaching to the point where the argument of every automation tool being rejected is that "it's still maintainable." Just like #1340, I don't see why automating something, so we can focus on other things, is bad. I know something could go wrong, but the risks are minimal. The positives are larger compared to the negatives.

----

* Of course in terms of the code reviewing process being faster, not about the merge process itself

.github/workflows/danger.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
```
Hans5958, line 2: Remove this?

.github/workflows/danger.yml
```diff
@@ -0,0 +1,32 @@
...
+
+name: Danger CI
```
Hans5958, line 4: Rename this to "Danger Check" or something? It is already an CI.

.github/workflows/danger.yml
```diff
@@ -0,0 +1,32 @@
...
+    - uses: actions/checkout@v2
+    - name: Setup Node.js environment
+      uses: actions/setup-node@v2.1.5
+    - name: Generate token
+      uses: tibdex/github-app-token@v1
+      id: generate-token
+      with:
+        app_id: ${{ secrets.BOT_APP_ID }}
+        private_key: ${{ secrets.BOT_PRIVATE_KEY }}
+    - run: npm install prettier
```
Hans5958, line 26: We should find a way, so we can have `package.json` and the `Dangerfile` that is not on the main repository.

dangerfile.js
```diff
@@ -0,0 +1,49 @@
...
+const fs = require("fs");
+let pr = danger.github.pr;
+if (pr.body.length === 0) {
+  fail("Please add a description to your PR.");
+}
+
+var path = require("path");
+if (pr.author_association == "FIRST_TIME_CONTRIBUTOR") {
+  message(
+    "Since this is your first contribution, here are some links while you wait for a review.\n[Our contributing guidelines](https://github.com/ScratchAddons/ScratchAddons/blob/master/CONTRIBUTING.md)"
```
Hans5958, line 12: You should go with a friendly tone like "hello, this is your first contribution, blablabla". I'm actually going to make this later with https://probot.github.io/apps/welcome/, which also includes welcoming on their first PR and issue, but this could be it for now.

https://github.com/ScratchAddons/ScratchAddons/blob/master/CONTRIBUTING.md is quite dead, let's remove it for now?

dangerfile.js
```diff
@@ -0,0 +1,49 @@
...
+  const filePath = path.basename(file, extname);
+  if (extname == ".svg") return;
+  let check = prettier.check(readFile, { filepath: filePath + extname });
+  if (!check) badFiles.push(file);
+}
+if (badFiles.length > 0) {
+  warn(
+    `${danger.github.utils.fileLinks(
+      badfiles
+    )} need to be formatted with Prettier. To enable it, go to https://github.com/<yourusername>/ScratchAddons/actions and enable it.`
```
Hans5958, line 33: Change this so it also mentions that you can also do Prettier yourself. Something like this.

`... need to be formatted with Prettier. Please format your code using Prettier, or go to [link] to enable formatting automation.`

I still don't know how it works, but we should put the file list below, using `markdown()`.

GrahamSH-LLK:
I'm just brainstorming, but here are some checks we could add.

- Userscripts that go on /projects/* should have while(true)
- New addons should be in addons.json
- Addons.json should have proper JSON syntax
- Manifest.json should be checked against a JSON schema? https://github.com/PrsPrsBK/ffext-manifest-schema/blob/master/ffext.json looks pretty good.
And others as we go along.

dangerfile.js
```diff
@@ -0,0 +1,49 @@
...
+  );
+}
+if (!pr.title.includes("Translation update:")) {
+  const modifiedMD = danger.git.modified_files.join("- ");
+  message(`This PR has ${pr.additions} additions and ${pr.deletions} deletions.`);
```
Hans5958, line 38: Please no? This info is easy to access (the top-right corner). I think for this kind of info (size, contributors, etc) it should be put using `markdown()`. I don't know, just a speculation. I just hope he doesn't comment for every commit that is pushed.

Hans5958 (changes requested):
That's for now. This could be improved later, but I think this review would make this an MVP. 

----

> Standby for what?

This.

dangerfile.js
GrahamSH-LLK: No, it edits the original message!

Hans5958: Oh, well, I didn't saw the message being edited. Other than that, I think this is kinda useless, eh?

GrahamSH-LLK: Yeah, I'll remove it.

GrahamSH-LLK:
Hmm, it shouldn't make another comment...

apple502j:
Can I opt-out of this stupid stuff? I only want human comments

GrahamSH-LLK:
> Can I opt-out of this stupid stuff? I only want human comments

I can add an opt-out array if you want.

Hans5958:
@GrahamSH-LLK Why not using the bot token? It seems the edit works if it is from the bot, due to the limitations of `GITHUB_TOKEN`.

Also, I think it removes if there are no issues?

GrahamSH-LLK:
> @GrahamSH-LLK Why not using the bot token? It seems the edit works if it is from the bot, due to the limitations of `GITHUB_TOKEN`.
> 
Hmm, it was sending new messages every time. I couldn't get it to work so I reverted.
> Also, I think it removes if there are no issues?

Yep.

Issue #1673, opened 2021-02-26
Migrate All Addons to Auto-refresh


TheColaber:
tl;dr We need to make all addons not need a refresh to update.

After #1500 is merged (Which should be soon), we will have the ability to make all current theme addons that need a refresh to not have one when the addon is disabled while it is in use. Right now I think there is only two addons (One we have done already) that we need to migrate.

Secondly, after we have a PR that allows a new "late" event (An event that should be fired if the page has already been loaded and an addon was enabled "late"), this should allow all existing themes that need a refresh to be enabled no longer need one.

Finally, After a PR is created and merged for #783, we now have the ability for all addons to no longer need a refresh to update their state. I think from then on we would really want any new addon PRs to include these new events and variables. For example, if a userscript was added "late", we would have something such as `addon.self.late`, which would be a boolean. Some addons might need to do something different depending on if they were late or not. Addons that have already been added should try to get migrated too.

Weredime:
`default-project` won't need auto-refresh as it works only once (which is the projects creation flow) and if we cancel before default-project is finished it may cause errors

WorldLanguages:
> `default-project` won't need auto-refresh as it works only once (which is the projects creation flow)

What about File>New?

Weredime:
> > `default-project` won't need auto-refresh as it works only once (which is the projects creation flow)
> 
> What about File>New?

Forgot about that

Weredime:
Quick script I wrote  that calculates how many addons we need to migrate
```js
const fs = require("fs");

const addons = require("./addons/addons.json");

const fullRefresh = [];
const partialRefreshEnable = [];
const partialRefreshDisable = [];
const noRefresh = [];
for (const addon of addons) {
    if (addon.startsWith("//")) continue;
    const manifest = JSON.parse(
        fs.readFileSync(`addons/${addon}/addon.json`)
    );
    if (manifest.dynamicEnable && manifest.dynamicDisable) {
        fullRefresh.push(addon);
    } else if (manifest.dynamicEnable) {
        partialRefreshEnable.push(addon);
    } else if (manifest.dynamicDisable) {
        partialRefreshDisable.push(addon);
    } else {
        noRefresh.push(addon);
    };
}
for (const ref of fullRefresh) {
    console.log(`✅ ${ref} (https://github.com/ScratchAddons/ScratchAddons/tree/master/addons/${ref})`);
}
for (const ref of partialRefreshEnable) {
[...]

PR #1961, opened 2021-03-24
Dynamically enabling and disabling, changing "theme" tag functionality

### LINKED ISSUE #783: Userscript autorefresh

### LINKED ISSUE #1672: Rethink what the "theme" tag does
(Note: this is not a 100% complete idea, but writing helps a ton - and it's also a good idea to document how stuff works right now)

As of now, the "theme" tag does 4 things:
1. Settings page considers the addon to have theme category
2. Its userstyles are injected as soon as possible to the page, using inline `<style> /* */ </style>` elements
3. If the addon is enabled, disabled, or its settings change, all of its current userstyles are removed and readded according to the new value of addon settings (exception: if style starts with `/* sa-autoupdate-theme-ignore */`, it's only added to the page on first load)
4. Its userstyles always have more priority than non-theme userstyles.

Ideally specifying a tag shouldn't have side effects, so I suggest:
1. "theme" tag will only determine the category for the settings page
2. If an addon wants its userstyles to be injected as inline styles, it could specify that through a manifest property. It would apply for all styles.
3. If any addon is disabled, and its manifest specifies `"dynamicDisable": true`, its styles are removed/disabled - these styles can be inline or not. If reenabled, its styles are added back. This would also happen if the addon changes settings, which may result in different userstyles being matched. If the addon has userscripts, they get the `disabled` and `reenabled` events **after** its styles were updated, and `addon.settings` is up to date.
4. If any addon is enabled (but wasn't enabled when the page loaded), and its manifest specifies  `"dynamicEnable": true`, its styles are added to the page - these styles can be inline or not. If the addon has userscripts, they should be ran as well (see issue #783)
5. Only if the manifest specifies `"updateUserstylesOnSettingsChange": true`, the styles for the addon are rematched and reinjected if settings changed for the addon. Note that CSS variables would still update automatically even if this is set to false - this is about the styles themselves changing, like going from 3.Dark to 3.Darker on editor dark mode.
6. Whether an addon has a theme tag, or its styles are inline or not, should not impact its style priority. The priority should still be specified on `/addons/addons.json` - first addon in list has highest priority, and gets its styles injected last in the head (and eventually after body? see #701)

About the 2 current uses of `/* sa-autoupdate-theme-ignore */``:
1. `editor-theme3` uses it in its main css file. This addon's javascript adds some dynamic styles to its inline `<style>` tag (previously it added these styles at the end of the body, but it was moved there to preserve style priority with other addons), so it can only be enabled after a refresh, and disabled after a refresh. 
If we made this CSS file act normally, it would be removed after the addon is disabled, which would work. After the addon is reenabled, the `<style>` tag would be created again (TODO: if we disabled the style instead of removing it, would it work?), so these dynamic styles should be readded. If the addon is enabled but it was disabled when the page loaded, the userscript and userstyle could be injected normally and it would work just fine (but we'd need userscript autorefresh support).
To avoid changing a color setting addon from replacing the style element, thus removing the extra styles added by JavaScript, this addon would not set `updateUserstylesOnSettingsChange` to true.
2. I'm not 100% sure why `scratchr2` uses `/* sa-autoupdate-theme-ignore */` in the `set-default-var-value.js` file I made myself, lol

TheColaber:
#### Added Addon APIs
- addon.self.enabledLate (boolean)
- disabled (event)
- reenabled (event)
#### Added Manifest Properties
- dynamicEnable (boolean)
- injectAsStyleElt (boolean)
- updateUserstylesOnSettingsChange (boolean)
#### Migrated Addons
- confirm-actions
- editor-colored-context-menus
- scratchr2
- 60FPS
- block-switching
- block-palette-icons
- clones
- editor-stepping
- mouse-pos
- editor-searchable-dropdowns
- pause
- youtube-fullscreen - dynamicEnable only

#### Removed
- /* sa-autoupdate-theme-ignore */
#### Others
- data-sa-hide-if-disabled
- moved sourceURL and %addon-self-dir% to be evaluated in the background page

### Needs Discussion
- Rename `injectAsStyleElt`?
- Rename `updateUserstylesOnSettingsChange`?

### Related Issues
- #1673 - Migrating more addons to autorefresh
- #700 - Order of addon userstyle being append should be based on addons.json
- Resolves #783
- Resolves #1672


Explosion-Scratch:
> todo: explain what this PR is.

Yeah, I'm not sure what this does lol

background/get-userscripts.js
```diff
@@ -17,61 +17,104 @@ function getL10NURLs() {
   return urls;
 }
 
+scratchAddons.localEvents.addEventListener("addonEnabled", ({ detail }) => {
+  const { addonId, manifest } = detail;
+  chrome.tabs.query({}, (tabs) =>
+    tabs.forEach((tab) => {
+      if (tab.url || (!tab.url && typeof browser !== "undefined")) {
+        chrome.tabs.sendMessage(tab.id, "getInitialUrl", { frameId: 0 }, async (res) => {
```
WorldLanguages, line 25: I'm not sure Firefox likes async callbacks, I had trouble with them once

TheColaber: lol then should I send a promise?

background/get-userscripts.js
```diff
@@ -17,61 +17,104 @@ function getL10NURLs() {
...
+
+            console.log(userscripts, userstyles);
+            chrome.tabs.sendMessage(
+              tab.id,
+              {
+                dynamicAddonEnabled: {
+                  scripts: userscripts,
+                  userstyles,
+                  addonId,
+                  injectAsStyleElt: manifest.injectAsStyleElt,
```
WorldLanguages, line 37: `!!manifest.injectAsStyleElt`, might be `undefined` (?)

background/get-userscripts.js
```diff
@@ -17,61 +17,104 @@ function getL10NURLs() {
...
+      }
+    })
+  );
+});
+scratchAddons.localEvents.addEventListener("addonDisable", ({ detail }) => {
+  const { addonId, manifest } = detail;
+  chrome.tabs.query({}, (tabs) =>
+    tabs.forEach((tab) => {
+      if (tab.url || (!tab.url && typeof browser !== "undefined")) {
+        chrome.tabs.sendMessage(tab.id, "getInitialUrl", { frameId: 0 }, async (res) => {
```
WorldLanguages, line 54: (Duplicate) I'm not sure Firefox likes async callbacks, I had trouble with them once

background/get-userscripts.js
```diff
@@ -17,61 +17,104 @@ function getL10NURLs() {
...
+            chrome.tabs.sendMessage(tab.id, { dynamicAddonDisable: { addonId } }, { frameId: 0 });
+          }
+        });
+      }
+    })
+  );
+});
+
+async function getAddonData({ addonId, manifest, url }) {
+  const promices = [];
```
WorldLanguages, line 65: It's spelled `promises`

background/get-userscripts.js
```diff
@@ -17,61 +17,104 @@ function getL10NURLs() {
...
+  const userstyles = [];
+  for (const style of manifest.userstyles || []) {
+    if (userscriptMatches({ url }, style, addonId))
+      if (manifest.injectAsStyleElt) {
+        promices.push(
+          fetch(chrome.runtime.getURL(`/addons/${addonId}/${style.url}`))
+            .then((res) => res.text())
+            .then((text) => {
+              userstyles.push(text);
+            })
```
WorldLanguages, line 84: This pattern might cause race conditions, see #700

background/get-userscripts.js
```diff
@@ -17,61 +17,104 @@ function getL10NURLs() {
...
     l10njson: getL10NURLs(),
     globalState: {},
     addonsWithUserscripts: [],
     addonsWithUserstyles: [],
-    themes: [],
   };
-  const fetchThemeStylesPromises = [];
-
-  for (const { addonId, manifest } of scratchAddons.manifests) {
+  for (let i = 0; i < scratchAddons.manifests.length; i++) {
```
WorldLanguages, line 103: for of / for in should work
or even forEach!

background/handle-settings-page.js
```diff
@@ -24,24 +24,13 @@ chrome.runtime.onMessage.addListener(function (request, sender, sendResponse) {
     chrome.storage.sync.set({
       addonsEnabled: scratchAddons.localState.addonsEnabled,
     });
-
+    const manifest = scratchAddons.manifests.find((a) => a.addonId === addonId).manifest;
```
WorldLanguages, line 27: Let's try to call it something other than `a`, for consistency

background/get-userscripts.js
TheColaber: no i need a `for`, for a reason...
1. I need an index
2. iirc, I don't want to return or something, i forgot

background/handle-settings-page.js
```diff
@@ -24,24 +24,13 @@ chrome.runtime.onMessage.addListener(function (request, sender, sendResponse) {
...
-              target: "self",
-              name: newState ? "reenabled" : "disabled",
-              addonId,
-            },
-          })
-      )
-    );
+    if (dynamicEnable && newState)
+      scratchAddons.localEvents.dispatchEvent(new CustomEvent("addonEnabled", { detail: { addonId, manifest } }));
+    if (dynamicDisable && !newState)
```
WorldLanguages, line 32: Would prefer explicitly comparing newState, `newState === true` and `newState === false`, for clarity

content-scripts/cs.js
```diff
@@ -89,43 +89,48 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
-  for (const addon of userstyles || []) {
-    for (const userstyle of addon) {
+function addStyle(addon) {
+  for (let userstyle of addon.styles) {
+    if (addon.injectAsStyleElt) {
+      sessionEnabledThemes.add(addon.addonId);
+
+      // Replace %addon-self-dir% for relative URLs
+      userstyle = userstyle.replace(/\%addon-self-dir\%/g, chrome.runtime.getURL(`addons/${addon.addonId}`));
+      //userstyle += `\n/*# sourceURL=${styleUrl} */`;
```
WorldLanguages, line 105: Why was this commented? Is the `styleUrl` not available? It should, so that we can provide a source map

TheColaber: commented for now, until i send styleUrl back, it was removed bc we merged userstyles and themes.

content-scripts/cs.js
```diff
@@ -89,43 +89,48 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
-      const style = document.createElement("style");
-      style.classList.add("scratch-addons-theme");
-      style.setAttribute("data-addon-id", theme.addonId);
-      style.textContent = css;
-      if (document.body) document.documentElement.insertBefore(style, document.body);
-      else document.documentElement.appendChild(style);
-    }
+}
+function removeAddonStyles(addonId) {
+  document.querySelectorAll(`[data-addon-id='${addonId}']`).forEach((style) => style.remove());
```
WorldLanguages, line 127: Alternatively, we could set `.disabled = true`, works both for style and link elements. Not sure if it's worth doing, tho.

content-scripts/inject/module.js
```diff
@@ -36,6 +36,10 @@ const page = {
 
   fireEvent(info) {
     if (info.addonId) {
+      if (info.name === "enable") {
+        runAddonUserscripts({ ...info, enabledLate: true });
+        return;
+      }
```
WorldLanguages, line 42: I'd prefer a dedicated message for this - no need to make the background page directly call this.

content-scripts/inject/run-userscript.js
```diff
@@ -1,7 +1,7 @@
 import Addon from "../../addon-api/content-script/Addon.js";
 
-export default async function runAddonUserscripts({ addonId, scripts, traps }) {
-  const addonObj = new Addon({ id: addonId, traps });
+export default async function runAddonUserscripts({ addonId, scripts, traps, enabledLate }) {
```
WorldLanguages, line 3: Forgot a default value for `enabledLate`? Should be `false`.
Also, I think we can remove the `traps` parameter which is now unused.

background/get-userscripts.js
WorldLanguages: No. Make it a sync function. If you need to use await, use `.then()` or an async IIFE inside the non-async function.

background/get-userscripts.js
```diff
@@ -17,61 +17,117 @@ function getL10NURLs() {
...
+      }
+    })
+  );
+});
+scratchAddons.localEvents.addEventListener("addonDisable", ({ detail }) => {
+  const { addonId, manifest } = detail;
+  chrome.tabs.query({}, (tabs) =>
+    tabs.forEach((tab) => {
+      if (tab.url || (!tab.url && typeof browser !== "undefined")) {
+        chrome.tabs.sendMessage(tab.id, "getInitialUrl", { frameId: 0 }, (res) => {
```
WorldLanguages, line 54: You do not need the initial URL here

TheColaber:
awaiting @WorldLanguages to work on the reinjection part

background/get-userscripts.js
TheColaber: Not so sure... this is the userstyles between an addon... I don't think an addon's own files should conflict. But if they do, we can add a simple sort function based on how the manifest lists the userstyles.

TheColaber: It's also worth noting that this is how it worked before this PR...

webpages/settings/index.js
```diff
@@ -543,22 +543,22 @@ const vue = (window.vue = new Vue({
...
               ? []
-              : [...new Set([...res.userscripts, ...res.activeThemes, ...res.userstyles])].filter((runningAddonId) => {
-                  // Consider addons with "dynamicUserscriptDisable": true
+              : [...new Set([...res.userscripts, ...res.userstyles])].filter((runningAddonId) => {
+                  // Consider addons with "dynamicDisable": true
                   // If those are running on the page, their "is running on this tab"
                   // status should be the same as their "is enabled" status
                   const manifest = this.manifests.find((manifest) => manifest._addonId === runningAddonId);
                   if (manifest.dynamicDisable && !manifest._enabled) return false;
                   return true;
```
TheColaber, line 552: @WorldLanguages I need clarification on what's happening right here...

background/get-userscripts.js
WorldLanguages: > We can add a simple sort function based on how the manifest lists the userstyles.

It's as simple as reserving all needed indices of the array, then replacing the correct item depending in the order established in the manifest, instead of pushing to the end.

> It's also worth noting that this is how it worked before this PR...

No. As of now, theme userstyles (injected as style elements) are provided through a `styleUrls` array that is properly ordered, and a `styles` property is used to map URLs to the actual contents of the style element. See #700 for more info.

webpages/popup/index.js
```diff
@@ -48,21 +51,25 @@ const vue = new Vue({
       }
     },
     iframeSrc(addonId) {
+      // Problem here
+      if (addonId !== this.currentPopup._addonId) return;
```
WorldLanguages, line 55: What does this solve?

content-scripts/cs.js
```diff
@@ -108,43 +108,57 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
-      style.setAttribute("data-addon-id", theme.addonId);
-      style.textContent = css;
-      if (document.body) document.documentElement.insertBefore(style, document.body);
-      else document.documentElement.appendChild(style);
-    }
+}
+function removeAddonStyles(addonId) {
+  // Instead of actually removing the style/link element, we just disable it.
+  // That way, if the addon needs to be reenabled, it can just enable that style/link element instead of readding it.
+  // This helps with load times for link elements.
```
WorldLanguages, line 155: Source: :poop: :P

webpages/popup/index.js
TheColaber: I forgot to move the `// Problem here` but I need to look more into it... iframeSrc sometimes gets called for no reason...

content-scripts/cs.js
TheColaber: ?

addons/editor-dark-mode/addon.json
```diff
@@ -15,6 +15,10 @@
       "link": "https://github.com/mxmou/customize-scratch"
```
WorldLanguages, line 15: This file will conflict whenever editor dark mode refactor is merged

addons/dark-www/addon.json
```diff
@@ -8,6 +8,9 @@
       "link": "https://github.com/mxmou/customize-scratch"
```
WorldLanguages, line 8: This one as well

addons/editor-stepping/userscript.js
```diff
@@ -29,27 +29,30 @@ export default async function ({ addon, global, console }) {
...
           }
         });
-        if (!childblock && block && block.svgPath_) {
-          block.svgPath_.style.filter = "url(#blueStackGlow)";
-        }
       });
-    });
-    oldStep.call(this);
+    }
+    oldStep.call(this, ...args);
```
WorldLanguages, line 56: These edits shouldn't belong on thsi PR

addons/editor-theme3/theme3.js
```diff
@@ -105,5 +105,5 @@ export default async function ({ addon, global, console }) {
     }
   }
 
-  document.querySelector(".scratch-addons-theme[data-addon-id='editor-theme3']").textContent += stylesheet;
+  document.querySelector("[data-addon-id='editor-theme3']").textContent += stylesheet;
```
WorldLanguages, line 108: Bug: after reenable, color for custom blocks is not restored

WorldLanguages: This happens because the userscript modifies its own userstyle, and cs.js uses the text of the element to check if it's the same

content-scripts/cs.js
```diff
@@ -196,21 +220,46 @@ async function onInfoAvailable({ globalState, l10njson, addonsWithUserscripts, a
       setCssVariables(request.newGlobalState.addonSettings);
     } else if (request.fireEvent) {
       _page_.fireEvent(request.fireEvent);
+    } else if (request.dynamicAddonEnabled) {
+      const { scripts, userstyles, addonId, injectAsStyleElt, index } = request.dynamicAddonEnabled;
+      // TODO: simply adding the style won't cut it.
+      // Use the "index" variable to insert it or remove all and readd properly.
+      // But we also don't want to readd all... then "get-userscripts.js" will have to fetch a lot of data again.
```
WorldLanguages, line 227: Remove TODO

WorldLanguages:
Note, as of now:
When an addon is disabled/reenabled, its styles are first updated, then the userscript gets the respective event fired on `addon.self`. This is important for some addons like editor-stage-left, which expect to run JS with its CSS injected, and don't work properly otherwise.
However, when an addon uses `"updateUserstylesOnSettingsChange": true`, the userscript first gets the event fired on `addon.settings`, and a brief moment later, cs.js updates the userstyles accordingly. This shouldn't be that complex to fix, but it's not worth it because of the async nature of `getAddonData()` (we would need to wait until the `updateUserstylesSettingsChange` message is sent in order to send the `newGlobalState` and `fireEvent` messages to the content script). So it will stay like this, at least for now. As of now, only editor-dark-mode uses this feature and likely no other addon will

background/handle-settings-page.js
```diff
@@ -55,17 +44,19 @@ chrome.runtime.onMessage.addListener(function (request, sender, sendResponse) {
...
 
-    if (scratchAddons.manifests.find((obj) => obj.addonId === addonId).manifest.tags.includes("theme"))
-      scratchAddons.localEvents.dispatchEvent(new CustomEvent("themesUpdated"));
+    const manifest = scratchAddons.manifests.find((addon) => addon.addonId === addonId).manifest;
+    const { updateUserstylesOnSettingsChange } = manifest;
+    if (updateUserstylesOnSettingsChange)
+      scratchAddons.localEvents.dispatchEvent(
+        new CustomEvent("updateUserstylesSettingsChange", { detail: { addonId, manifest } })
+      );
+    // TODO: something idk
```
WorldLanguages, line 60: Remove todo

webpages/settings/index.js
```diff
@@ -543,22 +543,22 @@ const vue = (window.vue = new Vue({
...
                 });
-            // Addons/themes that were previously enabled on the tab (but not anymore)
+            // Addons that were previously enabled on the tab (but not anymore)
             // should go above enabled addons that are not currently running on the tab
             // so that it's easier to find them, even if the popup was closed.
             // Disabling then reenabling an addon is likely something common
             // so hopefully this saves some seconds of our users' lives :P
```
WorldLanguages, line 558: This logic is not working properly when an addon is dynamically disabled, it should stay at the top

WorldLanguages: This file will conflict with #1805

webpages/popup/index.js
WorldLanguages: Is this related to this PR...?

addons/60fps/userscript.js
```diff
@@ -16,30 +16,35 @@ export default async function ({ addon, global, console }) {
...
-        button.style.filter = mode ? "hue-rotate(90deg)" : "";
+        changeMode();
       }
     };
     button.addEventListener("click", (e) => flagListener(e));
     button.addEventListener("contextmenu", (e) => flagListener(e));
 
     const setFPS = (fps) => {
-      gloabal_fps = fps;
+      gloabal_fps = addon.self.disabled ? 30 : fps;
```
WorldLanguages, line 36: Now that you can, fix the misspelling

addons/editor-stepping/userscript.js
```diff
@@ -29,27 +29,30 @@ export default async function ({ addon, global, console }) {
...
-          if (!block) return;
-          while (b.childBlocks_.length) {
-            b = b.childBlocks_[b.childBlocks_.length - 1];
-            if (i === b.id) return true;
+    if (!addon.self.disabled) {
+      vm.runtime.threads.forEach((thread) => {
+        if (thread.target.blocks.forceNoGlow) return;
+        thread.stack.forEach((e) => {
+          let blockId = thread.target.blocks.getBlock(e)?.id;
+          if (!blockId) return;
```
RedGuy12, line 40: This fixes a bug where you can not run blocks in the block palate by clicking on them.

Issue #23, opened 2020-07-29
More options for userscript matches (not only URLs)


WorldLanguages:
Let's say a userscript only works on user profiles. The current match would currently be `https://scratch.mit.edu/users/*/`, why not just `userprofile`? And other potential filters for common use.

WorldLanguages:
I believe it's too late to implement this. It's not a big deal, either. Closing

Hans5958:
> too late

TheColaber:
I believe this one is worth reopening. Once we implement #390, addons that work on the editor/project page would also need to work on turbowarp's editor/project page. If we simply said for the match URL to be `project`, then we could have the match URL patterns for both turbowarp and scratch to link to htpps://scratch.mit.edu/projects/* and https://turbowarp.org/* (excluding the https://turbowarp.org/privacy.html page). And, if we end up supporting addons on another site, then it would be useful there too. This would also help with #646. And I think it's just easier if all the user needed to add in their addon, was just `search` or `user`

RedGuy12:
And phosphorous

WorldLanguages:
> I believe this one is worth reopening. Once we implement #390, addons that work on the editor/project page would also need to work on turbowarp's editor/project page. If we simply said for the match URL to be `project`, then we could have the match URL patterns for both turbowarp and scratch to link to htpps://scratch.mit.edu/projects/* and https://turbowarp.org/* (excluding the https://turbowarp.org/privacy.html page). And, if we end up supporting addons on another site, then it would be useful there too. This would also help with #646. And I think it's just easier if all the user needed to add in their addon, was just `search` or `user`

I honestly doubt userscript made for the Scratch editor could work "as-is" on TurboWarp, so it wouldn't make sense to merge both origins into one permission - an addon might only work on the Scratch editor

RedGuy12:
Well maybe userstyles

Paul Reid

________________________________

TheColaber:
> I honestly doubt userscript made for the Scratch editor could work "as-is" on TurboWarp, so it wouldn't make sense to merge both origins into one permission - an addon might only work on the Scratch editor

I'm sure that most do, and some (like poject-notes-tabs) could easily be fixed by adding an if statement.

TheColaber:
> I honestly doubt userscript made for the Scratch editor could work "as-is" on TurboWarp, so it wouldn't make sense to merge both origins into one permission - an addon might only work on the Scratch editor

[...]

Issue #645, opened 2020-11-04
Add the ability to customize all parts of the Scratch Website


aetinx:
**Is your feature request related to a problem? Please describe.**

Add an option to change what the actual Scratch website looks like, as in colours and fonts (add the ability to upload fonts from the user's PC).

**Describe the solution you'd like**

The option to change colours and fonts of any part of the website.

**Describe alternatives you've considered**

null

**Additional context**

null

mxmou:
The ability to customize colors will probably be added to Scratch 2.0 → 3.0 in a future update. Changing fonts could be a new addon.

PR #1977, opened 2021-03-26
More consistent error pages with Scratch 2.0 → 3.0

### LINKED ISSUE #1976: 404 not converted
**Describe the bug**
404 does not get converted to 3.0

404 pages sometimes do not get converted from 2.0 to 3.0
**To Reproduce**
go to scratch.mit.edu/errorfourohfour

go to scratch.mit.edu/errorfourohfour
1. See error

**Expected behavior**

 expect to see my 3.0 color changed taskbar at top

**Screenshots**

Screenshot (26)
Screenshot (27)

- OS: windows 10
- Browser edge
- Version i actually dont know

**Additional context**

sometimes shows color sometimes doesn't

mxmou:
**Resolves**

Resolves #1976

**Changes**

Makes 404s more consistent when Scratch 2.0 → 3.0 is enabled. Some still don't have the 3.0 navigation bar because that would require #23.

### DIFF
--- a/addons/scratchr2/addon.json
+++ b/addons/scratchr2/addon.json
@@ -54,25 +54,7 @@
     },
     {
       "url": "scratchr2_errors.css",
-      "matches": [
-        "https://scratch.mit.edu/discuss/*",
-        "https://scratch.mit.edu/mystuff/",
-        "https://scratch.mit.edu/users/*",
-        "https://scratch.mit.edu/classes/*",
-        "https://scratch.mit.edu/studios/*",
-        "https://scratch.mit.edu/projects/*/remixes/",
-        "https://scratch.mit.edu/projects/*/studios/",
-        "https://scratch.mit.edu/projects/*/remixtree/",
-        "https://scratch.mit.edu/accounts/*",
-        "https://scratch.mit.edu/statistics/",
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Well, this has the downside that "Scratch 2.0→3.0" will be listed as running, even on 3.0 pages, but not a big deal I guess? :P
Maybe we should add a way to exclude matches and list all the 3.0 pages on there. scratch-www is open source so maintaining a list of 3.0 pages shouldn't be too hard, I think?

RedGuy12:
> Well, this has the downside that "Scratch 2.0→3.0" will be listed as running, even on 3.0 pages

isn't it already with the customizable navbar/links settings?

mxmou:
> > Well, this has the downside that "Scratch 2.0→3.0" will be listed as running, even on 3.0 pages
> 
> isn't it already with the customizable navbar/links settings?

Yes, it already is. #23 would solve some similar problems such as website addons being listed as running in the editor, but this one will remain "running" on all pages until #645 is done.

mxmou:
There will be no other reason for Scratch 2.0 → 3.0 to run on 3.0-styled pages after #645 is finished so what @WorldLanguages mentioned is actually a problem. Closing.

PR #1980, opened 2021-03-27
Fix auto-hide block palette closing animation

### LINKED ISSUE #1876: Auto-hide block palette hide animation doesn't work
**Describe the bug**

When the auto-hide block palette addon is enabled, showing the block palette has an animation but hiding it doesn't.

**To Reproduce**

2. Go to the editor and let the palette hide
3. It hides instantly, with no animation

**Expected behavior**

It would be animated.

SheepTester:
**Resolves**

Fixes #1876

**Changes**

I've replaced the CSS animation with a transition. This removes the need to have to define an animation for each thing animated, and CSS transitions are also capable of stopping and going in a different direction midway through a transition if, say, the user wiggles their mouse around a lot.

**Reason for changes**

I've noticed that the hide animation didn't seem to work for me; the palette just disappears with the lock trailing behind.

**Tests**

This is the current animation on `master`:

Old behaviour demonstrates the hide animation not working

This is the animation introduced by the PR:

New behaviour demonstrates a smoother animation

I tried the various animation durations and toggle behaviours, and they seem to work the same as before.

### DIFF
--- a/addons/hide-flyout/style.css
+++ b/addons/hide-flyout/style.css
@@ -1,3 +1,9 @@
+.blocklyFlyout,
+.blocklyFlyoutScrollbar,
+.sa-lock-image {
+  transition-property: transform;
+}
+
 .blocklyFlyout.sa-flyoutClose {
   transform: translateX(-200px) !important;
 }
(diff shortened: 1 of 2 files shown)


TheColaber (approved):
Huh... this doesn't happen on my device since my computer is larger ig... This looks ok to me!

SheepTester:
> Huh... this doesn't happen on my device since my computer is larger ig... This looks ok to me!

Oh, very interesting. I zoomed out to 50% and the `master` branch version worked. Perhaps it might be because I'm on a retina display with a device pixel ratio of 2? I'm not sure how that would affect this though because typically CSS is DPR-independent

TheColaber:
@SheepTester You said what happened to mxmou also happened to you. This change also fixes that right? Just confirming before I merge.

SheepTester:
> @SheepTester You said what happened to mxmou also happened to you. This change also fixes that right? Just confirming before I merge.

Yep, this change was intended to fix that

PR #1986, opened 2021-03-28
Share Button to Project 

### LINKED ISSUE #1821: Share Button to Project
**Describe the new addon**

This addon would add a share button on project pages that would allow users to easily share projects on sites such as Twitter or Facebook via the built-in intent prompts. (Could be merged with the copy link button)

image

Genius398:
**Resolves**

Resolves #1821

**Changes**
I have added a button which when clicked takes the user to a WhatsApp Webpage with a pre-written message for sharing quickly.

Also, forgive me if there are some errors, cause this is the first time I'm contributing. It's also the first time I'm writing JSON (I stick to JS ☹)

### DIFF
--- a/addon.json
+++ b/addon.json
@@ -1,0 +1,15 @@
+{
+    "name": "Share Button",
+    "description": "Provides a button with when clicked allows you to choose from a few websites and share a pre-framed message of your Scratch Project. Will be updated soon, but it can currently only supprot whatsapp",
+    "userstyles": [
+      {
+        "url": "button.css",
+        "matches": ["https://scratch.mit.edu/projects/*"]
+      }
+    ]
+,
+      "persistent_scripts": ["background.js"],
+      "tags": ["community"],
+      "enabled_by_default": true
+
+}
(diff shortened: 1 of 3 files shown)


addon.json
```diff
@@ -0,0 +1,15 @@
...
+    "name": "Share Button",
+    "description": "Provides a button with when clicked allows you to choose from a few websites and share a pre-framed message of your Scratch Project. Will be updated soon, but it can currently only supprot whatsapp",
+    "userstyles": [
+      {
+        "url": "button.css",
+        "matches": ["https://scratch.mit.edu/projects/*"]
+      }
+    ]
+,
+      "persistent_scripts": ["background.js"],
```
mxmou, line 11: This needs to be a userscript because it interacts with the page.

mxmou (changes requested):

addon.json
Genius398: What do you mean? Please explain

RedGuy12 (changes requested):
go to https://github.com/Genius398/ScratchAddons/actions and enable it so your code can be automatically formatted.

GrahamSH-LLK:
This is useless. People are going to write their own messages anyway.

PR #2013, opened 2021-03-30
Drag and drop into lists and the ask and wait answer field

### LINKED ISSUE #1988: Drop text files onto lists, maybe into "ask and wait" prompt
Not exactly a *new* addon request, but building on the awesome work in #1916: how about an addon or option to allow for dropping text files into lists (and maybe the "ask and wait" prompt)? Lists are already pretty much the standard way to exchange data between the system (or some external resource) and a Scratch project, via the "import" and "export" options, or simply copying to clipboard from a save string shown as an item in the list. So, it'd be pretty cool to be able to drop a saved text file (such as exported from a list or downloaded from another program/site) right into a list watcher! Ditto for the "ask and wait" prompt, for single-line text files anyway.

This is based on my oooold feature request [here](https://github.com/LLK/scratch-gui/issues/268), BTW :P

(See ~~*spooky imminent issue reference*~~ #1990 for an infinitely stranger suggestion based off this.)

SheepTester:
**Resolves**

Resolves #1988 

**Changes**

I made it so that you can drag and drop text files into the list monitors. This is done by simulating a right click > Import, then inserting the dropped files into the file input that Scratch creates. This is rather janky, but Scratch uses a CSV parsing library that I don't want to re-include.

Dragging into lists and the answer box

I also made it so that you can drag files into the ask and wait answer field. It inserts the files as text and turns the newlines into spaces, like when you paste text with multiple lines into that input. It doesn't automatically submit the text so the user can modify it as they please.

Dragging into the answer box (updated)

I also fixed the import dialog spinning indefinitely if you drag highlighted text into the sprites pane, etc.

**Tests**

List monitor drag/drop testing procedure:

1. Dragging a text file into the list monitor in the editor should work
2. Also if a right click menu is open on that monitor
3. It should import to the correct list monitor when a right click menu is open on a different monitor
4. In fullscreen mode in the editor, dropping a file should not show the blue indicator and, when dropped, should behave as if the addon was disabled (opens the file in a new tab)
5. Same in player mode

(This PR passes those steps with flying colours)

I dragged and dropped text files into the ask and wait box. Dragging highlighted text inside the input around still behaves as normal. The `answer` block contains the inserted text upon submission.

I also tested this on Firefox!


RedGuy12 (commented):
I assume it works with any text file, you just used html/py/js as examples?

SheepTester:
> I assume it works with any text file, you just used html/py/js as examples?

It works with any text file; it works the same as if you right click > Import, make the file selector select all files, then select a non-txt/csv/tsv file

addons/drag-drop/userscript.js
```diff
@@ -43,4 +56,80 @@ export default async function ({ addon, global, console }) {
...
+      // Get the monitor's context menu ID from the list name
+      // https://github.com/LLK/scratch-gui/blob/develop/src/components/monitor/monitor.jsx#L37
+      const contextMenuId = `monitor-${listMonitor.querySelector('div[class*="monitor_list-header"]').textContent}`;
+      droppable(listMonitor, async (files) => {
+        // Force react-contextmenu's context menu to open using their global
+        // events:
+        // https://github.com/vkbansal/react-contextmenu/blob/v2.9.4/src/ContextMenuTrigger.js#L114-L119
+        // https://github.com/vkbansal/react-contextmenu/blob/v2.9.4/src/actions.js#L27-L29
+        window.dispatchEvent(
+          new CustomEvent("REACT_CONTEXTMENU_SHOW", {
```
towerofnix, line 72: This looks like a fine way to show the menu, since it matches react-contextmenu's public `actions.js` code, like you linked. I'm curious if simulating a plain right-click event would also work, though? If so, you'd be effectively using the exact same interface as the user themselves do, making the code more reliable in the long-term, avoiding breakage if e.g. the specifics of this CustomEvent (i.e. its event name or expected `detail` format) change.

addons/drag-drop/userscript.js
```diff
@@ -43,4 +56,80 @@ export default async function ({ addon, global, console }) {
...
+          })
+        );
+        const contextMenu = await addon.tab.waitForElement("body > .react-contextmenu.react-contextmenu--visible");
+        // Override DOM methods to import the text file directly
+        // See: https://github.com/LLK/scratch-gui/blob/develop/src/lib/import-csv.js#L21-L22
+        const appendChild = document.body.appendChild;
+        document.body.appendChild = (fileInput) => {
+          // Restore appendChild to <body>
+          document.body.appendChild = appendChild;
+          document.body.appendChild(fileInput);
```
towerofnix, line 86: Can you guarantee the next call of `appendChild` will be with the file input? Or would it be safer to only reset `appendChild` once the passed element is definitely the file input?

(Since they're a bit less intrusive—no global function overriding—I was going to suggest using a mutation observer, or possibly `waitForElement` if that's based on observers, but I'm guessing it's important here that your overrides e.g. `fileInput.click = () => {}` be set immediately.)

addons/drag-drop/userscript.js
```diff
@@ -43,4 +56,80 @@ export default async function ({ addon, global, console }) {
...
+        // https://github.com/vkbansal/react-contextmenu/blob/v2.9.4/src/actions.js#L27-L29
+        window.dispatchEvent(
+          new CustomEvent("REACT_CONTEXTMENU_SHOW", {
+            detail: {
+              id: contextMenuId,
+              position: { x: 0, y: 0 },
+            },
+          })
+        );
+        const contextMenu = await addon.tab.waitForElement("body > .react-contextmenu.react-contextmenu--visible");
```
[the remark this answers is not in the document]
SheepTester: Ah yes, I probably should move the `display: none;` a bit earlier for clarity (it's currently set once the file input is added, which I believe is done synchronously, so moving it shouldn't change anything)

SheepTester:
Ah, I allowed it to import files while in player mode to facilitate importing text files, but since it can be used for cheating in multiplayer games (and exporting still only works in the editor), I shall change that.

addons/drag-drop/userscript.js
SheepTester: I believe because [`importCSV`](https://github.com/LLK/scratch-gui/blob/develop/src/containers/monitor.jsx#L166) is called immediately (and synchronously) [on click](https://github.com/LLK/scratch-gui/blob/develop/src/components/monitor/monitor.jsx#L96), it should be guaranteed, but I can add a check just in case

SheepTester:
Hmm, I think the code would be a lot simpler if it just imported the text directly into the list via the VM trap rather than right click > Import. Getting the list from the VM should be fairly easy because it has just occurred to me that I can just use the list name from the monitor.

The only thing we would lose is CSV support, but to be honest, Scratch's CSV support is kind of janky (it's just an alert box asking "which column do you want?"), exporting from a list doesn't produce CSVs anyways, and since Scratch detects CSVs regardless of the file type, their CSV parser would interfere with ordinary files containing commas

PR #2016, opened 2021-03-30
Fix profile statistics graph issues

### LINKED ISSUE #2005: Profile statistics issues
* Points on the graph are 8 days apart instead of 7
* The last point is always the current date, which means the one before is actually newer. The previous one should either be removed or the list should be sorted before it's displayed.

mxmou:
**Resolves**

Resolves #2005

**Changes**

* Changes `segment` to 6 (for some reason 7 is actually 8 days) and `range` to 364 (so that it's a multiple of 7) in the call to the graph API
* Removes the last point from the graph so that there aren't two points for the same day

@WorldLanguages These changes should probably also be made on scratchstats.com.

### DIFF
--- a/addons/scratchstats/userscript.js
+++ b/addons/scratchstats/userscript.js
@@ -74,9 +74,10 @@ export default async function ({ addon, msg, console }) {
           msg("most-views")
         )
       );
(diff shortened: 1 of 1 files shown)


WorldLanguages:
@mxmou I'll revert part of this bc this will be fixed in ScratchDB's side

PR #2037, opened 2021-03-31
Add dark mode to statics graphs

TheColaber:
Try on https://scratch.mit.edu/statistics/

image

### DIFF
--- a/addons/dark-www/experimental_statistics.css
+++ b/addons/dark-www/experimental_statistics.css
@@ -1,17 +1,31 @@
-.box .box-content {
-  background: #282828;
+.stats li.data {
+  background-image: url("%addon-self-dir%/assets/meow.png");
+  background-size: 20px 15px;
 }
-text {
+.nvd3 text,
+#country_chart text {
   fill: white;
 }
-.nvtooltip {
-  background: #000;
-  color: #fff;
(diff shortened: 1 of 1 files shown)


addons/dark-www/experimental_statistics.css
```diff
@@ -1,5 +1,16 @@
 .box .box-content {
-  /* Charts don't look well on dark background */
-  background: #fff;
+  background: #282828;
   color: #575e75;
```
mxmou, line 3: You can remove this line and line 6.

mxmou (changes requested):
Could you also change the color of the grid lines? Maybe white for the black one and `#333` for the gray ones.

addons/dark-www/experimental_statistics.css
```diff
@@ -1,5 +1,16 @@
...
 }
+text {
+  color: white;
+  fill: currentColor;
+}
+.nvtooltip {
+  background: #000;
```
mxmou, line 10: Use `#282828`

TheColaber:
I think it's a bit hard to see at `#000000`
image
I think `#606060` is better maybe
image

mxmou (approved):

PR #2066, opened 2021-04-03
Remove ../

GrahamSH-LLK:
We use a lot of ../s to get to directories. These can be removed with / in most cases.


WorldLanguages:
Reasoning: code editors can't always guess what's the root of the project

PR #2083, opened 2021-04-05
S3blocks

GrahamSH-LLK:
Adds scratch 3.0 blocks to forums.
Todo:
- [ ] Figure out why menus have a few undefineds
- [ ] Use regex to find scratch assets instead of string.

Git is magic, and that's how I have commits from @RedGuy12 on my fork


apple502j:
Please do not ship another copy of scratchblocks.

PR #2084, opened 2021-04-05
scratchblocks-v3 using WIP and PLSEXPLAIN

### LINKED ISSUE #438: Forum Scratchblocks Injection
This would be an extension that injected the new scratchblocks into the forums.

apple502j:
Resolves #438


RedGuy12:
Can your rename the translation file to clarify it's the translations for scratchblocks?

libraries/translations-all-v3.5.js
```diff
@@ -0,0 +1,9 @@
+/**
```
WorldLanguages, line 1: Should this be added to the list of libraries?

apple502j: Prob not

WorldLanguages: Why not?

apple502j: It's part of scratchblocks, just 2 files

WorldLanguages:
I can't get asap.js to run consistently before vanilla 2.0 scratchblocks is ran, we need to find a better way to do this

GrahamSH-LLK:
> I can't get asap.js to run consistently before vanilla 2.0 scratchblocks is ran, we need to find a better way to do this

Same, I could consistently reproduce it not working at all.

PR #2091, opened 2021-04-06
Devtools cleanup update

### LINKED ISSUE #1943: Developer tools renames the clean up blocks option confusingly
**Describe the bug**

When the `editor-devtools` addon is enabled, the *Clean up Blocks* option in the workspace context menu is renamed to *Clean Up Blocks (+)*. Apparently the (+) is supposed to mean that it's better than Scratch's option, but at first I thought it means that + is a keyboard shortcut for it.

**To Reproduce**

1. Enable developer tools
2. Open the editor
3. Right click the workspace
4. The clean up option is named *Clean Up Blocks (+)*, although pressing + does not clean up blocks

**Expected behavior**

One of the following:
* Don't rename the option
* Name it something like *Clean up Blocks (improved)* so that it isn't confusing.

Either way it shouldn't change capitalization from that used by Scratch.

TheColaber:
Resolves #1943
Related #2072

--

Adds a setting to remove the old cleanup option and renames the item to `Clean Up +`. Also enables the addon by default because it no longer removes the element unless the user chooses to.


mxmou:
> Adds a setting to remove the old cleanup option and renames the item to `Clean Up +`.

This means that there are two clean up options by default, which is confusing. I think it should be a select type setting called "Preferred clean up style" with options "Vertical" (default), "2D" (for users who already have the addon enabled) and "Both" (which shows both options).

wgyt (approved):
Nice pull request, I like how this is now enabled by default and we can disable deleting the vanilla clean up.

addons-l10n/en/editor-devtools.json
```diff
@@ -41,7 +41,7 @@
   "editor-devtools/replace": "Griffpatch: Switch all {name} in this sprite for the variable named:",
   "editor-devtools/top": "send to top",
   "editor-devtools/bottom": "send to bottom",
-  "editor-devtools/clean-plus": "Clean Up Blocks (+)",
+  "editor-devtools/clean-plus": "Clean Up Blocks +",
```
WorldLanguages, line 44: It should be `Clean up Blocks +`, Scratch doesn't use uppercase for up

TheColaber:
See #2146

PR #2117, opened 2021-04-09
New Addon: Wiki Button

### LINKED ISSUE #721: Wiki Button

It's hard to acsess the forums without a navbar button. Same for the Wiki.

I propose a addon that adds a Wiki button to the navbar, just like Maximouse and World_Language's Discuss button.

8bitjake:
**Resolves**

Resolves #721.

**Changes**

Added a new addon, the Wiki Button. It works the same way as the Discuss Button, but it links to wiki.scratch.mit.edu and replaces the About button.

**Reason for changes**

Again, read #721.

**Tests**

It's been tested. I tested it by removing the original extension and enabling the custom addon.

**Other**

I'm also working on a Custom Header extension. It'll combine this addon and the Discuss Button addon, along with an option 
for custom URLs. That addon is probably gonna be more likely to be merged than this one, to be honest.


Hans5958:
Some say that the discuss button will be replaced with a completely customisable navbar.

8bitjake:
> Some say that the discuss button will be replaced with a completely customisable navbar.

I'm already working on that, I'm suggesting this in the meantime.

WorldLanguages:
> I'm also working on a Custom Header extension

I think that's one of those addons we never had consensus on how it should be properly done, don't expect it to get merged easily

PR #2118, opened 2021-04-09
Disable scrolling

8bitjake:
**Resolves**
Resolves an issue where the live featured project embed scrolls when it's not supposed to.

**Changes**
I added a single line of code. It adds the "scrolling, no" attributes to the iframe.

**Reason for changes**
Scrolling is not supposed to happen, according to TheColaber.
image

**Tests**
Yes, I tested it by simply reloading the page when I make a new change.


GrahamSH-LLK:
I'm not sure this is the answer. We might need to figure out whether the frame is too small?

PR #2120, opened 2021-04-10
Lint 20210410

apple502j:


popups/cloud-games/popup.js
```diff
@@ -18,7 +18,7 @@ import WebsiteLocalizationProvider from "../../libraries/website-l10n.js";
 
   await l10n.loadByAddonId("cloud-games");
 
-  const vue = new Vue({
+  new Vue({
```
WorldLanguages, line 21: Can we change this to `window.vue = new Vue({`?

apple502j: Why do we need debugging code?

WorldLanguages: we don't, but addons already own the global object on a popup, so I don't see any negative side to exposing vue as a global

PR #2131, opened 2021-04-11
Costume Editor Shape Opacity Addon

### LINKED ISSUE #2129: Alpha (Opacity) slider in costume editor.
Currently, there are only color, brightness and saturation sliders in the color picker. I suggest adding a new one for alpha (or opacity). In vanilla scratch the only way to get semi-opaque pixels is using bucket fill option and selecting a gradient between no color and a fully opaque color. This creates major headaches if you want a large section of the same opacity as you have to copy the pixel that has the opacity you want to create a large section of those pixels.

An example of opacity in the costum editor created with the bucket fill gradient method.
image

TheColaber:
**Resolves**

Resolves #2129

**Changes**

Adds an input to the costume editor when a shape is selected. Also works with multiple selected shapes. Does not work with bitmap.

**Tests**

Works.

### DIFF
--- a/addons/shape-opacity/addon.json
+++ b/addons/shape-opacity/addon.json
@@ -1,0 +1,18 @@
+{
+  "name": "Costume Editor Shape Opacity",
+  "description": "Select a shape in the costume editor and a new input will apear to change the shape's opacity.",
+  "credits": [
+    {
+      "name": "TheColaber",
+      "link": "https://scratch.mit.edu/users/TheColaber"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["editor"],
+  "l10n": true
+}
--- a/addons/shape-opacity/userscript.js
+++ b/addons/shape-opacity/userscript.js
@@ -1,0 +1,61 @@
+export default async function ({ addon, global, console, msg }) {
+  let appended = false;
+  const clss = addon.tab.scratchClass;
+
+  const labelContainer = document.createElement("label");
+  labelContainer.className = clss("label_input-group");
+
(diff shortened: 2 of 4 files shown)


addons/shape-opacity/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Costume Editor Shape Opacity",
```
mxmou, line 2: Other addon names use sentence case.

addons/shape-opacity/userscript.js
```diff
@@ -0,0 +1,61 @@
...
+      if (action.selectedItems.length === 0) {
+        labelContainer.style.display = "none";
+      } else {
+        // Scratch returns undefined for bitmapMode when in bitmap mode...
+        if (action.bitmapMode !== false) return;
+        labelContainer.style.display = "";
+        if (!appended) {
+          appended = true;
+          const appendRow = document.querySelector("[class*=paint-editor_mod-mode-tools]");
+          appendRow.append(labelContainer);
```
mxmou, line 32: Could the input be next to Fill and Outline? That's where it makes the most sense. For consistency it should be present but disabled when no object is selected.

mxmou (changes requested):

WorldLanguages:
TODO: test whether this works with every addon interacting with color pickers (there are many)

TheColaber:
> TODO: test whether this works with every addon interacting with color pickers (there are many)

Shouldn't interfere with anything bc it's a separate variable...

TheColaber:
closing for now. better implmentation later

WorldLanguages:
@TheColaber What happened?

RedGuy12:
He gave up

TheColaber:
I dont think I gave up, but that some internal scratch stuff needs to change so it doesn't remove the opacity once you change to a different color for example.

PR #2116, opened 2021-04-09
New addon: initialise sprite position

pufferfish101007:
**This needs #2142 to be merged for this to work**

**Resolves**

Resolves #1759 

**Changes**

Adds a new addon that sends sprites to a specified position when they are added.

**Reason for changes**

People (including me) want it.

**Tests**

Works on chrome for all methods of adding sprites afaik

~~**Todo:**~~
- [ ] ~~differentiate between sprites selected from the library and those selected using the surprise sprite thing~~

PR #2142, opened 2021-04-13
Automatically typecast number settings to numbers

pufferfish101007:
**Resolves**

Resolves https://github.com/ScratchAddons/ScratchAddons/pull/2116#discussion_r611223125

**Changes**

Adds the number attr to number settings inputs

**Reason for changes**

The value will now be automatically typecasted to a number so that addon.settings.get eeturns the correct type.

**Tests**

Tested on #2116 without the typecasting on the actual addon. Also logged typeof addon.settings.get blah blah blah andit was all correct.


WorldLanguages:
I'm confused by this PR, and is `number` even a valid HTML attribute for inputs?

WorldLanguages:
AFAIK we already store numbers for number settings?

TheColaber:
> AFAIK we already store numbers for number settings?

Nope.
image
image

pufferfish101007:
> I'm confused by this PR, and is `number` even a valid HTML attribute for inputs?

It's a vue thing, see https://v1.vuejs.org/guide/forms.html#number



> AFAIK we already store numbers for number settings?

For defaults, yes, but when you change it it just v-models it into the settings object which is sent to the background; v-model only uses numbers if youuse the number attr in vue 1. You can try it out by console.logging `typeof addon.settings.get("xyz")` - for me it always is a string without this pr.

RedGuy12:
number is not just a vue thing, it is part of the html spec.

TheColaber:
> number is not just a vue thing, it is part of the html spec.

I don't think that's true...

pufferfish101007:
> > number is not just a vue thing, it is part of the html spec.
> 
> I don't think that's true...

Yeah I can't find it anywhere on MDN, and it's in the vue 1 docs.

PR #2144, opened 2021-04-13
Bugfix - Event should have no default background color

griffpatch:
**Resolves**

Naughty background color on events in Find menu - should only have a background color when hovered over


TheColaber:
#2146 seems to have this bug fix. Should this be closed?

TheColaber:
Closing because #2146 seems to have the same commit in the PR.

griffpatch:
Yep sorry - should have closed that one - Still learning about git lol

PR #2146, opened 2021-04-13
Enable / Disable features in settings of Dev Tools

griffpatch:
**Resolves**

Bugfix - Event should have no default background color
Enhancement - Enable / Disable features in settings


addons-l10n/en/editor-devtools.json
```diff
@@ -41,7 +41,7 @@
   "editor-devtools/replace": "Griffpatch: Switch all {name} in this sprite for the variable named:",
   "editor-devtools/top": "send to top",
   "editor-devtools/bottom": "send to bottom",
-  "editor-devtools/clean-plus": "Clean Up Blocks (+)",
+  "editor-devtools/clean-plus": "Clean up blocks +",
```
TheColaber, line 44: Make the "b" in "blocks" capital to stay inline with scratch 
image

addons/editor-devtools/DevTools.js
```diff
@@ -1622,6 +1631,10 @@ export default class DevTools {
       return;
     }
 
+    if (typeof this.addon.settings !== undefined && !this.addon.settings.get("enableBlockInjector")) {
```
TheColaber, line 1634: "enableBlockInjector" doesn't seem to be a setting.

RedGuy12:
Yay!!
From: griffpatch ***@***.***>
Sent: Tuesday, April 13, 2021 2:32:17 PM
To: ScratchAddons/ScratchAddons ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [ScratchAddons/ScratchAddons] Enable / Disable features in settings of Dev Tools (#2146)


Resolves

Bugfix - Event should have no default background color
Enhancement - Enable / Disable features in settings

You can view, comment on, or merge this pull request online at:

  https://github.com/ScratchAddons/ScratchAddons/pull/2146

Commit Summary

  *   Bugfix - Event should have no default background color
  *   Enhancement - Enable / Disable features in settings

File Changes

  *   M addons-l10n/en/editor-devtools.json<https://github.com/ScratchAddons/ScratchAddons/pull/2146/files#diff-1748b451da5a9e4fe32e993e335b7c983ecf6283b88039121f00363c8f012032> (4)
  *   M addons/editor-devtools/DevTools.js<https://github.com/ScratchAddons/ScratchAddons/pull/2146/files#diff-43d928c2d28cbb183b4bd0e3404ded341647def6274253ba7e5aa36036596ad2> (97)
  *   M addons/editor-devtools/addon.json<https://github.com/ScratchAddons/ScratchAddons/pull/2146/files#diff-8041d88420f8884c0a0060c016c2070a8b46ee0441c40488f49e962b770dbb00> (41)
  *   M addons/editor-devtools/userscript.css<https://github.com/ScratchAddons/ScratchAddons/pull/2146/files#diff-424c5cc3eed5f8e9e56c8aff67d0e3eca110e29d2cd006b339574c4d2389cad5> (1)

Patch Links:

  *   https://github.com/ScratchAddons/ScratchAddons/pull/2146.patch
  *   https://github.com/ScratchAddons/ScratchAddons/pull/2146.diff

addons/editor-devtools/DevTools.js
griffpatch: Indeed - These flags don't exist yet - WL said to hang back on these so they will default to enabled for the time-being until key bindings are configurable...

addons-l10n/en/editor-devtools.json
griffpatch: Addressed thanks

addons/editor-devtools/DevTools.js
TheColaber: @griffpatch This check you are doing here always returns true because undefined needs to be a string. Try `"undefined" !== undefined` in the console.
image

RedGuy12: but it shouldn't ever be undefined? or am I missing something

TheColaber: Yes, it is undefined when running the addon through the extension. Remember, the extension and the addon both use the same code.

RedGuy12: ah, so on the extension, all features are always enabled?

griffpatch: > Yes, it is undefined when running the addon through the extension. Remember, the extension and the addon both use the same code.

Hmm, I just stripped these one - Are they required after all (only with quotes). I was assuming that perhaps they weren't needed anymore.

TheColaber: > Hmm, I just stripped these one - Are they required after all (only with quotes). I was assuming that perhaps they weren't needed anymore.

I would assume that they would be needed because the devtools extension does not support setting, but unless we really wanted to add them, they could be in editor settings, or maybe settings in the browser popup. In the extension, it should be erroring out saying "Cannot read property 'get' of undefined".

TheColaber: W_L reminded me that there is an alternative, by checking `addon.self._isDevtoolsExtension` [which is true for the devtools extension](https://github.com/ScratchAddons/DevtoolsExtension/blob/master/inject/run-addon.js#L47).

griffpatch:
Yes let's do that

PR #2148, opened 2021-04-13
Devtools conflict with theme3

### LINKED ISSUE #1968: Using find functionality can leave found occurrences highlighted

TheColaber:
**Resolves**

Resolves #1968

**Changes**

Instead of storing the color that devtools thinks it should be, it just sets its fill to "" when it wants to clear the highlight.

### DIFF
--- a/addons/editor-devtools/blockly/BlockFlasher.js
+++ b/addons/editor-devtools/blockly/BlockFlasher.js
@@ -9,12 +9,11 @@ export default class BlockFlasher {
   static flash(block) {
     if (myFlash.timerID > 0) {
       clearTimeout(myFlash.timerID);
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Was fixed by #2146

PR #2149, opened 2021-04-13
New addon: gamepad/controller support

GarboMuffin:
Adds an addon that lets you interact with Scratch projects using a controller. I don't know if anyone has asked for this in SA, but some people have for TW and I've been writing it as an addon, so I'm opening this here if you want it.

It works by binding gamepad buttons to keyboard keys or mouse down, similar to programs like joytokey. It must be manually configured by each user for each game, which is far from ideal.

It adds a button:

chrome_r2mO2dxPSh

Which opens an ugly configuration modal:

image


addons/gamepad/addon.json
```diff
@@ -22,7 +22,7 @@
       "matches": ["https://scratch.mit.edu/projects/*"]
     }
   ],
-  "tags": [],
+  "tags": ["beta"],
```
WorldLanguages, line 25: It's definitely an editor addon?

GarboMuffin: I'm not sure. This addon is essential JoyToKey as an addon. You could use it in the editor, but I suspect more people will use it outside of the editor

jeffalo:
could there be a way to write syntax in a project description which could prompt the addon to use that control scheme? that would make it useful for project creators to decide the controls.

mxmou:
> could there be a way to write syntax in a project description which could prompt the addon to use that control scheme? that would make it useful for project creators to decide the controls.

Could it be in a list inside the project (called something like "sa-gamepad")? Otherwise it would be confusing for non-SA users.

addons/gamepad/addon.json
RedGuy12: 60fps, mute project, and etc are often used outside the editor, yet they have `editor` tags.
Plus, it needs either `community` or `editor` because of how the settings page is made, and `editor` is definitely more applicable here than `community`.

addons/gamepad/gamepadlib.js
```diff
@@ -0,0 +1,835 @@
+let console = window.console;
```
RedGuy12, line 1: should this be moved to the `libraries` folder?

GarboMuffin: no

addons/gamepad/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Gamepad support",
-  "description": "Interact with projects using a controller/gamepad.",
+  "description": "Interact with projects using a USB or Bluetooth controller/gamepad.",
```
WorldLanguages, line 3: Chrome only addon?

GarboMuffin: The Gamepad API and Web Bluetooth/USB are different things. All major browsers support the gamepad API https://caniuse.com/?search=gamepad

PR #2156, opened 2021-04-14
Make block palette category icons not draggable

GarboMuffin:
Block palette category icons should not be draggable.


GarboMuffin:
#2159 is probably the better solution

PR #2159, opened 2021-04-14
migrate block-pallate-icons to a userstyle

RedGuy12:
this is a better way of doing it. if you don't need a userscript, don't use a userscript (imo)

(the only reason I changed the icon files was so i could apply some styles in the svg files)


addons/block-palette-icons/userstyle.css
```diff
@@ -0,0 +1,47 @@
...
+.scratchCategoryId-motion .scratchCategoryItemBubble {
+  background-image: url(chrome-extension://npijjahcnpemcijdgoioaclneonckman/addons/block-palette-icons/icons/motion_icon.svg);
+}
+
+.scratchCategoryId-looks .scratchCategoryItemBubble {
+  background-image: url(chrome-extension://npijjahcnpemcijdgoioaclneonckman/addons/block-palette-icons/icons/looks_icon.svg);
+}
+
+.scratchCategoryId-sound .scratchCategoryItemBubble {
+  background-image: url(chrome-extension://npijjahcnpemcijdgoioaclneonckman/addons/block-palette-icons/icons/sound_icon.svg);
```
GarboMuffin, line 18: url() is relative. please use that, this probably won't work on anyone else's computer.
```suggestion
  background-image: url(sound_icon.svg);
```

RedGuy12: oh yeah i was going to but forgot

p.s. actually it's a theme so it's injected into the page. but I can use `%addon-self-dir%`

addons/block-palette-icons/addon.json
```diff
@@ -1,14 +1,17 @@
 {
   "name": "Block palette category icons",
-  "description": "Replaces the colored circles that identify block categories with custom icons.",
-  "tags": ["editor"],
+  "description": "Adds icons inside the colored circles that normally identify block categories.",
```
WorldLanguages, line 3: Explain why the description change?

RedGuy12: just for clarity. the addon doesn't remove the circles, they're still there. it just adds icons inside them.

WorldLanguages: Ok, but can you remove `normally`? Not sure why you added that keyword

Issue #1934, opened 2021-03-22
Settings presets


TheColaber:
**Is your feature request related to a problem? Please describe.**

A lot of ppl like the fact that with scratch addons, you can make the editor look like 2.0 and sorta 1.4 as well. It would be nice if the user could just click a single button to get that addon presets.

**Describe the solution you'd like**

Buttons to load SA preset addons.

**Describe alternatives you've considered**

Do it's manually? Meh.

**Additional context**

This would be easy if we just load premade settings like this for example:
<details>
<summary>2.0 settings</summary>
<br>
<pre>
{
  "core": { "version": "1.12.0-prerelease" },
  "addons": {
    "custom-block-shape": {
      "enabled": true,
      "settings": { "cornerSize": 150, "notchSize": 75, "paddingSize": 70 }
    },
    "editor-dark-mode": {
      "enabled": false,
      "settings": { "selectedMode": "3-darker", "textShadow": false }
    },
    "editor-stage-left": { "enabled": true, "settings": {} },
    "editor-theme3": {
      "enabled": true,
      "settings": {
        "Pen-color": "#0e9a6c",
        "control-color": "#e1a91a",
        "custom-color": "#632d99",
        "data-color": "#ee7d16",
        "data-lists-color": "#cc5b22",
        "events-color": "#c88330",
        "looks-color": "#8a55d7",
        "motion-color": "#4a6cd4",
        "operators-color": "#5cb712",
        "sensing-color": "#2ca5e2",
        "sounds-color": "#bb42c3"
      }
    },
    "animated-thumb": {
      "enabled": false,
      "settings": { "persistentThumb": false }
    },
    "block-palette-icons": { "enabled": false, "settings": {} },
    "clones": { "enabled": false, "settings": {} },
    "editor-colored-context-menus": { "enabled": false, "settings": {} },
    "editor-messages": { "enabled": false, "settings": {} },
    "hide-flyout": {
      "enabled": false,
      "settings": { "speed": "default", "toggle": "category" }
    },
    "mediarecorder": { "enabled": false, "settings": {} },
    "pause": { "enabled": false, "settings": {} },
    "variable-manager": { "enabled": false, "settings": {} }
  }
}
</pre>
</details>

![image](https://user-images.githubusercontent.com/72760579/111924687-1d795200-8a7c-11eb-9934-486159f14901.png)
vs
[...]

PR #2161, opened 2021-04-14
Initial debugger

### LINKED ISSUE #693: Code Debugger
**Describe the new addon**

Basically debugs your code.
Features in mind:

- Option in right-click to debug a stack of blocks
  - Shows variables and parameters (for custom blocks) values
  - Shows what part of that code is currently running (has been done already)
- Adds a Console
  - Reports performance issues and tries to find areas that cause these bugs.
  - If it doesn't find any bugs, create checks on might be not working
  - sa-log(stuff) - to log data in the console

**Link to the new addon**
null

Oh yeah, and trust me, this addon would have major value.

GrahamSH-LLK:
**Resolves**

Resolves #693

**Changes**

Adds an intial console/debugger addon
**Reason for changes**

Debuggers cool
**Tests**

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -85,5 +85,6 @@
   "dango-rain",
   "custom-zoom",
   "wrap-lists",
-  "blocks2image"
+  "blocks2image",
+  "console"
 ]
--- a/addons/console/addon.json
+++ b/addons/console/addon.json
@@ -1,0 +1,31 @@
+{
+  "name": "Console",
+  "description": "Adds blocks to log data to a Scratch console.",
+  "info": [
+    {
+      "type": "notice",
+      "text": "Protip: remove logs once you're done debugging",
+      "id": "logprotip"
+    }
+  ],
+  "credits": [
+    {
+      "name": "GrahamSH"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["editor", "recommended"],
+  "enabledByDefault": false,
+  "l10n": true
+}
--- a/addons/console/debug.svg
+++ b/addons/console/debug.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" class="h-6 w-6" viewBox="0 0 24 24" fill="#ffae00" stroke="#d89400">
+  <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 9l3 3-3 3m5 0h3M5 20h14a2 2 0 002-2V6a2 2 0 00-2-2H5a2 2 0 00-2 2v12a2 2 0 002 2z"/>
+</svg>
--- a/addons/console/style.css
+++ b/addons/console/style.css
@@ -1,72 +1,72 @@
 .debug-btn {
-    width: 2rem;
-    height: 2rem;
-    padding: 0.375rem;
-    border-radius: 0.25rem;
-    -webkit-user-select: none;
-    -moz-user-select: none;
-    -ms-user-select: none;
-    user-select: none;
-    user-drag: none;
-    cursor: pointer;
-  }
-  
-  .debug-btn:hover {
-    background-color: hsla(215, 100%, 65%, 0.15);
-  }
-  .debug {
-    display: none;
-  }
-  .debug {
-    position: absolute;
-    z-index: 1000;
-    background-color: #f1f1f1;
-    border: 1px solid #d3d3d3;
-    min-height: 25rem;
-    min-width: 25rem;
-    max-height: 25rem;
-  }
-  
-  .debug h1 {
-    padding: 10px;
-    cursor: move;
-    z-index: 10;
-    color: #fff;
-  }
-  .log {
-      --message-border-color: rgb(240 240 240);
-      --warning-border-color: hsl(50deg 100% 88%);
-      --error-border-color: hsl(0deg 100% 92%);
-      --error-text-color: red;
-      --monospace-font-size: 12px;
-      --monospace-font-family: consolas, lucida console, courier new, monospace;
-      --source-code-font-size: 12px;
-      --source-code-font-family: consolas, lucida console, courier new, monospace;
-      font-family: var(--monospace-font-family);
-      font-size: var(--monospace-font-size) !important;
-      word-wrap: break-word;
-      user-select: text;
-      box-sizing: border-box;
-      min-width: 0;
-      min-height: 0;
-      display: flex;
-      border-top: 1px solid var(--message-border-color);
-      border-bottom: 1px solid transparent;
-      background: white;
-      word-wrap: break-word;
-      white-space: pre-wrap;
-      flex: auto;
-      line-height: 1.2;
-      padding-left: 10px;
-      border-top-right-radius: revert;
-      border-top-left-radius: revert;
-      padding: revert;
-      cursor: text;
-  }
-  .logs {
-      overflow-y: scroll;
-      height: 25rem;
-      cursor: default;
-      border-radius: 0;
-      padding: 0;
-  }
+  width: 2rem;
+  height: 2rem;
+  padding: 0.375rem;
+  border-radius: 0.25rem;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+  user-drag: none;
+  cursor: pointer;
+}
+
+.debug-btn:hover {
+  background-color: hsla(215, 100%, 65%, 0.15);
+}
+.debug {
+  display: none;
+}
+.debug {
+  position: absolute;
+  z-index: 1000;
+  background-color: #f1f1f1;
+  border: 1px solid #d3d3d3;
+  min-height: 25rem;
+  min-width: 25rem;
+  max-height: 25rem;
+}
+
+.debug h1 {
+  padding: 10px;
+  cursor: move;
+  z-index: 10;
+  color: #fff;
+}
+.log {
+  --message-border-color: rgb(240 240 240);
+  --warning-border-color: hsl(50deg 100% 88%);
+  --error-border-color: hsl(0deg 100% 92%);
+  --error-text-color: red;
+  --monospace-font-size: 12px;
+  --monospace-font-family: consolas, lucida console, courier new, monospace;
+  --source-code-font-size: 12px;
+  --source-code-font-family: consolas, lucida console, courier new, monospace;
+  font-family: var(--monospace-font-family);
+  font-size: var(--monospace-font-size) !important;
+  word-wrap: break-word;
+  user-select: text;
+  box-sizing: border-box;
+  min-width: 0;
+  min-height: 0;
+  display: flex;
+  border-top: 1px solid var(--message-border-color);
+  border-bottom: 1px solid transparent;
+  background: white;
+  word-wrap: break-word;
+  white-space: pre-wrap;
+  flex: auto;
+  line-height: 1.2;
+  padding-left: 10px;
+  border-top-right-radius: revert;
+  border-top-left-radius: revert;
+  padding: revert;
+  cursor: text;
+}
+.logs {
+  overflow-y: scroll;
+  height: 25rem;
+  cursor: default;
+  border-radius: 0;
+  padding: 0;
+}
--- a/addons/console/style.css
+++ b/addons/console/style.css
@@ -1,0 +1,72 @@
+.debug-btn {
+    width: 2rem;
+    height: 2rem;
+    padding: 0.375rem;
+    border-radius: 0.25rem;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+    user-drag: none;
+    cursor: pointer;
+  }
+  
+  .debug-btn:hover {
+    background-color: hsla(215, 100%, 65%, 0.15);
+  }
+  .debug {
+    display: none;
+  }
+  .debug {
+    position: absolute;
+    z-index: 1000;
+    background-color: #f1f1f1;
+    border: 1px solid #d3d3d3;
+    min-height: 25rem;
+    min-width: 25rem;
+    max-height: 25rem;
+  }
+  
+  .debug h1 {
+    padding: 10px;
+    cursor: move;
+    z-index: 10;
+    color: #fff;
+  }
+  .log {
+      --message-border-color: rgb(240 240 240);
+      --warning-border-color: hsl(50deg 100% 88%);
+      --error-border-color: hsl(0deg 100% 92%);
+      --error-text-color: red;
+      --monospace-font-size: 12px;
+      --monospace-font-family: consolas, lucida console, courier new, monospace;
+      --source-code-font-size: 12px;
+      --source-code-font-family: consolas, lucida console, courier new, monospace;
+      font-family: var(--monospace-font-family);
+      font-size: var(--monospace-font-size) !important;
+      word-wrap: break-word;
+      user-select: text;
+      box-sizing: border-box;
+      min-width: 0;
+      min-height: 0;
+      display: flex;
+      border-top: 1px solid var(--message-border-color);
+      border-bottom: 1px solid transparent;
+      background: white;
+      word-wrap: break-word;
+      white-space: pre-wrap;
+      flex: auto;
+      line-height: 1.2;
+      padding-left: 10px;
+      border-top-right-radius: revert;
+      border-top-left-radius: revert;
+      padding: revert;
+      cursor: text;
+  }
+  .logs {
+      overflow-y: scroll;
+      height: 25rem;
+      cursor: default;
+      border-radius: 0;
+      padding: 0;
+  }
--- a/addons/console/userscript.js
+++ b/addons/console/userscript.js
@@ -18,19 +18,19 @@ export default async function ({ addon, global, console, msg }) {
   let debug;
   const addItem = (thread) => {
     let logs = document.querySelector(".debug > .logs");
-    let div = document.createElement('div')
-    div.innerText = thread.stackFrames[0].params.text
-    div.classList = `log ${addon.tab.scratchClass('sprite-info_sprite-info')}`
-    logs.appendChild(div)
+    let div = document.createElement("div");
+    div.innerText = thread.stackFrames[0].params.text;
+    div.classList = `log ${addon.tab.scratchClass("sprite-info_sprite-info")}`;
+    logs.appendChild(div);
     console.log(thread.stackFrames[0].params.text);
   };
   const addConsole = () => {
     document.querySelector("body").insertAdjacentHTML(
       "afterbegin",
       `
-    <div class="debug ${addon.tab.scratchClass('card_card')} ">
-    <h1 class="${addon.tab.scratchClass('card_header-buttons')}">Debugger</h1>
-    <div class="${addon.tab.scratchClass('sprite-info_sprite-info')} logs">
+    <div class="debug ${addon.tab.scratchClass("card_card")} ">
+    <h1 class="${addon.tab.scratchClass("card_header-buttons")}">Debugger</h1>
+    <div class="${addon.tab.scratchClass("sprite-info_sprite-info")} logs">
     </div>
     </div>
     `
--- a/addons/console/userscript.js
+++ b/addons/console/userscript.js
@@ -1,0 +1,190 @@
+export default async function ({ addon, global, console, msg }) {
+  const img = document.createElement("img");
+  img.className = "debug-btn";
+  img.src = addon.self.dir + "/debug.svg";
+  img.draggable = false;
+  img.title = msg("debug");
+  img.addEventListener("click", () => showConsole());
+
+  const vm = addon.tab.traps.vm;
+  let blockColorOverrides = {};
+  blockColorOverrides["log %s"] = {
+    color: "#43cfca",
+    secondaryColor: "#3aa8a4",
+    tertiaryColor: "#3aa8a4",
+  };
+
+  let injected;
+  let debug;
+  const addItem = (thread) => {
+    let logs = document.querySelector(".debug > .logs");
+    let div = document.createElement('div')
+    div.innerText = thread.stackFrames[0].params.text
+    div.classList = `log ${addon.tab.scratchClass('sprite-info_sprite-info')}`
+    logs.appendChild(div)
+    console.log(thread.stackFrames[0].params.text);
+  };
+  const addConsole = () => {
+    document.querySelector("body").insertAdjacentHTML(
+      "afterbegin",
+      `
+    <div class="debug ${addon.tab.scratchClass('card_card')} ">
+    <h1 class="${addon.tab.scratchClass('card_header-buttons')}">Debugger</h1>
+    <div class="${addon.tab.scratchClass('sprite-info_sprite-info')} logs">
+    </div>
+    </div>
+    `
+    );
+    let debug = document.querySelector(".debug");
+
+    var pos1 = 0,
+      pos2 = 0,
+      pos3 = 0,
+      pos4 = 0;
+    document.querySelector(".debug h1").onmousedown = dragMouseDown;
+
+    function dragMouseDown(e) {
+      e = e || window.event;
+      e.preventDefault();
+      // get the mouse cursor position at startup:
+      pos3 = e.clientX;
+      pos4 = e.clientY;
+      document.onmouseup = closeDragElement;
+      // call a function whenever the cursor moves:
+      document.onmousemove = elementDrag;
+    }
+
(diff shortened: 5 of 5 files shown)


retronbv:
_moving from convo with colaber_

maybe a way to save the logs (?)
maybe as a scratch list (?)

GrahamSH-LLK:
> _moving from convo with colaber_
> 
> maybe a way to save the logs (?)
> maybe as a scratch list (?)

Saving logs is ezpz, scratch list is ah

retronbv:
> > _moving from convo with colaber_
> > maybe a way to save the logs (?)
> > maybe as a scratch list (?)
> 
> Saving logs is ezpz

just to clarify, saving in memory so you can reload the page and see past logs. OR, saving as a file and allowing opening the file (?)

RedGuy12:
heres a bunch of suggestions. you don’t need to listen to me theese are just my ideas



add colaber and easic to credits



change notice on pause addon to reflect new block name



i feel like the blocks should be #ff7b26 to reflect the sa logo but then it'd be too similar to variables

of course people could make any two colors the same with the customizable colors addon

so maybe make the blocks a custom color?



the console should automatically scroll to the bottom

mxmou:
I don't think this needs to be separate from the pause button.

TheColaber:
> I don't think this needs to be separate from the pause button.

yeah, i agree, we will probably merge the two and call the addon debugger or such

TheColaber:
Should we merge editor-stepping or other addons into this addon? What setting should be enabled by default? I think we want the console to be enabled by default. Should we have a send feedback button?

WorldLanguages:
@GrahamSH-LLK can you share a project that allows for testing? try to cover every possible log type, edge cases etc

GrahamSH-LLK:
> @GrahamSH-LLK can you share a project that allows for testing? try to cover every possible log type, edge cases etc

https://scratch.mit.edu/projects/515191926/editor/

WorldLanguages:
1. Logs should scroll to bottom by default (if you're already at the bottom, and a new log is added, it should stay scrolled at the bottom)
2. Blocks will need `sa-` prefix
3. Would be cool to have timestamps next to logs, for example, including # of project run (incremental, amount of times green flag was pressed), seconds and milliseconds.
4. To improve performance, logs might benefit from `content-visibility: auto`
5. Logs window should probably have more padding
6. Logs window shouldn't be able to be dragged outside of the viewport - users will often want to keep it exactly in a corner
7. It's a bad idea to register an event listener for each log (click event listener for sprite name that links to the block that

content-scripts/prototype-handler.js
```diff
@@ -14,6 +14,7 @@ function injectPrototype() {
       Object.prototype.hasOwnProperty.call(args[0], "runtime")
     ) {
       onceMap.vm = args[0];
+      window.__scratchAddonsTraps.dispatchEvent(new Event("gotvm"));
```
WorldLanguages, line 17: Probably not the proper way to do it

addons/a11y/userscript.js
```diff
@@ -198,21 +99,9 @@ export default async function (o) {
   async function renderLoop() {
     while (true) {
       cleanUp();
-      for (let e of Object.entries(a11yObjects)) {
-        updateAria(vm.runtime.targets.filter((a) => a.id === e[0])[0]);
+      for (let e in a11yObjects) {
```
WorldLanguages, line 102: a11y updates in another PR please

addon-api/content-script/blocks.js
```diff
@@ -0,0 +1,206 @@
...
+              `</top>`,
+            "text/xml"
+          )
+          .querySelectorAll("block, label"),
+      ];
+    });
+    return originalShow.call(this, xml);
+  };
+
+  // We use Scratch's extension category mechanism to replace the data category with our own.
```
WorldLanguages, line 125: This code was taken from `data-category-tweaks-v2` right?

content-scripts/prototype-handler.js
GrahamSH-LLK: cc @easrng

easrng: Why not? It's simple and it works.

addons/a11y/userscript.js
easrng: They aren't updates. It just make it use the new blocks api. Same for pause, tho I did add the breakpoint block. (Same as sa-pause)

addon-api/content-script/blocks.js
easrng: Taken from a11y which took from data category tweaks.

easrng:
> * Blocks will need `sa-` prefix

Wait, what? I thought we were phasing out sa-*.

easrng:
> * The new `addBlock` API is cool and such but it might take some time to make sure we're doing things right

The only issue is l10n which will be very hard to do. This isn't an issue with the api per se, pause and a11y had it too.

content-scripts/prototype-handler.js
apple502j: VM should be available asap for runAtComplete addons, other addons can listen for redux
I think I had that code until 1.5 or so when we refactored addon stuff, plz don't add it back

easrng: This is used in addon-api, how would I use redux for this?

apple502j: `init` should be something addons call explicitly. Not all editor addons add blocks.

addons/a11y/userscript.js
WorldLanguages: @easrng but wasn't the code that handled a11y blocks commented/impossible to run? why are you updating code that will not run?

WorldLanguages: > we are phasing sa- out

Why?

WorldLanguages:
> Wait, what? I thought we were phasing out sa-*.

Source?

addons/a11y/userscript.js
easrng: > wasn't the code that handled a11y blocks commented/impossible to run?

The code to add the blocks to the editor was disabled but it always processed the block runs. If you want to hide the a11y blocks from the flyout again that will be easy to do.

easrng: > > we are phasing sa- out
> 
> Why?

Idk, it was discussed in the discord.

easrng:
> > Wait, what? I thought we were phasing out sa-*.
> 
> Source?

Read from https://discord.com/channels/751206349614088204/751207339545198682/831872159927894057

addons/debugger/addon.json
```diff
@@ -0,0 +1,47 @@
...
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Export format",
+      "id": "exportFormat",
+      "type": "string",
+      "default": "${sprite}: ${content} (${type})"
```
RedGuy12, line 41: i think this is too much. why would you need to change the log output format lol

addons/debugger/userscript.js
```diff
@@ -0,0 +1,255 @@
...
+  addon.tab.addBlock("sa-log %s", ["content"], ({ content }, targetId, blockId) => {
+    workspace = Blockly.getMainWorkspace();
+    addItem(content, targetId, blockId, "log");
+  });
+  addon.tab.addBlock("sa-warn %s", ["content"], ({ content }, targetId, blockId) => {
+    workspace = Blockly.getMainWorkspace();
+    addItem(content, targetId, blockId, "warn");
+  });
+  addon.tab.addBlock("sa-error %s", ["content"], ({ content }, targetId, blockId) => {
+    workspace = Blockly.getMainWorkspace();
```
RedGuy12, line 20: why don't you get the workspace inside of the `addItem` function (to avoid duplication)

addons/debugger/addon.json
jeffalo: scratch addons is all about customization. more options like this = good.

mxmou: @jeffalo Depends, see #1262. More options can also cause problems.

mxmou:
@easrng Can you explain the reason why the blocks don't have the sa- prefix here, for those who aren't in the Discord?

addons/debugger/addon.json
jeffalo: i think the customization provided by a custom log format is useful

WorldLanguages:
@mxmou It basically was just @apple502j saying sa- prefix is stupid, would be great to have Apple write down here why he said that? (he didn't explain in the discord)

addons/debugger/userscript.js
GrahamSH-LLK: Because idk. It was causing weird bugs.

retronbv:
Should we add the new blocks into the block switching addon?

WorldLanguages:
Logs seem broken rn?

GrahamSH-LLK:
> Logs seem broken rn?
> (nvm, the testing project probably should use sa- prefix and it isn't)

Yep, I forgot to fix it

WorldLanguages:
If anyone is still working on this, see https://github.com/ScratchAddons/ScratchAddons/pull/2161#issuecomment-821902497

GrahamSH-LLK:
> If anyone is still working on this, see [#2161 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/2161#issuecomment-821902497)

I am, I'll try to resolve some of these.

WorldLanguages:
Merging for next milestone: v1.15.0 (~May 14th)

content-scripts/prototype-handler.js
GrahamSH-LLK: @easrng It would be great if you could fix this.

TheColaber:
We need consistency here, but I will work on this, and other things
image

RedGuy12:
Just add a background and border to the debugger button, then it'll be consistent with the other nearby buttons.

TheColaber:
Rethinking merging other addons with the addon to call it "Debugger" instead of "Logs"...

#1934, could keep these addons separate, but there being a preset called "Debugger" which has Logs, Pause, Stepping, etc in the preset.

Just an idea and alternative, to merging multiple addons that do different things.

WorldLanguages:
Probably not going to make it for v1.15.0 (specially if we want TurboWarp support, and clean code)

WorldLanguages:
I don't think export format should be a setting, it's too specific. We could keep the functionality but only when the export button is shift+clicked, and using a `prompt()`

WorldLanguages:
We may keep the "breakpoint" block, but we should provide a way to unpause the project within the logs window in case the pause addon is not enabled. Alternatively, the breakpoint could be ignored if there's no pause button, but that would be a bit confusing.
Also, not sure why it's not `sa-breakpoint`.
`sa-pause` should be removed from the pause button

WorldLanguages:
We should find a proper CSS fix for https://github.com/ScratchAddons/ScratchAddons/pull/2161/commits/2f5aac2ebbefd2dff9e937df51e738945afad1c3

WorldLanguages:
And maybe most importantly, I don't think we should have a Scratch Addons block category. Debugger addon should have its own category, and maybe we could clarify it's from Scratch Addons within the category name (but not in the short name displayed below the category icon). Code to add a new category could be added into a library and we could have data-category-tweaks-v2 use it.

mxmou:
> We should find a proper CSS fix for [2f5aac2](https://github.com/ScratchAddons/ScratchAddons/commit/2f5aac2ebbefd2dff9e937df51e738945afad1c3)

Done. I don't think it's a good idea to use Scratch classes just to set text and background colors: the code will be easier to understand if classes are only used for their intended purposes.

Some comments: 
* I think the icon should be dark (`#575e75`) and thicker to match the full screen icon. I'll change the way it and the gamepad icon is added to the stage header to improve dark mode compatibility.
* Could the header be blue instead of green? It isn't a modal but it isn't a tutorial either.
* The added blocks should either appear in the normal custom block color or a color that has good text contrast. In the latter case support for it should be added to custom block colors.

GarboMuffin:
I tried this out today and apparently it has a very expensive loop that runs every animation frame. In https://scratch.mit.edu/projects/463553665/ and you're doing 9ms of main thread work every animation frame when the project isn't even running.

Anyways, I'm going to give a good look at this (at least the API parts) today/tonight

GarboMuffin:
You probably want these commits: https://github.com/GarboMuffin/ScratchAddons/tree/debugger2

Cleaned up large parts of API, fixed performance

WorldLanguages:
We were close, but I'd prefer merging for v1.16.0

addons/debugger/userscript.js
```diff
@@ -0,0 +1,296 @@
...
+  buttonImage.className = addon.tab.scratchClass("stage-header_stage-button-icon");
+  buttonImage.draggable = false;
+  buttonImage.src = addon.self.dir + "/debug.svg";
+  buttonContent.appendChild(buttonImage);
+  buttonContainer.appendChild(buttonContent);
+  container.appendChild(buttonContainer);
+  buttonContainer.addEventListener("click", () => toggleConsole(true));
+
+  const vm = addon.tab.traps.vm;
+  addon.tab.addBlock("sa-log %s", ["content"], ({ content }, thread) => {
```
apple502j, line 20: Please, do, not, name, this, "sa". Just do a proper conflict handling.

addons/debugger/userscript.js
```diff
@@ -0,0 +1,296 @@
...
+    workspace = Blockly.getMainWorkspace();
```
apple502j, line 21: Why do we re-assign this on every call? Can't we do this downstream, when needed?

addons/debugger/userscript.js
```diff
@@ -0,0 +1,296 @@
...
+            : inputBlockFill;
+        inputSpan.innerText = inputBlock.toString();
+        inputSpan.className = "console-variable";
+        inputSpan.style.background = inputBlockColor;
+        wrapper.append(inputSpan);
+      }
+    }
+    let string = addon.settings
+      .get("exportFormat")
+      .replace("{sprite}", targetName)
```
apple502j, line 265: This may log incorrect stuff for sprites named {content}. Consider regex.

addons/gamepad/userscript.js
```diff
@@ -33,12 +33,10 @@ export default async function ({ addon, global, console, msg }) {
   const height = renderer._yTop - renderer._yBottom;
   const canvas = renderer.canvas;
 
-  const spacer = document.createElement("div");
```
apple502j, line 36: Separate PR please

apple502j (changes requested):

addons/debugger/userscript.js
mxmou: What's wrong with naming it sa, and what's "proper conflict handling"?

addons/gamepad/userscript.js
mxmou: The change is for compatibility with the addon this PR adds.

addons/debugger/userscript.js
GrahamSH-LLK: Wouldn't replacing content first fix this?

addon-api/content-script/blocks.js
```diff
@@ -0,0 +1,174 @@
...
+  // https://github.com/LLK/scratch-gui/blob/ddd2fa06f2afa140a46ec03be91796ded861e65c/src/containers/blocks.jsx#L344
+  // https://github.com/LLK/scratch-vm/blob/a0c11d6d8664a4f2d55632e70630d09ec6e9ae28/src/engine/runtime.js#L1381
+  const originalGetBlocksXML = vm.runtime.getBlocksXML;
+  vm.runtime.getBlocksXML = function (target) {
+    const result = originalGetBlocksXML.call(this, target);
+    result.unshift({
+      id: "sa-blocks",
+      xml:
+        "<category" +
+        ` name="${escapeHTML(scratchAddons.l10n.get("extensionName", null, "Scratch Addons"))}"` +
```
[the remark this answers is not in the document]
easrng: It exists in other localizations and there's a fallback so it should be fine.

addon-api/content-script/blocks.js
```diff
@@ -0,0 +1,174 @@
...
+};
+
+let inited = false;
+export async function init(tab) {
+  if (inited) {
+    return;
+  }
+  inited = true;
+
+  if (tab.clientVersion !== "scratch-www") {
```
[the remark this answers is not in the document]
easrng: We only want to init once and not on r2 pages

addon-api/content-script/blocks.js
```diff
@@ -0,0 +1,174 @@
...
+    xml +=
+      '<block type="procedures_call" gap="16"><mutation generateshadows="true" warp="false"' +
+      ` proccode="${escapeHTML(proccode)}"` +
+      ` argumentnames="${escapeHTML(JSON.stringify(names))}"` +
+      ` argumentids="${escapeHTML(JSON.stringify(ids))}"` +
+      ` argumentdefaults="${escapeHTML(JSON.stringify(defaults))}"` +
+      "></mutation></block>";
+  }
+  if (xml.length === 0) {
+    const message = scratchAddons.l10n.get("noAddedBlocks", null, "No addons have added blocks.");
```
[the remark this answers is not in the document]
easrng: Yes I know but this should get localized. IDK how adding l10n messages works.

easrng:
The **Logs** title seems way too big and out of place

easrng:
Also there should be a little dot or badge on the corner of the logs icon when there's unread logs.

easrng:
> The added blocks should either appear in the normal custom block color or a color that has good text contrast. In the latter case support for it should be added to custom block colors.

I agree that the current color isn't great but I don't think we should use the custom block colors. The color should be easily distinguishable from other block types. Right now it is easy to distinguish but like you said the contrast isn't great. If we used the SA color it would be easy to confuse with variables.

A screenshot showing color differences between blocks

addons/debugger/userscript.js
WorldLanguages: > Just do a proper conflict handling.

Garbo made it so that if projects have a `sa-log` custom block, it won't run at all if this addon is enabled. So if we keep this behavior we'll need to keep the sa prefix.

GarboMuffin: > Garbo made it so that if projects have a sa-log custom block, it won't run at all if this addon is enabled

Prior to this, it would run the custom block with its arguments all set to their default value. This is a more complicated problem to fix than it seems. I'll take another look at some point

addon-api/content-script/blocks.js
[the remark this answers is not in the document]
apple502j: Then it should also fail on non-project www pages

apple502j:
Idea: rename to these
- sa-log -> "Add debug log"
- sa-warn -> "Add debug warning"
- sa-error -> "Add debug error"

RedGuy12:
Is the block compatible with custom block colors, newly the black and dark themes?

easrng:
> Idea: rename to these [snip]

How about 
 - log %s to console
 - log %s to console as error
 - log %s to console as warning
This seems more in line with Scratch naming conventions.

mxmou:
I think the block names should remain short.

WorldLanguages:
Writing TODOs again lol

- [ ] Rename addon to "debugger"
- [ ] Rename block category to "debugger"
- [ ] Consider changing block color for this category
- [ ] Consider changing the category icon, from SA icon to that color
- [ ] Move "export format" logic to shift+click on "export"
- [ ] Remove `sa-pause` from pause addon - pause addon should only add the UI element. Debugger should handle pausing separately. If `sa-pause` or `sa-breakpoint` are ran, debugger addon should provide an alternative way to unpause the project, in case the user doesn't have the pause addon enabled.
- [ ] Ideally, but not required: `Also there should be a little dot or badge on the corner of the logs icon when there's unread logs.` (by @easrng)

Also, I think that, at least for now, we should keep the block names as `sa-pause`, `sa-breakpoint`, `sa-log`, `sa-warn` and `sa-error`. If we change opinion in the future, we can provide backwards compatibility without worrying about potential conflicts, since these block names are so specific because of the `sa-` prefix. Going all in with short names can be problematic, but not the other way.
All of this is, of course, as long as user defined custom blocks don't run if the addon is enabled. If we find a way for a custom block that is simply called `log` to not only log something within the addon, but also run the potentially existing `log` project author defined custom block, we could remove the prefix just fine. In other words, we can name the blocks whatever, if we make it impossible for a project to distinguish between a user with debugger disabled, and a user with debugger enabled (without considering tiny differences in time blocks take to run or stuff like that - I'm talking simple stuff, like custom blocks running and getting the correct arguments).

Feel free to discuss about any of the TODOs I mentioned above before I try to tackle them.
If you disagree about releasing with the block prefix, please make it so that a potentially existing custom block runs as well as our code (with the correct arguments passed in to the custom block), or alternatively pressure someone else to find a way to do that :P

easrng:
> * Ideally, but not required: `Also there should be a little dot or badge on the corner of the logs icon when there's unread logs.` (by @easrng)

This was done in c18553b846e0815875ec9e949cac41e4a5a75dbf

addons/gamepad/userscript.js
```diff
@@ -227,11 +224,11 @@ export default async function ({ addon, global, console, msg }) {
   gamepad.addEventListener("mousemove", handleGamepadMouseMove);
 
   while (true) {
-    const stageHeaderWrapper = await addon.tab.waitForElement('[class*="stage-header_stage-menu-wrapper"]', {
+    const stageHeaderSizeControls = await addon.tab.waitForElement('[class*="stage-header_stage-size-row"]', {
```
GarboMuffin, line 227: The changes this PR makes to the gamepad addon makes the settings button never appear in fullscreen mode. That's a bug.

GarboMuffin (changes requested):

addons/gamepad/userscript.js
GarboMuffin: I will fix this in #2658. Would recommend reverting all the gamepad changes here because there's going to be conflicts otherwise.

WorldLanguages: Done

WorldLanguages:
What's the reasoning behind this notice?
image

WorldLanguages:
Also, for the icons, we probably want to include them as part of the addon, instead of providing a scratch.mit.edu URL. This in case the dynamic part of the URL changes, and so that TurboWarp doesn't need to change these manually

RedGuy12:
We don’t want non-sa users seeing the blocks and asking about them on the forums

WorldLanguages:
Yeah, now that I think about it, we should probably launch with a prefix, but not `sa-`, since it could be considered advertising, specially if people go asking what it means...
Some other options, with "log" as an example:
- `tw-log` (tw for turbowarp? lol prob not)
- `__log`
- `log__`
- `@log`
- `🞄 log`
- `🐛 log`
- `​​​​log​​​` (you can't see it, but it's surrounded by two zero width spaces `​` each side, U+200B)

RedGuy12:
No emojis pls, some aren’t supported on some platforms

TheColaber:
I like the last one or  `@log`. I don't really think sa- could cause issues anyway since many people in the forums have wrote "SA" and the mods didn't care

RedGuy12:
> * log %s to console
> * log %s to console as error
> * log %s to console as warning

this pls

to be consistent with scratch

WorldLanguages:
Hmm I guess that a very specific and long block name like that could also work instead of a prefix.

mxmou:
> > * log %s to console
> > * log %s to console as error
> > * log %s to console as warning
> 
> this pls
> 
> to be consistent with scratch

I think these names are too long - the error and warning blocks don't need "to console" in their name.

apple502j:
merge conflict

GrahamSH-LLK:
> merge conflict

Not anymore

WorldLanguages:
Block names for this release are:
- `sa-pause` (backwards compatibility)
- `\u200B\u200Bbreakpoint\u200B\u200B`
- `\u200B\u200Blog\u200B\u200B`
- `\u200B\u200Bwarn\u200B\u200B`
- `\u200B\u200Berror\u200B\u200B`

These block names aren't localized.
We might try to localize them and/or want to change the block names in the future. If so, we can provide backwards compatibility without any issues.

addon-api/content-script/blocks.js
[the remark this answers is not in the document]
WorldLanguages: Will not happen anyway

WorldLanguages (approved):

addon-api/content-script/blocks.js
```diff
@@ -119,7 +119,7 @@ const injectWorkspace = (ScratchBlocks) => {
       id: "sa-blocks",
       xml:
         "<category" +
-        ` name="${escapeHTML(scratchAddons.l10n.get("extensionName", null, "Scratch Addons"))}"` +
```
GrahamSH-LLK: Please no. This makes it so that only this addon can use the blocks api

WorldLanguages: And that will be the case for now anyway

WorldLanguages:
Wide variable bug (not turbowarp bug):

image

WorldLanguages:
CSS looks a bit sketchy and is definitely not RTL friendly but this is launching as beta anyway.

WorldLanguages:
Merging so that strings are sent for translations. If anyone wants to make any more little changes feel free to create another PR.

PR #2162, opened 2021-04-14
Lower addons.json conflicts

GrahamSH-LLK:
In a sense, it allows a trailing comma, so conflicts will be way easier to resolve.


TheColaber:
Will be resolved when #1961 is merged

PR #2169, opened 2021-04-15
Invite button above studio comments

### LINKED ISSUE #2150: Invite Curator button next to comments on studios
**Describe the new addon**

Similar to "Copy Link", this addon would add a "Invite Curator" button to comments in studios that you manage. If , the invitation is successful, the text changes to "Done!" Otherwise, it'd say either "This user is already a curator" or "Whoops, something went wrong", depending on the response given. 

This addon would be really useful for managers of active studios. It would enable us to efficiently invite people who ask without needing to open the curators tab, paste the username of the curator, and go back to the comments tab and reply each time.

A mockup of the extension:

Screenshot 2021-04-13 145435

Screenshot 2021-04-13 145944

Screenshot 2021-04-13 150815

Screenshot 2021-04-13 150630

CubeyTheCube:
**Resolves**

Resolves #2150

**Changes**

Added a button that shows up when you hover over a studio comment, and when clicked it will attempt to invite the comment author the the studio.

**Tests**

Tested on Google Chrome
Note: I tested this by inserting an edited version of the userscript into the javascript console, in theory should not be an issue but should be tested to be safe


RedGuy12 (changes requested):
Please test your addons, and i don't mean just copy-pasting the code. Actually load the branch into chrome.

RedGuy12:
Also please enable prettier here: https://github.com/CubeyTheCube/ScratchAddons/actions

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+    inviteButton.style.color = 'rgb(157, 157, 157)';
+
+    comment.querySelector('.actions-wrap').appendChild(inviteButton);
+    comment.addEventListener('mouseover', () => {
+      inviteButton.style.visibility = 'visible'; // should only be visible when the mouse is hovering over the comment
+    });
+    comment.addEventListener('mouseout', () => {
+      inviteButton.style.visibility = 'hidden';
+    });
+    const listener = inviteButton.addEventListener('click', () => {
```
TheColaber, line 33: "listener" should be set to the function, not the whole listener.

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+        try {
+          const parsed = JSON.parse(resp);
+          // handle the different possible responses
+          if (parsed.status === 'success') {
+            inviteButton.innerHTML = 'Invited!';
+          } else if (parsed.message.endsWith('is already a curator of this studio')) {
+            inviteButton.innerHTML = 'This user is already a curator';
+          } else { throw Error(0); }
+        } catch (err) {
+          inviteButton.innerHTML = 'Whoops, something went wrong';
```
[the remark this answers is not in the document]
TheColaber: Why not?

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+        }
+        comment.querySelector('.reply').click();
```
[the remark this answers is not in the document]
retronbv: > Why?

You weren’t there for the convo in the discord I think.. 

It’s so people can respond with “Added” or “Invited”

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+        inviteButton.innerHTML = inviteButton.innerHTML.bold();
```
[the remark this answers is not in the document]
TheColaber: Bolds it. I would try not to use innerHTML...

addons/studio-invite-button/userscript.js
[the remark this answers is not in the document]
TheColaber: Why not?

addons/studio-invite-button/userscript.js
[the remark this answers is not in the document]
TheColaber: Bolds it. But you should try to avoid using innerHTML.

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+    body: null,
+    method: 'PUT',
+  });
+}
+// checks if the user is a manager and the URL is correct
+if (document.getElementById('description').classList.contains('editable') && urlRegex.test(window.location.href)) {
+  Array.from(document.getElementsByClassName('comment')).forEach((comment) => {
+    const inviteButton = document.createElement('span'); // create the button
+    inviteButton.innerHTML = 'Invite';
```
TheColaber, line 21: Wherever you set the innerHTML to just text, change it to be innerText.

GrahamSH-LLK: Also, use l10n

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+    inviteButton.style.visibility = 'hidden';
+    inviteButton.style.color = 'rgb(157, 157, 157)';
+
+    comment.querySelector('.actions-wrap').appendChild(inviteButton);
+    comment.addEventListener('mouseover', () => {
+      inviteButton.style.visibility = 'visible'; // should only be visible when the mouse is hovering over the comment
+    });
+    comment.addEventListener('mouseout', () => {
+      inviteButton.style.visibility = 'hidden';
+    });
```
TheColaber, line 32: If you add the "report" class to the inviteButton, you don't need this code.

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,54 @@
+export default async function ({ addon, global, console }) {
+// RegExp to test for a valid url and get the studio id
+const urlRegex = /^https?:\/\/scratch\.mit\.edu\/studios\/(\d+)\/comments($|\/)/;
```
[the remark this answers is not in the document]
GrahamSH-LLK: > Change the match patterns, don't use a regex

In this case they're testing to see if the user is on comments, but there should be a better way to do that

addons/studio-invite-button/userscript.js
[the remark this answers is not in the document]
GrahamSH-LLK: This can easily be done with CSS.

addons/studio-invite-button/manifest.json
```diff
@@ -0,0 +1,12 @@
...
+  "name": "Studio Invite Button",
+  "description": "Adds a button to add comment authors to a studio.",
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/studios/*"]
+    }
+  ],
+  "tags": ["community"],
+  "enabled_by_default": false
```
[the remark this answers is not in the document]
GrahamSH-LLK: That's the second time someone has done this, is there a guide that uses this case?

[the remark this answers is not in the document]
Secret-chest: Yes

addons/studio-invite-button/userscript.js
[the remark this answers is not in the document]
mxmou: > In this case they're testing to see if the user is on comments, but there should be a better way to do that

For now doing it using a match pattern is fine because studio tabs are handled server-side, but when the studio page is migrated to `scratch-www` the best way will be `waitForElement`.

addons/studio-invite-button/manifest.json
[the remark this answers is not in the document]
RedGuy12: where?

addons/studio-invite-button/userscript.js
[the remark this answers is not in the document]
RedGuy12: oh i was thinking the `catch` was only run on the `throw Error(0);`. but i guess this is good since now it can catch other errors, such as if the `JSON.parse(resp)` fails.

addons/studio-invite-button/manifest.json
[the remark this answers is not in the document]
GrahamSH-LLK: > where?

I already found and fixed it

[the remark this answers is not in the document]
mxmou: @RedGuy12 In the Creating an addon tutorial. It's now fixed.

ZI-Youtuber (approved):

addons/studio-invite-button/userscript.js
CubeyTheCube: nope, to remove the listener I need to set a variable to the whole thing

TheColaber: Please read the docs https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

addons/studio-invite-button/userscript.js
CubeyTheCube: I don't want to risk other side affects that could happen when I do that

TheColaber: there are no "side affects"

addons/studio-invite-button/userscript.js
CubeyTheCube: > Wherever you set the innerHTML to just text, change it to be innerText.

thanks

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,64 @@
...
+        },
+        body: null,
+        method: "PUT",
+      }
+    );
+  }
+  addon.tab.waitForElement("#description").then((description) => {
+    // checks if the user is a manager and the URL is correct
+    if (description.classList.contains("editable") && urlRegex.test(window.location.href)) {
+      Array.from(document.getElementsByClassName("comment")).forEach((comment) => {
```
TheColaber, line 24: you shouldn't need an Array.from here i think

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,64 @@
...
+                // handle the different possible responses
+                if (parsed.status === "success") {
+                  inviteButton.innerText = "Invited!";
+                } else if (parsed.message.endsWith("is already a curator of this studio")) {
+                  inviteButton.innerText = "This user is already a curator";
+                } else {
+                  throw Error(0);
+                }
+              } catch (err) {
+                inviteButton.innerText = "Whoops, something went wrong";
```
TheColaber, line 53: localize https://github.com/ScratchAddons/ScratchAddons/wiki/Localizing-addons

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,64 @@
...
+        method: "PUT",
+      }
+    );
+  }
+  addon.tab.waitForElement("#description").then((description) => {
+    // checks if the user is a manager and the URL is correct
+    if (description.classList.contains("editable") && urlRegex.test(window.location.href)) {
+      Array.from(document.getElementsByClassName("comment")).forEach((comment) => {
+        const inviteButton = document.createElement("span"); // create the button
+        inviteButton.innerText = "Invite";
```
TheColaber, line 26: localize https://github.com/ScratchAddons/ScratchAddons/wiki/Localizing-addons

addons/studio-invite-button/userscript.js
TheColaber, line 24: This won't work for newly created comments. Do a `addon.tab.waitForElement` for `.comment`.

addons/studio-invite-button/userscript.js
CubeyTheCube: done

addons/studio-invite-button/userscript.js
CubeyTheCube: yes you do, NodeLists are not Arrays

TheColaber: Even though, the user should be seeing it on all comments.

addons/studio-invite-button/userscript.js
CubeyTheCube: okay

addons/studio-invite-button/userscript.js
TheColaber: oops... but forEach does work for querySelectorAll

CubeyTheCube: Okay, but how would addon.tab.waitForElement handle new comments?

TheColaber: put it in a forever loop and use an `await`... smth like
```js
while (true) {
const comment = await addon.tab.waitForElement(".comment", { markAsSeen: true });
...
}
```

CubeyTheCube: huh, I never knew that.

CubeyTheCube: do I have to wrap it in an async IIFE?

mxmou: > do I have to wrap it in an async IIFE?

No, everything is already inside an async function.

RedGuy12: you need to convert
image
to async tho

TheColaber: You should actually remove that...

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,58 @@
...
+        const listener = inviteButton.addEventListener("click", () => {
+          inviteToStudio(comment.querySelector(".name").textContent.trim())
+            .then((resp) => resp.text())
+            .then((resp) => {
+              try {
+                const parsed = JSON.parse(resp);
+                // handle the different possible responses
+                if (parsed.status === "success") {
+                  inviteButton.innerText = msg("invited");
+                } else if (parsed.message.endsWith("is already a curator of this studio")) {
```
RedGuy12, line 41: does this change with other languages?

addons/studio-invite-button/userscript.js
CubeyTheCube: @RedGuy12 told me to use addon.tab.waitForElement in that case, are you saying I should not?

addons/studio-invite-button/userscript.js
CubeyTheCube: no, the scratch api's responses are always the same language afaik

addons/studio-invite-button/userscript.js
RedGuy12: i said for the comment not for the desc

TheColaber: If it's needed in order to check if you are in a studio, then you can keep it and make it async... and also remove the regex.

CubeyTheCube: What's wrong with the regex? I need to test if it's in the studio's comments, and find the studio ID.

mxmou: > What's wrong with the regex? I need to test if it's in the studio's comments, and find the studio ID.

You can change the match pattern for the script to `https://scratch.mit.edu/studios/*/comments/` and get the ID as `location.pathname.split("/")[2]`.

CubeyTheCube: Thanks, someone in the discord server told me to just test if the user was on /comments from inside the code

addons/studio-invite-button/userscript.js
```diff
@@ -0,0 +1,57 @@
...
+        },
+        body: null,
+        method: "PUT",
+      }
+    );
+  }
+  addon.tab.waitForElement("#description").then(async (description) => {
+    // checks if the user is a manager and the URL is correct
+    if (description.classList.contains("editable")) {
+      for (;;) {
```
RedGuy12, line 22: use while?

RedGuy12 (approved):

addons/studio-invite-button/userscript.js
CubeyTheCube: `for(;;)` is more readable imo, if it's the general practice to use while in scratchaddons ill switch though

PR #2173, opened 2021-04-15
For GrahamSH to merge

easrng:
just dewit

### DIFF
--- a/addons/pause/userscript.js
+++ b/addons/pause/userscript.js
@@ -72,10 +72,11 @@ export default async function ({ addon, global, console, msg }) {
       pausedThreadState = new WeakMap();
     }
   };
-
-  addon.tab.addBlock("sa-pause", [], (role, target) => {
+  function pauseBlock(role, target) {
     setPaused(true);
-  });
+  }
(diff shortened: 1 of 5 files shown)


addons/pause/userscript.js
```diff
@@ -72,15 +72,11 @@ export default async function ({ addon, global, console, msg }) {
...
-      setPaused(true);
-      return;
-    }
-    return originalStepToProcedure.call(this, thread, proccode);
-  };
+  function pauseBlock(role, target) {
+    setPaused(true);
+  }
+  addon.tab.addBlock("sa-pause", [], pauseBlock, true);
+  addon.tab.addBlock("breakpoint", [], pauseBlock);
```
RedGuy12, line 79: what's the difference between these two blocks

GrahamSH-LLK: Nothing, one's for backwards compatability

mxmou: > Nothing, one's for backwards compatability

Does the name need to be changed? If it does, the `sa-` prefix should be re-added, otherwise it would break projects with a block called "breakpoint".

PR #2174, opened 2021-04-15
Issues in editor when user logged out

### LINKED ISSUE #2171: Set thumbnail button is shown when the user is logged out
**Describe the bug**

The set thumbnail button always shows in the editor, even if logged out.

**To Reproduce**

1. Make sure that the thumbnail setter is enabled
2. Log out
3. Go to the editor

**Expected behavior**

It would not be visible.

TheColaber:
**Resolves**

Resolves #2171

**Changes**

animated-thumb: returns if there is no input to change the project title.
last-edit-tooltip: fix error when logged out

**Tests**

works i think


addons/animated-thumb/userscript.js
```diff
@@ -5,33 +5,28 @@ export default async function ({ addon, global, console, msg }) {
...
-      thumbspan.innerText = msg("set-thumbnail");
-      thumbcontent.append(thumbspan);
-      nav.append(setthumb);
-      setthumb.addEventListener("click", function (e) {
-        const setter = new ThumbSetter((key) => msg(`/${key}`));
-        setter.addFileInput();
-        setter.showInput();
-      });
-    }
+    if (!document.querySelectorAll("[class*='project-title-input_title-field']")) return;
```
RedGuy12, line 8: why not check `addon.auth.isLoggedIn`?

mxmou: @RedGuy12 That would be a better idea because Scratch sometimes needs a while to check if the user is logged in or not.

WorldLanguages:
> editor-msgs: only fetch msg count when the element is actually appended

I don't see how that's the case, current master and this PR keep fetching the count even after going to player mode.

addons/animated-thumb/userscript.js
WorldLanguages: This previously used whether the project says "by username" to decide if the button should be added or not, which probably uses the same API request as checking whether the editor should have editable title 🤔

Secret-chest (approved):
This looks good!

PR #2184, opened 2021-04-16
Better handle key/mouse events in 60fps

### LINKED ISSUE #2153: 60 FPS Sticky Toggling
**Describe the bug**

When you alt + click the green flag on a Scratch project and have the 60 FPS addon enabled, 60 FPS is toggled. However, once the alt key is no longer being pressed, it often still toggles 60 FPS instead of restarting the project when you click the green flag. This continues for an unpredictable number of times until it restarts the project like it's supposed to.

**To Reproduce**

1. Go to any Scratch project
2. Hold down the alt key and click on the green flag.
3. Stop pressing the alt key and click again.
5. Eventually, if you click enough, it'll do what it's supposed to do, which is restart the project.

**Expected behavior**

When the alt key is not pressed, clicking the green flag should not toggle 60 FPS.

**Screenshots**

https://user-images.githubusercontent.com/79656307/114634969-997c4980-9c78-11eb-93dc-6edf62d25855.mp4

(note that I pressed the alt key for the first time I pressed the green flag in the upper left, but not any of the other clicks)

- OS: Windows 10
- Browser: Chrome
- Chrome version 89.0.4389.114

**Additional context**

To work around this bug, you can toggle 60 FPS and then open full screen mode. You can restart the project without the undesired FPS toggling. However, the bug remains when you go back to small screen.

TheColaber:
Resolves #2153 properly
In most devices, it checks whether `e.altKey` is true when clicking the green flag
In chromebooks, it instead assumes a right click in the green flag is an alt+click

### DIFF
--- a/addons/60fps/userscript.js
+++ b/addons/60fps/userscript.js
@@ -14,7 +14,10 @@ export default async function ({ addon, global, console }) {
       button.style.filter = mode ? "hue-rotate(90deg)" : "";
     };
     const flagListener = (e) => {
-      if (e.altKey && !addon.self.disabled) {
+      if (addon.self.disabled) return;
+      const isAltClick = e.type === "click" && e.altKey;
+      const isChromebookAltClick = navigator.userAgent.includes("CrOS") && e.type === "contextmenu";
+      if (isAltClick || isChromebookAltClick) {
         e.cancelBubble = true;
         e.preventDefault();
         changeMode();
--- a/addons/60fps/userscript.js
+++ b/addons/60fps/userscript.js
@@ -4,18 +4,6 @@ export default async function ({ addon, global, console }) {
 
   let global_fps = 30;
   const vm = addon.tab.traps.vm;
-  let altPressesCount = 0;
-  let altPressedRecently = false;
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Needs testing in a real chromebook

GarboMuffin:
It seems that alt+clicking on a Chromebook emits a contextmenu event with altKey: false, so this probably won't work

Perhaps you could do what TW has been doing since October and just make all contextmenu events toggle 60 FPS mode

TheColaber:
> Perhaps you could do what TW has been doing since October and just make all contextmenu events toggle 60 FPS mode

mute-project listens for this event... how would we do this without toggling 60FPS when we actually want to mute the project?

WorldLanguages:
> mute-project listens for this event

...but only does something if `e.ctrlKey` is true

WorldLanguages:
@TheColaber I suggest checking whether `navigator.userAgent.includes("CrOS")`, if so, right click (aka contextmenu event) should be identical to alt+click (making sure to `e.preventDefault()` to avoid the context menu from showing)

PR #2192, opened 2021-04-16
Revert "60FPS - temp fix - reduce timeout time"

WorldLanguages:
Would merge #2184 but it needs to be tested in a chromebook


WorldLanguages:
Closed in favor of #2184

PR #2193, opened 2021-04-16
v1.13.0 set update notification and new addons

WorldLanguages:


mxmou:
Shouldn't all addons mentioned in the notification be pinned, including editor dark mode?

WorldLanguages:
@mxmou Eh good point, but it's not strictly a "new" addon and pinning it to the top will mean it will take its settings will take the whole viewport if the addon is enabled because it will be expanded by default. I'll try to find a workaround (like not extending it by default, even if it's enabled) and pin it to the top as a new addon in a new PR rn

PR #2200, opened 2021-04-17
Fix loading presets for colors & allowTransparency:false not working

### LINKED ISSUE #2097: Color picker doesn't prevent setting transparency if `allowTransparency` is set to false
**Describe the bug**

Yes, another color picker bug. Omitting the `allowTransparency` property of a color setting in `addon.json` adds a `no_alpha` attribute to the web component, which has no effect: <del>the color picker source code doesn't even mention "no_alpha" anywhere</del> I was wrong, it does. This means that the user can still make the color transparent.

**To Reproduce**

1. Clone [this branch](https://github.com/mxmou/ScratchAddons/tree/dark-presets) or just remove `allowTransparency` from an existing `addon.json`
2. Open the settings page
3. Change the transparency of any editor dark mode setting except the border color, or the color from which you removed `allowTransparency`
4. The transparency changes and the transparent setting can be accessed by the addon normally

**Expected behavior**

It would not be possible to make colors that don't enable transparency transparent.

### LINKED ISSUE #2196: Loading presets for addons with color settings is not possible
**Describe the bug**

Caused by #2191

Loading any preset on any addon with color settings (e.g. editor dark mode) results in the following error:
```
Uncaught (in promise) TypeError: Cannot read property '_valueChanged' of undefined
    at VueComponent.value (index.js:75)
```

**To Reproduce**

1. Open the settings page
2. Load any editor dark mode preset
3. Open the console - there is an error
4. Try to interact with the settings page - it doesn't work

**Expected behavior**

Presets could be loaded.

WorldLanguages:
Resolves #2196 
Resolves #2097

Do not call `._valueChanged()` in the color picker if it doesn't exist
Doesn't cause any bugs because when created, the color picker will grab the updated color from `this.value` through the `v-bind:value` HTML attribute

### DIFF
--- a/webpages/settings/index.html
+++ b/webpages/settings/index.html
@@ -42,7 +42,7 @@
               v-show="isOpen"
               v-el:pickr
               :formats="formats"
-              :no_alpha="no_alpha"
+              :no_alpha="noAlphaString"
             ></color-picker>
           </template>
         </button>
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -49,6 +49,11 @@ const ColorInput = Vue.extend({
       }
     });
   },
+  computed: {
+    noAlphaString() {
+      return String(this.no_alpha);
+    }
+  },
   methods: {
     toggle(addon, setting, value = !this.isOpen) {
       if (!this.loadColorPicker) return;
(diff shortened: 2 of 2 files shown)


WorldLanguages:
@mxmou please find more bugs :P

PR #2205, opened 2021-04-18
Lint 20210418

apple502j:


addons-l10n/en/editor-devtools.json
```diff
@@ -10,7 +10,7 @@
...
   "editor-devtools/improved-tidy-up": "Improved Code Tidy Up",
-  "editor-devtools/improved-tidy-up-desc": "Right click on the scripts window to pop up the menu and the Clean up Blocks option will have been replaced by a Clean up Blocks + option. Se this to tidy your scripts and it will preserve your scripts columns as well as attempt to align the comments and remove all those orphaned variables, etc. You'll like this a lot I guarantee!",
+  "editor-devtools/improved-tidy-up-desc": "Right click on the scripts window to pop up the menu and the Clean up Blocks option will have been replaced by a Clean up Blocks + option. Set this to tidy your scripts and it will preserve your scripts columns as well as attempt to align the comments and remove all those orphaned variables, etc. You'll like this a lot I guarantee!",
```
WorldLanguages, line 13: Now that we're on it, let's remove `You'll like this a lot I guarantee!`? It's so out of context, this is a help modal :P

mxmou: Wasn't it meant to be "use this", not "set this"?

WorldLanguages: @mxmou huh, looks like you're right

PR #2207, opened 2021-04-18
Add dark presets to theme3

### LINKED ISSUE #1806: Block dropdowns menus and connected comment outlines having the wrong color
**Describe the bug**

Dropdown and comments connected to a block outlines are the original block colors instead of the modified ones.

**To Reproduce**

Open a project and create a comment connected to a block. Or get a block with dropdowns. With dropdowns get a custom color a bit different than the original. With comments make the custom color very see-able from yellow.

**Expected behavior**

read describe the bug

**Screenshots**

lancer

- OS: Windows 10
- Browser: Google Chrome
- Version: idk

**Additional context**

no

### LINKED ISSUE #2095: Extension block error in developer tools
Blocks from Scratch extensions, such as music and pen, appear with transparent backgrounds in the middle-click pop-up menu in griffpatch's developer tools.

**To Reproduce**

1. Enable griffpatch's developer tools
2. Add an extension (such as music) to your Scratch project
3. Middle-click the code area
4. Search for a block from the extension
5. See error

**Expected behavior**
Extension blocks should have colored backgrounds.

**Screenshots**
Screenshot of bug

- OS: Windows 10
- Browser: Chrome
- Version: 89.0.4389.114 (Chrome)

mxmou:
**Resolves**

Resolves #1806
Resolves #2095

**Changes**

Multiple changes to customizable block colors:
* a new setting to change the background color of block inputs
* a new setting to change the text color to black or colored on black/white background
* two new presets designed to be used with editor dark mode:
  * image
  * image

None of these changes are very important so this can (should) be merged after the release.

**Reason for changes**

https://github.com/ScratchAddons/ScratchAddons/pull/1805#issuecomment-815788261

**Tests**

Tested on Chrome and Firefox.


mxmou:
This contains commits from #1805 because I started working on it before #1805 was merged.

WorldLanguages:
@mxmou Should I wait until this is not a draft to review?

mxmou:
@WorldLanguages I still need to remove the changes I made to the auto refresh code in the userscript while working on this, but the rest is done.

mxmou:
All presets now define all colors, because the previous behavior was confusing.

webpages/settings/index.html
```diff
@@ -61,7 +61,7 @@
           </li>
           <template v-for="preset in addon.presets">
             <li
-              v-if="preset.values.hasOwnProperty(setting.id) && preset.values[setting.id] != setting.default"
+              v-if="preset.values.hasOwnProperty(setting.id) && preset.values[setting.id].toLowerCase() != setting.default.toLowerCase()"
```
WorldLanguages, line 64: Why?

mxmou: Some addons use uppercase and lowercase hex inconsistently, including this one.

WorldLanguages:
@mxmou Do you think we should mark the addon as "new" for v1.14.0 so that people notice the presets? 🤔 
Or do you have any better ideas on how to handle this for the feature
Maybe an "updated addons" group on the settings page, that lets us describe the update?

RedGuy12:
I think we should rename the "New" group to "New and updated"

WorldLanguages:
> I think we should rename the "New" group to "New and updated"

Yeah, not now, but maybe for next release

RedGuy12:
> > I think we should rename the "New" group to "New and updated"
> 
> Yeah, not now, but maybe for next release

but if we did it now, addons like cat blocks, ocular, and this one would fit into the category a bit more

mxmou:
@WorldLanguages If we don't do that now (I think a separate "updated" category is better), I think we should mark theme3 and ocular as new because they have important new features.

ZenithRogue:
@mxmou 
> All presets now define all colors, because the previous behavior was confusing.

The presets don't need to change all the settings. This was an intentional feature of the presets. It allows you to change only groups of the needed settings, while not resetting other user values

mxmou:
@ZenithRogue I think all presets should affect the same group of settings. Previously choosing "2.0 Colors", then "3.0 Tweaks" would use the original 2.0 colors for everything except events, control, and custom blocks, which is likely not what any user wanted. The addition of these new presets made things even more confusing, because switching to "Dark", then "Black" (many people would probably do this when playing with the settings because the two presets were next to each other) made the text almost black on a black background.

I agree that the ability to leave certain settings unchanged when switching presets is sometimes useful - I don't want #2415 to be affected by presets - but I don't think it makes sense in this addon, especially because users can now use the reset button to only load certain colors from a preset.

PR #2211, opened 2021-04-18
Fix waitForElement MutationObserver options

GarboMuffin:
The waitForElement MutationObserver used to specify `attributes: false`, but it seems it was accidentally changed to `attributes: true` in #2061 which is *very bad* for performance, especially in the editor.

This shouldn't cause any addons to misbehave as this was the old behavior.


WorldLanguages:
Nice catch should have checked tbh

PR #2241, opened 2021-04-21
Make editor dark mode not affect the stage

### LINKED ISSUE #2239: [editor-darkmode]: Page Background should not affect the stage monitor
**Describe the bug**

PAGE BACKGROUND, INPUT BACKGROUND, and ACCENT BACKGROUND are affecting variables, lists, and ask inputs in the stage which is should not happen. One reason is that it does not fit the project's scheme of things.

**Expected behavior**

Either other settings or just not be affected at all.

**Additional context**

One person also mentioned to me that "SPRITE LIST BACKGROUND" sounded like it was going to change the background of the lists on the stage. I think it might need to be renamed a bit for clarity.

mxmou:
**Resolves**

Resolves #2239

**Changes**

* Makes ask prompts and stage monitors not affect the stage
* Renames two settings for clarity

**Reason for changes**

Feedback

### DIFF
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -113,7 +113,7 @@
       "default": "#151515"
     },
     {
-      "name": "Sprite list background",
+      "name": "Sprite pane background",
       "id": "selector",
       "type": "color",
       "default": "#202020"
@@ -125,7 +125,7 @@
       "default": "#202020"
     },
     {
-      "name": "Selected list item background",
+      "name": "Selected sprite/asset background",
       "id": "selectorSelection",
       "type": "color",
       "default": "#111111"
--- a/addons/editor-dark-mode/experimental_editor.css
+++ b/addons/editor-dark-mode/experimental_editor.css
@@ -45,20 +45,13 @@
 .gui_body-wrapper_-N0sA,
 .gui_tab_27Unf:not(.gui_is-selected_sHAiu):hover,
 .backpack_backpack-item_hwqzQ > div,
-.monitor_monitor-container_2J9gl,
-.custom-procedures_workspace_1d2uW .blocklySvg,
-.monitor_list-body_2OFZ6 {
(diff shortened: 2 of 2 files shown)


WorldLanguages:
Probably going for v1.13.1?

TheColaber:
> Probably going for v1.13.1?

Nah, this is not mandatory, this is an "enhancement"... but we do end having any more have bugs to fix, you can tag this a long in there.

WorldLanguages:
@mxmou So if I understood this correctly, this means nothing above the canvas/stage should be styled from now? Confirm if I'm correct

Looks like there's a few problems with this PR:
1. It doesn't fully avoid styling stuff above the canvas
2. Some text is now unreadable in editor mode with `dark-www` disabled (note that this is probably indirectly caused by Scratch devs not noticing these were inhereting color from body, it's probably unintentional that they do that)

Project player with `editor-dark-mode` disabled, `dark-www` disabled (AKA vanilla Scratch):
image

Project player with `editor-dark-mode` enabled and `dark-www` disabled:
image
1. Ask prompt has gray border instead of blue
2. Answer button is gray instead of blue
3. Monitor names are white (only in editor/fullscreen mode)

Project player with both `editor-dark-mode` and `dark-www` enabled also has problems 1 and 2 above (but not problem 3):
image

WorldLanguages:
@mxmou Also nevermind about v1.13.1, we will probably directly move to v1.14.0

mxmou:
@WorldLanguages I'll fix the white monitor names. The answer input is intentionally still affected by the highlight color because I don't think that could cause any problems.

WorldLanguages:
> The answer input is intentionally still affected by the highlight color because I don't think that could cause any problems.

Huh, not sure what you mean by problems? It would be sad if the addon didn't style anything above the canvas *except a slight border of the ask and wait modal and the checkbox button* 🤔

PR #2251, opened 2021-04-21
Make cat blocks a non-easter-egg addon

TheColaber:


WorldLanguages:
TODO: still list "cat blocks" in easter egg category, even tho it's not a category

Issue #1972, opened 2021-03-25
Upcoming studio changes


GrahamSH-LLK:
The ST is working on a scratch-www studio page. This will affect any addons that interact with the studios page and possibly other addons, as this will bubble into other pages. This will also help resolve #1967.

mxmou:
We should probably change addons that interact with studios to support both studio pages when the new studio page is complete, and remove code for the old studio page when it's actually released.

jeffalo:
we could also watch studio development and when we believe it's getting close, release an update that removes studio styling in anticipation of the update. not a great solution but it wont leave unused code.

WorldLanguages:
IMO it wouldn't be that complex to support URL matches like `https://scratch.mit.edu/studios_r2/*` and `https://scratch.mit.edu/studios_www/*` if really needed 🤔 (`https://scratch.mit.edu/studios/*` would still match both)

mxmou:
> IMO it wouldn't be that complex to support URL matches like `https://scratch.mit.edu/studios_r2/*` and `https://scratch.mit.edu/studios_www/*` if really needed 🤔 (`https://scratch.mit.edu/studios/*` would still match both)

This would only be need temporary so I think it would be better to just handle it in a userscript.

WorldLanguages:
@mxmou Sure, but RIP theme autoupdate unless dynamicEnable works by then :P
Also, it's always a bad idea to hide the styles until the userscript can run, might cause a flash of unstyled (actually, "unthemed" would be a more acurrate word) content

mxmou:
At the moment manually injecting both scratchr2 and scratch-www styles (from scratchr2 and dark-www addons) into the new studio page works well. Even if this changes, adjusting some selectors will probably be enough to be compatible with both studio pages.
![image](https://user-images.githubusercontent.com/51849865/113404298-38b35e00-93a8-11eb-918c-042aff9df6ea.png)

*(If anyone wants to test it, note that the studio comments endpoint is currently only available to Scratch Team members. I manually redirected it to `https://api.scratch.mit.edu/users/andresmh/projects/104/comments`)*

mxmou:
The studio changes will probably be out soon. Could the next version of SA add the necessary changes?

RedGuy12:
Moved to #2484

PR #2257, opened 2021-04-22
More options for userscript matches

### LINKED ISSUE #23: More options for userscript matches (not only URLs)

TheColaber:
70 changed files ahh

Resolve #23

Please discuss with me what I should add/remove and if I have messed up any addons since I'm sleepy. I'm more specifically focused on dark-www and scratchr3.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -262,6 +262,29 @@ function userscriptMatches(data, scriptOrStyle, addonId) {
 }
 
 function urlMatchesPattern(pattern, url) {
+  switch (pattern) {
+    case "all":
+      pattern = "https://scratch.mit.edu/*";
+      break;
+    case "projects":
+      pattern = "https://scratch.mit.edu/projects/*";
+      break;
+    case "profiles":
+      pattern = "https://scratch.mit.edu/users/*/";
+      break;
+    case "studio":
+      pattern = "https://scratch.mit.edu/studios/*";
+      break;
+    case "forums":
+      pattern = "https://scratch.mit.edu/studios/*";
+      break;
+    case "forums_topic":
+      pattern = "https://scratch.mit.edu/discuss/topic/*";
+      break;
+    case "studio_comments":
+      pattern = "https://scratch.mit.edu/studios/*/comments";
+      break;
+  }
   const patternUrl = new URL(pattern);
   const urlUrl = new URL(url);
   // We assume both URLs start with https://scratch.mit.edu
--- a/addons/2d-color-picker/addon.json
+++ b/addons/2d-color-picker/addon.json
@@ -11,13 +11,13 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["https://scratch.mit.edu/projects/*"]
+      "matches": ["projects"]
     }
   ],
   "userstyles": [
     {
       "url": "style.css",
-      "matches": ["https://scratch.mit.edu/projects/*"]
+      "matches": ["projects"]
     }
   ],
   "libraries": ["tinycolor2"],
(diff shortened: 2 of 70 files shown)


apple502j:
think we can do it better

mxmou:
I don't think everything needs to have named matches - for most pages matching by URL is sufficient. The following would probably be helpful:
* `project`: matches `/projects/*/`, `/projects/*/editor` and `/projects/*/fullscreen` but not `/projects/*/remixes` or `/projects/*/studios`
* `forums`: matches `/discuss/*` but not `/discuss/m/*`
* `403`, `404`, `500`: error pages (other matches, including URL patterns, should never match error pages)
* `scratchr2` and `scrathchwww`: matches by page type
* `studios_scratchr2` and `studios_scratchwww`: see https://github.com/ScratchAddons/ScratchAddons/issues/1972#issuecomment-808743331

RedGuy12:
You’re not in the discord…we’re also going to use theese for settings page filtering in the future.

apple502j:
Closing; I will implement better version this weekend

Hans5958:
...and this is why feature discussions should take on GitHub, not Discord.

mxmou:
> You’re not in the discord…we’re also going to use theese for settings page filtering in the future.

I don't think this is a good idea. Addons should be able to define how they appear on the settings page separately from their functionality.

TheColaber:
> I don't think this is a good idea. Addons should be able to define how they appear on the settings page separately from their functionality.

Hmm, but if you have matches for the forums, why would you also add a tag when the settings page could just look for the forum matches?

mxmou:
> Hmm, but if you have matches for the forums, why would you also add a tag when the settings page could just look for the forum matches?

One reason is that there are addons that don't set their match to the forums but are closely related to them (discuss button). There are also addons that have forum matches to implement forum-specific functionality but aren't otherwise forum addons (infinite scrolling).

Another reason is that if we add subcategories, editor addons will need to set them explicitly anyway. Why would website addons work differently?

Hans5958:
Should we move to #23 for the time being?

TheColaber:
> Closing; I will implement better version this weekend

I think you missed your chance. Reopening till you actually open one.

background/get-userscripts.js
```diff
@@ -262,6 +262,29 @@ function userscriptMatches(data, scriptOrStyle, addonId) {
...
+    case "profiles":
+      pattern = "https://scratch.mit.edu/users/*/";
+      break;
+    case "studio":
+      pattern = "https://scratch.mit.edu/studios/*";
+      break;
+    case "forums":
+      pattern = "https://scratch.mit.edu/studios/*";
+      break;
+    case "forums_topic":
```
GrahamSH-LLK, line 281: Shouldn't these be camelCase?

TheColaber: For what reason?

mxmou: @TheColaber Everything in `addon.json` is currently camel case.

Hans5958:
Why should we do this? I think we shouldn't make matches difficult. Having those URL is already simple, but it seems we're going to make this a little be abstract. Why can't we have simple things and keep it as it is?

mxmou:
> Why should we do this? I think we shouldn't make matches difficult. Having those URL is already simple, but it seems we're going to make this to be a little abstract. Why can't we have simple things and keep it as it is?

I think it's best to only add an alternative where it would be more convenient than matching by URL - https://github.com/ScratchAddons/ScratchAddons/pull/2257#issuecomment-824921416

WorldLanguages:
Closed in favor of #2305

PR #2258, opened 2021-04-22
Componentize the settings page

TheColaber:
**Changes**

As of writing this, this PR converts the addon body and the addon settings into components.

**Reason for changes**

For organizations so functions for just settings or just addons stay in one place rather than with everything else. Also, so we don't have to pass in `addon` or `setting` into functions.

**Tests**

Probably needs more testing just in case.

**Discuss**
What else might need to be componentized?


GrahamSH-LLK:
This is awesome! One thing I want to consider is figuring out how to have separate files for components.

GrahamSH-LLK:
> What else might need to be componentized?

Categories and icons

WorldLanguages:
> One thing I want to consider is figuring out how to have separate files for components.

Keep in mind that a file with an inline `<script></script>` tag won't work because of the extension CSP. We'll need separate .html and .js files if we want to have components as separate files (and if we want to have separated `Vue.extend()`s as well

webpages/settings/index.js
```diff
@@ -474,64 +579,28 @@ const vue = (window.vue = new Vue({
...
-          this.updateOption(property.id, property.default, addon);
-        }
-        console.log(`Loaded default values for ${addon.id}`);
-      }
-    },
     closePickers(e, leaveOpen) {
-      for (let child of this.$children) {
-        if (child.isOpen && child.canCloseOutside && e.isTrusted && child.color && child !== leaveOpen) {
-          child.toggle(child.addon, child.setting, false);
+      for (let addon of this.$children) {
```
WorldLanguages, line 583: Maybe `this.$children.filter(c => c.$options.name === "addon-body")`

WorldLanguages: This is ok for now :P

PR #2263, opened 2021-04-22
RTL support

### LINKED ISSUE #1844: Add RTL support
Our UI doesn't support RTL for languages like Arabic
Add RTL Support
idk
**Additional context**
image

### LINKED ISSUE #1877: Palette is initially visible when auto-hide block palette is enabled
**Describe the bug**

Auto-hide block palette doesn't hide the palette when opening the editor, which doesn't make sense.

**To Reproduce**

2. Open the editor
3. The palette is visible
4. Move your mouse over the palette and back out
5. It's now hidden

**Expected behavior**

It would be initially hidden.

mxmou:
**Resolves**

Resolves #1844
Resolves #1877

**Changes**

Adds support for right-to-left scripts to the settings page, popup and editor addons. Website addons don't need it because the website itself doesn't have proper RTL support. Notes:
* Untranslated strings sometimes display weirdly. Untranslated descriptions on the settings page are truncated incorrectly: the end of the description is visible instead of the beginning.
* Comments in the messaging popup are always left-to-right to make sure that they're displayed the same way as on Scratch.
* Quite a few changes to the `hide-flyout` addon were necessary, so I also fixed a few bugs.

**Reason for changes**

#1844

**Tests**

Tested by changing the browser and Scratch language.


popups/scratch-messaging/popup.css
```diff
@@ -77,27 +78,32 @@ button {
...
   overflow: hidden;
-  margin-right: auto;
-  margin-left: 15px;
+  margin-__MSG_@@bidi_end_edge__: auto;
+  margin-__MSG_@@bidi_start_edge__: 10px;
   max-width: 75%;
 }
 .float-right {
-  margin-left: 6px;
+  margin-__MSG_@@bidi_start_edge__: 6px;
```
WorldLanguages, line 86: Can we use CSS properties like `margin-block-start` that are supported outside of extension contexts, except of this?

WorldLanguages:
@mxmou King 👑 
Was thinking on the bus about this, I arrive home and 2 hours later this PR appears out of nowhere 😂

popups/scratch-messaging/popup.css
mxmou: Done. There is still one instance of `__MSG__` because `float: inline-start` and `float: inline-end` are only supported on Firefox.

WorldLanguages: @mxmou As long as `__MSG__` is only used in non-addon code, it's OK. The main problem of using that in addon context is that TurboWarp would need to do magic to support that. And it also wouldn't work in addon userstyles with `injectAsStyleElt: true`

mxmou: @WorldLanguages It's only used in the messaging popup and the settings page. Addons handle RTL the same way as the Scratch editor itself - using `[dir="rtl"]` selectors.

WorldLanguages: Ohh, I see. Why can't these new CSS properties be used for addons as well? Just to do it the same way as Scratch (so no need to mess with style priority)?

mxmou: > Why can't these new CSS properties be used for addons as well? Just to do it the same way as Scratch (so no need to mess with style priority)?

It's easier to work with if it's done the same way as Scratch. The main advantage of logical properties is that they use the direction of the element itself (or of its parent), but using that of `.gui` wasn't a problem in any editor addon.

apple502j:
What happens when someone enables editor-stage-left?

mxmou:
@apple502j It moves the stage to the right.

WorldLanguages:
> Website addons don't need it because the website itself doesn't have proper RTL support.

Oh wow.

> Untranslated strings sometimes display weirdly. Untranslated descriptions on the settings page are truncated incorrectly: the end of the description is visible instead of the beginning.

We probably don't want to merge this before those are solved - for RTL users, our settings page probably got worse, not better, at the current state of this PR. We don't have complete translations for Arabic or Hebrew yet anyway (this is a bit of a mouse and cat game anyway, I know). So I guess that for the settings page, we'll not only have to support LTR and RTL, but RTL with LTR translation fallbacks as well, which kinda hurts, but is strictly needed, at least for now.

> * Quite a few changes to the `hide-flyout` addon were necessary, so I also fixed a few bugs.

Your changes add new calls to `addon.tab.waitForElement`, so we'll need to coordinate efforts with #2268

WorldLanguages:
> What happens when someone enables editor-stage-left?

@apple502j If you want, this would be a good moment to add support for translator notes inside `addons-l10n`. editor-stage-left doesn't have a file on there (all of its strings are in addon.json), but we could create one that stores the notes reminding Arabic/Hebrew translators that they need to use "right" instead of "left" in both the addon name and description.

mxmou:
Is displaying punctuation like this wrong?
image

WorldLanguages:
> Is displaying punctuation like this wrong?
> image

If it's easy to fix, do it. If not, create an issue and we'll throw it at the backlog :P

mxmou:
@WorldLanguages It could be fixed by adding `dir="auto"` to all elements with this issue, but I'm not sure if that's a very good idea.

WorldLanguages:
@mxmou Why is it a bad idea?

mxmou:
@WorldLanguages When a new element with punctuation is added, it would also need the attribute.

WorldLanguages:
Well, that's not ideal, but what are the alternatives?

mxmou:
> Well, that's not ideal, but what are the alternatives?

I don't know any.

WorldLanguages:
@mxmou 
- If you're ok with this, I'd prefer merging this **but with RTL off for our own webpages**. Arabic and Hebrew are 10% and 0% translated respectively, and it doesn't make sense to ship RTL webpages yet. Of course we can keep the HTML/JS/CSS changes, but the \<body> element should have dir="ltr" for all languages for now (you can hardcore this on `vue.direction` method). Apple and I will try to make a few changes on how translation works soon (stuff has changed since v1.4.0) and specially try to get consistent Arabic/Hebrew translators to make this worth it
- About dir="auto", we should use that in the editor to make sure punctuation isn't messed up. We don't need to care that much about using it in our own webpages for now, because of point above
- Weird scrollbars:
image

addons/hide-flyout/userscript.js
```diff
@@ -118,12 +128,21 @@ export default async function ({ addon, global, console }) {
     // Only append if we don't have "categoryclick" on
     if (toggleSetting === "hover") tabs.appendChild(lockDisplay);
 
+    let blocklyDiv = await addon.tab.waitForElement(".injectionDiv");
```
GarboMuffin, line 131: should be able to just querySelector?

addons/hide-flyout/userscript.js
```diff
@@ -118,12 +128,21 @@ export default async function ({ addon, global, console }) {
...
+
     if (toggleSetting === "hover") {
       placeHolderDiv.onmouseenter = onmouseenter;
       blocklySvg.onmouseenter = onmouseleave;
     }
 
-    if (toggleSetting === "cathover") onmouseleave(null, 0);
+    if (toggleSetting === "cathover") {
+      let categoryMenu = await addon.tab.waitForElement(".blocklyToolboxDiv");
```
GarboMuffin, line 139: should be able to just querySelector?

addons/hide-flyout/userscript.js
```diff
@@ -38,6 +48,10 @@ export default async function ({ addon, global, console }) {
     lockDisplay.classList.add("sa-flyoutClose");
     lockDisplay.style.transitionDuration = `${speed}s`;
     setTimeout(() => Blockly.getMainWorkspace().recordCachedAreas(), speed * 1000);
+    if (toggleSetting === "category") {
+      for (let category of document.querySelectorAll(".scratchCategoryMenuItem.categorySelected"))
```
GarboMuffin, line 52: This will be slow, but it's probably fine

TheColaber: This part can mostly be rewritten to use Blockly's data

RedGuy12:
@WorldLanguages
I disagree with not shipping RTL webpage yet. It can't hurt anything (unless I'm wrong) and it'll just be another step we'd have to remember when we fo implement it.

WorldLanguages:
> It can't hurt anything (unless I'm wrong)

IMO, having 90% or 100% of strings being English, means the layout should be LTR, even tho the browser is in RTL mode.
In other words, IMO, LTR text in LTR mode > LTR text in RTL mode. That's pretty much my point.

mxmou:
> If you're ok with this, I'd prefer merging this **but with RTL off for our own webpages**. Arabic and Hebrew are 10% and 0% translated respectively, and it doesn't make sense to ship RTL webpages yet. Of course we can keep the HTML/JS/CSS changes, but the <body> element should have dir="ltr" for all languages for now (you can hardcore this on `vue.direction` method). Apple and I will try to make a few changes on how translation works soon (stuff has changed since v1.4.0) and specially try to get consistent Arabic/Hebrew translators to make this worth it

Another option would be to use a hardcoded list of languages for which RTL should be used and only put languages that are at least 50% translated there.

> About dir="auto", we should use that in the editor to make sure punctuation isn't messed up. We don't need to care that much about using it in our own webpages for now, because of point above

I didn't notice any punctuation issues in the editor.

> Weird scrollbars

Weird. `overflow: hidden` on the body fixes it, but I don't know what causes this. It doesn't happen in Chrome so it might be a Firefox bug.

WorldLanguages:
> Another option would be to use a hardcoded list of languages for which RTL should be used and only put languages that are at least 50% translated there.

Would work

> I didn't notice any punctuation issues in the editor.

Did you check the devtools help modal?

mxmou:
Direction is now set in `set-lang.js` and is always LTR.

WorldLanguages:
Probably will be merged after v1.14.0 release

WorldLanguages:
@mxmou Can you resolve the conflicts?

mxmou:
@WorldLanguages Note that some CSS added after this PR was made needs to be changed before merging.

WorldLanguages:
@mxmou Could we split hide-flyout changes into another PR? Also, please change the added CSS when you can so that we can create separate files for each Vue component using #2424

mxmou:
> Could we split hide-flyout changes into another PR?

Maybe I could move all changes to hide-flyout, including the ones that add RTL support? They're bigger than changes to other addons.

WorldLanguages:
Yeah you can move all changes to hide-flyout or just some, you decide

WorldLanguages:
@mxmou ?

mxmou:
Why did Prettier make changes to unrelated files?

WorldLanguages:
> Why did Prettier make changes to unrelated files?

It does that sometimes lol ¯\_(ツ)_/¯

mxmou:
I think this PR is now finished. Note that it adds some uses of `inset-inline`, which is only supported in Chrome 87+.

WorldLanguages:
Do not merge yet

WorldLanguages:
> I think this PR is now finished. Note that it adds some uses of `inset-inline`, which is only supported in Chrome 87+.

TODO: is this ok, alternatives? 🤔

mxmou:
> > I think this PR is now finished. Note that it adds some uses of `inset-inline`, which is only supported in Chrome 87+.
> 
> TODO: is this ok, alternatives? 🤔

If it isn't, the best alternative is
```css
.something {
  left: 10px;
}
[dir="rtl"] .something {
  left: 0;
  right: 10px;
}
```

webpages/settings/style.css
```diff
@@ -144,30 +144,23 @@ h1 {
 }
 
 .clear-button {
-  border-radius: 0 4px 4px 0px;
-  border-left: none !important;
+  border-start-radius: 0;
```
WorldLanguages, line 147: I think this line can be removed. CSS property is invalid and all 4 corner radiuses are set below

WorldLanguages:
We currently want core to be functional in Chrome 80+ and Firefox 74+ (both released ~1 year from now)

The CSS properties I found used here that are not supported in those browsers are:
- `inset-inline-*`: Chrome 87+, Firefox OK
- `border-[start/end]-[start/end]-radius`: Chrome 89+, Firefox OK

I guess a good compromise we could do is officially support RTL only for Chrome 89+
And in order to support Chrome 80+, we could continue to set mapped physical properties (example: `border-top-left-radius`) but above the new better properties (example: `border-start-start-radius`), so that everything looks the same in Chrome < 88 for LTR users. 

For example:
```css
el {
  inset-inline-end: 25px;
}
```
would become:
```css
el {
  left: 25px;
  inset-inline-end: 25px;
}
```
Which would now support LTR for Chrome 80+ and all Firefox versions, as well as RTL in Chrome 89+ and all Firefox versions.

@mxmou What do you think? It will be easy to migrate when our minimum browser version requirements change (all we need to do is remove properties) and there's no need to do `[dir="rtl"]` selectors

mxmou:
@WorldLanguages This would work, but I'm not sure if having different version requirements for LTR and RTL layouts makes sense. If we want core to work in a certain browser version, it should probably work with that version regardless of the language.

WorldLanguages:
@mxmou What do you suggest then? 🤔 We haven't supported RTL until now, so it feels like a good compromise to avoid separate `[dir="rtl"]` selectors (which will be able to be removed in a few months anyway) by only making RTL available in newer Chrome versions, which most of our users have anyway
I considered using `__MSG_@@bidi_start_edge__` and `__MSG_@@bidi_end_edge__` in the middle of the name of CSS properties, which would probably work, but it would use the browser configuration, not the dir attribute of the root element we can determine on runtime.

mxmou:
@WorldLanguages I guess it would work, I just don't think there's anything wrong with adding `[dir="rtl"]` selectors and removing them later.

WorldLanguages:
Then do that I guess :P

webpages/settings/style.css
mxmou: Removed.

WorldLanguages:
Solve conflicts and merge?

RedGuy12:
> * If you're ok with this, I'd prefer merging this **but with RTL off for our own webpages**. Arabic and Hebrew are 10% and 0% translated respectively, and it doesn't make sense to ship RTL webpages yet. Of course we can keep the HTML/JS/CSS changes, but the <body> element should have dir="ltr" for all languages for now (you can hardcore this on `vue.direction` method). Apple and I will try to make a few changes on how translation works soon (stuff has changed since v1.4.0) and specially try to get consistent Arabic/Hebrew translators to make this worth it

@WorldLanguages 17 months later, Arabic and Hebrew are 54% and 46% translated respectively. I think we forgot to enable this :P

WorldLanguages:
@RedGuy12 Those languages didn't hit the necessary threshold for them to be included. See https://github.com/ScratchAddons/ScratchAddons/tree/master/_locales

But we should pay close attention, they might appear soon automatically

PR #2265, opened 2021-04-22
block-palette-icons: fix list icon

GarboMuffin:
Fixes a regression from #2159 that made the list icon look a bit strange

firefox_rEqS8YQgGi


GarboMuffin:
Some parts of the path got changed too, probably on accident.

PR #2268, opened 2021-04-23
Use the new waitForElement spec and remove unnecesary waitForElement calls to improve performance

GarboMuffin:
- Makes a lot of addons use the new waitForElement spec (#2182) to reduce the amount of querySelectorAlls that are run in response to any DOM change. This can significantly reduce Scratch Addons' impact on performance in large projects.
 - Removes unnecessary waitForElement usages

I did a test where I dragged some blocks around in the player sprite of https://scratch.mit.edu/projects/60917032/editor/ with a bunch of addons enabled:

Before: 30+ querySelectorAll / 134ms per DOM change (https://share.firefox.dev/3ayMSgu)
After: 2 querySelectorAll / 10ms per DOM change (https://share.firefox.dev/2QHjgXx)

The difference is immediately noticeable. The remaining two querySelectorAlls are in hide-flyout (only when mode is category click or category hover) and addon.tab.scratchClass.

The likelihood of some bugs sneaking into this PR is **very high**. Consider testing it out and especially making sure that addons work after doing things like entering/leaving the editor, fullscreen, etc.

### DIFF
--- a/addons/discuss-button/userscript.js
+++ b/addons/discuss-button/userscript.js
@@ -15,7 +15,7 @@ export default async function ({ addon, global, console }) {
         markAsSeen: true,
         condition: () => {
           if (!addon.tab.redux.state) return false;
-          if (!addon.tab.redux.state.scratchGui) return false;
+          if (!addon.tab.redux.state.scratchGui) return true;
           return addon.tab.redux.state.scratchGui.mode.isPlayerOnly;
         },
       });
--- a/addons/discuss-button/userscript.js
+++ b/addons/discuss-button/userscript.js
@@ -11,7 +11,14 @@ export default async function ({ addon, global, console }) {
 
   if (addon.tab.clientVersion === "scratch-www") {
     while (true) {
-      const el = await addon.tab.waitForElement("div#navigation div.inner ul:not(.production)", { markAsSeen: true });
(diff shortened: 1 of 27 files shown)


addons/discuss-button/userscript.js
```diff
@@ -11,7 +11,14 @@ export default async function ({ addon, global, console }) {
 
   if (addon.tab.clientVersion === "scratch-www") {
     while (true) {
-      const el = await addon.tab.waitForElement("div#navigation div.inner ul:not(.production)", { markAsSeen: true });
+      const el = await addon.tab.waitForElement("div#navigation div.inner ul:not(.production)", {
+        markAsSeen: true,
+        condition: () => {
+          if (!addon.tab.redux.state) return false;
+          if (!addon.tab.redux.state.scratchGui) return false;
+          return addon.tab.redux.state.scratchGui.mode.isPlayerOnly;
```
WorldLanguages, line 19: You may use `?.` if wanted

GarboMuffin: The code diff here has a logical error in it. Once fixed ?. probably won't work

WorldLanguages (approved):
Epic job. Couldn't get anything to break, but even if some little bug sneaks in, this is a clear net positive

PR #2278, opened 2021-04-24
Update all settings at once when loading presets

### LINKED ISSUE #2177: Update all settings at once when setting presets
Right now, every setting is updated individually, causing several writes to `chrome.storage` and many "change" events on `addon.settings`.
Instead, they should all be set at once. Addons don't have any way to listen for *single* setting changes so that's not a problem.
This also causes a weird cascade effect when setting a preset in a theme and seeing it update realtime, you can see it load step by step:
Mb7l4MJC8U

mxmou:
**Resolves**

Resolves #2177

**Changes**

Changes preset loading to update all settings at once.


WorldLanguages:
We should update all at once when resetting all to default
And fix some console logs that broke after #2258:
image

TheColaber (approved):

Issue #1146, opened 2020-12-31
Improvements on the update notification modal


Hans5958:
There is this new thing where there will be a notification modal/banner that shows every update. The modal is shown below.

![image](https://user-images.githubusercontent.com/11584103/103412730-22c7ae80-4ba9-11eb-97bf-666d65f38067.png)

Unfortunately, I have comments with the design and the system, and I feel that this is done hastily, I think this can be improved for the sake of the users. 

Firstly, I wanted to introduce the prototype that I made to help me explaining the points that I am going to talk soon.

![image](https://user-images.githubusercontent.com/11584103/103412366-5acdf200-4ba7-11eb-9ef6-93f7975dd9cd.png)
https://codepen.io/Hans5958/pen/oNzpxNy

You may see that this resembles the design of Windows window design. **Thought this design can improved, that's not the point I want to discuss mainly**, but I have also made multiple examples with different color schemes, shown below.

On the top portion of this modal, there is the title bar. It contains the title of the modal and the close button, like the one's that we have. I also added the logo as a decoration, but that can be removed if you really don't want to copy the Windows design.

Below the title bar is the where the content is. I divided it into three parts: the intro, the highlights, and the footer. Here's the explanation of those three parts.

1. The intro contains, well, the introduction/summary of the modal. It just says that it is updated into this certain version. Pretty weird why this didn't included, and the version number is just included below, as a part of a link. This also works as the only content if there are no highlights. The reason of why this is a point will be explained below.
2. The highlights contains the feature/addon that wanted to be highlighted. It contains the image part and the text part, but can be adjusted for various updates. The elements are aligned center so that the text is aligned in the center when the content is short, which examples will be provided below. The image can also be excluded, just in case there are none.
3. The footer contains relevant links (and texts maybe) for the notification. It only contains links with short phrases, because I think it is unnecessary to make it long. I didn't include the disclaimer because, well, do we really need to warn them like hundreds of times because they might didn't understand about the implications?

Now that I have explained the structure of my idea of the modal, I will continue giving suggestions for the mechanism and some other aspects of it.

1. The notification should be shown for every single update, including patch versions.
I have no idea why we don't tell the users about the patch updates. What if there is a significant and popular bug that we fixed in a patch version, and we didn't tell them? That would be sad.
What about the highlights? We can just hide the image if it is only text, or even remove it completely and just show the intro and the footer. I have also put this into consideration on the design, which is shown on the example below.

[...]

PR #2279, opened 2021-04-24
Improve update notification

mxmou:
**Resolves**

None

**Changes**

Changes to the update notification:
* Changes background color and close button to match settings page design
* Adds a new string instead of using CSS to make the changelog link title case (currently it's capitalized incorrectly in some languages, including Spanish, French, German and Slovenian)


WorldLanguages:
Epic, but have you seen #1146? This might be a good opportunity to get closer to one of the designs Hans suggested. I'm not sure if they are compatible with our typical design choices, tho

mxmou:
@WorldLanguages I think most ideas in #1146 are good, but not the visual appearance. It's different from both our UIs and Scratch.

PR #2286, opened 2021-04-24
Add remove stage borders addon

### LINKED ISSUE #2277: Remove curved stage borders
Right now in Scratch, you can't see the very corners of the stage because of a beveled border. It can be useful to be able to see these areas of the stage. This plugin would remove the curved border of the stage, allowing you to see the whole stage.

I have made this plugin to do this so I could use it myself but don't know if this deserves it's own plugin or not, so I haven't created a pull request.
https://github.com/Tacodude7729/ScratchAddons/tree/master/addons/remove-curved-stage-border

Before
image

After
image

Tacodude7729:
Resolves #2277 

Should be compatible with everything.


addons/remove-curved-stage-border/addon.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "Remove Curved Stage Border",
+    "description": "Removes the curved border around the stage, allowing you to see the very corners.",
+    "tags": [
+        "editor"
```
RedGuy12, line 5: add `theme` tag

addons/remove-curved-stage-border/addon.json
```diff
@@ -0,0 +1,25 @@
...
+    ],
+    "enabledByDefault": false,
+    "userstyles": [
+        {
+            "url": "remove-borders.css",
+            "matches": [
+                "https://scratch.mit.edu/projects/*/",
+                "https://scratch.mit.edu/projects/*/editor",
+                "https://scratch.mit.edu/projects/*/fullscreen",
+                "https://scratch.mit.edu/projects/editor"
```
RedGuy12, line 21: just do `https://scratch.mit.edu/projects/*`, the rest will be automatic

RedGuy12 (changes requested):

RedGuy12:
also please enable prettier here: https://github.com/Tacodude7729/ScratchAddons/actions

Tacodude7729:
@RedGuy12 The format code using Prettier task says it errored with the following log, is this a problem?
```
Checking formatting...
[warn] addons/remove-curved-stage-border/addon.json
[warn] addons/remove-curved-stage-border/remove-borders.css
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
Error: Process completed with exit code 1.
```

RedGuy12:
No, that's normal.

addons/remove-curved-stage-border/addon.json
mxmou: > just do `https://scratch.mit.edu/projects/*`, the rest will be automatic

This will also match `https://scratch.mit.edu/projects/*/remixes` and `https://scratch.mit.edu/*/studios`, but in this case that isn't a problem.

addons/remove-curved-stage-border/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Remove Curved Stage Border",
+  "description": "Removes the curved border around the stage, allowing you to see the very corners.",
```
RedGuy12, line 3:
```suggestion
  "description": "Removes the curved border around the stage, allowing you to see into the corners.",
```

addons/remove-curved-stage-border/addon.json
```diff
@@ -0,0 +1,18 @@
...
+      "link": "https://github.com/Tacodude7729"
+    }
+  ],
+  "enabledByDefault": false,
+  "userstyles": [
+    {
+      "url": "remove-borders.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ]
```
RedGuy12, line 17: also please add `l10n: true`
```suggestion
  ],
  "l10n": true
```

addons/remove-curved-stage-border/addon.json
Hans5958: "the very corners" is valid. What did you see in the corner? We want to see the right-angle corners, not something inside of them.

RedGuy12: "the very corners" just sounds odd imo

addons/remove-curved-stage-border/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Remove Curved Stage Border",
```
Hans5958, line 2: Do not use title caps. Use "Remove curved stage border."

addons/remove-curved-stage-border/addon.json
Hans5958: If you think it is, then "the corners" is enough, "into the corners" would be weirder/misleading. Otherwise, it is fine.

TheColaber:
1. Add dynamicEnable and dynamicDisable
2. This does not support TW; You need to change the css selectors to smth like `[class*=__]`

WorldLanguages:
> Add dynamicEnable and dynamicDisable

Smh it's not mandatory

TheColaber:
> Smh it's not mandatory

ofc not but *why* not

Hans5958:
> > Smh it's not mandatory
> 
> ofc not but _why_ not

That could wait tommorow

Tacodude7729:
>     2. This does not support TW; You need to change the css selectors to smth like `[class*=__]`

@TheColaber What do you mean by not supporting turbo warp? On turbowarp.org the class for the wrapper with the border is still `stage_stage_1fD7k`?

RedGuy12:
@Tacodude7729
The '_1fD7k' part changes with every update if scratch and I don't think TW has it at all. So instead of using a class selector,  do [class*=stage_stage]. Same for the other selector.

(Since that has a lower specificity, you may need to add 'injectAsStyleElt:true' to your json and/or use !important)

mxmou:
> @Tacodude7729 The '_1fD7k' part changes with every update if scratch and I don't think TW has it at all. So instead of using a class selector, do [class*=stage_stage]. Same for the other selector. (Since that has a lower specificity, you may need to add 'injectAsStyleElt:true' to your json and/or use !important)

The hash never changes unless the CSS file in which the class is defined is moved to a different location. The reason why we avoid hardcoding them is that they're different in TurboWarp Desktop.

Attribute selectors have the same specificity as class selectors.

RedGuy12 (approved):

astrosticks:
Thanks so much!

PR #2288, opened 2021-04-25
Remove unnecessary typecasting in onion skinning

pufferfish101007:
**Resolves**

none

**Changes**

Removes unnecessary typecasting from onion skinning

**Reason for changes**

#2142 means these typecasts are not needed, and no me gusta nada unneeded code.

**Tests**

Works in chrome


apple502j:
Eh, doesn't cause harm, let's keep it

WorldLanguages:
@apple502j Why? Number settings are guaranteed to return numbers now anyway

PR #2292, opened 2021-04-25
Support customCssVariables using text_color library

### LINKED ISSUE #2099: Addons should be able to change how settings are exposed to CSS

Currently settings are provided to CSS as custom properties. If any other values (typically colors) are needed by a userstyle, they need to be set by a userscript (like I did in #1805) which means that users have to wait for the correct colors. This can be a problem for website dark mode where users frequently navigate between pages.

A new `addon.json` property with the following syntax:
```json
{
  "cssVariables": [
    {
      "name": "setting1",
      "value": "setting1"
    },
    {
      "name": "setting1-text",
      "value": {
        "type": "textColor",
        "source": "setting1",
        "black": "rgba(0, 0, 0, 0.5)",
        "white": "rgba(255, 255, 255, 0.5)"
      }
    }
  ]
}
```

`"name"` would be the name of the CSS variable. `"value"` would be a setting name or an object with a `"type"` property describing a color operation and additional properties depending on the type. Most types would have a `"source"` property which could be a setting name or another such object. Possible values for the `"type"` could be:
* `"textColor"`: automatically selects an appropriate text color (or border color, image filter etc.) based on the source. `"black"` would be the value that is used for light background colors (the default could be `#575e75`) and `"white"` the value for dark backgrounds (defaults to `#ffffff`). An optional property called `"threshold"` would be used to specify how bright/dark the background color needs to be to be considered bright/dark.
* `"multiply"`: makes a source color darker or more transparent. Additional properties would define the coefficients for red, green, blue and alpha valules.
* `"brighten"`: similar to the above, but makes the source brighter or less transparent. Would do the same thing as line 38 of `text_color.js` in #1805. (Probably rarely useful but exactly what I need for generating the secondary color in editor dark mode.)
* `"alphaBlend"`: would have two source colors, the first one would need to be opaque, the second one could be transparent. Would return an opaque color equivalent to drawing the second source on top of the first one.

If this replaces the implicitly created variables it would also reduce the amount of them.

This could be extended to work with other setting types, not just colors.

WorldLanguages:
Resolves #2099 

@mxmou TL;DR is:
1. Your goal is achieved, colors are loaded way faster
2. I modified a single file of scratchr2 to test it. It's up to you to decide whether you want to move to this inside this PR or create another one. I'd migrate myself but I'm not sure I wouldn't mess up.
3. Feel free to add more custom variable types. You mentioned `alphaBend` but I didn't see it in the library.
4. There's still no way to opt out of getting string/number/color variables as CSS variables in the page, but now variables are only added for active userstyles, heavily decreasing the amount of added CSS variables.


WorldLanguages:
Schema will need to be updated. @mxmou please confirm which properties are optional and which aren't

GrahamSH-LLK:
Can this PR remove the weird system of putting every SINGLE setting (and past setting) into the dom??

RedGuy12:
@GrahamSH-LLK

I assume you mean only make css custom properties for addons that explicitly request it (probably via a new key in the addon manifest)? I agree, the long list of settings can lengthen the body tag unnecessarily. There's not much we're can do about the 20 dark mode settings, but we can eliminate SOME unused ones.

WorldLanguages:
> Can this PR remove the weird system of putting every SINGLE enabled setting (and past setting) into the dom??

That's already the case. This PR only adds variables for addons that have userstyles on the page. This is infinitely times better than how it worked until now.

mxmou:
I think it's best to change editor dark mode to use this later and other operations when needed.

The structure of `addon.json` properties currently looks like this:
* `customCSSVariables` (optional, defaults to `[]`)
   * `name` (required)
   * `value` (required)
      * `type` (required). Possible values:
         * `"settingValue"`
            * `settingId` (required)
         * `"textColor"`
            * `black` (required)
            * `white` (required)
            * `source` (required)
         * `"multiply"` or `"brighten"`
            * `r` (optional, defaults to 1)
            * `g` (optional, defaults to 1)
            * `b` (optional, defaults to 1)
            * `a` (optional, defaults to 1)
            * `source` (required)

GrahamSH-LLK:
> > Can this PR remove the weird system of putting every SINGLE enabled setting (and past setting) into the dom??
> 
> That's already the case. This PR only adds variables for addons that have userstyles on the page. This is infinitely times better than how it worked until now.
Can you make it enabled addons only?

GrahamSH-LLK:
> > Can this PR remove the weird system of putting every SINGLE enabled setting (and past setting) into the dom??
> 
> That's already the case. This PR only adds variables for addons that have userstyles on the page. This is infinitely times better than how it worked until now.
Can you make it enabled addons only?

mxmou:
> Can you make it enabled addons only?

It already is enabled addons only.

WorldLanguages:
> Can you make it enabled addons only?

Do I need to explain disabled addons can't have active userstyles on the page...? :P

WorldLanguages:
@mxmou Confirm whether this can be merged as-is

mxmou:
@WorldLanguages I think it can be merged.

PR #2296, opened 2021-04-26
Remove sourceMap from Comlink

GrahamSH-LLK:
Since the sourceMap doesn't exist at this location (it's not even in the repo!), this is 100% unneccesary.
It also removes a warning because there's no missing source map.


WorldLanguages:
Don't we want to use the library as-is? We'd be modifying it by doing this

TheColaber:
Didn't you convert it from ts to js anyway?

GrahamSH-LLK:
> Didn't you convert it from ts to js anyway?

No, that was comlink-extension.

> Don't we want to use the library as-is? We'd be modifying it by doing this

Yes, BUT this adds a warning in the console whenever SA is enabled.

WorldLanguages:
> Yes, BUT this adds a warning in the console whenever SA is enabled.

Then let's add an empty source map file so that it won't complain :P

GrahamSH-LLK:
> > Yes, BUT this adds a warning in the console whenever SA is enabled.
> 
> Then let's add an empty source map file so that it won't complain :P

I'm just gonna add the actual one.

RedGuy12:
> > > Yes, BUT this adds a warning in the console whenever SA is enabled.
> > 
> > 
> > Then let's add an empty source map file so that it won't complain :P
> 
> I'm just gonna add the actual one.

that references other files that aren't in the repo tho...wouldn't devtools complain about that

WorldLanguages:
Exactly, at that point an empty source map would be better

GrahamSH-LLK:
> Exactly, at that point an empty source map would be better

Yes, but I bet devtools will complain about that!

apple502j:
People need to learn to ignore warnings.

WorldLanguages:
> Yes, but I bet devtools will complain about that!

Have you tested if it does...?

PR #2305, opened 2021-04-27
Add improved userscript matching

### LINKED ISSUE #23: More options for userscript matches (not only URLs)

### LINKED ISSUE #646: https://scratch.mit.edu/ at the beginning of match patterns should be optional

All userstyle and userscript matches start with https://scratch.mit.edu/.

That should be optional.

apple502j:
Resolves #23 
Resolves #646
Closes #2257 

New syntax is:

- `*`: allow any URL on Scratch origin.
- string starting with `^https:`: treat as regex that matches origin+pathname, aka absolute regex
- string starting with `^` but not `^https:`: treat as regex that matches pathname, aka relative regex
- Array of:
  - absolute or relative regexes (mentioned above)
  - shortcut: `projects` (excl. embeds), `projectEmbeds`, `studios`, `studioComments`, `profiles`, `topics`, `newPostScreens` (topics + new topic), `editingScreens` (newPostScreens + signature setting), `forums` (excl. mobile forums)), `scratchWWWNoProject` (all scratch-www pages, except project page and embeds)
  - shortcut matchers: `isNotScratchWWW`, which can be used like "is scratchr2?"
  - legacy match (aka current match) is kept untouched

NOTE: To use origin-less URL, regex must be used (not current glob-style match.) Legacy matching style should no longer be used.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -259,7 +259,7 @@ const WELL_KNOWN_PATTERNS = {
   profiles: /^\/users\/[\w-]+\/?$/,
   topics: /^\/discuss\/topic\/\d+\/?$/,
   newPostScreens: /^\/discuss\/(?:topic\/\d+|\d+\/topic\/add)\/?$/,
-  editingScreens: /^\/discuss\/(?:topic\/\d+|\d+\/topic\/add|settings\/[\w-]+)\/?$/,
+  editingScreens: /^\/discuss\/(?:topic\/\d+|\d+\/topic\/add|post\/\d+\/edit|settings\/[\w-]+)\/?$/,
   forums: /^\/discuss(?!\/m(?:$|\/))(?:\/.*)?$/,
 };
 
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -261,6 +261,14 @@ const WELL_KNOWN_PATTERNS = {
   newPostScreens: /^\/discuss\/(?:topic\/\d+|\d+\/topic\/add)\/?$/,
   editingScreens: /^\/discuss\/(?:topic\/\d+|\d+\/topic\/add|post\/\d+\/edit|settings\/[\w-]+)\/?$/,
   forums: /^\/discuss(?!\/m(?:$|\/))(?:\/.*)?$/,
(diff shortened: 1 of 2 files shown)


mxmou:
Are `projectEmbeds`, `studios`, `studioComments`, `profiles` and `topics` really necessary? See https://github.com/ScratchAddons/ScratchAddons/pull/2257#issuecomment-824921416

Could you also add a way to match by page type (scratchr2, scratch-www)? This would solve #1976.

mxmou:
The reason why we need this is not to make matches shorter. It's to prevent issues where the page is something else than what is usually on that URL (e.g. a 404).

apple502j:
@mxmou Currently we cannot check if it is 404 - the script does not have access to status code, DOM, etc; only the URL. We can prevent stuff from loading if the URL itself is invalid (`/projects/project-id-is-not-integer/`), and that's what the regexes do.

mxmou:
@apple502j If so, the way matching works should be changed. Otherwise this doesn't fully solve the problem.

apple502j:
@mxmou This does not fully solve the problem - yeah, i know, but detecting 404 error is not easy. scratchr2 404 can be detected from content script, but scratch-www 404 itself is client-side rendered and we don't know when they will be.

The problem this PR solves is mainly these two:
- Duplications between addon.json files
- Handling of URLs that are obviously invalid

mxmou:
@apple502j Detecting scratchr2 404s is enough to solve #1976.

Does this fix #646 too?

apple502j:
@mxmou #646 is fixed if regex is used. Legacy behavior is kept untouched.

This PR's purpose is not to fix #1976; fixing 1976 requires major code refactoring and needs to be in a separate PR, since it breaks the assumption that CS knows only runnable addons, not addons that may or may not run.

WorldLanguages:
@mxmou I moved #646 to the v1.15.0 milestone, I'll work on it
Having 404 matches properly is hard but not impossible. And it will be easier to do above this PR.

WorldLanguages (approved):
Good. If someone wants to discuss the name/regex expressions for the word matches this PR adds before this gets merged, feel free to do so

apple502j:
Working on "scratch-www pages" regex.

PR #2260, opened 2021-04-22
Settings page: add addon groups and subcategories

TheColaber:
Resolves #2218 
Resolves #2281
Resolves #2303
Resolves #1505
TODOs
- [x] decide how to order each group
- [x] handle search with groups
- [x] handle hidden addons with groups (when searching)
- [x] css
- [x] not sure why everything has "for website" or "for editor" visible tag
- [x] handle small settings page in popup
- [x] bug: cannot close color pickers
- [x] disabled easter egg addons
- [x] "hidden" tag (same group as beta)
- [x] expanded should be a prop, not an addon prop
- [x] bug: beta group acts weird since it's default disabled
- [x] remove filters
- [x] subcategories
- [x] more tags and categories
- [x] location.hash with groups
- [x] different categories in iframe mode
- [x] Order when searching?
- [x] Consistent tag ordering
- [x] Delete "remove on v1.14.0" code
- [x] Fix being able to open color pickers when addon disabled
- [x] what to do with the emptyness when opening the popup

PR #2326, opened 2021-04-30
Tag all addons

TheColaber:
(this is being written by @WorldLanguages)
- Removed `beta` from some addons
- Added `beta` tag to feature-unshared, should be tested more deeply
- Slightly changed the definition of what a "recommended" addon is: now, themes like dark mode and editor-stage-left are recommended. true-youtube-links isn't recommended anymore, because most users don't even know Scratch has a YouTube pseudoplayer.
- Used tags for first time: `danger`, `codeEditor`, `costumeEditor`, `projectPlayer`, `editorMenuBar`, `projectPage`, `profiles`, `studios`, `comments`, `popup`
- Make sure all addons have either `editor`, `community` or `popup`
- Add `"versionAdded": "1.14.0"` to new addons, instead of hardcoding on settings page. This property can be added later to all addons.
- Make sure enabled by default addons and recommended addons don't ping servers by default. Only change was making exact-count not set exact number of posts on the forums by default (this only affects the default setting value for new users)


WorldLanguages:
TODOs:
- Popup should be a tag
- Make sure ALL addons have either: community, editor or popup
- Write spec for updated schema

WorldLanguages:
TODO: rename `editorNavbar` to `editorMenuBar` (cc @mxmou, re: https://github.com/ScratchAddons/ScratchAddons/pull/2260#discussion_r624512948)

PR #2349, opened 2021-05-02
Lint 20210503

apple502j:


addons/gamepad/gamepadlib.js
```diff
@@ -561,7 +561,7 @@ class GamepadEditor {
       return;
     }
     this.selector.hidden = false;
-    for (const [id, gamepadData] of gamepads) {
+    for (const [id, _] of gamepads) {
```
RedGuy12, line 564: if it's not needed, why even destructure it?

PR #2359, opened 2021-05-02
Fix missing "of"

apple502j:


WorldLanguages:
AAAAAA we already discussed this

mxmou:
I don't think "new addons of this update" is correct.

mxmou:
I would use "*in* this update".

WorldLanguages:
> I don't think "new addons of this update" is correct.

IMO, it's correct but it sounds weird. Grammatically, it's no different than "new features of this update" which is an exact phrase that returns over 250,000 results on Google

> I would use "_in_ this update".

Too late now :P

PR #2370, opened 2021-05-04
Color picker fixes

GrahamSH-LLK:
Upgrades TinyColor and lit-html.
The upgraded lit-html seems to have really improved performance!


GrahamSH-LLK:
@WorldLanguages This is tested and should work well.

PR #2371, opened 2021-05-04
Fix grammar

GrahamSH-LLK:
Fixes grammar in editor-devtools string


TheColaber:
Grammarly thinks `Right-click a block, and three new options are available to Copy All, Copy Block, and Cut Block. The Copy All will copy to the clipboard everything, including and below the block you clicked on. Copy block will only copy the current block and its contents, but nothing below. And cut block will copy it and remove it from the workspace.` is the best grammar. So the "and" in there is fine and is proper grammar.

PR #2372, opened 2021-05-04
Fix lags on some profiles

Sly-Little-Fox:
Agggh fixing conflicts idk


RedGuy12:
These don't work. Both times, instead of
```js
  setTimeout(() => {
```
use 
```js
  setTimeout(async () => {
```

Sly-Little-Fox:
> These don't work. Both times, instead of
> 
> ```js
>   setTimeout(() => {
> ```
> 
> use
> 
> ```js
>   setTimeout(async () => {
> ```
> 
> (that should also fix prettier failing)

Oh, yes. Also, I thought I was sending a PR in my repository.

PR #2315, opened 2021-04-28
Close color picker after reclicking input (settings page)

GrahamSH-LLK:
Resolves #2201

**Changes**

Changes code
**Reason for changes**

Because
**Tests**

No

PR #2374, opened 2021-05-04
clicking on color picker closes it

TheColaber:
temp fix: always open color picker, when it's button, is pressed (which has the color picker in it


mxmou:
I think you reverted an intentional change (#2315).

TheColaber:
@GrahamSH-LLK Test your PRs

PR #2385, opened 2021-05-05
Fix reduxEvents regressions

GarboMuffin:
- maybe fix race condition on page load (please test)
 - fix addons not running in response to language change

This API really needs to be cleaned up to get rid of all this code repetition.


TheColaber:
Could `reduxEvents: ["scratch-gui/mode/SET_PLAYER", "fontsLoaded/SET_FONTS_LOADED", "scratch-gui/locales/SELECT_LOCALE"]` be a property, instead of many addons repeating this code?

apple502j:
Merging as this PR should not introduce bugs. It might not fix it but that's fine.

PR #2408, opened 2021-05-06
Fix ocular reaction wrapping

mxmou:
**Resolves**

https://scratch.mit.edu/discuss/post/5199993/

**Changes**

Reactions are no longer overlapped or split into two when there are too many of them to fit on a single line.

**Tests**

Tested with this post: https://scratch.mit.edu/discuss/post/5199335/

Before:
image

Now:
image

### DIFF
--- a/addons/my-ocular/reactions.css
+++ b/addons/my-ocular/reactions.css
@@ -1,5 +1,6 @@
-.postfootright {
-  padding-bottom: 4px; /* so that the buttons have some space */
+.postfootright ul {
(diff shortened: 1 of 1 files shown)


jeffalo:
you read my mind, i was about to do this when i got home (same issue on ocular itself)

jeffalo (approved):
looks good. just wondering, was adding just inline-block not enough?

mxmou:
@jeffalo Adding `inline-block` fixes both issues, but the result looks weird - the buttons don't overlap, but they still touch each other.

PR #2416, opened 2021-05-06
Match userscripts/styles to scratch.mit.edu/[statusCode] if not 2xx

### LINKED ISSUE #1976: 404 not converted

WorldLanguages:
This is an easy implementation of #1976 (Resolves #1976)
If the status code of the page is not 2xx, instead of userscripts and userstyles being matched to the real url (e.g. `scratch.mit.edu/nope`) they are matched against slash + status code (`scratch.mit.edu/404`).

Some details about this implementation:
- Does not work on iframes to avoid complexity
- Addons aren't able to target specific URLs if they don't return 2xx. Not sure why an addon would want to do that, tho
- This implementation does not try its best to run the userscripts and userstyles as fast as if the page returned a 2xx status code, not because we need to wait to get the status code, but because **the content script is in charge** of requesting contentScriptInfo again if the original getContentScriptInfo request returned, along with other data, that the status code of the page was not 2xx. We could make the background page be in charge of this, but it would make the whole matching process more complex, since we don't know what status code the page will return in the moment we cache (onBeforeRequest). **So we're compromising a bit of potential flashing if the user has dark mode or 2.0→3.0, in exchange of way simpler logic and code.** The csInfoCache logic remains mostly unchanged, and we don't have to match against every potential status code. We don't have to maintain a cache of every potential status code and make sure to keep it up to date if addon settings changes. either
- dynamicEnable matches against the status code URL, not the real one

Some global details, independent of implementation:
- There are situations where we won't know the status code of the page. In these cases, we'll assume it returned 2xx
- I'd be OK changing the matching name to simply `404`
- I'd be OK restricting the status codes that can be matched against (403, 404, 503)
- scratch.mit.edu/500 returns 503 status :P

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -221,7 +221,7 @@ chrome.webRequest.onResponseStarted.addListener(
   },
   {
     urls: ["https://scratch.mit.edu/*"],
-    types: ["main_frame", "sub_frame"],
+    types: ["main_frame"],
   }
 );
 
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -130,6 +130,7 @@ async function getAddonData({ addonId, manifest, url }) {
 async function getContentScriptInfo(url) {
   const data = {
     url,
+    httpStatusCode: null, // Set by webRequest onResponseStarted listener
     l10njson: getL10NURLs(),
     globalState: {},
     addonsWithUserscripts: [],
@@ -209,6 +210,20 @@ chrome.alarms.onAlarm.addListener((alarm) => {
     });
   }
 });
+
+chrome.webRequest.onResponseStarted.addListener(
+  (request) => {
+    const identity = createCsIdentity({ tabId: request.tabId, frameId: request.frameId, url: request.url });
+    const cacheEntry = csInfoCache.get(identity);
+    if (cacheEntry && cacheEntry.loading === false) {
+      cacheEntry.info.httpStatusCode = request.statusCode;
+    }
+  },
+  {
+    urls: ["https://scratch.mit.edu/*"],
+    types: ["main_frame", "sub_frame"],
+  }
+);
 
 chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
   if (!request.contentScriptReady) return;
--- a/addons/scratchr2/addon.json
+++ b/addons/scratchr2/addon.json
@@ -69,6 +69,7 @@
         "https://scratch.mit.edu/403/",
         "https://scratch.mit.edu/404/",
         "https://scratch.mit.edu/500/",
+        "https://scratch.mit.edu/503/",
         "https://scratch.mit.edu/login",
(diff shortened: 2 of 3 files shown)


WorldLanguages:
@mxmou Tell me what you think, and I'm curious, is there extra code to handle 404 to cover the previous behavior? If so, you can remove those matches/code

@apple502j Opinion pls

WorldLanguages:
Resolves #1976

background/get-userscripts.js
```diff
@@ -210,6 +211,20 @@ chrome.alarms.onAlarm.addListener((alarm) => {
   }
 });
 
+chrome.webRequest.onResponseStarted.addListener(
+  (request) => {
+    const identity = createCsIdentity({ tabId: request.tabId, frameId: request.frameId, url: request.url });
```
RedGuy12, line 216: wouldn't it be easier to do
```suggestion
    const identity = createCsIdentity(request);
```

mxmou:
@WorldLanguages If this does not affect how fast userstyles load on non-error pages (if I understand correctly it doesn't), this is a good change. It will allow reverting the part of #1259 that made `dark-www/experimental_scratchr2` run on all pages - it requires lots of `:not()`s in that file and causes multiple minor bugs.

WorldLanguages:
> If this does not affect how fast userstyles load on non-error pages (if I understand correctly it doesn't)

Userstyles load the same way/speed in pages that return 2xx status codes.

WorldLanguages:
Also, a note I forgot to mention in the first post: some 404s are fake 404s, and the number of fake 404s will keep growing as Scratch migrates to scratch-www. For example, project 404s actually return 200, because whether the project ID is valid or not is checked in the client side (as long as a number is passed)

mxmou:
> Also, a note I forgot to mention in the first post: some 404s are fake 404s, and the number of fake 404s will keep growing as Scratch migrates to scratch-www.

This is actually a good thing - if it didn't work like this there would be no way to differentiate between scratchr2 404s and scratch-www 404s.

> For example, project 404s actually return 200, because whether the project ID is valid or not is checked in the client side (as long as a number is passed)

Even if no number is passed: https://scratch.mit.edu/projects

background/get-userscripts.js
WorldLanguages: Related: https://jakearchibald.com/2021/function-callback-risks/

PR #2427, opened 2021-05-07
Ask before moving a project to the trash

### LINKED ISSUE #2413: Ask before moving a project to trash.

I always accidentally click the delete button so if you could add that to the "Confirm actions" addon, that would be great.

Add a checkbox inside of the "Confirm actions" addon to confirm before deleting a project.

FunctionalMetatable:
**Resolves**

Resolves #2413

**Changes**

adds a confirmation for before trashing the project

**Tests**

tested


Shluffy:
Maybe say 'trash' rather then 'delete'. Also, is there a specific reason it is `confirm-actions/trashproject` rather then `confirm-actions/deleteproject` if it using delete rather then trash in other places?

PR #2430, opened 2021-05-07
Change default msg-count-badge color, make it customizable

WorldLanguages:
Moved from #1914


wgyt (approved):
TLDR This is my code but with presets. I approve.

PR #2439, opened 2021-05-08
Prevent purple potassium fire

apple502j:
Potassium, or purple potassium in particular, is difficult to handle. Recent investigations by the Chemical Wrongdoing Search led to a warning to the Scratch Addons team, which mentioned that potassium, due to its sensitive nature, must be handled with care, and should not be used unless it is necessary. This PR fixes the issue by removing unnecessary use of the metal.


Secret-chest:
Why do you want to remove it?

apple502j:
@Secret-chest For safety.

Endermaster3690:
whats wrong with bitmap copy and clipboard apis?

GrahamSH-LLK:
I assume that "Purple potassium" is a chrome web store code (having gotten rejected before)?

WorldLanguages:
Yeah, issue here is that `clipboardWrite` permission means literally nothing for Chrome extensions - you can copy anything, at any time and context, without permissions, using `navigator.clipboard`. Adding that permission does nothing in Chrome. However, it's needed in Firefox to copy images, because `navigator.clipboard.write` is only supported in Chrome - Firefox only supports `navigator.clipboard.writeText`.

WorldLanguages:
Instead of removing the addon entirely, it might be better overall to keep it Chrome only. Which sucks, tbh.
Maybe someone can port this addon into its own extension and publish it on the Firefox store.

GrahamSH-LLK:
MOZMANIFEST WOULD SOLVE THIS

WorldLanguages:
IKR! It's so sad that Mozilla didn't provide a way to give Firefox a different manifest.json. Because of that, Chrome users are missing on a feature right now, being able to use Scratch Addons on incognito :/

apple502j:
Gonna do mozmanifest.

PR #2470, opened 2021-05-11
New addon - adjust script run behavior

### LINKED ISSUE #591: Fix Scratch bug with running duplicated scripts
**Describe the new addon**

When duplicating a block and placing it inside of a script, the addon disables the triggering of the event that would cause that script to run.

_The second part of this issue was migrated to #2471._

lisa-wolfgang:
With this addon, we take one step closer to a complete 2.0 editor.
It addresses a 2-and-a-half-year-old bug that took mere hours to fix.

**Before:**
SA fix-pasted-scripts Demo (Before)

**After:**
SA fix-pasted-scripts Demo (After)

**Resolves**

Resolves #591 (and a duplicate)

**Changes**

This addon fixes a bug in the Scratch editor that unexpectedly runs scripts when placing duplicated blocks inside of them.
It also fixes this behavior when pasting blocks using griffpatch's developer tools.
There's also an option to completely disable running scripts via clicking, or to only run on CTRL + click.

**Reason for changes**

Because it should have been fixed 2.5 years ago. Smh Scratch devs

**Tests**

Achieves its purpose without negative side effects.
Doesn't conflict with any addons from my testing.
Do your best to break it and let me know if you find anything.

**Known bugs**
- None currently


addons/fix-pasted-scripts/userscript.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  // Overwrite the old functions with our new ones
+  vm.editingTarget.blocks.constructor.prototype.blocklyListen = newBlocklyListen;
+
+  // When the setting is toggled, update the variable that contains it
+  addon.settings.addEventListener("change", function () {
+    runMode = addon.settings.get("runMode");
+  });
+}
```
TheColaber, line 60: Set runMode in your newBlocklyListen function since it's only used there... it will remove this code.

TheColaber:
> @GarboMuffin should make a PR with TurboWarp/scratch-blocks@5c5c120

Couldn't you use that code in your PR?

lisa-wolfgang:
> > @GarboMuffin should make a PR with [TurboWarp/scratch-blocks@5c5c120](https://github.com/TurboWarp/scratch-blocks/commit/5c5c120)
> 
> Couldn't you use that code in your PR?

I did originally, but it didn't allow me to disable script running in general, so I switched back to my implementation. I considered using both, but I wanted to keep it to just modifying one of the two (either the VM or scratch-blocks) if possible.

TheColaber:
> didn't allow me to disable script running

Why does it work in garbo's pr then?

lisa-wolfgang:
> > didn't allow me to disable script running
> 
> Why does it work in garbo's pr then?

I should clarify -- GarboMuffin's PR works fine, but it only works for fixing the duplicate bug. The idea is that this addon can disable intentional clicks as well.

TheColaber:
> The idea is that this addon can disable intentional clicks as well.

I'm sure you could also return when that happens?

lisa-wolfgang:
> I'm sure you could also return when that happens?

Yes, but that would disable any interaction with blocks at all -- dragging, deleting, etc.

lisa-wolfgang:
I just reworked this addon -- there should no longer be any bugs! 🎉

addons/fix-pasted-scripts/userscript.js
```diff
@@ -0,0 +1,45 @@
...
+  document.addEventListener("keydown", function (e) {
+    if (e.ctrlKey) {
+      ctrlKeyPressed = true;
+    }
+  });
+  document.addEventListener("keyup", function (e) {
+    if (!e.ctrlKey) {
+      ctrlKeyPressed = false;
+    }
+  });
```
TheColaber, line 20: doing this doesn't feel right... can you check for e.metaKey in onMouseDown_?

lisa-wolfgang: The CTRL check is actually in `newBlocklyListen(e)`, and unfortunately, `metaKey` isn't a property of the `e` passed into that function.

addons/fix-pasted-scripts/addon.json
```diff
@@ -0,0 +1,51 @@
...
+          "id": "default",
+          "name": "On direct click"
+        },
+        {
+          "id": "ctrl",
+          "name": "On CTRL + click"
+        },
+        {
+          "id": "fullDisable",
+          "name": "Never"
```
GarboMuffin, line 42: I'm a little curious. Why would somebody ever want to completely disable this?

lisa-wolfgang: I personally wouldn't use it, but there might be people who appreciate the option, even though there's a CTRL+click option.

WorldLanguages:
1. Removed "never" option, too dangerous. Users can just set it to ctrl+click and not press ctrl key
2. I was getting "vm.editingTarget is null", so I fixed it thanks to GarboMuffin
3. I made dynamicDisable also disable the bug fix, not just ctrl+click
4. I removed dynamicEnable because I couldn't get it to work
5. Added recommended tag

WorldLanguages:
Note: dynamicEnable does work but you need to go into another tab (e.g. customes) then back into script tab. I don't remember right now but I think an addon also had this problem and there was a method in blockly that forced a refresh? @apple502j help?

RedGuy12:
I think it was hide-flyout that had this problem.

WorldLanguages:
@lisa-wolfgang agreed on having this addon just fix the bug. They'll send another PR later that lets users customize what click/key combination runs blocks. So this addon will just be a bug fix, nothing else

PR #2472, opened 2021-05-11
Manual translation update: 2021/05/11

apple502j:
New requirement for translations is now in place:

- For current translations, this PR drops all that do not meet the new rules.
- New languages will simply be not added on this repo until it meets the requirement.
- If a language is translated enough at one point, then it will not be removed in the future, even if new strings cause them to not meet the requirements.

Requirements:
- 90% translated for _locales
- 80% translated for addons-l10n
- No language can have only addons-l10n (i.e. if addons-l10n exists, _locales must exist)

Also fixes a bug where translations for a deleted addon, `search-load-more-fix`, weren't removed.


WorldLanguages (approved):

WorldLanguages:
Can we have a file that lists all the locale codes? It would be useful for many cases, such as https://github.com/ScratchAddons/extension-store-description-tool and avoiding some types of 404 errors in the console

apple502j:
Not right now

PR #2485, opened 2021-05-13
Match scratch-www studios differently

WorldLanguages:
This only changes content script code
Behavior does not change in any way if URL doesn't match studio regex
If the URL matches a studio (and it didn't 404), `onInfoAvailable` waits until `document.head` is available before doing _anything_. This may add a slight delay before injecting userscripts and userstyles, even before Scratch transitions studios to scratch-www, but it's barely noticeable, if at all.
After the head is available, it checks if `addon.tab.clientVersion` would return `"scratch-www"`. If so, similarly to #2416 it will set the pseudoUrl of the page to `/studios_www/...` instead of `/studios/...`. Similarly to #2416, userscripts and userstyles will take a bit more time to get injected after studios become scratch-www (and now it can be probably seen by eye, specially with website dark mode). This isn't ideal but good enough to have a clean transition (of course, after the transition by Scratch, we can remove all of this and we won't have any other timing problems like this)

TODOs:
- [x] Adapt addons so that they can use this (cc @mxmou, feel free to push into this PR)
- [ ] Change regexes to consider `/studios_www/...` to be a scratch-www page


content-scripts/cs.js
```diff
@@ -223,24 +223,37 @@ function setCssVariables(addonSettings, addonsWithUserstyles) {
   }
 }
 
+function waitForDocumentHead() {
+  if (document.head) return Promise.resolve();
+  else {
+    return new Promise((resolve) => {
+      const observer = new MutationObserver(() => {
+        if (document.head) resolve();
```
apple502j, line 231: disconnect!

apple502j:
I think this is kinda overkill.

mxmou:
@apple502j It's better than using complex selectors to prevent CSS conflicts. `scratchr2.css` and `scratchwww.css` are separate because running both on all pages causes bugs.

Hans5958:
what a bodge

WorldLanguages:
The art of the bodge

RedGuy12:
>  Adapt addons so that they can use this (cc @mxmou, feel free to push into this PR)
apple did that in #2493

WorldLanguages:
> apple did that in #2493

I meant doing it for themes, not addons that use JS

WorldLanguages:
> * [ ]  Change regexes to consider `/studios_www/...` to be a scratch-www page

Note: this TODO wasn't done. But no addon uses `scratchWWWNoProject` or `isNotScratchWWW` matches, so it doesn't matter

PR #2493, opened 2021-05-14
Make existing addons work on new studios

apple502j:
This breaks some laws of dot updates (particularly, "no string update") but it's worth it.

Please test this on current studios (and make sure I didn't leave debugging code.)

### DIFF
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -10,7 +10,7 @@ export default async function ({ addon, console, msg }) {
         cancelMessage = msg("share");
       } else if (addon.settings.get("followinguser") && e.target.closest("#profile-data .follow-button")) {
         cancelMessage = msg("follow");
-      } else if (addon.settings.get("joiningstudio") && e.target.closest("a.accept")) {
+      } else if (addon.settings.get("joiningstudio") && (e.target.closest("a.accept") || e.target.textContent === addon.tab.scratchMessage("studio.curatorAcceptInvite"))) {
         cancelMessage = msg("joinstudio");
       }
       if (cancelMessage !== null) {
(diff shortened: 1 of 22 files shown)


addons/infinite-scroll/buttonScroll.js
```diff
@@ -14,12 +15,13 @@ async function commentLoader(addon, heightControl, selector) {
 }
 
 export default async function ({ addon, global, console }) {
-  if (
-    window.location.pathname.split("/")[1] === "studios" &&
-    window.location.pathname.split("/")[3] === "comments" &&
-    addon.settings.get("studioScroll")
-  )
-    commentLoader(addon, "#content", "#comments > div:nth-child(2) > ul > div");
+  if (window.location.pathname.split("/")[1].startsWith("studios") && addon.settings.get("studioScroll")) {
```
WorldLanguages, line 18: `startsWith` could be `===` just fine, I assume this was to test it properly. Will still work ofc

addons/more-links/userscript.js
```diff
@@ -26,6 +26,8 @@ export default async function ({ addon, console }) {
       break;
 
     case "studios":
+    case "studios-playground":
```
WorldLanguages, line 29: Same as above

addons/confirm-actions/userscript.js
```diff
@@ -10,7 +10,11 @@ export default async function ({ addon, console, msg }) {
         cancelMessage = msg("share");
       } else if (addon.settings.get("followinguser") && e.target.closest("#profile-data .follow-button")) {
         cancelMessage = msg("follow");
-      } else if (addon.settings.get("joiningstudio") && e.target.closest("a.accept")) {
+      } else if (
+        addon.settings.get("joiningstudio") &&
+        (e.target.closest("a.accept") ||
+          e.target.textContent === addon.tab.scratchMessage("studio.curatorAcceptInvite"))
```
WorldLanguages, line 16: This is truthy when clicking anything without a textContent (for example an image) in any non-studio page without that locale (for example, explore page before new studios were shipped - not sure if it will have the locale after it)

PR #2501, opened 2021-05-15
Use JSON API for retrieving/replying/deleting comments in projects and studios (scratch-messaging)

### LINKED ISSUE #2046: scratch-messaging: support JSON API for comments
Right now, we're still using the old `site-api` API for comments, even for project comments. Studio pages and profiles still use that, so they cannot be removed, but the `site-api/project` endpoint could be.
We should migrate to the JSON API for project comments.
Whenever we need to switch studios and profiles, the switch will also be easier if project comments already work with the JSON API.

WorldLanguages:
Resolves #2046 
- [x] Show project and gallery comments using JSON API
- [x] Send and delete comments using JSON API
- [x] Test everything works the same as before in both Chrome and Firefox

Caveats:
- Links aren't supported yet. No links will be clickeable, including scratch.mit.edu links

### DIFF
--- a/addons/scratch-messaging/background.js
+++ b/addons/scratch-messaging/background.js
@@ -322,8 +322,36 @@ export default async function ({ addon, global, console, setTimeout, setInterval
     return value;
   }
 
-  function sendComment({ resourceType, resourceId, content, parent_id, commentee_id }) {
-    return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


jeffalo:
note that new studios comment endpoint seems to be frozen right now.

WorldLanguages:
`&nocache=${Date.now}` is used when fetching a parent comment's replies to avoid the aggresive caching.
No need to add that for requesting single comment info, since that doesn't change with time, unlike the array of replies

PR #2508, opened 2021-05-16
Add reduxCondition to more addons

GarboMuffin:
**Changes**

I missed some addons in #2268

This adds reduxCondition and similar optimizations to:

 - comments-linebreaks
 - remix-tree-button
 - more-links
 - infinite-scroll
 - copy-message-link
 - true-youtube-links
 - better-emojis

Has been at least briefly tested

### DIFF
--- a/addons/better-emojis/userscript.js
+++ b/addons/better-emojis/userscript.js
@@ -1,42 +1,27 @@
 export default async function ({ addon, global, console }) {
-  const body = document.body;
-
-  // Check if there are any emojis on the page when this script loads.
-  // On Scratch 2.0-esque pages, emojis are given a different class and image directory than on Scratch 3.0-esque pages.
-
-  // For 2.0 pages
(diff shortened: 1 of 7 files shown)


RedGuy12:
Some conditions for addons that interact with project comments were removed in #2493 because the same code will run on www studio comments, which doesn't have redux

GarboMuffin:
> Some conditions for addons that interact with project comments were removed in #2493 because the same code will run on www studio comments, which doesn't have redux

The new www studio comments *will* have redux. All scratch-www pages have redux, even contact-us does. It won't be a problem anyways because everything that can run on those pages checks `if (!state.scratchGui) return true;`

apple502j:
Note that reduxCondition is fool-proof so even if we run it on scratchr2 page that should be fine.

WorldLanguages:
TODO: check URL matches for true-youtube-links, probably shouldn't be wildcard

PR #2478, opened 2021-05-12
Multiple meta updates

Hans5958:
I have did this since a while ago, but I think it's the best time to do this, especially since I need to do something here for website-v2.

Here are a list of changes. Let me know if I missed a few and you need a clarification.

- Remove Danger JS and Labeler until they can do outside PRs or we have an alternative solution.
- Remove separate "Help" and "Discussions" links from the "create new issue" page because it is a long fad since the pre-Discussions era and I think it is now unused
- Improve issue and PR templates (readme, links, use headings (`### Description`) instead of bold text (`**Description**`), etc)
- Improve Prettify workflow by replacing it with an Action and use SAB inside of this repo
- Add community Discord server on the "create new issue" page (also put above the other server to prioritize) 
- Rename Schema Validation workflow
- Swap website repo on workflow to website-v2
- Link the community Discord server instead of the other on README.md

It's been a while since the last comms on this repo so I might missed something important. Comments (and probably objections) are appreciated, as always.

PS. For some reason I am quite sick right now. Maybe it's because of the "microcheeps" (it's a reference) or it's just because I have a bad sleep last night. Hopefully I can be fine tommorow.

PS 2. Wow. For the something-th time. apple502j decided to ~~remove~~ disable Codacy! I wonder if it's because of his hatred of this kind of this kind of software or some kind, even I could guess that the majority fine with it.

PS 3. Oh, yeah. The message of the Prettier workflow is now `Format code` instead of `style: format code`.

PR #2516, opened 2021-05-17
Multiple meta updates (2nd try)

Hans5958:
Second try, everyone. Wish me luck. Succeeds #2478.

----

I have did this since a while ago, but I think it's the best time to do this, especially since I need to do something here for website-v2.

Here are a list of changes. Let me know if I missed a few and you need a clarification.

- Remove Danger JS and Labeler until they can do outside PRs or we have an alternative solution.
- Remove separate "Help" and "Discussions" links from the "create new issue" page because it is a long fad since the pre-Discussions era and I think it is now unused
- Improve issue and PR templates (readme, links, use headings (`### Description`) instead of bold text (`**Description**`), etc)
- Improve Prettify workflow by replacing it with an Action and use SAB inside of this repo
- Add community Discord server on the "create new issue" page (also put above the other server to prioritize) 
- Rename Schema Validation workflow
- Swap website repo on workflow to website-v2
- Link the community Discord server instead of the other on README.md

It's been a while since the last comms on this repo so I might missed something important. Comments (and probably objections) are appreciated, as always.

PS. For some reason I am quite sick right now. Maybe it's because of the "microcheeps" (it's a reference) or it's just because I have a bad sleep last night. Hopefully I can be fine tommorow.

PS 2. Wow. For the something-th time. apple502j decided to ~~remove~~ disable Codacy! I wonder if it's because of his hatred of this kind of this kind of software or some kind, even I could guess that the majority fine with it.

PS 3. Oh, yeah. The message of the Prettier workflow is now `Format code` instead of `style: format code`.


Hans5958:
6 commits and it's now working. Haha.

PR #2518, opened 2021-05-17
Redirect users to /unsupported if version is unsupported

TheColaber:
Checks on extension load, and when the user goes to the popup or settings page.

TODO: Create page on website (https://github.com/ScratchAddons/website-v2/issues/11)

### DIFF
--- a/background/handle-unsupported-version.js
+++ b/background/handle-unsupported-version.js
@@ -1,0 +1,18 @@
+function handleUnsupported() {
+  const checkIfUnsupported = () => {
+    const getVersion = () => {
+      let userAgent = /(Firefox|Chrome)\/([0-9.]+)/.exec(navigator.userAgent);
+      if (!userAgent) return { browser: null, version: null };
+      return { browser: userAgent[1], version: userAgent[2].split(".")[0] };
+    };
+
+    let { browser, version } = getVersion();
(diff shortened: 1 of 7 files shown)


background/handle-unsupported-version.js
```diff
@@ -0,0 +1,18 @@
...
+  const checkIfUnsupported = () => {
+    const getVersion = () => {
+      let userAgent = /(Firefox|Chrome)\/([0-9.]+)/.exec(navigator.userAgent);
+      if (!userAgent) return { browser: null, version: null };
+      return { browser: userAgent[1], version: userAgent[2].split(".")[0] };
+    };
+
+    let { browser, version } = getVersion();
+    return (browser === "Chrome" && version < 81) || (browser === "Firefox" && version < 74);
+  };
```
GrahamSH-LLK, line 11: To maintain consistency with other parts of code, use
```js
      if (typeof browser !== "undefined") 
``` 
for browser checking instead of ua

TheColaber: no because the ua has a multipurpose, getting the browser AND the version

GrahamSH-LLK: No, we want to maintain consistency. PLUS, I'm not sure if this will work in other Chromium browsers.

TheColaber: no wl said it will, also there is not other way to get the version (except ff)

PR #2522, opened 2021-05-18
Capture event handlers in tab.traps

### LINKED ISSUE #2521: Capture event listeners in tab.traps

Not related to any issue so far (as far as I know).

I have already drafted a working piece code for this, using the same sort of method as the extension uses to get the vm:

```js

//Works
const listeners = (__scratchAddonsTraps.eventListeners = new Array);

const oldAddListener = Element.prototype.addEventListener;
Element.prototype.addEventListener = function (...args) {
  listeners.push({
    event: args[0],
    callback: args[1],
    options: args[2],
    element: this,
  });

  oldAddListener.bind(this)(...args);
};

//Also works now...
const oldRemoveListener = Element.prototype.removeEventListener;
Element.prototype.removeEventListener = function (...args) {
  let listener = listeners.splice(
    listeners.indexOf(
      listeners.filter(
        (listener) =>
          listener.event === args[0] &&
          listener.callback === args[1] &&
          (typeof args[2] === "object" && typeof listener.options === "object" ? JSON.stringify(args[2]) === JSON.stringify(listener.options) : args[2] === listener.options) &&
          this === listener.element
      )[0]
    ),
    1
  );

  oldRemoveListener.bind(this)(...args);
};
``` 

This will allow us to manually remove and/or use listeners on elements that are usually hidden. For example the file uploads. They have change listeners on them, but they run before the event can be prevented keeping us from changing them. If, for example, I wanted to change how files were uploaded (I am working on an addon that will allow images to be uploaded at full quality, but I need this) I would need to remove the current listener and add a new one that would trigger the other listener after I have manipulated the file. With this, we could remove the event listener (because we would have the function used), and be able to re-assign it to a different, custom, event.

I haven't found any alternatives yet...

### Additional context

Api suggestions:

- `addon.tab.traps.getListenersForElement(selector)`
  - Example method:
```js
getListenersForElement = (selectorOrElement) =>
  __scratchAddonsTraps.eventListeners.filter(
    typeof selectorOrElement === "string" ?
      (listener) => listener.element === document.querySelector(selectorOrElement) :
      (listener) => listener.element === selectorOrElement
  )
```
  - Example use:
```js
const listeners = addon.tab.traps.getListenersForElement('input[type="file"]');

const listeners = addon.tab.traps.getListenersForElement(document.querySelectorAll('input[type="file"]')[0]);
```

ErrorGamer2000:
Resolves #2521

### Reason for changes

Expands the api to include capturing event handlers to allow addons to replace/change how they work.

### Tests

I have tested this, mainly just from the console though. It captured (I think) all of the event listeners (about 117 of them XD), and I was able to successfully remove them and add them back without changing functionality (I removed and replaced the "mousedown" listener on the main canvas).

Please ignore any commits not on May 18, 2021 (before a8bb364), I have removed the others. Because of the way I have set up my forks, I had to delete one of my pending addons.


content-scripts/prototype-handler.js
```diff
@@ -21,6 +21,40 @@ function injectPrototype() {
       return oldBind.apply(this, args);
     }
   };
+
```
RedGuy12, line 24: this would only run on project pages?

ErrorGamer2000: Oof, true...

addon-api/content-script/Trap.js
```diff
@@ -25,6 +25,19 @@ export default class Trap extends Listenable {
     return __scratchAddonsTraps._onceMap.vm;
   }
 
+  get eventListeners() {
+    return __scratchAddonsTraps.listeners;
+  }
+
+  getListenersForElement(selectorOrElement) {
+    return this.eventListeners.filter(function (listener) {
```
RedGuy12, line 33: this.eventListeners is undefined

content-scripts/prototype-handler.js
ErrorGamer2000: Hadn't hought of that. Just a sec.

addon-api/content-script/Trap.js
ErrorGamer2000: Ummm, no? I added a `get` before.

RedGuy12:
> I have tested this, mainly just from the console though.

that doesn't count. actually checkout the branch and test it.

addon-api/content-script/Trap.js
RedGuy12: oh, i missed that

content-scripts/prototype-handler.js
ErrorGamer2000: How do you suggest I move this? Should I make a new `__scratchAddons` object in the window?

RedGuy12: move it to `cs.js`

ErrorGamer2000: Okay.

ErrorGamer2000: I am thinking about adding a `window.__scratchAddonsEventListeners` variable, sound okay?

RedGuy12: no, use scratchAddons.localState.eventListeners

ErrorGamer2000: Okay.

ErrorGamer2000: Umm.... No can do? It is an object in the bg, this needs to be for content-scripts.

RedGuy12: oh wait

move it to `content-scripts/inject/module.js` and use `scratchAddons.eventListeners`

ErrorGamer2000:
I will check it later . . . I have a LOT of work to do for school right now, sorry.

apple502j:
Four arguments:
1. This will hurt performance a lot. Storing listeners ourselves could prevent them from getting GCed, causing memory leak. At least WeakMap that.
2. removeEventListener is also doing function calls that are O(n^2) in the worst case, if i'm correct.
3. This is the third argument; [there is a third argument.](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters)
4. There is also a fourth argument related to your [use case](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent).

ErrorGamer2000:
I thought I had taken care of the third argument ... and stored it.

apple502j:
You don't need this at all. Just `{capture: true}`

ErrorGamer2000:
Yeah, but I need to modify an existing listener, and save the original handler

ErrorGamer2000:
I am trying to change a file before it is uploaded as a costume.

apple502j:
You can either trap the handleUpload method (see folders addon) or implement costume uploading yourself, capture the event and call stopPropagation.

ErrorGamer2000:
Huh. Hadn't thought of that, thanks!

PR #2524, opened 2021-05-18
Use console.debug instead of console.log

Sly-Little-Fox:
This PR doesn't resolve any issues.

### Changes

Replaced console.log in userscript loader with console.debug.

### Reason for changes

The userscript loader output is verbose, so it will fit better in "debug" tab.

### Tests

I went to the Scratch website and opened console. Everything works.


apple502j:
Doesn't really matter.

PR #2525, opened 2021-05-18
Better image uploads addon

ErrorGamer2000:
### Changes

Added the "Better image uploads" addon, which automatically converts any uploaded image (in the editor) to a svg so that you do not lose quality because it is not converted to bitmap.

### Tests

Costumes:

https://user-images.githubusercontent.com/70162741/118713072-c3c4a800-b7de-11eb-8f74-2b6328f07ed7.mp4

Sprites and Backdrops:

https://user-images.githubusercontent.com/70162741/118713773-90cee400-b7df-11eb-82d3-2fae11692352.mp4


WorldLanguages:
cc @SheepTester

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Better image uploads",
+  "description": "Automatically converts uploaded images into svg images to keep them from losing quality.",
+  "tags": ["editor", "beta"],
+  "credits": [
+    {
+      "name": "ErrorGamer2000",
+      "link": "https://scratch.mit.edu/users/ErrorGamer2000/"
+    }
+  ],
```
RedGuy12, line 10: False

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,22 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"],
+      "runAtComplete": true
+    }
+  ],
+  "dynamicDisable": true,
+  "dynamicEnable": true,
+  "l10n": false
```
RedGuy12, line 21: True

addons/better-img-uploads/addon.json
TheColaber: wut

RedGuy12: oh i did it on the wrong line oops

i was referring to `enabledByDefault` which is currently true.

addons/better-img-uploads/addon.json
ErrorGamer2000: ??? Make the `l10n` true?

TheColaber: Yes, every addon should have this true

RedGuy12: even tho you don't use `msg`, l10n needs to be true so the addon name and description can be translated.

addons/better-img-uploads/userscript.js
```diff
@@ -0,0 +1,101 @@
...
+
+    let input = await addon.tab.waitForElement('input[type="file"][accept*=".svg"]', {
+      markAsSeen: true,
+    }); //Wait for the next image file input
+
+    input.addEventListener(
+      "change",
+      (e) => {
+        if (!enabled) return;
+        if (!ignore) (ignore = true) && onchange(e);
```
apple502j, line 18: No one-liner please. Also call stopPropagation here maybe?

ErrorGamer2000: How is it now?

addons/better-img-uploads/userscript.js
```diff
@@ -0,0 +1,101 @@
...
+    let processed = new Array();
+
+    for (let file of files) {
+      if (file.type.includes("svg")) {
+        //The file is already a svg, we should not change it...
+        processed.push(file);
+        continue;
+      }
+
+      let blob = await new Promise((resolve) => {
```
apple502j, line 42:
```js
let blob = file;
```

addons/better-img-uploads/userscript.js
```diff
@@ -0,0 +1,101 @@
...
+      );
+    }
+
+    el.files = arrayToFileList(processed); //Convert processed image array to a FileList, which is not constructible.
+
+    el.dispatchEvent(e); //Start the event over again, but allow scratch to receive it this time.
+  }
+
+  function arrayToFileList(arr) {
+    let filelist = new ClipboardEvent("").clipboardData || new DataTransfer(); //This "creates" a FileList that we can add files to
```
apple502j, line 97: `new DataTransfer();` works fine on all versions

ErrorGamer2000: okay

addons/better-img-uploads/userscript.js
ErrorGamer2000: That doesn't work, just throws an error...

ErrorGamer2000: In order to correctly add the image to an svg for scratch it has to be a dataURL

GarboMuffin:
Bitmaps embedded in SVG are noticeably worse in several ways:

 - Can't be edited with bitmap tools without first converting to bitmap
   - Extra button click per costume
   - Might also have to manually resize the image (humans will do a worse job at this than computers)
   - Also seems like right now converting the "better" image uploads to bitmap will make them blurry
 - File sizes will be increased. Not only is base64 inefficient, but you include the image source twice for some reason (surely that isn't necessary?)

Perhaps only do this on big images that would actually be otherwise resized or find ways to resolve those issues

ErrorGamer2000:
Oh, yeah that was something I was testing at the beginning, I don't need the normal `href`. Also, I will add n automatic resize into the 360 x 480 window.

ErrorGamer2000:
> Extra button click per costume
Alright then, maybe instead of just overriding the default upload add a new button on the slideup menu, and an option to make this the default?

> Also seems like right now converting the "better" image uploads to bitmap will make them blurry
That is because they are being uploaded at full size. the sizing will fix that.

SheepTester:
> Alright then, maybe instead of just overriding the default upload add a new button on the slideup menu, and an option to make this the default?

If you're adding a new button, you can use the icon I made when I worked on something similar:

[image](https://github.com/SheepTester/ScratchAddons/blob/png-as-svg/addons/high-res-png/icon.svg)

ErrorGamer2000:
Looks good, thanks!

ErrorGamer2000:
Do you have an svg of it?

SheepTester:
> Do you have an svg of it?

Yes, the image was a link to https://github.com/SheepTester/ScratchAddons/blob/png-as-svg/addons/high-res-png/icon.svg

addons/better-img-uploads/userscript.js
apple502j: ok

ErrorGamer2000:
Ah, okay thanks. Sorry it took so long to respond, I was taking my AP exam.

ErrorGamer2000:
Alright, @GarboMuffin will you take a look now? I think I fixed everything. I added the new buttons (and tooltips), and added options for the image fitting. The blurriness is a problem with scratch-paint, and it shows fine on the stage itself. As for the inefficiency of base64, I can't really help that. Also, scratch-paint itself uses base64. Example:

1. Upload a bitmap image
2. Convert it to a vector
3. Change the file extension (I used `.xml`)
4. Find the image tag

The image's `xlink:href` is going to be a base64 url.

ErrorGamer2000:
Here is a vid of it:

https://user-images.githubusercontent.com/70162741/119058993-a9c8c800-b98c-11eb-8b61-2c28e1bdca1c.mp4

addons/better-img-uploads/userscript.js
```diff
@@ -0,0 +1,160 @@
...
+
+  //The class name for the menu
+  let c = addon.tab.scratchClass("action-menu_more-buttons");
+
+  while (true) {
+    //Catch all upload menus as they are created
+    let menu = await addon.tab.waitForElement(`.${c}`, { markAsSeen: true });
+    let button = menu.parentElement.previousElementSibling.previousElementSibling; //The base button that the popup menu is from
+
+    let id = button.ariaLabel.replaceAll(" ", "_");
```
GarboMuffin, line 33: does not work in firefox
probably use getatribute instead

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,48 @@
...
+  "name": "Better image uploads",
+  "description": "Automatically converts uploaded images into svg images to keep them from losing quality.",
+  "tags": ["editor", "beta"],
+  "credits": [
+    {
+      "name": "ErrorGamer2000",
+      "link": "https://scratch.mit.edu/users/ErrorGamer2000/"
+    }
+  ],
+  "enabledByDefault": true,
```
GarboMuffin, line 11: probably not?

GarboMuffin (changes requested):
As long as it adds a separate button and does not affect the default upload button then my complaints are mute. My primary concern was that for an addon called "Better image uploads", it was making bitmap uploads a lot worse.

Maybe rename the addon to "HD bitmap costume upload" or something less vague.

The "HD" in the icon is way too small to read.

Tooltip appears too high https://user-images.githubusercontent.com/33787854/119060654-447ad400-b998-11eb-83b7-d77ddece3115.mp4

> The blurriness is a problem with scratch-paint, and it shows fine on the stage itself

~~If I convert an HD image upload to bitmap then it still becomes blurry on the stage too. Scratch does not have this problem, so it is probably a bug in your code. (ok, it's probably a Scratch bug, but it's something you should be able to workaround)~~

haven't taken a hard look at the code or much other testing

addons/better-img-uploads/addon.json
ErrorGamer2000: Oh, sorry!

RedGuy12:
Instead of

    .replaceAll(" ", "_");
Do
    .replace(\/s\g, "_")

Because replaceAll is not supported by most browsers yet.

ErrorGamer2000:
AH, ok

RedGuy12:
Oops I got it wrong. Do _this_ instead:

```js
.replace(/\s/g, "_")
```

ErrorGamer2000:
@GarboMuffin

> If I convert an HD image upload to bitmap then it still becomes blurry on the stage too. Scratch does not have this problem, so it is probably a bug in your code. (ok, it's probably a Scratch bug, but it's something you should be able to workaround)

It is not really a bug with scratch that way... with the HD upload it uploads a scaled vector image. When you convert to bitmap via scratch-paint it converts the image to a `480` x `360` png image **no matter what the original size was.** This is what is making your image blurry/pixely(pixeley?) on the stage. For it to keep full quality it needs to stay in vector.

> Maybe rename the addon to "HD bitmap costume upload" or something less vague.

How about "HD Image Uploads"?

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,54 @@
...
+    {
+      "name": "SheepTester",
+      "link": "https://scratch.mit.edu/users/Sheep_maker/"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"],
+      "runAtComplete": true
```
RedGuy12, line 22: remove this line. `"runAtComplete": true` is the default, so it is not needed.

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,54 @@
...
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Image Fitting",
```
RedGuy12, line 33:
```suggestion
      "name": "Image resizing",
```

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,54 @@
...
+      "id": "fitting",
+      "type": "select",
+      "potentialValues": [
+        {
+          "id": "full",
+          "name": "Full Size"
```
RedGuy12, line 39:
```suggestion
          "name": "Original size"
```

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        {
+          "id": "fill",
+          "name": "Fill Stage"
```
RedGuy12, line 43:
```suggestion
          "name": "Stretch to fill stage"
```

addons/better-img-uploads/addon.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        {
+          "id": "fit",
+          "name": "Fit to Stage"
```
RedGuy12, line 47:
```suggestion
          "name": "Shrink to fit stage"
```

RedGuy12 (approved):

ErrorGamer2000:
Thanks

GarboMuffin:
So, things becoming blurry apparently when converted to bitmap does happen in unmodified Scratch so not something this addon can/should change

icon.svg currently has syntax errors and does not render  in any browser, so, probably try to fix that

ErrorGamer2000:
> icon.svg currently has syntax errors and does not render in any browser, so, probably try to fix that

RIP okay, done. 😂

WorldLanguages:
It stresses me out how a 480x270 JPG gets converted to a 481x271 SVG when uploaded the same way (as a costume), and is positioned slightly differently to the same image but being uploaded normally 🤔 Is this intentional, a simple bug to fix?

WorldLanguages:
You may not be rounding the same way as Scratch?
I forgot to mention, the original JPG is 1025x576. It gets converted to 480x270 by vanilla Scratch and to 481x271 by your addon. It should probably be resized the same way as vanilla Scratch to be consistent

ErrorGamer2000:
Hmmm, that is a new issue. That may be part of my moving the image `0.5` px down in order to prevent the `1px` whitespace that otherwise shows up at the bottom of the image... I will check on monday.

WorldLanguages:
Releasing for next version in around 2 weeks (v1.16.0)

ErrorGamer2000:
Okay.

Explosion-Scratch:
OK NOW THAT'S JUST EPIC FEEL FREE TO MARK MY COMMENT AS SPAM OR OFF TOPIC WL

WorldLanguages:
@ErrorGamer2000 Do you think you could try to make it so that your addon downscales images the same way as Scratch? It will be easier to showcase the difference that way, even tho it's just a minor detail

ErrorGamer2000:
Ah, I found it. https://github.com/LLK/scratch-paint/blob/2a9fb2356d961200dc849b5b0a090d33f473c0b5/src/helper/bitmap.js#L358

```js
const getHitBounds = function (raster, rect) {
    const bounds = rect || raster.bounds;
    const width = bounds.width;
    const imageData = raster.getImageData(bounds);
    let top = 0;
    let bottom = imageData.height;
    let left = 0;
    let right = imageData.width;

    while (top < bottom && rowBlank_(imageData, width, top)) ++top;
    while (bottom - 1 > top && rowBlank_(imageData, width, bottom - 1)) --bottom;
    while (left < right && columnBlank_(imageData, width, left, top, bottom)) ++left;
    while (right - 1 > left && columnBlank_(imageData, width, right - 1, top, bottom)) --right;

    // Center an empty bitmap
    if (top === bottom) {
        top = bottom = imageData.height / 2;
    }
    if (left === right) {
        left = right = imageData.width / 2;
    }

    return new paper.Rectangle(left + bounds.left, top + bounds.top, right - left, bottom - top);
};
```

Will implement soon.

GarboMuffin:
> **Edit**: Actually, that is just the centering code, it is here: https://github.com/LLK/scratch-paint/blob/2a9fb2356d961200dc849b5b0a090d33f473c0b5/src/helper/view.js#L169

That's not it either. That just does viewport centering. I think you're for https://github.com/LLK/scratch-svg-renderer/blob/develop/src/bitmap-adapter.js

RedGuy12:
Just add it as a lib

RedGuy12:
You can include @scratch/paper as a lib

GarboMuffin:
> You can include @scratch/paper as a lib

not really and its not needed anyways

ErrorGamer2000:
> That's not it either. That just does viewport centering. I think you're looking for https://github.com/LLK/scratch-svg-renderer/blob/develop/src/bitmap-adapter.js#L69

Oh, jeez. That is exactly what I needed @GarboMuffin, thanks!

ErrorGamer2000:
Will someone test it now? I am busy.

addons/better-img-uploads/userscript.js
```diff
@@ -0,0 +1,203 @@
...
+    "action-menu_tooltip"
+  )} sa-better-img-uploads-tooltip" id="sa-${id}-HD Upload" data-id="tooltip" >${m("upload")}</div>
+</div>`;
+
+  //The class name for the menu
+  let c = addon.tab.scratchClass("action-menu_more-buttons");
+
+  while (true) {
+    //Catch all upload menus as they are created
+    let menu = await addon.tab.waitForElement(`.${c}`, { markAsSeen: true });
```
WorldLanguages, line 30: You should not use the return value of `addon.tab.scratchClass` to then run querySelector on it. It should only be used as a class name you add to elements the addon creates and owns.
You should just use an attribute selector `[class^="action-menu_more-buttons"]`

WorldLanguages: Even with this issue the addon appears to work most of the time. But it failed here once

WorldLanguages:
The issue didn't appear to get solved. The original image is 1080x732. First is uploaded with vanilla upload, second with HD upload.
This might look like a small detail but it makes it harder to showcase how HD upload enhances images since pixels move a bit

image

WorldLanguages:
I might have forgotten to pull? I tested again on "shrink to fit stage" mode and it's double the expected size: 
image

addons/better-img-uploads/userscript.js
ErrorGamer2000: Will do...

ErrorGamer2000:
I'll work on that...

ErrorGamer2000:
can you try again? I think I fixed it...

addons/better-img-uploads/userscript.js
```diff
@@ -0,0 +1,200 @@
...
+  </button>
+  <div class="__react_component_tooltip place-${right ? "left" : "right"} type-dark ${addon.tab.scratchClass(
+    "action-menu_tooltip"
+  )} sa-better-img-uploads-tooltip" id="sa-${id}-HD Upload" data-id="tooltip" >${m("upload")}</div>
+</div>`;
+
+  while (true) {
+    //Catch all upload menus as they are created
+    let menu = await addon.tab.waitForElement('[class^="action-menu_more-buttons"]', { markAsSeen: true });
+    let button = menu.parentElement.previousElementSibling.previousElementSibling; //The base button that the popup menu is from
```
WorldLanguages, line 28: It's not working for me now because button is null

ErrorGamer2000: Oh. It needs to be
```js
let menu = await addon.tab.waitForElement('[class*="action-menu_more-buttons"]', { markAsSeen: true });
```

Daniel4-Scratch (approved):

Daniel4-Scratch:
> The "HD" in the icon is way too small to read.
i made the hd bigger

WorldLanguages:
Appears to crash, probably wrong amount of `parentElement` or `previousElementSibling`
Probably can be done easier with `.closest()`
image

ErrorGamer2000:
Alright, I will fix it either tomorrow or Tuesday. Also, I am almost done creating the `addon.messaging` api. I have already implemented the `sendMessage` and `onMessage` parts and am working on the ports now.

ErrorGamer2000:
Fixed it...

addons/better-img-uploads/userscript.js
```diff
@@ -0,0 +1,201 @@
...
+  </button>
+  <div class="__react_component_tooltip place-${right ? "left" : "right"} type-dark ${addon.tab.scratchClass(
+    "action-menu_tooltip"
+  )} sa-better-img-uploads-tooltip" id="sa-${id}-HD Upload" data-id="tooltip" >${m("upload")}</div>
+</div>`;
+
+  while (true) {
+    //Catch all upload menus as they are created
+    let menu = await addon.tab.waitForElement('[class*="action-menu_more-buttons_"]', { markAsSeen: true });
+    console.log(menu);
```
GarboMuffin, line 28: dont log things unnecessarily

ErrorGamer2000: Oh, I forgot it was there, thanks.

PR #2526, opened 2021-05-18
Some light mode changes

GrahamSH-LLK:
fixes up some minor stuffs

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -787,8 +787,7 @@ const vue = (window.vue = new Vue({
     },
     setTheme(mode) {
       chrome.storage.sync.get(["globalTheme"], function (r) {
-        let rr = true; //true = light, false = dark
-        rr = mode;
+        let rr = mode;
         chrome.storage.sync.set({ globalTheme: rr }, function () {
           if (rr && r.globalTheme !== rr) {
             document.head.appendChild(lightThemeLink);
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -787,7 +787,7 @@ const vue = (window.vue = new Vue({
     },
     setTheme(mode) {
       chrome.storage.sync.get(["globalTheme"], function (r) {
-        let rr = mode;
+        let rr = mode || true;
         chrome.storage.sync.set({ globalTheme: rr }, function () {
           if (rr && r.globalTheme !== rr) {
             document.head.appendChild(lightThemeLink);
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -787,7 +787,7 @@ const vue = (window.vue = new Vue({
     },
     setTheme(mode) {
       chrome.storage.sync.get(["globalTheme"], function (r) {
-        let rr = mode || true;
+        let rr = mode ?? true;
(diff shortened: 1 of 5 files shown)


webpages/settings/index.js
```diff
@@ -787,8 +787,7 @@ const vue = (window.vue = new Vue({
     },
     setTheme(mode) {
       chrome.storage.sync.get(["globalTheme"], function (r) {
-        let rr = true; //true = light, false = dark
-        rr = mode;
+        let rr = mode;
```
TheColaber, line 790: why did you remove this logic?

ErrorGamer2000: It seems to me that it will be changed no matter what, and it is just unnecessary.

GrahamSH-LLK: Yeah, if you look at the old logic it sets it to true then proceeded to change it to mode

TheColaber: im now realizing that the logic is broken

webpages/settings/index.js
```diff
@@ -787,8 +787,7 @@ const vue = (window.vue = new Vue({
     },
     setTheme(mode) {
       chrome.storage.sync.get(["globalTheme"], function (r) {
-        let rr = true; //true = light, false = dark
-        rr = mode;
+        let rr = mode || true;
```
TheColaber, line 790: This makes it true no matter what

ErrorGamer2000: Why?

ErrorGamer2000: maybe try `mode??true`?

TheColaber: that will work

webpages/settings/index.js
```diff
@@ -787,8 +787,7 @@ const vue = (window.vue = new Vue({
     },
     setTheme(mode) {
       chrome.storage.sync.get(["globalTheme"], function (r) {
-        let rr = true; //true = light, false = dark
-        rr = mode;
+        let rr = mode ?? true;
```
WorldLanguages, line 790: We can use `?.` but not `??`, that's only Chrome 85+

WorldLanguages, line 790: lol `??` is chrome 80+, we almost cant use it

ErrorGamer2000: Oof

WorldLanguages:
Regression: 
image

mxmou:
@WorldLanguages That could be fixed by using the `addon-description-full` class.

Did this PR intentionally change the layout of the modal? It didn't use different classes for no reason.

GrahamSH-LLK:
> @WorldLanguages That could be fixed by using the `addon-description-full` class.
> 
> Did this PR intentionally change the layout of the modal? `settings-*` classes weren't used for no reason.

Sorry, I made an assumption that they were the same because they were when I designed it 😳

mxmou:
> Sorry, I made an assumption that they were the same because they were when I designed it 😳

I changed them quite a long time ago  (#1418), but they're still mostly the same. I think using the same classes is a good idea, but layout-related parts of the previous modal-specific styles should be kept.

PR #2539, opened 2021-05-19
Adds bookmarklet versions of a couple addons

brourbeinsus:
### Additions

This simply adds a bookmarklet version of a couple addons, I was only able to find 2 that worked correctly, editor stage left and account settings capitalize. Account settings capitalize is kind of useless as a bookmarklet but I thought I would add it just in case.

### Why did I add these?

Mostly for people on mobile devices who can’t download browser extensions and don’t have a developer console, so I went through to see what addons worked as a bookmarklet and found 2.


retronbv:
But how will this help?

RedGuy12:
No

WorldLanguages:
You may want to take a look at @Scratch-Bookmarklets

xItzDiamondx:
You know about us @WorldLanguages?

Shluffy:
> You know about us @WorldLanguages?

it used to be a scratch forum topic, but that got dustbinned, possibly because of the extension policy

PR #2552, opened 2021-05-21
Make Transparent orphan blocks support more blocks

TheColaber:
Resolves https://github.com/ScratchAddons/ScratchAddons/discussions/2551#discussioncomment-767955

### Changes

Adds 2 new setting options, plus the name change to support dragged blocks, and all other blocks in general.

### Reason for changes

> Snap (snap.berkeley.edu) solved on[e] of the more annoying Scratch problems, that is dragging a reporter block in the right argument, making the dragged block semitransparent.

### Tests

I have tested it, it works great.

### DIFF
--- a/addons/transparent-orphans/addon.json
+++ b/addons/transparent-orphans/addon.json
@@ -29,7 +29,7 @@
       "type": "integer",
       "min": 0,
       "max": 100,
-      "default": 0
+      "default": 25
     },
     {
       "name": "Dragged transparency (%)",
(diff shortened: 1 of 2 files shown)


ErrorGamer2000:
Sounds like a good idea!

WorldLanguages:
Is there any reason to support changing transparency for normal blocks?

ErrorGamer2000:
Actually, I quite like it. I have them at 75%, so they look mostly solid but I can see overlaps. That is probably the most useful part of it, seeing if one block is over another, like some people do when hiding code.

PR #2555, opened 2021-05-22
New addon - Ctrl+Click to run scripts

lisa-wolfgang:
This is a feature that was originally part of #2470.

### Changes

When the addon is enabled, scripts will only run when they are clicked while holding down the Ctrl key.

### Reason for changes

W_L approved

### Tests

Achieves its purpose without negative side effects.
Do your best to break it and let me know if you find anything.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -75,6 +75,7 @@
   "gamepad",
   "remove-curved-stage-border",
   "fix-pasted-scripts",
+  "ctrl-click-run-scripts",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/ctrl-click-run-scripts/addon.json
+++ b/addons/ctrl-click-run-scripts/addon.json
@@ -6,6 +6,11 @@
       "type": "notice",
       "text": "If you just want to fix the bug where placing a duplicated block into a script runs the script, you should enable \"Do not automatically run duplicated blocks\" instead.",
       "id": "redirectToDuplicateFix"
+    },
+    {
+      "type": "notice",
+      "text": "If you're on macOS, use the Cmd key instead of the Ctrl key.",
+      "id": "macContextDisabled"
     }
   ],
   "credits": [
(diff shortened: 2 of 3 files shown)


lisa-wolfgang:
Also, I apologize to whoever has to make the website thumbnail for this addon

TheColaber:
Im not sure why this isn't part of the fix addon, since this is only useful when that addon is enabled.

lisa-wolfgang:
> Im not sure why this isn't part of the fix addon, since this is only useful when that addon is enabled.

This addon actually doesn't depend on the duplicate fix addon.

TheColaber:
But this addon is useless, because you can already click on blocks to run them

lisa-wolfgang:
> ... scripts will **only** run when they are clicked while holding down the Ctrl key.

@TheColaber The point is to make running scripts require a more distinct input, since normal clicking is used for various other tasks. This would appeal to people who accidentally run scripts or who don't like the script running feature of Scratch.

TheColaber:
I'm still not entirely convinced. This addon basically makes it possible to ctrl click blocks, when clicking is still an option. ¯\\\_(ツ)_/¯

lisa-wolfgang:
> I'm still not entirely convinced. This addon basically makes it possible to ctrl click blocks, when clicking is still an option. ¯\_(ツ)_/¯

But it's not still an option when the addon is enabled. Read the description again lol

> ... scripts will **only** run when they are clicked while holding down the Ctrl key.

> ... scripts will **only** run ... Ctrl key.

> ... **only** ... Ctrl key.

> ... **only** ...

TheColaber:
Wait ... I think I forgot the "never" feature was removed

WorldLanguages:
TODO: handling ctrl+click in macs, where that triggers a right click (contextmenu)

lisa-wolfgang:
> TODO: handling ctrl+click in macs, where that triggers a right click (contextmenu)

@WorldLanguages Maybe switch to Alt+Click or Shift+Click?

WorldLanguages:
Alt+Click will cause problems in Chromebooks, and it's harder to workaround than Ctrl+Click in a Mac
Shift would work, but I still prefer Ctrl tbh, we should probably try to find a workaround instead

lisa-wolfgang:
Maybe Macs use Shift and everything else uses Ctrl? Or maybe make it an option.

CST1229:
Or maybe an option to make double clicking run scripts like early 1.x/0.x versions.

WorldLanguages:
Interesting, didn't know you had to double click in old versions

PR #2574, opened 2021-05-23
Fix addons not running on browser startup

WorldLanguages:
Surprisingly, my usual test case of opening a Scratch tab on startup with focus by default stopped working.
It looks like the tab is loading before we register `chrome.runtime.onMessage`, which is weird, one would expect extensions would be able to run sync operations before any tabs loading but whatever :P
Fix is ugly but lets us avoid edge cases like opening "a Scratch tab with an iframe on it on startup with focus". (edit: redid properly, tested succesfully in edgecase)
@mxmou Maybe this is what was causing `Unchecked runtime.lastError: The message port closed before a response was received.` for you?

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -268,6 +268,15 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
     return true;
   }
 });
+// In case a tab messaged us before we registered the event above,
+// we notify them they can resend the contentScriptInfo message
+chrome.tabs.query({}, (tabs) =>
+  tabs.forEach((tab) => {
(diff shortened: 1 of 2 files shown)


mxmou:
Unfortunately this did not fix the `lastError` - it's still happening. I don't know how the error could be debugged because Chrome doesn't provide a traceback.

WorldLanguages:
@mxmou Don't you have any more details on when this happens? And did you check the background page logs?

WorldLanguages:
Also, I'm assuming there's an error **and the addons don't load**? If it's only the first, then I don't see the problem

mxmou:
@WorldLanguages Addons don't load when the error appears. It appears to happen unpredictably and there are no errors in the background page console.

PR #2584, opened 2021-05-24
Improve compatibility with new studios

mxmou:
### Changes

Changes to `dark-www` and `scratchr2` to improve compatibility with the new studio page, and to `studio-tools` for compatibility with these addons.

Also makes `discuss-button` work correctly in development versions of `scratch-www`.


RedGuy12:
The new studio page will be released mid July.

(Source: LLK Tranisfex org announcements)

mxmou:
image

PR #2598, opened 2021-05-24
Fix reenabled event not launching when dynamicDisable is true

TheColaber:


WorldLanguages:
I'll do it a bit differently and merge tomorrow

WorldLanguages:
This fixes the "reenabled" event not firing for addons with `dynamicDisable:true` and `dynamicEnable:true`

PR #2612, opened 2021-05-25
load-vue-components: support <style> element

### LINKED ISSUE #2607: Allow vue templates to include css
It would be simple enough to append the styles to the page when it's loaded.

WorldLanguages:
Resolves #2607

Like #2610 but order is consistent in all executions


WorldLanguages:
souceURL commit above ^ I did mainly for fun and meemz
Not convinced this is the way to go, but if it is, now source maps work properly, including referencing the correct line number:
image
image
image

PR #2619, opened 2021-05-26
Expanding search bar addon

CST1229:
### Changes

Adds the "Expanding search bar" addon (expanding-search bar).
The addon expands the website search bar when you click on it:
example
(here demonstarted by my bookmarklet version)
I wanted to add dynamic disabling/enabling, but the events just didn't want to remove so I scrapped that.

One thing that might need looking at is the scratch-www search bar detection. I just use a static class check, and the class name makes it seem like it's gonna change.
And also it unnecessarily runs in the editor, but I'm already writing this when I noticed it and I don't know how to do negative matches (if they even exist, and there's a lot of pages with the search bar) so...

### Tests

Tested mostly on Firefox and a bit on Chrome. Should work with both styles of pages and the scratchr2 and discuss-button addons.

(this is my first PR so I don't know the standards very well)


addons/expanding-search-bar/addon.json
```diff
@@ -0,0 +1,22 @@
...
+  "userscripts": [
+    {
+      "url": "expanding-search-bar.js",
+      "matches": ["https://scratch.mit.edu/*"]
+    }
+  ],
+  "versionAdded": "",
+  "tags": ["community"],
+  "enabledByDefault": false,
+  "runAtComplete": true,
```
TheColaber, line 20: This belongs under `userscripts[0]`

addons/expanding-search-bar/expanding-search-bar.js
```diff
@@ -0,0 +1,69 @@
+export default async function ({ addon, global, console }) {
+
+	///scratch-www checking
+	var exsearch_isWWW; //True if scratch-www, false if not
+	if (document.getElementsByClassName("site-nav")[0]) {
+		exsearch_isWWW = false;
+	} else {
+		exsearch_isWWW = true;
+	}
```
TheColaber, line 9: You probably don't want to check once, since scratchr2 is an addon. I would use `waitForElement` and a while loop

addons/expanding-search-bar/expanding-search-bar.js
```diff
@@ -0,0 +1,69 @@
...
+			exsearch_NoDispIfExist("about"); //About
+			exsearch_NoDispIfExist("discuss"); //SA Discuss button compatibility
+		}
+		function exsearch_clickOut() { //Clicking out of  the search bar
+			exsearch_DelDispIfExist("create"); //Create
+			exsearch_DelDispIfExist("explore"); //Explore
+			exsearch_DelDispIfExist("ideas"); //Ideas
+			exsearch_DelDispIfExist("about"); //About
+			exsearch_DelDispIfExist("discuss"); //SA Discuss button compatibility
+		}
```
TheColaber, line 50: I'm sure you can do better than that... Just get all the `.link`s before the `.search`

addons/expanding-search-bar/addon.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Expanding search bar",
+  "description": "Expands the website search bar when you click on it.",
```
TheColaber, line 3: Maybe change this to `Expands the website search bar when your focused on it.`?

addons/expanding-search-bar/addon.json
```diff
@@ -0,0 +1,22 @@
...
+    }
+  ],
+
+  "userscripts": [
+    {
+      "url": "expanding-search-bar.js",
+      "matches": ["https://scratch.mit.edu/*"]
+    }
+  ],
+  "versionAdded": "",
```
TheColaber, line 17: you can probably set this to `1.16.0`

TheColaber (changes requested):
Nice job on your first addon!

addons/expanding-search-bar/addon.json
Shluffy: wouldn't that need hover then?

GrahamSH-LLK:
Have you checked if this is possible with `:focus` + fun CSS?

addons/expanding-search-bar/expanding-search-bar.js
```diff
@@ -0,0 +1,72 @@
+export default async function ({ addon, global, console }) {
+
+	///scratch-www checking
+	var exsearch_isWWW; //True if scratch-www, false if not
+	if (document.getElementsByClassName("site-nav")[0]) {
+		exsearch_isWWW = false;
+	} else {
+		exsearch_isWWW = true;
+	}
```
GrahamSH-LLK, line 9: Don't we have an API for this????

CST1229:
> Have you checked if this is possible with `:focus` + fun CSS?

Not yet.

addons/expanding-search-bar/expanding-search-bar.js
GrahamSH-LLK: `addon.tab.clientVersion` should work, unless I'm misunderstanding what this does.

addons/expanding-search-bar/expanding-search-bar.js
CST1229: Assuming you're talking about the scratch-www detection, changed the script to use the client version API.

addons/expanding-search-bar/expanding-search-bar.js
```diff
@@ -0,0 +1,61 @@
...
+		//Functions
+		function exsearch_getLinks() { //Gets all header links
+			let e = document.getElementsByClassName("link")[0]; //The first link
+			let list = []; //The list
+			
+			while (e.classList.contains("link")) { //Repeat until the element is NOT a link
+				list.push(e); //Add
+				e = e.nextSibling; //Next 
+			}
+			return list;
```
TheColaber, line 22: You can just query `.link:not(.search ~ *)` and probably just switch to using css

addons/expanding-search-bar/expanding-search-bar.js
CST1229: So should I wait for the search bar to load before doing the stuff after?

TheColaber:
@CST1229 When I meant
> probably just switch to using css

i mean delete your js file and turn it into css... It probably very simple

CST1229:
> 
> 
> @CST1229 When I meant
> 
> > probably just switch to using css
> 
> i mean delete your js file and turn it into css... It probably very simple

I think that would be pretty hard, since I have to style a _previous_ element if another element is focused...

mxmou:
I'm not sure if there is a non-hacky way to do this with just CSS.

addons/expanding-search-bar/expanding-search-bar.js
CST1229: Anyways, waited to load the search bar.

addons/expanding-search-bar/expanding-search-bar.js
CST1229: Also done.

addons/expanding-search-bar/expanding-search-bar.js
CST1229: Also done. I don't think switching entirely to a style would be simple.

WorldLanguages (approved):
@CST1229 Good job!
Leaving open in case someone wants to suggest a better addon name/description
I felt "expanding search bar" was not very easy to translate to other languages because of the weird (imo) use of "expanding". So I changed it to "expandable search bar".
I also changed the description to `Makes the search bar on the navbar wider when it's focused.`, instead of `Expands the website search bar when it's focused.`

Issue #2531, opened 2021-05-19
Settings page takes way too long to load


GarboMuffin:
### Description

The settings page can take seconds to load. There is no reason for it to take this long.

### Reproduction steps

1. Open the settings page

### Expected behavior

It should very quickly, especially as it should not need to make any network requests. First paint should happen in under a second at worst.

### Actual behavior

Tested on a several year-old Chromebook:

 - With all addons disabled, the settings page takes more than **three seconds** to load.
 - With all addons enabled, the settings page takes more than **five seconds** to load.

These are dangerously close to or past the point where a majority of users will give up and not bother. These underpowered devices probably represent a very significant chunk of Scratch Addons users.
 
### Environment

 - Scratch Addons version: Release 1.14.3
 - Browser: any, worse in Firefox
 - Operating system: any

apple502j:
On Firefox 88, with Windows 10 and Intel Core i5 2.2GHz CPU and 8GB of RAM, `PerformanceNavigationTiming.duration`(aka time took to finish firing onload) is 7600ms, with all addons enabled.

Here are some data:
- setLang (indicating load time of page) to index.js load: 100ms. This is probably DOM parsing.
- setLang to index.js end: 440ms
- setLang to vue ready() function call: 440ms
- setLang to chrome.runtime.onMessage call (when data is received from background): 560ms
- PerformancePaintTiming started at 7400ms.

It took around 6-7 seconds to render the DOM, apparently.

apple502j:
Duplicate of #2090

GarboMuffin:
This is not a #2090 duplicate. Even after completely removing the color picker, it still takes the same amount of time to load. It seems like #2090 has been fixed already.

WorldLanguages:
Not color picker. And I can confirm it being worse in Firefox for some reason.
We could probably try to optimize little pieces of our Vue logic that probably waste a few hundred of milliseconds, but not sure if that's the main cause of this.
It taking more time when all addons are enabled probably means that the `<addon-body>` elements being expanded adds to the initial DOM build time

WorldLanguages:
If the main overhead is **painting** every `<addon-body>` element, including those out of initial view, `content-visibility: auto` could help (but Chrome only for now I think).
Not sure if converting the addon list into a virtualized list would be a good idea or not

WorldLanguages:
[...]

Issue #2339, opened 2021-05-01
Enhance addon search by keywords on settings page


WorldLanguages:
Jeffalo has suggested using a library like [Fuse.js](https://fusejs.io/) that will let us support search properly - sort addons by priority, allow addons to include keywords a user might use to search for it in the manifest (not sure how localization will play with that tho), among other things.
For example, right now, searching for "smart block switching" returns no results, even tho an addon is named "block switching".

PR #2650, opened 2021-06-01
Optimize the settings page for relevance-ordered search

### LINKED ISSUE #2654: Component styles override `light.css`
### Description

Caused by #2612

Light mode on the settings page doesn't work properly because `light.css` is inserted into the `<head>` before component styles. This doesn't happen when switching to light mode after page load.

### Repoduction steps

1. Open the settings page.
2. If you're using dark mode, switch to light mode and reload.

### Expected behaviour

Icons and the text of addon descriptions would be black.

### Actual behaviour

Icons and addon descriptions are hard to see.

WorldLanguages:
Work in progress.
Related to #2339 - trying to get ready for ordered search
Related to #2531 - not really trying too hard to make stuff faster other than making use of the virtual DOM, but instead "load in steps", no need to wait until everything is ready until the UI can be shown

- Reuse the same Vue components for both displaying addon groups and search, for better performance
- Reuse the same DOM elements when addon order changes, instead of destroying and creating again
- Show placeholder when loading data:
image

TODOs:
- [x] Resolves #2654
- [x] Whoops I broke search midway through, sort works but unmatched addons still visible
- [x] Fix addon group margin bottom
- [x] Fix ^ not working properly sometimes: https://github.com/ScratchAddons/ScratchAddons/pull/2650/commits/6d5fbb26625a9c8097a9a8d3e678b5bf36d61d12#r646010296
- [x] Fix "thumb" not matching "thumbnail setter" (weird)
- [x] Fix bug shown by graham
- [x] Fix \<addon-body> expanded not working properly
- [x] Readd "open full screen settings page" modal in popup
- [x] Keep old easter egg logic
- [x] Wait until search bar isn't typed on for ~250ms before trying to display results edit: now 150ms
- [x] Fill `addonListObjs` with example pseudoaddon after full event loop (for faster first paint)
- [x] Maybe, move some stuff to their own component (edit: should be down in another PR)
- [x] pointer-events none in addon list when displaying example manifest (to avoid user enabling a pseudoaddon and weird stuff like that)
- [x] More optimizations? Such as rendering one addon group at a time for smoother loading (edit: discarded for now. not implemented)
- [x] Make sure `#addon-addon-id` still works
- [x] Fix search input autofocus madness
- [x] Fix disabled addon search in popup (regression by this PR)


GrahamSH-LLK:
That placeholder is TRULY ugly. Maybe a little more `border-radius` and accent-4? Adding a skeleton loader would also be cool.

mxmou:
I made some changes to the placeholder:
* The addon name placeholder is now smaller and rounded:
image
* It now has the correct color in light mode
* There is now padding above the placeholder to fill the space normally occupied by the first group header
* Placeholder addons are now enabled because enabled addons will be shown at the top after they're loaded

webpages/settings/style.css
```diff
@@ -99,6 +99,10 @@ h1 {
   height: 100vh;
 }
 
+.addons-container.placeholder {
+  padding-top: 43px;
+}
+
```
WorldLanguages, line 106: Instead of doing this, we could directly show the `<addon-group-header>` for the "new" group, which will probably end up being the first div in the container anyway (at least in stable releases)

WorldLanguages: This is actually a UI/UX decision, not a technical one. UI/UX wise I prefer this. But just mentioning we could prerender the header for the new group instead of this just fine

mxmou: Yet another option is to show an empty group header.

GrahamSH-LLK:
image
Why the extra placeholder addons?

mxmou:
> Wait until search bar isn't typed on for ~250ms before trying to display results

I usually use search to find addons quickly in the popup, so showing results immediately is important in some cases.

WorldLanguages:
> I usually use search to find addons quickly in the popup, so showing results immediately is important in some cases.

Sure, but blocking the main thread isn't any better. It will feel laggy and even worse, the user won't be able to see what they've typed in realtime, if their computer is slow enough

webpages/settings/style.css
```diff
@@ -51,8 +51,8 @@ h1 {
   align-items: stretch;
 }
 
-.addon-group {
-  margin-bottom: 20px;
+.addon-group:not(:first-child) {
+  margin-top: 20px;
```
WorldLanguages, line 55: This will not work properly, because a group header might be hidden with `display:none`. We'd want to target the first visible child, not the first child.

mxmou:
I think the 250 ms delay is too long. Anything loner than about 100 ms feels slow.

WorldLanguages:
Let's do 150 ms for now then :P

WorldLanguages:
Keep in mind this PR caused (as a bug) that enabled by default addons are not expanded by default
So should the placeholders look like this in fullscreen mode...?
image

mxmou:
@WorldLanguages I don't think "Credit to: Credit" should be there. I would make them the same as before, but with the expand icon reverted.

WorldLanguages:
Do you think we could keep the `Credit to: Credit` but hide it with CSS visibility prop?

mxmou:
> Do you think we could keep the `Credit to: Credit` but hide it with CSS visibility prop?

I think the placeholder looks better if there isn't too much empty space.

WorldLanguages:
Keep in mind some browsers might not have the time to load the images at all if the main thread is blocked at the correct time

image

WorldLanguages:
Also, I forgot the first ~4 addons in the settings page will be the new category, and it's most likely not all first 4 of them will be enabled. So maybe we shouldn't even revert the expand icon. Just keep placeholder addon unexpanded

webpages/settings/style.css
WorldLanguages: Lol there are so many bad ways to go about this and I don't know which one to start with

mxmou: Is hiding the element by removing it instead of `display: none` also a bad way?

webpages/settings/data/example-manifest.js
```diff
@@ -0,0 +1,16 @@
+export default {
+  name: "",
+  description: "",
+  credits: [
+    {
+      name: "Credit",
```
apple502j, line 6: Can't we remove this or set it to blank (not "Credits")?

WorldLanguages: Yes, it can be removed. I will remove it later today.

webpages/settings/style.css
WorldLanguages: Solved

PR #2651, opened 2021-06-01
Replace editor dark mode userscript with `customCssVariables`

mxmou:
### Changes

Removes the editor dark mode userscript and uses `customCssVariables` to set the values instead. Other changes related to CSS variables:
* Makes the text on hovered inactive tabs transparent to match Scratch design
* Makes inactive tab icons brighter when the background is dark to match the text color 
* Styles modal back and close icons correctly
* Fixes the sound icon of the selected sound using `selector2-filter` instead of the correct `selectorSelection-filter`
* Makes the block palette affected by the border color setting

### `addon.json` format changes

* The `black` and `white` properties of `"type": "textColor"` objects are now sometimes omitted, so they should be marked as optional in the schema. They can now be objects with the same format as `value` or `source`.
* The `value` of a custom CSS variable or any other object with the same format can now be a hex color.
* There is a new `type`, `"alphaBlend"`, with two required properties: `opaqueSource` and `transparentSource`.

### Reason for changes

To make loading colors faster.


addons/editor-dark-mode/addon.json
```diff
@@ -55,6 +600,19 @@
         "value": true
       }
     },
+    {
+      "url": "hide_dots.css",
+      "matches": [
+        "https://scratch.mit.edu/projects/*/",
+        "https://scratch.mit.edu/projects/*/editor",
+        "https://scratch.mit.edu/projects/*/fullscreen",
+        "https://scratch.mit.edu/projects/editor"
```
WorldLanguages, line 609: You could simply use `projects` word match

mxmou: I think it's better to do that for all editor addons at once.

WorldLanguages (approved):
Good
I've never edited the schema so I doubt this will be my first time doing it :P You can merge this before the schema is updated if you wish

mxmou:
@WorldLanguages I want to make another addition to `customCssVariables` soon anyway, so I'm merging this PR.

PR #2667, opened 2021-06-04
Redirect to main forums if on mobile forums

### LINKED ISSUE #2664: Redirect mobile forums to regular forums
Exactly what it sounds like. Google indexes mobile forums a lot, can get annoying.

FunctionalMetatable:
Resolves #2664

### Changes

Adds a redirect if on mobile forums addon.

### Reason for changes

Google.

### Tests

Tested Locally


WorldLanguages (approved):
OK, but as an easter egg addon, **for now**.
In the future, we should find a better way to showcase addons like this and account-settings-capitalize, so that they are hidden but not _that hidden_.

Issue #972, opened 2020-12-07
Activity Status Addon


Mast3rGenius:
I looked through (unofficial) rejected suggestions and issues and didn't see this anywhere, which surprised me, but I really think that this would be a really nice feature to have. If you don't know what I mean by this, basically it would put an "Active Now" status on peoples' profiles who had scratch addons if they are on scratch.

Mast3rGenius:
Update: Checked again and found the Should there be an IsOnline addon. I really think this would be a useful feature.

jeffalo:
#342 #504

WorldLanguages:
#504

Issue #342, opened 2020-09-27
Should there be an isOnline addon?


JeffaloBob:
so i remember back when isOnline was at its peak and it was super useful to tell if someone was online.

i was wondering, maybe there could be a super simple iO implementation (probably not with the recommended tag and with a new name) which wouldn't have any of the friends list features.

basically, ping server every few minutes with username (or maybe user id), and on the server, if someone has not pinged in x minutes, then they are offline.

then when visiting a profile, fetch an api from the server and decide whether they are online, offline, or not using the extension. 

isonline had some other features that let you know if your friends were online, but something like that could cause privacy concerns blah blah. it also had a ghost mode, which would also be fairly simple to implement in this simple iO.

i think even i know how to make a simple iO implementation and if you want i can make the PR to add it.

_there might also have to be a privacy policy and stuff like that 😬_

JeffaloBob:
i made a demo server on repl.it

https://io-meme.jeffalobob.repl.co/ will display how many users are online
https://io-meme.jeffalobob.repl.co/online?user=username will make a user online for 10 seconds
https://io-meme.jeffalobob.repl.co/isonline?user=username will check if they're online

JeffaloBob:
and the code https://repl.it/@jeffalobob/io-meme

jemunm378:
I think it’s a great idea.

tb148:
This is AWESOME! However we should also have a "non-SA user" status so people don't get confused why online users have an "offline" status.
This might be done by setting an `isUser` status the first time they ping.

YOYITsM3M8:
Yas

JeffaloBob:
we will also have to handle verification if the user pinging is really them. any ideas on how to implement this?

JeffaloBob:
@WorldLanguages ^^

jemunm378:
Maybe the users can verify using their scratch account. Maybe something similar to this?: [https://github.com/ScratchVerifier/ScratchVerifier](url)

WorldLanguages:
Keep in mind, we won't be adding any online indicator/notifications for now, while we focus in other parts of the extension.

JeffaloBob:
i can focus though

WorldLanguages:
I need to focus on whether your focus is appropriate :P

JeffaloBob:
i was thinking of something that may be better, instead of displaying offline users, only online users are displayed. 

this makes it a bit simpler to code and allows focus or something idk just let me make it

[...]

Issue #184, opened 2020-09-03
[New addon] Markdown support for comments


NoobTracker:
**Describe the new addon**

As you might know, comments on Scratch don't have any sort of formatting options. Besides using unicode fonts but you know what I mean. Now how should that work?

When you write a comment, there would be a markdown switch next to the submit button. When you enable that switch, you get those buttons you also have here on GitHub. At the same time, it sets something up so some nonsense will be added to your comment to show that it uses markdown, an invisible unicode character, or maybe even a visible one since that extension will just hide it for everyone who has it enabled.
For file uploading Jeffalo's image uploader would be used.
The addon could add support for blank lines.

There would be options to enable Markdown for comments by defauls and for interpreting Markdown even if there is not that character.

Why Markdown and not BBCode, you might ask. The reason is simple: Markdown uses less characters and since comments have a limit that's kinda neat.

**Link to the new addon**

Doesn't exist so far.

Hans5958:
ST wouldn't like those hacky things done on the website. Just look what happens on the first version of isOnline a while ago.

NoobTracker:
Well, adding the support for _this_ and **this** isn't exactly a really hacky thing - it's like adding emoji support if the comments had none - anyone also without the extension can read it, but those with it can do it better.

Hans5958:
I still afraid, but I don't know, I'm a guy who is not using Scratch for a while. Probably @WorldLanguages knows better than me.

NoobTracker:
As I said, that's like replacing :) with that emoji so not really extrem stuff. Yep, let's ask @WorldLanguages .

NoobTracker:
Has anyone an idea where you can get a markdown interpreter on GitHub?

NoobTracker:
Okay, I asked Za-Chary.

NoobTracker:
Also, has anyone an idea how to let the script run for every comment, also when you load new ones? @JeffaloBob

YOYITsM3M8:
Forever loop?

NoobTracker:
I'm not sure if that's a good idea ... it would be better when it just executes once as soon as a new element that belongs to a certain class is loaded.

WorldLanguages:
> ST wouldn't like those hacky things done on the website. Just look what happens on the first version of isOnline a while ago.

Unrelated.

WorldLanguages:
Anyway, please no. This is useless if only a few people are able to see the markdown.

[...]

PR #2676, opened 2021-06-05
Add OneState

### LINKED ISSUE #2076: i s O n l i n e
# isonline
## disclaimer
this has been discussed in discord but I am opening this anyways.
## how
if the user turns on the addon, show online / offline
get data from a scratch project that proxies requests to a server
add an option to share online / offline
this option would also go through the proxy project
## why
i s O n l i n e
## by the way
- I have a server that we could use that is open source

wgyt:
Resolves #2076 #972 #184 #342 #1820 and #504

### Changes

Added an addon called OneState, that is like isOnline but with rich presence as an option

### Reason for changes

Add isOnline

### Tests

It works. Tested on Chrome 91 Windows 10, and on Chrome OS 90


apple502j:
Rejected (at least for Rich Presence part): https://discord.com/channels/751206349614088204/755131743136383107/850489773696221245

wgyt:
@apple502j Would it be approved if I remove rich presence?

PR #2677, opened 2021-06-05
OneState No RPC edition

wgyt:
Supercedes #2676

Same thing but without rpc and stuff


WorldLanguages:
Probably won't make it to stores - if we ever provide addons that a) need to be enabled by default to be worth adding and b) need a central server, it will need to be a coordinated effort and be hosted under the scratchaddons.com domain.

RedGuy12:
Probably it's own extention

wgyt:
> Probably it's own extention

@RedGuy12 yeah.... that is what I was thinking, currently working on that at the moment, but I'll keep this open for the time being

wgyt:
Note: Working on a seperate chrome extension for OneState with RPC:
image

RedGuy12:
is it onedotprojects/onestate-extention?

wgyt:
> is it onedotprojects/onestate-extention?

no, its wgyt/onestate-extention

wgyt:
wont happen, might as well close

PR #2682, opened 2021-06-06
Create TurboWarp Player Addon

### LINKED ISSUE #164: Turbowarp player in editor (player)
This replaces the existing player with turbowarp by grabbing the id from url and embedding turbowarp where the normal player would be

**Link to the new addon**

[here](https://github.com/YOYITsM3M8/ScratchAddons)

TheColaber:
Resolves #164

### Changes

Creates a button on project pages that lets you toggle from tw to scratch player or open a new tw tab.

### Tests

This happens after seeing inside :/
image
The iframe size is not 100% fine yet


addons/turbowarp-player/userscript.js
```diff
@@ -0,0 +1,62 @@
...
+    const row = await addon.tab.waitForElement(".preview .project-buttons", {
+      markAsSeen: true,
+      reduxEvents: ["scratch-gui/mode/SET_PLAYER", "fontsLoaded/SET_FONTS_LOADED", "scratch-gui/locales/SELECT_LOCALE"],
+      reduxCondition: (state) => state.scratchGui.mode.isPlayerOnly,
+    });
+    row.prepend(button);
+
+    scratchStage = document.querySelector("[class^='stage-wrapper_stage-wrapper']");
+
+    twIframe.src = "//turbowarp.org/" + window.location.pathname.split("/")[2] + "/embed";
```
GarboMuffin, line 42: could also send `&username=...` to control the username block

GarboMuffin (changes requested):
I'm still not fond of the idea of putting TurboWarp into an iframe. Too many hard or impossible to solve issues.

Very selfish request: make "Open tab" the default mode. Or don't. It's not really my choice to make.

addons/turbowarp-player/userscript.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+    } else {
+      window.open("//turbowarp.org/" + window.location.pathname.split("/")[2], "_blank");
+    }
+  };
+
+  while (true) {
+    const row = await addon.tab.waitForElement(".preview .project-buttons", {
+      markAsSeen: true,
+      reduxEvents: ["scratch-gui/mode/SET_PLAYER", "fontsLoaded/SET_FONTS_LOADED", "scratch-gui/locales/SELECT_LOCALE"],
```
GarboMuffin, line 45: remove reduxEvents - Scratch will recreate elements when window is resized to certain levels without emitting redux events. reduxCondition is fine.

addons/turbowarp-player/userscript.js
```diff
@@ -0,0 +1,66 @@
...
+  button.onclick = (event) => {
+    if (addon.settings.get("action") == "player") {
+      playerToggled = !playerToggled;
+      if (playerToggled) {
+        twIframe.style.display = "";
+        scratchStage.style.display = "none";
+        button.classList.add("scratch");
+        if (!wasEverToggled) {
+          twIframe.src =
+            "//turbowarp.org/" + window.location.pathname.split("/")[2] + "/embed?username=" + addon.auth.username;
```
GarboMuffin, line 28: seems to send ?username=null when signed out

addons/turbowarp-player/userscript.js
```diff
@@ -0,0 +1,66 @@
...
+  let playerToggled = false;
+  let wasEverToggled = false;
+  let scratchStage;
+  let twIframe = document.createElement("iframe");
+  twIframe.style.border = "none";
+  twIframe.style.display = "none";
+  twIframe.scrolling = "no";
+  twIframe.setAttribute("allowtransparency", "true");
+  twIframe.setAttribute("allowfullscreen", "true");
+  // The stage is "bouncing" after switching
```
GarboMuffin, line 11: Not a complete solution, but

```
position: relative;
right: 8px;
bottom: 3px;
```

at least makes the stage itself not move

GarboMuffin (changes requested):
other feedback:

 - In replace player mode, if Scratch either Scratch or TurboWarp are running, it keeps running when mode is switched. That's not ideal.
   - For Scratch you could stop or pause the project
   - For TurboWarp you could remove the iframe from the DOM instead of just hiding it

WorldLanguages:
> This happens after seeing inside :/

Will be solved by #2852

addons/turbowarp-player/userscript.js
ErrorGamer2000: try
```js
          ...
          twIframe.src =
            `//turbowarp.org/${window.location.pathname.split("/")[2]}/embed${addon.auth.username ? `?username=${addon.auth.username}` : ""}`;
          ...
```

FunctionalMetatable:
Can there be the ability to load from different mods, e.g forkphorus?

FunctionalMetatable:
+ For TurboWarp, could there be a setting for enabling 60fps/high quality pen?

WorldLanguages:
TODOs:

- [ ] We'll keep "new tab" as the default, so I don't see the point on having auto replace. The feature should be removed altogether.
- [ ] `div.project-info-alert` should be hidden when on iframe mode ("this project uses username block" or "you cannot use block data", sometimes both)
- [ ] Let's make the iframe background a bit lighter or darker (depending on whether website dark mode is enabled) when on iframe mode. Plus border-radius.
- [ ] Iframe sizing needs to be enhanced.
- [ ] We should probably ask the user to confirm that they want to go inside the project when there's an iframe, since it will be destroyed (and maybe the other way, since the project will be stopped)
- [ ] Good addon description and notices, so that it's clear that 1) addons won't work in iframe mode 2) turbowarp will get your username in iframe mode 3) it only works on player mode, not editor mode.

Daniel4-Scratch:
idea: if the project notes contains a turbowarp link with settings the button redirects to that one

addons/turbowarp-player/userscript.js
```diff
@@ -0,0 +1,59 @@
...
+  let scratchStage;
+  let twIframe = document.createElement("iframe");
+  twIframe.scrolling = "no";
+  twIframe.setAttribute("allowtransparency", "true");
+  twIframe.setAttribute("allowfullscreen", "true");
+  twIframe.className = "sa-tw-iframe";
+
+  const button = document.createElement("button");
+  button.className = "button see-inside-button sa-tw-button";
+  button.title = "TurboWarp";
```
FunctionalMetatable, line 12: l10n?

addons/turbowarp-player/userscript.js
```diff
@@ -34,11 +34,23 @@ export default async function ({ addon, console }) {
...
+    const seeInside = await addon.tab.waitForElement(".see-inside-button", {
       markAsSeen: true,
       reduxCondition: (state) => state.scratchGui.mode.isPlayerOnly,
     });
+
+    seeInside.addEventListener("click", function seeInsideClick(event) {
+      if (!playerToggled || !showAlert) return;
+
+      if (confirm("Are you sure you want to see inside? Doing this will destroy the TurboWarp player.")) {
+        showAlert = false;
```
WorldLanguages, line 48: What does `showAlert` do here? Is this to intentionally not ask the user a second time if they want to destroy the player?

WorldLanguages: @TheColaber Please explain this

WorldLanguages:
@mxmou Could you 1) tell me if addon name is OK 2) tell me if addon description is OK 3) tell me what you think about the `background-color` applied to `.sa-tw-iframe`, both in light mode and dark mode (currently both `rgba(0, 0, 0, 0.1)`)?

addons/turbowarp-player/userscript.js
TheColaber: Correct

mxmou:
> @mxmou Could you 1) tell me if addon name is OK 2) tell me if addon description is OK 3) tell me what you think about the `background-color` applied to `.sa-tw-iframe`, both in light mode and dark mode (currently both `rgba(0, 0, 0, 0.1)`)?

I like the title and description, but like #2799 changes the remix tree button addon, I think the project page should not be mentioned in the title - maybe add all information about the button's location to the description instead.

In notices, the "replace player" mode should probably be called *the "replace player" mode*, not just *"replace player" mode*. I'm not sure if "embed" can be used as a noun - maybe "replaces the Scratch player with an embedded TurboWarp player"?

I don't think the background is necessary, and it doesn't fit with Scratch's UI. I think it's better to put the iframe in a container with `overflow: hidden` and add a negative margin to align it with the rest of the page. If the background is important for some reason, a better color choice might be white in light mode and `#282828` in dark mode, with a border. It could also be styled similarly to the project description, with a transparent blue background and border.

addons/turbowarp-player/userscript.js
WorldLanguages: Okay then, keeping this logic

WorldLanguages:
TODO: English grammar check for notices

WorldLanguages:
English notices approved by @TheColaber and Grammarly

WorldLanguages (approved):
Merging tomorrow

Daniel4-Scratch:
> Merging tomorrow

Nice

mxmou:
This iframe background looks good, but I still think there should be at least an option to make the TurboWarp stage the same size as Scratch's, with no background.

retronbv:
> Merging tomorrow

Frick yea my addons idea is being added

WorldLanguages:
> This iframe background looks good, but I still think there should be at least an option to make the TurboWarp stage the same size as Scratch's, with no background.

I don't think there's a way to do that. Keep in mind the margin you see between the stage and the border is not artificially added by this addon. It's margin inside the iframe. So we can't make it smaller.
We do make the iframe 2px less wide and 2px less tall for the border, tho. But we don't add any more margins. We can't keep the stage the exact same size.

mxmou:
> > This iframe background looks good, but I still think there should be at least an option to make the TurboWarp stage the same size as Scratch's, with no background.
> 
> I don't think there's a way to do that. Keep in mind the margin you see between the stage and the border is not artificially added by this addon. It's margin inside the iframe. So we can't make it smaller.
> We do make the iframe 2px less wide and 2px less tall for the border, tho. But we don't add any more margins. We can't keep the stage the exact same size.

We can. I'll make a PR when I have time.

PR #2684, opened 2021-06-06
SECURITY: Sanitize HTML returned from ScratchDB

apple502j:
Defense in depth; not actual vulnerability.

### DIFF
--- a/addons/forum-search/userscript.js
+++ b/addons/forum-search/userscript.js
@@ -5,6 +5,22 @@ let currentQuery;
 let currentSort = "relevance";
 let locationQuery = "";
 
+const ALLOWED_TAGS = {
+  A: ["href"],
+  BR: [],
+  BLOCKQUOTE: [],
+  DIV: ["class", "style"],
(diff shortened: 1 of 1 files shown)


jeffalo:
unrelated, but you can/should remove the `cleanPost` function. it was a temporary fix in ocular because scratchindexer was vulnerable to xss. thats fixed now.

apple502j:
Well, that function will stay.

WorldLanguages:
@jeffalo Realistically there has never been and there will never be XSS, but it's still _potential_ XSS, if something crazy happens like Cloudflare getting hacked.

apple502j:
Or more realistically, Mozzy's server gets hacked (e.g. domain expired)

WorldLanguages (approved):

WorldLanguages:
We may be able to use the [window.Sanitizer](https://developer.mozilla.org/en-US/docs/Web/API/Sanitizer) API in the future.

PR #2687, opened 2021-06-06
Allow string min/max length to be specified by addons

### LINKED ISSUE #2639: Broken Discuss Button When The Text is Too Long

Yep. When you enable the discuss button, you can name it whatever you want, even a randomly generated password with 256 characters that makes the other buttons faze off the screen.

Have a limit to the length of the name of the discuss button.

### Additional context

broken discuss button scratchaddons

### LINKED ISSUE #2679: Bug when discuss field is left empty
### Description
When you leave the discuss field empty on the addon _Discuss Button_, it shows the word "Discuss" but upon reloading it actually is just empty and apparently has a margin on top as everything else is pushed downwards.

### Repoduction steps
1. Search "discuss"
2. Click _Discuss button_
3. Turn it on without typing anything in the textbox

### Expected behaviour
The button should have replaced Ideas and said "Discuss".

### Actual behaviour
The button replaces Ideas, but it's empty and the other buttons are all pushed out of place.

### Screenshots
Screen Shot 2021-06-05 at 3 00 40 PM
Screen Shot 2021-06-05 at 3 00 47 PM

### Environment

- Browser: Chrome
- Scratch Addons version: around v1.15.0
- Operating system: Apple

apple502j:
Resolves #2679 
Resolves #2639 
Schema already updated

### DIFF
--- a/webpages/settings/components/addon-setting.html
+++ b/webpages/settings/components/addon-setting.html
@@ -38,13 +38,15 @@
       type="text"
       class="setting-input string"
       v-model="addonSettings[addon._addonId][setting.id]"
-      @input="updateSettings()"
+      @change="checkValidity() || updateSettings()"
       :disabled="!addon._enabled"
       v-if="setting.type === 'string'"
       :placeholder="setting.default"
       :maxlength="setting.max || 100"
       :minlength="setting.min || 0"
       :required="!!setting.min"
+      :data-addon-id="addon._addonId"
+      :data-setting-id="setting.id"
     />
     <input
       type="text"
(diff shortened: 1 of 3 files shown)


webpages/settings/components/addon-setting.html
```diff
@@ -38,11 +38,15 @@
...
       class="setting-input string"
       v-model="addonSettings[addon._addonId][setting.id]"
-      @input="updateSettings()"
+      @change="checkValidity() || updateSettings()"
       :disabled="!addon._enabled"
       v-if="setting.type === 'string'"
       :placeholder="setting.default"
-      maxlength="100"
+      :maxlength="setting.max || 100"
+      :minlength="setting.min || 0"
```
WorldLanguages, line 46: Default should probably be 1?

apple502j: Maybe, not high prio as discuss button is the only use for this

PR #2688, opened 2021-06-06
Fix addon body CSS

### LINKED ISSUE #2681: Better unexpanded <addon-body> CSS
Making an issue instead of fixing myself bc I'm bad at CSS lol
These have been bothered me for a while now

1. If long addon names get wrapped, they should not be taking more width than they need to. This issue also causes some tags to wrap, like "for editor" in the screenshot. Fixing this kills 2 birds with one stone.
2. Addon descriptions should not have a max width. They should all reach the same point with an ellipsis (unless those that are shorter and couldn't fill the space, of course)

How it looks now:
image

How it should look: **(except first addon name should wrap, if I could figure out how to do that, I'd PR instead :P)**
image

Alternatively to the addon name wrapping, we could make it never wrap, and instead make the description shorter

css lords welp

mxmou:
Resolves #2681

### Changes

* Removes `max-width` from collapsed addon descriptions
* Tags no longer wrap
* Addon names no longer wrap until the window is so small that the description becomes invisible

### DIFF
--- a/webpages/settings/components/addon-body.html
+++ b/webpages/settings/components/addon-body.html
@@ -82,10 +82,12 @@
   .addon-description {
     margin-inline: 10px;
     color: gray;
-    max-width: 35vw;
     overflow: hidden;
     white-space: nowrap;
     text-overflow: ellipsis;
+    width: 0; /* don't include the description
+                 when calculating addon-body size */
+    flex-grow: 1;
   }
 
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Regression:
image

Should wrap like previously, or the black background should cover all the text.

PR #2699, opened 2021-06-07
Add two new settings to editor dark mode

### LINKED ISSUE #2415: Bring back the dark lists & variables but as an option

Yes! As to update 1.14, It resolved the stage being affected by dark mode, but to me, this seemed intentional and I'm fond of it.

An option to switch on/off dark mode impacting the stage.

Better3- nah haven't tried any alternatives.

**Additional context**

I thought it was intentional :,D

mxmou:
Resolves #2415
Adds a feature described in https://github.com/ScratchAddons/ScratchAddons/pull/1805#issuecomment-827611354

### Changes

* Adds the Boolean setting described in #2415
* Adds a setting to change the color of blue text and icons (ideally the tool buttons in the paint editor would also be affected but those icons also use other colors and I'm not sure if there is an SVG filter that would only affect a specific color)
* Makes the selection background and handles in the sound editor affected by the highlight color setting
* Removes now unused camera modal styles

### `addon.json` change

There is a new type for custom CSS variables, `"recolorFilter"`, which returns a filter function that can be used to change the color of monochrome icons.

### Reason for changes

To allow more customization.

### Tests

Tested with different values of both settings.

### DIFF
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -614,6 +614,19 @@
       }
     },
     {
+      "url": "stage.css",
+      "matches": [
+        "https://scratch.mit.edu/projects/*/",
+        "https://scratch.mit.edu/projects/*/editor",
+        "https://scratch.mit.edu/projects/*/fullscreen",
+        "https://scratch.mit.edu/projects/editor"
+      ],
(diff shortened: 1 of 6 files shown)


WorldLanguages (approved):
Can be merged

PR #2700, opened 2021-06-07
scratch-messaging: linkify mentions and external URLs

### LINKED ISSUE #1701: "More links" addon should also work in the extension popup

If someone posts a link in a comment, and I view it in the messaging popup, I need to open the comment in a new tab to go to that link. (The link isn't clickable!)

image

The "More Links" addon should also work in the popup.

None yet.

**Additional context**
something like that:

image

### LINKED ISSUE #2566: scratch-messaging: linkify links from JSON API comments
title
We might also linkify links to external sites. It's fine since we "own" the popup, not Scratch

apple502j:
Resolves #1701
Resolves #2566 

Pingifiers are tested, but did not test the code for the addon. Note that linkifying is optional and pingifying is not. So if linkifying is turned off, Scratch URLs will not be linkified for JSON but will be for HTML. (Didnt mess with that)

Also some changes to prevent triple serialization (getting HTML, fixCommentContent, and popup) of HTML. Also no more raw HTML manipulation in fixCommentContent.

@WorldLanguages what's preventing us from moving all these logics to popup?

### DIFF
--- a/addons/scratch-messaging/background.js
+++ b/addons/scratch-messaging/background.js
@@ -1,10 +1,13 @@
 import commentEmojis from "../scratch-notifier/comment-emojis.js";
+import { linkifyTextNode, pingifyTextNode } from "../../libraries/cs/fast-linkify.js";
 
 export default async function ({ addon, global, console, setTimeout, setInterval, clearTimeout, clearInterval }) {
   let lastDateTime;
   let data;
   let pendingAuthChange = false;
   let addonEnabled = true;
+  // reuse one DOMParser
+  const parser = new DOMParser();
 
   const getDefaultData = () => ({
     messages: [],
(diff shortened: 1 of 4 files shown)


WorldLanguages:
> Also some changes to prevent triple serialization

Note that it might not have been for sanitization. \&quot;

addons/scratch-messaging/background.js
```diff
@@ -1,10 +1,13 @@
 import commentEmojis from "../scratch-notifier/comment-emojis.js";
+import { linkifyTextNode, pingifyTextNode } from "../../libraries/cs/fast-linkify.js";
```
WorldLanguages, line 2: You forgot `/common`

WorldLanguages:
The popup seems to be crashing for me, recursion probably gone bad :P

Also, instead of making linkifying any URL an addon setting, we should check `scratchAddons.localState.addonsEnabled["more-links"] === true` in the background page.

WorldLanguages:
> What's preventing us from moving all these logics to popup?

In the near future, only logic that should stay in the background is updating the messages array. Loading comments, posting comments, marking messages as read, loading ST alerts, and others should be moved to the popup.

WorldLanguages:
@apple502j Wanna take a look when you can? I fixed the import and it seems to be crashing the popup

apple502j:
Not a recursion but something similar: iterating over `childNodes` while adding a new node. Oops.

apple502j:
@WorldLanguages There's probably some people who only want to enable More Links on popups. We can probably change defaults for those, tho.

WorldLanguages:
I'll make some changes and merge later today

PR #2702, opened 2021-06-07
Update 60FPS

Time-dragon:
Resolves nothing

### Changes

Edited for a little clarity.

### Reason for changes

So that it looks better


GrahamSH-LLK:
This doesn't make it that much clearer.

PR #2719, opened 2021-06-09
Lint 20210609

apple502j:


addons/forum-search/userscript.js
```diff
@@ -234,7 +234,7 @@ function appendSearch(box, query, page, term, msg) {
 }
 
 export default async function ({ addon, global, console, msg }) {
-  await addon.tab.loadScript(addon.self.lib + "/thirdparty/cs/scratchblocks-v3.5.1-min.js");
```
GrahamSH-LLK, line 237: Just remove the version number, then you don't keep having to change files.

GrahamSH-LLK (changes requested):

addons/forum-search/userscript.js
apple502j: That's fine

WorldLanguages (approved):

PR #2722, opened 2021-06-09
Remove useless setting type:select from dark-www

WorldLanguages:
This got outdated, we won't be supporting more "versions"

@mxmou Please approve

Before:
image

After:
image


apple502j (approved):
LGTM, waiting for @mxmou

mxmou (commented):
I'm not sure about this. If it was done about half a year ago (in #876) I would agree but now it seems like an unnecessary change. The setting will be removed in the near(?) future anyway.

apple502j:
Okay. People refactor code for unnecessary reasons, so I'll merge.

PR #2745, opened 2021-06-11
Changes to debugger addon

apple502j:
Read commit messages for details.

image
- [ ] after merging make sure to copy-paste the userscript


addons/debugger/userscript.js
```diff
@@ -97,6 +108,11 @@ export default async function ({ addon, global, console, msg }) {
     myFlash.block = block;
 
     function _flash() {
+      if (!myFlash.block.svgPath_) {
+        myFlash.timerID = count = 0;
+        flashOn = true;
+        return;
+      }
```
TheColaber, line 115: This code is taken from devtools. is it worth bringing this change there?

apple502j: Gonna check that, bc that means devtools can crash...

apple502j: Cannot repro crash, so not gonna fix in this PR.

addons-l10n/en/debugger.json
```diff
@@ -5,5 +5,12 @@
   "debugger/clear": "Clear",
   "debugger/close": "Close",
   "debugger/enter-format": "Enter export format:",
-  "debugger/unpause": "Unpause"
+  "debugger/unpause": "Unpause",
+  "debugger/deleted-sprite": "(deleted sprite)",
+  "debugger/deleted-sprite-desc": "The sprite is deleted, so you cannot find the source.",
```
WorldLanguages, line 10: Unnecessary

addons-l10n/en/debugger.json
```diff
@@ -5,5 +5,12 @@
...
+  "debugger/clone-of": "Clone of {spriteName}",
+  "debugger/clone-desc": "This message is logged by a clone.",
```
WorldLanguages, line 12: Unnecessary

WorldLanguages:
@mxmou Commit above ^ (CSS `!important`) because dark-www made the card title color gray.

mxmou:
> @mxmou Commit above ^ (CSS `!important`) because dark-www made the card title color gray.

`!important` works, but wouldn't setting the color on the card itself instead of the heading work as well?

WorldLanguages:
Yes. Apple decided to go this way :P But sure that's also valid

PR #2759, opened 2021-06-12
Handle various errors from message fetching

### LINKED ISSUE #2727: Investigate stuck "loading messages..." in scratch-messaging
Not sure what triggers it, but I think it once happened to me and I've seen feedback about it
I don't think we made any recent code changes that should trigger something like this either, maybe a browser update changed behavior

apple502j:
Resolves #2727


WorldLanguages:
This is good, but I'd like to enhance the error checking logic in the future. Currently it's a little weird. Moving some logic to the popup will solve some of this

PR #2767, opened 2021-06-13
Add YouTube changelog video link to update notification

WorldLanguages:
Also, always shows banner if the URL hash is `#sa-update-notif`, so that it's easier to do future PRs like this

image


Shluffy:
how does forcing it to display work?

apple502j:
can you tinypng it?

WorldLanguages:
> how does forcing it to display work?

Just go to https://scratch.mit.edu/#sa-update-notif

> can you tinypng it?

I already compressed it by around 50% with https://squoosh.app

Shluffy:
> > how does forcing it to display work?
> 
> Just go to https://scratch.mit.edu/#sa-update-notif

doesn't work...

PR #2775, opened 2021-06-14
debugger: fix #2771 and some other bugs

### LINKED ISSUE #2771: [v1.16.1] debugger log/warn/error blocks logging empty string if sprite that logged is not focused
Demo: https://scratch.mit.edu/projects/544252633/

Error: block is not defined in userscript.js:318, because block isn't in the workspace

GarboMuffin:
- Changes how the block "previews" are generated to not rely on an actual Blockly block existing.
 - don't try to scroll a block in a flyout into view.
 - make the button not disappear when entering and leaving fullscreen.

### DIFF
--- a/addons/editor-theme3/theme3.js
+++ b/addons/editor-theme3/theme3.js
@@ -88,7 +88,11 @@ function updateSettings(addon, newStyle) {
       }
       #s3devIDD > li.${prop} {
         background-color: var(--editorTheme3-${settingName}Color);
-      }`;
+      }
+      .console-variable[data-category="${prop}"] {
+        background-color: var(--editorTheme3-${settingName}Color) !important;
+      }
+      `;
       if (prop === "custom") {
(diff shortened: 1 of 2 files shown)


addons/editor-theme3/theme3.js
```diff
@@ -88,7 +88,11 @@ function updateSettings(addon, newStyle) {
       }
       #s3devIDD > li.${prop} {
         background-color: var(--editorTheme3-${settingName}Color);
-      }`;
+      }
+      .console-variable[data-category="${prop}"] {
+        background-color: var(--editorTheme3-${settingName}Color) !important;
+      }
+      `;
```
WorldLanguages, line 95: What does this do? I'm not sure how theme3 behaves with this new category

GarboMuffin: It makes custom block colors work in here:

image

Issue #2760, opened 2021-06-12
Slow-Mo, Fast forward and super fast forward


Genius398:
*Idea*

So we can have a Slow-mo, Fast Forward and Super Fast Forward button in the same area where the pause button is. We can group them up in a "More" section. I mean that when 1 button is clicked, it shows or hides the other buttons. The button can be useful in skipping a part of a project.

apple502j:
Hmm, is it like FPS presets?

WorldLanguages:
Kinda, we'd need to make sound go faster too, good luck with that lol

apple502j:
I mean we can just run pitch sound effect

fierin:
Someone change the title to include **Rewind** and ### _**close**_ #3120.

PR #2784, opened 2021-06-15
allow fps lower then 31 to 60fps

### LINKED ISSUE #1963: Rename "60FPS player mode"

The "Alt+GreenFlag" looks strange, and the FPS can be changed (it's not necessarily 60)

Change to "Alternate FPS player mode" or something similar

Remove "Alt+GreenFlag" from name, it's a little awkward to put how to use the addon in the name of the addon
For example the mute addon is not named "Ctrl+GreenFlag mute project"

### LINKED ISSUE #2778: Have projects move slower addon

projects can only run at 30-240 fps using the 60fps addon. You can't make it run slower

A new addon like this

remove the minimum for the 60fps addon

### Additional context

can this be done by just changing the min to 1, max to 29, and change the default?

Shluffy:
Resolves #2778 Fixes Slow-mo part of #2760 resolves #1963

### Changes
change the `addon.json` file to have the minimum be 1, changes description and changes name to be "custom fps player mode"
 
### Reason for changes
because this is just the opposite that makes sense to have
 
### Tests
I have not tested this.


RedGuy12:
Changing the name and desc is fine IMO, just not the id.

Shluffy:
All right, ty. I will change it

RedGuy12:
Resolves #1963

Shluffy:
@RedGuy12 is the name fine to change, just not the id?

Shluffy:
is this going to be like this 4ever?

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "Alt+GreenFlag 60FPS player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Custom FPS player mode",
+  "description": "Alt+Click the green flag to toggle Custom FPS.",
```
mxmou, line 3: I don't think "Custom FPS" should be capitalized.

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "Alt+GreenFlag 60FPS player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Custom FPS player mode",
```
mxmou, line 2: "Player mode" should be changed to "project player mode" for consistency with the mute addon.

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
...
+  "description": "Alt+Click the green flag to toggle Custom FPS.",
   "info": [
     {
       "type": "notice",
-      "text": "Projects will run twice as fast at 60FPS, because they typically run at 30FPS.",
+      "text": "Projects will run at a different speed when in use. Keep in mind that projects typically run at 30 fps",
```
mxmou, line 7: Notices should end with a dot, like descriptions. I would change it to "Enabling the custom FPS mode changes the project speed. Projects typically run at 30 FPS."

mxmou (changes requested):

addons/60fps/addon.json
Shluffy: fps was caps before, and custom is first word of name. Player and Mode should maybe be caps

WorldLanguages:
> is this going to be like this 4ever?

I'll review this in a few days, I'm taking a little break

addons/60fps/addon.json
CST1229: I think he was referring to the description.

Shluffy: @mxmou were you? i can change that

mxmou: > I think he was referring to the description.

I was.

Shluffy: i made the `C` in custom lowercase, but the fps was caps in the notice as well

WorldLanguages:
IMO slow down / go faster buttons would be very useful, but they should be part of another addon. I don't think letting users enter less than 30 FPS on this addon makes sense. It would fix the problem temporarily, but don't forget this addon is enabled by default and 99% of its users will keep it at fixed 60

WorldLanguages:
If you want to achieve this currently, you can run this in the settings page:

```js
vue.addonSettings["60fps"].framerate = 19
vue.updateSettings(vue.manifestsById["60fps"])
```

return-misaPuding:
This is a useful feature that some people experimenting with and debugging some enhanced scratch projects could use and won´t harm other users (because as you already said, almost all of them (even me) don't even change the FPS and the ones which do so just won't set it to FPS lower than 30 unless they want it for some reason)

PR #2804, opened 2021-06-16
debugger: fix more bugs

### LINKED ISSUE #2749: debugger: window can't be dragged when not fully visible
1. Open log window
2. Move it to an edge
3. Shrink the window in that edge

Window can only be dragged one way and cannot be closed without resizing the window back

GarboMuffin:
- Fixes #2749
 - Fixes https://discord.com/channels/751206349614088204/755131743136383107/854819125770584065
 - Fix all other cases of console blocks not being called
 - Fixes addon sometimes not working when going from project page to editor
 - Fixes going to block not working immediately after leaving editor and returning to editor
 - Improve performance of logging items by probably close to 100x (the browser painting those items in the list is still painfully slow and always will be until it paginates or uses virtual lists, not something I want to do right now but might later if you want)
 - Fix unread logs icon not appearing reliably when a bunch of items are being logged (and also fix performance issues caused by that)
 - Activate the code tab when going to a block


addons/debugger/userscript.js
```diff
@@ -198,14 +205,19 @@ export default async function ({ addon, global, console, msg }) {
...
-  let pos1 = 0,
-    pos2 = 0,
-    pos3 = 0,
-    pos4 = 0,
-    maxX,
-    maxY;
   consoleTitle.addEventListener("mousedown", dragMouseDown);
 
+  let isScrolledToEnd = true;
+  extraContainer.addEventListener("wheel", (e) => {
```
apple502j, line 211: needs to be passive

addons/debugger/userscript.js
```diff
@@ -198,14 +205,19 @@ export default async function ({ addon, global, console, msg }) {
...
+    // When user scrolls up, stop automatically scrolling down
+    if (e.deltaY < 0) {
+      isScrolledToEnd = false;
+    }
+  });
+  extraContainer.addEventListener("scroll", () => {
```
apple502j, line 217: here too

apple502j (changes requested):

addons/debugger/userscript.js
```diff
@@ -71,6 +70,14 @@ export default async function ({ addon, global, console, msg }) {
     // Don't scroll to blocks in the flyout
     if (block.workspace.isFlyout) return;
 
+    // Make sure the code tab is active
+    if (addon.tab.redux.state.scratchGui.editorTab.activeTabIndex !== 0) {
+      addon.tab.redux.dispatch({
+        type: "scratch-gui/navigation/ACTIVATE_TAB",
+        activeTabIndex: 0,
+      });
+    }
```
WorldLanguages, line 79: Block won't flash here, console will error `Cannot read property 'blockDragSurface_' of null` when calling `getRelativeToSurfaceXY()`. Block should either flash properly (setTimeout of 0?) or we should not try to make it flash.

addons/debugger/userscript.js
```diff
@@ -219,32 +239,41 @@ export default async function ({ addon, global, console, msg }) {
...
+    lastX = x;
+    lastY = y;
+    const width = document.documentElement.clientWidth || document.body.clientWidth;
+    const height = document.documentElement.clientHeight || document.body.clientHeight;
+    const clampedX = Math.max(0, Math.min(x - mouseOffsetX, width - consoleWrapper.offsetWidth));
+    const clampedY = Math.max(0, Math.min(y - mouseOffsetY, height - consoleWrapper.offsetHeight));
+    consoleWrapper.style.left = clampedX + "px";
+    consoleWrapper.style.top = clampedY + "px";
+  }
+
```
WorldLanguages, line 267: This is letting me drag the logs window exactly 1px more to the bottom that I should, resulting in a vertical scrollbar in the editor

PR #2817, opened 2021-06-17
studio-tools: add errors related to manager limit

apple502j:
Please make sure to run the userscript after merging.

Untested because the API change that backs the manager limit isn't available yet.


apple502j:
self merging

Issue #2380, opened 2021-05-05
Addon/option to disable selecting a sprite when double clicking inside editor?


towerofnix:
**Describe the new addon**

A lot of games involve repeatedly clicking! In Scratch, double-clicking on a sprite in the editor makes it become the focused sprite, which causes a bit of lag when loading in the new blocks, and makes you have to return to the sprite you're working in to go back to coding. (Plus, when using the sprite folders addon, if the sprite you're double-clicking is in a collapsed folder, it'll un-collapse the folder. This is annoying and basically makes it so I never collapse the folder a sprite I regularly click is in, since I just keep having to re-collapse it. :P)

This is a pretty small feature so maybe it fits better as an addition to an existing addon—but it's also probably a regularly sought tweak, so unless it already exists and I'm missing it somewhere, it may actually be worth having it as its own thing. Up for discussion!

(It could be added as part of a more substantial "treat stage like outside editor" addon, alongside other options, like "disable dragging sprites in editor". [I implemented a basic version of this in Scrap a few years ago!](https://github.com/towerofnix/scrap-mod/pull/49))

ErrorGamer2000:
I just put it in fullscreen mode, that fixes it.

ErrorGamer2000:
I have a suggestion. What if we put an invisible `div` over the project area that can be toggled on and off?

ErrorGamer2000:
Aaannnd.... It Works! just paset this code into the console:

```js
((doc) => {
  let c = document.querySelector("canvas");
  let rect = c.getBoundingClientRect();
  let div = document.createElement("div");
  div.style.position = "fixed";
  div.style.top = rect.top + "px";
  div.style.left = rect.left + "px"
  div.style.height = rect.height + "px"
  div.style.width = rect.width + "px"
  doc.appendChild(div);
})(document.body || document.documentElement)
```

retronbv:
> paset

paste*

ErrorGamer2000:
oof

GrahamSH-LLK:
> Aaannnd.... It Works! just paset this code into the console:
> 
> ```js
> ((doc) => {
>   let c = document.querySelector("canvas");
>   let rect = c.getBoundingClientRect();
>   let div = document.createElement("div");
>   div.style.position = "fixed";
>   div.style.top = rect.top + "px";
>   div.style.left = rect.left + "px"
>   div.style.height = rect.height + "px"
>   div.style.width = rect.width + "px"
[...]

PR #2818, opened 2021-06-17
Added disable-stage-drag-select addon

### LINKED ISSUE #2709: Enable/disable clicking on sprites on stage from selecting them.

Alot of my projects involve clicking and dragging sprites to move them around. However if I do that not in full screen it changes the focus to that sprite which is frustrating when I need to test it by moving the sprite but always have to return the focus to the original sprite afterwards. It also drags it in the editor, which i'd like to override with my own dragging.

I'd like for a button in the addon which you could enable/disable "draggable in editor", which would be in the addons menu.

Going full screen is one alternative but not really a solution, as it still wastes time, when i'm making lots of iterative changes.
Currently I make it so that you can click near the sprite and move it rather than actually clicking

### Additional context

I was considering how to actually implement this, and it's not trivial. You want to disable the dragging/change focus event but you don't want to disable the onclick event entirely or mouse sensing wouldn't work.
Additionally you'd need to get access to the event listeners of a canvas object and i'm not sure how possible that is.

Alteratively we could modify the scratch code itself. The code we need to change is in either of these I think:

- Rendered Sprite code: https://github.com/LLK/scratch-vm/blob/f405e59d01a8f9c0e3e986fb5276667a8a3c7d40/src/sprites/rendered-target.js

- Mouse code: https://github.com/LLK/scratch-vm/blob/f405e59d01a8f9c0e3e986fb5276667a8a3c7d40/src/io/mouse.js

- Stage GUI code: https://github.com/LLK/scratch-gui/blob/6300e4b819c3bf1a77b613517469d6e631604ab1/src/containers/stage.jsx

Thank you to anyone willing to help. If you don't have time but you know how to solve it tell me and I'll try implementing it.

Joeclinton1:
Resolves #2709 

### Changes

I've added a new addon called disable-stage-drag-select. All it does is redefine the dragStart and dragStop functions to return instantly, essentially deleting them.

### Reason for changes

This adds the functionality to drag sprites in the editor and not select that sprite or call the dragStart or dragStop functions in the virtual machine.

### Tests

I tested it on one of my projects which I need it for: https://scratch.mit.edu/projects/524385137/. The result was that it did exactly what I needed it to do.

However, since I don't have access to the stage container and the methods contained within it, I couldn't remove the onDragStart or onDragStop methods, so the sprite will still go the top and be slightly offsetted when dragged. It will also still be selected on double click. This is only a minor inconvenience and most people wouldn't notice this is happening unless they know it's there.

### DIFF
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Disable sprite dragging in editor",
-  "description": "Makes dragging sprites in the project player behave exactly as in fullscreen while in the editor.",
+  "description": "Removes the ability to drag sprites around in the stage, except those explicitly set as draggable.",
   "credits": [
     {
       "name": "Chrome_Cat",
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -5,6 +5,9 @@
     {
       "name": "Chrome_Cat",
       "link": "https://scratch.mit.edu/users/Chrome_Cat"
+    },
+    {
+      "name": "GarboMuffin"
     }
   ],
   "userscripts": [
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Disable sprite dragging in editor",
+  "name": "Non-draggable sprites in editor",
   "description": "Removes the ability to drag sprites around in the stage, except those explicitly set as draggable.",
   "info": [
     {
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -1,18 +1,18 @@
 {
-    "name": "Disable sprite dragging in editor",
-    "description": "Makes dragging sprites in the project player behave exactly as in fullscreen while in the editor.",
-    "credits": [
-        {
-          "name": "Chrome_Cat",
-          "link": "https://scratch.mit.edu/users/Chrome_Cat"
-        }
-    ],
-    "userscripts": [
-        {
-          "url": "userscript.js",
-          "matches": ["https://scratch.mit.edu/projects/*"]
-        }
-    ],
-    "tags": ["editor"],
-    "enabledByDefault": false
+  "name": "Disable sprite dragging in editor",
+  "description": "Makes dragging sprites in the project player behave exactly as in fullscreen while in the editor.",
+  "credits": [
+    {
+      "name": "Chrome_Cat",
+      "link": "https://scratch.mit.edu/users/Chrome_Cat"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+  "tags": ["editor"],
+  "enabledByDefault": false
 }
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -14,5 +14,8 @@
     }
   ],
   "tags": ["editor"],
-  "enabledByDefault": false
+  "enabledByDefault": false,
+  "dynamicDisable": true,
+  "dynamicEnable": true,
+  "versionAdded": "1.17.0"
 }
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -20,7 +20,7 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["https://scratch.mit.edu/projects/*"]
+      "matches": ["projects"]
     }
   ],
   "tags": ["editor"],
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -1,6 +1,13 @@
 {
   "name": "Disable sprite dragging in editor",
   "description": "Removes the ability to drag sprites around in the stage, except those explicitly set as draggable.",
+  "info": [
+    {
+      "type": "notice",
+      "text": "Hold Shift while dragging a sprite to move it normally, even if it's not set as draggable.",
+      "id": "shift"
+    }
+  ],
   "credits": [
     {
       "name": "Chrome_Cat",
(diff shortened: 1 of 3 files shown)


RedGuy12:
> However, since I don't have access to the stage container and the methods contained within it, I couldn't remove the onDragStart or onDragStop methods, so the sprite will still go the top and be slightly offsetted when dragged. It will also still be selected on double click. This is only a minor inconvenience and most people wouldn't notice this is happening unless they know it's there.

Add a notice about this in the addon manifest.


Also, please remove all the comments in the userscript file.

RedGuy12:
Maybe you could make it only drag if `alt` is selected?

Joeclinton1:
> Add a notice about this in the addon manifest.

I can do that.

> Also, please remove all the comments in the userscript file.

I left the comment because I believe it's helpful. It provides a second way of achieving the same functionality.

Joeclinton1:
> Maybe you could make it only drag if `alt` is selected?

I can do that too.

WorldLanguages:
This is nice, but we need to name this addon properly, otherwise users might expect something else 🤔

GarboMuffin:
I'm a bit confused what you're trying to accomplish, but it sounds like your goals are:

1. dragging a sprite should not select it in the editor
2. dragging a sprite in the editor should work like it does in fullscreen or in player only mode -- can only drag sprites marked as draggable

The approach this PR currently uses completely breaks dragging sprites that are actually marked as draggable, so here's my proposal:

```js
export default async ({addon, console}) => {

    // Do not focus sprite after dragging it
    const oldStopDrag = vm.stopDrag;
    vm.stopDrag = function (...args) {
        const setEditingTarget = this.setEditingTarget;
        this.setEditingTarget = () => {};
        const r = oldStopDrag.call(this, ...args);
        this.setEditingTarget = setEditingTarget;
        return r;
    };

    // Don't let the editor drag sprites that aren't marked as draggable
    const oldGetTargetIdForDrawableId = vm.getTargetIdForDrawableId;
    vm.getTargetIdForDrawableId = function (...args) {
        const targetId = oldGetTargetIdForDrawableId.call(this, ...args);
        if (targetId !== null) {
            const target = this.runtime.getTargetById(targetId);
            if (target && !target.draggable) {
                return null;
            }
        }
        return targetId;
    };
};
```

Has a side effect of resolving #2380 for sprites that aren't draggable

Joeclinton1:
@WorldLanguages Sure, what do you have in mind?

Joeclinton1:
@GarboMuffin That is true, I neglected draggable sprites because I thought anyone who has this addon turned on clearly doesn't care about them. With your solution you went with my proposed solution of editing the drag function, which is fine, I just am worried it could break in the long term if it became incompatible, with the code. (though that is pretty unlikely and it would be an easy fix)

The second part of modifying the get editing target is a good idea. Just trick it into thinking it's not hovering over anything if it's not draggable. Are you sure that has no negative side effects? `.getTargetIdForDrawableId` is used for alot more than just this.

Apart from those concerns I think it is better, I'll change it to that if you can confirm the second part affects nothing else negatively.

GarboMuffin:
getTargetIdForDrawableId is used in exactly two spots:

https://github.com/LLK/scratch-gui/blob/336c60b2f7f1f5578095650cce1ff0323c9b5033/src/containers/stage.jsx#L170
https://github.com/LLK/scratch-gui/blob/336c60b2f7f1f5578095650cce1ff0323c9b5033/src/containers/stage.jsx#L345

It is safe to override

Joeclinton1:
> getTargetIdForDrawableId is used in exactly two spots:
> 
> https://github.com/LLK/scratch-gui/blob/336c60b2f7f1f5578095650cce1ff0323c9b5033/src/containers/stage.jsx#L170
> https://github.com/LLK/scratch-gui/blob/336c60b2f7f1f5578095650cce1ff0323c9b5033/src/containers/stage.jsx#L345
> 
> It is safe to override

Ah, I guess it's not used alot then, I must've confused it with another function. Ok, I'll change it to that. Then just need to make the changes to do the other things.

Shluffy:
Shouldn't prs not be made off the master branch?

WorldLanguages:
@mxmou Tell me what you think about the addon name and description I chose, so that it fits with the PR to be merged soon

WorldLanguages:
TODO: add shift+drag to addon description/info

WorldLanguages:
TODO: credit garbo

addons/disable-stage-drag-select/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Disable sprite dragging in editor",
+  "description": "Makes dragging sprites in the project player behave exactly as in fullscreen while in the editor.",
```
mxmou, line 3: This description is a bit weird:
- it sounds like you can still drag sprites in full screen, but differently;
- "stage" would be a more appropriate word than "project player" here because it's talking about sprites.

Maybe "Removes the ability to drag sprites around in the editor"?

WorldLanguages: What about the name? I thought we would stop using verbs?

WorldLanguages:
@mxmou Tell me if the name of the addon is OK: `Disable sprite dragging in editor`
If it's OK we can merge

WorldLanguages:
This addon could have `projectPlayer` subcategory, but I don't think it fits. Technically it only affects the project player, but it doesn't add any buttons, unlike all others in that subcategory, so I prefer to leave it inside "others"

RedGuy12:
> This addon could have `projectPlayer` subcategory, but I don't think it fits. Technically it only affects the project player, but it doesn't add any buttons, unlike all others in that subcategory, so I prefer to leave it inside "others"

i don't agree. it's changes the project player



> @mxmou Tell me if the name of the addon is OK: `Disable sprite dragging in editor`
> If it's OK we can merge

`Non-draggable sprites`? Not sure what would be good. but pls not a verb

WorldLanguages:
@RedGuy12 We already have "disable auto-save", so it depends on what mxmou decided to do with that name.

mxmou:
> @RedGuy12 We already have "disable auto-save", so it depends on what mxmou decided to do with that name.

I didn't change "Disable auto-save" - I don't think there is a better name for it. In this case @RedGuy12's suggestion ("Non-draggable sprites") is a good way to avoid using a verb and it's also shorter, so I think it should be used.

WorldLanguages:
I'll go with `Non-draggable sprites in editor`, so that it's clear by the name it only affects the stage while editing.

Joeclinton1:
Hi guys, awesome work!I started a new job (software engineer) recently and haven't had much time to do anything, but I'm so glad to see you guys took it upon yourselves to complete the feature. Very cool!

PR #2822, opened 2021-06-18
Add new option to waitForElement

apple502j:
This fixes an issue where copy-message-link does not work when logged out.

Other solutions:
- using waitForElement on div.comments, markAsSeen: does not work when new comments are loaded
- using waitForElement on report button: does not work when logged out
- rewriting Backbone template: does not work when logged out, because Scratch has authentication check in Scratch.Comments.prototype.showCommentOptions
- polluting showCommentOptions: while this is the best solution, it's very racy.
- MutationObserver: kinda works, so I've made it into the API


WorldLanguages:
Closing in favor of #2831 at least for now

PR #2827, opened 2021-06-18
Mark studio-tools as beta when new studios are alive

apple502j:


WorldLanguages:
I'd prefer marking it as beta normally

PR #2830, opened 2021-06-18
Settings popup: fix disabled searched addons staying after clearing search

WorldLanguages:
Easy fix: no addons group the `_iframeSearch` pseudogroup should match search if the search input is empty. Addons not from that pseudogroup continue to match search if the search input is empty for obvious reasons.

To reproduce:
1. Open "addons" tab in popup
2. Search for a disabled addon with the search bar
3. Clear the search term - all search results stay

image


RedGuy12:
Thanks for reporting @thecrewmate876!

PR #2832, opened 2021-06-18
Merge redirects into one addon

### LINKED ISSUE #2718: New addon type: redirection
Redirection addons handle redirects, that's it. This is different from using userscripts, because redirections will happen either at background (tabs.update) or content script, which makes things very fast.

FunctionalMetatable:
Closes #2718

### Changes

Adds one addon for redirecting

### Reason for changes

So the easterEggs won't get cluttered up with redirections. 

### Tests

Tested locally.


apple502j:
Please do not change addon id.

Issue #2407, opened 2021-05-06
Don't show dangerous addons at the top or allow disabling this behavior


mxmou:
**Is your feature request related to a problem? Please describe.**

I don't want to be reminded that I'm using a "dangerous" addon (disable auto-save) every time I open the settings page.

**Describe the solution you'd like**

Just don't show them at the top. I think the warning that appears when enabling them is enough.

**Describe alternatives you've considered**

Make it an option in More settings that is enabled by default.

WorldLanguages:
It shows at the top... of the list of your enabled addons. The sorting algorithm simply prefers dangerous/beta addons at the top. @mxmou Why do you think this is a big deal?

lisa-wolfgang:
It's probably a good idea to continuously keep those addons on users' minds. There's a good reason that they're labeled "Dangerous".
Users who are troubled by it probably shouldn't be using those addons in the first place.

WorldLanguages:
IMO, "reminded" is an overstatement, it's simply preferred by the sorting algorithm. Would you prefer being shown another random enabled addon there instead? (possibly an editor addon whose name starts with letter "a")

PR #2835, opened 2021-06-19
Mark studio-tools as beta addon

apple502j:
Still enabled by default, but can change.


mxmou:
This will be annoying because of #2407.

PR #2844, opened 2021-06-19
JSdoc intellisense test

WorldLanguages:
I admit I'm doing this for fun


addons/2d-color-picker/paint-editor.js
```diff
@@ -4,7 +4,7 @@
 import { normalizeHex } from "../../libraries/common/cs/normalize-color.js";
 import RateLimiter from "../../libraries/common/cs/rate-limiter.js";
 
-export default async ({ addon, console, msg }) => {
+export default /** @param {Addon.Userscript} */ async ({ addon, console, msg }) => {
```
RedGuy12, line 7: put it on the line above because

- it looks better
- prettier might get confused if it's there
- it still works above

addon-api/types.d.ts
```diff
@@ -0,0 +1,25 @@
+import UserscriptAddon from "./content-script/Addon.js";
+import PersistentScriptAddon from "./background/Addon.js";
+
+interface Script {
+  global: Object;
+  console: Console;
+  msg: () => String | null;
```
RedGuy12, line 7: string lowercase

addons/2d-color-picker/paint-editor.js
WorldLanguages: It's ugly to start the file with a comment IMO :P 
Not the case with the file you commented this on, but will happen in most other files

addon-api/types.d.ts
```diff
@@ -0,0 +1,25 @@
+import UserscriptAddon from "./content-script/Addon.js";
+import PersistentScriptAddon from "./background/Addon.js";
+
+interface Script {
+  global: Object;
```
RedGuy12, line 5: `Object` means `any`
do `{ [key: string]: string }` to specify objects

addon-api/types.d.ts
WorldLanguages: Ah, good catch. I got confused because typescript didn't like lowercase `object`

addon-api/types.d.ts
WorldLanguages: heh, typescript is weird

addons/2d-color-picker/userscript.js
```diff
@@ -1,7 +1,6 @@
 import paintEditorHandler from "./paint-editor.js";
 
-export default async (api) => {
-  const { addon } = api;
+export default /** @param {Addon.Userscript} */ async ({ addon, console }) => {
```
TheColaber, line 3: This breaks the addon

WorldLanguages: Right

content-scripts/cs.js
```diff
@@ -130,10 +165,22 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
+ *   injectAsStyleElt: boolean;
+ *   index: number;
+ * }} addon
+ */
 function addStyle(addon) {
   const allStyles = [...document.querySelectorAll(".scratch-addons-style")];
   const addonStyles = allStyles.filter((el) => el.getAttribute("data-addon-id") === addon.addonId);
 
+  /**
+   * @param {HTMLStyleElement} el
```
TheColaber, line 181: Could be link element?

TheColaber:
I hate this tbh

.github/workflows/format-code.yml
```diff
@@ -18,21 +18,25 @@ jobs:
...
+
+      - run: npm install prettier-plugin-jsdoc prettier
+
       - name: Prettify code (ScratchAddons)
         if: steps.generate-token.outcome == 'success'
         uses: creyD/prettier_action@v3.1
         with:
           prettier_options: --write .
           commit_message: Format code
+          prettier_plugins: "prettier-plugin-jsdoc"
```
Hans5958, line 30: This won't work since prettier_plugins only work with libs that starts with `@prettier/`. I need to do some changes with this. Stand by.

Hans5958 (changes requested):

.github/workflows/format-code.yml
WorldLanguages: I'm not sure if we want to prettify JSDoc in the first place tbh

WorldLanguages:
This PR went too far, I'll work on this again after v1.17.0 release.

RedGuy12:
noo i spent forever on this i dont want it to go to waste

WorldLanguages:
Then I need to revert all the added `?.`s lol
I know many changes are good, like the new types JSdoc

PR #2848, opened 2021-06-20
Add some JSDoc

apple502j:


WorldLanguages:
I think we should reuse the first sentence from the description of every property/method from the API docs in the JSDoc for consistency (except for cases where they aren't documented).

RedGuy12:
doesn't matter imo

RedGuy12:
Merged into #2844

PR #2852, opened 2021-06-20
New API: addon.tab.appendToSharedSpace

### LINKED ISSUE #2846: addon.tab.appendToSharedSpace API
Related to #2110

Lower order means element goes before others with higher order number.

Example:

```js
// addons/gamepad/userscript.js
addon.tab.appendToSharedSpace({space: "stageHeaderSizeControls", element: gamepadButton, order: 0});

// addons/debugger/userscript.js
addon.tab.appendToSharedSpace({space: "stageHeaderSizeControls", element: debuggerButtonContainer, order: 1});
```

Would guarantee this, no matter what addon calls the function first:
image

Currently, both addons race condition on the order and it's up to luck. (and dynamicEnable makes it even less predictable)

WorldLanguages:
Resolves #2846

- Add API (see docs below)
- Hide debugger in small stage mode
- Support RTL margin in gamepad & debugger buttons
- mouse-pos static width to avoid clone count moving around costantly when moving cursor

Docs: https://scratchaddons.com/docs/reference/addon-api/addon.tab/addon.tab.appendtosharedspace/

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -306,8 +306,8 @@ export default class Tab extends Listenable {
       beforeRemixButton: {
         element: () => q(".project-buttons"),
         from: () => [],
-        until: () => [q(".project-buttons > .remix-button")]
-      }
+        until: () => [q(".project-buttons > .remix-button")],
+      },
     };
 
     const spaceInfo = sharedSpaces[space];
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -249,9 +249,9 @@ export default class Tab extends Listenable {
         element: () => q("[class^='stage-header_stage-size-row']"),
         from: () => [],
         until: () => [
-          // Small/big stage buttons
+          // Small/big stage buttons (for editor mode)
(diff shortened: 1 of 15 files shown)


ErrorGamer2000:
Can you make this work with the tab area? Addons like `variable-manager` that need (or, will need since this is the only addon that currently adds a new tab) a lot of space will need to make a new tab. I say this mainly because I am planning on adding a new editor tab for my `text-editor` addon.

WorldLanguages:
Good idea!

ErrorGamer2000:
Thanks @WorldLanguages!

WorldLanguages:
@mxmou What's the purpose of `.sa-gamepad-spacer`? Why not just use `margin-left: auto`? I need to know this for a potential `fullscreenStageHeader` space.

GarboMuffin:
> What's the purpose of `.sa-gamepad-spacer`? Why not just use `margin-left: auto`? I need to know this for a potential `fullscreenStageHeader` space.

It's used to right-align a button when the fullscreen mode DOM doesn't have an easy spot to just append one. sa-gamepad-spacer was just how I decided to do that. If margin-left: auto works then you can use that instead

PR #2874, opened 2021-06-28
Change bottom border radius on banner add-on

Daniel4-Scratch:
I changed the bottom border radius for banner add-on just to look more appealing to the eye. 

The two cornered radius borders look ugly in the 3.0 add-on with the banner add-on. I think it should be changed

Add this bit of code down below to the 3.0 add-on css file for the profile page
```css
.box#profile-data .box-head {
    border-bottom-left-radius: 10px;
    border-bottom-right-radius: 10px;
}
```

Much better

### Tests
It works when a inspect the scratch website.
When i disable the banner and apply the code it looks normal. Nothing changed without the banner on.

@mxmou this is your add-on so I think i need approval from you first.

### DIFF
--- a/addons/scratchr2/profile.css
+++ b/addons/scratchr2/profile.css
@@ -154,3 +154,9 @@
 .portrait [data-control="edit"] {
   border-top-left-radius: 5px;
 }
+
+.box#profile-data .box-head {
+    border-bottom-left-radius: 10px;
+    border-bottom-right-radius: 10px;
+}
+
--- a/addons/scratchr2/profile.css
+++ b/addons/scratchr2/profile.css
@@ -159,4 +159,5 @@
   border-bottom-left-radius: 10px;
   border-bottom-right-radius: 10px;
 }
+ 
 
--- a/addons/scratchr2/profile.css
+++ b/addons/scratchr2/profile.css
@@ -155,7 +155,7 @@
   border-top-left-radius: 5px;
 }
 
-#profile-data .box-head {
+#profile-data .box-head::before {
   border-bottom-left-radius: 10px;
   border-bottom-right-radius: 10px;
 }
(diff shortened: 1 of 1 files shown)


addons/scratchr2/profile.css
```diff
@@ -154,3 +154,9 @@
 .portrait [data-control="edit"] {
   border-top-left-radius: 5px;
 }
+
+.box#profile-data .box-head {
```
mxmou, line 158: I like this, but it actually does make a change to the bottom left corner of the profile picture when the banner addon is off. Adding `::before` to the end of the selector makes it target the image itself, which isn't there when the banner is off.

`.box` at the beginning is unnecessary.

mxmou (changes requested):

addons/scratchr2/profile.css
Daniel4-Scratch: soo you want it to be `::before #profile-data .box-head`?

Daniel4-Scratch (commented):
‏‏‎ ‎

addons/scratchr2/profile.css
mxmou: The correct selector is `#profile-data .box-head::before`.

Daniel4-Scratch: okie dokie

mxmou (approved):

Daniel4-Scratch:
How to I make the format check not fail?

TheColaber:
You may go to https://github.com/Daniel4-Scratch/ScratchAddons/actions and enable actions. Then just make a dummy commit by adding a space at the end of the file or something.

Daniel4-Scratch:
> You may go to https://github.com/Daniel4-Scratch/ScratchAddons/actions and enable actions. Then just make a dummy commit by adding a space at the end of the file or something.

Thanks it worked 😊

Daniel4-Scratch:
most of my commits have just been border-radius changes lol

Issue #762, opened 2020-11-16
Replace "Tutorials" or "About" button in scratch with "Convert" Button


SparkScratch-P:
**Replace "Tutorials" or "About" button in scratch with "Convert"  Button**

**I am frustrated when I have to go through a lengthy process to convert sb3 to exe**

**I am here as a coder for 7 yrs. So just reporting a bug doesn't enjewel me always. I am basically to suggest another button replace like "Ideas" with "Discuss" button.i want either the "Tutorials" button in editor, or the "about" button on web to be replaced with convert button. It will be much easier and helpful to convert to .exe directly from the web than encountering a frustrating journey all over ur PC.I promise to provide all codes needed but anyone pls make a code to replace the buttons**

**Pls don't close it for I will give further codes for this to be carried out :)**

JeffaloBob:
so this would have two parts?
1. customize navbar (this would be cool as its own addon)
2. sb3 to exe converter

perhaps it could avoid having to make the converter, by just letting people customize the navbar, and letting them add a link to one that already exists.

pufferfish101007:
Also, thinking of more converters we could have: 
* [Leopard (scratch -> javascript)](https://leopard.now.sh) 
* [Snapinator (Scratch -> Snap!)](https://snapinator.github.io)
* Sb3 -> swf? - it is possible but I don't know if there are any converters out there
* Sb3 -> apk? - probably not possible, but you never know 
* sb3 -> python - I think I saw that someone was making one recently, but it was uncompleted/buggy

SparkScrach-P:
> I want it to happen in the following way:
1. The "Convert" Button is clicked
2. It shows a pop-up providing options that would have choices of different converters
** There would be separate options for sb3 to HTML for .exe standalone and .exe Web-view App**
3. When the user clicks his choice, it would lead him to that web page !

pufferfish101007:
I think that rather than replacing an existing tab with a convert button, it would be better to have it in the file/edit menu in the editor (probably file as that's where all the downloading stuff goes), as then we don't need to handle file uploads and reading all the data - it's already all there for us in the vm (or scratch-storage or somewhere else - but it is there). I'll go and figure out how to add an item to the menu so you can start working on the converyer code if you'd like...

SparkScrach-P:
> Don't forget to view this HTMLifier for web-view exe -
[AI-HTMLifier](https://sparkscrach-p.github.io/AI-HTMLifier/)
[...]

PR #1342, opened 2021-01-16
Custom Navbar Addon + Dynamic Length Settings

TheColaber:
**Resolves**

Resolves #1333
Resolves #951
Resolves #721
Resolves #255

**Changes**

Added addon
Remove discuss button addon

**Reason for changes**

It's epic and I like it
image

**Tests**

~~It works ;)~~ Not even close to done.

PR #2875, opened 2021-06-28
Table Setting + Custom Navbar Addon

### LINKED ISSUE #255: dynamic length options

I want to make an addon for the forums where you have predifined strings you can send. I don't want to have to limit the amount to 5 or something and make each string option manually.

a new option type that had a dynamic length of multi strings if that makes sense.

I'll send a mockup soon™️

creating a custom options menu for my addon but that's bad UX, probably not good and technically not exactly very possible

**Additional context**

i send mockup soon

### LINKED ISSUE #721: Wiki Button

### LINKED ISSUE #951: Custom Navbar

### LINKED ISSUE #1333: Removes the 'Tutorials' button
Hello,
when we activate all the functions of the editor, the bar at the top is too big compared to the rest of the editor. I therefore suggest that you make an option that removes the 'Tutorials' button.
thank you,
Miala

Text translated from French by Google Translate

TheColaber:
Resolves #951
Resolves #721
Resolves #255

Exceeds #1342
Related ScratchAddons/manifest-schema#40
Merge after this: ScratchAddons/manifest-schema#50
Merge after this: ScratchAddons/l10n-script#7

### Changes

- turn discuss-button to Custom Navbar
- background logic for when a user goes from using old setting to table settings
- remove background logic for previous versions
- Move handleKeySettings to addon-settings component
- removes forums tag... why do we need it??
- adds images/icons/drag.svg
- adds libraries/thirdparty/sortable.min.js
- simpler way to do checkValidity

### Manifest Spec
```jsonc
{
  "name": String,
  "id": String,
  "type": String,
  "default": Array of Arrays of Strings, Booleans, and Numbers,
  "row": Array of Settings // excluding type table setting without default prop
  "presets": Array of { name: String, values: Arrays of Strings, Booleans, and Numbers }
}
```

### TODO
- [x] add the library info
- [x] description and name
- [x] don't make javascript URLs work
- [x] lightmode
- [x] presets
- [x] i10n - apple said they'd work on this soon

### Tests

https://user-images.githubusercontent.com/72760579/123692567-f7a2dc80-d824-11eb-9ed8-b5a731fa83eb.mp4


mxmou:
I see this when I open the settings page after reloading the extension. What causes this?
image

Could the delete, add and drag icons have hover backgrounds, just like the reset settings one (but with light mode support)?

Possible name and description:
- Customizable navigation bar
- Choose a list of links to be displayed in the navigation bar.

TheColaber:
What were the settings set to?
```js
JSON.parse(JSON.stringify(vue.addonSettings["discuss-button"]))
```
Any errors in the console?

TheColaber:
That's weird... I didn't create something called `links`... and Items is supposed to be an array, not an object...
image
What are the value of `items`

mxmou:
> What are the value of `items`

image

TheColaber:
I may have messed up the background transition code...

TheColaber:
I was able to repro, and got my settings looking like this:
image
Turns out I messed up my transition code

TheColaber:
@mxmou fixed. you can try setting your settings to `buttonName: Discuss` and `removeIdeasBtn: true` and then reload the extension

mxmou:
> @mxmou fixed. you can try setting your settings to `buttonName: Discuss` and `removeIdeasBtn: true` and then reload the extension

It works now.

Could the background color of items be `#333333` to avoid adding another color that isn't used anywhere on the page yet?

TheColaber:
@mxmou are you fine with the current design of the table? You can commit any changes you want to improve it.

TheColaber:
Also this is weird...

https://user-images.githubusercontent.com/72760579/123882210-4f6c4100-d914-11eb-8a32-5220dcf961f4.mp4

I'm guessing this happens to new addons only? We may need an issue for this.

TheColaber:
mxmou, btw, items in the popup now look like this
image
Is there any way we can try to keep "name" on one line?

mxmou:
> mxmou, btw, items in the popup now look like this
> image
> Is there any way we can try to keep "name" on one line?

It doesn't look like that for me.

mxmou:
The transition code should put Discuss before About.

mxmou:
Reloading the extension seems to make `items` an object.

lisa-wolfgang:
A couple of things I noticed:
- The popup no longer loads... that might be a tiny problem.
- The settings on the settings page also don't load...
- Oh, that's why:
image

TheColaber:
Fixed, I believe

addons/discuss-button/addon.json
```diff
@@ -1,14 +1,6 @@
 {
-  "name": "Discuss button",
-  "description": "Adds the discuss button to the navbar.",
-  "credits": [
```
RedGuy12, line 4: was removing credits intentional?

addons/discuss-button/userscript.js
```diff
@@ -1,29 +1,50 @@
...
     }
-  } else {
-    const el = await addon.tab.waitForElement("div#topnav ul.site-nav");
-    if (addon.settings.get("removeIdeasBtn")) el.getElementsByTagName("li")[2].remove();
-    el.insertBefore(link, el.getElementsByTagName("li")[2]);
+    while (!searchItem.previousSibling.classList.contains("logo")) {
+      searchItem.previousSibling.remove();
+    }
+  }
+  async function init() {
```
RedGuy12, line 37: why async?

background/get-addon-settings.js
```diff
@@ -7,49 +7,33 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
       const settings = addonSettings[addonId] || {};
       let madeChangesToAddon = false;
       if (manifest.settings) {
-        if (addonId === "editor-dark-mode") {
```
RedGuy12, line 10: maybe we should keep old addon transition code?

webpages/settings/style.css
```diff
@@ -133,6 +133,12 @@ h1 {
   vertical-align: middle;
   filter: invert(0) brightness(10);
   margin-inline-end: 8px;
+  user-drag: none;
```
RedGuy12, line 136: image (caniuse)

so remove all instances of `user-drag` and use `draggable="false"` instead

webpages/settings/style.css
```diff
@@ -133,6 +133,12 @@ h1 {
...
+  user-select: none;
+  -moz-user-select: none;
+  -webkit-user-drag: none;
+  -webkit-user-select: none;
+  -ms-user-select: none;
```
RedGuy12, line 141: lol not ms

we don't support ie or old edge

webpages/settings/style.css
```diff
@@ -133,6 +133,12 @@ h1 {
   vertical-align: middle;
   filter: invert(0) brightness(10);
   margin-inline-end: 8px;
+  user-drag: none;
+  user-select: none;
+  -moz-user-select: none;
```
RedGuy12, line 138: remove prefixed versions, `user-select` covers all browsers we support

RedGuy12 (changes requested):

addons/discuss-button/addon.json
TheColaber: yes

background/get-addon-settings.js
TheColaber: why, its from like 5 versions

webpages/settings/style.css
RedGuy12: wait, nvm

edge needs -webkit- still
but remove -moz-

webpages/settings/style.css
TheColaber: i can remove it but adding `draggable="false"` is uneeded

RedGuy12: it's only supported in chromium with `-webkit-` and no other browsers support it

why is draggable=false unneeded? that's literally what this does

TheColaber: Because -webkit-user-drag works

RedGuy12: only in chromium

and it's not reccomended

RedGuy12 (approved):

addons/discuss-button/userscript.js
```diff
@@ -1,29 +1,50 @@
...
-        markAsSeen: true,
-        reduxCondition: (state) => {
-          if (!state.scratchGui) return true;
-          return state.scratchGui.mode.isPlayerOnly;
-        },
-      });
-      if (addon.settings.get("removeIdeasBtn")) el.getElementsByTagName("li")[3].remove();
-      el.insertBefore(link, el.getElementsByTagName("li")[3]);
+    const a = document.createElement("a");
+    a.href = url.startsWith("javascript") ? "" : url;
```
apple502j, line 17:
```suggestion
    a.href = /^https?:\/\//g.test(url) ? url : "";
```

apple502j (changes requested):

addons/discuss-button/userscript.js
TheColaber: What if I wanted to put the sa settings page url? Why not just exclude js urls?

apple502j:
```suggestion
    a.href = /^(?:https?|(?:chrome|moz)-extension):\/\//g.test(url) ? url : "";
```

WorldLanguages: > What if I wanted to put the sa settings page url? Why not just exclude js urls?

I don't think scratch.mit.edu can directly link to chrome-extension://*

FunctionalMetatable:
Pathnames in urls do not work

addons/discuss-button/userscript.js
TheColaber: @apple502j Your commit has broken pathnames.

TheColaber: Does this work ok?
```suggestion
    a.href = /\s*javascript:/g.test(url) ? "" : url;
```

RedGuy12: FYI Apple, on edge it's just `extention://` without the prefix

But why exclude js urls anyway?

TheColaber: Yeah keeping it would be cool. Tho it's a bad idea for many reasons.

apple502j, line 17:
```suggestion
    a.href = ["http:", "https:"].includes(new URL(url, location.href).protocol) ? url : "";
```

RedGuy12:
> Resolves #762

it does not?

TheColaber:
It does, read my comment in that issue.

Samq64:
custom-navbar-addon
I think there should be 2 items per row on bigger windows.

RedGuy12:
Wouldn't it just be display inline-block?


Samq64:
I noticed that resetting a value in a table causes it to go blank instead of using the default value even if there is one. Is this intentional?

TheColaber:
@WorldLanguages @mxmou How do you think the ui for presets should be? Maybe something like this?

https://user-images.githubusercontent.com/72760579/143478180-a595ec8b-a1e8-49cb-9031-68afcd3fa312.mp4

mxmou:
> @WorldLanguages @mxmou How do you think the ui for presets should be? Maybe something like this?

There is no indication that clicking one of the buttons adds a new row - I think a dropdown menu next to the add row button would be better.

> Also, what should our presets be?

Discuss and wiki are commonly requested, so those should certainly be there. You could also add the default buttons as presets to allow them to be easily re-added after removing one of them.

TheColaber:
@mxmou Is this similar to what you meant?

https://user-images.githubusercontent.com/72760579/143526341-b38717e5-7cf9-426f-a703-10ec3c2c41e2.mp4

mxmou:
> @mxmou Is this similar to what you meant?

Yes, this is exactly what I meant. I changed the design a bit so that it looks more similar to other buttons.

> Should the inputs be aligned? image

I don't think so. Other settings aren't aligned in the popup either.

> These preset dropdowns are here... They probably shouldn't since the default is "", and future addons probably won't want a preset dropdown either? image

Perhaps the dropdowns could contain the row presets? Future addons might also want to set the default values for a new row, and the ability to reset to those values would be useful.

TheColaber:
> Perhaps the dropdowns could contain the row presets? Future addons might also want to set the default values for a new row, and the ability to reset to those values would be useful.

Hmm... It would be better if it reset the entire row, and not just one input in the row? Is there a good way to do that?

CST1229:
Are tables of colors are supported? It would be useful for stuff like a color swatches addon where you can customize the colors.

TheColaber:
> Are tables of colors are supported? It would be useful for stuff like a color swatches addon where you can customize the colors.

All setting types should be supported
image

apple502j (approved):

mxmou:
> > Perhaps the dropdowns could contain the row presets? Future addons might also want to set the default values for a new row, and the ability to reset to those values would be useful.
> 
> Hmm... It would be better if it reset the entire row, and not just one input in the row? Is there a good way to do that?

If the reset button is separate for each input, it isn't very difficult to reset an entire row - just click the reset button for every input.

mxmou (approved):

TheColaber:
It's kind of hard to add the dropdowns since the way presets work for tables and all settings are completely different...

addons/discuss-button/addon.json
```diff
@@ -1,14 +1,6 @@
...
-    {
-      "name": "Maximouse"
-    },
-    {
-      "name": "World_Languages"
-    }
-  ],
+  "name": "Customizable navigation bar",
+  "description": "Choose a list of links to be displayed in the navigation bar.",
```
RedGuy12, line 3: Sounds off

libraries/license-info.json
```diff
@@ -10,5 +10,6 @@
   "color-picker-web-component": "MIT",
   "comlink": "Apache-2.0",
   "Sora": "OFL",
+  "sortable": "MIT",
```
RedGuy12, line 13: Don't forget to update the LICENSES.md

TheColaber:
@mxmou It seems that addon-buttons are causing the popup scrollbar to hide a bit...
image
It seems to be caused by https://github.com/ScratchAddons/ScratchAddons/pull/2875/files#diff-55c7d01e0e4e387212119cc7d7e2a0e8137c5a6f0ea7ce8b4ac9a87db76dea77R181 but I'm not sure the proper fix.

mxmou:
> @mxmou It seems that addon-buttons are causing the popup scrollbar to hide a bit...

Fixed. It wasn't actually caused by `addon-buttons`.

Samq64:
Will this make it for 1.23?

lisa-wolfgang:
I saw the commit and my first reaction was "he didn't..." and then I realized that you actually didn't lol

addons/discuss-button/addon.json
```diff
@@ -18,21 +10,58 @@
...
+          "name": "Wiki",
+          "values": { "name": "Wiki", "url": "https://scratch-wiki.info/" }
+        }
+      ]
     }
   ],
-  "versionAdded": "1.0.0",
-  "tags": ["community", "forums"],
+  "versionAdded": "1.23.0",
+  "tags": ["community", "recommended"],
```
WorldLanguages, line 65: Featured, not recommended

versionAdded should stays as 1.0.0 or maybe we can discuss whether 1.24.0 fits

Mbrick2:
IT HAS BEEN MERGED

Secret-chest:
WIKI LINK SOON!

Secret-chest:
NO MORE FORCED ABOUT PAGE LINK!

PR #2877, opened 2021-06-29
Make Settings Page, Popup, and Addon Popups Keyboard Friendly

### LINKED ISSUE #2532: Addon settings page can not be used properly with a keyboard or other accessibility tools
### Description

The addon settings page can not be properly navigated with a keyboard by pressing tab. It is also probably unusable for people who rely on other accessibility tools.

To be fair, though, these users will also have a hard time using the Scratch editor or interacting with Scratch projects (which is outside the scope of this issue)

### Reproduction steps

1. Open the settings page
2. Press tab a few times

### Expected behavior

It should let you toggle addons being enabled/disabled and reset addons, at a minimum.

### Actual behavior

It does not let you toggle addons as the switches are not focusable. Similarily addons cannot be reset because the button is not focusable.

### Environment

 - Browser: any
 - Operating system: any

TheColaber:
Resolves #2532

### Changes

- [x] Settings Page
- [x] Popup
- [x] Messaging Addon
- [x] Cloud Games Addon
- [x] Lightmode?

### Reason for changes

I was having some fun. And I'm pretty used to enabling addons with my keyboard now. This is also a cool thing to do lol.

### Tests

This may lead to some regressions but hopefully not :)

### DIFF
--- a/popups/cloud-games/popup.html
+++ b/popups/cloud-games/popup.html
@@ -6,7 +6,6 @@
     <meta charset="UTF-8" />
     <title></title>
     <style>
-      @import url("../../webpages/styles/components/tabbable.css");
       @import url("../../webpages/styles/components/popup.css");
     </style>
     <script src="../../webpages/set-lang.js" type="module"></script>
--- a/popups/cloud-games/popup.html
+++ b/popups/cloud-games/popup.html
@@ -6,6 +6,7 @@
     <meta charset="UTF-8" />
     <title></title>
     <style>
+      @import url("../../webpages/styles/components/tabbable.css");
       @import url("../../webpages/styles/components/popup.css");
     </style>
     <script src="../../webpages/set-lang.js" type="module"></script>
--- a/popups/scratch-messaging/light.css
+++ b/popups/scratch-messaging/light.css
@@ -1,21 +1,17 @@
-body {
-  background: white;
-  color: #222;
+:root {
+  --accent-1: #f7f7f7;
+  --accent-2: #f7f7f7;
+  --accent-3: #ecebeb;
+  --accent-4: #555;
+  --accent-5: #222;
 }
 
-button {
-  border: 1px solid #8f8f8f;
-  background: #e0e0e0;
-  color: #222;
-}
 #bottom-bar {
   background: #e3e3e3;
 }
 
 .message-type {
-  border: 1px solid #cfcfcf;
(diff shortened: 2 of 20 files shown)


WorldLanguages:
Shouldn't we be adding `tabindex` attr instead of a `tababble` class? I'm confused

TheColaber:
@WorldLanguages There are 2 options when wanting to make stuff tabable.
1. take your div and put a `tabindex` on it. then also put listeners for when enter is clicked and then make it run your `@click` function
2. take your div and turn it into a button and add a class so it doesn't look like a button
The second one is much simpler.

mxmou:
> @WorldLanguages There are 2 options when wanting to make stuff tabable.
> 
> 1. take your div and put a `tabindex` on it. then also put listeners for when enter is clicked and then make it run your `@click` function
> 2. take your div and turn it into a button and add a class so it doesn't look like a button
>    The second one is much simpler.

I also think using `<button>`s is the right way, but I think it's better to add necessary styles to existing CSS instead of a generic `.tabbable` class.

WorldLanguages:
I agree with mxmou

TheColaber:
I guess maybe I've named the class wrong. `tabbable` is supposed to remove the default button look and also show a border on focus.

mxmou:
> I guess maybe I've named the class wrong. `tabbable` is supposed to remove the default button look and also show a border on focus.

I understand that, but I think defining the focused appearance separately for every button would be better.

Could `:focus-visible` be used for the focus styles? It's supported in Chrome 86+ and Firefox 85+ and only matches focused elements when the keyboard is being used for navigation. In older versions the default focus outline would be shown.

TheColaber:
> defining the focused appearance separately for every button would be better.

there are a lot of buttons, this would be a lot of repetitive lines.

> Could :focus-visible be used for the focus styles?

Good idea

mxmou:
> > defining the focused appearance separately for every button would be better.
> 
> there are a lot of buttons, this would be a lot of repetitive lines.

I think the focused appearance should match the hovered one when possible.

Using a class to override the default appearance just doesn't seem like a good idea - it would be if you wanted all buttons to look the same, but you don't so each should have its own style. Even if it's a bit repetitive.

WorldLanguages:
Will merge #3036 first, then we can come back to this.

TheColaber:
going to create a new pr for this since conflicts are going crazy and I need to remove the tabable class anyway :(

PR #2881, opened 2021-06-29
Not a pull request

lnsanityGames:
But what is this


CST1229:
magnifier.potatophant.net addon. iirc w_l rejected it, because it uses an external server.

apple502j:
We do allow addons that use external server. We do not allow pull requests that are self-contradicting, though.

lnsanityGames:
Closing because why is this in the master directory

CST1229:
> 
> 
> Closing because why is this in the master directory

potatophant made it like that, not insanitygames

PR #2888, opened 2021-06-30
Fix bugs with urlChange event

### LINKED ISSUE #2885: Using browser back and forward buttons doesn't trigger `addon.tab.urlChange`
`addon.tab.urlChange` is only triggered when the current URL is changed by Scratch, not when back or forward buttons are used after that. To reproduce:

1. Enable `studio-tools`
2. Open a new studio page
3. Go to the curators tab
4. Press Back. The UI added by studio-tools is not removed.
5. Click a tab other than Curators. The UI is removed.

apple502j:
Resolves #2885 
Fixes a bug where same-URL navigation added `/undefined` at the end of the `newUrl`

### DIFF
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -176,7 +176,7 @@ else bodyIsEditorClassCheck();
 const originalReplaceState = history.replaceState;
 history.replaceState = function () {
   const oldUrl = location.href;
-  const newUrl = new URL(arguments[2], document.baseURI).href;
+  const newUrl = arguments[2] ? new URL(arguments[2], document.baseURI).href : oldUrl;
   const returnValue = originalReplaceState.apply(history, arguments);
   _cs_.url = newUrl;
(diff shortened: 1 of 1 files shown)


content-scripts/inject/module.js
```diff
@@ -199,6 +199,16 @@ history.pushState = function () {
...
 };
 
+// replaceState or pushState will not trigger onpopstate.
+window.addEventListener("popstate", () => {
+  const newUrl = (_cs_.url = location.href);
+  for (const eventTarget of scratchAddons.eventTargets.tab) {
+    // There isn't really a way to get the previous URL from popstate event.
+    eventTarget.dispatchEvent(new CustomEvent("urlChange", { detail: { oldUrl: "", newUrl } }));
+  }
+  bodyIsEditorClassCheck();
```
WorldLanguages, line 209: Note that `bodyIsEditorClassCheck();` doesn't fix any bugs that we're aware about. But it's still a good idea.

PR #2900, opened 2021-07-02
Fix addon changing order when toggling addons

TheColaber:
Resolves #2898

### Changes

`manifest._wasEnabledBeforeSearch` which gets evaluated when addons get loaded and when the search is updated.

### Tests

Tested. Works perfectly.

EDIT: #2900!!!!!

PR #2909, opened 2021-07-04
Fix addons changing order after toggle on search

WorldLanguages:
Better fix than #2900, by avoiding giving Fuse an object owned by Vue
I'll clean up this later, I promise, it's pretty messy rn


WorldLanguages:
In the future, manifest should be read only. Modifiable state (addon settings, enable state) should be in a separate object.

PR #2911, opened 2021-07-05
uppercase addon

MondoComputazionale:
Resolves #

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -80,6 +80,7 @@
   "redirect-mobile-forums",
   "expanding-search-bar",
   "debugger",
+  "uppercase",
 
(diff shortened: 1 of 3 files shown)


Shluffy:
You should probably first open an issue about this, then not make the pr off the `master` branch, instead maybe something like `uppercase-addon`. tags `dsa`, `uppercase`, and `dyslexia` do not exist.

Shluffy:
and maybe keep the names shorter?

WorldLanguages:
@MondoComputazionale For very specific addons like this, it's probably better to write them as a userstyle for userstyle managers like Stylus, rather than an addon.

PR #2918, opened 2021-07-06
New studio stuff

apple502j:
Note that this PR is for v1.16.x, master will be updated after merging

PR is draft


addons/exact-count/studio.js
```diff
@@ -24,14 +24,8 @@ export default async function ({ addon }) {
...
-    });
-  } else {
-    countProjects(apiUrlPrefix, 0, initialDelta, function (count) {
-      document.querySelector("[data-count=projects]").innerText = count;
-    });
-  }
+  const countElement = await addon.tab.waitForElement(".studio-tab-nav > a:first-child .tab-count");
+  countProjects(apiUrlPrefix, 0, initialDelta, function (count) {
+    countElement.innerText = `(${count})`;
+  });
```
WorldLanguages, line 30: Note: apparently this is not just "exact", but accurate! Scratch shows 13 projects for https://scratch.mit.edu/studios/1060340/, but there's only 3. It's probably counting unshared projects.

PR #2942, opened 2021-07-08
editor-messages: add support for browser containers

lisa-wolfgang:
The current implementation of the addon gets the message count from the account that the messaging extension is signed in to. This works fine for most cases, but when Firefox's containers feature is utilized to sign in to an alt account, the message count does not show the correct number.

### Changes
The addon now determines which user is signed in to the editor tab and makes a direct API call to Scratch to get the message count.

### Reason for changes
An incorrect message count shouldn't be displayed near the alt's username:

image

### Tests
Tested in Chromium when signed in and when signed out.
Tested in Firefox across containers.

### DIFF
--- a/addons/editor-messages/userscript.js
+++ b/addons/editor-messages/userscript.js
@@ -15,7 +15,7 @@ export default async function ({ addon, global, console, msg }) {
     );
     const msgCount = Number(msgRes.count);
     messageCount.innerText = msgCount;
-    if (msgCount === 0 || msgCount === 'NaN') {
+    if (msgCount === 0 || msgCount === "NaN") {
       messageCount.setAttribute("style", `display: none;`);
     } else {
       messageCount.setAttribute("style", "");
--- a/addons/editor-messages/userscript.js
+++ b/addons/editor-messages/userscript.js
@@ -15,7 +15,7 @@ export default async function ({ addon, global, console, msg }) {
     );
     const msgCount = Number(msgRes.count);
     messageCount.innerText = msgCount;
-    if (msgCount === 0) {
+    if (msgCount === 0 || msgCount === 'NaN') {
       messageCount.setAttribute("style", `display: none;`);
     } else {
       messageCount.setAttribute("style", "");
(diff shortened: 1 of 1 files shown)


apple502j:
I'm sure there are so many things that can break with this. Maybe we need to re-design the core.

lisa-wolfgang:
The worst thing that can happen is the message count turning to `NaN`, which would only happen if the API couldn't be connected to, so the likelihood of something actually breaking is extremely slim.

To cover the case of the API being unavailable, the message count could either not show (which is the same behavior as the extension badge and the same behavior as before) or display a red badge with an exclamation point that on click/hover would tell the user that the message count is currently unavailable. I'm going ahead with the first option for now.

lisa-wolfgang:
The Scratch API call is also cached by the browser, so this is even more of an edge case, but now all scenarios should be accounted for.

RedGuy12:
we use `nocache=${Date.now()}` so caching is not a problem.

apple502j:
We assume that the backend and frontend share the cookies globally; there's more that could break. For example, addons that use `addon.auth` API would break.

addons/editor-messages/userscript.js
```diff
@@ -9,9 +9,13 @@ export default async function ({ addon, global, console, msg }) {
   messageCount.classList.add("sa-editormessages-count");
   messages.appendChild(messageCount);
   const setMessages = async () => {
-    const msgCount = Number(await addon.account.getMsgCount());
+    let user = document.querySelector("[class*='account-nav_profile-name']").innerHTML;
```
RedGuy12, line 12: innerText

RedGuy12 (changes requested):
tbh, i think it'd be better to figure out how to make `addon.account.getMsgCount()` supported with containers

addons/editor-messages/userscript.js
mxmou: textContent

WorldLanguages:
@apple502j Can't we get the same message count that scratch-www uses with Redux? As long as the count is still fetched while inside the editor (if not, it would be useless for our usecase)

apple502j:
Closing for now; will do myself, with redux probably

PR #2955, opened 2021-07-08
Add dynamic enable/disable to some addons

mxmou:
See commits for the list of changed addons.


addons/bitmap-copy/userscript.js
```diff
@@ -14,5 +13,8 @@ export default async ({ addon, console }) => {
       .copyImage(dataURL)
       .then(() => console.log("Image successfully copied"))
       .catch((e) => console.error(`Image could not be copied: ${e}`));
-  });
+  };
+  addon.tab.redux.addEventListener("statechanged", handleCopy);
+  addon.self.addEventListener("disabled", () => addon.tab.redux.removeEventListener("statechanged", handleCopy));
+  addon.self.addEventListener("reenabled", () => addon.tab.redux.addEventListener("statechanged", handleCopy));
```
TheColaber, line 19: couldn't you have just added an if check to return if it's disabled? there are a lot of other examples of this below.

TheColaber (changes requested):

addons/bitmap-copy/userscript.js
WorldLanguages: Simply returning is recommended as long as it produces cleaner code. No need to worry about memory leaks at this scale.

WorldLanguages:
We probably want to merge this together with #2852, so that dynamicEnable doesn't mess with the order of the UI.

addons/copy-message-link/profile.css
```diff
@@ -0,0 +1,7 @@
+#comments .comment .report {
+  display: block;
+  opacity: 0;
+}
+#comments .comment:hover .report {
+  opacity: 1;
+}
```
WorldLanguages, line 7: Explain these changes...?

addons/drag-drop/userscript.js
```diff
@@ -125,5 +125,11 @@ export default async function ({ addon, global, console }) {
...
   };
 
-  document.addEventListener("dragover", globalHandleDragOver, true);
+  document.addEventListener("dragover", globalHandleDragOver, { useCapture: true });
+  addon.self.addEventListener("disabled", () =>
+    document.removeEventListener("dragover", globalHandleDragOver, { useCapture: true })
+  );
+  addon.self.addEventListener("reenabled", () =>
+    document.addEventListener("dragover", globalHandleDragOver, { useCapture: true })
+  );
```
WorldLanguages, line 134: `if (addon.self.disabled) return;` inside the event listener would be cleaner.

addons/remove-sprite-confirm/userscript.js
```diff
@@ -17,4 +17,7 @@ export default async ({ addon, console, msg }) => {
     );
     return Promise.resolve();
   };
+  vm.deleteSprite = newDeleteSprite;
+  addon.self.addEventListener("disabled", () => (vm.deleteSprite = oldDeleteSprite));
+  addon.self.addEventListener("reenabled", () => (vm.deleteSprite = newDeleteSprite));
```
WorldLanguages, line 22: Same as above

WorldLanguages: Here, it's not just about cleaner code - if 2 addons both pollute this, and because of luck this one polluted first, it means the second addon polluted this addon's pollution. If this addon does this, it might affect other addons, by removing their polluted method from the VM.

addons/youtube-fullscreen/userscript.js
```diff
@@ -1,5 +1,12 @@
...
-  player.src += ""; // reload video
+
+  function setFullscreenEnabled(enabled) {
+    if (enabled) player.setAttribute("allowfullscreen", true);
+    else player.removeAttribute("allowfullscreen");
+    player.src += ""; // reload video
+  }
+  setFullscreenEnabled(true);
+  addon.self.addEventListener("disabled", () => setFullscreenEnabled(false));
+  addon.self.addEventListener("reenabled", () => setFullscreenEnabled(true));
```
WorldLanguages, line 11: Is it reasonable to reload the YouTube iframe dynamically? IMO this addon should have never had dynamicEnable, nor it should also get dynamicDisable

addons/copy-message-link/profile.css
mxmou: `displayNoneWhileDisabled` wouldn't work if `display: none` is used to hide the buttons when not hovered.

addons/youtube-fullscreen/userscript.js
mxmou: I'll remove `dynamicEnable` and `dynamicDisable` from this addon.

addons/remove-sprite-confirm/userscript.js
WorldLanguages: These 2 lines should have been removed.

addons/remix-tree-button/main.js
```diff
@@ -15,6 +15,8 @@ export default async function ({ addon, global, console, msg }) {
 
           const remixtreeSpan = document.createElement("span");
           remixtreeSpan.innerText = msg("remix-tree");
+
+          addon.tab.displayNoneWhileDisabled(remixtree);
```
WorldLanguages, line 19: Why empty line?

WorldLanguages (approved):

mxmou:
I don't think Redux needs to be initialized explicitly: https://github.com/ScratchAddons/ScratchAddons/blob/master/addon-api/content-script/ReduxHandler.js#L12

apple502j:
Redux must be initialized because that operation is racy.

PR #2958, opened 2021-07-08
General-Purpose redirection addon

### LINKED ISSUE #2718: New addon type: redirection

FunctionalMetatable:
Closes #2718

### Changes

Adds one addon for redirecting

### Reason for changes

So the easterEggs won't get cluttered up with redirections. 

### Tests

Tested locally.

This is also based off of @nuxtjs/redirect-module

cc @Shluffy I've added your redirect here


WorldLanguages:
It's unlikely we'll accept any redirects other than /m/discuss→/discuss as addons, even if part of a general purpose redirects addon.

PR #2961, opened 2021-07-09
Improve warning and notice colors

### LINKED ISSUE #2948: Notices are more visible than warnings in dark mode
In dark mode, the background color of notices stands out more than that of warnings.

mxmou:
Resolves #2948

### Changes

Changes the colors used for warnings and notices.

Dark mode screenshot
Light mode screenshot

### DIFF
--- a/webpages/settings/components/addon-body.html
+++ b/webpages/settings/components/addon-body.html
@@ -193,7 +193,8 @@
   }
   .addon-notice,
   .addon-warn {
-    padding: 10px;
+    padding: 8px;
+    border: 2px solid;
     border-radius: 5px;
     font-weight: bold;
     margin: 10px 0;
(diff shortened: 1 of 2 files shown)


Hans5958:
My initial suggestion is to have the notice have a blue tint (notices aren't quite important compared to warnings) and the warnings have an orange/red tint (usually for things you really need to know). Both are relatable colors that often be found in lots of places.

WorldLanguages:
IMO red is too red, brown reminds me of 💩 
sorry I'm being honest

mxmou:
> IMO red is too red

I could make it orange, but I wanted it to look like a warning.

> brown reminds me of 💩

@Hans5958 suggested a blue color, which would work better on a dark background.

Hans5958:
Or, you know, just make it yellow like a yellow warning sign.

Secret-chest:
Well try Material Design colors because the first one is transparent #FFFF00 and the second one is #FF0000 and default HTML colors don't work well

mxmou:
Update:
image
image

I didn't change the colors themselves because I think they make sense (notice = important = yellow, warning = dangerous = red).

Hans5958:
Another alternative is to just don't use transparency? Have it so it has the same orange on dark and light.

mxmou:
> Another alternative is to just don't use transparency? Have it so it has the same orange on dark and light.

That's possible, but it would probably make notices look brighter than warnings in dark mode again.

PR #2964, opened 2021-07-09
Add Popup.getSelectedTabUrl

apple502j:
Implements a new API that can be used by popups like cloud-games.

### DIFF
--- a/addon-api/popup/Popup.js
+++ b/addon-api/popup/Popup.js
@@ -14,4 +14,21 @@ export default class Popup {
   get isLightMode() {
     return scratchAddons.isLightMode;
   }
+
+  /**
+   * Gets the URL of the Scratch page that is selected, or null.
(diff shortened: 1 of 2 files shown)


addon-api/popup/Popup.js
```diff
@@ -14,4 +14,21 @@ export default class Popup {
...
+   * Gets the URL of the Scratch page that is selected, or null.
+   * @returns {Promise<?string>} - the URL
+   */
+  getSelectedTabUrl() {
+    return new Promise((resolve) => {
+      chrome.tabs.query(
+        {
+          active: true,
+          currentWindow: true,
+          url: "https://scratch.mit.edu/*",
```
WorldLanguages, line 28: Don't forget in older Firefox versions we do support, we cannot query tabs by URL because we don't have access to the URL. We might want to do an ugly workaround that involves content scripts to make it work there.

PR #2967, opened 2021-07-09
editor-stepping: Allow changing highlight color

### LINKED ISSUE #2957: [editor-stepping] Ability to change the highlighted color
I don't want blue to be the only color available for the highlighting blocks addon.

Setting to change color for the highlight in the editor-stepping addon

CST1229:
Resolves #2957

### Changes

Adds a color option to recolor the border in the editor-stepping (Highlight currently executing blocks) addon. Also slightly rewords the addon to not use the word blue. The default color is still #0000FF blue.

### Reason for changes

Some people may not like the blue or think it's too strong:
>**FunctionalMetatable wrote:**
>
>I don't want blue to be the only color available for the highlighting blocks addon.

### Tests

Tested on Firefox, should work on Chrome.
redgreenorangeblackwhite

### DIFF
--- a/addons/editor-stepping/userscript.js
+++ b/addons/editor-stepping/userscript.js
@@ -1,11 +1,9 @@
 export default async function ({ addon, global, console }) {
   const vm = addon.tab.traps.vm;
 
-  function stepping_updateColor() {
-    document.getElementById("editor-stepping-flood").setAttribute("flood-color", addon.settings.get("highlight-color")); //Change the color
-  }
+  const setColor = () =>
+    document.getElementById("editor-stepping-flood").setAttribute("flood-color", addon.settings.get("highlight-color"));
 
-  // Insert this amazing filter
   document.body.insertAdjacentHTML(
     "beforeend",
     `
@@ -31,7 +29,7 @@ export default async function ({ addon, global, console }) {
 </svg>
 `
   );
-  stepping_updateColor(); //Set the color
+  setColor();
   // Wait for Blockly, as it tends to not be ready sometimes...
   await addon.tab.traps.getBlockly();
   const elementsWithFilter = new Set();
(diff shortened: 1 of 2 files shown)


addons/editor-stepping/userscript.js
```diff
@@ -15,7 +15,7 @@ export default async function ({ addon, global, console }) {
       </feFuncA>
     </feComponentTransfer>
 
-    <feFlood flood-color="blue" flood-opacity="1" result="outColor">
+    <feFlood flood-color="` + addon.settings.get("highlight-color")  + `" flood-opacity="1" result="outColor">
```
RedGuy12, line 18: Do we need to worry about xss here?

RedGuy12 (changes requested):

addons/editor-stepping/userscript.js
apple502j: Yes.

apple502j:
Cannot merge due to sec regressions

addons/editor-stepping/userscript.js
CST1229: It's just a color input, however modifying the config file could do something here. But I don't think it's such a big thing to worry about

CST1229:
> 
> 
> Cannot merge due to sec regressions

Can you explain?

addons/editor-stepping/userscript.js
apple502j: We won't merge this until this is fixed.

WorldLanguages:
IMO xss isn't a big deal here...
Update: recent commit is cleaner code and "safer" so problem solved.

CST1229:
> IMO xss isn't a big deal here...
Yeah I was thinking that, it's literally a color input.

WorldLanguages:
Well, it's most likely nothing bad would happen with the previous version, but a chain of bugs could happen and an addon doing this could be the last one on the chain

WorldLanguages:
You can enable Prettier in your fork's action tab, or alternatively tell me when this is ready to review and I'll lint myself

WorldLanguages (approved):

PR #2981, opened 2021-07-11
Fix lint errors

apple502j:


addon-api/content-script/Tab.js
```diff
@@ -361,7 +361,7 @@ export default class Tab extends Listenable {
     let insertAfter = null;
 
     const children = Array.from(spaceElement.children);
-    for (let indexString in children) {
+    for (let indexString of children) {
```
TheColaber, line 364: I dont think this works

apple502j: whoops

PR #2988, opened 2021-07-12
v1.17 update notification

WorldLanguages:
Image TODO, merging to get strings into translation

image


WorldLanguages:
@mxmou Do you think we should make it so that the default setting for old-studio-layout changes to scratch2.0→3.0 if the user had that addon enabled at the time of the update?

mxmou:
> @mxmou Do you think we should make it so that the default setting for old-studio-layout changes to scratch2.0→3.0 if the user had that addon enabled at the time of the update?

I think that's a good idea.

PR #3001, opened 2021-07-13
Fix dropdown on addon-body look like it's hovering

TheColaber:
I'm not sure what to call this, but it causes this:
image
The mouse being pointer is ok, but the dropdown should not look like it's being hovered.

Regressed by #2949, I'm sure this wasn't intensional.

### DIFF
--- a/webpages/settings/components/addon-body.html
+++ b/webpages/settings/components/addon-body.html
@@ -175,7 +175,7 @@
     padding: 5px;
     border-radius: 4px;
   }
-  .clickeable-area:hover .btn-dropdown {
+  .btn-dropdown:hover {
     background: rgba(255, 255, 255, 0.05);
   }
 


mxmou:
> Regressed by #2949, I'm sure this wasn't intensional.

It's meant to indicate that the addon will be expanded when clicked.

TheColaber:
There is already a `cursor: pointer`... Why do we need to make something look hovered, when it's not?

WorldLanguages:
Current behavior on master looks OK to me.

WorldLanguages:
> I'm sure this wasn't intensional.

It was

mxmou:
> > I'm sure this wasn't intensional.
> 
> It was

It was, although I should have mentioned it in the PR description.

The previous behavior was strange because there is no difference between clicking the title and clicking the dropdown button, but only the dropdown button changed on hover.

PR #3034, opened 2021-07-15
better-img-uploads: remove gif from acceptable files

### LINKED ISSUE #3027: Problems when GIF loaded in HD
image
The gif plays in the costume preview because we haven't split the gif into many images like scratch.

apple502j:
Resolves #3027 

Reasons for this change:
- GIF files are already of poor resolution in most cases
- Most GIF files are animated and hard to deal with (do we see static GIF anywhere?)
- We can't add omggif (GIF parsing library used by Scratch) just for this small feature


ErrorGamer2000:
LOL, how did I not think about that... Maybe we can split it like scratch normally does?

CubeyTheCube:
> LOL, how did I not think about that... Maybe we can split it like scratch normally does?

Apple said:
> We can't add omggif (GIF parsing library used by Scratch) just for this small feature

Issue #2298, opened 2021-04-26
Redo theme system


GrahamSH-LLK:
There should be a library for setting theme, and the code should be cleaned up (it's really confusing rn).
I'll tackle this soon.

mxmou:
Which theme system? Light/dark mode for the settings page and popup?

If so, changing the settings page to use CSS variables for all colors would make working with themes much simpler.

GrahamSH-LLK:
> Which theme system? Light/dark mode for the settings page and popup?
> 
> If so, changing the settings page to use CSS variables for all colors would make working with themes much simpler.

That is what I mean. For the most part, I'll fix the JS, do you want to help variablify the rest of the css 🤞?

TheColaber:
This is resolved, right? Is there any more that needs to be done?

mxmou:
> This is resolved, right? Is there any more that needs to be done?

The JS for setting the theme wasn't changed yet.

WorldLanguages:
IIRC it was cleaned up in some places but not all

Issue #2593, opened 2021-05-24
Weird popup styling issues in Chrome, just sometimes


WorldLanguages:
Appears to be a Chrome bug, not ours. Going to devtools, unchecking any style near the affected area, then rechecking it, appears to make it return to normal. So Chrome is probably not rerendering/recalculating/whatever when it should.
Screenshots were taken by clicking the popup icon many times until the issue appeared

Text wrapping in navbar
![image](https://user-images.githubusercontent.com/17484114/119379740-3aa5e900-bc96-11eb-9e42-d01bd32d338c.png)

Tab padding inexistent
![image](https://user-images.githubusercontent.com/17484114/119379770-4396ba80-bc96-11eb-8e81-507e0be43d62.png)

WorldLanguages:
When the styling issue in the first screenshot happens, running this in the console appears to fix it:
```js
const header = document.querySelector("#header");
header.style.display = "block";
requestAnimationFrame(() => header.style.display = "");
```
Of course this would not be the only way. Any way of forcing Chrome to reconsider would work

Issue #257, opened 2020-09-10
[Question] What the heck is this checkbox?! lol


Explosion-Scratch:
What is the checkbox on the top at the far right? lol
![image](https://user-images.githubusercontent.com/61319150/92761594-780c5580-f357-11ea-9d1d-bbc8d77ac046.png)

BoomerScratch:
Probably the text doesn't show up :lol:

Explosion-Scratch:
> Probably the text doesn't show up :lol:

What does the text say then? :P

JeffaloBob:
replace stage?

Explosion-Scratch:
> replace stage?

Oh, lol, the checkboxes are on the _right_ of the text. I for some reason thought it was the other way around.

JeffaloBob:
perhaps this is a UX issue

WorldLanguages:
Lol we should probably center the text

WorldLanguages:
(or align it to the right)

towerofnix:
Perhaps a light background transition could be applied when hovering over individual option tiles?

PR #3036, opened 2021-07-16
CSS and design improvements

### LINKED ISSUE #2198: Inconsistant Scrollbars between browsers
This seems to be within the webpages and here is an example:
image
Left: FF
Right: Chrome
They should be styled to look like chrome.

mxmou:
Resolves #2198
Related to #2298
Related to #2603

### CSS changes

* Defines all colors and shadows using CSS variables, making light theme CSS much simpler.
* Uses CSS variables in popup styles.
* Removes some unused CSS and moves some CSS to places where it makes more sense (but I didn't want to do too much of that in this PR).
* Moves Chrome scrollbar styles from individual popups to `styles/components/popup.css`.
* Uses `brightness(0) invert(1)` instead of `brightness(<large number>)` to make icons white, which would also work if the icons were black.
* Moves addon group header CSS to the component template.

### Other code changes

* Removes some inline CSS and incorrect use of `&nbsp;` and `<br />`
* Uses `v-cloak` in popups instead of `style="display: none" v-show="true"` and similar hacks.
* Removes `v-cloak` from `<body>` on the settings page and adds it to some child elements instead.
* Adds `v-cloak` to some elements on the licenses and site access pages.

### Design changes

Most of these are small, but I tried to list all of them anyway.

#### General

* Improves the way addon settings are displayed in the popup.
* Uses `:focus-visible` for focus styles. This means that people using browsers that support it (Chrome 86+, Firefox 85+) will only see the focused appearance when keyboard navigation is being used. Older browsers will show the default focus outline.
* Adds a focus style to buttons
* Changes the font size of `"type": "select"` settings, popup tabs, the "Loading messages..." text, and the version number in the popup to 12px.
* Makes the button in the messaging popup look like those on the settings page (which probably didn't exist yet when the popup was designed).
* Changes the comment reply form to be more similar to our addon settings and not a copy of Scratch's.
* Makes the popup tab pop out icon change background to white on hover.
* Adds 0.2s transitions to some elements that change background color on hover, and changes 0.25s transitions to 0.2s.
* Makes the "explore all 80+ addons" message in the popup look like an addon.

#### Colors

* Changes the background color of the page in light mode to `#fcfcfc`, which is more similar to the color used before the first CSS refactor and in the popup (`#ffffff`). This is the same as the search toolbar background, so I added a border to the toolbar. The border also prevents text from looking cut off if part of it is behind the toolbar.
* Changes the popup background color from `#ffffff` to `#fcfcfc`.
* Changes the border colors of addons in light mode to `#cfcfcf`, which was used before the first CSS refactor and is still used in the popup.
* Adds a bottom border to the popup tab bar. Especially the Addons popup looks weird without it.
* Uses the same colors for buttons, select settings, and popup tabs.
* Changes the color of all blue links to `#1aa0d8` in dark mode and `#175ef8` in light mode.
* Makes scrollbars brighter in light mode. Uses the same color as addon descriptions.
* Makes the transparent black background of hovered categories and addon/message category expand buttons in light mode use the same opacity as the white one in dark mode.
* Changes the bottom border of sticky message category headers from transparent black to that used for the border of the whole category.
* Changes the username in the user's own comments to `#53b31c` in dark mode.
* Makes the messaging popup bottom bar appearance match message types.
* Uses orange in the addon blink animation.
* Uses the same shadow for tooltips and color pickers as for the top bar.
* Makes tooltips light in light mode.

#### Layout

* Removes `max-width: 75%` from titles of message categories and cloud game projects.
* Makes the padding of project titles in the cloud games popup same as in the messaging popup. Adjusts the padding of the username lists to match.
* Changes the size of the addon/message category expand button from 33x33 to 32x32 to match other buttons.
* Increases the size of collapsed message categories in the full screen messaging popup to match the settings page.
* Decreases the size of collapsed addons in the popup from 52px to 46px and slightly increases it on the settings page (from 52px to 54px).
* Aligns titles and content of non-collapsible message categories.
* Adds 5px padding below the title area of expanded addons and message categories. Decreases the padding below the last item of studio/forum activity lists.
* Makes the padding of alerts the same as that of normal comments.
* Changes the height of the messaging popup bottom bar to 32px to match buttons.
* Makes the version number in the popup aligned with the title.
* Increases the padding before collapsed addon descriptions.
* Makes spacing around addon/message category expand buttons the same on all four sides.
* Aligns addon group titles with addon type icons.
* Increases the maximum width for small mode to prevent horizontal scrollbars from appearing.
* Aligns this:  
  image
* Changes the height of select settings to match buttons.
* Makes the modal content itself scrollable, not the backdrop.
* Makes the search input smaller to match all other inputs.
* Reduces the padding of items in the small mode category menu.


mxmou:
This is now mostly finished, but still a draft because I need to check if it's compatible with other open PRs.

RedGuy12:
Can you also do something about #257? That's still a problem. I remember being confused when I first installed it

mxmou:
> Can you also do something about #257? That's still a problem. I remember being confused when I first installed it

I think moving checkboxes to the left is the best solution because that's probably what most users expect.

RedGuy12:
I was thinking moving the text so the checkboxes could still be aligned.

mxmou:
> I was thinking moving the text so the checkboxes could still be aligned.

In essentially every other software, labels are on the right of checkboxes and radio buttons.

What do you think about something like this?
image

mxmou:
The layout of settings could also be improved - addons with many settings currently look messy. Letting addons specify the layout themselves is probably too complex, so one option would be to simply use a fixed width for all settings.

mxmou:
Marked as ready for review. I'll probably have to work on compatibility when the other PRs are merged.

WorldLanguages:
> * Removes `v-cloak` from `<body>` on the settings page and adds it to some child elements instead.

I originally intended to do this, and I don't remember why I ended up adding v-cloak to body. It was probably related to the fact the sidebar of the fullscreen settings page is visible for a few ms in the popup:

image

Could this be fixed with CSS media queries?

WorldLanguages:
I think this PR is also a good opportunity to fix 2 issues:

- [ ] Ugly `rr = r` style code for handling dark mode in various places of the extension
- [ ] Fix chrome bug #2593 ourselves, by forcing rerender of the affected areas after load

WorldLanguages:
Honestly this looks weird (but probably not worse than before):

image

WorldLanguages:
> * Changes the username in the user's own comments to `#53b31c` in dark mode.

Is this for better contrast ratio?

WorldLanguages:
Excellent PR
> I'll probably have to work on compatibility when the other PRs are merged.

This might be merged first anyway, probably. This one is the most important :P

TheColaber:
> Honestly this looks weird (but probably not worse than before):
> 
> image

I don't like this. I like how it looked like a table before.

mxmou:
> I don't like this. I like how it looked like a table before.

The settings of this addon already look a bit messy and not really like a table:
image

I think making all settings the same width (or using two or so different widths) would make them much better.

mxmou:
> > * Removes `v-cloak` from `<body>` on the settings page and adds it to some child elements instead.
> 
> I originally intended to do this, and I don't remember why I ended up adding v-cloak to body. It was probably related to the fact the sidebar of the fullscreen settings page is visible for a few ms in the popup:
> 
> (image)
> 
> Could this be fixed with CSS media queries?

It could.

> Fix chrome bug [Weird popup styling issues in Chrome, just sometimes #2593](https://github.com/ScratchAddons/ScratchAddons/issues/2593) ourselves, by forcing rerender of the affected areas after load

I think that bug is related to iframes because it often happens when switching tabs. If it is, it will be fixed by #3038.

> > * Changes the username in the user's own comments to `#53b31c` in dark mode.
> 
> Is this for better contrast ratio?

Yes - the current color (`#008000`) is quite dark.

mxmou:
Possible setting layout:
image

wgyt:
So far, this looks great! Would like to see a better settings page design than the current one, but that would be a huge change, probably wouldn't fit with this PR.

WorldLanguages:
@mxmou Do you think we could keep settings layout unchanged in this PR? It will probably take a lot of tweaking until we all agree on something that looks good, and the rest can be merged. Onion skinning settings now look good, but Scratch Notifier is pretty weird:
image

CST1229:
> 
> 
> Possible setting layout:
> image

Having checkboxes on the left and everything else on the right just feels off and inconsistent.

TheColaber:
I wonder if keeping the checkbox on the right but keeping it close to the text is a better solution. I liked how the first items in the rows were all aligned.

mxmou:
> Having checkboxes on the left and everything else on the right just feels off and inconsistent.

> I wonder if keeping the checkbox on the right but keeping it close to the text is a better solution. I liked how the first items in the rows were all aligned.

Checkboxes are typically placed to the left of the label. Moving them to the right is possible, but they should probably look like switches in that case:
image

> @mxmou Do you think we could keep settings layout unchanged in this PR? It will probably take a lot of tweaking until we all agree on something that looks good, and the rest can be merged. Onion skinning settings now look good, but Scratch Notifier is pretty weird:

I think reverting the settings layout changes for now, except the popup ones, is a good idea - new `addon.json` functionality would probably need to be added for a new layout to work with all addons, which isn't what this PR is about.

WorldLanguages:
@mxmou I think a v-cloak change on this PR broke search bar autofocus on settings page in some cases. On Chromium, by opening the settings through the popup (clicking settings cog in top right), search bar is focused for a few milliseconds then loses focus

webpages/settings/components/category-selector.html
```diff
@@ -17,7 +17,10 @@
 
 <style>
   .category {
-    transition: all 0.2s ease;
+    transition: background-color 0.2s ease,
+      padding 0.2s ease;
+      height 0.2s ease,
+      opacity 0.2s ease;
```
WorldLanguages, line 23: #3532

PR #3041, opened 2021-07-16
Live comment reloading addon

### LINKED ISSUE #2215: Live Comment Refreshing
**Describe the new addon**

This addon would get the comments on a studio, project, or profile every minute and show them without the user having to reload the page. It could work with the infinite scrolling addon, possibly, or be an option on that addon.

CubeyTheCube:
Resolves #2215

### Changes

- On project and studio pages, new comments now load automatically

### Notes

- This handles most possible edge cases, but it's possible that there's a much easier way to do this that I missed
- ~~the code is pretty messy~~
- The code uses innerHTML since otherwise it would be a mess of `document.createElement`s
- This doesn't remove deleted comments, because that might generate confusion
- This doesn't work on user pages, because of https://github.com/ScratchAddons/ScratchAddons/issues/2215#issuecomment-822208275
- When you reply to a newly loaded comment, and get muted, the mute modal doesn't show up (I thought that was too big and too unnecessary an edge case to include)
- I'm not sure if it conflicts with other comment-related addons

### Tests

- Tested on Firefox
- Tested minimally on Chrome


FunctionalMetatable:
just use redux:

```javascript
    const { redux } = addon.tab;
    let [ , type, id ] = location.pathname.split('/')
    
    if (type == 'studios') {
        await redux.waitForState(state => redux.state.studio.infoStatus == "FETCHED", {
            actions: ['SET_INFO']
        })
    } else {
        await redux.waitForState((state) => state.preview.status.project === "FETCHED", {
            actions: ['SET_PROJECT_INFO']
        })
    }
    
    const URL = `https://api.scratch.mit.edu/${type == 'projects' ? `users/${redux.state.preview.projectInfo.author.username}/projects` : `studios`}/${id}/comments`
    

    setInterval(async () => {
        var loopOffset = 40
        var offset = -loopOffset
        while (true) {
            offset += loopOffset
            let res = await fetch(URL + `?offset=${offset}&limit=${loopOffset}`)
            let json = await res.json()
            if (json.length == 0) return
            let loadedCommentIds = redux.state.comments.comments.map(e => e.id)

            for (let j in json) {
                let comment = json[j]

                if (loadedCommentIds.indexOf(comment.id) !== -1) {
                    break
                }

                console.log('omg new comment')
                redux.dispatch({ type: "ADD_NEW_COMMENT", comment })
            }
        }
    }, 1000)
}
```

FunctionalMetatable:
why close

CubeyTheCube:
@FunctionalMetatable Since you can do it with redux, I felt that you should create a new PR since my code is entirely different.

PR #3055, opened 2021-07-17
Message passing

### LINKED ISSUE #1671: addon.messaging API
Maybe not that exact name but this would allow communication through:
persistent <> userscripts
persistent <> popup

userscripts <> userscripts
popup <> popup
persistent <> persistent
userscripts <> popup

ErrorGamer2000:
Resolves #1671

### Changes

Added the `addon.messaging` api (no jsDoc yet though).

API:
- `messaging.sendMessage`
  - [x] tested
  - Userscripts and Popups only
  - Example:
```js
...
addon.messaging.sendMessage({greeting: "hello"}, function(response) {
  console.log(response.farewell);
});
...
```
- `messaging.connect`
  - [x] tested
  - Userscripts and Popups only
  - Example:
```js
...
const port = addon.messaging.connect({name: "knockknock"});
port.postMessage({joke: "Knock knock"});
port.onMessage.addListener(function(msg) {
  if (msg.question === "Who's there?")
    port.postMessage({answer: "Madame"});
  else if (msg.question === "Madame who?")
    port.postMessage({answer: "Madame... Bovary"});
});
...
```
- `messaging.onMessage`
  - [x] tested
  - Persistient-scripts only
  - Example:
```js
...
addon.messaging.onMessage.addListener(
  function(request, sender, sendResponse) {
    if (request.greeting === "hello")
      sendResponse({farewell: "goodbye"});
  }
);
...
```
- `messaging.onConnect`
  - [x] tested
  - Persistient-scripts only
  - Example:
```js
...
addon.messaging.onConnect.addListener(function(port) {
  console.assert(port.name === "knockknock");
  port.onMessage.addListener(function(msg) {
    if (msg.joke === "Knock knock")
      port.postMessage({question: "Who's there?"});
    else if (msg.answer === "Madame")
      port.postMessage({question: "Madame who?"});
    else if (msg.answer === "Madame... Bovary")
      port.postMessage({question: "I don't get it."});
  });
});
...
```

I have not been able to test the popup api for this, can someone else please do it? Also, I have not Added a jsDoc for the api yet.

### Reason for changes

The `addon.messaging` api has been requested many times, and will finally allow addons that are not popups to communicate with their persistent scripts to access data in the background or saved data.

> Examples from [chrome developers](https://developer.chrome.com/docs/extensions/mv3/messaging/)

Hoping to maybe add an `addon.tabs` api soon that will allow the persistent scripts to initiate connections.


ErrorGamer2000:
Wow, my timing is waaaay off, maybe because of the way I have my PC time set... XD

ErrorGamer2000:
Every time I comment it says `1 hour ago`.

addon-api/background/MessagePasser.js
```diff
@@ -0,0 +1,141 @@
...
+ */
+
+export default class MessagePasser {
+  constructor(addonObject) {
+    this._addonId = addonObject.self.id;
+
+    this._onMessageListeners = [];
+    this._onConnectListeners = [];
+
+    let self = this;
```
RedGuy12, line 12: just use `this`? why do you need a `self` variable?

apple502j:
Please use EventTarget

addon-api/background/MessagePasser.js
```diff
@@ -0,0 +1,141 @@
+/**
+ * Handles message passing. Mimics chrome.runtime message passing.
```
RedGuy12, line 2: > Mimics chrome.runtime message passing.

pls don't

`chrome.runtime`'s is too confusing

just do two functions: `addon.messaging.send` and `addon.messaging.onRecieve`. Make them both available in all environments (popup, userscript, and bg)

```js
export default async function({ addon }) {
  addon.messaging.send({
    data: "hi",
    otherData: "hello",
  })
}
```

```js
export default async function({ addon }) {
  addon.messaging.onRecive((data) => console.log(data)) // { data: "hi", otherData: "hello" }
}

```

RedGuy12 (changes requested):

addon-api/background/MessagePasser.js
TheColaber: best if `onRecieve` is an event listener, so you can remove it and stuff.
```js
addon.messaging.addEventListener("data", ({ data } => console.log(data)))
```

addon-api/background/MessagePasser.js
ErrorGamer2000: Because the functions are not going to be called immediately always, so this will either be the window or undefined.

addon-api/background/MessagePasser.js
ErrorGamer2000: > best if `onRecieve` is an event listener, so you can remove it and stuff.
> 
> ```js
> addon.messaging.addEventListener("data", ({ data } => console.log(data)))
> ```

Okay, but I need to add the `sendResponse` function as well. How would I fix that?

and @apple502j, the same problem if I use `EventTarget`.

TheColaber: > Okay, but I need to add the sendResponse function as well. How would I fix that?

By just adding it?

> and @apple502j, the same problem if I use EventTarget.

What's the problem with an event target? All the other addon.* are event targets.

ErrorGamer2000: Nevermind...

ErrorGamer2000: Finished.

WorldLanguages:
This will probably remain as a draft for a while because current code isn't good enough to merge and message passing isn't high prio

ErrorGamer2000:
Looking through the content-scripts folder, I realize that it might be better to use Comlink . . . Any objections?

apple502j:
Please do not use Comlink for between addons or content scripts to backgrounds.

PR #1297, opened 2021-01-12
Fix Semicolon Bug

GrahamSH-LLK:
This is the better fix b/c it looks the same on all pages, scratchr2 and www, light and dark modes.
Resolves #1291

PR #3068, opened 2021-07-19
dark-www: fix `semicolon` addon showing white background and incorrect font color

### LINKED ISSUE #1604: Semicolon with darkmode bug
image

FunctionalMetatable:
Resolves #1604

### Changes
Adds dark background to semicolon when both semicolon and dark-www addons are enabled.

### Tests
Tested locally

### DIFF
--- a/addons/dark-www/experimental_scratchwww.css
+++ b/addons/dark-www/experimental_scratchwww.css
@@ -19,12 +19,13 @@ a.social-messages-profile-link,
 .studio-project-tile .studio-project-username,
 #footer,
 #donor,
+.semicolon,
 .subactions .share-date,
 .developers #faq p,
 .social-form,
 body:not(.sa-body-editor) .join-flow-title,
 .initiatives-section .initiatives-tools .subsection-tag.collaborator {
-  color: #fff !important;
+  color: #fff;
 }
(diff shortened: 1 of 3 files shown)


mxmou:
Apparently it always being light is the intended behavior. See #1297

addons/dark-www/experimental_scratchwww.css
```diff
@@ -24,7 +24,7 @@ a.social-messages-profile-link,
 .social-form,
 body:not(.sa-body-editor) .join-flow-title,
 .initiatives-section .initiatives-tools .subsection-tag.collaborator {
-  color: #fff;
+  color: #fff !important;
```
mxmou, line 27: Add `.semicolon` to the list of selectors (after `#donor`) instead of `!important` (which would likely break things).

mxmou (changes requested):
I'll review the PR anyway because I think the semicolon should be dark in dark mode, like everything else.

mxmou (approved):

WorldLanguages:
@GrahamSH-LLK ?

WorldLanguages:
Well, if someone really wants to, they may PR to bring back semicolon white background on dark mode, but hopefully not :P

PR #3075, opened 2021-07-21
Dark speech bubbles

mxmou:
### Changes

Changes editor dark mode to use the accent background for speech bubbles if the `affectStage` setting is enabled.

### DIFF
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -597,6 +597,17 @@
   "injectAsStyleElt": true,
   "versionAdded": "1.0.0",
   "updateUserstylesOnSettingsChange": true,
+  "userscripts": [
+    {
+      "url": "bubbles.js",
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Looks good
One thing we should consider tho, is be careful in case code related to this changes in a future Scratch update. "Editor dark mode" might potentially break the editor now :P
Since most editor dark mode users will have affectStage off, do you think you could refactor the code a bit so that the vm method is only trapped if affectStage was/is enabled (either on first load or settings change event)?

mxmou:
> Since most editor dark mode users will have affectStage off, do you think you could refactor the code a bit so that the vm method is only trapped if affectStage was/is enabled (either on first load or settings change event)?

Done. This means that enabling the setting after page load won't change existing text bubbles, but I don't think that's a problem because enabling the addon after page load has the same behavior.

PR #3076, opened 2021-07-21
Perform message checks every minute

apple502j:
This does not implement MV3 changes because it simply does not make sense under current implementation.


WorldLanguages:
We probably want to tackle how requesting an updated messaging count works in future versions, when getting ready for manifest V3.

PR #3077, opened 2021-07-21
Dark costume editor

mxmou:
### Changes

Makes editor dark mode affect the costume editor.

Screenshots with the default preset:
image
image

### `addon.json` changes

Adds a new `type` for `customCssVariables`, called `"makeHsv"`. It has the properties `h`, `s`, and `v` - each can be a number (in range 0 to 360 for `h` and 0 to 1 for the other two), a hex color or an object of the usual format.


addons/editor-dark-mode/addon.json
```diff
@@ -597,6 +597,18 @@
...
   "versionAdded": "1.0.0",
   "updateUserstylesOnSettingsChange": true,
+  "userscripts": [
+    {
+      "url": "paper.js",
+      "matches": [
+        "https://scratch.mit.edu/projects/*/",
+        "https://scratch.mit.edu/projects/*/editor",
+        "https://scratch.mit.edu/projects/*/fullscreen",
+        "https://scratch.mit.edu/projects/editor"
```
TheColaber, line 607: can be simplified to "editor"

TheColaber (changes requested):

addons/editor-dark-mode/addon.json
mxmou: I might make a PR later that changes the matches for all editor addons to "projects".

mxmou:
The contrast of the grid is currently quite small with the dark editor highlight color - I want to improve that. The center crosshair also looks a bit weird right now.

mxmou:
The grid should now look good with most (reasonable) combinations of highlight colors and accent backgrounds.
image

WorldLanguages:
Schema needs to be updated

WorldLanguages:
Funny related feedback:
> the update is pretty good but some addon is making the transparent background of some random sprite turn black I already looked and I looked and I didn't think I wanted you to help and say that addon is this.

RedGuy12:
There was also a few suggestions to change the color of the background independently of the accent color.

mxmou:
> There was also a few suggestions to change the color of the background independently of the accent color.

I think the addon already has too many settings, but this would be useful.

WorldLanguages:
> I have a suggestion for the editor dark mode. Perhaps add a setting that changes only the color of the costume background? While it can be changed with the accent color, it looks strange if paired with dark mode colors, hence why I am suggesting this. Thank you, and keep up the good work with Scratch Addons!

> I have a problem with editor dark mode. In the most recent update I think, you changed the costume editor to be dark instead of light. This is a problem for me, as I draw in bitmap with black outlines. When I'm drawing, it's really hard to see the shape I'm making. Please add an option to bring the lighter costume editor back please. :)
\
P. S. I know you can just change the accent color in the settings, but that doesn't effect just the costume editor, and all the white kind of hurts my eyes. (Also it doesn't look very nice when you have normal dark mode but there are just random spots of white.)

At this point it's probably expected we do _something_ (either make the feature togglable, the color customizable, or both) for v1.19

PR #3080, opened 2021-07-22
redirect-mobile-forums: use chrome.webRequest.onBeforeRequest for faster redirects

FunctionalMetatable:
### Changes

Refactor redirect-mobile-forums into background script which uses chrome.webRequest API

### Reason for changes

Faster redirects instead of making a request to the forums again, also because its scratchr2

### Tests

Tested locally (edge)


WorldLanguages:
This will work for now, but because of manifest V3's peculiarities, supporting persistent scripts like these might get complex. So I'd prefer we do userscript redirects until we migrate to manifest V3.

PR #3091, opened 2021-07-24
exact-count: use fetch instead of XHR, omit credentials

WorldLanguages:
Omitting credentials makes it possible for Scratch to return a cached HTML source, because there's no login data.


TheColaber:
Why do we want smth caches for an _exact_ count?

PR #3104, opened 2021-07-26
Clean up in DevTools remove unused lists.

### LINKED ISSUE #3103: Developer tools Clean up Blocks+ should ask remove unsued local lists too.
Right now, with developer tools enabled when you clean up a sprite it searches for unused local variables. If it finds some, it prompts the user to delete them. It makes scene that it would also search for unused "for this sprite only" lists and prompt the user to delete those too.

Tacodude7729:
Resolves #3103

### Changes

Added a new confirmation message the unused variable one variables asking to delete unused lists. Also added unused-list translation. 
Do I need to do something to add translations for other languages?

### Tests

Tested manually, with many used and unused lists, private and public and ensuring it doesn't delete lists with uses.


WorldLanguages (approved):
Works, waiting for second review

apple502j (approved):

PR #3110, opened 2021-07-26
Fix dark bubbles crashing when renderer is not available

mxmou:
### Changes

Makes the dark bubbles userscript not crash if the renderer is not available yet.

### Tests

Not tested yet because I don't know how to make the problem happen consistently.


GarboMuffin:
Consider using https://github.com/ScratchAddons/ScratchAddons/blob/ac65a554fde4bca65748e0026db746a52cbcb495/addons/gamepad/userscript.js#L7-L10 for a simpler/less invasive way to wait for the renderer to exist

WorldLanguages:
@mxmou Please merge if you're OK with my changes.

PR #3112, opened 2021-07-26
Added Fix CSS

UtterDonkey:
I have added an option to fix CSS bugs in the scratch editor. I came across these and they really annoyed me. They are mainly scrollbar issues.


addons/addons.json
```diff
@@ -84,7 +84,7 @@
   "disable-stage-drag-select",
   "studio-followers",
   "ctrl-enter-post",
-
```
RedGuy12, line 87: Add a new line after your addon

WorldLanguages:
Hmm @mxmou what do you think of this?

mxmou:
> Hmm @mxmou what do you think of this?

I think a problem with very small fixes like these is that people won't know if and why they need them. It would make sense to make such addons always enabled and hidden on the settings page.

addons/css-fix/userstyle.css
```diff
@@ -0,0 +1,18 @@
...
+.sprite-selector_items-wrapper_4bcOj.box_box_2jjDp {
+overflow: hidden;
+}
+.sprite-selector_items-wrapper_17MEf {
+overflow: hidden; /* Hide scrollbars */
+}
+.s3devSTT {
+border-top: 1px solid hsla(0, 0%, 0%, 0.15);
+padding: 8px 12px;
+}
```
mxmou, line 14: Looks like this is styling an element added by the developer tools addon? If so, you should change that addon, not add a new one that changes it.

WorldLanguages:
@mxmou So as an easter egg enabled by default?

mxmou:
> @mxmou So as an easter egg enabled by default?

Something similar, but not toggleable. I'm not sure how useful this is though - scrollbars appearing for no reason are annoying, but wouldn't a scratch-gui/scratch-blocks PR be more helpful? It would fix them for everyone, not just Scratch Addons users.

addons/css-fix/userstyle.css
```diff
@@ -0,0 +1,18 @@
+
+.blocklyWidgetDiv .goog-menu {
+overflow: hidden; /* Hide scrollbars */
+}
+.sprite-selector_items-wrapper_4bcOj.box_box_2jjDp {
```
mxmou, line 5: The hash at the end should not be included (use something like `[class^="sprite-selector_items-wrapper_"]`) for compatibility with TurboWarp Desktop.

mxmou (changes requested):

addons/css-fix/userstyle.css
```diff
@@ -0,0 +1,18 @@
...
+overflow: hidden; /* Hide scrollbars */
+}
+.s3devSTT {
+border-top: 1px solid hsla(0, 0%, 0%, 0.15);
+padding: 8px 12px;
+}
+.react-contextmenu-item:hover {
+background: hsla(215, 100%, 65%, 1);
+color: white;
+}
```
mxmou, line 18: What is this for?

WorldLanguages:
> Something similar, but not toggleable. I'm not sure how useful this is though - scrollbars appearing for no reason are annoying, but wouldn't a scratch-gui/scratch-blocks PR be more helpful? It would fix them for everyone, not just Scratch Addons users.

Yeah, I had the same thought.

@UtterDonkey You'll have to prove the Scratch Team is aware of the specific CSS bugs you're fixing, and show that it's unlikely it will be fixed soon by them.

addons/css-fix/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  "dynamicDisable": true,
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["https://scratch.mit.edu/projects/*"]
+    }
+  ],
+
+  "credits": [
+    { "name": "Utter Donkey", "link": "https://pixelbulb.online" }
```
RedGuy12, line 16: Add a trading slash to your website URL

RedGuy12:
Go to https://github.com/UtterDonkey/ScratchAddons/actions and enable them

addons/css-fix/userstyle.css
UtterDonkey: Sorry, that code is supposed to be something else, I will fix it as soon as possible

addons/css-fix/userstyle.css
UtterDonkey: I have removed that, it's not supposed to be there

addons/css-fix/addon.json
UtterDonkey: Do you mean trailing slash? If so, I have done that

PR #3138, opened 2021-07-28
Run 60fps, mute-project, and pause in embeds

mxmou:
Makes 60fps, mute-projects, and pause addons run in embeds again, allowing them to be used together with live featured project.

### DIFF
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -23,7 +23,7 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["projects"]
+      "matches": ["projects", "projectEmbeds"]
     }
   ],
   "settings": [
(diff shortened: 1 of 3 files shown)


FunctionalMetatable:
https://github.com/ScratchAddons/ScratchAddons/blob/45bf398599bec0ee90ec903f9631e3c0140409c4/content-scripts/cs.js
I'm not sure if this blocks project embeds addons from working in other sites

WorldLanguages:
Whoops, good catch! I knew I was forgetting something when merging that other PR but I forgot what it was :P

WorldLanguages:
Lol: (don't worry, it overflows in vanilla even without pause button and muted project indicator, as long as turbo mode is on)
image

PR #3142, opened 2021-07-28
cloud-games: show notice about incorrect data

apple502j:


apple502j:
Bug fixed upstream, no need

PR #3143, opened 2021-07-28
Add Zebra stripes addon by TheBrokenRail

### LINKED ISSUE #2742: Zebra Striping nested blocks from the same category
https://github.com/LLK/scratch-gui/issues/646

This would be cool

TheColaber:
Resolves #2742

### Changes

Adds @TheBrokenRail userscript, with support for dynamicEnable and disable.

### Tests

Tested both dyamicEnable and dynamicDisable

### DIFF
--- a/addons/zebra-stripes/addon.json
+++ b/addons/zebra-stripes/addon.json
@@ -1,0 +1,21 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/dist/schema.json",
+  "name": "Nested blocks zebra stripes",
+  "description": "Nested blocks become zibra striped.",
+  "credits": [
+    {
+      "name": "TheBrokenRail",
+      "link": "http://github.com/TheBrokenRail"
+    }
+  ],
+  "tags": ["editor"],
+  "versionAdded": "1.18.0",
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ]
+}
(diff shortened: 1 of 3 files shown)


addons/zebra-stripes/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/dist/schema.json",
+  "name": "Nested blocks zebra stripes",
+  "description": "Nested blocks become zibra striped.",
```
RedGuy12, line 4: Zibra

Also what does "zebra striping" even mean?

RedGuy12 (changes requested):

WorldLanguages:
@TheColaber Where was the code taken from?

addons/zebra-stripes/addon.json
CST1229: Basically blocks in another block with the same category alternate between becoming light and dark

TheColaber:
A post on the llk issue... https://github.com/LLK/scratch-gui/issues/646#issuecomment-325002472

WorldLanguages:
It will be challenging to make this work with editor-theme3
Will also be challenging to have an understandable addon name and description :P
And we also need to make sure block flashing works OK with these blocks (assuming one of the blocks affected by this addon can be somehow get flashed through editor-devtools or debugger, maybe it's not possible, since this affects C blocks only)

CST1229:
This has a number of issues...
- Doesn't work at all when `editor-theme3` is enabled (W_L said previously)
- Doesn't zebra-stripe if a reporter/boolean is in a stack block of the same category
- Sometimes doesn't zebra-stripe
- Looks like it doesn't zebra-stripe with extensions and some other categories like lists
- And probably more

> since this affects C blocks only

Nope! It works with reporters too.
reporters
(Not that this addon isn't already horribly broken in an SA context anyways.)

TheColaber:
closing till we have a proper implementation

PR #2947, opened 2021-07-08
Add old studio layout addon

mxmou:
Resolves #2935

### Changes

Adds an addon that reverts studios to either the old design or the 2.0 → 3.0 design. Adjusts studio-tools for compatibility with the added addon and attempts to fix a bug causing the leave button to not always appear.

### Screenshots

Default:
image

Scratch 2.0 → 3.0:
image

Adding projects:
image

PR #3144, opened 2021-07-28
More items per row addon

mxmou:
See https://github.com/ScratchAddons/ScratchAddons/pull/2947#issuecomment-877235781

### Changes

Adds an addon that allows customizing the number of items shown in a single row of a grid.


addons/items-per-row/classes.js
```diff
@@ -0,0 +1,30 @@
...
+    setClasses("studio-curators", addon.settings.get("studioCurators"), { min: [4, 5], max: [2] });
+    setClasses("projects", addon.settings.get("projects"), { min: [6, 7], max: [4, 3] });
+    setClasses("studios", addon.settings.get("studios"), { min: [6, 7], max: [3] });
+    setClasses("users", addon.settings.get("users"), { min: [12, 15], max: [] });
+  };
+  updateClasses();
+  addon.settings.addEventListener("change", () => {
+    updateClasses();
+    // Handle scratchr2's lazy image loading
+    if (addon.tab.clientVersion === "scratchr2") window.dispatchEvent(new CustomEvent("scroll"));
```
WorldLanguages, line 28: Why `CustomEvent` instead of `Event`?

WorldLanguages:
Infinite scrolling works inconsistently with 6 curators per row

WorldLanguages:
A bit offtopic but I would also like to see a reset button for number settings. Would be useful for this addon, as well as the recently added `default-project` addon. For that addon, it would be great because it has presets, and here, it would let you reset a single setting without affecting all of them.

WorldLanguages:
Do you think we should add a notice to `old-studio-layout` hinting users to set studio projects per row to 4 through this addon?

addons/items-per-row/classes.js
mxmou: There isn't really a difference in this case, but I don't think `Event` is meant to be initialized directly.

mxmou:
> Infinite scrolling works inconsistently with 6 curators per row

Is limiting the number of curators per row to 5 a good solution? The current maximum makes the cards really small anyway.

> A bit offtopic but I would also like to see a reset button for number settings. Would be useful for this addon, as well as the recently added `default-project` addon. For that addon, it would be great because it has presets, and here, it would let you reset a single setting without affecting all of them.

Reset buttons would probably be useful for all types of settings other than select and Boolean.

> Do you think we should add a notice to `old-studio-layout` hinting users to set studio projects per row to 4 through this addon?

Probably.

addons/items-per-row/classes.js
apple502j: `Event` can be constructed so we should use that

WorldLanguages:
> Is limiting the number of curators per row to 5 a good solution? The current maximum makes the cards really small anyway.

Hmm, it also works inconsistently for me with 4 or 5 curators per row.

WorldLanguages:
The problem is caused by #2980, which itself fixed a bug
We could decrease the number of pixels from 500 to, for example, 250, and it would work with up to 4 curators per row.
Or we could rethink the logic being careful not to cause the bug that same PR fixed.

WorldLanguages:
The addon itself is fine, so I'm merging and opening an issue for the infinite-scroll "bug"

PR #3146, opened 2021-07-28
Localized ScratchAddons.com URLs and UTM tags

### LINKED ISSUE #3123: Open localized versions of scratchaddons.com for farwell, welcome, feedback, etc.
We currently open `https://scratchaddons.com/welcome` on first install, but we could open `https://scratchaddons.com/es/welcome` instead, if the user has their browser set to Spanish.

FunctionalMetatable:
Resolves #3123

### Changes

Localized URLs in farewells, transition and welcome.

### Tests

Untested as i dont know how to test this sort of thing


WorldLanguages:
Okay, we need to figure out what language code to put in the URL so that this works. It might need a few tweaks in the website itself until we can merge this.

PR #3147, opened 2021-07-28
We can keep the stage the exact same size

mxmou:
See https://github.com/ScratchAddons/ScratchAddons/pull/2682#issuecomment-888570607

### Changes

Makes the TurboWarp stage the same size as Scratch's stage and removes the now unnecessary background and border from the iframe.

### DIFF
--- a/addons/turbowarp-player/style.css
+++ b/addons/turbowarp-player/style.css
@@ -39,9 +39,8 @@
 }
 
 .sa-tw-iframe {
-  width: calc(100% + 8px);
+  width: 100%;
   height: calc(100% + 10px);
-  margin-left: -4px;
   margin-top: -3px;
   border: none;
 }
--- a/addons/turbowarp-player/style.css
+++ b/addons/turbowarp-player/style.css
@@ -40,7 +40,12 @@
 
 .sa-tw-iframe {
   width: 100%;
+  /* TW scratch-gui/src/lib/screen-utils.js, line 57
+     The stage height should be 360px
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Negative margin, I knew, but I'm a CSS noob 😂

WorldLanguages:
I added the background/border to have some UI indication that this is not the vanilla Scratch player, do you think we should remove these entirely or try to remind the user in some other way?

addons/turbowarp-player/style.css
```diff
@@ -26,19 +26,22 @@
...
+
+  /* Hide part of loading screen to make it the same size as Scratch's */
+  overflow: hidden;
+}
+
+.sa-tw-iframe {
+  width: calc(100% + 8px);
+  height: calc(100% + 10px);
+  margin-left: -4px;
+  margin-top: -3px;
```
WorldLanguages, line 45: And can you add comments on where these magic numbers come from? Or is it by eye?

mxmou:
> I added the background/border to have some UI indication that this is not the vanilla Scratch player, do you think we should remove these entirely or try to remind the user in some other way?

I don't think reminding the user is necessarily - there isn't a way to always automatically replace the Scratch player with the TurboWarp one, so a button needs to be clicked for the replacement to happen anyway.

WorldLanguages:
Do I merge or will you add CSS comments?

addons/turbowarp-player/style.css
mxmou: I added the comments.

PR #3038, opened 2021-07-16
Turn popup iframes into vue components

TheColaber:
Resolves #2959
Resolves #1316 
Resolves #2593

### Changes

todo

### Todo
- [x] fullscreen
- [x] settings page
- [x] new light mode style feature
- [x] scrollbar :(
- [x] rm stuff

### Reason for changes

to resolve stuff

### Tests

todo

Issue #2785, opened 2021-06-15
Popups `settingsMatch`


RedGuy12:
### Is your request related to a problem? Please describe.

I want to add a popup version of image uploader, but some users might not want a popup.

### Describe the solution you'd like

Allow more customization of popups with `settingMatch`, like we have for USs, but for popups

WorldLanguages:
> I want to add a popup version of image uploader

Can you elaborate why?

RedGuy12:
So I can upload images to other fields (notes and credits, wiwo, etc) easierly

Or we could make a drag and drop system for those to. But the suggestion still applies. Popup settingsMatch could be helpful for future addons.
________________________________

PR #3151, opened 2021-07-29
Replace `settingsMatch` with more general `if`

RedGuy12:
Instead of
```json
"settingMatch": {
  "id": "persistentThumb",
  "value": true
}
```
You now do
```json
"if": {
  "settings": { "persistentThumb": true }
}
```
Which has the same result.

You can also do
```json
"if": {
  "addonEnabled": "debugger"
}
```
Which will only load the asset if debugger is enabled.
It can also be an array of strings instead of a string. If any in the array matches, the whole condition matches.
If there's both types of ifs, as long as one matches, the whole if matches.

---
You can add more keys to `addons` or `settings` and you can use both together. 
Please note that **all** conditions must be met.
However you can do
```json
"settings": { "alternativePlayer": ["turbowarp", "forkphorus"] }
```
to load the asset if `alternativePlayer` is `turbowarp` OR `forkphorus`.

---

You can also add the `if` key on settings. Usage is same as above. Please note that the setting can and will still have a value.

**Using `if` here only hides the setting fron the user**. The setting just shouldn't change anything in that state. Ie, the zoom controls addon. "Autohide duration" has no effect if "Autohide" is off.

---

I added settings for debugger addon in theme3 and block switching, **but I did not write any code for them.  They are only for testing the new option.** Ideally I'd add code for them but I don't know how xD. **Before this is merged, the seems should either be removed or code should be added to implement them**

---

Will add #2785 in the same way after #3038.

Not interested in making it compatible with #2875. Not sure if it'd even be possible with my current implementation. Nor would it be useful.

---

I can write docs, can try to update manifest schema


WorldLanguages:
@mxmou The new capability to hide certain settings depending on a condition will break an assumption on the settings page, mainly that all settings are shown all the time.

WorldLanguages:
Also, it is not recommended to use settingsMatch/if for userscripts, since that doesn't play very well with dynamicDisable/dynamicEnable. It's ok to use it with userstyles.

PR #3152, opened 2021-07-29
infinite-scroll: Fix donor text with "fix footer"

### LINKED ISSUE #1600: Footer part missing if fixed footer
For the "fix footer" in "infinite scrolling" it doesn't show this part at the bottom.
image

*Note: This is because the bottom text is a different div than the footer.

CST1229:
Resolves #1600

### Changes

In the `infintie-scroll` addon, this adds a simple userscript for the home page that simply moves the donor text (the "Scratch is available for free..." text) into the footer, allowing it to show when the footer is hovered over instead of being at the bottom of the page.
This also adds a CSS variable for the hovered footer's height so that the userscript can change it to make the text visible once it runs.

### Reason for changes

Not having the donor text be with the footer just looks weird.

### Tests

Seems to work.
pic
The donor text script moves the text about 1-2 secs after the footer style loads which is okay.

### DIFF
--- a/addons/infinite-scroll/moveDonor.js
+++ b/addons/infinite-scroll/moveDonor.js
@@ -1,0 +1,8 @@
+export default async function ({ addon, global, console }) {
+	const root = document.documentElement; //The :root element
+	const footer = await addon.tab.waitForElement("#footer"); //The footer
+	const donor = await addon.tab.waitForElement("#donor"); //The donor text
+	
+	footer.appendChild(donor); //Move the donor text to inside the footer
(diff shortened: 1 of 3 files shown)


addons/infinite-scroll/moveDonor.js
```diff
@@ -0,0 +1,8 @@
+export default async function ({ addon, global, console }) {
```
FunctionalMetatable, line 1:
```suggestion
export default async function ({ addon, global, console }) {
```

PR #2799, opened 2021-06-16
Use more consistent language and extended search

mxmou:
Resolves #2243

### Changes

- Makes some addon names shorter (#2764).
- Changes most addon names to nouns (#2764).
- Changes addon names and descriptions to use consistent words for parts of Scratch, and generally specifies the locations of added UI features more accurately. Removes some unnecessary mentions of Scratch.
- Enables extended search in fuse.js, mainly to allow things like searching for "record video" to get "video recorder". "Smart block switching" no longer works, but a better solution for that is telling users to remove some keywords and link words when there are no results. Note that different fields (name, description, ...) are searched separately, which is not ideal but it's how fuse.js currently works. I'm not sure if there is a good way to work around that.

No Title Case though. It would cause reading the list of addons to take longer because you stop for a moment at Each Capitalized Word.

### Reason for changes

To make addon names and descriptions more consistent.

### Tests

You can still search for the previous addon names and get results.

PR #2908, opened 2021-07-04
Enable recommended addons by default, new "featured" tag

WorldLanguages:
Resolves #2617

- Create featured tag
- Use yellow color for featured tag
- Use new lightblue color for "for website" tag, that previously used the yellow
- Create featured addon group
- Create forums addon group, and have it unexpanded by default. This makes the "others" addon group cleaner. We have around 10 forum addons, and most of our users will never use them.
- Rename `New!` addon group to `New`. Tag is still called `New!`. Allow translators to use plurals in languages that support it (example: Spanish `recomendados` will be the addon group name, and `recomendado` will continue to be the tag).

Note that addons that say "enabled by default" will only be enabled by default for new users.

\- 2d-color-picker: others → featured (changes UI, cannot be recommended)
\- account-settings-capitalize: easter egg → others (small addon)
\- better-featured-project: recommended → featured (widely used, cannot be recommended)
\- block-pallete-icons: others → featured (widely used)
**\- block-switching: enable by default** (scratch 2.0 feature, was already recommended)
\- cat-blocks: others → featured (widely used)
\- clones: others → featured (widely used)
**\- color-picker: enable by default** (useful, additive, was already recommended)
**\- confirm-actions: enable by default** (only setting enabled by default is confirming sharing projects, others require optin)
**\- copy-message-link: enable by default** (useful, additive, was already recommended)
\- custom-block-shape: recommended → featured (used to achieve 2.0 looks, cannot be recommended)
\- dark-www: recommended → featured (widely used, cannot be recommended)
\- data-category-tweaks-v2: recommended → featured (scratch 2.0 feature, cannot be recommended)
\- debugger: beta → featured (useful)
**\- drag-drop: enable by default** (useful, additive, was already recommended)
\- editor-colored-context-menus: others → featured (scratch 2.0 feature)
\- editor-dark-mode: recommended → featured (widely used, cannot be recommended)
\- editor-messages: recommended → featured (widely used, cannot be recommended because it might be annoying/distracting to the user)
**\- editor-searchable-dropdowns: enable by default** (useful, additive, was already recommended)
\- editor-stage-left: recommended → featured (scratch 2.0 feature, cannot be recommended)
\- editor-stepping: others → featured (useful)
\- editor-theme3: recommended → featured (used to achieve 2.0 looks, cannot be recommended)
\- exact-count: recommended → featured (useful, cannot be recommended)
\- expanding-search-bar: others → featured (useful, scratch wiki feature)
[...]

PR #3155, opened 2021-07-29
Rename to "ocular integration"

Hans5958:
Addendum to #2799 using my phone since I'm not satisfied.

Can we have a vote now? Thumbs up if you agree, thumbs down if you don't.

### DIFF
--- a/addons/my-ocular/addon.json
+++ b/addons/my-ocular/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "ocular.jeffalo.net integration",
+  "name": "ocular integration",
   "description": "Shows ocular statuses and forum post reactions. Uses my-ocular.jeffalo.net to obtain data.",
   "credits": [
     {


WorldLanguages:
The main reasons I prefer `ocular.jeffalo.net integration`:
- Ever since post reactions were the main feature of the my-ocular addon, the addon name had the full domain on it
- I've seen people ask me multiple times on my profile "why does it say Hmmm next to Argentina"
- Above happened even tho the addon was called `Show my-ocular status` at the time
- Addon name never referred to `ocular`, it always has been `my-ocular` or `ocular.jeffalo.net`
- It discourages people from turning the addon on if they don't know what ocular is.
- Ocular is perfectly usable on its own website without the extension, so above applies more than it normally would.
- Jeffalo doesn't really care about the addon name
- It's the only addon that's entirely an integration, so anything we decide for my-ocular would apply to future integrations
- It's literally another type of addon unlike any other, naming it the same way just for consistency is not a valid reason
- ocular is not a unique name, unlike ocular.jeffalo.net
- Weirder/longer name makes people think twice before enabling the addon, making it less likely number 2 happens (kinda a duplicate of point 5 above)

Shluffy:
First, google "ocular". It returns this:
image
So medicine statuses? I think we should keep it as ocular.jeffalo.net because if you search that, you get this:
image
so maybe keep it as ocular.jeffalo.net

Shluffy:
8 yes, to 1 no

mxmou:
> * It discourages people from turning the addon on if they don't know what ocular is.

If they don't know what ocular is, they won't know what ocular.jeffalo.net is either.

mxmou:
> * Weirder/longer name makes people think twice before enabling the addon, making it less likely number 2 happens (kinda a duplicate of point 5 above)

That's probably true, but we should find a better way to make people think twice before enabling it, for example a badge and a notice about using an external service, with a link to its privacy policy.

Hans5958:
Finally have time to answer some points, in no particular order.

> It's the only addon that's entirely an integration, so anything we decide for my-ocular would apply to future integrations

In the name, yes, but there the profile statistics "integrates" with ScratchDB, and Scratch Messaging "integrates" with the messages on Scratch. I don't think just because it has "integration" in its name, it is going be a different kind of addon.

> It's literally another type of addon unlike any other, naming it the same way just for consistency is not a valid reason

See above.

> It discourages people from turning the addon on if they don't know what ocular is.

I think there are better ways to improve it better, such as including the description of ocular on the description instead of asking them to search about what it is.

> I've seen people ask me multiple times on my profile "why does it say Hmmm next to Argentina"

I don't think changing addon names will improve recoginition in this case, so this is unrelated.

> ocular is not a unique name, unlike ocular.jeffalo.net

Within the context of Scratch forums and the fact that we can put more on the descriptions, I don't think this matters.

> Jeffalo doesn't really care about the addon name

This is also unrelated since he won't partake on our discussion.

> Ever since post reactions were the main feature of the my-ocular addon, the addon name had the full domain on it

There are a lot of addons who have the same addon names up until now, so this is unrelated.

> Ocular is perfectly usable on its own website without the extension, so above applies more than it normally would.

I'm not sure about how this is related. May I get an explanation.

mxmou:
> * Ocular is perfectly usable on its own website without the extension, so above applies more than it normally would.

Ocular is not "perfectly usable" without an easy way to access a specific post, which search and the browse feature aren't.

WorldLanguages:
> That's probably true, but we should find a better way to make people think twice before enabling it, for example a badge and a notice about using an external service, with a link to its privacy policy.

The problem is not the external service itself. That would affect other addons that use ScratchDB.
The main point here is that users should not enable the addon without first knowing what ocular is.
The usefulness of ocular comes from its account system, not really from the data it provides. All of its data is generated by ocular users.
It's the only addon whose main purpose is to show user generated data that does not originate from Scratch.
This makes it different to any other addon currently in existence. Comparisons like the one I'll reply to below are not helpful to the discussion.

> In the name, yes, but there the profile statistics "integrates" with ScratchDB, and Scratch Messaging "integrates" with the messages on Scratch. I don't think just because it has "integration" in its name, it is going be a different kind of addon.

- It's obvious Scratch Messaging will need to "integrate" with Scratch.
- Profile statistics is not locked to ScratchDB forever. If at some point some competition arrives, the same addon could potentially allow the user to switch to another source of information. This is not the case with the my-ocular addon.
- It has always been a different kind of addon, even before it had "integration" in the name. That's why the addon name has included `ocular.jeffalo.net` for months now, per my request to Jeffalo.

> I think there are better ways to improve it better, such as including the description of ocular on the description instead of asking them to search about what it is.

Do you really think people will read the description for an addon named `ocular integration`? See my point below.

> I don't think changing addon names will improve recoginition in this case, so this is unrelated.

It will, because it's most likely users will understand what the addon is about, read the description and addon settings if the addon name is long, weird, or just different to others. `ocular integration` can go unnoticed more easily between the 90+ addons currently available.

WorldLanguages:
Honestly, if you want to make a good point, compare my-ocular to the recently added turbowarp-player. Not to addons that use ScratchDB.
Would be stupid if I didn't highlight some differences in advance tho:
- TurboWarp is not useful just because of its data, but because of its features, mainly its compiler.
- TurboWarp is a unique name, unlike ocular, and ranks first on Google when searching "TurboWarp".
- TurboWarp is more widely known in the Scratch community.
- TurboWarp does not have any account system.
- The default behavior of the addon is opening TurboWarp.org in a new tab, which is what we want (users to explore what TurboWarp is on the web first, then decide if they want to use the embedded player)

mxmou:
> * TurboWarp is more widely known in the Scratch community.

The ocular addon has a forums tag, which indicated that it's intended for forum users. Most forumers that use Scratch Addons also know what ocular is.

#2908 adds a "Forums" group, which isn't a bad idea if it's placed above Others and not below (there should be a clear difference between useful addons that only few users need and not very useful addons). Once that PR is merged and ocular integration is in that group, most people who consider enabling it will be forum users and already know ocular, so maybe the name could be changed then?

WorldLanguages:
> #2908 adds a "Forums" group, which isn't a bad idea if it's placed above Others and not below (there should be a clear difference between useful addons that only few users need and not very useful addons). Once that PR is merged and ocular integration is in that group, most people who consider enabling it will be forum users and already know ocular, so maybe the name could be changed then?

Okay, that seems like a good compromise

Hans5958:
> The usefulness of ocular comes from its account system, not really from the data it provides. All of its data is generated by ocular users.
> It's the only addon whose main purpose is to show user generated data that does not originate from Scratch.
> This makes it different to any other addon currently in existence. Comparisons like the one > I'll reply to below are not helpful to the discussion.

I don't think this kind of behavior warrants it as a unique/different addon.

> It will, because it's most likely users will understand what the addon is about, read the description and addon settings if the addon name is long, weird, or just different to others. `ocular integration` can go unnoticed more easily between the 90+ addons currently available.

In my opinion, I think what mostly applies here is that they will either turn on all addons blindly or take a time to consider for each of it. So, they either won't read anything on the description, or they will open and take time to read regardless of the name. I think it's mostly a black-and-white situation.

> TurboWarp is a unique name, unlike ocular, and ranks first on Google when searching "TurboWarp".

Again, just like SDB, this is insignificant as argued above.

> TurboWarp is not useful just because of its data, but because of its features, mainly its compiler.

> TurboWarp does not have any account system.

> The default behavior of the addon is opening TurboWarp.org in a new tab, which is what we want (users to explore what TurboWarp is on the web first, then decide if they want to use the embedded player)

So what? Again, I don't think ocular integration is a unique/different addon.

> It's obvious Scratch Messaging...

Those are examples to prove that just because is integrates in someway, does not mean that it warrants it to be some kind of a different/unique kind of addon.

apple502j:
superseded by #3261

PR #3164, opened 2021-07-30
msg-count-badge: add indicator when signed out

### LINKED ISSUE #1756: Connection indicator in popup

Since Scratch experiences frequent outages, it would be nice to be able to see at a glance if Scratch is up or not.

At the very left of the tabs, there could be a small circular indicator that's green when online and red when offline. Additionally, the extension could show a small message when Scratch is down and another message when it comes back up. This would look something like the Firefox Private Network extension:

image
image

The indicator could also appear on the extension icon -- again, in a similar manner to Firefox Private Network.
There could also be an option to tell if it's not just Scratch but the entire device/network -- this could be achieved by trying to ping scratchaddons.com.

lisa-wolfgang:
Resolves #1756.

Currently, it is impossible to tell the difference between not having any messages and not being signed in. Because Scratch sessions can expire without action on the user's behalf, a user may falsely believe that they have no messages under the notion that they're still logged in.

It's also useful to know when the Scratch website experiences downtime and/or if the network connection as a whole is down.

### Changes

A new option has been added to "Message count on extension icon". It's disabled by default.

image

When the option is enabled, a special badge appears when the user is offline, signed out, or having other connection issues to Scratch.

image

Additional options are available to customize this offline badge.

image
image

The default red for the setting has also been added as a preset for the regular badge color. Additionally, the "Old light green" preset was renamed to "Legacy" and given a tooltip description. Presets (as of now) only affect the regular badge color, not the offline badge.

image

### Reasons for changes

~~You can put a dango in the badge~~
image

### Tests

Tested in Chromium 92 and Firefox 90.
Does not impact previous behavior when the "Indicate when not signed in" option is disabled.
Being signed out, signing in, and signing back out all function as expected.

### To-do
- [ ] When #3151 is merged, use the `if` property to hide the last two addon settings when `showOffline` is `false`.

### DIFF
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -30,7 +30,7 @@
       "id": "offlineText",
       "type": "string",
       "default": "?",
-      "min": "1"
+      "min": 1
     }
   ],
   "presets": [
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -9,7 +9,7 @@
   "settings": [
     {
       "name": "Badge color for message count",
-      "id": "countColor",
+      "id": "color",
       "type": "color",
       "default": "#008000"
     },
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -8,7 +8,7 @@
   "versionAdded": "1.0.0",
   "settings": [
     {
-      "name": "Badge color for message count",
+      "name": "Badge color",
       "id": "color",
       "type": "color",
       "default": "#008000"
@@ -18,19 +18,6 @@
       "id": "showOffline",
       "type": "boolean",
       "default": false
-    },
-    {
-      "name": "Badge color when not signed in",
-      "id": "offlineColor",
-      "type": "color",
-      "default": "#e90404"
-    },
-    {
-      "name": "Badge indicator when not signed in",
-      "id": "offlineText",
-      "type": "string",
-      "default": "?",
-      "min": 1
     }
   ],
   "presets": [
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -8,32 +8,59 @@
   "versionAdded": "1.0.0",
   "settings": [
     {
-      "name": "Badge color",
-      "id": "color",
+      "name": "Badge color for message count",
+      "id": "countColor",
       "type": "color",
       "default": "#008000"
+    },
+    {
+      "name": "Indicate when not signed in",
+      "id": "showOffline",
+      "type": "boolean",
+      "default": false
+    },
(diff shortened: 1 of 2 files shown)


addons/msg-count-badge/addon.json
```diff
@@ -8,32 +8,59 @@
...
+      "default": "?",
+      "min": "1"
     }
   ],
   "presets": [
     {
-      "name": "Old light green",
+      "name": "Legacy",
       "id": "oldLightGreen",
+      "description": "The badge color from the original Scratch Messaging extension.",
```
RedGuy12, line 40: Wait this works?

lisa-wolfgang: Yup, it creates a tooltip on hover.

addons/msg-count-badge/addon.json
```diff
@@ -8,32 +8,59 @@
...
+      "id": "offlineColor",
+      "type": "color",
+      "default": "#e90404"
+    },
+    {
+      "name": "Badge indicator when not signed in",
+      "id": "offlineText",
+      "type": "string",
+      "default": "?",
+      "min": 1
```
TheColaber, line 33: you can probably remove `min: 1` entirely

TheColaber (approved):

addons/msg-count-badge/addon.json
lisa-wolfgang: I was originally thinking that leaving it blank would make the indicator disappear, which could be confusing to some users.

But then I realized that I could just manually add the space if the settings input is left blank.

But then I noticed that leaving the text field blank looks like this:

image

The `?` is the default, but it looks like it's still there, which could cause more confusion.

TheColaber: try making `min: 0` then. i thought min was 1 by default.

lisa-wolfgang: `min: 0` has the same effect.

WorldLanguages (approved):

WorldLanguages:
@lisa-wolfgang I removed 2 settings you added that I considered were excessive customization

lisa-wolfgang:
That's fine, but I don't think black is a very good default color for an offline indicator. (It's also one of the preset colors...)

WorldLanguages:
Hmm, you're right. I'll change it to a dark red (but not pure red)

Secret-chest (commented):
STOP REMOVING SETTINGS PLEASE, THE WHOLE POINT OF SA IS TO BE CUSTOMIZABLE! WHAT DO YOU WANT?

WorldLanguages:
@Secret-chest There's a customization ceiling, if you want to modify _the text the badge gets when you are logged out_ you are better modifying the source itself than making that customizable.

PR #3178, opened 2021-07-31
Fix TurboWarp Addon Error and Devtools clean up msg being wrong

TheColaber:
### Changes

- An error that occers with the turbowarp addon when seeing inside
- The clean up message in the context menu when you change the setting for devtools.

### Tests

Yep, self merging, these changes are only on master, not on stable.

### DIFF
--- a/addons/editor-devtools/DevTools.js
+++ b/addons/editor-devtools/DevTools.js
@@ -53,12 +53,12 @@ export default class DevTools {
         oldCleanUpFunc.call(this);
       }
     };
-    
+
     let orginialMsg = blockly.Msg.CLEAN_UP;
     if (this.addon.settings.get("enableCleanUpPlus")) blockly.Msg.CLEAN_UP = this.m("clean-plus");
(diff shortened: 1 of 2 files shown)


addons/editor-devtools/DevTools.js
```diff
@@ -53,9 +53,12 @@ export default class DevTools {
         oldCleanUpFunc.call(this);
       }
     };
+
+    let orginialMsg = blockly.Msg.CLEAN_UP;
     if (this.addon.settings.get("enableCleanUpPlus")) blockly.Msg.CLEAN_UP = this.m("clean-plus");
     this.addon.settings.addEventListener("change", () => {
       if (this.addon.settings.get("enableCleanUpPlus")) blockly.Msg.CLEAN_UP = this.m("clean-plus");
+      else blockly.Msg.CLEAN_UP = orginialMsg;
```
WorldLanguages, line 61: `originial`?

TheColaber: ah man

PR #3202, opened 2021-08-02
Add debugger block support to editor-theme3

### LINKED ISSUE #2787: No support for debugger blocks in Custom block colors
image

mxmou:
Resolves #2787

### Changes

Adds supports for debugger blocks to the customizable block colors addon. The setting is called "Scratch Addons blocks" because other addons might use the color in the future.


apple502j:
Note to self; do not merge this until we work on 1.19

RedGuy12:
> The setting is called "Scratch Addons blocks" because other addons might use the color in the future.

**in the future**. no other addon adds blocks rn. if another one is added that does, then we can change the name.

---

can you use #3151?

WorldLanguages (approved):
@apple502j Please confirm if this can be merged, I forgot what your reasoning was to add status:pending. Was it "if" support?

addons/editor-theme3/addon.json
```diff
@@ -104,6 +104,16 @@
       "default": "#0FBD8C",
       "allowTransparency": true
     },
+    {
+      "name": "Scratch Addons blocks",
```
RedGuy12, line 108: Please change this

WorldLanguages: To what? And why?

mxmou: To "Debugger" because that's currently the only addon that adds new blocks. I'm not sure which name is better.

WorldLanguages: There's an open PR that I'll probably modify so that it's part of a separate addon instead of debugger (comment blocks)
So I don't think this should be changed

GarboMuffin: I would personally prefer anything that doesn't mention "Scratch Addons" by name but will change it myself for TW if that doesn't happen

mxmou: "Addon blocks" is another option, but I think "Scratch Addons blocks" is clearer.

PR #3205, opened 2021-08-02
New addon: hide-delete-button

apple502j:
This adds a new addon, hide-delete-button. This is one of the addons where manifests are larger than the actual style... Implements https://discord.com/channels/806602307750985799/806609030935740477/871486900261752943


WorldLanguages:
Waiting for #3261 to be merged

WorldLanguages:
@mxmou Ugh having no minor tag/group makes addons like this weird IMO

apple502j:
@WorldLanguages This addon is very useful - if the problem is that we don't have a category, then we can make one about sprite management.

WorldLanguages:
> This addon is very useful

I'm missing something then, how is it useful? "Confirm sprite deletion" seems superior?

WorldLanguages:
Hmm does that addon ask if sure when deleting costumes or sounds?

apple502j:
Confirmations are annoying - sure, it's less annoying than losing sprites, but is annoying when a popup appears out of nowhere. Plus it only works for sprites, not other assets.

PR #3212, opened 2021-08-02
Add link to YouTube video in v1.18 update notification

WorldLanguages:


FunctionalMetatable:
I'd like to note that ctrl-enter-post and customize-avatar-border are in the same section of the video (its named as Ctrl+Enter to post but theres supposed to be two different sections

PR #3218, opened 2021-08-03
Remove labels from reset buttons

mxmou:
### Changes

#3156 increased the number of reset buttons and dropdowns, which take up quite a bit of space. The icon is enough to make their purpose obvious, so this makes them smaller by removing the labels.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -140,6 +140,9 @@
   "viewLicenses": {
     "message": "View library licenses"
   },
+  "reset": {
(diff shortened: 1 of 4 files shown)


WorldLanguages:
This is good, but let's keep the "reset" string and use it as the alt/title of the button.

WorldLanguages:
We might further tweak this for v1.19

WorldLanguages (approved):
Will merge and release in 30 min or so

PR #3219, opened 2021-08-03
ctrl-enter-post: stop erroring in console when topic is closed

FunctionalMetatable:
### Changes

Stops errors in console when ctrl-enter-post in forums are enabled and topic is closed

### Reason for changes

Errors fill up the console

### Tests

Tested locally

### DIFF
--- a/addons/ctrl-enter-post/forums.js
+++ b/addons/ctrl-enter-post/forums.js
@@ -1,4 +1,4 @@
-export default function ({ addon }) {
+export default async function ({ addon }) {
   let type = location.pathname.split("/")[2];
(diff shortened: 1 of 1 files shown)


addons/ctrl-enter-post/forums.js
```diff
@@ -1,9 +1,10 @@
-export default async function ({ addon }) {
+export default function ({ addon }) {
```
apple502j, line 1: Uhh what?

FunctionalMetatable: oops

PR #3226, opened 2021-08-03
better-img-uploads: fix button showing up in sounds tab

apple502j:
Fix issue where HD uploads button shows up in sounds tab when using non-English locale


apple502j:
Self-merging, tested myself

PR #3241, opened 2021-08-05
New addon: Disable script bumping

GarboMuffin:
Addon that disables script bumping so that https://cdn.discordapp.com/attachments/806602307750985803/872726733164535808/yLd6dge2wj.gif doesn't happen. Instead, you can just place scripts anywhere and nothing will move around

Was requested here: https://discord.com/channels/806602307750985799/806602307750985803/872726735710457896

I'm not particularly attached to this addon and don't really care if it gets added or not


WorldLanguages:
We can probably add this as a "minor" addon and we'll need to try to explain what it does properly :P

lisa-wolfgang:
"Do not automatically space overlapping scripts"

PR #3250, opened 2021-08-05
Add intent system

apple502j:
WIP. We need to add a fake scratch page that says "message cleared, close?"

### DIFF
--- a/addon-api/content-script/Account.js
+++ b/addon-api/content-script/Account.js
@@ -1,11 +1,6 @@
 import AccountCommon from "../common/Account.js";
 
 export default class Account extends AccountCommon {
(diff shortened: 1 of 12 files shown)


addon-api/content-script/Account.js
```diff
@@ -1,11 +1,6 @@
 import AccountCommon from "../common/Account.js";
 
 export default class Account extends AccountCommon {
-  constructor(addon) {
-    super();
-    this._addon = addon;
```
RedGuy12, line 6: Was this intentional?

apple502j: Yes.

PR #3262, opened 2021-08-08
Fix variable tab getting inserted after devtools find bar sometimes

GarboMuffin:
It's a very inconsistent, but sometimes this could happen:

image

fix is simple: s3devToolBar is an ID, not a class


WorldLanguages:
Good catch! Thanks

PR #3263, opened 2021-08-09
drag-drop: do not HD-upload images

apple502j:
Resolves a bug where all drag-drop uploads were HD when better-img-uploads is enabled. In case anyone wants old behavior, this is now a setting (off by default, only shows if someone is using HD upload addon).

This code should not break in most circumstances - if better-img-uploads is not available when the settings is true, then it'll just fallback to normal upload (aka existing code path). Dynamic enabling of better-img-uploads should also be fine. It might break when better-img-uploads was dynamically disabled (drag-drop would still upload as HD), but better-img-uploads isn't dynamically disableable yet, so who cares

### DIFF
--- a/addons/better-img-uploads/userscript.js
+++ b/addons/better-img-uploads/userscript.js
@@ -15,7 +15,7 @@ export default async function ({ addon, console, safeMsg: m }) {
   }" height="10", width="10">
      <input accept=".svg, .png, .bmp, .jpg, .jpeg" class="${addon.tab.scratchClass(
        "action-menu_file-input"
-     )}" multiple="" type="file">
+     )} sa-better-img-uploads-input" multiple="" type="file">
   </button>
   <div class="__react_component_tooltip place-${right ? "left" : "right"} type-dark ${addon.tab.scratchClass(
     "action-menu_tooltip"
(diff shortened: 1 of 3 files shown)


TheColaber:
But what if you want them to be?

mxmou:
> 
> 
> But what if you want them to be?

... which most people who enabled the HD upload addon probably do.

apple502j:
@TheColaber @mxmou In that case the last commit can be reverted. @WorldLanguages will you reconsider that?

https://discord.com/channels/751206349614088204/751207339545198682/874283234370945044

WorldLanguages:
> ... which most people who enabled the HD upload addon probably do.

I wouldn't be so sure. We discovered this behavior because someone sent feedback about them wanting to use the normal upload.
Also, "HD image uploads" is beta, sometimes makes images blurrier (source: this same feedback and other people), and the description says it "adds a button".
If we keep this behavior as a setting it should be off by default, even tho it's a ""breaking change""

apple502j:
it's off by default dont worry

WorldLanguages:
Well then the setting can be kept. I kinda forgot we have "if addonEnabled" now so we should keep the setting.

PR #3265, opened 2021-08-09
debugger: do not pause in project page

apple502j:


WorldLanguages:
Is redux really needed for this?

apple502j:
IMO we shouldn't have to worry about redux errors. Very rare imo, only happens if SA gets updated (which breaks other things) or if they have Redux DevTools extension installed.

PR #3277, opened 2021-08-11
Change logic of matchesIf

apple502j:
See the code below for changes:  (do not cheat and read the comments in diff!)

```js
!((injectable.if.addonEnabled?.length && (Array.isArray(injectable.if.addonEnabled) ? injectable.if.addonEnabled : [injectable.if.addonEnabled]).every((addon) => !scratchAddons.localState.addonsEnabled[addon])) || (injectable.if.settings && Object.keys(injectable.if.settings).some((settingName) => (Array.isArray(injectable.if.settings[settingName]) ? injectable.if.settings[settingName] : [injectable.if.settings[settingName]]).every((possibleValue) => settings[settingName] !== possibleValue))))
```

Should not affect current addons, but might affect future addons.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -308,7 +308,19 @@ function matchesIf(injectable, settings) {
    * )
    */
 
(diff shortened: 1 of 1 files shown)


retronbv:
> Should not affect current addons

Did you test?

PR #3284, opened 2021-08-14
Custom block arg switching

### LINKED ISSUE #557: Allow right-clicking a function parameter call (in the function's definition) to replace it with a dropdown, like variables.
I'm not sure how it would work when it's in a block of code floating outside of the function, it could possibly allow selection of parameters from all functions, but it would be an improvement over having to drag another one in.

pufferfish101007:
Resolves #557

### Changes

Adds an option to block-switching to replace custom block args. Options are off (no arg switching), only in custom block definitions, and all (every arg can be replaced with any other arg anywhere). Note that boolean and string/num args can't be switched between.

### Reason for changes

'Tis useful and easy.

### Tests

Tested on chrome.

### Additional notes

The settings probably need a change of wording.

### DIFF
--- a/addons/block-switching/addon.json
+++ b/addons/block-switching/addon.json
@@ -90,22 +90,27 @@
     {
       "name": "Custom block arguments",
       "id": "customargs",
+      "type": "boolean",
+      "default": true
+    },
+    {
+      "name": "Shown custom block arguments options",
+      "id": "customargsmode",
       "type": "select",
       "default": "defOnly",
       "potentialValues": [
         {
-          "id": "off",
-          "name": "none"
-        },
-        {
           "id": "defOnly",
-          "name": "only in custom block definitions"
+          "name": "Arguments in own custom block"
         },
         {
           "id": "all",
-          "name": "all"
+          "name": "Arguments in all custom blocks in sprite"
         }
-      ]
+      ],
+      "if": {
+        "settings": { "customargs": true }
+      }
     }
   ],
   "libraries": ["scratch-blocks"],
--- a/addons/block-switching/addon.json
+++ b/addons/block-switching/addon.json
@@ -4,6 +4,10 @@
   "credits": [
     {
       "name": "GarboMuffin"
+    },
+    {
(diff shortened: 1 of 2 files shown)


GarboMuffin:
merge https://github.com/pufferfish101007/ScratchAddons/pull/4 and I approve

WorldLanguages:
> The settings probably need a change of wording.

Can you explain what the difference between "only in custom block definitions" and "all" is? I haven't tested this and I can't seem to be able to figure out without trying it out.

pufferfish101007:
> merge [pufferfish101007#4](https://github.com/pufferfish101007/ScratchAddons/pull/4) and I approve

Merged, thanks.

> > The settings probably need a change of wording.
> 
> Can you explain what the difference between "only in custom block definitions" and "all" is? I haven't tested this and I can't seem to be able to figure out without trying it out.

I'll show you some screenshots. In each one, I've circled the block that's been right-clicked.
- "only in custom block definitions":
Screenshot_20210816-101657_Browser
Screenshot_20210816-101344_Browser
Screenshot_20210816-101249_Browser
- "all": 
Screenshot_20210816-102100_Browser
Screenshot_20210816-102026_Browser
Screenshot_20210816-102003_Browser


Basically in definition only mode, you'll only get the switch menu when the arg is in a custom block definition and belongs to that custom block. In all mode, you can switch any arg to any other arg of the same type (in the same sprite), regardless of where in the code it is.

WorldLanguages:
Oof, that's something complex to explain...

WorldLanguages:
Okay, settings name is longer, but it's easier to dismiss now because checkbox and mode are separate.

image

WorldLanguages (approved):

pufferfish101007:
> Okay, settings name is longer, but it's easier to dismiss now because checkbox and mode are separate.
> 
> image

Coolio, thanks :D

RedGuy12:
I was thinking "only trigger in custom blocks" for the setting name

Too late now lol

PR #3290, opened 2021-08-15
Create "featured" addon group, separate "forums" group, rename ocular addon

WorldLanguages:
Okay, my 3rd PR :P

We can deal on whether some addons should be enabled by default or not in a future PR, please leave that for the future.
Recommended tag has not changed - there are still many recommended addons that aren't enabled by default.

Whether an addon is recommended, featured, or none, is very opiniated. I suggest you avoid discussing addons that fall in a gray area. We'll soon have a source of truth for what addons people use because of #3283, so let's just wait until we can use that. In general, recommended addons should be those that more than 75% of users use, and featured ones should be idk, more than 20%? We'll see how it plays out.
Featured tag is not visible as a tag - it's only used for grouping. Because of this reason, forum tag (green background tag, not literal manifest tag) could be removed, sure (just noticed while writing this, after commiting)

Ocular addon was renamed per mxmou's suggestion https://github.com/ScratchAddons/ScratchAddons/pull/3155#issuecomment-890038913
Some easter egg addons are now under "other" group
Clone counter is now recommended, it was previously under "others" group
For whole list of changes see changed files, I'll just list downgrades (addons that go from recommended to featured)
- better-featured-project: I don't like recommended themes
- custom-block-shape: let's try to keep 2.0 look addons under featured
- dark-www: same as below
- editor-dark-mode: same as better-featured-project. I didn't remove scratchr2 from recommended tho, because it brings consistency.
- editor-stage-left: same as custom-block-shape
- editor-theme3: same as above
- gamepad: not everyone has a controller

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -44,6 +44,9 @@
   "recommended": {
     "message": "Recommended"
   },
+  "featured": {
+    "message": "Featured"
+  },
   "beta": {
     "message": "Beta"
   },
--- a/addons/account-settings-capitalize/addon.json
+++ b/addons/account-settings-capitalize/addon.json
@@ -7,6 +7,6 @@
       "matches": ["https://scratch.mit.edu/*"]
     }
   ],
-  "tags": ["easterEgg", "community"],
+  "tags": ["community"],
   "versionAdded": "1.9.0"
 }
--- a/addons/account-settings-capitalize/addon.json
+++ b/addons/account-settings-capitalize/addon.json
@@ -7,6 +7,6 @@
       "matches": ["https://scratch.mit.edu/*"]
     }
   ],
-  "tags": ["community"],
+  "tags": ["easterEgg", "community"],
   "versionAdded": "1.9.0"
 }
--- a/addons/better-featured-project/addon.json
+++ b/addons/better-featured-project/addon.json
@@ -30,5 +30,5 @@
     }
   ],
   "versionAdded": "1.5.0",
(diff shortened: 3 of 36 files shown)


WorldLanguages:
This is the amount of addons in each group on a fresh install (and making sure no addons fall inside "new", in other words, this is how v1.20.0 will look if it had no new addons on it)
image

WorldLanguages:
For comparison, master looks like this, using same logic (no new addons, fresh install):
image

TheColaber (approved):

WorldLanguages:
lol but apple
mxmou's review...

mxmou:
> Whether an addon is recommended, featured, or none, is very opiniated. I suggest you avoid discussing addons that fall in a gray area. We'll soon have a source of truth for what addons people use because of #3283, so let's just wait until we can use that. In general, recommended addons should be those that more than 75% of users use, and featured ones should be idk, more than 20%? We'll see how it plays out.

This doesn't make much sense. Recommended and featured addons should be addons we want people to use, not ones they already use.

WorldLanguages:
> This doesn't make much sense. Recommended and featured addons should be addons we want people to use, not ones they already use.

Well you've got a point, I made a little paradox hehe. Featured addons will naturally have more users because they are featured. So, let me rephrase.
If there's an addon under "others" that is widely used compared to other addons under that category, we should probably bump it to "featured".
And if an addon that has been featured for a while doesn't have many users, it can probably be put into "others".

PR #3291, opened 2021-08-15
Customizable variable background

Cool-Scratcher:
This is my first time coding an addon or contributing to anything, please spare me the hate
Adds an addon that removes the blue overlay when creating a new variable, and replaces it with a more modern blur.


WorldLanguages:
@mxmou What do you think lol
This addon was mainly made for fun so we could just reject the idea

WorldLanguages:
Okay, nevermind what I just said.

image

mxmou:
> Okay, nevermind what I just said.
> 
> image

Replacing it with a blur makes the problem worse, and I don't think it's more modern either.

Wouldn't an "overlay color" setting in editor dark mode be more useful.

WorldLanguages:
> Wouldn't an "overlay color" setting in editor dark mode be more useful.

I probably agree (but would need to test myself)

PR #3299, opened 2021-08-16
Add extra keys support addon

### LINKED ISSUE #3298: New addon that allows 'enter' key to be selected in 'when <> key pressed' and 'key <> pressed?' blocks.
Right now in vanilla Scratch, it is possible to detect the 'enter' key with the 'when <> key pressed' and 'key <> pressed?' blocks. The only problem is for some reason the 'enter' option is not available in the drop downs of these blocks and can only be accessed through editing the `project.json`. See [this project](https://scratch.mit.edu/projects/560948099/) for a demo of these hacked blocks in action.

 I suggest a plugin which allows the enter option to be selected in these two blocks, simply by adding it to the drop down just like the other keys. 

As far as I'm aware, this trick only works for enter and doesn't work on other keys like shift, backspace, etc.

Tacodude7729:
Resolves #3298

### Changes

Added a new addon "Enter Key Support" which allows the user to choose the normally hidden options for the enter key and many symbol keys in the 'when <> key pressed' and 'key <> pressed?'. See the #3298 for details.

### Tests

Tested manually, 'hacked' blocks created with this addon work in vanilla as expected.


CST1229:
I think it might be a better idea to append to the `options` arrays of the inputs instead of completely overriding them.

Tacodude7729:
> I think it might be a better idea to append to the `options` arrays of the inputs instead of completely overriding them.
I too thought this but was unsure how to achieve it given it's defined inline in a function (Javascript isn't my forte). It's also desirable to have the enter key option below the arrow keys. I doubt this will ever cause compatibility issues as is; feel free to show me how it's done if you think it could be improved, though. :)

Tacodude7729:
I have discovered that this trick works for many symbol keys too, I've added support for them into the plugin. Note that it works for symbols like '#' that need shift key + '3' key, but it's buggy. If you release the shift key first scratch will still report that your holding '#", so I didn't include them. I have also update the project https://scratch.mit.edu/projects/560948099/ to include all the other special keys so you can see them in action, working with or without the plugin.

GrahamSH-LLK:
You could replace this.jsonInit with a function that appends the thingies.

Tacodude7729:
Ok this seems better now, I'm appending the keys by replacing the `jsonInit` and `init` functions. I have an issue, though. Even though when you drag the blocks out of the sidebar you can see the extra options, the overwritten init function never gets called for the blocks actually in toolbar on the left, so they don't have the options.
image
Any idea how I could fix that?

addons/editor-extra-keys/userscript.js
```diff
@@ -0,0 +1,50 @@
...
+      if (this.jsonInitOriginal === undefined) this.jsonInitOriginal = this.jsonInit;
+      this.jsonInit = function (obj) {
+        appendKeys(obj.args0[0].options);
+        this.jsonInitOriginal(obj);
+      };
+      this.initOriginal();
+    },
+  };
+
+  addon.tab.traps.vm.emitWorkspaceUpdate();
```
GarboMuffin, line 49: This should make it work everywhere

```suggestion
  const workspace = Blockly.getMainWorkspace();
  if (workspace) {
    const allBlocks = [
      ...workspace.getAllBlocks(),
      ...workspace.getFlyout().getWorkspace().getAllBlocks()
    ];
    for (const block of allBlocks) {
      if (block.type !== "event_whenkeypressed" && block.type !== "sensing_keyoptions") {
        continue;
      }
      const input = block.inputList[0];
      if (!input) {
        continue;
      }
      const field = input.fieldRow[1];
      if (!field) {
        continue;
      }
      const menuGenerator = field.menuGenerator_;
      appendKeys(menuGenerator);
    }
  }

```

WorldLanguages:
I could not test the backtick key pressed block because of my keyboard layout.
Which is a good way to start talking about whether we should add symbols that some users would not be able to use. Including those sounds like a bad idea, even if Scratch supports them. I'm not an expert, but this probably includes backslash \\ brackets [] and backtick \`

WorldLanguages:
Let's initially only support enter, dot, comma and hyphen. In most keyboards and OSs these work normally.

WorldLanguages (approved):

Tacodude7729:
@WorldLanguages Not sure if this is the right place to put this now that it's merged; but could we have a setting for the plugin (disabled by default), something like 'show unstable keys', which re-enables the keys you disabled. A warning  could be attached stating that some users may not be able to press these keys, and use at your own risk. I agree that having keys some people can't hit may not be a great idea if it's unclear, but I also think that we shouldn't take away these options for people who might want to use them (like, for a text editor or something).

mxmou:
> Let's initially only support enter, dot, comma and hyphen. In most keyboards and OSs these work normally.

! ? + * / and = would also be useful and are probably on all keyboards.

WorldLanguages:
@mxmou I've been told keys that need the Shift key wouldn't work on MacOS. Slash can be inserted with numpad but some people don't have one.

addons/editor-extra-keys/userscript.js
RedGuy12: Not fixed according to #3525

PR #3302, opened 2021-08-16
Separate "featured new addons" and "other new addons" addon groups

WorldLanguages:
"New" group is now collapsed on page load, unless user comes from the update notification.
If an addon is either recommended or featured, it gets into "featured new" instead of "new".

How it looks with "new" expanded, as if user came from update notification:
image

With "new" collapsed:
image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -253,11 +253,11 @@
     "description": "Tag for addons. Singular."
   },
   "newGroup": {
-    "message": "New",
+    "message": "Other new addons",
     "description": "Used as a category name for new addons"
   },
   "featuredNew": {
-    "message": "Featured new",
+    "message": "Featured new addons",
     "description": "Used as a category name for featured new addons"
   },
   "enabled": {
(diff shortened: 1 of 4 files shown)


WorldLanguages:
lol someone suggest a better name for "featured new"

FunctionalMetatable:
Newly featured / Featured.

Also the newly merged scratchblocks addon doesn't have a versionAdded

WorldLanguages:
Whoops, my bad

lisa-wolfgang:
"New featured" or "New and featured".

Also, the "New" section should be renamed to "All new" or "All new addons".

WorldLanguages:
What about `New featured` and `All new`?

WorldLanguages:
@lisa-wolfgang Eh, but it's not all new addons. Featured new aren't inside new. We could change that, tho.

lisa-wolfgang:
Or maybe "Other new addons" or something similar to the "Others" section.

mxmou:
> 
> 
> Or maybe "Other new addons" or something similar to the "Others" section.

"Featured new addons" and "Other new addons"?

WorldLanguages:
> "Featured new addons" and "Other new addons"?

Kinda long, but it sounds weird without "addons" at the end, so ok :P

Issue #3304, opened 2021-08-16
`scratchblocks` Addon Ignores Newline Characters


Samq64:
### Description

The `scratchblocks` addon does not work with scripts more than one block tall. It acts as though the newline characters doesn't exist.

### Repoduction steps

1. Enable the `scratchblocks` addon
2. Find a script that uses more than one block.
3. Notice how new line characters are ignored and the script is all on one line.

### Expected behaviour

Newline characters should not be ignored.

### Actual behaviour

Newline characters are ignored.

### Screenshots

Enabled:
![Screenshot (19)](https://user-images.githubusercontent.com/81489795/129597389-ee705adc-4694-4d95-bbd1-adbe046e8b84.png)

Disabled:
![Screenshot (20)](https://user-images.githubusercontent.com/81489795/129597428-4dbe3c37-47e1-4fd5-8b3d-7465a5b4f5f5.png)

### Environment

- Browser: Edge 92.0.902.73
- Scratch Addons version: 1.19.0 pre-realese
- Operating system: Windows 11 running build 22000.132

### Additional Info

Can repo with all addons on and with all addons off except for `scratchblocks`

PR #3306, opened 2021-08-16
Do not load scratchblocks addon

apple502j:
scratchblocks addon will no longer be loaded, enabled or even recognized by the SA. Code will remain. If you can fix this mess, please make a PR. If nobody can before v1.19 releases, then the code will be removed as well.

We can't ship a broken addon.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -109,6 +109,5 @@
   "dark-www",
   "old-studio-layout",
   "compact-messages",
-  "customize-avatar-border",
-  "scratchblocks"
+  "customize-avatar-border"
 ]


FunctionalMetatable:
#3308 fixes this. InnerText somehow works as a fix.

CubeyTheCube:
innerText preserves newlines, textContent does not

apple502j:
#3306 is in no way related to #3304; the addon literally does not work for me. Sometimes the addon loads faster than cs.js, sometimes the addon loads but then original one loads and overwrites, sometimes the addon misses the blocks because they aren't loaded yet, etc.

FunctionalMetatable:
> #3306 is in no way related to #3304; the addon literally does not work for me. Sometimes the addon loads faster than cs.js, sometimes the addon loads but then original one loads and overwrites, sometimes the addon misses the blocks because they aren't loaded yet, etc.

oh

apple502j:
@FunctionalMetatable That doesn't fix if the addons run before cs.js loads, or if the addons run before DOM loads.

apple502j:
On Discord, apple: "and we still havent figured out how to do sb". W_L: "fine, remove the addon for now we have many already so 1.20.0 milestone anyway"

Issue #558, opened 2020-10-23
Change setting redirect link from Scratch to our domain


Hans5958:
I'm not sure why we have https://scratch.mit.edu/scratch-addons-extension/settings for linking the addon settings page, while we can do something in our domain like https://scratchaddons.com/settings.

JeffaloBob:
we don't have access to scratchaddons, and a change like that might disable the addon for everyone until they accept

WorldLanguages:
> we don't have access to scratchaddons.com, and a change like that might disable the addon for everyone until they accept

Yeah, we don't want to request any more host permissions, and we can only read the URL from tabs where we have host permissions. Websites can't open `chrome-extension://` URLs directly, they require help from an extension.

Hans5958:
I see. Wouldn't it be shorter if it's like https://scratch.mit.edu/#sa-settings, or https://scratch.mit.edu/#scratch-addons-settings (so it won't link a 404 page), or https://scratch.mit.edu/scratch-addons/settings?

WorldLanguages:
Well you see, if it links to a 404 it will likely load faster. And yeah it could be shorter, I just preferred to make it longer since we're basically using Scratch's domain for our own purposes so better safe than sorry.

Hans5958:
> if it links to a 404 it will likely load faster.

I don't think so. I got redirected way before it loads anything. It is instantaneous, so it doesn't matter where we pointing to.

WorldLanguages:
Yeah, actually, the scratchr2 page takes more to load than the homepage, since it's SSR vs react based. I think that in order to answer which one is actually faster we'd need to know more about how browsers work tho lol, I won't be looking into that.

Explosion-Scratch:
That's cool. I didn't know we had that loll

WorldLanguages:
@Hans5958 You should reconsider to close this, after what I said on that issue.

Hans5958:
I'll keep this open for future possibilities.

mxmou:
> I'll keep this open for future possibilities.

I don't think there are any possibilities for this other than adding an unnecessary permission.

Hans5958:
You are not a fortune-teller, and I'm also not.

Hans5958:
Very overdue update, but the only certain opportunity of the domain gets added is when https://scratch.org is needed to be added. Not sure if that will happen sometime in the future, or when, but it's good to know that there is one.

[...]

PR #3327, opened 2021-08-19
Account switching

### LINKED ISSUE #1074: Popup for switching accounts
Description:
I would like an addon that shows a popup in Scratch and allows you to easily switch accounts to be added to Scratch Addons.

Picture:
I have added a picture of what the popup should look like below. (Profile thumbnails should be placed next to usernames)
Screenshot 2020-12-19 205612

### LINKED ISSUE #1671: addon.messaging API

TheColaber:
I completely understand what World_Languages has said
> If someone is interested in developing it the proper way (alternate session IDs need to be stored as an HTTP only cookie, and session IDs should never go outside of the extension background script) as a standalone extension, they can license it under GPL and have us publish it.

I PRed this as an addon because I want W_L to take a look and see if it's ok. He may say it needs to be an extension... he may not. Regardless, the code needs to be reviewed.

### Resolves
Resolves #1671
Resolves #1074

Exceeds #3055
Not using window.postMessage anymore.

### Changes
- addon.messaging API
- new addon: account switching
- addon.auth.sessionId (ONLY IN BG)

Thank you copilot for writing half of my code :)
Thanks, ErrorGamer for writing the addon.messaging. I had to rewrite it since the API wasn't simple enough.


Samq64:
What about if the addon was added, but the dangerous tag was added to it?

TheColaber:
No it wouldn't bc it's not dangerous

apple502j:
At least please do not use userscripts. Use popup.

TheColaber:
@apple502j I think that would make the addon less convenient.

FunctionalMetatable:
Maybe adding a modal to switch accounts?

WorldLanguages:
> add msg counts

THIS!

WorldLanguages:
> At least please do not use userscripts. Use popup.

Not really an issue because all userscript can do is add accounts to the list.
If we really want we can force the user to explicitly accept adding an account to the list through a chrome-extension:// page to make everything completely safe.

TheColaber:
Make sure you remember that the only way that "other scripts" can add my account to the list is if they click the button. I could add an event.isTrusted if you want. The userscript doesn't even tell the bg what account it wants to add. The BG gets the current username. I think this is like very safe.

TheColaber:
(guys guys will this actually become an addon and not an extension :D)

addons/account-switching/userscript.js
```diff
@@ -0,0 +1,111 @@
...
+    await addon.tab.waitForElement(".divider");
+    const username = await addon.auth.fetchUsername();
+
+    function sendBGMsg(type, username) {
+      return async function (event) {
+        if (!event.isTrusted) return;
+        addon.messaging.sendMessage({ type, username }, () => {
+          window.location.reload();
+        });
+      };
```
FunctionalMetatable, line 34: Probably doesn't need async function for now (could the messaging api in this PR be using promises instead of callbacks?)

addons/account-switching/userscript.js
```diff
@@ -0,0 +1,111 @@
...
+          }
+          dropdown.insertBefore(
+            el(
+              "li",
+              elProps,
+              el(
+                "a",
+                {
+                  href: "javascript:void(0)",
+                  listeners: {
```
FunctionalMetatable, line 66:
```suggestion
                {
                  href: "#",
                  listeners: {
```

addons/account-switching/userscript.js
```diff
@@ -0,0 +1,111 @@
...
+        if (!didDivider) {
+          elProps.className = "divider";
+        }
+        let item = dropdown.insertBefore(el("li", elProps), dropdown.querySelector(".divider:last-child"));
+
+        if (accounts.includes(username)) {
+          item.append(
+            el("a", {
+              href: "javascript:void(0)",
+              textContent: msg("remove-account"),
```
FunctionalMetatable, line 91:
```suggestion
            el("a", {
              href: "#",
              textContent: msg("remove-account"),
```

addons/account-switching/userscript.js
```diff
@@ -0,0 +1,111 @@
...
+              listeners: {
+                click: sendBGMsg("remove-account"),
+              },
+            })
+          );
+        } else {
+          item.append(
+            el("a", {
+              href: "javascript:void(0)",
+              textContent: msg("add-account"),
```
FunctionalMetatable, line 101:
```suggestion
            el("a", {
              href: "#",
              textContent: msg("add-account"),
```

addon-api/background/MessagePasser.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Handles message passing for background scripts.
+ */
+
+export default class MessagePasser extends EventTarget {
+  constructor(addonObject) {
+    super();
+    this._addonId = addonObject.self.id;
+  }
```
FunctionalMetatable, line 9: Probably no use to extend EventTarget if we dont take advantage of their APIs
```suggestion
/**
 * Handles message passing for background scripts.
 */

export default class MessagePasser {
  constructor(addonObject) {
    this._addonId = addonObject.self.id;
  }
```

addons-l10n/en/account-switching.json
```diff
@@ -0,0 +1,4 @@
+{
+  "account-switching/add-account": "Add Account",
+  "account-switching/remove-account": "Remove Account"
```
FunctionalMetatable, line 3:
```suggestion
  "account-switching/add-account": "Add account",
  "account-switching/remove-account": "Remove account"
```

addon-api/background/MessagePasser.js
RedGuy12: consistency

addons/account-switching/userscript.js
TheColaber: Imo callbacks are better for message apis especially since responses might not come back right away. That's my opinion, if anyone thinks differently, let me know, I can add both callback and promise.

addons/account-switching/userscript.js
TheColaber: I think JavaScript:void let's the user know that some JavaScript is going to run instead of going to a local/different url. Plus why change the url when you can just run a void onclick.

addon-api/background/MessagePasser.js
TheColaber: This should actually be extending listenable so the bg can access it in scratchAddons to be consistent with userscripts.

addons/account-switching/userscript.js
RedGuy12: yeah, so use promises to get the result inline with `async`/`await`, but if one person wants to use callbacks, they can use `.then`

addons/account-switching/userscript.js
CubeyTheCube: @TheColaber https://stackoverflow.com/a/1293130

TheColaber: I don't really agree and think changing a url is worse but I think I'll just get rid of the href attribute altogether. Actually it was copilot that decided to do a JavaScript url, and I # intensionally didn't use a hash.

addons/account-switching/userscript.js
TheColaber: But there's no good reason to use awaits when messages aren't meant to get responce right away.

WorldLanguages:
> (guys guys will this actually become an addon and not an extension :D)

Source? :P

TheColaber:
> Source? :P

This not being closed/rejected yet

RedGuy12:
He said "will this" not "this will"

WorldLanguages:
> He said "will this" not "this will"

good point

ianfife:
Just tested; works great and looks great. My one complaint is that it is not exactly obvious how it works unless you sign out and add an account.  Maybe like an "add new account" button?

TheColaber:
What would said button do?

retronbv:
> What would said button do?

It should add a new account to the switcher?

TheColaber:
@retronbv right, but lets say i wanted to add griffpatch's account, I would have to be logged in to it and not just get prompted for what username i'd like to add.

WorldLanguages:
You need to log in at least once, how would it know the session ID if you don't log in?

ianfife:
> @retronbv right, but lets say i wanted to add griffpatch's account, I would have to be logged in to it and not just get prompted for what username i'd like to add.

log them out, prompt them to log in and automatically add that new account that they logged in with.

TheColaber:
I'm not sure how useful that feature would be.

ianfife:
It would basically just be for UX purposes.

WorldLanguages:
I'm marking as a draft, I can probably make this into a separate extension within the next 2 months.

ianfife:
> I'm marking as a draft, I can probably make this into a separate extension within the next 2 months.

Why does it need to be a separate extension?

WorldLanguages:
@ianfife Scratch Addons never interacts with the scratchsessionsid cookie and that's great. Moving this to a separate extension means we can keep that limitation, publish the standalone extension once instead of updating it every few weeks and keep the source small and stable so that anyone can check how session ids are stored.

WorldLanguages:
@ianfife Scratch Addons never interacts with the scratchsessionsid cookie and that's great. Moving this to a separate extension means we can keep that limitation, publish the standalone extension once instead of updating it every few weeks and keep the source small and stable so that anyone can check how session ids are stored.

ianfife:
I am considering maintaining a fork that has this included then, because I could really use this and don't want to have to use another extension wasting more RAM.

hello-smile6:
> I'm marking as a draft, I can probably make this into a separate extension within the next 2 months.

How did that timeline go?

lisa-wolfgang:
Note that the recent removal of persistent scripts has broken this implementation.

WorldLanguages:
I didn't really have the time to look deeply into this yet, but I was wondering something.

Ideally, this addon/extension should respect these two design decisions:
1. The session IDs should be stored in the same way as Scratch's so they have the same level of security. This way, we can easily tell the user that they should consider themselves to be logged in to all accounts that this addon displays as available for login. We also avoid storing plain text passwords or having to encrypt them properly. 
So, like this PR does, we just create a new HTTP-only cookie with the same properties as the original `scratchsessionsid` cookie.
2. Users of the addon/extension should be able to easily remove additional stored session IDs. This isn't a problem until the user uninstalls the extension - any session IDs that were stored in our own cookie (named `accounts` in this PR) will stay in memory forever until the user clears all the storage for `scratch.mit.edu` (or the cookie expires).

It's hard to accomplish number 2 if we store the session IDs in the scratch.mit.edu context. We could store the session IDs in `chrome.storage.local` instead, but that would mean rule number 1 is not followed. So we should mitigate the issue somehow. Some ideas:
1. Somehow cause the `accounts` cookie to be cleared when the user logs out, even if the extension is not running. Probably impossible.
2. Use short expiry dates for our own cookie (a week?) and renew it once per hour when the extension is running.
3. Store half of each session ID in the cookie and the other halves in `chrome.storage` (sounds ridiculous I know, I don't mean it literally, you get the point - reminds me of bitcoin multisig)

RedGuy12:
> 2\. This isn't a problem until the user uninstalls the extension - any session IDs that were stored in our own cookie (named `accounts` in this PR) will stay in memory forever until the user clears all the storage for `scratch.mit.edu` (or the cookie expires).

is there any way to run a script on uninstall or smth?

RedGuy12:
if we had scratchaddons.com host perms (related: #558) we could save the cookie there and remove it when users visit https://scratchaddons.com/farewell

WorldLanguages:
> we could save the cookie there

And GitHub would get our users' session ids 🤦‍♂️

lisa-wolfgang:
> is there any way to run a script on uninstall or smth?

This exists, but it seems like a bit of a shaky method: https://stackoverflow.com/a/47361379

BroJac5246:
> Scratch Addons never interacts with the scratchsessionsid cookie and that's great. Moving this to a separate extension means we can keep that limitation, publish the standalone extension once instead of updating it every few weeks and keep the source small and stable so that anyone can check how session ids are stored.

Could SA somehow run this addon in a container?
I really have no idea, but just a thought. That way the addon could be theoretically implemented without being its own extension.

Also, cybersecurity is a good thing, but what are the odds of people stealing Scratch logins? Like, why and how would that happen? If attackers are doing it via an extension, then they should just steal the cookies that the Scratch website places. If they're doing it via a website, then we just have to stop SA from interacting with that website. And if the user has a virus on their computer, then, well, getting their Scratch logins stolen should be the least of their worries. I think we should just give users a warning of what could _theoretically_ happen if they enable this.

That said, I am by no means a cybersecurity expert, so that's just my thought process.

apple502j:
There is a consensus that this should be in a separate extension for addon store review purposes. https://discord.com/channels/751206349614088204/751207339545198682/1007771284270227456

PR #3333, opened 2021-08-19
v1.19 update notification

WorldLanguages:
New YouTube video ID coming in a few days

image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -236,14 +236,14 @@
   "extensionHasUpdated": {
     "message": "Scratch Addons has updated to v$1"
   },
-  "extensionUpdateInfo1_v1_18": {
-    "message": "New addon: \"TurboWarp button\", which adds a button on project pages to play the project in TurboWarp.org. Go to $1 to enable it."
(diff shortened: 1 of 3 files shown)


FunctionalMetatable:
#3333 🎉

RedGuy12:
> The new \"name scripts **before** placing in backpack\" addon asks you to name scripts **after** you drag them into the backpack

before or after?


also "hide delete button" is very out of context. the addon name should probably be changed to "hide **sprite** delete button"

WorldLanguages:
> before or after?

After being dragged but before it's actually placed in it :P

> also "hide delete button" is very out of context. the addon name should probably be changed to "hide sprite delete button"

Not sure about that, naming addons differently in the notification is weird (translators might get confused). And it also hides delete buttons for costumes, backdrops and sounds.

RedGuy12:
> After being dragged but before it's actually placed in it :P

Should be consistent



> Not sure about that, naming addons differently in the notification is weird (translators might get confused).

I meant change the addon.json too.



> And it also hides delete buttons for costumes, backdrops and sounds.

Ah, maybe "hide editor delete buttons"

PR #3335, opened 2021-08-20
New addon: Fix forum links

FunctionalMetatable:
### Changes

Adds small addon that fixes replit repl source URLs in forums (@'s get `encodeURIComponent`\`d).

### Reason for changes

Repl.it is a good place to share code on scratch

### Tests
Tested locally


mxmou:
This will probably break as many links as it will fix.

PR #3344, opened 2021-08-21
New addon: disable-paste-offset

### LINKED ISSUE #3343: Addon that makes paste in the paint editor's paste position directly on the original item.

not really but it could help people

an addon that makes the paste go directly on the original position of copying instead of what it does in the video below.

haven't considered

### Additional context
this is the video showing what i don't like. please ignore watermark.

https://user-images.githubusercontent.com/80355301/130332182-54123071-a2f8-45f1-81c2-c3e5cbc998f8.mp4

GarboMuffin:
Resolves #3343

This is another one of those small addons that I'm not personally attached to but that at least one person seems to want for whatever reason

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -95,6 +95,7 @@
   "forum-copy-code",
   "paint-by-default",
   "hide-project-stats",
+  "disable-paste-offset",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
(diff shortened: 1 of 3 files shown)


apple502j:
Not sure if we can/should modify redux state ourselves.

GarboMuffin:
I believe this works just fine, but if it makes you uncomfortable, it could instead dispatch a `scratch-paint/clipboard/CLEAR_PASTE_OFFSET`

PR #3365, opened 2021-08-25
Support host transfer

### LINKED ISSUE #3355: Studio host transfer modal dark-www support
The ST recently added support to change the studio host, which added a new modal. it will need dark-www support.

apple502j:
Tested, everything works fine. :) Also adds promotion support to message-filters

Resolves #3355


RedGuy12:
This is now in the API. This should be merged ASAP.

addons/dark-www/experimental_scratchwww.css
```diff
@@ -62,7 +62,8 @@ body:not(.sa-body-editor) .join-flow-title,
 .download .installation,
 body:not(.sa-body-editor) .gender-radio-row,
 .wedo .banner,
-.studio-report-modal .studio-report-tile-header {
+.studio-report-modal .studio-report-tile-header,
+.studio-info-box.studio-info-box-success {
```
mxmou, line 66: I don't think green boxes should be made gray.

dark-www currently has little support for various messages and errors, so it would probably be better to fix all of them at once.

WorldLanguages: Should all dark-www changes be removed from this PR?

addons/message-filters/userscript.js
```diff
@@ -17,12 +17,18 @@ export default async function ({ addon, global, console, msg }) {
     [msg("follow")]: "mod-follow-user",
     [msg("invite")]: "mod-curator-invite",
     [msg("forum")]: "mod-forum-activity",
+    [msg("host")]: "mod-become-host",
```
WorldLanguages, line 20: Where do studio promotions fall?

addons/dark-www/experimental_scratchwww.css
WorldLanguages: and scratchr2 changes?

mxmou: Those specific to the promote modal itself can stay.

apple502j: @mxmou The color is not Scratch dark green (or "extension green"/"ui-aqua"); it's more lighter than that. IMO we should change it to a darker color.

I'll remove that line from this PR but the color definitely needs some changes.

addons/message-filters/userscript.js
apple502j: Nowhere rn

PR #3366, opened 2021-08-25
Debugger blocks switching

pufferfish101007:
Resolves https://github.com/ScratchAddons/ScratchAddons/issues/2787#issuecomment-886261866

### Changes

- Adds a setting to allow switching of debugger blocks, only works when debugger addon is enabled
- fix border on top item sometimes when noops disabled
- Display border in data_listcontents switches too
- Fully dynamic settings

### Reason for changes

They're useful.

### Tests

Tested on chromium 93.

@GarboMuffin if you could take a look at this to make sure it's all fine, that would be great.


addons/block-switching/userscript.js
```diff
@@ -3,558 +3,621 @@ import blockToDom from "./blockToDom.js";
...
-    ];
-  }
-
-  // Switching for these is implemented by Scratch. We only define them here to optionally add a border.
-  // Because we don't implement the switching ourselves, this is not controlled by the data category option.
-  blockSwitches["data_variable"] = [];
-  blockSwitches["data_listcontents"] = [];
+
+  // TODO: it isn't **really** necessary to know whether debugger is enabled or not
+  const hasDebuggerEnabled = (await addon.self.getEnabledAddons()).includes("debugger");
```
apple502j, line 14: Don't think we need this

addons/block-switching/userscript.js
```diff
@@ -3,558 +3,621 @@ import blockToDom from "./blockToDom.js";
...
+        },
+        noopSwitch,
+      ];
+    }
+
+    if (addon.settings.get("sa") && hasDebuggerEnabled) {
+      const logProc = "\u200B\u200Blog\u200B\u200B %s";
+      const warnProc = "\u200B\u200Bwarn\u200B\u200B %s";
+      const errorProc = "\u200B\u200Berror\u200B\u200B %s";
+      // TODO: this will not work in Japanese or German
```
GarboMuffin, line 566: comment is now outdated and wrong

pufferfish101007: removed, thanks

addons/block-switching/userscript.js
pufferfish101007: removed

PR #3386, opened 2021-08-27
Fix `scratchWWWNoProject` not matching explore page popular mode

mxmou:
### Changes

Makes the `scratchWWWNoProject` regex match `/explore/*/*/popular`.

### Reason for changes

To fix themes not working on those URLs. Bug reported by Scratch user VRPS.


WorldLanguages:
@mxmou Whoops, what about recent? https://scratch.mit.edu/explore/projects/all/recent is not styled.

TheColaber:
neither is https://scratch.mit.edu/explore/projects/all/trending ...

PR #3387, opened 2021-08-27
user id addon compatible with profile banner

LankyBox01:
profile banner *x* user id

### Changes

you can see yourself, i just added a "color: white" property for the sa-user-id class

### Reason for changes

cuz https://discord.com/channels/751206349614088204/751207339545198682/880806031121338468

### Tests

yes i downloaded the thing and it works


WorldLanguages:
It doesn't become white because of CSS priority. Try `!important`?

mxmou:
> It doesn't become white because of CSS priority. Try `!important`?

Better featured project is kind of a theme, so moving it to the bottom of `addons.json` might be better.

WorldLanguages:
> so moving it to the bottom of `addons.json` might be better.

Agreed, let's do that instead of !important

PR #3405, opened 2021-08-31
Dont update post count if number on scratch is higher than scratchdb's count

FunctionalMetatable:
### Changes
Stops erroring in console if scratchdb is down

### DIFF
--- a/addons/exact-count/forum.js
+++ b/addons/exact-count/forum.js
@@ -2,7 +2,7 @@ export default async function ({ addon, global, console }) {
   while (true) {
     let userbox = await addon.tab.waitForElement(".postleft dl", { markAsSeen: true });
     if (userbox.childNodes[6] && userbox.childNodes[6].nodeValue.includes("+")) {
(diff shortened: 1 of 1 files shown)


addons/exact-count/forum.js
```diff
@@ -2,10 +2,13 @@ export default async function ({ addon, global, console }) {
   while (true) {
     let userbox = await addon.tab.waitForElement(".postleft dl", { markAsSeen: true });
     if (userbox.childNodes[6] && userbox.childNodes[6].nodeValue.includes("+")) {
+      const postCountReal = Number(/(\d\d?\d?\d?)\+? posts?/.exec(userbox.childNodes[6].nodeValue)[1]);
```
TheColaber, line 5: `posts` would need to be translated...

mxmou: Most of the forums isn't translated.

FunctionalMetatable: > 
> 
> `posts` would need to be translated...

Its never translated. I even checked by switching to a different language

apple502j, line 5: `(\d+)\+? posts?` ?

FunctionalMetatable: fixed

PR #3406, opened 2021-08-31
Add "delete-others-option" addon

Joeclinton1:
Does not resolve any issue as I created the addon without an issue.

### Changes

Created new addon called "delete-others-option" addon which adds a "delete others" option to the context menu that appears when you right click a backdrop selector or costume selector.

### Reason for changes

This makes it much faster to delete all the costumes or backdrops in a sprite. For people like me who work on projects with hundreds of costumes in a sprite this saves lots of time.

### Tests

Tested on both costumes tab and backdrop tab by right clicking, selecting "delete others" and checking that all costumes/backdrops were deleted except for the one I right clicked on.


addons/delete-others-option/userscript.js
```diff
@@ -0,0 +1,76 @@
...
+                type: 'scratch-gui/restore-deletion/RESTORE_UPDATE',
+                state: { restoreFun: restoreCostumeFun, deletedItem: 'Costume' }
+            });
+        }
+       
+    }
+}
+
+function addDeleteOthersOption(contextMenu){
+    const deleteNode = queryByText(contextMenu, "delete");
```
CST1229, line 28: I don't think this would work when any language other than English is used.

Joeclinton1: That's a good point. Unfortunately delete has no id. I'll change it to place it in the 3rd position.

Joeclinton1: unfortunately it seems delete moves about a bit so putting it in the 3rd position would some time make it out of place.

Joeclinton1: I found out delete is always the third node with the class "react-contextmenu-item" so I can now get it without the text "delete"

addons/delete-others-option/userscript.js
```diff
@@ -0,0 +1,63 @@
+var addonGlobal;
```
RedGuy12, line 1: you can put all your functions into the `export default async funtion` and then they'll have access to `addon` already

Joeclinton1: There should be only one export default function.

RedGuy12: i mean
```js
import ... from "./utils.js"
export default async ({ addon, console }) => {
  function deleteOtherCostumes(contextMenu) { /* ... */ }
  function addDeleteOthersOption(contextMenu) { /* ... */ }
  document.addEventListener("contextmenu", (e) => { /* ... */ });
};
```

Joeclinton1: fixed

Joeclinton1:
@WorldLanguages I'm fixing an issue where the context menu doesn't close after clicking "delete others" this is caused by react reloading the tab and since the context menu state is unchanged it just reappears after I close it. So please don't merge just yet

Joeclinton1:
Was getting annoyed with git at the end there because it kept forcing me to merge my changes before pushing probably because that format code thing keeps changing my code after I push. 

@WorldLanguages it's finally done.

lisa-wolfgang:
No need to ping W_L in the future -- he will never merge a PR that's marked as a draft, and he usually merges most PRs on the same date before a release.

TheColaber:
Unless they are addons, we would never merge an addon right before a release because it needs some time to be translated.

RedGuy12:
> Unless they are addons, we would never merge an addon

@TheColaber lol wut

Joeclinton1:
> No need to ping W_L in the future -- he will never merge a PR that's marked as a draft, and he usually merges most PRs a couple of days before every release.
PR's should be merged as soon as possible, not days before release. The longer you wait for more likely a merge conflict can occur. Also if someone else wants to work on the same area of code, without explicitly cloning your fork/branch, they need to wait till it's merged, so merging sooner than later speeds up development.

RedGuy12:
Wl has school


Joeclinton1:
> Wl has school

We should be able to merge our own pr, after a reviewer with merge rights has verified the pr. That's how things normally work

Also why don't you and all the other under 18's have school? (reply on discord pls)

PR #3417, opened 2021-09-02
Better BBCode for inserted image

Mr-Anonymous75:
# Changes

Instead of inserting `[img]{url}[/img]`, it now inserts `[img={url}]`.

# Motivation

It's much cleaner BBCode.

# Tests

Not tested - it's a very simple change.


WorldLanguages:
I think this should be consistent with the vanilla "picture" button, which does `[img]{url}[/img]`. So we should not change this.

Mr-Anonymous75:
> I think this should be consistent with the vanilla "picture" button, which does `[img]{url}[/img]`. So we should not change this.

So maybe we should add an addon that changes the vanilla image button's output to `[img={url}]` or something?

PR #3429, opened 2021-09-04
Implement Tab.createEditorContextMenu

### LINKED ISSUE #3439: folders addon's implementation of reordering has off-by-one error
`folders` addon re-implements reordering in L1042: https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/folders/userscript.js#L1042

`newIndex` is clamped between `0` - `guiItems.length`. This is an off-by-one error - `guiItems[guiItems.length]` is always `undefined`. This will cause `itemAtNewIndex` to be null if `zeroIndexed` is true and `newIndex` is clamped to `guiItems.length`. Luckily, this will be caught in the "should never happen" assertion.

As far as I know, no vanilla code supplies out-of-bounds index into reordering; however this is different from the vanilla behavior which correctly clamps: https://github.com/LLK/scratch-vm/blob/develop/src/sprites/rendered-target.js#L629

test code: `target.reorderCostume(0, Infinity)` (works in vanilla, fails with folders addon)

apple502j:
Closes #3406 
Closes #3418 
Resolves #3439 
Also adds new addon "Move costume to top or bottom" (move-to-top-bottom) - a testing addon but this can be kept.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -31,9 +31,6 @@ export default class Tab extends Listenable {
     this.traps = new Trap(this);
     this.redux = new ReduxHandler();
     this._waitForElementSet = new WeakSet();
-
-    this._reactContextMenus = [];
-    this._reactContextMenusInited = false;
   }
   addBlock(...a) {
     blocks.init(this);
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -4,6 +4,7 @@ import Listenable from "../common/Listenable.js";
 import dataURLToBlob from "../../libraries/common/cs/data-url-to-blob.js";
 import getWorkerScript from "./worker.js";
 import * as blocks from "./blocks.js";
+import { addContextMenu } from "./contextmenu.js";
 
 const DATA_PNG = "data:image/png;base64,";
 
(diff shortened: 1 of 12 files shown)


addon-api/content-script/Trap.js
```diff
@@ -60,6 +60,18 @@ export default class Trap extends Listenable {
...
+   * Gets react internal key.
+   * @param {HTMLElement} elem - the reference
+   * @returns {string} the key
+   */
+  getInternalKey(elem) {
+    if (!this._react_internal_key) {
+      this._react_internal_key = Object.keys(elem).find((key) => key.startsWith(this.REACT_INTERNAL_PREFIX));
+    }
+    return this._react_internal_key;
+  }
```
TheColaber, line 73: The folders addon could use this api.

WorldLanguages:
TODOs:
- [ ] Single contextmenu event for all addons
- [ ] Make sure this is compatible with right clicking sprite, costume and sound folders made by "folders" addon.

apple502j:
@WorldLanguages Requesting re-review; also @GarboMuffin for folders change and compat
After that I'll push the big boring changes (removing stuff from DevTools)

Not sure if I want to mess with folders addon in this PR. Maybe as a future refactoring PR.

GarboMuffin:
Folders change looks okay

addon-api/content-script/Tab.js
```diff
@@ -30,6 +31,9 @@ export default class Tab extends Listenable {
     this.traps = new Trap(this);
     this.redux = new ReduxHandler();
     this._waitForElementSet = new WeakSet();
+
+    this._reactContextMenus = [];
+    this._reactContextMenusInited = false;
```
WorldLanguages, line 36: These 2 look unused?

addons/move-to-top-bottom/userscript.js
```diff
@@ -1,4 +1,15 @@
 export default async ({ addon, console, msg }) => {
+  const types = ["sound"];
+
+  // Costumes should not be affected if user is running DevtoolsExtensio before version 1.20.0
```
TheColaber, line 4: "DevtoolsExtensio"

TheColaber (changes requested):

mxmou:
Shouldn't most references to "context menu" in the code, including the function name, actually be called "context menu item"?

WorldLanguages:
Yeah, it is this way to be consistent with `createBlockContextMenu` which has existed for a few months

PR #3446, opened 2021-09-07
Copy JSON from settings page

Mr-Anonymous75:
## Changes

I added another button between the _Import_ and _Export Settings_ buttons called _Copy Settings to Clipboard_, that, well, does exactly what it says.
For the button's text, I added a new... Thing? To the `messages.json` file, but only for english.
For obvious reasons, I also added a function to `webpages\index.js`.

## Motivation

It's always a good idea to paste the settings JSON in a bug report issue. 

But to do that, you have to download the file, open it in some kind of text editor, copy it, and then come back and paste it - that's not very fun, so a button in the settings page to copy the contents of the file would be quite useful.

## Tests

**Tested on Windows 10 21H1 x64 build 19043**

* Microsoft Edge 94.0.992.9 Beta 64-bit
* Google Chrome 93.0.4577.63 64-bit
* Mozilla Firefox 91.0.2 64-bit

---
<sub><sup>oops I accidentally posted before completing it</sup></sub>


webpages/settings/index.js
```diff
@@ -313,6 +313,11 @@ chrome.storage.sync.get(["globalTheme"], function ({ globalTheme = false }) {
           downloadBlob("scratch-addons-settings.json", blob);
         });
       },
+      copySettings() {
+        serializeSettings().then((serialized) => {
+          navigator.clipboard.writeText(serialized);
```
apple502j, line 318: I thought clipboard cannot be used on Promise callback?

Mr-Anonymous75: I don't know - I copied that clipboard code from somewhere. I mean, it works... Honestly, I don't even know that much JS.

lisa-wolfgang:
It would actually be even more useful if Scratch Addons could paste it automatically in a new bug report issue on this repo.

Mr-Anonymous75:
> It would actually be even more useful if Scratch Addons could paste it automatically in a new bug report issue on this repo.

Well, yeah, but aren't issue templates fixed? How can they access data from the user's browser?

Mr-Anonymous75:
What's wrong with Prettier? Why does it think my code isn't formatted properly?

RedGuy12:
why not add a button to upload json settings as well? it's annoying for me when ppl post the json and I have to create a file to upload them

Mr-Anonymous75:
> why not add a button to upload json settings as well? it's annoying for me when ppl post the json and I have to create a file to upload them

Hmm... Good idea - but I probably won't be able to implement that.

webpages/settings/index.js
Mr-Anonymous75: Well, the same code is used in forum-copy-code.

WorldLanguages: There's no promise callback on forum-copy-code

WorldLanguages:
I'm unsure on whether the button itself would be useful.
Assuming it was, I'm not sure how to guarantee a gigantic string will actually be copied in time to the clipboard. `serializeSettings` is an async function that needs to read from storage and might take more time than the browser is willing to wait to allow a user click to change the clipboard.

TheColaber:
I don't think this is useful in anyway. If your really wanting to copy the json, download the file and copy the json.

Mr-Anonymous75:
> I don't think this is useful in anyway. If your really wanting to copy the json, download the file and copy the json.

That's precisely the reason it is useful - you don't _need_ to download the file and copy it's contents from a text editor if you have a button in the settings page that just puts it in your clipboard straight away.

TheColaber:
> > I don't think this is useful in anyway. If your really wanting to copy the json, download the file and copy the json.
> 
> That's precisely the reason it is useful - you don't _need_ to download the file and copy it's contents from a text editor if you have a button in the settings page that just puts it in your clipboard straight away.

You still haven't told me why having your settings on the clipboard is useful. I understand why adding a button to copy is more useful than download a file and then copying it from there.

RedGuy12:
Posting bug reports


Mr-Anonymous75:
> Java is nothing like Javascript

Well, it does have some similarities in syntax - like the way you use the `extends` keyword to inherit classes.

Mr-Anonymous75:
Does anyone know how to change the code in pull requests? I'm kinda new to Github PRs too...
If I make changes in my `Mr-Anonymous75:copypaste-json` branch, will this PR update too?

Mr-Anonymous75:
> Yup! GitHub has now linked the branch in your fork to this PR, so any new commits will also appear in this PR.

Hmm... I have a feeling it won't automatically update here - there's a button saying `Compare and pull request` in my branch, and it hasn't updated here yet.

Do I have to make a new PR?

CST1229:
> I'm positive I made changes to the right branch. I still checked anyways, but the changes were on the correct one. Looking at the _Commits_ tab of this PR, though, it seems like two changes did update a few days after it was opened - but my most recent change still hasn't. *visible confusion*

If you're using GitHub Desktop, make sure to push the changes to GitHub.

TheColaber:
@Mr-Anonymous75 my point is, let's say I put my copied json in my issue. If you wanted to test it, you would have to copy my json from the issue, create a json file, paste it in, and then load it to sa. that's harder than just using a json file on the issue. And yes, it is possible to attach non image files to gh.

Mr-Anonymous75:
Oof, I just realized I tested the wrong copy of my code and the updates I made don't actually work - they just copy `[object Promise]` to the clipboard.

RedGuy12:
Why I asked for an upload as json button as well

Mr-Anonymous75:
> Why I asked for an upload as json button as well

I'm still working on an upload button.

Samq64:
What about an option in more settings to copy the string generated by the feedback page and then adding a spot in the bug issue template to paste it?

Mbrick2:
@TheColaber Can you reopen this. #4360 Is proof that its needed

Chiroyce1:
Why not create a blob link with the JSON, and open the link in a new tab? Then the user can copy it manually.

```js
ScratchAddonsSettings = JSON.stringify({
    "a demo JavaScript object":"Foo Bar"
});

let blob = new Blob([ScratchAddonsSettings], {type: "application/json"});
let url  = URL.createObjectURL(blob);
window.open(url);
```

Firefox users would have to manually click "Raw Data" and then copy it

image

PR #3460, opened 2021-09-10
New addon: block cherry-picking

### LINKED ISSUE #2338: Holding right click will only drag that block
> Idea: holding right click instead of left click and dragging a block out of a script will only take out that block and not everything underneath it.

https://github.com/LLK/scratch-blocks/blob/develop/core/block_svg.js#L745 might be useful but who knows...

lisa-wolfgang:
Resolves #2338

### Changes

When the addon is enabled and the Ctrl key (or Cmd key on Macs) is held down, blocks that are in the middle of scripts can be dragged out without the blocks beneath it being pulled out as well. (This does not work on the tops of scripts.)

Before:

https://user-images.githubusercontent.com/43426138/132908195-220330e8-d483-4f37-be7a-ee51fbfab336.mp4

After:

https://user-images.githubusercontent.com/43426138/132907076-d7b20cf8-7b01-4375-87da-f89d53248f6a.mp4

There is also an option that enables cherry-picking on regular dragging:

image

### Reason for changes

Normally, dragging out a block from the middle of a script takes 3 gestures -- one to detach the bottom scripts, one to drag out the middle block, and one to replace the bottom scripts. This addon makes this entire process much simpler.

### Tests

Tested on Chromium 93.


lisa-wolfgang:
Also, this is my first independently-developed addon! 🥳

addons/block-cherry-picking/userscript.js
```diff
@@ -0,0 +1,41 @@
...
+  document.addEventListener("keydown", function (e) {
+    if (e.metaKey) {
+      ctrlKeyPressed = true;
+    }
+  });
+  document.addEventListener("keyup", function (e) {
+    if (!e.metaKey) {
+      ctrlKeyPressed = false;
+    }
+  });
```
GarboMuffin, line 27: any reason these can't be shortened to eg.?

```js
  document.addEventListener("keydown", function (e) {
    ctrlKeyPressed = e.ctrlKey || e.metaKey;
  });
```

addons/block-cherry-picking/userscript.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // `opt_healStack` is a built-in option in scratch-blocks that enables cherry-picking behavior.
+  // All this function does is enable that built-in option for every block.
+  BlocklyInstance.BlockSvg.prototype.unplug = function (opt_healStack) {
+    if (ctrlKeyPressed != invertDrag) {
```
GarboMuffin, line 32: Doesn't really matter here but I think we generally prefer `!==`

I think you need something like `&& !addon.self.disabled` to make dynamicDisable work

addons/block-cherry-picking/addon.json
```diff
@@ -0,0 +1,35 @@
...
+    {
+      "type": "notice",
+      "text": "On macOS, use the Cmd key instead of the Ctrl key.",
+      "id": "macContextDisabled"
+    }
+  ],
+  "credits": [
+    {
+      "name": "lisa_wolfgang",
+      "link": "https://github.com/lisa-wolfgang"
```
GarboMuffin, line 14: Your link in custom-block-shape points to your Scratch profile instead. Would probably be best to pick one and be consistent

Secret-chest:
> Chromium 93

@lisa-wolfgang Did you switch to Linux? Nice!

lisa-wolfgang:
> @lisa-wolfgang Did you switch to Linux? Nice!

Nope, just using a Chromium-based browser.

addons/block-cherry-picking/userscript.js
lisa-wolfgang: I just copied it over from the Ctrl + click to run scripts addon -- I'll definitely condense that to what you're suggesting.

addons/block-cherry-picking/userscript.js
```diff
@@ -0,0 +1,19 @@
+export default async function ({ addon, global, console }) {
+  const BlocklyInstance = await addon.tab.traps.getBlockly();
+  const originalObject = BlocklyInstance.Block.prototype.unplug;
+
+  // Necessary to detect the CTRL/CMD key
+  let ctrlKeyPressed = false;
+  document.addEventListener("keydown", function (e) {
+    ctrlKeyPressed = e.ctrlKey || e.metaKey;
+  });
```
WorldLanguages, line 9: Don't forget `keyup`.

addons/block-cherry-picking/addon.json
```diff
@@ -0,0 +1,35 @@
...
+  "credits": [
+    {
+      "name": "lisa_wolfgang",
+      "link": "https://scratch.mit.edu/users/lisa_wolfgang"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/projects/*"]
```
WorldLanguages, line 20: Use word matcher `"projects"`

WorldLanguages:
I'm not so sure `Enable without holding Ctrl (hold Ctrl for regular drag)` is a good idea. I can already see some users posting on Bugs & Glitches.

mxmou:
> I'm not so sure `Enable without holding Ctrl (hold Ctrl for regular drag)` is a good idea. I can already see some users posting on Bugs & Glitches.

I think it's useful - could a tooltip saying something similar to "hold Ctrl to move entire stack" be added to the block while it's being dragged if the setting is enabled? It would mean that pressing Ctrl after starting dragging needs to work too.

lisa-wolfgang:
> 
> 
> > I'm not so sure `Enable without holding Ctrl (hold Ctrl for regular drag)` is a good idea. I can already see some users posting on Bugs & Glitches.
> 
> I think it's useful - could a tooltip saying something similar to "hold Ctrl to move entire stack" be added to the block while it's being dragged if the setting is enabled? It would mean that pressing Ctrl after starting dragging needs to work too.

Not sure about that -- the tooltip could seem more annoying than helpful to many users, and even without that factor, being able to switch in or out of cherry-picking mid-drag would be relatively complicated, feel a little janky to perform, and possibly break other things.

lisa-wolfgang:
> I'm not so sure `Enable without holding Ctrl (hold Ctrl for regular drag)` is a good idea. I can already see some users posting on Bugs & Glitches.

@WorldLanguages Having users posting on Bugs & Glitches shouldn't be a reason for us to limit functionality. A better solution would be to implement that recently changed settings feature we were discussing.

WorldLanguages:
Okay, let's keep the setting, but add a notice that explains in more details what it does.

WorldLanguages:
We ended up with these after some discussion on Discord:
image

WorldLanguages (approved):
Merging soon

lisa-wolfgang:
Lol we really need per-addon documentation

Issue #3238, opened 2021-08-04
Some users can't scroll studio-followers modal if infinite-scroll is enabled


WorldLanguages:
Cannot reproduce myself.

> Sorry if my comment is the same as the one bellow but the followers bar on studios IS AWSOME AND I LOVE THE FORMAT!!! But 1 probem it only shows 28 of your followers but i have 900+, Im not sure why. 

https://scratch.mit.edu/users/World_Languages/#comments-153270248

There was some feedback on scratchaddons.com/feedback as well but I won't try to search for it.
Pretty sure this bug existed before 1.18.0 release

WorldLanguages:
New feedback today:

> The browse followers glitch
Today when I was creating a new studio, I wanted to invite all my followers but surprisingly i could only add that much followers that were on my screen i couldnt scroll down or up to invite more people. I would be happy if this glitch is being fixed. thank you

WorldLanguages:
If this isn't fixed by v1.19.0 maybe we should disable infinite scrolling in that part of studios

Also, I'm not sure if people reporting this have old-studio-layout on or off. That might change things.

WorldLanguages:
All feedback reports from modern Chrome (91 and 92)

FunctionalMetatable:
cant repro

maybe its because i dont have that much followers? idk

WorldLanguages:
Whoops I forgot lol

WorldLanguages:
Well, the bug itself is "fixed" as we removed the feature...
See #5419 as well

PR #3462, opened 2021-09-10
Disable infinite-scroll for studio-followers

WorldLanguages:
Related to #3238

Some users couldn't load more followers, so it's better to remove infinite-scroll support until we find a solution.

Weirdly enough, infinite scrolling works differently for the vanilla "browse projects" modal and our own "browse followers" modal? Strangely, what we do is make infinite-scroll tell browse-followers to handle the infinite scrolling by itself:
https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/studio-followers/userscript.js#L139

### DIFF
--- a/addons/infinite-scroll/buttonScroll.js
+++ b/addons/infinite-scroll/buttonScroll.js
@@ -70,11 +70,14 @@ export default async function ({ addon, global, console }) {
     commentLoader(addon, "#view", ".studio-activity .studio-grid-load-more > button", "activity");
 
   // Enable scrolling for studio-followers
+  // Disabled, see #3238
+  /*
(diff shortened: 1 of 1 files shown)


WorldLanguages:
We can probably attempt to make infinite-scroll work the same way it does on the vanilla "browse projects" modal, instead of doing this. We haven't received feedback about that being broken, so doing that would probably fix the issue. So we should not merge this for now.

WorldLanguages:
> We can probably attempt to make infinite-scroll work the same way it does on the vanilla "browse projects" modal

It's possible, but problematic, because the same yProvider needs to be used for both listing "followers" and "following".

PR #3473, opened 2021-09-12
editor-extra-keys: use quotes in description

apple502j:


apple502j:
For some reason TX registers `<` as a placeholder syntax

RedGuy12:
It's for html placeholders.

PR #3489, opened 2021-09-13
Instead of hiding settings when the `if` evaluates to `false`, just show them as disabled

RedGuy12:
Resolves no open issues

### Changes

Instead of hiding settings when the `if` evaluates to `false`, just show them as disabled

### Reason for changes

To stop other settings from jumping around. also lets the user know that the option is there, it's just not compatible with their other settings.

### Tests

tested locally

### DIFF
--- a/webpages/settings/components/addon-setting.html
+++ b/webpages/settings/components/addon-setting.html
@@ -1,11 +1,11 @@
 <template>
   <div
-    v-show="show"
     class="addon-setting"
     :class="{
     'boolean-setting': setting.type === 'boolean',
     'number-setting': setting.type === 'integer' || setting.type === 'positive_integer',
   }"
+    v-bind:class="[show ? '' : 'disabled']"
   >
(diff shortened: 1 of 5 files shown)


WorldLanguages:
I'm unsure about this. Users will rarely actually see the settings jump around. Also, we need to figure out a good UI so that users understand why the setting is not togglable.
Also, don't forget boolean settings can stay true even if the setting "if" evaluates to false. For example, user turns on "scratch addons blocks", then disables debugger. Then the user is stuck with an enabled checkbox they can't uncheck, which is kinda sketchy. I know it's still checked even if we hide it, but that's no the important part of my point.

mxmou:
I think it's better to add a separate property (`enabledIf`?) that would only be used for settings that depend on the values of other settings of the same addon. Others should never visibly jump around.

WorldLanguages:
We should probably discuss in an issue first

PR #3491, opened 2021-09-14
Category Columns Addon

TheColaber:
- [x] See inside works
- [x] Switching Tabs works
- [x] Language switching works
- [x] RTL 😬
- [x] hide-flyout (currently, the flyout can't hide
- [x] Extensions with icons bg
- [x] Move extensions icon
- [x] Categories are listed in the wrong order

Guys... We're almost there....
sa

Please suggest things below...


TheColaber:
> Um... no? I think users with that addon enabled really want their category icons. Plus, we don't want addons to make each other redundant if at all possible. If they don't like icons with this addon, they can just disable the icons addon.

It's just... The icons would be shown in that tiny little rectangle... You'd barely even see them. I mean it's obvious that having this addon enabled means that something like that wouldn't work.

mxmou:
This addon should not replace the category bubbles with rectangle, and it shouldn't change the selected category background. Both should be part of a 2.0 theme. That solves both the icons issue and the editor-theme3 one until that addon is actually made.

> editor-theme3 (it needs changes in its addon... it should probably be rewritten to use Blockly anyway...)

Different blocks define their colors in different ways, so I think overriding `jsonInit` is the only way to implement that addon using Blockly.

> Move extensions icon to right side

If possible, I think it would be better to add an additional "category" at the end called "Add extension".

CST1229:
> 
> 
> > Um... no? I think users with that addon enabled really want their category icons. Plus, we don't want addons to make each other redundant if at all possible. If they don't like icons with this addon, they can just disable the icons addon.
> 
> It's just... The icons would be shown in that tiny little rectangle... You'd barely even see them. I mean it's obvious that having this addon enabled means that something like that wouldn't work.

Maybe have the little rectangle at the left of each category be a square when that addon is enabled?
Also I'd like to have an option to have the categories be like [a 2017 concept sketch](https://en.scratch-wiki.info/w/images/Scratch_3.0_second_mockup.png) (the rectangles are circles and the background is gray when selected).

mxmou:
> Also I'd like to have an option to have the categories be like [a 2017 concept sketch](https://en.scratch-wiki.info/w/images/Scratch_3.0_second_mockup.png) (the rectangles are circles and the background is gray when selected).

That should be the default - 2.0-style categories should not be added until a Scratch 2.0 theme is added.

CST1229:
> 
> 
> > Also I'd like to have an option to have the categories be like [a 2017 concept sketch](https://en.scratch-wiki.info/w/images/Scratch_3.0_second_mockup.png) (the rectangles are circles and the background is gray when selected).
> 
> That should be the default - 2.0-style categories should not be added until a Scratch 2.0 theme is added.

I wrote that comment before seeing yours.

mxmou:
Is there a reason to set inline styles? If not, I'll change it to use a userstyle.

mxmou:
Not sure what could be done about the order, because pen is an extension in 3.0 so it can't be exactly the same as in 2.0. Would just using `columns: 2` instead of grid (so half of the items is in the first column and the other half is in the second) be good enough?

TheColaber:
@mxmou I notice this happens to the add extensions button when switching languages...
image

> Different blocks define their colors in different ways, so I think overriding jsonInit is the only way to implement that addon using Blockly.

Not really... This affects everything but the bubbles in the flyout...
```js
b = await scratchAddons.eventTargets.tab[0].traps.getBlockly();
b.Colours.motion.primary = "#000000";
b.Colours.motion.secondary = "#000000";
b.Colours.motion.tertiary = "#000000";
b.mainWorkspace.getToolbox().dispose();
b.mainWorkspace.getToolbox().workspace_ = b.mainWorkspace;
b.mainWorkspace.getToolbox().init();
vm = __scratchAddonsTraps._onceMap.vm;
vm.emitWorkspaceUpdate();
```

> Not sure what could be done about the order, because pen is an extension in 3.0 so it can't be exactly the same as in 2.0. Would just using columns: 2 instead of grid (so half of the items is in the first column and the other half is in the second) be good enough?

I'm not really worried about where the pen category it located, and that's something the 2.0 addon could do anyway. But I think that sounds should be under looks should be under motion and so on. That's more of the 2.0 feel.

CST1229:
> Not really... This affects everything but the bubbles in the flyout...
> 
> ```js
> b = await scratchAddons.eventTargets.tab[0].traps.getBlockly();
> b.Colours.motion.primary = "#000000";
> b.Colours.motion.secondary = "#000000";
> b.Colours.motion.tertiary = "#000000";
> b.mainWorkspace.getToolbox().dispose();
> b.mainWorkspace.getToolbox().workspace_ = b.mainWorkspace;
> b.mainWorkspace.getToolbox().init();
> vm = __scratchAddonsTraps._onceMap.vm;
> vm.emitWorkspaceUpdate();
> ```
> 

Maybe that could be combined with a userstyle to recolor the bubbles in the flyout?

mxmou:
> I'm not really worried about where the pen category it located, and that's something the 2.0 addon could do anyway. But I think that sounds should be under looks should be under motion and so on. That's more of the 2.0 feel.

Is this enough of the 2.0 feel?
image

WorldLanguages:
I haven't read this thread in detail.
But this should only add the category columns. The 2.0 style should be achieved by a different addons.
In other words, I'd like to see the category columns in 3.0 style like mxmou posted above.

TheColaber:
@WorldLanguages sorry mxmou already did that :P

TheColaber:
BUG: Somehow variables don't show anymore when you check them

mxmou:
RTL now looks like this:
image

mxmou:
> weird stuff is happening with variables...

Can reproduce, but only on page load. Switching to project page and back or switching languages fixes it. Probably caused by the `toolbox.init()` call.

TheColaber:
> RTL now looks like this:

Hmm... The toolbox is a different size in RTL

mxmou:
> > RTL now looks like this:
> 
> Hmm... The toolbox is a different width in RTL

Scratch sets the position of the flyout incorrectly in RTL. This addon makes the width the same (311px) in all languages.

addons/columns/style.css
```diff
@@ -20,7 +20,7 @@
 .scratchCategoryItemBubble,
 .scratchCategoryItemIcon {
   margin: 0;
-  margin-right: 0.5rem;
+  margin-inline-end: 0.5rem;
```
WorldLanguages, line 23: Editor addons typically don't use these inline-start/end etc. CSS properties, what do you think @mxmou?

RedGuy12: mxmou *did* that xd 

https://github.com/ScratchAddons/ScratchAddons/pull/3491/commits/4b880f96b6d5755459c850757dda0e96427d6187

WorldLanguages: lmao

mxmou: None of these horizontal paddings and margins override Scratch's ones, so using logical properties is fine (and makes the code shorter).

mxmou:
> weird stuff is happening with variables...

I think I fixed that now.

TheColaber:
When enabling and disabling the category icons addon... 
image

mxmou:
> When enabling and disabling the category icons addon...

I think category icons should not change the size of the bubbles. That would fix bugs like this.

WorldLanguages:
A few of us discussed on the Discord server about the order of the categories. We want to make sure the first version of this addon ships a good order of categories, since users will have muscle memory by the time we want to change it.

Is it complex to make it so "sensing" is always the first item on the second column? It might be because this addon appears to be mostly CSS and reordering with CSS is not always possible, so we might need to trap some blockly methods.

For example, it would be better if this never happened:
image
Instead, pen should go on the left column, and sensing should continue to be the first category on the right column.

Here, "debugger" should be below control.
image

mxmou:
@WorldLanguages The addon could use row first order and change the order of built-in categories to motion, sensing, looks, operators, ... but any addon that adds new categories (data category tweaks, and possibly an addon that makes pen a built-in category as suggested in a comment on this PR) would have to check if this addon is enabled and put the added category in a different position (probably at the end to avoid having to rearrange the rest of the categories). If that's a problem, a more complex version of this could change the order automatically.

Another way is to have two `.scratchCategoryMenu`s, one for built-in categories and the other for extensions, both sorted the same way as the menu currently is. The empty space after an odd number of built-in categories could be used for the add extension button. This wouldn't require changes to other addons, but the way extensions would be rearranged when a new one is added might be confusing.

The current way is the simplest though.

WorldLanguages:
@Samq64 That looks good to me.

Endermaster3690:
if i'm correct this should resolve my old issue or discussion, horizontal block categories

WorldLanguages:
> Another way is to have two `.scratchCategoryMenu`s, one for built-in categories and the other for extensions, both sorted the same way as the menu currently is. The empty space after an odd number of built-in categories could be used for the add extension button. This wouldn't require changes to other addons, but the way extensions would be rearranged when a new one is added might be confusing (unless the extension part would use row first order).

Anyone interested on coding this?

Samq64:
Will this get in for 1.22?

TheColaber:
@mxmou @WorldLanguages I have done mxmou's suggestion:
> Another way is to have two `.scratchCategoryMenu`s, one for built-in categories and the other for extensions, both sorted the same way as the menu currently is. The empty space after an odd number of built-in categories could be used for the add extension button. This wouldn't require changes to other addons, but the way extensions would be rearranged when a new one is added might be confusing (unless the extension part would use row first order).

Tell me know what you think. We could add little things like a line in between. The button to add extensions can't really be moved (right mxmou?) since it's not in the same div and not even a blocky element. We could create our own button if we wanted to...

My implementation works with RLT, but I haven't tested dynamicEnable and dynamicDisable yet.

Samq64:
@TheColaber Everything looks good except that dynamic disable hides all the extension categories that are currently loaded. All loaded extensions re-appear once a new extension that wasn't already loaded before the addon was disabled is loaded from the extension library.

TheColaber:
Currently:
- With no extensions loaded:
image
- With extensions loaded:
image
- With the list category enabled and no extensions loaded:
image
- With the list category enabled and extensions loaded:
image
compare with https://github.com/ScratchAddons/ScratchAddons/pull/3491#issuecomment-937891515

Samq64:
I think there should be a separator but other than that this looks great!

CST1229:
I think the add extension button should be between the main and extension spaces, filling in the empty space if there's one or being on its own line if there isn't one.

RedGuy12:
That could be a bad layout shift tho


Secret-chest:
> I think the add extension button should be between the main and extension spaces, filling in the empty space if there's one or being on its own line if there isn't one.

But it will cause a layout shift

TheColaber:
I think the Add Extensions button always being on the bottom, makes sense and is fine where it is. @WorldLanguages is there anything else I need to do for this addon so it can get merged?

Samq64:
I think extensions should be ordered differently, because right now they all get shifted when adding a third one.

mxmou:
> That could be a bad layout shift tho

The number of built-in categories doesn't change often (only when enabling or disabling data-category-tweaks-v2).

mxmou (approved):

lisa-wolfgang:
> 
> 
> > I think there should be a separator but other than that this looks great!
> 
> I'm not sure if a separator would actually look good because most projects only use one or two extensions. image

Even without the line, there is still the separation of empty space, and it does look a little strange not having a line there.

Samq64:
> > > I think there should be a separator but other than that this looks great!
> > 
> > 
> > I'm not sure if a separator would actually look good because most projects only use one or two extensions.

Maybe it could be optional?

mxmou:
> > > I think there should be a separator but other than that this looks great!
> > 
> > 
> > I'm not sure if a separator would actually look good because most projects only use one or two extensions.
> 
> Maybe it could bean option, but disabled by default?

It's small and doesn't change the addon's functionality, so I don't think a setting for it is necessary.

Samq64:
> > weird stuff is happening with variables...
> 
> I think I fixed that now.

I can reproduce this bug with the latest commits.

TheColaber:
@Samq64 that should have been fixed, can u describe what you see.

Samq64:
> @Samq64 that should have been fixed, can u describe what you see.

https://user-images.githubusercontent.com/81489795/140540720-7b2b61d0-e354-4ef9-aed0-647fe47749f4.mp4

TheColaber:
@Samq64 I can't reproduce what you've posted above, tho I did run into a strange bug during my testing...

For some reason, sometimes switching sprites changes the category that is selected...

PR #3497, opened 2021-09-14
Blue switches

mxmou:
Related to #257

### Changes

Makes checkboxes on the setting page look like this:
image

### Reason for changes

To make it obvious that the label is on the left side.

### Tests

Works on Chrome and Firefox.


WorldLanguages:
Well, around a year ago, I wanted these not to be the same as enabled/disabled addon switches. But I don't think I had a real reason for that.

PR #3230, opened 2021-08-03
studio-tools: actually update redux instead of reloading

FunctionalMetatable:

PR #3500, opened 2021-09-15
Revert "studio-tools: actually update redux instead of reloading"

apple502j:
Reverts #3230


apple502j:
Self-merging. This is just reverting a PR.

CST1229:
Why revert?

RedGuy12:
It caused crashes.

PR #3513, opened 2021-09-17
More reliable ctrl key detection

GarboMuffin:
block-cherry-picking and ctrl-click-run-scripts now detect the control key more reliably. Notably, devtool's paste context menu item will no longer cause issues. It will also avoid issues when alt+tabbing while holding control.


TheColaber:
On this ctrl key topic, sometimes pressing ctrl z or Ctrl shift z focuses on the devtools input after doing the command try typing stuff in the devtools box and placing blocks down and then pressing ctrl z

Issue #3140, opened 2021-07-28
Player:FullScreen


fierin:
### Description
Add the ability to Turn on a native Full-Screen Similar to the addon linked below (in sources).

### Solution
Add a second maximize button to apply native Full-Screen.

#### Author:
 Fierin Deathbomb
#### Sources:
https://chrome.google.com/webstore/detail/scratch-fullscreen/gihphjdlkigognnhkankfjcclijhadpe

TheColaber:
This is a dupe of an issue made by @pufferfish101007, tho I can't find the issue right now.

fierin:
Oh! Thank you!

mxmou:
> This is a dupe of an issue made by @pufferfish101007, tho I can't find the issue right now.

#587?

fierin:
yes. Thank you.

WorldLanguages:
Duplicate of #587

PR #3515, opened 2021-09-18
New addon: enhanced fullscreen

### LINKED ISSUE #587: Improved fullscreen mode
**Describe the new addon**

- Makes fullscreen mode have a black background.
- Makes the project player respond to window size changes in fullscreen.
- Perhaps integrated with browser fullscreen modes? If so, the flag/stop/exit fullscreen toolbar could be removed in fullscreen mode as well.

lisa-wolfgang:
Resolves #587 (and the duplicate #3140)

### Changes

When this addon is enabled, it makes a number of improvements to the project player's fullscreen mode.
- Without enabling any settings, the player's border and extra spacing are removed, and it also has the ability to resize dynamically when changing the window size after fullscreen is enabled.
- When the `Sync browser's and Scratch's fullscreen modes` option is enabled, clicking on the player's fullscreen button will make the browser enter fullscreen mode. This change is reversed when exiting fullscreen. It is also possible to toggle the player's fullscreen mode without clicking the button by using the browser's native fullscreen feature, such as when pressing the `F11` key.
  - This option has a sub-option, `Only toggle browser fullscreen in editor with stage button`, which will prevent the browser's native fullscreen feature from toggling the project player fullscreen mode when the user is in the editor. Fullscreen can still be activated with the player's fullscreen button.
- When the `Darken background` option is enabled, the background behind the project player is changed to a very dark grey, regardless of the user's theme settings in other addons. This option can simply be disabled to use theme settings from other addons for the background.
- When the `Hide toolbar` option is enabled, it hides the toolbar above the player that includes the green flag, the stop sign, the fullscreen button, and any other components added by addons, which allows for a truly filled screen in combination with other settings.

### Reason for changes

Because they're useful and cool

### Tests

Tested with a browser that uses Chromium 93.


towerofnix:
This is maybe only an adjacent suggestion, but how about a "hide cursor in fullscreen" option? It could be a nice addition for games with custom cursor effects, without outright adding new SA-exclusive behavior to code (e.g. a special "SA: hide cursor" variable, which is *another* idea but has the same problems as all projects implementing SA-exclusive features).

WorldLanguages:
A feature to hide the cursor because the project author put a custom one as a sprite would be interesting.
I was also thinking of mouse lock for first person shooter games, but I don't think Scratch projects can detect raw mouse input :P

WorldLanguages:
> * When the `Darken background` option is enabled, the background behind the project player is changed to a very dark grey, regardless of the user's theme settings in other addons. This option can simply be disabled to use theme settings from other addons for the background.

What do you think of this @mxmou?

WorldLanguages:
> * This option has a sub-option, `Only toggle browser fullscreen in editor with stage button`, which will prevent the browser's native fullscreen feature from toggling the project player fullscreen mode when the user is in the editor. Fullscreen can still be activated with the player's fullscreen button.

Is this setting even necessary? I don't know why someone would want pressing F11 in the editor to put the project in fullscreen mode. I doubt anyone would disable it understanding what it does.

addons/fullscreen/addon.json
```diff
@@ -0,0 +1,104 @@
...
+  "credits": [
+    {
+      "name": "lisa_wolfgang",
+      "link": "https://github.com/lisa-wolfgang"
+    }
+  ],
+  "info": [
+    {
+      "type": "notice",
+      "text": "If you hide the toolbar and don't have the fullscreen sync option enabled, you'll need to use the Esc key on your keyboard to exit the project player's fullscreen mode.",
```
WorldLanguages, line 13: Maybe we can shorten this to `You can always use the Esc key to exit fullscreen mode.`?

WorldLanguages: Whoops, the notice below 😂

addons/fullscreen/addon.json
```diff
@@ -0,0 +1,104 @@
...
+      "id": "hideToolbarNotice"
+    },
+    {
+      "type": "notice",
+      "text": "If you use the F11 key to enter fullscreen, the Esc key will not fully exit browser fullscreen. If you enable the editor fullscreen option and you're in the editor, when you use F11 and then click the stage button, neither the Esc key nor the stage button will exit browser fullscreen.",
```
WorldLanguages, line 18: Is there no JavaScript fix for this? This notice is so weird and hard to understand. Is it actually impossible to make this work as we'd intend?

addons/fullscreen/addon.json
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "tags": ["editor", "theme", "projectPlayer"],
```
WorldLanguages, line 102: I don't think we should consider this a theme.
`"featured"` tag would work here too.

WorldLanguages:
Do we go for `fullscreen` or `full screen`?
image

WorldLanguages:
TODO: this addon has dynamicDisable, so we need to make sure users can't get "trapped" on full screen mode if "hide controls" is enabled (haven't tested this yet)

CST1229:
How Youtube does fullscreen because W_L told me to post it:

```
F11          + normal player     + windowed site   = normal player, fullscreen site
F11          + normal player     + fullscreen site = normal player, windowed site
F11          + fullscreen player ................. = normal player, windowed site
Video button + normal player     + windowed site   = fullscreen player, fullscreen site
Video button + normal player     + fullscreen site = fullscreen player, fullscreen site
Video button + fullscreen player ................. = normal player, windowed site
```

WorldLanguages:
Also, full screen probably works differently on Firefox, so we need to test everything there too.

Samq64:
> Do we go for `fullscreen` or `full screen`?
Fullscreen.

mxmou:
> > * When the `Darken background` option is enabled, the background behind the project player is changed to a very dark grey, regardless of the user's theme settings in other addons. This option can simply be disabled to use theme settings from other addons for the background.
> 
> What do you think of this @mxmou?

I don't like this because it makes a feature of another addon redundant (and its preview inaccurate) - could the full screen background and full screen header background settings be moved from editor dark mode to this addon instead?

> Do we go for `fullscreen` or `full screen`? image

Full screen. There aren't currently any uses of "fullscreen" in the extension, but we have YouTube full screen and full screen background.

RedGuy12:
"fullscreen" sounds better to me as a native English speaker


lisa-wolfgang:
> I don't like this because it makes a feature of another addon redundant (and its preview inaccurate) - could the full screen background and full screen header background settings be moved from editor dark mode to this addon instead?

Whoops, I didn't realize that there was a dedicated setting for this already. I don't think it should be moved because then users who have already changed the setting would have to go into the addon and change it again. It also wouldn't be affected by any of the theme presets in that addon. I'll just remove the setting from this addon and include a reference to the editor theme addon.

lisa-wolfgang:
> 
> 
> TODO: this addon has dynamicDisable, so we need to make sure users can't get "trapped" on full screen mode if "hide controls" is enabled (haven't tested this yet)

`Hide toolbar` is just CSS, so disabling will automatically show the toolbar again.

addons/fullscreen/addon.json
lisa-wolfgang: This is unfortunately the case. Browser-level full screen cannot be controlled by JS, so when the user enters browser full screen with F11 and then tries to use the Esc key to exit full screen, it is impossible to also toggle browser full screen.

The best solution is probably to disable the F11 key behavior and rely solely on Esc and the full screen button. A plus with this is that the editor setting is rendered completely unnecessary because the user has complete control over their browser full screen by default. This doesn't really fix the problem entirely, as the user can still use F11 and then click the full screen button, but at least the user won't expect browser full screen to be toggled automatically for them in that case.

WorldLanguages:
> `Hide toolbar` is just CSS, so disabling will automatically show the toolbar again.

lol you're right 🤦‍♂️

lisa-wolfgang:
I am working on fixes for all of these issues, but I haven't had much time lately. I'll push my changes when I can.

lisa-wolfgang:
I think it would be nice to have a hotkey to enter full screen mode (that isn't `F11`) -- to stay consistent with YouTube, we could use the `F` key.

retronbv:
> I think it would be nice to have a hotkey to enter full screen mode that isn't `F11`. To stay consistent with YouTube, we could use the `F` key.

What about games that require the `f` key?

Samq64:
> I think it would be nice to have a hotkey to enter full screen mode that isn't `F11`. To stay consistent with YouTube, we could use the `F` key.

I think just F11 is fine.

addons/fullscreen/addon.json
lisa-wolfgang: Never mind, `e.preventDefault()` can be used, and we can then toggle document-level full screen ourselves. However, I still think that keeping `F11` out of the picture makes the addon easier to understand.

mxmou:
Another reason to use "full screen" and not "fullscreen":
image

PR #3526, opened 2021-09-18
Run autoprefixer and fix syntax error

RedGuy12:
i ran autoprefixer which removed some outdated vendor prefixes. it also found a css syntax error.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,1 @@
+chrome > 0, edge > 0, and_chr > 0, firefox > 0, opera > 0 and last 1 year
--- a/addons/clones/style.css
+++ b/addons/clones/style.css
@@ -1,13 +1,7 @@
 .clone-container-container {
   display: none;
-  -webkit-box-align: center;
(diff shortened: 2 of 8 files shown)


.browserslistrc
```diff
@@ -0,0 +1 @@
+chrome > 0, edge > 0, and_chr > 0, firefox > 0, opera > 0 and last 1 year
```
RedGuy12, line 1: this finds chrome, edge, firefox, opera, and android chrome (for kiwi) versions in the last year

apple502j:
Can you make a separate PR to fix the error?

.browserslistrc
WorldLanguages: can this be only chrome, and_chr and firefox? opera and edge are chromium anyway

WorldLanguages:
We probably don't need an autoprefixer, the only CSS property I saw we had trouble with was `fit-content` which is not supported in Firefox to this day. We have official browser support requirements now, which are very flexible.

PR #3171, opened 2021-07-30
editor-devtools: add Alt + Click as middle-click alternative

lisa-wolfgang:
Resolves #2321 (specifically https://github.com/ScratchAddons/ScratchAddons/issues/2321#issuecomment-829874507)

### Changes

Alt + Click is now capable of triggering the middle-click popup and the "jump to" feature.

The help modal has also been tweaked to accommodate this change. I don't know if I'm supposed to change the strings in Transifex or not, and I don't have access to Transifex anyway, so I just changed them in `addons-l10n/en/editor-devtools.json` for now.

### Reason for changes

Currently, for users without a mouse or a touchpad where a triple-finger-tap simulates a middle-click, the "jump to" feature is unusable.
It's available for the middle-click popup as well for simplicity code-wise and consistency.
I would have just used Ctrl + Space, but it may have caused issues for existing Ctrl + Space users using the shortcut with their cursor inadvertently positioned over a block.

### Tests

Tested in Chromium 92 and Firefox 90.
Middle-clicking and Ctrl + Space are unaffected by this change.

PR #3541, opened 2021-09-21
New addon: Alt+click drag to duplicate block

GarboMuffin:
~~Alt+click is already used by editor-devtools, so that can't be used here~~

Alt+ctrl+drag (cmd+option+drag on macOS) will duplicate only the current block, like block-cherry-picking


WorldLanguages:
Should this addon also have a "flip controls" setting, similar to "grab single block with Ctrl key"? It would require you to hold Ctrl as well as Shift to duplicate the whole stack, and would duplicate a single script by default.
What do users do more often? Duplicate a whole stack, or a single block?

WorldLanguages:
Okay, I'm talking with griffpatch and we might actually want to use Alt instead of Shift, which makes way more sense. But we'd first need to fix the hotkey madness. So I'm not really rushed to merge this.

Currently, devtools uses Alt+Click for everything middle click, that is:
- Alt+Click variable: open in search bar
- Alt+Click custom block: go to definition
- Alt+Click in workspace: type block name

The first one is probably not widely used, and the last one could be kept.
We could also consider having this new addon duplicate blocks when being dragged with mouse middle click. But we'd need to remove the "middle click to go to custom block definition" feature of devtools.

Suffering from success, folks.

GarboMuffin:
> Should this addon also have a "flip controls" setting, similar to "grab single block with Ctrl key"?

Sure

> we might actually want to use Alt instead of Shift

What's the proposed replacement for the alt+click devtools shortcuts? Currently those shortcuts are the easiest or only way use those features on some systems.

> We could also consider having this new addon duplicate blocks when being dragged with mouse middle click. But we'd need to remove the "middle click to go to custom block definition" feature of devtools.

I don't think removing devtools features is a good idea. It could be possible for this addon to use middle click without completely breaking that feature because dragging and clicking are different actions.

On that topic, if this addon can be bound to middle mouse, then surely block-cherry-picking should be able to too. Maybe these addons can get merged together into a "Better dragging" addon and have a bunch of settings for "middle mouse", "control+click", etc. so people can configure it however they want

RedGuy12:
Or a new "keybind" setting type :eyes:


lisa-wolfgang:
If we want to keep middle-click/alt-click, we could change all of the Devtools usages to only occur when the keybind is released in order to differentiate between a click and a drag.

lisa-wolfgang:
On a separate note, it would probably be a good idea to provide an API to handle keybinds and key presses.

WorldLanguages:
> If we want to keep middle-click/alt-click, we could change all of the Devtools usages to only occur when the keybind is released in order to differentiate between a click and a drag.

I like this idea. Devtools could wait until Alt+Click is released (keyup), so both addons could use the Alt hotkey at the same time.

GarboMuffin:
It will be looked into

WorldLanguages:
Thanks!
We need to keep in mind some users use DevtoolsExtension, so an older version of devtools that doesn't use keyup could in theory run at the same time as this addon. So we probably want to ship a new devtools extension version that uses keyup a few days earlier to make sure that doesn't happen.

CST1229:
> 
> 
> Thanks! We need to keep in mind some users use DevtoolsExtension, so an older version of devtools that doesn't use keyup could in theory run at the same time as this addon. So we probably want to ship a new devtools extension version that uses keyup a few days earlier to make sure that doesn't happen.

Or maybe set a global variable in keyup-using devtools versions and don't run this addon if the devtools elements are still present without the variable being set.

lisa-wolfgang:
I still don't understand why we continue to update the Devtools extension. It seems to be doing more harm than good.

WorldLanguages:
> I still don't understand why we continue to update the Devtools extension. It seems to be doing more harm than good.

I don't think whether we update it would change the user count. It has looked pretty much the same yet it still grew. If we update it, at least we can make it compatible with other addons.

WorldLanguages:
Griffpatch is unsure about having Alt+Click and Alt+drag use the same key.

I suggested and he agreed:
1. Devtools has had middle-click features for a while, and #3171 added Alt+Click as an alternative for users that don't have middle click, not too far back. It's not ideal, but we could change that to Shift+Click, which is currently unused.
2. Make this addon use Alt+drag to duplicate the whole stack, and Alt+Ctrl+drag to duplicate a single block. This is consistent with vector editors, which use Alt for duplicating.

The only real pain is number 1, and we could hint users that it changed to Shift if devtools sees them Alt+Clicking blocks without dragging them.
In the future, we'll probably discontinue the devtools extension and hope most users migrate to Scratch Addons.

The final list of all addons and hotkeys would be:
1. Devtools uses middle click and Shift+Click for "go to custom block definition" and others.
2. Ctrl+Click to run scripts only runs scripts if Ctrl is held (no changes)
3. Grab single block with Ctrl key (no changes)
4. Duplicate block with Shift key (duplicates whole stack by default, unless Ctrl is held)

WorldLanguages:
> We need to keep in mind some users use DevtoolsExtension, so an older version of devtools that doesn't use keyup could in theory run at the same time as this addon. So we probably want to ship a new devtools extension version that uses keyup a few days earlier to make sure that doesn't happen.

I'll be releasing a new version of DevtoolsExtension in the next few days, to avoid any problems, since theoretically a user could run an old DevtoolsExtension version that listens to Alt+Click at the same time as this new addon.

PR #3554, opened 2021-09-25
New Addon: Fullscreen Project

e-coders:
New Addon: Fullscreen Project
Adds an option to let we get an button and click it to open in fullscreen ( Like https://scratch.mit.edu/projects/378507713/embed/
But If we clicks from an Project that is not containing "/" Forward Slash at the end, # 404 because the URL will be https://scratch.mit.edu/projects/378507713embed. help!!
Also help that after merging, In the addons.json file there my addon at top stops working others. Please Help!!


RedGuy12:
this just adds another fullscreen button...we already have a fullscreen button

e-coders:
No!! I've added another one button that is able to be load project as Embed. See https://scratch.mit.edu/projects/ProjectID/embed

e-coders:
:`(

RedGuy12:
not useful

WorldLanguages:
@e-coders This is a good first PR tho

CST1229:
Also, you can't put comments in JSON.
Don't modify the manifest (usually) - `__MSG_extensionName__` is automatically replaced by something else via localization stuff, and the version number is usually only changed on a "bump to `[version]`" PR.
New non-theme addons should be put above the themes (above `// NEW ADDONS ABOVE THIS ↑↑`).

PR #3557, opened 2021-09-25
block-cherry-picking: allow cherry-picking the top block of a script

GarboMuffin:
Previously, dragging the top block of a script would always drag the entire script regardless of control being pressed or not. Now it's possible to drag just the top block.

The "flip controls" setting is always disabled when dragging the top block. This matches the current behavior which allows people to easily move scripts around without pressing control when that setting is enabled. They can still drag just the top block by pressing control, which is weird, but I'm not aware of a better solution. Adding another setting seems unnecessary.


lisa-wolfgang:
Now you really need to add yourself to the credits... :)

Issue #1811, opened 2021-03-11
Add more specific buttons to the audio editor maybe, also add echo effect back


Endermaster3690:
**Describe the new addon**

This would separate the Faster and Slower buttons in the audio editor. Faster and Slower would only change the speed, not pitch, and there are other buttons below those labelled something like Higher and Lower which only change the pitch, not speed. And a way to set how much you slow and speed it up. also add echo effect back.

What do you think?

lisa-wolfgang:
The nice thing with assets in Scratch is that we can do almost anything we want with their corresponding editors because the assets are stored in standard file formats, which everybody sees and hears the same regardless of whether or not they have Scratch Addons.

areeq0325:
What about a "Compress" button? It would competes the audio file while still keeping it of similar quality. It would be helpful against "Project could not save" errors.
Of course, you could also use the "Faster" button and "Slower" button, but that's slow.

Endermaster3690:
> What about a "Compress" button? It would competes the audio file while still keeping it of similar quality. It would be helpful against "Project could not save" errors.
> Of course, you could also use the "Faster" button and "Slower" button, but that's slow.

Good idea we need that!

Genius398:
Yes we need it

@WorldLanguages  see this

TheColaber:
@Genius398 Please Stop pinging W_L on everything you think is something you need.
1) He's not going to make it... someone else will
2) He is annoyed enough going through all 6969 messages in his inbox from this project.
3) We aren't you slaves. Don't tell us to do stuff. We will when we have time. W_L personally looks at every new issue/discussion/pr/etc.

Explosion-Scratch:
I mean... It's two buttons that you have to click as is, maybe we can just push those two? How much time is this actually going to save?

areeq0325:
> It's two buttons that you have to click as is, maybe we can just push those two?

Not many people actually know about it.

SheepTester:
I'm not sure how elegantly manipulating the sound editor can be, but here's some notes on how to implement this if anyone is looking at this in the future.

[...]

PR #3562, opened 2021-09-26
New addon: echo-effect

apple502j:
See #1811 (and yes, echo effect is not removed from code, yet.)


addons-l10n/en/echo-effect.json
```diff
@@ -0,0 +1,3 @@
+{
+  "echo-effect/echo": "Echo"
```
TheColaber, line 2: Did they remove the msg but not the code?

addons/echo-effect/addon.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "Echo effect button",
+  "description": "Brings back the Echo button in the sound editor.",
+  "tags": ["editor"],
```
TheColaber, line 4: dont forget to add versionAdded and maybe make this featured?

TheColaber (changes requested):

addons-l10n/en/echo-effect.json
apple502j: Hmm, seems like "yes".

WorldLanguages:
Anyone know why it was removed?

CST1229:
> Anyone know why it was removed?

IIRC it was removed because there were too many buttons in the sound editor, and it needed a bit of cleanup. The echo effect was the one that was decided to go.

mxmou:
> > Anyone know why it was removed?
> 
> IIRC it was removed because there were too many buttons in the sound editor, and it needed a bit of cleanup. The echo effect was the one that was decided to go.

[This post](https://scratch.mit.edu/discuss/post/5271499) by a Scratch Team member confirms that. Interestingly, the [update announcement](https://scratch.mit.edu/discuss/topic/365737/) doesn't mention echo at all, and [the PR](https://github.com/LLK/scratch-gui/pull/4974) doesn't include the reason either.

lisa-wolfgang:
If the developers really wanted to clean up the sound editor, removing just one button while adding many more wasn't the way to do it.

WorldLanguages:
Ok, good. I wanted to make sure it wasn't removed for some weird reason, like idk, using this effect too many times causing a ear rape sound idk :P

PR #3571, opened 2021-09-27
confirm-actions: confirm project unshare

apple502j:
Parity with the editor unshare button which already has confirmation.

Suggested by CreativeCat at https://discord.com/channels/806602307750985799/806609030935740477/891940176845217812

### DIFF
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -8,6 +8,8 @@ export default async function ({ addon, console, msg }) {
         e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-button")
       ) {
         cancelMessage = msg("share");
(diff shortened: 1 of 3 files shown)


addons/confirm-actions/userscript.js
```diff
@@ -8,6 +8,8 @@ export default async function ({ addon, console, msg }) {
         e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-button")
       ) {
         cancelMessage = msg("share");
+      } else if (addon.settings.get("projectunsharing") && e.target.closest(".media-stats a.unshare")) {
```
TheColaber, line 11: This has a different selector than the share button. Should the share button use `.media-stats a.shared`?

apple502j: There is no share button in My Stuff?

TheColaber: Ah yes, I thought this was in the editor whoops

PR #3581, opened 2021-09-29
Capitalise "Sign out"

PoIygon:
### Changes

I have changed the "Capitalize Account Settings" add-on so it also capitalizes "Sign out" button.
### Reason for changes

The changes should be made because the "Sign out" button is also not capitalized and it doesn't fit with the other buttons.
### Tests

I have tested this pull request by using the "load temporary add-on" feature in Firefox.


WorldLanguages:
@mxmou What do you think? Is there a specific reason you didn't do this too when submitting this addon initially?

addons/account-settings-capitalize/addon.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "Capitalized Account Settings",
-  "description": "Changes \"Account settings\" in the account menu to \"Account Settings\".",
+  "name": "Capitalized Account Settings and Sign Out",
```
RedGuy12, line 2:
```suggestion
  "name": "Capitalized account menu items",
```

addons/account-settings-capitalize/addon.json
```diff
@@ -1,6 +1,6 @@
...
+  "description": "Changes \"Account settings\" and \"Sign out\" in the account menu to \"Account Settings\" and \"Sign Out\".",
```
RedGuy12, line 3:
```suggestion
  "description": "Changes \"Account settings\" and \"Sign out\" in the account menu to \"Account Settings\" and \"Sign Out\", respectively.",
```

addons/account-settings-capitalize/style.css
```diff
@@ -1,4 +1,5 @@
 .account-nav .dropdown > li:nth-child(3),
-ul.user-nav > li:nth-child(3) {
+ul.user-nav > li:nth-child(3), .account-nav .dropdown > li:nth-child(3),
```
RedGuy12, line 2: i think this is supposed to be
```suggestion
ul.user-nav > li:nth-child(3), .account-nav .dropdown > li:nth-child(4),
```
?

RedGuy12 (changes requested):
maybe this addon could also capitalize `Sign in` in the navbar?

mxmou:
@WorldLanguages "Sign in" and, in the editor, "Fade in" and "Fade out" aren't capitalized either, so this would just make things less consistent.

PoIygon:
This is not active at all

WorldLanguages:
I'm unsure whether this change makes it more consistent or not.
`Account Settings` is obvious because the page title is title case:
image
But if we capitalized `Sign Out`, then why not also capitalize `Sign in`? And we have an endless chain of consistency.

PR #3583, opened 2021-09-29
New addon: `emoji-picker`

### LINKED ISSUE #966: Comment Emoji's Dropdown
Back at it again with a new idea!

**Describe the new addon.**

When your commenting and want to show an emoji, you would usually have to manually type in _ meow _, _ B) _, _ 10mil _, so on and so forth.

But this addon would add a little dropdown menu with  buttons labeled "Cat" for _ meow _, "Confetti" for _ 10mil _, ect. 

These labels would also have a the emoji display itself on the label. The labels, when pressed, would add the text, such as _ 10mil _, for you instead of manually typing it in.

**Resolves:**

Nothing

**Reasons to add it:**

Just a handy little quality of life feature.

CST1229:
Resolves #966

### Changes

Adds the "emoji picker for comments" addon, which adds a button to the comment input to add both Scratch emojis and Unicode emojis that Scratch supports into the comment input.
Works in projects, studios and profiles.

#### Projects
project
project2
#### Profiles
user
user2
#### Addon image
website-addons-image

### Tests

Tested on Firefox.


WorldLanguages:
Screenshots looks awesome!
Are you sure those are all supported Unicode emojis? If you've used my "hidden emojis" Scratch projects as reference, keep in mind I probably forgot a few.

CST1229:
> 
> 
> Screenshots looks awesome! Are you sure those are all supported Unicode emojis? If you've used my "hidden emojis" Scratch projects as reference, keep in mind I probably forgot a few.

Probably. I included a few from [this](https://en.wikipedia.org/wiki/Emoji#Unicode_blocks) section of a Wikipedia article.
(Also, I checked by posting the list (seperated by ") into a comment and checking for any "" and there weren't any.)

FluddTheWadder8:
Yes please merge! This would be very convenient for shops!

CST1229:
> 
> 
> Yes please merge! This would be very convenient for shops!

Shops are for the forums, and this doesn't touch the forums at all (yet). Unless you copy paste it.

addons/emoji-picker/userscript.js
```diff
@@ -0,0 +1,551 @@
...
+    pickerDivider.style.display = "none";
+
+    setSeeMoreText();
+
+    this.appendChild(emojiPicker);
+
+    setEmojiPickerPos();
+  };
+
+  const insertTextToTextArea = function (insertText, textBox) {
```
apple502j, line 382: text-field-edit?

RedGuy12: (Apple is referring to the library that we already ship for image-uploader)

CST1229: I didn't know.

addons/emoji-picker/userscript.js
```diff
@@ -0,0 +1,551 @@
...
+    },
+    {
+      text: "_blm_",
+      image: "/images/emoji/blm.png",
+      imager2: "//cdn.scratch.mit.edu/scratchr2/static/__9e4044de46c7852aec750b6571cceb92__/images/easter_eggs/blm.png",
+    },
+  ];
+  //Scratch only supports Unicode emojis inside the Basic Multilingual Plane.
+  //This array contains all of those emojis.
+  //"br" is automatically converted to a line break.
```
RedGuy12, line 183:
```suggest
  //"br" is automatically converted to a section break.
```

addons/emoji-picker/userscript.js
```diff
@@ -0,0 +1,551 @@
...
+    "◽️",
+    "◾️",
+  ];
+
+  const emojiPickerOffset = addon.tab.clientVersion === "scratchr2" ? 30 : 52;
+
+  //Functions
+
+  const setEmojiPickerPos = function () {
+    emojiPicker.style.top = emojiPickerOffset + "px";
```
RedGuy12, line 360: Put this below in an else statement?

addons/emoji-picker/userscript.js
```diff
@@ -0,0 +1,551 @@
...
+
+  document.addEventListener("mouseup", function (event) {
+    if (!emojiPicker.contains(event.target)) {
+      emojiPicker.remove();
+    }
+  });
+
+  const addEmoji = function () {
+    try {
+      var textBox;
```
RedGuy12, line 415: Use ternary statement? Also then you can use const

addons/emoji-picker/userscript.js
```diff
@@ -0,0 +1,551 @@
...
+      //emojis added from an empty comment using this button
+      //as an empty comment
+      textBox.dispatchEvent(new Event("input", { bubbles: true }));
+    } catch (error) {}
+  };
+
+  //Addon
+
+  //Create the emoji picker
+  var emojiPicker = document.createElement("div");
```
RedGuy12, line 438: Why can't you use const?

addons/emoji-picker/userscript.js
```diff
@@ -0,0 +1,551 @@
...
+      item.textContent = emoji;
+      item.classList.add("sa-emoji-picker-item-inner");
+      container.appendChild(item);
+      unicodeContainer.appendChild(container);
+    }
+  });
+  unicodeContainer.style.display = "none";
+  emojiPicker.appendChild(unicodeContainer);
+
+  var seeMoreButton = document.createElement("button");
```
RedGuy12, line 487: Why not comst

RedGuy12 (changes requested):

addons/emoji-picker/userscript.js
WorldLanguages: comst lmao

addons/emoji-picker/userscript.js
CST1229: I thought you couldn't modify properties of const variables.

RedGuy12: You can.

addons/emoji-picker/userscript.js
CST1229: Actually, my function inserts spaces if there aren't already next to the added text. (I don't know how to do that with the library insert function)

addons/emoji-picker/userscript.js
CST1229: What do you mean?

addons/emoji-picker/userscript.js
```diff
@@ -0,0 +1,541 @@
...
+    "⬜️",
+    "⚪️",
+    "⚫️",
+    "▪️",
+    "▫️",
+    "◻️",
+    "◼️",
+    "◽️",
+    "◾️",
+  ];
```
WorldLanguages, line 353: You might want to split all this data into a separate js file.

WorldLanguages:
@mxmou Do you have any CSS tricks we could use so that the emoji picker button can stay aligned with the end of the textarea on scratchr2 comments? I tried display:flex on container with margin-left:auto and it almost works:
image

WorldLanguages:
@mxmou We also need to know how to properly add dark mode support to the new elements added by this addon. Should we do it on addons/dark-www/experimental_scratchr2_comments.css?

addons/emoji-picker/userscript.js
CST1229: Done

mxmou:
> Do you have any CSS tricks we could use so that the emoji picker button can stay aligned with the end of the textarea on scratchr2 comments? I tried display:flex on container with margin-left:auto and it almost works

That works, but the button needs to be moved after the `<span class="notification">` and a bit of extra CSS is necessary to match textarea width in replies:
```css
#comments .comment form > .control-group:last-child {
  width: 412px;
}
```
This would need to be overridden in `addons/scratchr2/scratchr2_comments.css` so that the width is `auto` when 2.0 → 3.0 is enabled.

Setting `display: flex` on the container removes the space between post and cancel buttons, so that needs to be re-added using CSS.

Dark mode CSS should be added to `experimental_scratchwww.css` and `experimental_scratchr2_comments.css`.

PR #3585, opened 2021-09-30
Updates to scratch messaging ui

RedGuy12:
Resolves no open issues

### Changes

- linked the "please sign in" message. it's annoying when I need to sign in but it takes more than a single click to get there
- always show the "reload" button & updated css accordingly

### Reason for changes

bc

### Tests

yes


WorldLanguages:
Nope (and it might be worse in other languages)
image

mxmou:
The popout icon shouldn't become orange on hover.

I think it would make more sense for the buttons to be under each other and not side by side.

mxmou:
This is definitely not intentional (apparently caused by the merge):
Two reload buttons

TheColaber (approved):

mxmou (approved):

PR #3603, opened 2021-10-05
animated-thumb: add modal

### LINKED ISSUE #2334: Enhance thumbnail setter
1. This should be a "community" addon, not an "editor" addon - editor addons are about the project itself, and thumbnails are only a concept as part of the community site. Adding the "set thumbnail" button to the editor makes some sense, but not really. It's probably better to add a "set thumbnail" button next to "add to studio" and "copy link", or somewhere else in the project page. Not sure how we should transition so that users find the button after the change (specially difficult because this is enabled by default).
2. Should have better UX. Show the user the updated thumbnail. Clarify that if someone looks wrong, but the request 200ed, it's Scratch's fault, not ours (Scratch's thumbnail server sometimes works weirdly). Clarify within the UI that resaving the project will change the thumbnail (unless "DON'T SET THUMBNAILS AUTOMATICALLY" setting is on)

apple502j:
Resolves #2334

https://user-images.githubusercontent.com/33279053/136035391-02c14822-5740-44c1-9ecb-7b32ef039a40.mp4

### DIFF
--- a/addons-l10n/en/animated-thumb.json
+++ b/addons-l10n/en/animated-thumb.json
@@ -7,9 +7,10 @@
   "animated-thumb/use-stage": "Use Stage",
   "animated-thumb/select-file": "Select File",
   "animated-thumb/close": "Close",
-  "animated-thumb/successful": "Thumbnail was uploaded. Note that it takes up to one day before users can see the new thumbnail.",
+  "animated-thumb/successful": "Thumbnail was uploaded.",
   "animated-thumb/server-error": "Thumbnail could not be uploaded due to server error. Please try again later.",
   "animated-thumb/error": "Thumbnail could not be uploaded.",
   "animated-thumb/too-big": "The file is too big.",
-  "animated-thumb/gif": "Projects with animated thumbnails might not appear on the front page or in the search result. Upload anyway?"
+  "animated-thumb/gif": "Projects with animated thumbnails might not appear on the front page or in the search result. Upload anyway?",
+  "animated-thumb/if-unsuccessful": "If this is not the thumbnail you uploaded just now, the server might be having a problem. Please wait a few days and try again."
 }
--- a/addons-l10n/en/animated-thumb.json
+++ b/addons-l10n/en/animated-thumb.json
@@ -2,7 +2,7 @@
   "animated-thumb/set-thumbnail": "Set Thumbnail",
   "animated-thumb/added-by": "Button added by Scratch Addons browser extension",
   "animated-thumb/description": "Upload a thumbnail from a file, or click \"Use Stage\" to use the current stage as the thumbnail.",
-  "animated-thumb/keep-thumb": "Stop Thumbnail Auto-saving (Recommended)",
+  "animated-thumb/keep-thumb": "Stop thumbnail auto-saving (Recommended)",
   "animated-thumb/keep-thumb-desc": "This stops the editor from overwriting the uploaded thumbnail with the stage's contents when saving the project.",
   "animated-thumb/use-stage": "Use Stage",
   "animated-thumb/select-file": "Select File",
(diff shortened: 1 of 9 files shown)


addons-l10n/en/animated-thumb.json
```diff
@@ -1,4 +1,16 @@
...
+  "animated-thumb/keep-thumb": "Stop Thumbnail Auto-saving (Recommended)",
+  "animated-thumb/keep-thumb-desc": "This stops the editor from overwriting the uploaded thumbnail with the stage's contents when saving the project.",
+  "animated-thumb/use-stage": "Use Stage",
+  "animated-thumb/select-file": "Select File",
+  "animated-thumb/close": "Close",
+  "animated-thumb/successful": "Thumbnail was uploaded. Note that it takes up to one day before users can see the new thumbnail.",
+  "animated-thumb/server-error": "Thumbnail could not be uploaded due to server error. Please try again later.",
+  "animated-thumb/error": "Thumbnail could not be uploaded.",
+  "animated-thumb/too-big": "The file is too big.",
+  "animated-thumb/gif": "Projects with animated thumbnails will not appear on the front page. Upload anyway?",
```
RedGuy12, line 14: or explore, or search

Samq64: How about "Projects with animated thumbnails may not appear on the front page, the explore page, or in search results."?

mxmou:
> Resolves #2334

It doesn't move the button to the project page, so it only fixes half of the issue.

apple502j:
Moving the position is not in this PR because of possible user confusion, since this addon is probably used by almost everyone with SA (because of its enabled-by-default state)

mxmou:
@TheColaber That much customization just causes confusion.

WorldLanguages:
About the position of the button: we should probably move it to scratch-www instead of adding it to scratch-gui, if it's technically possible. The stage is still visible on project player mode, so that's not an issue. Modal probably won't be able to use scratch-gui styles then (or will it?)

Secret-chest:
> @TheColaber That much customization just causes confusion.

@mxmou It's just a toggle. Not everyone can fork this.

apple502j:
New flow: it now has a second stage where the user can see the uploaded thumbnail.

https://user-images.githubusercontent.com/33279053/139537600-be3eb682-67a0-4629-9922-d44d30cd1650.mp4

WorldLanguages:
We should probably add a little tooltip over the button that says something like `Scratch Addons extension now allows you to set the project thumbnail to any image by clicking this button. [OK]` and use localStorage to only show it once. This works both for old users of this addon that expect the button inside the editor, as well as users that did not enable animated-thumb manually (this is an enabled by default addon) and might think the button was added by vanilla Scratch.

WorldLanguages:
@mxmou Do you think you could help make a tooltip that looks good and supports dark mode? Scratch doesn't have its own tooltips (AFAIK) so we need to create our own, right?

Samq64:
Why not just add a browser tooltip that displays `Button added by Scratch Addon browser extension.` like studio tools for now.

WorldLanguages:
@Samq64 It would not be surprising if the Scratch Team added a "leave studio" button, but it would be if they added a "set thumbnail" one.

mxmou:
I think the connection between the "Stop Thumbnail Auto-saving" checkbox (which probably shouldn't be title case - Scratch's "For all sprites" isn't) and the explanation below should be more obvious. Making both left aligned and moving them under the Select File and Use Stage buttons might help:
image

apple502j:
@mxmou Should everything be left aligned?
image

mxmou:
> @mxmou Should everything be left aligned?

Everything except the thumbnail.

WorldLanguages:
> Could we use the classes from [this component](https://github.com/seotts/scratch-www/blob/develop/src/components/forms/validation-message.jsx)?

Yeah, I think I remember how these looked and they seemed fine, let me check

apple502j:
Positioning is a hell.
image

apple502j:
image
Yes

PR #3622, opened 2021-10-09
Fix editor-dark-mode throwing error during dynamic settings change

apple502j:
Tested and works.


WorldLanguages:
lmao it was a single commit and i didnt check so it merged with the name of the commit which is incorrect instead 😂 
@RedGuy12 what on earth is a "dynamic settings change" anyway :P

RedGuy12:
i was too tired to think about grammar

I was referring to the fact that it only is throwing with the dynamic settings update functionality and it wasn't a bug with the settings page like my initial thought was

PR #3632, opened 2021-10-10
Stop using window.console in addons

RedGuy12:
Found a few refrences to window.console in various addons. This pr makes them use the console passed them.


apple502j:
`window.console` can be used outside project pages.

wgyt:
> `window.console` can be used outside project pages. Sentry-polluted console.log is only used on project pages, not on r2 or other www pages.

Sentry pollution is one of the reasons to not use real console, another is to use better console.

TheColaber:
This still appears to be untested.

apple502j:
Closing as stale; will open another PR later.

PR #3644, opened 2021-10-12
add comment blocks

### LINKED ISSUE #3614: Comment blocks
Moved from #3134

It would be great to have comment blocks as an independent addon. However, currently only `debugger` adds its own blocks to the editor, so that's the name of the addon category. We should change that.
We should also discuss the name for the block. Should it be `//`, `comment` or `comment:`?

image

MrEgggga:
Resolves #3614 

### Changes

This addon adds a block to the Debugger category called "comment" which takes one argument (a string) and does nothing, imitating a comment.

### Reason for changes

Some people don't like vanilla comments and have to create an empty custom block per sprite for this.

### Tests

I have tested this addon by loading the extension on Firefox and checking that:
- the block exists
- it does nothing
- The addon listing looks fine
I didn't make any other tests so there might be problems.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -105,6 +105,7 @@
   "block-duplicate",
   "disable-paste-offset",
   "emoji-picker",
+  "comment-blocks",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/comment-blocks/addon.json
+++ b/addons/comment-blocks/addon.json
@@ -1,0 +1,18 @@
+{
+  "name": "Comment blocks",
+  "description": "Adds comment blocks as an alternative to vanilla comments.",
+  "credits": [
+    {
+      "name": "MrEgggga"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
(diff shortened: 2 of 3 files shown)


RedGuy12:
This has been rejected iirc because these blocks actually affect project performance (so do debugger blocks, but that's why you're supposed to remove them before share). Plus, our addons already fix normal comments and go farther.

TheColaber:
@RedGuy12 there is a performance drop? Why?

RedGuy12:
https://github.com/ScratchAddons/ScratchAddons/issues/3614#issuecomment-939296305

TheColaber:
Apple is not a reliable source for that :P. A project that runs a million empty custom blocks should run instantly.

Samq64:
> Apple is not a reliable source for that :P. A project that runs a million empty custom blocks should run instantly.

image

image

MrEgggga:
> > A project that runs a million empty custom blocks should run instantly.
> 
> image
> 
> image
> 
> image
> 
> image

they might have meant debugger blocks, i'll try a performance check.

MrEgggga:
update: this does have a performance impact, closing

WorldLanguages:
Oh wow... Comment blocks might not be the best of ideas after all.

Issue #2396, opened 2021-05-05
FPS and Volume slider on project player


TheColaber:
Right now, to change FPS, you have to do so through the settings page. I have gotten requests to create a slider (similar to scratch's variable sliders... input[type=range]). 

Also, you can't change the volume for the whole project... only set it to 100% or 0%.

I propose to create new settings on these addons, that add sliders somehow. Maybe hide them later because they kinda might take up space 🤷. However, keep the hotkeys since... they have already been features for a while and people might get confused if they get removed.

PR #3656, opened 2021-10-14
New addon: FPS counter

Tacodiva:
Idea from here: https://discord.com/channels/806602307750985799/806609030935740477/897971782076739635

This addon adds an FPS counter next to the green flag of a project player when enabled. Works fine with the 60fps plugin.

image

Tested in Chromium on Ubuntu


GrahamSH-LLK:
the `=` is inconsistent.
we should use `:`

Tacodude7729:
I don't know if that's super clear either. 'Lag', to me at least, seems like a bad you should always reduce, but a lot of CPU heavy projects run much faster if Scratch isn't yielding every 0.3 seconds to render the screen. You would also see things like turbo mode increasing 'lag', which doesn't really make sense. I think most people understand what FPS is, and people who don't probably don't have much use for the plugin anyways (if your gonna reduce it to just three categories it's kinda pointless because at that point people can just look at the project and see if it's laggy or not).
I think you're right about the description probably needing a bit more elaboration, though.

lisa-wolfgang:
@Tacodude7729 What will the fate of this addon be now that we have the epic and amazing debugger tabs?

Tacodude7729:
@lisa-wolfgang I still think there's an argument to be made for this plugin. Seeing the FPS on the graph with the debugger isn't as simple as this, and it doesn't work when you're not in the editor like this does (it's also less performance heavy than the debugger). I would still vote that it should go in as an option for people (at least I would have it enabled), but I guess it's not really up to me :stuck_out_tongue:

WorldLanguages:
The FPS counter should hide itself in small stage mode.

Also, if I switch to another tab, then come back, I see the FPS counter go from ~10 FPS to 30 FPS pretty slowly. Since the drop of framerate is caused by changing tabs, could something be done about that?

Is the green color the same one as the green flag?

Should the FPS counter be visible in project player mode at all? Should it be togglable and the default be disabled?

lisa-wolfgang:
I'm pretty sure that the readout is the debugger-green color.

I think there's value in being able to see the FPS of a project from the project page, since performance is often better there. However, adding yet another button (plus this readout) seems like a lot of clutter for one feature, especially since we're already running out of room. Perhaps there could be an option to only see the FPS on your own projects?

Tacodude7729:
Could I get some advice on what to do with this? I would prefer this PR gets closed or we work out what needs to be done for it to be merged, as I for some reason I made the changes on the master branch of my fork and it's annoying me that's it's still here lol.

BroJac5246:
> Could I get some advice on what to do with this? I would prefer this PR gets closed or we work out what needs to be done for it to be merged, as I for some reason I made the changes on the master branch of my fork and it's annoying me that's it's still here lol. 

I can't say say for sure, but it looks like the main thing is to change when it displays.

Mr-Anonymous75:
How about something like this for the design?

BroJac5246:
> How about something like this for the design?
> 
> 

I'm fine with it.

lisa-wolfgang:
The problem with that design is that it doesn't give the user enough info about what is being displayed. It's just a dial icon and a number.

lisa-wolfgang:
I still think it's worth looking into adding a very thin section below the stage and moving this there as well as the clone counter and mouse position. The controls bar isn't meant to display information -- that's why there was a separate bar for the mouse position in Scratch 2.

lisa-wolfgang:
Given the already low space in the project controls bar and the fact that this could be confused with the max FPS setting, I think it would be best to hold on merging this addon until a decision is made on https://github.com/ScratchAddons/ScratchAddons/issues/5155 (creating our own shared space somewhere below the stage.)

Mr-Anonymous75:
> The problem with that design is that it doesn't give the user enough info about what is being displayed. It's just a dial icon and a number.

How about a show icon only toggle in the settings like the clone counter addon?

Samq64:
Linking #2396 since the volume slider exists now.

TheColaber:
@Tacodiva @WorldLanguages Do you guys think this addon is good to be merged?

Tacodiva:
Ok I messed up git really bad because I tried to revert a prettier commit which messed everything up so I force pushed and then bad git things happend, but I think I've fixed it now. As for the actual contents of this PR, WL has some points I think need to be discussed:

>Should the FPS counter be visible in project player mode at all? Should it be togglable and the default be disabled?

and also

> Also, if I switch to another tab, then come back, I see the FPS counter go from ~10 FPS to 30 FPS pretty slowly. Since the drop of framerate is caused by changing tabs, could something be done about that?

I can't reproduce this second issue so someone who can might need to look into it.

Samq64 (commented):
There's a small [shared library](https://github.com/ScratchAddons/ScratchAddons/blob/master/libraries/common/cs/small-stage.js) for handling small stage mode now.

Samq64:
If there's still any interest in this I can take things from here.

Samq64:
> Should the FPS counter be visible in project player mode at all? Should it be togglable and the default be disabled?

I thought the point of this addon over the debugger is that it can be used without the overhead of opening the editor. I guess it could be an option though.

> Also, if I switch to another tab, then come back, I see the FPS counter go from ~10 FPS to 30 FPS pretty slowly. Since the drop of framerate is caused by changing tabs, could something be done about that?

This happens with the debugger too:
[Screencast_20240803_131842.webm](https://github.com/user-attachments/assets/e4523283-b7d7-45a5-af29-82d41225d627)

Samq64 (approved):

TheColaber:
i changed the way it calculates the FPS, although I feel like it may be slightly inaccurate since sometimes the value goes above 60...

Samq64 (changes requested):
When 60fps is enabled the counter says displays infinity.

Tacodiva:
> i changed the way it calculates the FPS, although I feel like it may be slightly inaccurate since sometimes the value goes above 60...

What was wrong with how it was calculated before?

TheColaber:
It felt a little "delayed"

Samq64:
I can't repro any more but when exiting small stage it does display infinity but only for a second.

TheColaber (approved):

TheColaber:
ready to merge after updating versionAdded and fixing conflicts 🥳

addons/fps/addon.json
```diff
@@ -0,0 +1,26 @@
...
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["projects"]
+    }
+  ],
+  "tags": ["editor", "projectPlayer", "featured"],
+  "versionAdded": "1.39.0"
```
WorldLanguages, line 25:
```suggestion
  "versionAdded": "1.42.0"
```

addons/fps/userscript.js
```diff
@@ -0,0 +1,59 @@
...
+    lastRender = now;
+
+    // Show/Hide the element based on if there are any threads running
+    if (runtime.threads.length === 0) {
+      if (wasRunning) fpsCounterElement.classList.remove("show");
+      wasRunning = false;
+      return;
+    }
+    if (!wasRunning) fpsCounterElement.classList.add("show");
+    if (fps !== lastFps || !wasRunning) fpsCounterElement.innerText = msg("fpsCounter", { fps: (lastFps = fps) });
```
WorldLanguages, line 48: The clone counter addon sets an HTML attribute instead of directly setting the text. One or the other may be more performant based on registered MutationObserver's including ours and ones added by other browser extensions.

WorldLanguages (approved):

WorldLanguages:
@Samq64 For some reason GitHub was bugging out, but it seems this had 2 approvals already, yours and Colaber's.

addons/fps/addon.json
```diff
@@ -0,0 +1,26 @@
...
+      "matches": ["projects"]
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["projects"]
+    }
+  ],
+  "tags": ["editor", "projectPlayer", "featured"],
```
WorldLanguages, line 24:
```suggestion
  "tags": ["editor", "player", "featured"],
```

Samq64:
This still isn't the most accurate but it's good enough.

Samq64 (approved):

PR #3667, opened 2021-10-16
fullscreen: Fix everything that is wrong with this addon

lisa-wolfgang:
Resolves [this bug](https://discord.com/channels/806602307750985799/826250884279173162/898816304025325588) (reported by Private I#4427)
Resolves [this bug](https://discord.com/channels/751206349614088204/755131743136383107/898991419853795368) and [this observation](https://discord.com/channels/751206349614088204/819307061493760031/898947988699897876)

### Changes

Redux is now guaranteed to initialize before the `statechanged` event listener is set, which should remove the inconsistency of this addon.
Removing the scroll bars on the full screen page is now handled by JS, allowing it to only remove the scroll bars in full screen mode and not the editor.

### Reason for changes

Removing the scroll bars in the editor makes certain editor features inaccessible to users with smaller screen sizes.
The addon working inconsistently is also pretty bad.

### Tests

Tested in browser w/ Chromium 94.
Scroll bar fix tested in Firefox 93.0.

### DIFF
--- a/addons/fullscreen/userscript.js
+++ b/addons/fullscreen/userscript.js
@@ -63,6 +63,7 @@ export default async function ({ addon, global, console }) {
 
   // Changing to or from Scratch fullscreen is signified by a state change
   // (URL change doesn't work when editing project without project page)
+  addon.tab.redux.initialize();
   addon.tab.redux.addEventListener("statechanged", (e) => {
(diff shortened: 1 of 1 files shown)


lisa-wolfgang:
> 
> 
> Can you quote the bugs instead of just linking to discord

They're included in the Changes section.

PR #3674, opened 2021-10-18
Global Theme Library

### LINKED ISSUE #2298: Redo theme system

TheColaber:
Resolves #2298

Hopefully this is ok.


TheColaber:
(Btw is the permissions.html page even used?)

RedGuy12:
> ~(Btw is the permissions.html page even used?)~ Yes it is ig

it is used when site access is not granted

PR #3695, opened 2021-10-23
blocks2image: respect customisable block colours

### LINKED ISSUE #2379: "Save blocks as image" addon doesn't capture custom block colors
**Describe the bug**

The "save blocks as image" addon captures blocks as they look in default Scratch, without colors applied by "customizable block colors". 

**To Reproduce**

1. Make sure both "save blocks as image" and "customizable block colors" are enabled.
2. Change the block color settings to something besides the default (e.g. one of the presets).
3. Export blocks from the editor (as PNG or SVG, all or only selected—same result either way).
4. View the downloaded image. Note it has the default block colors, not the customized ones.

**Expected behavior**

Generally, captures of blocks should probably show as close to they actually look in the editor as possible. If for any reason a user desires otherwise (i.e. would rather capture blocks as they look in the default editor, like current), the fix could be implemented as a toggle option.

**Screenshots**

Compare the colors below (note the "when I receive" block as well as custom blocks):

On left, the editor, with brown "when I receive" and deep purple custom blocks. On right, an image capture of the same script, with the default editor colors, yellow "when I receive" and coral custom blocks.

- OS: Linux/Debian
- Browser: Firefox (Developer Edition)
- Version: 89

**Additional context**

Notably, the "custom block shape" addon [plays nicely](https://user-images.githubusercontent.com/9948030/117090934-21151f80-ad30-11eb-9cb8-42efdf5c43bf.png) with "save blocks as image" (though custom colors are still not captured).

pufferfish101007:
Resolves #2379

### Changes

Chamges fill and stroke attributes of elements in the svg so that it respect the user's chosen colours if customisable block colours is enabled.

### Reason for changes

Quality of life + consistency

### Tests

Tested on chromium 94

### DIFF
--- a/addons/blocks2image/userscript.js
+++ b/addons/blocks2image/userscript.js
@@ -78,7 +78,7 @@ export default async function ({ addon, global, console, msg }) {
       const insertBeforeIndex =
         makeSpaceItemIndex !== -1
           ? // If "make space" button exists, add own items before it
-            makeSpaceItemIndex 
+            makeSpaceItemIndex
           : // If there's no such button, insert at end
             items.length;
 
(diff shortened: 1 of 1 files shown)


WorldLanguages:
@pufferfish101007 What did you find?

pufferfish101007:
> @pufferfish101007 What did you find?

Eep I forgot abput this

Right so I tried garbo's idea and it didn't work - I've got no idea why, as I made sure to copy across the variable declarations stylesheet as well, but oh well.
Now that you've reminded me of this, I'm trying out my current approach but getting the selectors from the stylesheet rather than copying across a bunch of theme3's code.
Hopefully I'll get this done today.

GrahamSH-LLK:
imeanwecouldjustmakealib

mxmou:
I think the actual problem is that theme3 uses CSS to style blocks - it's difficult and can easily cause bugs. I tried rewriting the addon by replacing `Blockly.Colours` as suggested [here](https://github.com/ScratchAddons/ScratchAddons/pull/3491#issuecomment-919095721) and various field methods, which works better than the current version and handles most of this automatically. The only problem I noticed with that implementation is that devtools compatibility would require a lot of additional CSS.

WorldLanguages:
@mxmou Even if devtools would get more complex, it's probably worth it to start doing that.

WorldLanguages:
Closing, #3805 will fix this. Correct me if I'm wrong.

PR #3701, opened 2021-10-23
New addon: better editor comments

### LINKED ISSUE #2475: New addon - fix moving comments
**Describe the new addon**

This addon would fix comments not saving correctly when they're moved, which results in something like this:

_before switching sprites_
image

_after switching sprites_
image

lisa-wolfgang:
Resolves #2475 (https://github.com/LLK/scratch-blocks/issues/1566 with duplicates https://github.com/LLK/scratch-gui/issues/4776 and https://github.com/LLK/scratch-gui/issues/5948)

### Changes

In vanilla Scratch, when dragging a block with a comment attached to it, the position of the comment is not saved correctly, causing it to jump back to its original position when the sprite or project is reloaded. This addon fixes this bug.

Also, when the contents of comments are clipped due to the comment bubble's size, a scroll bar does not appear on the comment. This addon implements this scroll bar.

In addition, comments not attached to blocks cannot be edited directly -- the user must click the comment for it to gain focus, and only then can it be directly edited. This addon makes it so that the cursor can interact with the text without having to select the comment first.

This addon does NOT implement the comment previews that I recently showcased. Those will be part of a separate addon.

All three of these adjustments already exist in TurboWarp.

### Reason for changes

Usage of comments on Scratch has sharply declined since the debut of Scratch 3.0 due to bugginess and incompleteness. It is frequently suggested to fix these bugs. Some users even make custom blocks as stand-ins for comments, which has a negative impact on project performance.

### Tests

Tested using
- browser w/ Chromium 95
- Firefox 93.0

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -105,6 +105,7 @@
   "block-duplicate",
   "disable-paste-offset",
   "emoji-picker",
+  "fix-editor-comments",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/fix-editor-comments/addon.json
+++ b/addons/fix-editor-comments/addon.json
@@ -1,0 +1,62 @@
+{
+  "name": "Better editor comments",
+  "description": "Makes numerous changes to comments in the Scratch editor: fixes a bug where comments attached to blocks don't save their positions correctly after dragging, adds a scroll bar on comments where text is clipped, and allows you to edit standalone comments by clicking directly where you want to start typing.",
+  "tags": ["editor", "codeEditor"],
+  "credits": [
+    {
+      "name": "GarboMuffin",
+      "link": "https://github.com/GarboMuffin"
+    },
+    {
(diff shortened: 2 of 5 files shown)


RedGuy12:
@CST1229 Community discord #general a few days ago


WorldLanguages:
There's some weird incompatibility with editor-dark-mode that adds a line:
image

WorldLanguages:
Addon looks very nice!

WorldLanguages:
image
image

Reproduces inconsistently, but try:
1. Load Scratch editor with "add scroll bars to clipped comments" enabled and "editor dark mode" disabled
2. Create a comment that overflows
3. Enable dark mode
4. Disable dark mode

lisa-wolfgang:
Seems to only be a problem with macOS's scroll bars? (It looks like you're using macOS.)

WorldLanguages:
Bug I mentioned previously is probably some weird Chrome rendering bug, so doesn't matter. If a user reports this we can investigate it. I wasn't able to reproduce it again but I only tried once.

PR #3703, opened 2021-10-24
`editor-devtools`: Shortcut to not close the floating find bar

CST1229:
### Changes

If ~~Ctrl~~ Shift is held while clicking on the workspace or dragging a block from the middle-click find bar, it will not  be closed. The layer of the floating find bar is also lowered so that it appears below dragged blocks.

### Reason for changes

A little quality-of-life feature to allow rapid creation of scripts.

### Tests

Tested.

~~This might have some issues combined with block cherry picking, but hopefully it should be fine.~~

### DIFF
--- a/addons/editor-devtools/DevTools.js
+++ b/addons/editor-devtools/DevTools.js
@@ -1507,9 +1507,12 @@ export default class DevTools {
     if (this.floatInp && !e.target.closest("#s3devIDDOut")) {
       if (
         !e.shiftKey ||
+		// Clicking on the code area should always make multi-inject work
         (!document.getElementsByClassName("injectionDiv")[0].contains(e.target) &&
+		// Focused inputs are not part of the injectionDiv to the user they are part of the code area so make multi-inject work there
           !e.target.classList.contains("blocklyHtmlInput")) ||
-        e.target.matches(".blocklyFlyoutButton, .blocklyFlyoutButton *, .blocklyTouchTargetBackground")
+		  // This selection targets workspace buttons (Make a Block etc.) and the extension (!) buttons, which most commonly triggers a popup window so always close the dropdown
+		  e.target.matches(".blocklyFlyoutButton, .blocklyFlyoutButton *, .blocklyTouchTargetBackground")
       ) {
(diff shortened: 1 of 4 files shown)


addons/editor-devtools/userscript.css
```diff
@@ -440,7 +440,7 @@ div.s3devDDOut.vis ul.s3devDD {
   border-radius: 4px;
   box-shadow: rgba(0, 0, 0, 0.3) 0 0 3px, rgba(0, 0, 0, 0.2) 0 3px 10px;
 
-  z-index: 1000001;
+  z-index: 999;
```
WorldLanguages, line 443: Can you explain this change? z-index is madness and even tho 999 might appear to work just fine at first, it might not in some edge case.

addons/editor-devtools/DevTools.js
```diff
@@ -1505,8 +1505,15 @@ export default class DevTools {
     }
 
     if (this.floatInp && !e.target.closest("#s3devIDDOut")) {
-      // If we click outside the dropdown, then instigate the hide code...
-      this.hideFloatDropDown();
+      if (
+        !e.shiftKey ||
+        (!document.getElementsByClassName("injectionDiv")[0].contains(e.target) &&
+          !e.target.classList.contains("blocklyHtmlInput")) ||
+        e.target.matches(".blocklyFlyoutButton, .blocklyFlyoutButton *, .blocklyTouchTargetBackground")
```
WorldLanguages, line 1512: Could you add a code comment for this? That selector is scary lol

addons/editor-devtools/userscript.css
CST1229: It's so that dragged blocks appear above the floating find bar, which I think is better when the bar is not closed the instant a block is added.

addons/editor-devtools/DevTools.js
CST1229: Done, along with other parts of that condition

WorldLanguages:
Waiting for @griffpatch approval

WorldLanguages:
Griffpatch says we should also consider a "pin" button, but that can be added in a separate PR.

PR #3713, opened 2021-10-29
Forum user agent addon

imdevarsh:
It doesn't resolve anything as far as I know. This is a new addon.

### Changes

I've added a new addon (forum-user-agent) which adds an anchor tag to each post on the Scratch discussion forums that when clicked gets the user agent of the poster through ScratchDB by DatOneLefty and the Scratch API.

### Reason for changes

This can help people in the Bugs and Glitches forums if they need more information. It can save the hassle of going to a website like ScratchStats. It can also help people trying to give tech support because then they can know which operating system the person uses.

### Tests

I've tested this on Chrome 95.0.4638.54 and it works good. **The design right now probably needs fixing.**

I've attached a video of the current version.

https://user-images.githubusercontent.com/75148774/139388613-489123be-a5a6-4c21-a02b-f513da155288.mov

### DIFF
--- a/addons/forum-user-agent/forums.js
+++ b/addons/forum-user-agent/forums.js
@@ -12,28 +12,38 @@ export default async function ({ addon, global, console, msg }) {
     addon.tab.displayNoneWhileDisabled(line_br);
     uabutton.innerText = "Get user agent";
     uabutton.id = "user-agent-btn";
-    uabutton.title = "When you click this button, the user agent will display here.";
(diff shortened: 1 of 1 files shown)


webdev03:
Well yeah there isn't much of a use because you can just ask people their device...

WorldLanguages:
At first glance, I agree on that this addon is a bit overkill.

webdev03:
> At first glance, I agree on that this addon is a bit overkill.

Yeah...

PR #3722, opened 2021-10-31
New addon: Edit variable scope and type after creating it

### LINKED ISSUE #1469: The ability to change the "scope" of a variable after it has been created.
**Describe the new addon**
Once you create a variable, it is impossible to change it from "For all sprites" to "For this sprite only". You have to delete the variable than recreate it. It would be nice to be able to change this without having to go through that process. Ideally, a plugin would warn you if you are trying to change a global variable to a local one and it is being used in multiple sprites.

### LINKED ISSUE #3524: Toggle Cloud Variable after creation

If I make a variable, and want to turn it into a cloud variable later (or vice versa) I have to create a new variable, then manually replace every instance of that variable with the new variable

In the "rename variable" menu I should be able to toggle whether or not the variable is cloud. This won't work for "This Sprite Only" variables.

Manually replacing the variable as described above, but that's tedious.

### Additional context

Mockup:
samockup

GarboMuffin:
Closes #1469, Closes #3524

https://user-images.githubusercontent.com/33787854/139592299-1876c96a-142d-4f09-a0d7-ef03b24c7ba9.mp4

and also see: https://github.com/ScratchAddons/ScratchAddons/pull/3722#issuecomment-956782841

Variable value and monitor state are preserved. Undo/redo supported.

Variables cannot be converted if:

 - You are trying to make a cloud variable local
 - You are trying to make a local variable in the stage
 - You are trying to make a global variable used by multiple sprites local
 - You are trying to make a global variable used by an unfocused sprite local
 - You are trying to make a local variable global but the name would conflict with a local variable in another sprite


CST1229:
I think this would work better if this was in the rename variable modal, renaming that to "Edit Variable".

Explosion-Scratch:
Neato

CubeyTheCube:
Shouldn't you also be able to make ordinary variables cloud, and vice versa?

Mbrick2:
would this work in the rename popup?

GarboMuffin:
I will be adding some things to the rename variable prompt soon

GarboMuffin:
The addon now modifies the "Rename variable" prompt to allow changing a variable's scope and whether or not it's a cloud variable. The addon's name and description need to get rewritten, or the cloud variable toggling can be moved into a separate addon. The context menu option to swap between local/global still exists, but maybe it should be removed.

image

I am a little concerned people will misinterpret these buttons as "renaming a variable for every sprite" and "renaming a variable but only for this sprite" instead of changing the variable's scope.

CubeyTheCube:
Can "rename variable" be changed to "edit variable"?

WorldLanguages:
Maybe we could add a separate "Variable scope:" section to the "Rename Variable" modal, similarly to how "cloud variable" has a gray line separator above it.

TheColaber:
@WorldLanguages i think it was meant to be a replica of the "Create Variable" prompt. If we only change this one, it would be inconsistent.

WorldLanguages:
I don't think consistency is that important here. Garbo is right that it could be misunderstood as "rename in all sprites". When creating new variables that confusion cannot happen.

WorldLanguages:
This is the addon name and description I came up with. "Scope" is not good enough for the addon name because few Scratchers know what it means.
image

Joeclinton1:
Awesome stuff, looking forward to using this.

PR #3726, opened 2021-11-02
benchmark angles on rotational input addon

### LINKED ISSUE #3699: Holding shift while using rotating input should stop every 45 degree intervals
https://discord.com/channels/806602307750985799/806609030935740477/891288358800736276
> shift + rotating this to only stop at 0, 45, 90, 135, 180, -135, -90, -45
image
 
useful

TheColaber:
Resolves #3699

The name and description can be reworded if needed.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -105,6 +105,7 @@
   "block-duplicate",
   "disable-paste-offset",
   "emoji-picker",
+  "shift-rotate-input",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/shift-rotate-input/addon.json
+++ b/addons/shift-rotate-input/addon.json
@@ -1,0 +1,17 @@
+{
+  "name": "Benchmark angles on rotational input",
+  "description": "Hold shift while dragging the rotational input to set the angle to perfect benchmark angles (such as 45 degrees, 90 degrees, -45 degrees, or -90 degrees).",
+  "credits": [
+    {
+      "name": "TheColaber",
+      "link": "https://scratch.mit.edu/users/TheColaber"
+    }
+  ],
+  "tags": ["editor"],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ]
+}
--- a/addons/shift-rotate-input/userscript.js
+++ b/addons/shift-rotate-input/userscript.js
@@ -1,0 +1,11 @@
+export default async function ({ addon, msg, global, console }) {
+  while (true) {
+    const dial = await addon.tab.waitForElement("[class^=dial_container]", { markAsSeen: true });
+    const key = addon.tab.traps.getInternalKey(dial);
+    const originalDir = dial[key].return.stateNode.directionToMouseEvent;
+    dial[key].return.stateNode.directionToMouseEvent = function (e) {
+      const old = originalDir.call(this, e);
+      return e.shiftKey ? Math.round((old + this.directionOffset) / 45) * 45 - this.directionOffset : old;
+    };
+  }
+}


WorldLanguages:
We will need to make it clear that this addon is only for the dial on the right, not the one on the left (see screenshot):
image

WorldLanguages:
One on the left already stops on 15 degree intervals without Shift, but maybe to be consistent we can make this addon work on those too?

TheColaber:
> One on the left already stops on 15 degree intervals without Shift, but maybe to be consistent we can make this addon work on those too?

Hmm or I wonder if this addon should just make the one on the right have the same function as the one on left. The average user probably doesn't even use the input on the left, and if they do, they probably want to set it to a number like 90 or -90.

RedGuy12:
They should be consistent


WorldLanguages:
> Hmm or I wonder if this addon should just make the one on the right have the same function as the one on left. The average user probably doesn't even use the input on the left, and if they do, they probably want to set it to a number like 90 or -90.

I'm not really sure. This addon is pretty "small" already and you can already type in numbers fast enough if you need to change direction often.

TheColaber:
maybe this just shouldnt be an addon?

WorldLanguages:
Closing. There's already a number input to set the angle, and typing is faster.

sfederici:
> Closing. There's already a number input to set the angle, and typing is faster.

I don't agree. When you have students that don't know about angles (mine don't, and are not asked to) having a way to quickly select a benchmark angle would be a good help and relief to them. I guess both dials could need the possibility to select a benchmark angle and to select one of the 360 available directions. Using shift to activate the two selections would be a good solution.

Usually the right dial is used to work on the sprite's direction on the Stage, to change it to the one you like. So, a precise orientation would be good. But, if you need a benchmark orientation (and you don't know about angles), being able to select it by pressing shift would help a lot. It doesn't matter if you can type very quickly the value: some students (e.g. dyslexic students, that are about 20% of all students) would usually type the wrong value. To give an example, it is excellent when in Scratch you get the precise x/y coordinates in all Movement blocks when you move a sprite to the desired location.

As for the left dial, you would usually type a valuein it, but if you don't, and you use the dial, it is more common that you want to use one of the benchmark positions. So, the current behaviour is correct. If we prefer to have the same behaviour for both dials, having a standard fine-grained selection and a shift benchmark selection would be a good solution.

sfederici:
Moreover, as mentioned in https://github.com/ScratchAddons/ScratchAddons/discussions/5809, it would be great to have labels of the benchmark angles visible around on the dial. The stye I prefer is positive values to the right and negative values to the left, but this could be an option in the addon (the alternative: positive values only, 0-360, for the labels)

BroJac5246:
> Moreover, as mentioned in https://github.com/ScratchAddons/ScratchAddons/discussions/5809, it would be great to have labels of the benchmark angles visible around on the dial. The stye I prefer is positive values to the right and negative values to the left, but this could be an option in the addon (the alternative: positive values only, 0-360, for the labels)

If we did do that, then the labels wouldn't be customizable (just toggleable), since we would follow Scratch's degrees system.

sfederici:
> If we did do that, then the labels wouldn't be customizable (just toggleable), since we would follow Scratch's degrees system.

Scratch's degree system has always been double. In 1.4 and 2.0 was more oriented to positive/negative, in 3.0 to all positive. So, two toggable alternatives would be fine:  0, 45, 90, 135, 180, 225, 270, 315 or 0, 45, 90, 135, 180, -135, -90, -45.

BroJac5246:
> > If we did do that, then the labels wouldn't be customizable (just toggleable), since we would follow Scratch's degrees system.
> 
> Scratch's degree system has always been double. In 1.4 and 2.0 was more oriented to positive/negative, in 3.0 to all positive. So, two toggable alternatives would be fine:  0, 45, 90, 135, 180, 225, 270, 315 or 0, 45, 90, 135, 180, -135, -90, -45.
> 
> 

3.0 uses positive/negative, so we should stick with that.

PR #3727, opened 2021-11-02
New addon: editor comment previews

### LINKED ISSUE #3132: New addon - label C blocks at bottom

When very large scripts are created, C and E blocks can often stack overwhelmingly, to the point where when you need to add specific code below one, it's hard to figure out which is which:

image

This is also a problem in other programming languages.

At the bottom of each C and E block, this addon would add a small label reminding the user which block it is.
It would also show the contents of the conditional slot, if any.

image

When hovering over the bottom area, a preview of the top of the block could appear:

image

lisa-wolfgang:
Partially addresses #3132

### Changes

This addon introduces editor comment previews, which are small tiles that appear on hover of a comment or block. They are capable of displaying the first ~325 characters of a comment. They empower editor comments (specifically block comments) to be extremely useful to project creators.

*I left myself a quick note, but I didn't want it to take up space in the editor, so I collapsed it. I can simply hover over any collapsed comment to preview it!*

*This block is really horizontally long, but thanks to this preview, I don't have to scroll one bit to see what the comment says!*

*This is a dreaded image for every Scratcher. There are so many if statements inside of each other that I couldn't possibly tell the difference at the bottom by myself. Now I can just use comment previews to find the one I'm looking for!*

*I can also hover over any custom block to preview the comment attached to its definition block. It even works in the block palette!*

*Perhaps best of all, comment previews are fully compatible with editor dark mode!*

The addon also has a number of options regarding the look and feel of the previews:

`DELAY DURATION` (None | Short | Long), `FOLLOW MOUSE`, `REDUCE TRANSPARENCY`, `REDUCE ANIMATION`

### Reason for changes

- Reasons listed above
- Positive reception in community Discord server
- It looks pretty

### Tests

Tested in browser w/ Chromium 95.


mxmou:
I don't think this fixes #3132, because it still requires adding a comment to every conditional or loop block to be able to identify them.

lisa-wolfgang:
> I don't think this fixes #3132, because it still requires adding a comment to every conditional or loop block to be able to identify them.

This is true. Maybe I'll update this addon at some point with some kind of preview of the conditional that the loop block contains.

PR #3741, opened 2021-11-06
New addon: hide Tutorials button

lisa-wolfgang:
There are two modes in this addon.

The first mode changes the Tutorials button to act as a link. The user can specify the label and URL. The default values are set to griffpatch's YouTube channel:

image

The second mode hides the Tutorials button altogether:

image

### Reason for changes

W_L wanted me to reimplement this PR: https://github.com/ScratchAddons/ScratchAddons/pull/3673

### Tests

Tested in browser w/ Chromium 95.


Mbrick2:
could we do custom tutorials in the same format avalible??

CST1229:
Idea: A third mode to only hide the label (but keep the icon).

WorldLanguages:
> The first mode changes the Tutorials button to act as a link. The user can specify the label and URL. The default values are set to griffpatch's YouTube channel

I don't think anyone will actually use this. Browser bookmarks can achieve the same thing.

PR #3769, opened 2021-11-09
Higher resolution thumbnails

### LINKED ISSUE #3239: High resolution thumbnails addon

Scratch's website embeds a variety of bitmap thumbnails across the site: user profile pictures and project/studio covers are the main ones. Generally, Scratch aims not to waste users' internet data, so the images are actually thumbnails, hard-coded to (approximately) the portion of the screen the image will actually take up. But for some people, data usage isn't a significant issue, and their screens display at a higher resolution than the images were intended to be seen at! This leaves blurry/pixelly images that are a bit of a sore to look at, and miss out on a lot of the original detail.

(Retina/high-DPI screens are the biggest example since they're automatically using 4x the pixels to display as Scratch expects, so images look like your resolution's just a normal non-retina display for the most part. Personally, I don't have a retina display, but I usually have browser pages zoomed 120-150%, because I have bad eyesight and like reading to be easy instead of terrible. :P That means images are scaled up too though, so I get blurry images as well.)

An addon which automatically updates the source URLs for images to a higher resolution would be very nice! Just hardcoding to 2x the default dimensions (so 4x pixels) would be plenty I think, though a slider to let the user pick could be cool too. (Some people probably only need 1.2x or so for example, while others could use higher if they zoom in a lot.)

It should definitely have a warning that clearly shows it'll use more data to load images too, though. (I haven't actually fact-checked this but I'd guess it's probably exponential, so 3x upscale would be a way bigger file than 2x, which itself is already significantly larger than the default 1x?)

Honestly, there aren't any good alternatives I can think of besides just looking at blurry images. Which is fine, but suboptimal, and something an addon could do well. :P

I guess one alternative is badgering the Scratch Team to do bigger thumbnails, or spooky dynamic image loading :ghost:, themselves, but they probably won't because the data impact is too big (on both users' data plans and the ST's own CDNs) and the code could be too risky to publish to millions of active users. It'd work better as an addon because then users can choose to opt into high-res thumbnails themselves, with an understanding of data usage consequences.

### Additional context

Oh wait one more alternative is always using the site at 100% zoom on non-retina displays.

Except that's bad because then I can't see.

:shipit:

towerofnix:
Resolves #3239.

### Changes

Adds a new addon, "Higher resolution thumbnails"! This includes two primary features:

- Increase resolution of thumbnails embedded across the site. This works dynamically as images are added to the page (e.g. when a reply box is created) and takes into account lazy loading too.
- Fix thumbnails which are initially loaded at a lower resolution than they are actually displayed at (fixing blurry embeds even at 100% zoom and no high DPI/retina).

Here's the settings pane (with default values), including a warning notice about data usage:

hi-res-thumb-settings

Probably on most connections/data plans, extra usage from larger thumbnails is hardly an issue, but I thought it still worth pointing out.

### Reason for changes

Things look better! :sparkles: 

Examples (click image to see at original screenshot resolution):

*User avatar in messages, 150% zoom, addon disabled:*

old-thumb

*User avatar in messages, 150% zoom, 150% resolution scaling, "Fix initially low-res thumbnails?" checked:*

new-thumb

<details><summary>MOAR EXAMPLES</summary>

*User avatar in reply box, 100% zoom, addon disabled:*

reply-low

*User avatar in reply box, 100% zoom, 100% (i.e. no) resolution scaling, "Fix initially low-res thumbnails?" checked:*

reply-high

*User avatars in forum posts, 130% zoom, addon disabled:*

forum-low

*User avatars in forum posts, 130% zoom, 130% resolution scaling:*

forum-high

*Project thumbnails on user-subpage, 100% zoom, "More row items" addon set to 3 per row, hi-res addon disabled:*

3row-low

*Project thumbnails on user-subpage, 100% zoom, "More row items" addon set to 3 per row, 100% (i.e. no) resolution scaling, "Fix initially low-res thumbnails?" checked:*

3row-high

</details>

### Tests

Buncha manual testing :P

Thanks to @FunctionalMetatable in https://github.com/ScratchAddons/ScratchAddons/issues/3239#issuecomment-893326923 for putting together the general userscript outline!

Also, their example has code for thumbnails from uploads.scratch.mit.edu I think, but I'm not exactly sure where that URL pattern appears on the site? Input and further thumbnail locations worth watching out for would be appreciated!


addons/hi-res-thumbnails/userscript.js
```diff
@@ -0,0 +1,77 @@
...
+
+        // Use greater of both scales to choose the dimension which needs the most
+        // scaling and to maintain the ratio between the rendered and loaded
+        // dimensions.
+        const scale = Math.max(rect.width / width, rect.height / height);
+
+        if (scale > 1) {
+          width *= scale;
+          height *= scale;
+        }
```
towerofnix, line 44: One point of discussion re: implementation of this resolution fix: maybe it makes sense to *down*scale images which are by default loaded in at a higher resolution than they're actually displayed at? (Before applying the user-specified multiplier, of course.)

The main example I found of this is on the new studio pages, where for some reason thumbnails are loaded at 480x360, even though they're displayed at only about 217x203 in CSS pixels on a 1920p screen. This addon would actually save a little data usage if you set the res setting to 100% and had this fix enabled, then, because it would fetch the images at 217x203 instead of 480x360!

Although apparently, studio thumbnails load a higher res because the rendered size is semi-flexible, i.e. at thinner displays (e.g. by shrinking the window) it'll only show two projects per row, and this can get the thumbnails to around 345x260 CSS pixels. Still less than 480x360, though not by as much. Just the same, it points out a flaw with downscaling to the rendered size: if the image later dynamically *expands*, it'll look lower resolution than the original developers were giving excess dimensions as a way of intentionally allowing for!

addons/hi-res-thumbnails/userscript.js
```diff
@@ -0,0 +1,77 @@
+export default async function ({ addon, console }) {
+  const DEBUG = true;
```
towerofnix, line 2: oops

addons/hi-res-thumbnails/userscript.js
```diff
@@ -0,0 +1,66 @@
...
+      image.dataset.original = newSrc;
+    }
+
+    if (image.src) {
+      image.src = newSrc;
+    }
+  }
+
+  function scaleDimensions(width, height) {
+    const factor = 0.01 * addon.settings.get("multiplier");
```
towerofnix, line 56: I decided to make this a setting instead of trying to detect the browser zoom level (or "real" image render size) in some way. Partly because 1) [it's really difficult (possibly impossible) to actually detect that](https://css-tricks.com/can-javascript-detect-the-browsers-zoom-level/), but more importantly because 2) that either wouldn't work with dynamic zooming (i.e. as user zooms further in, image becomes blurry again), or it *would* detect zooming and would cause a TON of slightly-different-resolution image requests as the user adjusts their zoom level.

Leaving it up to the user solves both those problems while also giving them specific control over how big their thumbnails are going to end up, allowing for a maximum "nice" zoom and letting the user decide what they want that to be.

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,42 @@
...
+    }
+  ],
+  "settings": [
+    {
+      "name": "Dimension Multiplier (%)",
+      "id": "multiplier",
+      "type": "integer",
+      "min": 100,
+      "max": 300,
+      "default": 200
```
towerofnix, line 18: This is a percentage rather than a decimal (e.g. 2.00x) because I'm not exactly sure decimal settings exist, and because integers are generally nicer for users to play with (citation needed).

It'd be nice to have something like `"step": 10` here (for nicer non-typing controls), but I don't *think* there exists support for that in the current settings viewer?

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "Higher resolution thumbnails",
```
towerofnix, line 2: I decided on naming it "Higher resolution thumbnails" instead of "High resolution thumbnails" for a sort of cheeky reason: this way, when you read it, you wonder, "how *much* higher?"—and when you click to find out, it turns out the answer is you get to choose!

The default (200% width & height) is still a sensible value for anyone on a retina screen. If you're the sort of person who browses at a higher than normal zoom level, you probably already understand the specifics of what makes images blurry when you're zoomed in, and will know to set the value to however much you're normally zoomed.

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,42 @@
...
+      "default": 200
+    },
+    {
+      "name": "Fix initially low-res thumbnails",
+      "id": "fixLowRes",
+      "type": "boolean",
+      "default": true
+    }
+  ],
+  "tags": ["community"],
```
towerofnix, line 27: Maybe more tags would be better? I don't really know where a list of tags or when they're appropriate is, so I'd appreciate review from someone more experienced here :P

FunctionalMetatable:
> Also, their example has code for thumbnails from uploads.scratch.mit.edu I think, but I'm not exactly sure where that URL pattern appears on the site? Input and further thumbnail locations worth watching out for would be appreciated!

The profile pictures on the "What's happening"  section of the profile uses uploads: https://github.com/LLK/scratch-www/search?q=%22uploads.scratch.mit.edu%22&type=

addons/hi-res-thumbnails/userscript.js
mxmou: I think `devicePixelRatio` and not the zoom level is what you would need, but that still doesn't solve problem 2.

addons/hi-res-thumbnails/addon.json
mxmou: I agree that there should be a `step` option - it would also be helpful for addons that already have percentage settings, such as block transparency.

addons/hi-res-thumbnails/addon.json
mxmou: A slightly outdated list (the "featured" tag is missing) is [here](https://scratchaddons.com/docs/reference/addon-manifest/#tags-array-required). This addon isn't for specific areas of the website and isn't a theme, so I don't think any other tags are necessary.

mxmou:
> > Also, their example has code for thumbnails from uploads.scratch.mit.edu I think, but I'm not exactly sure where that URL pattern appears on the site? Input and further thumbnail locations worth watching out for would be appreciated!
> 
> The profile pictures on the "What's happening" section of the profile uses uploads: https://github.com/LLK/scratch-www/search?q=%22uploads.scratch.mit.edu%22&type=

Project and studio thumbnails on the front page are also loaded from `uploads.scratch.mit.edu`, but I don't think those need to be resized at all - they're already loaded at their original size.

towerofnix:
> > 
> > The profile pictures on the "What's happening" section of the profile uses uploads: https://github.com/LLK/scratch-www/search?q=%22uploads.scratch.mit.edu%22&type=
> 
> Project and studio thumbnails on the front page are also loaded from `uploads.scratch.mit.edu`, but I don't think those need to be resized at all - they're already loaded at their original size.

Oh, interesting! Thanks both. It seems this is the case for user avatars too—on my homepage right now is [deck26](https://uploads.scratch.mit.edu/users/avatars/1892712.png), [piano_miles](https://uploads.scratch.mit.edu/users/avatars/10494864.png), and [djdolphin](https://uploads.scratch.mit.edu/users/avatars/395764.png), and these are all [already](https://cdn2.scratch.mit.edu/get_image/user/1892712_9999x9999.png) [max](https://cdn2.scratch.mit.edu/get_image/user/10494864_9999x9999.png) [resolution](https://cdn2.scratch.mit.edu/get_image/user/395764_9999x9999.png).

This would be another case for deliberately lowering the resolution re: https://github.com/ScratchAddons/ScratchAddons/pull/3769#discussion_r745289783 (we'd have to manually reconstruct the appropriate cdn2 URL, which @FunctionalMetatable's original code does). But if we aren't for going lower than the intended (base) resolution, because of the problems discussed there, it'd probably just be inconsistent to target these avatar, project, etc images on `uploads` too.

addons/hi-res-thumbnails/userscript.js
towerofnix: Yeah, `devicePixelRatio` is *theoretically* the solution, but it [isn't consistently supported with regards to zooming](https://css-tricks.com/can-javascript-detect-the-browsers-zoom-level/#there-is-pixel-density).

Another point against dynamically detecting zoom is that users might deliberately zoom further in for the sake of reading *text* more easily (or inspecting smaller layout details)—if a user doesn't care about the image resolution here, currently they can just leave the resolution setting at the max value they deem relevant, instead of loading a bunch of data unnecessarily.

Problem 2 could also cause a lot of lag (replacing image sources frequently), and either we update *all* images (= even more lag and data usage) or we lazily target only images on- or near-screen (= less lag but pop-in as you scroll, and even more code complexity on top of dynamic zoom detection and image updating in the first place).

addons/hi-res-thumbnails/addon.json
towerofnix: Thanks for the link, and that makes sense to me!

WorldLanguages:
Related: "profile page banner" addon also automatically makes the resolution higher for featured project images. Watch out for conflicts?

micahlt:
Whoa, just gotta say that this is super cool!  Awesome job 🎉

towerofnix:
@WorldLanguages Good point, thanks! I made a small adjustment to that addon's code to support 480x360ifying the thumbnail regardless of its current size. I think it's better to do it this way rather than try to not target the featured project thumbnail from this new addon, only when the *other* addon is detected as enabled (if that's even possible).

addons/hi-res-thumbnails/addon.json
RedGuy12: yeah, the thing is, now it's a verb, and we want to try and keep addon names as nouns wherever possible

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "Higher resolution thumbnails",
+  "description": "Replaces thumbnails across the website with higher resolution images. Useful on retina displays or browsing at >100% zoom.",
```
RedGuy12, line 3: `more than 100%` instead of `>100%`

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,42 @@
...
+  "info": [
+    {
+      "type": "warning",
+      "text": "Higher resolution images require larger filesizes. Take care not to accidentally incur data expenses if your network plan is limited.",
```
RedGuy12, line 7: worded oddly

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,42 @@
...
+      "id": "imagefilesize"
+    }
+  ],
+  "settings": [
+    {
+      "name": "Dimension Multiplier (%)",
```
RedGuy12, line 13: it's a bit unclear what this setting does

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,42 @@
...
+      "id": "multiplier",
+      "type": "integer",
+      "min": 100,
+      "max": 300,
+      "default": 200
+    },
+    {
+      "name": "Fix initially low-res thumbnails",
```
RedGuy12, line 21: shouldn't be a setting imo

but if you really want to keep it,
```suggestion
      "name": "Resize initially undersized thumbnails",
```

also, mxmou made an exhaustive list of all sizes of images here fyi: https://github.com/LLK/scratch-www/issues/5598#issuecomment-864118903

RedGuy12 (changes requested):

towerofnix:
Thanks for the feedback @RedGuy12! I updated the strings across the addon.json file, so they should be clearer and easier to understand right away.

I sort of disagree that "Higher resolution thumbnails" is a verb—*higher* isn't a verb, it's an adjective—but "High resolution thumbnails" is punchy and keeping addon names sounding similar is good anyway, so I switched it to that. I also removed the setting for toggling that fix (and instead mentioned it in the addon description)—you're right that toggling it *off* doesn't really serve much purpose.

addons/hi-res-thumbnails/addon.json
WorldLanguages: #3860

addons/hi-res-thumbnails/userscript.js
```diff
@@ -0,0 +1,68 @@
+export default async function ({ addon, console }) {
+  main: while (true) {
```
WorldLanguages, line 2: Oooh, named loops! :P

addons/hi-res-thumbnails/addon.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "High resolution thumbnails",
+  "description": "Replaces thumbnails across the website with higher resolution images. Useful for Retina displays and browsing at greater than 100% zoom, and scales up thumbnails that are undersized by default.",
```
WorldLanguages, line 3: @towerofnix The description is weird. `Retina` sounds too much like an Apple only thing, maybe "high pixel density displays"? And the last sentence looks out of place, and I'm not sure what it tries to convey.

towerofnix: Yeah, I'm all for "high pixel density displays" :P It's not Apple-exclusive, obviously, but Retina displays afaik are the most like, well-known instance of high DPI displays? I went with it for recognizability—but teaching users to recognize that their Retina display is of the more general (and accurate) class/description "high pixel density display" is worthwhile.

The latter part is about scaling up images that are by default loaded at a *lower* resolution than what they're displayed at, so that they match that display (render) size. So for example, if an avatar is loaded at 38x38 but is displayed at 45x45, it'd automatically replace the dimensions with the larger ones (and then multiply by the user-selected scaling factor).

I'm not sure how to convey that in a simple sentence though, or if it's necessarily a key part of the addon (despite the fact that it's useful even if you don't use the high DPI part of the addon at all). Feedback would be appreciated!

addons/hi-res-thumbnails/userscript.js
towerofnix: Stolen frommm....

@FunctionalMetatable's example code :P https://github.com/ScratchAddons/ScratchAddons/issues/3239#issuecomment-893326923

addons/hi-res-thumbnails/addon.json
micahlt: How about `Useful for Retina (high pixel density) displays`?  That seems simple enough and it clarifies exactly what you mean by Retina.

towerofnix: Yeah, that'd be good! Another option is `Useful for high pixel density (e.g. Retina) displays`. Same difference really, but it points out explicitly that the more general term is high pixel density.

RedGuy12 (approved):
wl told me to

PR #3786, opened 2021-11-12
`emoji-picker`: add limited Sinhala letter support

CST1229:
### Changes

Adds limited support to Sinhala letters to the emoji picker.

### Reason for changes

It's very sus and also people on the discord wanted it.

### Tests

Works

(This will probably get closed but I'll try)


CST1229:
apple sabotaged
sus deactivated

PR #3789, opened 2021-11-12
editor-comment-previews: fix some bugs

GarboMuffin:
- Don't needlessy force the flyout to completely rerender
 - Various other minor optimizations
 - Fix comment previews in the flyout not working after leaving & entering editor
 - Don't show preview for comments that only contain whitespace
 - Right clicking should hide preview as left clicking already did
 - Improve preview readability in browsers that don't support backdrop-filter (Firefox)
 - Hovering over a text input will now show the comment attached to the block that the input is in (previously it would show nothing, don't know if that was intentional)
 - Fix various other bugs


lisa-wolfgang:
It's only missing one change... you can probably guess what I'm going to say lol

PR #3790, opened 2021-11-12
swap-local-global: theoretical bug fix

GarboMuffin:
The old canUserUseCloudVariables() was unreliable in certain situations, although as far as I'm aware Scratch Addons itself wasn't affected (not entirely sure, there might be some weird edge case). This makes it work reliably everywhere and should be more resistant to any Scratch changes.


apple502j:
Can't we just use `typeof vm.runtime.ioDevices.cloud.provider !== "undefined"` ?

GarboMuffin:
That is always null when the project doesn't already have cloud variables, even if the user is able to make them

PR #3795, opened 2021-11-13
Split project-info addon

### LINKED ISSUE #3742: Make block count a separate addon

The "sprite and script count" addon has an option to show a block count in the editor. Because it's shown in a completely different place, I don't think it should be part of the same addon as the sprite and script count.

Split the addon into two - "sprite and script count" and "block count".

Keep it as a single addon, but change the name and the description to mention the block count. Possible name and description:
* Sprite, script, and block count
   * Shows the number of sprites and scripts a project has next to the Remix button, and the number of blocks in the editor menu bar.

mxmou:
Resolves #3742

### Changes

Makes block-count a separate addon from project-info.

### Reason for changes

The addons are related, but the UI they add is in completely different places.

### Tests

Both addons still work like before.


CST1229:
What about adding transition code to transition project-info users into the new split addons?

WorldLanguages:
Yes, we'd need to add transition code to get-manifests.js

WorldLanguages:
Transition handled by issue #3861

WorldLanguages:
@mxmou Should versionAdded be 1.4.0? I think it would be nice to include this addon under "other new addons"

mxmou:
> @mxmou Should versionAdded be 1.4.0? I think it would be nice to include this addon under "other new addons"

I think showing it as a new addon is better because people might not know about it - if we decide to do that, the description should mention that it was previously part of the sprite and script count addon.

If not, `versionAdded` should be 1.7.0, which is when the setting was added.

PR #3801, opened 2021-11-14
New addon: Profile message count

itta611:
### Changes
This addon shows user message count on profile page.
demo

### Tests
Tested on Google Chrome 95.0.4638.69


itta611:
oh

PR #3802, opened 2021-11-14
New addon: Profile message count

itta611:
### Changes
This addon shows user message count on profile page.
demo

### Tests
Tested on Google Chrome 95.0.4638.69

### DIFF
--- a/addons-l10n/en/profile-msg-count.json
+++ b/addons-l10n/en/profile-msg-count.json
@@ -1,0 +1,3 @@
+{
+  "profile-msg-count/unread-messages": "Unread messages"
+}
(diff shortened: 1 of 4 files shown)


lisa-wolfgang:
Why under the follow button? Just curious.

itta611:
Just a hunch. I think it looks cluttered when displayed on the left.

itta611:
@WorldLanguages Any updates?

WorldLanguages:
I'm unsure about this addon's concept. Message counts should ideally stay private, they simply aren't because of performance reasons. And I don't think this has been suggested by any users.

WorldLanguages:
Closing, if someone is interested, create an issue and we can discuss whether we want this as an addon or not, then we can reconsider.

TheColaber:
This is something you really shouldn't or don't need to know, if you don't own the account. I think knowing message counts for other accounts you own could be implemented in an account switching addon, which I have a pr open somewhere.

Issue #3782, opened 2021-11-12
Comment straightening + leashing


lisa-wolfgang:
https://github.com/ScratchAddons/ScratchAddons/pull/3731 does not currently implement any functionality for comments. The ideal solution would be to add a "straightening" feature that would ensure that all block comments' connections are perfectly horizontal. It would also be nice to ensure that block comments don't stray too far from the actual block, which does happen quite often in sprawling workspaces.

Comment leashing should probably go in `fix-editor-comments` because it's pretty much a bug fix. The only problem is deciding where to put comment straightening. There are 2 main options:
1. In `script-snap`, which would make sense because they make comments neater, but the comment leash feature wouldn't fit here.
2. In `fix-editor-comments`, which would definitely make sense for the comment leash because the user doesn't typically expect block comments to be created about three screens away from the block, but maybe not the best option for the comment straightening since that isn't something everyone would want. But then again, this addon is all about enhancing comments...

It could alternatively be its own addon. It would be a rather small one, and I personally think it would fit better in another addon, but I'm just jotting this down to consider.

Samq64:
> ...maybe not the best option for the comment straightening since that isn't something everyone would want.

Why is that a problem? Each bug fix in that addon is already togglable, so this one could just go there and be disabled by default.

lisa-wolfgang:
> 
> 
> > ...maybe not the best option for the comment straightening since that isn't something everyone would want.
> 
> Why is that a problem? Each bug fix in that addon is already togglable, so this one could just go there and be disabled by default.

I suppose.

PR #3804, opened 2021-11-14
fix-editor-comments: add comment snapping and leashing

lisa-wolfgang:
Resolves https://github.com/ScratchAddons/ScratchAddons/issues/3782

### Changes

Two new features have been added to `fix-editor-comments`: "Always spawn block comments next to block" (leashing) and "Keep block comment connections straight" (snapping).

**Before leashing:**

https://user-images.githubusercontent.com/43426138/141669491-2ce2bdcc-2f8f-496b-afa6-27c2c3815033.mp4

**After leashing:**

https://user-images.githubusercontent.com/43426138/141669490-028debc8-349e-4620-b5b1-2fb728121d6e.mp4

Here's what snapping looks like:

https://user-images.githubusercontent.com/43426138/141669488-03ca0f95-0e54-4c11-9551-9a537f374895.mp4

### Reason for changes

Off-screen comments are extremely annoying. They are one of the deterrents that keep people from using block comments.
Snapping also makes comments a lot neater-looking while still allowing the user to move them horizontally, which also makes block comments more appealing to use.

### Tests

Tested on browser w/ Chromium 95 and Firefox 94.0.1.


addons/fix-editor-comments/addon.json
```diff
@@ -9,7 +9,7 @@
     },
     {
       "name": "lisa_wolfgang",
-      "link": "https://github.com/lisa-wolfgang"
+      "link": "https://scratch.mit.edu/lisa_wolfgang"
```
Samq64, line 12:
```suggestion
      "link": "https://scratch.mit.edu/users/lisa_wolfgang/"
```

lisa-wolfgang:
TODO: When https://github.com/ScratchAddons/ScratchAddons/pull/3811 is merged, add `latestUpdate` property to manifest.

addons/fix-editor-comments/addon.json
```diff
@@ -52,6 +52,18 @@
       "id": "standalone-edit",
       "type": "boolean",
       "default": true
+    },
+    {
+      "name": "Always spawn block comments next to block",
+      "id": "leash",
+      "type": "boolean",
+      "default": true
+    },
```
WorldLanguages, line 61: Are we sure all old users of this addon will be fine with this being enabled automatically?

WorldLanguages: Leashing does not happen that often, and it's easy to trace the behavior to this addon, so I'd say it's fine.

lisa-wolfgang: Very few people would want a new comment to be created off-screen automatically. There is little purpose to it, and it is extremely annoying in most cases.

PR #3805, opened 2021-11-14
Rewrite editor-theme3 using less CSS

### LINKED ISSUE #2379: "Save blocks as image" addon doesn't capture custom block colors

### LINKED ISSUE #3750: "Customizable block colors" doesn't update the color of block comment borders
### Description

There's a little-recognized feature of scratch-blocks which makes the border of comments attached to blocks match the primary color of the block. The "Customizable block colors" addon updates those colors for actual blocks, but any attached comments will be bordered with the original block color, causing a mismatch.

### Repoduction steps

1. Use "Customizable block colors" to change the color of a block category to a distinctly different color than the 3.0 default.
2. Drag out a block from that category.
3. Right click on it and choose "Add Comment" from the menu.
4. Look at the color of the comment border and compare it to the color of the block (as well as the color used for that block's category in 3.0 default).

### Expected behaviour

Comment borders should match the updated color!

### Actual behaviour

They always use 3.0's default colors, and cause a visual mismatch.

### Screenshots

Comment attached to a custom block. The custom block is a purple similar to in Scratch 2.0, but the comment shows default 3.0's coral-red color as its border.

### Environment

- Browser: Firefox 91.3.0esr
- Scratch Addons version: 1.21.2
- Operating system: Linux (Debian)

### Additional context

None! :space_invader:

mxmou:
Resolves #2379
Resolves #3750

### Changes

Removes most of the CSS in the editor-theme3 addon and replaces `Blockly.Colours` (suggested by @TheColaber [here](https://github.com/ScratchAddons/ScratchAddons/pull/3491#issuecomment-919095721)) and `.init` methods of various fields instead.

There is a new custom CSS variable type: `"alphaThreshold"`.

### Reason for changes

It will be easier to work with than the current implementation of the addon. It also loads faster.

### Tests

The addon functionality is mostly unchanged. Transparency now works.

### To do

* [ ] Is there a better way to reinitialize all blocks? (I want to avoid workspace update too)
* [x] Decide what to do with transparency
* [x] A better way to communicate with other addons that need colors
* [x] Even better way to communicate with other addons that need colors


addons/editor-theme3/theme3.js
```diff
@@ -1,398 +1,282 @@
...
+
+    // Reload toolbox
+    toolbox.dispose();
+    toolbox.workspace_ = workspace;
+    toolbox.init();
+    // Connects events to VM
+    // https://github.com/LLK/scratch-gui/blob/ba76db7350bd43b79119cac2701bc10f6c511f0c/src/containers/blocks.jsx#L250-L254
+    const flyoutWorkspace = workspace.getFlyout().getWorkspace();
+    flyoutWorkspace.addChangeListener(vm.flyoutBlockListener);
+    flyoutWorkspace.addChangeListener(vm.monitorBlockListener);
```
TheColaber, line 273: like in this commit, there should specific things you want to reload, instead of disposing and initing, which probably is slower. plus, its probably better not to touch vm here.

addons/editor-theme3/theme3.js
```diff
@@ -1,398 +1,282 @@
...
+    devtoolsId: "operators",
+  },
+  {
+    id: "variables",
+    settingId: "data-color",
+    colorId: "data",
+    devtoolsId: "data",
+  },
+  {
+    id: null,
```
TheColaber, line 59: Hmm, this is currently `null`? Does this mean you can't currently change the color of the lists category?

addons/editor-theme3/theme3.js
```diff
@@ -1,398 +1,282 @@
...
+    id: "operators",
+    settingId: "operators-color",
+    colorId: "operators",
+    devtoolsId: "operators",
+  },
+  {
+    id: "variables",
+    settingId: "data-color",
+    colorId: "data",
+    devtoolsId: "data",
```
TheColaber, line 56: If devtools uses a different id, I think it should be the one to change here.

addons/editor-theme3/theme3.js
```diff
@@ -1,398 +1,282 @@
...
+    colorId: "more",
+    devtoolsId: null,
+  },
+  extensionsCategory,
+];
+
+export default async function ({ addon, console }) {
+  const Blockly = await addon.tab.traps.getBlockly();
+
+  let originalColors = JSON.parse(JSON.stringify(Blockly.Colours));
```
TheColaber, line 76: I'm curious why you need to stringify and parse

TheColaber (changes requested):

addons/editor-theme3/theme3.js
mxmou: The `id` property is the Blockly category ID and it's only used to recolor the category bubbles. I'll add the correct ID later because it's only required for compatibility with data category tweaks.

addons/editor-theme3/theme3.js
mxmou: I'm not sure what the best way to handle devtools compatibility is, but I'll probably remove these `devtoolsId`s.

addons/editor-theme3/theme3.js
mxmou: Without this, changing the colors would also affect `originalColors`.

addons/editor-theme3/theme3.js
mxmou: Is it possible to redraw every block in the flyout without recreating the workspace (which requires interacting with the VM)?

addons/editor-theme3/theme3.js
```diff
@@ -1,398 +1,329 @@
...
+    oldFieldVariableGetterInit.call(this);
+    this.textElement_.style.fill = textColor(this);
+  };
+
+  const oldFieldMatrixUpdateMatrix = Blockly.FieldMatrix.prototype.updateMatrix_;
+  Blockly.FieldMatrix.prototype.updateMatrix_ = function () {
+    // Matrix inputs
+    oldFieldMatrixUpdateMatrix.call(this);
+    console.log("updating matrix");
+    console.log("still updating matrix");
```
WorldLanguages, line 241: Remove console logs?

WorldLanguages:
Some things we need to check before merging:
- Test that the addon can handle settings that previously had transparency
- A better way to make addons communicate than using a global `window._devtoolsGetColorsForCls`
- Make sure standalone devtools extension will work properly with this cross addon communication
- Keep in mind a user might run Scratch Addons v1.23.0 and older devtools extension at the same time - what would happen in those cases? The opposite might also happen.

mxmou:
> Test that the addon can handle settings that previously had transparency

The addon can, but the settings page can't - it doesn't let the user remove the transparency without resetting the color. I'm not actually sure if fixing that is easier than just overriding the relevant Blockly method to support transparency.

> Keep in mind a user might run Scratch Addons v1.23.0 and older devtools extension at the same time - what would happen in those cases? The opposite might also happen.

Older versions of devtools defined the colors using CSS, so they would use the default ones. In the opposite case, `_devtoolsGetColorsForCls` and `_devtoolsGetColorsForCategory` would be undefined, which would also result in the default colors being used.

Samq64:
I don't think the bug label should belong on PRs. I think a new bug fix label should be made instead.

TheColaber:
This isn't a bug fix anyway, but the same tags are meant to be used on all issues, prs, and discussions.

WorldLanguages:
I'm not sure when we will be ready for this, since it needs coordination with a DevtoolsExtension update.

mxmou:
Transparency is now supported, even for block inputs:
image

Debugger needed a way to get the color so I added an addon API for that. The devtools extension could probably interact with it using the global `scratchAddons` object.

addons/editor-theme3/theme3.js
TheColaber: @mxmou How does this sound... I tested it briefly and the flyout still seemed to work.
```suggestion
    toolbox.flyout_.dispose();
    toolbox.createFlyout_();
    toolbox.showAll_();
    toolbox.flyout_.position();
```

GarboMuffin (changes requested):

addons/editor-theme3/theme3.js
mxmou: There is still a bug: category bubbles are not updating.

WorldLanguages:
@GarboMuffin Are the issues you mentioned now resolved?

RIMOPA:
Is this still "less CSS"?
image

WorldLanguages:
@RIMOPA Even if this ends up being more lines of code, this method to color blocks fixes many bugs and makes more sense

TheColaber (approved):

TheColaber:
This PR will fix bug where currently changing the settings of this addon does not update the styles, and instead disables them.

apple502j (approved):

PR #3710, opened 2021-10-26
Convert website dark mode to presets

mxmou:
Resolves #645 (not the fonts part, but there is another issue for that: #1646)
Adds a separator between comments as suggested in #792
Resolves #1833
Resolves #2740
Resolves #2741
Resolves #3079
Resolves #3829

### Dark mode changes

Adds color settings to website dark mode, with presets for the current "Experimental Dark" style and Scratch defaults. Dark WWW (#517) is also a preset.
image

On scratchr2 pages, variants of the selected colors are generated to make sure that the Scratch's default colors preset doesn't change those pages.

To do:
* Find and fix bugs
* [x] Make the colored areas setting affect banners added by Scratch 2.0 → 3.0
* Migrate existing settings (#3861)

### Other changes

* The `threshold` property of a `textColor` custom CSS variable can be a color provider
* `brighten` properly supports changing the opacity of a color
* Fixed the `isNotScratchWWW` regular expression not matching `/educators/register` and `/educators/waiting`
* Slightly changed the design of `message-filters` and `project-notes-tabs`

PR #3457, opened 2021-09-10
hide-project-stats: fix padding, show own stats option, My Stuff support

lisa-wolfgang:
Resolves #3376
Resolves #3374

### Changes

The padding between the buttons when the labels are hidden has been adjusted to match the appearance of vanilla Scratch.

image

It is now also possible to show statistics on the project page when the user owns the project they're viewing:

image

When the user is signed out, and the love and favorite counts are set to be hidden, the love and favorite buttons won't appear at all.

In addition, the addon now also supports the My Stuff page. If "show stats on my own projects" is enabled, all statistics will show in My Stuff, since they are all projects owned by the user. As the comment count and (broken) studio count for a project is viewable in My Stuff, new settings for these have also been added.

image

### Reason for changes

The new appearance of the love and favorite buttons is much improved from the existing version, which has inconsistent spacing depending on the love/favorite counts.
Also, users may still want to be able to see stats on their own projects without having to disable and reenable the addon.
In addition, for a truly comprehensive project statistic hiding experience, the user would also want the stats on the My Stuff page to be hideable as well.

### Tests

Tested to work on Chromium 106 and Firefox 105.0.3.

### Known issues
None currently.

Issue #2384, opened 2021-05-05
addon.json property: `history`


TheColaber:
Example:
```json
{
  "history": [
    {
      "event": "added",
      "version": "1.10.0"
    },
    {
      "event": "update",
      "info": "Added settings",
      "version": "1.11.0"
    },
    {
      "event": "update",
      "info": "Fixed elements not appearing sometimes",
      "version": "1.11.1"
    },
    {
      "event": "update",
      "info": "Now has presets",
      "version": "1.13.0"
    }
  ]
}
```
1. This would be nice for devs
2. This would be nice for the user if we implemented #1424 so that way they won't have to go to changelog for updated addons
3. History can be kept in the settings page in the form of #475
4. This might be a better solution then `versionAdded`

Hans5958:
Hello, 2021! Somehow found this issue out of nowhere.

It would be better to implement it on the documentation instead. The plans on ScratchAddons/website-v2#260 would have the changelog section, which let devs describe the changes better.

WorldLanguages:
Agree with Hans - this won't belong in the addon manifest.

PR #2010, opened 2021-03-29
editor-stage-left: add setting to have 2.0 button order (fullscreen on left)

WorldLanguages:
PRing on behalf of djdolphin
Right now, refresh is needed to get disabling/enabling this setting to take effect. We might want to wait until dynamicDisable and dynamicEnable work properly so that's not the case.

Image with setting enabled, see how fullscreen icon is on the left, just like 2.0
image

PR #3811, opened 2021-11-15
Settings page: UI for feature additions in existing addons

### LINKED ISSUE #1424: "Update" Tooltip
Update-scratchaddons-tooltip

I thought when a tool/modification gets updated recently it will show what's new in that tool/modification in a update toolotip.
Or
There also should be a update log on the tool's/modification's box.

This could allow users to pay more attention to the new feature in the tool/modification.

lisa-wolfgang:
Resolves #1424

### Changes to settings page

The settings page now looks like this:

image

Addons that are designated as updated now appear at the bottom of the "New" categories, which have been renamed accordingly to "Featured new addons and updates" and "Other new addons and updates". As a consequence, the tag sort order has been tweaked.

A description can be specified for the update, which appears at the top of the notices section. Addon settings can also be designated as updated and will have a New tag next to them.

image

Just like the "New" tag, these indicators will automatically disappear in subsequent releases.

### Changes to addon manifest

A new property has been added.

**`latestUpdate` (object, optional)**
Provides information to the settings page about the latest update to this addon. Responsible for the "New options" tag and the New tag on individual addon settings. If not specified, no update-specific tags will be added.

Sub-properties:
- `"version"` (string, required) The version that this update applies to. If the value is the same as the current version of the extension, the addon will get the "New options" tag and display the indicators (if any) specified in `temporaryNotice` and `newSettings`.
- `"isMajor"` (boolean, optional) Indicates whether the addon should be included in the "Featured new addons and updates" category. Use the same discretion that you would for the Featured tag on regular new addons. Defaults to `false` if not specified.
- `"temporaryNotice"` (string, optional) A text description of the new features or changes. This will appear as a notice-style banner at the top of the info section (above any warnings or notices) when the addon has the "New options" tag. If not specified, no banner will be displayed.
- `"newSettings"` (array of strings, optional) The IDs of any settings that should be designated as new in this update. These settings will display the New tag when the addon has the "New options" tag. If not specified, no settings will display the New tag.

Example:
```json
"latestUpdate": {
  "version": "1.22.0",
  "isMajor": true,
  "temporaryNotice": "You can now set a whole lotta new settings.",
  "newSettings": ["test-2", "test-3"]
},
```
If these changes are merged, the manifest schema will need to be updated.

### Reason for changes

Feature additions to addons are currently not showcased anywhere on the settings page. As some new features are particularly noteworthy, there should be a way to showcase them alongside the new addons.

### Tests

Tested on browser w/ Chromium 95 and Firefox 94.0.1.


apple502j:
Is there a way to indicate an update without options?

mxmou:
> Is there a way to indicate an update without options?

A better name for the tag might be "Updated" with a tooltip that describes what is updated, as suggested in #1424.

mxmou:
Could updated addons be a separate group on the setting page?

lisa-wolfgang:
> Is there a way to indicate an update without options?

Not at the moment. We would need to discuss the best UI for this.

This shouldn't be a tooltip because users wouldn't expect that kind of information to go there, so few people would hover over the tag. The latest update should be alongside the warnings and notices in the respective extension version, and in future versions, it should be automatically appended to the addon description. Of course, if we decide that this is the ideal strategy, we would need to implement something like #2384.

> A better name for the tag might be "Updated" with a tooltip that describes what is updated, as suggested in #1424.

I chose the name "New options" because it hints to the user that they need to open the addon settings to enable and/or learn about the features that the update adds. "Updated" doesn't even suggest that new functionality was added, leading people to think that it's not important enough to deserve their attention, which would defeat the purpose of this feature.

> Could updated addons be a separate group on the setting page?

I dislike this because it further increases the perceived importance gap between updates and new addons. Some addon updates are significantly more useful than some non-featured addons. That's why I added `isMajor` -- most updates will probably go under the "Other" category, but anything feature-worthy has the option to be displayed in the "Featured" category.

lisa-wolfgang:
More ideas for UI:
- Having the update description display below the addon title

image

- Auto-expanding the addon (suggested by Biker2000: https://discord.com/channels/806602307750985799/809063831841407018/910008475076165632)

lisa-wolfgang:
I think it might be better to place that update description notice at the top of the notices section inside the addon. Hiding the addon description would make it harder for users who haven't seen the addon to figure out what it does.

Automatically appending the update descriptions to the main addon description in future versions also seems like a bad idea in retrospect.

I don't think the lack of update description support should prevent this addon from being merged for v1.23 --  https://github.com/ScratchAddons/ScratchAddons/pull/2010, https://github.com/ScratchAddons/ScratchAddons/pull/3825, https://github.com/ScratchAddons/ScratchAddons/pull/3804, https://github.com/ScratchAddons/ScratchAddons/pull/3457, and https://github.com/ScratchAddons/ScratchAddons/pull/3710 are features that would greatly benefit from this, especially since customizable website colors have been very highly requested. (It would also help beef up the new addon section this release...)

lisa-wolfgang:
Known issue: the update description banner has a slightly greater height than regular notices and warnings.

image

This is because the New tag that is used as an icon has a greater height than the icons for notices and warnings, and I don't think there's a simple way to fix this. (Besides possibly removing the New tag? Most update descriptions will be pretty obvious that they're talking about updates, and the purple color is the same as the New tag, although color isn't the most accessible indicator.)

lisa-wolfgang:
Something else to discuss is how to decide how "important" an update has to be to go alongside new addons.

Obviously, new features and settings should be recognized as well as newly split addons.
Not so sure about bug fixes and performance enhancements, unless they're really significant. Same goes for presets and preset updates.
Definitely not things related to the presentation on the settings page such as modified tags, with the possible exception of when an addon moves out of beta.

CST1229:
What about showing the update description instead of the addon description for _people who have the addon enabled_?

lisa-wolfgang:
> What about showing the update description instead of the addon description for _people who have the addon enabled_?

I like the idea, but it might be a little inconsistent/confusing to have enabled addon updates and disabled addon updates side-by-side.

addons/test-to-remove/addon.json
```diff
@@ -0,0 +1,81 @@
+{
+  "name": "Test to remove",
+  "description": "This is a test for the update label PR.",
+  "tags": ["editor", "beta"],
+  "info": [
+    {
+      "type": "warning",
+      "text": "This addon should be deleted before the PR is merged, and its entry in addons.json should also be removed.",
+      "id": "wl-dont-forget"
```
WorldLanguages, line 9: LOL

apple502j:
Manifest: ScratchAddons/manifest-schema#51

PR #3825, opened 2021-11-18
`forum-id`: add automatically adding link, add dynamic enable/disable

### LINKED ISSUE #2468: Link to quoted post

Scratch doesn't provide an easy way to jump to the quoted post from a quote.

An option in the Better quoter addon to automatically add a link when quoting a post.

It could also be its own addon.

CST1229:
Resolves #2468

### Changes

Adds an option in `forum-id` to automatically add a link to the post when it is quoted.
Also adds dynamic enable and disable to it.
(Maybe needs discussion: should we use post number or some other text?)

### Reason for changes

More convienient than having to click another link every time.

### Tests

Tested on Firefox.

(Actually real PR this time)


WorldLanguages:
> (Maybe needs discussion: should we use post number or some other text?)

What are the alternatives?

CST1229:
> 
> 
> > (Maybe needs discussion: should we use post number or some other text?)
> 
> What are the alternatives?

Something like "Original post" or "view post".

RedGuy12:
Or make it a setting


CST1229:
> Or make it a setting

Done

addons/forum-id/userscript.js
```diff
@@ -8,10 +8,41 @@ export default async function ({ addon, console, msg }) {
...
+    $.ajax("/discuss/post/" + id.substr(1) + "/source/").done(function (data) {
+      paste(
+        "[quote=" +
+          username +
+          "][small](" +
+          getIDLink(
+            id.substring(1),
+            addon.settings.get("auto_add_use_number")
+              ? post["0"].querySelector(".box-head > .conr").textContent
+              : "View post",
```
WorldLanguages, line 39: "View post" would need to be localized here (kind of - the forums themselves aren't localized completely).

WorldLanguages:
@CST1229 To avoid having to localize "view post", could we just use the post number as originally intended, instead of making it togglable?

CST1229:
> @CST1229 To avoid having to localize "view post", could we just use the post number as originally intended, instead of making it togglable?

Oh yeah I completely forgot about l10n.

WorldLanguages:
I think it's better to make it always use the post number.

CST1229:
> I think it's better to make it always use the post number.

Yes I thought that too.

CST1229:
> I think it's better to make it always use the post number.

Done

PR #3831, opened 2021-11-20
Bump to v1.23.0-prerelease

WorldLanguages:


WorldLanguages:
This is my last week of school, so I'll be way more active starting Saturday Nov 27.
v1.23.0 should reach users around December 7th.

Issue #3314, opened 2021-08-17
Reset dropdown `defaultPresetName` option


mxmou:
> > The only problem(?) would be that 3.Darker would be called "Default" in the reset menu.
> 
> The addon will still be called "Editor dark mode [...]" so I don't think that's a problem, tbh.
> We can work around this and make it so it says "3.Darker" instead of "Default", somehow

It depends on how the dropdown menu would work. In cases like this, a way to tell that default settings have a name in `addon.json` (maybe a property called `defaultPresetName`) so they would appear with that name in the reset menu would be useful.

_Originally posted by @mxmou in https://github.com/ScratchAddons/ScratchAddons/issues/1805#issuecomment-797306800_

WorldLanguages:
This would be good. But I think it should say `3.Darker (default)`, not just `3.Darker`. That's an implementation detail tho

PR #3873, opened 2021-12-03
`script-snap`: remove the default preset

### LINKED ISSUE #3864: Snap scripts addon should not have a "default" preset

The new "snap scripts to grid" addon has a preset named "Default" which resets the grid size setting to its default value (40px). Addons shouldn't do this because there is already a "reset to default settings" button next to the addon toggle.

One of the following options:
* Just remove the preset. The default value would still be accessible using the reset all button or the reset dropdown next to the input. This might make the functionality harder to access - the reset button is quite far away from the list of presets.
* Add an implicit "Default" or "Reset all" preset to every addon with presets. There would need to be a way to rename or remove it using `addon.json` (in editor dark mode it needs to be called 3.Darker - related: #3314).
* Move the reset all button below the settings:
image

Samq64:
Resolves #3864

### Changes

Removes the `Default` preset from `script-snap`.

### Reason for changes

Most add-ons don't have a default preset and the reset button does the same thing.

### Tests

I didn't test it because I'm on mobile right now.

### DIFF
--- a/addons/script-snap/addon.json
+++ b/addons/script-snap/addon.json
@@ -45,13 +45,6 @@
       "values": {
         "grid": 24
       }
-    },
-    {
-      "name": "Default",
-      "id": "default",
-      "values": {
-        "grid": 40
-      }
     }
   ],
   "dynamicEnable": true,


TheColaber:
> * Add an implicit "Default" or "Reset all" preset to every addon with presets. There would need to be a way to rename or remove it using `addon.json` (in editor dark mode it needs to be called 3.Darker - related: #3314).
> * Move the reset all button below the settings:
image

I think one of these should be implemented before we do this.

Samq64:
Might as well just remove the preset in the same PR as the rest of #3864

PR #3880, opened 2021-12-04
Integer stepping

### LINKED ISSUE #3860: "step" property for integer settings
towerofnix:
> It'd be nice to have something like `"step": 10` here (for nicer non-typing controls), but I don't think there exists support for that in the current settings viewer?

mxmou:
> I agree that there should be a `step` option - it would also be helpful for addons that already have percentage settings, such as block transparency.

source: https://github.com/ScratchAddons/ScratchAddons/pull/3769#discussion_r745293027

Samq64:
Resolves #3860

### Changes

Allows setting the `step` property through the add-on manifest and applies it to a few add-ons.

Some other values have also been changed:

- ~~The 2.0 block notch size is now 70 (used to be 75)~~
- The minimum zoom is now 10 (used to be 1)
- The default starting zoom is now 70 (used to be 68)

### Reason for changes

It makes changing integer values with the arrows easier.

### Tests

Tested on Edge 96.

### DIFF
--- a/addons/custom-block-shape/addon.json
+++ b/addons/custom-block-shape/addon.json
@@ -38,6 +38,7 @@
       "type": "integer",
       "min": 50,
       "default": 100,
+      "step": 10,
       "max": 200
     },
     {
@@ -46,6 +47,7 @@
       "type": "integer",
       "min": 0,
       "default": 100,
+      "step": 10,
       "max": 300
     },
     {
@@ -54,6 +56,7 @@
       "type": "integer",
       "min": 0,
       "default": 100,
+      "step": 10,
       "max": 150
     }
   ],
@@ -75,7 +78,7 @@
       "values": {
         "paddingSize": 70,
         "cornerSize": 150,
-        "notchSize": 75
+        "notchSize": 70
       }
     },
     {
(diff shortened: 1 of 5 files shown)


mxmou:
> * The 2.0 block notch size is now 70 (used to be 75)
> * The default starting zoom is now 70 (used to be 68)

I don't think these values should be changed - especially the second one, which matches Scratch's default zoom (67.5%) almost exactly. If you don't want the steps to be 68, 78, 88, 98, ..., you can change the step value (to 5 and 2).

WorldLanguages:
Can you explain why you're changing mins and defaults? I don't get why adding stepping would make us change those.

Samq64:
> Can you explain why you're changing mins and defaults? I don't get why adding stepping would make us change those.

Anytging that's not a multiple of the step us considered an invalid value.

WorldLanguages:
I now notice I never said it explicitly, but we'd still want to allow non multiples as valid values. The stepping would simply be for up/down controls.

mxmou:
> I now notice I never said it explicitly, but we'd still want to allow non multiples as valid values. The stepping would simply be for up/down controls.

I think we would need to implement our own up/down controls in that case.

PR #3882, opened 2021-12-04
Migration code for dark-www

apple502j:


WorldLanguages:
Tested, worked in all cases.

WorldLanguages:
Can you confirm the migration is correct?
- scratchr2's `primaryColor` is used for dark-www's `navbar` and `button`
- scratchr2's `linkColor` is used for dark-www's `link`
- scratchr2's `activeColor` is discarded, but it's not deleted from storage, unlike the above two.

mxmou:
> Can you confirm the migration is correct?
> 
> * scratchr2's `primaryColor` is used for dark-www's `navbar` and `button`
> * scratchr2's `linkColor` is used for dark-www's `link`
> * scratchr2's `activeColor` is discarded, but it's not deleted from storage, unlike the above two.

Yes, this is correct.

PR #3904, opened 2021-12-09
Update existing settings that use presets

### LINKED ISSUE #3861: get-addon-settings.js migrations for v1.23.0
There will be more than one migration this release, ~~so it's better to code them all at once~~.

- [x]  #3878 Users that had the `editorCount` setting of `project-info` set to true, need to have the new `block-count` addon enabled (#3795)
- [ ] TBD what should be done about users on the old TurboWarp Dark preset after the update (#3839)
- [x] #3882 / Migrate scratchr2 settings to dark-www (https://github.com/ScratchAddons/ScratchAddons/pull/3710#issuecomment-952652969)
- [ ] TBD what should be done about users with the old Experimental Dark preset after the update (#3710)

Discussion for how presets should migrate is happening on https://github.com/ScratchAddons/ScratchAddons/discussions/3855

We could also remove old migration code by now.
And related, I find it funny that some users tell me they see dango in my profile still. Maybe the addon should be removed or made non-functional by now.

apple502j:
Resolves #3861

### DIFF
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -1,3 +1,29 @@
+/**
+ Since presets can change independently of others, we have to keep track of
+ the versions separately. Current versions:
+
+ - editor-dark-mode 2 (bumped in v1.23 twice)
+ */
+
+const updatePresetIfMatching = (preset, settings, oldPreset, version) => {
+  if ((settings._version || 0) < version) {
+    /**
+     Version must be set even if transition is unnecessary;
+     1) User uses custom settings
+     2) User updates, transition aborts
+     3) User changes settings to old preset values
+     4) Without version, this change will revert after reload!
+     */
+    settings._version = version;
+    const map = {};
+    for (const key of Object.keys(oldPreset)) {
+      if (settings[key] !== oldPreset[key]) return;
+      map[key] = preset[key];
+    }
+    Object.assign(settings, map);
+  }
+};
+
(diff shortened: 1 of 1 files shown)


mxmou:
I think it would make sense to only check if settings related to the changed ones (such as accent background and workspace background, but not menu bar background, highlight color, or full screen background) have the same values as in the preset.

background/get-addon-settings.js
```diff
@@ -19,45 +45,60 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
         delete settings.editorCount;
         addonsEnabled["block-count"] = true;
       }
+      if (addonId === "editor-dark-mode") {
+        // Transition v1.22 to v1.23
+        // TurboWarp Dark preset changes:
+        updatePresetIfMatching(
+          manifest.presets.find((p) => p.id === "tw-dark"),
+          settings,
+          {
```
TheColaber, line 54: Listing the present settings here is probably a bad thing to do. You can get the data from the manifest.

TheColaber (changes requested):

background/get-addon-settings.js
WorldLanguages: Old presets are not available in the manifest.

apple502j:
@mxmou That'll work for Experimental Dark but not TW Dark (where only one setting actually changed).

background/get-addon-settings.js
TheColaber: Oh, I thought this was the new one

WorldLanguages:
Tested

PR #3905, opened 2021-12-09
Scrollable comments (new setting for comment-linebreaks addon)

### LINKED ISSUE #1870: Comment scroll
**Describe the new addon**
Some people compose very lengthy comments. Even with the 500-word limit, it is frustrating to read that lengthy comment while scrolling the entire page. This addon will make the comment div scrollable so that the whole page is not scrolled down.

Samq64:
Resolves #1870

### Changes

Adds a toggle to `comments-linebreaks` which adds a scroll bar to comments that are taller than the specified number of lines.

### Reason for changes

Some comments take up a lot of space when the addon is enabled.

### Tests

Tested on Edge 96.

image


RedGuy12:
Please enable actions here: https://github.com/Samq64/ScratchAddons/actions

Idea: make the comment height configurable?

Samq64:
> Please enable actions here: https://github.com/Samq64/ScratchAddons/actions

done.

> Idea: make the comment height configurable?

The problem is that if I make it an integer setting I would need to somehow append `px` to it with pure CSS or use JS.

RedGuy12:
calc(var(--settingName) * 1px)


addons/comments-linebreaks/addon.json
```diff
@@ -18,9 +18,43 @@
...
+      }
+    }
+  ],
+  "settings": [
+    {
+      "name": "Add scroll bars to tall comments",
+      "id": "scrollbars",
+      "type": "boolean",
+      "default": false
+    }
```
RedGuy12, line 38: Missing comma, that's why workflows are failing

addons/comments-linebreaks/addon.json
```diff
@@ -18,9 +18,43 @@
...
+          "scrollbars": true
+        }
+      }
     }
   ],
   "dynamicEnable": true,
   "versionAdded": "1.8.0",
+  "latestUpdate": {
+    "version": "1.23.0",
+    "newSettings": ["scrollbars"]
```
RedGuy12, line 57: And `"height"` too now

RedGuy12 (dismissed):

addons/comments-linebreaks/addon.json
Samq64: I noticed that a couple seconds before you reviewed.

Samq64:
Should I add myself to the credits?

RedGuy12:
You can if you want


Samq64:
image
It's probably another addon that's interfering, because that never happened before when only `comments-linebreaks` was enabled.

Samq64:
Moving the addon to the bottom of `addons.json` doesn't fix the overflowing text on profiles when Scratch 2.0 → 3.0 is enabled and I don't think there's an easy way to fix that since styling profile comments with CSS doesn't work, so I probbably can't use `!important`.

mxmou:
I'm working on some changes to Scratch 2.0 → 3.0 that should make this easier.

Samq64:
Could someone add the pending label to this?

Samq64:
The maximum height on profile comments is a bit off when `scratchr2` is enabled, but other than that, everything looks good.

mxmou:
@Samq64 I fixed the issue with 2.0 → 3.0.

Could the setting be enabled by default?

Samq64:
> @Samq64 I fixed the issue with 2.0 → 3.0.

Thanks!

mxmou (approved):

WorldLanguages:
There's a git conflict caused by #4828

Samq64:
Working on dynamic settings change.

Samq64:
I couldn't get dynamically enabling and disabling the scrollbar on scratch-www pages working because of `!important` but it's required to get the scrollbar to appear in the first place.

Secret-chest:
Why not 4em so it will always be 4 lines?

mxmou:
> Why not 4em so it will always be 4 lines?

The line height is more than 1em.

Secret-chest:
> > Why not 4em so it will always be 4 lines?
> 
> The line height is more than 1em.

But something em-based should be used.

WorldLanguages:
> I couldn't get `scratch-www` scrollbar dynamic enable/disable working because of `!important` but it's needed to get the scrollbar to appear in the first place.

Is this a blocker? Or can this be merged as-is?

WorldLanguages (approved):
LGTM
We should not merge, I'm not sure if what Samq64 mentioned is a blocker or not. I don't think I understood what the issue is
@mxmou Could you review again and merge if everything is OK?

mxmou:
@Samq64 `!important` is not the problem - `"updateUserstylesOnSettingsChange": true` needs to be added to the addon manifest to allow the setting to be dynamically enabled and disabled.

mxmou (approved):

PR #3907, opened 2021-12-11
Remove Scratch Addons ping from scratch-notifier

### LINKED ISSUE #3877: Remove SA notification sound from Scratch Notifier
Service workers do not allow playing audio on their own. We'll have to use the notification API's settings (system default, or none). Note that since Firefox does not support that part of the notification API, we'll have to hide that setting for Firefox users because there are no choices to be made.

Samq64:
Resolves #3877

### Changes

Removes `ping.mp3` from `scratch-notifier` and replaces the notification sound setting with a Boolean toggle.

This PR does not hide the setting for Firefox users.

### Reason for changes

> Service workers do not allow playing audio on their own.

Also, I don't think it's a very good idea to have different add-on options based on the browser which is why I removed it from both.

### Tests

Tested on Edge 96.


Samq64:
I know the transition code isn't great, I just wanted to get something that works before I made a pull request, feel free to change it.

background/get-addon-settings.js
```diff
@@ -100,6 +100,17 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
             delete scratchr2.linkColor;
           }
         }
+        if (addonId === "scratch-notifier") {
+          if (typeof settings.notification_sound !== "Boolean") {
```
apple502j, line 104:
```suggestion
          if (typeof settings.notification_sound !== "boolean") {
```

addons/scratch-notifier/addon.json
```diff
@@ -24,22 +21,8 @@
     {
       "name": "Notification sound",
```
RedGuy12, line 22: Rename to "Play sound"?

WorldLanguages:
We will be keeping it for now.

Samq64:
@WorldLanguages It looks like the non-breaking MV3 changes have the 1.28 milestone, so could you please re-open this?

apple502j:
We'll keep this for as long as we can.

WorldLanguages:
This is low prio.
I think we'll choose what to do about this depending on Firefox's implementation. It doesn't look like they are going to implement service workers for the background, but instead use "event pages", so `new Audio().play()` might still work there.

DNin01:
> This is low prio. I think we'll choose what to do about this depending on Firefox's implementation. It doesn't look like they are going to implement service workers for the background, but instead use "event pages", so `new Audio().play()` might still work there.

Since then, Firefox seems to be working on support for service workers, and it's probably not worth keeping a background page open just for this - they have some surprising differences from service workers.

PR #3909, opened 2021-12-11
fix(addon): `customize-avatar-border` - don't hide `Outline color` when `Fill transparent profile pictures` is on

CST1229:
Response to #3866

### Changes

In the "Customizable profile picture border" addon, also shows the "Outline color" option when the "Fill transparent profile pictures" option is on.

### Reason for changes

The "Fill transparent profile pictures" option also uses the "Outline color" option for its color, so it doesn't make sense to hide it when that option's on.

### Tests

Works.

### DIFF
--- a/addons/customize-avatar-border/addon.json
+++ b/addons/customize-avatar-border/addon.json
@@ -62,7 +62,7 @@
       "if": {
         "settings": {
           "hide-outline": false,
-		  "fill-transparent": true
+          "fill-transparent": true
         }
       }
     }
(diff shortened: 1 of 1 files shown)


apple502j:
note for future me: yes, this is fine.

- `if` in `userscripts` use AND, so `{"A":true, "B": false}` means A must be true and B must be false
- `if` in `settings` use OR, so so `{"A":true, "B": false}` means A must be true or B must be false

PR #3913, opened 2021-12-12
High res thumbnails: Downscaling + Special cases for larger-than-rendered dimensions

### LINKED ISSUE #3893: hi-res-thumbnails should scale images based on container size

If I enable `hi-res-thumbnails` and set the scaling to 100% I would expect all images across the site to be the same size as the container they're displayed in, but images that are displayed at high resolutions by default are untouched.

Make `hi-res-thumbnails` scale images based on container sizes instead of image sizes.

Make it togglable.

towerofnix:
Resolves #3893.

**Not a blocker for release,** but would be nice to have!

### Changes

Two major tweaks:

1. Downscale images when the default resolution is larger than it's actually displayed at.
2. Use arbitrarily determined dimensions as the basis for images which may dynamically be displayed larger than they currently are (only current case: studio project thumbnails).

~~**Still need to do this third change before worth merging:**~~ Done!

3. Auto-replace `projects.scratch.mit.edu` images with cdn2 480x360 images as a preprocessing step—this will generally be downscaled to whatever size the image is actually displayed at (e.g. on the homepage 145x110).

### Reason for changes

Save people's network data by not loading higher resolutions than needed! But also, in doing that, don't accidentally make images look bad (when dynamically displayed at larger than initial dimensions).

### Tests

Tested manually. To address the concerns I brought up [here](https://github.com/ScratchAddons/ScratchAddons/issues/3893#issuecomment-988213266):

> Make sure containers don't have smaller dimensions before the image is loaded. I.e, if a container is 200px by 0px before its content image has loaded, we don't want to scale down to that zero height dimension!

There don't seem to be any cases of this—Scratch specifically sets images to particular sizes so that the layout doesn't "blink" before the image loads.

> Make sure we're operating before images initially load, in the first place—downscaling is pointless if it only happens after the higher resolution has loaded.

We do! See `"runAtComplete": false`. I verified this by looking at the network log (with caching disabled) and made sure it wasn't loading duplicate images (i.e. not both the initial resolution and the modified one).

> We should deliberately replace images loaded from projects.scratch.mit.edu with corresponding cdn2 URLs, with the appropriate dimensions. Otherwise, we aren't saving nearly as much network data as we could be!

~~See changes section, this is a todo for this PR (hence marked draft).~~ Done!

### DIFF
--- a/addons/hi-res-thumbnails/userscript.js
+++ b/addons/hi-res-thumbnails/userscript.js
@@ -41,10 +41,7 @@ export default async function ({ addon, console }) {
     // account browser scaling or high-DPI screens. That's good, because we
     // want to leave the actual resolution-increase factor completely up to
     // the addon user/settings!
-    let {
-      width: widthRect,
-      height: heightRect
-    } = image.getBoundingClientRect();
+    let { width: widthRect, height: heightRect } = image.getBoundingClientRect();
 
     // If we've only got partial or no apparent displayed dimensions, and
     // width/height HTML attributes are provided, fall back to those. For some
--- a/addons/hi-res-thumbnails/userscript.js
+++ b/addons/hi-res-thumbnails/userscript.js
@@ -37,11 +37,25 @@ export default async function ({ addon, console }) {
   }
 
   function resizeToScreenRes(width, height, image) {
-    // Note this returns a value in CSS pixel units, i.e. it doesn't take into
+    // Note this returns values in CSS pixel units, i.e. it doesn't take into
     // account browser scaling or high-DPI screens. That's good, because we
     // want to leave the actual resolution-increase factor completely up to
     // the addon user/settings!
-    const rect = image.getBoundingClientRect();
+    let {
+      width: widthRect,
+      height: heightRect
+    } = image.getBoundingClientRect();
+
+    // If we've only got partial or no apparent displayed dimensions, and
+    // width/height HTML attributes are provided, fall back to those. For some
+    // reason, these attributes don't seem to always apply to the displayed
+    // image the instant it is added to the page DOM.
+    if (!widthRect || !heightRect) {
+      if (image.hasAttribute("width") && image.hasAttribute("height")) {
+        widthRect = +image.getAttribute("width");
+        heightRect = +image.getAttribute("height");
+      }
+    }
 
     // It's possible that the Scratch site is deliberately loading an image at
     // a greater resolution than it displays at - because with dynamic layout,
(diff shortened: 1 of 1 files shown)


towerofnix:
> There don't seem to be any cases of this—Scratch specifically sets images to particular sizes so that the layout doesn't "blink" before the image loads.

Found an exception: comment avatars on profiles, while loading, are apparently zero height. I'm not sure what causes this since they also have `width="45" height="45"`—I'll look into it!

addons/hi-res-thumbnails/userscript.js
```diff
@@ -41,10 +41,7 @@ export default async function ({ addon, console }) {
     // account browser scaling or high-DPI screens. That's good, because we
     // want to leave the actual resolution-increase factor completely up to
     // the addon user/settings!
-    let {
-      width: widthRect,
-      height: heightRect
-    } = image.getBoundingClientRect();
+    let { width: widthRect, height: heightRect } = image.getBoundingClientRect();
```
towerofnix, line 44: :angry:

towerofnix:
Okay, tested manually some more and browsed a bunch of pages across the site and all seems to be working. I fixed the issue with 0x0 images by falling back to `width`/`height`; now the only cases which could fail are those where said attributes are unset *and* no CSS is provided to automatically size the element. As far as I can tell, there don't appear to be any such conditions across the site—scratchr2 pages always set width/height and scratch-www uses ample CSS.

addons/hi-res-thumbnails/userscript.js
```diff
@@ -5,10 +5,29 @@ export default async function ({ addon, console }) {
...
+        src = `//cdn2.scratch.mit.edu/get_image/project/${id}_480x360.png`;
+      } else if (src.includes("users")) {
+        // Max user avatar size is 500x500.
+        src = `//cdn2.scratch.mit.edu/get_image/user/${id}_500x500.png`;
+      } else if (src.includes("galleries")) {
+        // Max studio thumbnail size is 500x500.
+        src = `//cdn2.scratch.mit.edu/get_image/gallery/${id}_500x500.png`;
+      }
+    }
+
```
towerofnix, line 30: I went with just rewriting `src` rather than applying values directly to `width` and `height` values because it makes reconstructing the URL quite a bit easier. This is more processing work than e.g. manually formatting the final width/height/id/type values (as we would need to), but it's minuscule—overall this only needs to run once for each (uploads) image, and at most there are 50 or those present total.

PR #3922, opened 2021-12-13
Lint 20211213

RedGuy12:
- Don't export-ignore LICENSE, we should ship it
- Mark locale files as generated code
- Run Prettier on Markdown and Yaml files
- Fix ESLint errors in `forum-id` adon


RedGuy12:
why not tho

PR #3923, opened 2021-12-13
Proper lint 20211213

apple502j:


addons/forum-id/userscript.js
```diff
@@ -1,3 +1,4 @@
+/* global $, paste */
```
RedGuy12, line 1: I did `window.paste` because this addon uses `window.paste` everywhere except that one line

PR #3948, opened 2021-12-18
Move `forum-quote-code-beautifier` to the bottom of `addons.json`

### LINKED ISSUE #3946: Customizable quotes and code blocks addon is not compatible with other themes
`forum-quote-code-beautifier` doesn't color properly anymore when `scratchr2` is also enabled (the border is always gray and the background is the same gray), which might be because of their order in `addons.json`.

_Originally posted by @CST1229 in https://github.com/ScratchAddons/ScratchAddons/issues/3938#issuecomment-996921777_

mxmou:
Resolves #3946

### Changes

Makes `forum-quote-code-beautifier` work when Scratch 2.0 → 3.0 or website dark mode is enabled.


CST1229:
Won't this make the "getting darker with nested quotes" effect no longer be a thing?

PR #3952, opened 2021-12-19
Refactor messaging system and remove persistent scripts

### LINKED ISSUE #1534: Persistent scripts: use chrome.alarms internally for setTimeout/setInterval/clearTimeout/clearInterval
Manifest v3 will use service workers instead of persistent scripts. setInterval/setTimeout are not honored in non-persistent background scripts if they trigger when the event page is dormant, instead, `chrome.alarms` should be used (source: https://developer.chrome.com/docs/extensions/mv2/background_migration/#timers). We can use the fact persistent scripts already get fake setTimeout/setInterval functions to abstract the complexity from addons, so that they can still use these methods, even tho the addon loader is internally using `chrome.alarms`.

### LINKED ISSUE #1662: Manifest v3: do not use XMLHttpRequest from background page / service worker
`scratch-messaging` uses XMLHttpRequest to send and delete comments. 

https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/scratch-messaging/background.js#L233

### LINKED ISSUE #2703: scratch-messaging should not have background script
see also #1662

if we truly need to send the request from background, we can just make a proxy script that only triggers from our own popups.

### LINKED ISSUE #3764: Allow messaging to work with multiple accounts.

I have 2 accounts and inorder to check my messages on both I am constantly swapping between them, each time I have to login so the messaging addon doesn't help much. 

I want the messaging api to automatically sign me out of my current account into my alternate account, check for messages, repeat for all other alternate accounts and then sign back into the original account. Then I'd like these messages to be shown in the messaging tab, with containers for each account.

it would also need a section in the addon to add your username and password for each account so that it can work properly.

Currently I just sign in and out of my accounts but that is slow and the whole purpose of the messaging api is to notify me when I get new messages.
I looked through the issues and i couldn't find an issue aiming to solve this specific problem.

### LINKED ISSUE #4042: [scratch-messaging]: Links in ST Alert do not go to scratch.mit.edu
This screen recording by a user in the Scratch Addons server showcases the bug:

https://user-images.githubusercontent.com/72760579/148692342-86579e2d-20ed-45dd-9898-793c9cd32138.mov

apple502j:
Apple does a big refactor again. This time, scratch-messaging.

Resolves #1534
Resolves #1662
Resolves #2703
Resolves #3764
Resolves #4042

User-facing changelog:
- Message counts in the badge will be updated every minute instead of every 30 seconds
- New messages will be notified every 5 minutes instead of every 2 minutes
- Error message has been added to indicate server issues and network problems in Scratch Messaging
- Performance improvements in Scratch Messaging when checking for project or studio comments
- Fixes a bug where Scratch Addons ping sound of Scratch Notifier played while muted
- Fixes a bug where Scratch Addons ping sound of Scratch Notifier played too many times and lagged the browser
- Scratch Messaging now works in Firefox incognito mode (Chrome incognito popup will still use the default profile, and you cannot use fullscreen in Chrome incognito)
- Scratch Messaging now works in tab mode in Firefox Containers
- Scratch Notifier notification now works after browser restart on Chrome, if Chrome was open at the time it was clicked (Does not work in Firefox, or when the browser was not open, due to browser restrictions)
- Mentioning welcomingcommittee in project/studio comment now links to the studio in Scratch Messaging

Technical changelog:
- Persistent scripts are removed completely. `persistentScripts` field in `addon.json` is unused.
- `Account` addon API is now userscript-only. `Account#getMessages` and `Account#clearMessages` are now removed, as it was only available to persistent scripts. To get messages from the API, use `MessageCache#fetchMessages`, or read from the IndexedDB (if it's not incognito). To clear the messages, use `MessageCache#markAsRead`, and if necessary update the badge using `BackgroundMessageCache#updateBadge`.
- `{forceBadgeUpdate: {store: "cookie store ID"}}` can be sent via IPC to force badge update.
- `MessageCache#updateMessages` can be used to fetch messages that are not yet seen by the extension. The return value should always be passed to Scratch Notifier. For IPC use `{notifyNewMessages: {store: "cookie store ID", messages: []}}`.
- Messages are cached in the IndexedDB. (In Firefox incognito, IndexedDB is unavailable; therefore, mock implementation is used, and the messages don't get cached.) The database table `messaging` has three object stores: `cache`, `count`, and `lastUpdated`. These are all keyed by the cookie store ID (always `0` in Chrome and `firefox-default`/`firefox-containers-X` in Firefox). `cache` stores the array of cached messages (latest first), `count` stores unread messages count (incl. alerts), and `lastUpdated` is when the last fetching occured in ms.
- Cached messages are invalidated 12 hours after `lastUpdated`, and instantly if session changed. Browser restart does not invalidate the cache by itself.
- `MessageCache#openMessageCache` initializes the database and performs the cleanup.

Roadmap:
- [x] perform API request from popup
- [x] addon.self.getEnabledAddons should work in popup
- [x] move fixCommentContent to popup
- [x] use new API wrapper
- [x] add way to render raw DOM as Vue component
- [x] use that to render comments instead of innerHTML
- [x] popup auth should inherit parent window, should allow #3764 combined with Firefox Container
- [x] fetch comments on popup
- [x] fetch messages on popup
- [x] cache messages on indexedDB
- [x] refactor notifier to use new cache
- [x] hardcode msg-count-badge
- [x] ~~add error for trying to use it on incognito - doesn't work because Firefox disabled IDB there~~
- [x] create ephemeral database for incognito
- [x] work around https://bugzilla.mozilla.org/show_bug.cgi?id=1747283
- [x] remove persistent script loading
- [x] remove addon API for badge and notifications
- [x] figure out whether XHR bug for Chrome is fixed or not - it's fixed from what I can tell


addon-api/content-script/Auth.js
```diff
@@ -1,11 +1,10 @@
...
-    super(...args);
-    this._refresh();
-  }
+import FetchableAuth from "../common/FetchableAuth.js";
 
+/**
+ * Authentication related utilities.
+ * @extends FetchableAuth
+ */
+export default class Auth extends FetchableAuth {
```
RedGuy12, line 7: It looks like this is empty except what's inherited from FetchableAuth now
..why?

RedGuy12: Oh I see, _getCookie is different in each subclass. Please add _getCookie to FetchableAuth and have it throw a "Subclasses must implement this" error.

RedGuy12: Oh I see, _getCookie is different in each subclass. Please add _getCookie to FetchableAuth and have it throw a "Subclasses must implement this" error.

addons/scratch-messaging/background.js
```diff
@@ -1,14 +1,8 @@
-import commentEmojis from "../scratch-notifier/comment-emojis.js";
-import { linkifyTextNode, pingifyTextNode } from "../../libraries/common/cs/fast-linkify.js";
-import formatProfileComments from "../../libraries/common/cs/format-profile-comments.js";
-
 export default async function ({ addon, global, console, setTimeout, setInterval, clearTimeout, clearInterval }) {
```
RedGuy12, line 1: I thought we wanted to remove this file?

webpages/settings/index.html
```diff
@@ -269,7 +269,7 @@ <h1 class="setting-heading">{{ msg("moreSettings") }}</h1>
           </p>
           <p>
             <a
-              href="./licenses.html?libraries=icu-message-formatter,vue,color-picker-web-component,comlink,Sora,fuse"
+              href="./licenses.html?libraries=icu-message-formatter,vue,color-picker-web-component,comlink,Sora,fuse,idb"
```
RedGuy12, line 272: Will it be used in settings or scratch messaging?

apple502j:
Please do not review draft PRs, it's WIP

apple502j:
Updates:

- incognito support will be missing/be partial. We can't use IndexedDB there (both technically and for privacy reasons), so on every page load all messages will be fetched, assuming I made it work in the first place.
- Firefox Container browserAction popup doesn't exist. It's default or incognito.
- You can open fullscreen popup in the container, but due to a bug, the code currently recognizes it as a default non-containered state despite being so logically. I might add a workaround for this.

WorldLanguages:
Tested dismissing an alert: works

apple502j:
Self-merging (with permission)

PR #3955, opened 2021-12-19
Move project player addons to a top-level category

### LINKED ISSUE #3014: Separate project player addons from editor addons

On the settings page, the project player category is a subcategory of Scratch Editor Features, but the project player is a part of both the website and the editor.

Make it a separate top-level category.

Samq64:
Resolves #3014

### Changes

Moves project player addons to a new top-level category under the Scratch Editor Features category. Since other the other top-level category tags are short, the `projectPlayer` tag has been renamed to just `player`.

### Reason for changes

It's confusing for project player addons to be a subcategory of editor addons since those addons also affect the project page and embeds.

### Tests
Tested on Edge 96.


webpages/settings/data/tags.js
```diff
@@ -44,6 +44,14 @@ export default [
       theme: true,
     },
   },
+  {
+    name: "forPlayer",
+    matchName: "player",
+    color: "darkgreen", // TODO: use a new tag color
```
RedGuy12, line 50: reminder

WorldLanguages:
@mxmou Do you like this? I'm not convinced. (asking you since you made the issue that suggested this)

webpages/settings/data/tags.js
Samq64: I'll do that later today or tomorrow.

Samq64:
There's only one theme in the project player themes, so maybe it would be better to get rid of the project player themes category completely.

WorldLanguages:
Honestly I don't really like this. IMO the project player is part of the editor.

mxmou:
> Honestly I don't really like this. IMO the project player is part of the editor.

It's also part of the website.

apple502j:
Manifest: ScratchAddons/manifest-schema#56

TheColaber:
In my opinion, I agree with this being a separate category. The player which can be found in 3 places (projects/*/, projects/*/editor, projects/*/embed). I don't think that to change something in an embed, you have to look under the editor section. And @mxmou argued that this could technically also be a website addon since you can find it there too. I think that when most people think of an "editor" addon, they think of scripts that can run on projects/*/editor. However, all of these scripts that run, run on "projects", which are all 3, project page, project editor, and project embed. This is why I think they should have their own category.

WorldLanguages:
I'm not convinced this should be merged, but I was more skeptical a few months ago.

Some issues I see:
1. There's less than 10 `projectPlayer` addons. And I'm not sure if clones and mouse-pos should be counted, because they only appear when inside the editor, and aren't helpful when viewing someone else's project.
2. The "Scratch Editor Features" would have 3 subcategories instead of four, making it less consistent with Scratch Website Features.
3. The sidebar of the settings page is already filled on low resolution screens. Adding one more button would hide the "more settings" button for some users.

mxmou:
> 1. There's less than 10 `projectPlayer` addons. And I'm not sure if clones and mouse-pos should be counted, because they only appear when inside the editor, and aren't helpful when viewing someone else's project.

Clones and mouse-pos are editor addons. I don't think this is an issue - there aren't many popup addons either.

> 2. The "Scratch Editor Features" would have 3 subcategories instead of four, making it less consistent with Scratch Website Features.

Why is that important? More subcategories can be added later if there are enough addons for the same part of the editor.

> 3. The sidebar of the settings page is already filled on low resolution screens. Adding one more button would hide the "more settings" button for some users.

The category buttons are very big - reducing all of them to the size of "small" ones (Credits, Help Translate, Send Feedback) would add enough space for adding another button. The header could be slightly smaller too.

WorldLanguages:
Sorry, I forgot to reply.

> Clones and mouse-pos are editor addons. I don't think this is an issue - there aren't many popup addons either.

I think I created the popup category because having "message count in extension icon" as a website addon was too weird. It did involve the Scratch website in a way, because the message count is taken from there, but in no way it modifies the Scratch website and it doesn't even link to it.
Having project player addons inside editor addons isn't as weird as that.

> Why is that important? More subcategories can be added later if there are enough addons for the same part of the editor.

Okay, good point. We have tags such as `comments` but they aren't shown as different subcategories because there's not enough addons with those tags.

> The category buttons are very big - reducing all of them to the size of "small" ones (Credits, Help Translate, Send Feedback) would add enough space for adding another button. The header could be slightly smaller too.

That could work.

WorldLanguages:
Am I missing any addon that could fit into this category? I previously said clones and mouse-pos are targeted for developing a project, not for watching projects, so those don't really fit. Also, please see what I say at the end about remove-curved-stage-border.

- 60fps
- fullscreen
- gamepad
- mute-project
- pause
- remove-curved-stage-border (would this addon stay as a theme? would we need to create a "themes for project player" subcategory? this one is very confusing)

webpages/settings/data/tags.js
Samq64: I think it's been decided there'll be no theme sub-category, so I'll just remove the tag completely.

Samq64:
Done. By the way, the project player themes sub-category already existed, but I removed it.

mxmou:
Added "pending" because the spacing of the category selector needs to be reduced first.

webpages/settings/data/tags.js
Samq64: I re-added the theme category, but now #4989 is open and if it gets merged, the colour of all theme tags will be the same.

Samq64:
Is this still pending discussion or can the pending label be removed since the padding has been reduced?

mxmou:
@Samq64 The pending label was for padding.

GrahamSH-LLK:
Is this still under development?

Samq64:
I don't think we got to a conclusion on if this will be accepted or not.

WorldLanguages:
There might have been a misunderstanding here: (does this PR remove the padding on the buttons or not?)
image

mxmou:
@WorldLanguages It does reduce the padding values in `category-selector.html`.

Samq64:
> There might have been a misunderstanding here: (does this PR remove the padding on the buttons or not?)

Yes it does: https://github.com/ScratchAddons/ScratchAddons/pull/3955/commits/122998b1ef50d16325f3c66add83bd71d90b4146

DNin01 (changes requested):
If the plan is to make this change, we will need to update this pull request so the latest addons such as `vol-slider` are categorized correctly.

DNin01 (approved):
I think we're all set.

addons/clones/addon.json
```diff
@@ -47,6 +47,6 @@
     "version": "1.31.0",
     "newSettings": ["projectpage"]
   },
-  "tags": ["editor", "projectPlayer", "recommended"],
+  "tags": ["editor", "recommended"],
```
iqnite, line 50: Clone counter works on the website, too.
```suggestion
  "tags": ["player", "recommended"],
```

addons/mouse-pos/addon.json
```diff
@@ -21,6 +21,6 @@
       "matches": ["projects"]
     }
   ],
-  "tags": ["editor", "projectPlayer", "recommended"],
+  "tags": ["editor", "recommended"],
```
iqnite, line 24: Same here. You should consider that someone might add website support in the future.
```suggestion
  "tags": ["player", "editor", "recommended"],
```

iqnite (changes requested):
Looks good! Could you take a look at the comments I've added? I don't know if it's intentional, but the addons that already had the "projectPlayer" tag shouldn't lose the "player" tag, because users might look for them in the "Player" category and wouldn't find them. Remember that addons can belong to multiple categories.

addons/mouse-pos/addon.json
Samq64: According to the [addon manifest docs](https://scratchaddons.com/docs/reference/addon-manifest/#required-factual-tags) addons can only have one top-level category tag.

addons/clones/addon.json
Samq64: It does now, but I think it's disabled by default outside the editor. I agree it should have the tag though.

addons/mouse-pos/addon.json
iqnite: > According to the [addon manifest docs](https://scratchaddons.com/docs/reference/addon-manifest/#required-factual-tags) addons can only have one top-level category tag.

Really? Oh, sorry then.

iqnite: Clone counter still needs the `player` tag.

iqnite (approved):
Neat!

Samq64:
This is still on the v1.36 milestone but Vue 3 is on v1.38. Is that the case or is this one also being pushed back to 1.38?

Samq64:
Is this still planned for 1.37? Both Mxmou and DNin previously approved.

WorldLanguages:
Hmm, I'll be looking at this after the MV3 release. Sorry.

mxmou (approved):

Samq64:
@Hans5958 For the manifest schema does it need to be a new version or do I edit the existing one?

DNin01:
> @Hans5958 For the manifest schema does it need to be a new version or do I edit the existing one?

I'm pretty sure we should bump the schema to a new version. The last schema hasn't changed for a few releases.

PR #3961, opened 2021-12-20
Hide inputs when custom block middle clicked

TheColaber:
Resolves TurboWarp/scratch-gui#390

### Changes

Blockly.hideChaff(), which scratch also uses for gestures: https://github.com/LLK/scratch-blocks/blob/893c7e7ad5bfb416eaed75d9a1c93bdce84e36ab/core/gesture.js#L483

I also import the devtools code to the debugger addon since they use the exact same code for going to a block and blinking it.


addons/editor-devtools/blockly/Utils.js
```diff
@@ -56,9 +59,8 @@ export default class Utils {
   /**
    * Based on wksp.centerOnBlock(li.data.labelID);
    * @param blockOrId {Blockly.Block|{id}|BlockInstance} A Blockly Block, a block id, or a BlockInstance
-   * @param [force] {boolean} if true, the view always moves, otherwise only move if the selected element is not entirely visible
```
WorldLanguages, line 59: Note: this was removed because it was unused
@TheColaber Correct?

WorldLanguages:
Using the block flasher from devtools in debugger makes code cleaner but I want to double check whether we introduced some different logic for debugger and forgot

WorldLanguages:
Devtools' `scrollBlockIntoView` was there since before the debugger addon existed, so that makes me wonder, why didn't we import the module from the beginning? 
Also, I thought the whole block flasher script was an entire chunk of code found as-is on devtools, but at the time the debugger 
addon was developed, it was already split into smaller methods in devtools!
Here's the commit: https://github.com/ScratchAddons/ScratchAddons/pull/2161/commits/6164eeb77ca857d333ebf818171438d4c5117776

addons/editor-devtools/blockly/Utils.js
TheColaber: yea

TheColaber:
>  why didn't we import the module from the beginning?

the state of the devtools addon code at that time could not be a simple import. The best way was to copy the code and leave a comment.

WorldLanguages:
> the state of the devtools addon code at that time could not be a simple import.

That's what I thought, but it looks like that's not the case? Go to the commit I linked to, browse files, addons→editor-devtools→blockly and there it is, BlockFlasher.js!

addons/debugger/userscript.js
```diff
@@ -307,65 +308,8 @@ export default async function ({ addon, global, console, msg }) {
...
-      count--;
-      if (count > 0) {
-        myFlash.timerID = setTimeout(_flash, 200);
-      } else {
-        myFlash.timerID = 0;
-      }
-    }
-
-    _flash();
+    const { scrollBlockIntoView } = new DevtoolsUtils(addon);
```
apple502j, line 311: @TheColaber Why assign?

apple502j (approved):
LGTM, but has one question. (note to future self: reusing DevtoolsUtils could have issue on re-entering, so this code is correct)

TheColaber:
> That's what I thought, but it looks like that's not the case? Go to the commit I linked to, browse files, addons→editor-devtools→blockly and there it is, BlockFlasher.js!

Well, I guess we didn't realize it.

PR #3968, opened 2021-12-21
Add .vscode/settings.json for json schema completion

RedGuy12:
Resolves none

### Changes

added VSC settings file

### Reason for changes

this adds automatic autocomplete and checking to all json files in this repo (only in vsc tho) without the need to specify $schema everywhere.

### Tests

yup


TheColaber:
You should also export ignore this, and pls stop creating branches on this repo 😭

RedGuy12:
> You should also export ignore this

all files and folders beginning with a `.` are already ignored

Weredime:
@RedGuy12 when

.vscode/settings.json
```diff
@@ -0,0 +1,53 @@
...
+        "additionalProperties": { "type": "string" }
+      }
+    },
+    { "fileMatch": ["/manifest.json"], "url": "https://json.schemastore.org/chrome-manifest.json" }
+  ],
+
+  "explorer.fileNesting.enabled": true,
+  "explorer.fileNesting.patterns": {
+    ".eslintrc.json": ".*",
+    "/webpages/settings/*.html": "/webpages/settings/${capture}.html"
```
RedGuy12, line 42: fsr this one doesn't want to work, no idea why

.vscode/settings.json
```diff
@@ -0,0 +1,53 @@
+{
+  "json.validate.enable": true,
+  "json.schemaDownload.enable": true,
+  "json.schemas": [
+    {
+      "fileMatch": ["/_locales/*/messages.json"],
+      "url": "https://raw.githubusercontent.com/RedGuy12/chrome-i18n-json-schema/master/schema.json"
```
TheColaber, line 7: This should be changed and then this can be merged

.vscode/settings.json
```diff
@@ -39,7 +39,7 @@
   "explorer.fileNesting.enabled": true,
   "explorer.fileNesting.patterns": {
     ".eslintrc.json": ".*",
-    "/webpages/settings/*.html": "/webpages/settings/${capture}.html"
+    "*.html": "${capture}.*"
```
WorldLanguages, line 42: Can you send a screenshot of example files that are grouped?
I remember considering grouping our pseudo-vue components like this, and discarding the idea because the UI/UX wasn't good enough.

RedGuy12: @thecolaber I didn't do it that way because then it'd nest all the files in the popup directories together which makes people have to open up another level of nesting.

TheColaber: That's a consistency issue. Popups and webpages should have the same type of folder system.

RedGuy12: So the current structure is:
```
webpages
- popups
  - cloud-games
    - index.html
    - index.js
    - index.css
  - scratch-messaging
   - index.html
   - index.js
   - index.css
```

We should change it to
```
webpages
- popups
  - cloud-games.html
  - cloud-games.js
  - cloud-games.css
  - scratch-messaging.html
  - scratch-messaging.js
  - scratch-messaging.css
```

apple502j: Popups are independent of each other, just like addons; 👎🏼 to this.

mxmou: Where would `popups/scratch-messaging/api.js` go?

GrahamSH-LLK:
I assume this is no longer under development?

WorldLanguages:
@GrahamSH-LLK Some minor things are blocked because of discussion. I think we should open another PR that creates a `.vscode` folder but we shouldn't attempt to set every setting at once. We can start with smaller things, like excluding translation strings from project search, and adding the manifest schema.

WorldLanguages:
I'm closing this as #6100 implemented a subset of these settings :)

Created issue #6440 to track some of the suggestions that this PR proposed

Issue #2599, opened 2021-05-24
Browser support discussion


WorldLanguages:
EDIT: THIS IS ALL OUTDATED

Let's keep 1 year of Chrome support, but reduce Firefox support in half to 6 months. This is because of Chromebooks which can't be updated, and the upcoming Manifest V3 update we're required to do.
I'm unsure if we're going to be able to keep the 1 year of Chrome support rule when we are required by the Chrome Web Store to migrate to manifest v3, but lets try I guess.

**Old requirements: Chrome 80+, Firefox 74+
New requirements: Chrome 87+ (Chrome 87+, Edge 87+, Opera 73+, Brave 1.19+, Vivaldi 3.5+), Firefox 89+**

We would need to kick some users out even tho most features will still work (unlike the update where we added minimum browser requirements, where the majority if not all users that were kicked out could not access settings page already because optional chaining JavaScript dyntax was unsupported), so we should also have a hidden way for users to try to bypass the new requirements through some hidden link and `chrome.storage.local`. Note that we should only do this if we don't add "breaking" changes, like dynamic imports on content scripts on Firefox. No reason to let users bypass if extension won't work anyway.

Chrome 87 (November 2020) vs Chrome 80 (February 2020):
- `String.prototype.replaceAll`
- `inset-inline-start` and `inset-inline-end`
- MV2 extensions can use background service workers (probably useless because Chrome 87 did not support module service workers - that's Chrome 92+ apparently, 91+ for normal non-extension service workers tho, and no dynamic import [source](https://stackoverflow.com/questions/66406672/chrome-extension-mv3-modularize-service-worker-js-file))
- CSS `revert` property
- Logical nullish assignment (`??=`) syntax support
- Not supported: `options.signal` parameter on addEventListener not yet supported, Chrome 90+
- Not supported: top level await, Chrome 89+

Firefox 89 (May 2021) vs Firefox 74 (March 2020):
- `String.prototype.replaceAll`
- `inset-inline-start` and `inset-inline-end` were already supported
- Probably no support for background service workers
- CSS `revert` property was already supported
- CSS `min()` and `max()` functions (some addons already use these, and Firefox 75+ supports these functions)
- DTF dateStyle and timeStyle (some addons already use these, Firefox 79+)
- Full `<link rel="preload">` support
- `options.signal` parameter on addEventListener
- `tabs` permission no longer required when host permissions are already present
- Dynamic import in content scripts (was blocker, Chrome has supported for a while)
[...]

PR #3979, opened 2021-12-21
[WIP] Split cs.js into modules, clean up content scripts, avoid polluting scope

WorldLanguages:
Because of #2599 we can now use dynamic import on content scripts.
However, dynamic import will slow down how fast code runs, so we need to be careful and only use it for code that isn't time sensitive.

Goals of this PR:
- Shrink the size of cs.js by moving non critical parts to other files. For explanation on why it was so big, see below; we intentionally made it a single file because separate content scripts would share variable scoping even if we use let/const, and not even ESLint would be able to catch potential errors there.
- Avoid content scripts other than cs.js from unintentionally polluting the global extension scope. All content scripts share the same scope unless they're dynamically imported modules, so `if(){}` blocks and IIFEs fix this. ESLint isn't smart enough to know this, so the easiest way to deal with this is to simply make sure only cs.js registers variables and functions at top level.
- Avoid code duplication

This PR will involve some copy pasting of code around, so I'll make sure to clearly mark code that has been changed after copy-pasting with `// comments like this //`

Because of Chrome, we need to add modules that we want to dynamically import on `web_accessible_resources`, even tho they run on extension context. Firefox does not have this limitation.

### DIFF
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -12,7 +12,7 @@ scratchAddons.eventTargets = {
 scratchAddons.session = {};
 const consoleOutput = (logAuthor = "[page]") => {
   const style = {
-    // Remember to change these as well on cs.js
+    // Remember to change these as well on cs-console.js
     leftPrefix: "background:  #ff7b26; color: white; border-radius: 0.5rem 0 0 0.5rem; padding: 0 0.5rem",
     rightPrefix:
       "background: #222; color: white; border-radius: 0 0.5rem 0.5rem 0; padding: 0 0.5rem; font-weight: bold",
--- a/content-scripts/comment-filter.js
+++ b/content-scripts/comment-filter.js
@@ -1,0 +1,195 @@
+
+// //
+import { escapeHTML } from "../libraries/common/cs/autoescaper.js";
+const DOLLARS = ["$1", "$2", "$3", "$4", "$5", "$6", "$7", "$8", "$9"];
+let initialUrl = location.href;
+let path = new URL(initialUrl).pathname.substring(1);
+if (path[path.length - 1] !== "/") path += "/";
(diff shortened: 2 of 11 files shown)


content-scripts/inject/module.js
```diff
@@ -12,7 +12,7 @@ scratchAddons.eventTargets = {
 scratchAddons.session = {};
 const consoleOutput = (logAuthor = "[page]") => {
   const style = {
-    // Remember to change these as well on cs.js
+    // Remember to change these as well on cs-console.js
```
RedGuy12, line 15: cant we make this an external lib already

WorldLanguages: Good point, we now could

FunctionalMetatable:
tbh one day it might be a good idea to bundle scripts all into one using rollup/webpack

WorldLanguages:
> tbh one day it might be a good idea to bundle scripts all into one content scripts using rollup/webpack

why?

FunctionalMetatable:
> > tbh one day it might be a good idea to bundle scripts all into one content scripts using rollup/webpack
> 
> why?

we can load addons faster

WorldLanguages:
> we can load addons faster

It won't make a noticeable difference.

FunctionalMetatable:
> > we can load addons faster
> 
> It won't make a noticeable difference.

but its better for addons like scratchblocks and stuff

WorldLanguages:
Dynamic import support for content scripts was added to Firefox since version 89, and we decided to support Firefox 86+ so this can't be done.
Also, there can be speed issues and flashes if themes like dark mode and scratch2.0→3.0 aren't added fast enough, and using dynamic import is probably slower, so I don't think we should use dynamic import() for the text-color library.

PR #3983, opened 2021-12-22
Remove swap variables in sprite

### LINKED ISSUE #3772: Move swap variables to swap-local-global

`editor-devtools` is too big and most of the things it does aren't even togglable.

> If we ever add a way to switch whether a variable is global or local after being created, it could be merged with this.

_Originally pasted by @WorldLanguages in #3700._

This got released as a new addon in v1.22, so it would make sense to move the swap variables option out of Editor devtools now.

Moving swap variables from `editor-devtools` to `swap-local-global` as a toggle.

Add a toggle to the Editor devtools addon.

### Additional context

Related: #3700

I should have brought this up before 1.22 was released.

Samq64:
Closes #3772

### Changes

Removes the swap variables in sprite context menu option from `editor-devtools`.

### Reason for changes

I'm pretty sure most people who swapped variables used it mostly for switching between local and global variables whitch can be done more efficiently with `swap-loclal-global`. It also means that devtools adds one less context menu option that can't be removed unless devtools is disabled. If people actually do use the swap variables feature for other purposes, it could just be added back as a toggle for `swap-local-global`.

### Tests

Tested on Edge 96. I might have missed some dead code because I didn't search it very well.


lisa-wolfgang:
I don't think this should be completely deleted. While it's true that this was often used for scope swapping, it's also helpful for refactoring purposes.

WorldLanguages:
lol whoops 👀

Mbrick2:
maybe just separate it. I also use it if i have 2 varibles and i need to replace one with the other you can use that

PR #3985, opened 2021-12-23
turbowarp-button: add setting to automatically open/replace with TurboWarp

### LINKED ISSUE #5480: Add an Option to Set the Turbowarp Player as Default in `Turbowarp button` - `Replace player` Option
### Why this would be helpful

Every time I reload the page, the player switches to Scratch player.
If there was an option to set the Turbowarp player by default, I don't need to click the button and let the Turbowarp player reload again.

### My suggestion

Add an option to set the Turbowarp player by default.

### Possible alternatives

_No response_

### Additional context

_No response_

lisa-wolfgang:
Resolves #5480

### Changes

A new setting has been added. When enabled, as soon as the addon loads, the TurboWarp player will immediately either open in a new tab or replace the Scratch player based on the `action` setting. This new setting is disabled by default.

Even when this setting is disabled, the TurboWarp button is now able to be toggled before the project finishes loading.

Also fixed a bug where the button would display the incorrect mode after exiting the editor.

### Reason for changes

For frequent users of TurboWarp, it is time-consuming to wait for large projects to load before using the TurboWarp button. In some cases, it may actually be faster to just copy the link and open TurboWarp manually, defeating the whole purpose of this addon.

### Tests

Tested in browser w/ Chromium 96 and Firefox 95.0.2.

One potential further optimization is blocking Scratch from loading the regular project while the TurboWarp project is loading. However, the user is already loading two projects in succession with the current addon, so this shouldn't be a merge blocker, in my opinion.

### DIFF
--- a/addons/turbowarp-player/addon.json
+++ b/addons/turbowarp-player/addon.json
@@ -22,7 +22,8 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["projects"]
+      "matches": ["projects"],
+      "runAtComplete": false
     }
   ],
   "userstyles": [
@@ -49,6 +50,12 @@
       "default": "link"
     },
     {
+      "name": "Automatically open/replace",
+      "id": "auto",
+      "type": "boolean",
+      "default": false
+    },
+    {
       "name": "Enable addons",
       "id": "addons",
       "type": "boolean",
@@ -59,5 +66,9 @@
     }
   ],
   "versionAdded": "1.18.0",
+  "latestUpdate": {
+    "version": "1.24.0",
+    "newSettings": ["auto"]
+  },
   "tags": ["community", "projectPage", "featured"]
 }
--- a/addons/turbowarp-player/userscript.js
+++ b/addons/turbowarp-player/userscript.js
@@ -16,7 +16,12 @@ export default async function ({ addon, console, msg }) {
 
   const button = document.createElement("button");
   button.className = "button sa-tw-button";
-  button.title = "TurboWarp";
+  if (addon.settings.get("auto")) {
+    button.title = "Scratch";
+    button.classList.add("scratch");
+  } else {
+    button.title = "TurboWarp";
+  }
 
   function removeIframe() {
     twIframeContainer.remove();
@@ -26,7 +31,7 @@ export default async function ({ addon, console, msg }) {
     button.title = "TurboWarp";
   }
 
-  button.onclick = async () => {
+  const buttonAction = async () => {
     if (action === "player") {
       playerToggled = !playerToggled;
       if (playerToggled) {
@@ -40,6 +45,7 @@ export default async function ({ addon, console, msg }) {
         }
         const iframeUrl = `https://turbowarp.org/${projectId}/embed?${usp}`;
         twIframe.src = "";
+        scratchStage = await addon.tab.waitForElement(".guiPlayer");
         scratchStage.parentElement.prepend(twIframeContainer);
         // Use location.replace to avoid creating a history entry
         twIframe.contentWindow.location.replace(iframeUrl);
@@ -54,6 +60,10 @@ export default async function ({ addon, console, msg }) {
     }
   };
 
+  if (addon.settings.get("auto")) buttonAction();
+
+  button.onclick = buttonAction;
+
   let showAlert = true;
   while (true) {
     const seeInside = await addon.tab.waitForElement(".see-inside-button", {
@@ -62,9 +72,16 @@ export default async function ({ addon, console, msg }) {
     });
 
     seeInside.addEventListener("click", function seeInsideClick(event) {
-      if (!playerToggled || !showAlert) return;
+      if (!playerToggled) return;
+      if (!showAlert) {
+        button.title = "TurboWarp";
+        button.classList.remove("scratch");
+        return;
+      }
 
       if (confirm(msg("confirmation"))) {
+        button.title = "TurboWarp";
+        button.classList.remove("scratch");
         showAlert = false;
       } else {
         event.stopPropagation();


WorldLanguages:
IIRC we were going to do this but decided against it

lisa-wolfgang:
UI King thinks it's fine because the user would have to toggle that setting on manually (and read it) instead of toggling the whole addon without reading and then suddenly having the TurboWarp player replace the Scratch player (which would actually require a second setting change.)

mxmou:
> For frequent users of TurboWarp, it is time-consuming to wait for large projects to load before using the TurboWarp button. In some cases, it may actually be faster to just copy the link and open TurboWarp manually, defeating the whole purpose of this addon.

If that's the only reason, wouldn't showing the button sooner be a better solution?

lisa-wolfgang:
I'm not entirely clear on the drawbacks of having the TurboWarp player load first, especially if the user has specified via this new setting that they'd like to always load TurboWarp first. It would be a major convenience to frequent TurboWarp users to never even have to click a button.

GrahamSH-LLK:
This seems implemented but waiting on discussion and merge, is that correct?

Samq64:
I think the setting should only apply to replace mode since a lot of browsers block opening tabs if they weren't activated by the user.

addons/turbowarp-player/addon.json
```diff
@@ -48,6 +49,12 @@
       ],
       "default": "link"
     },
+    {
+      "name": "Automatically open/replace",
+      "id": "auto",
+      "type": "boolean",
+      "default": false
+    },
```
DNin01, line 57: As Samq64 said, I think we may need to consider that people may have set the addon to open a new tab. How would that work out? It's intrusive to always open a new tab whenever you visit a project, especially if you just want to edit it or see a comment someone left.

DNin01 (changes requested):
In any case, it seems strange to always stray away from Scratch when opening projects on the Scratch website, especially those that didn't have TurboWarp in mind, which is most projects. It's also not ideal to not be able to pick up where you left off if you decide you want to open the editor, or the fact that you have to manually re-enter the TurboWarp player when you leave the editor. The experience is not that fun for someone who wants to use TurboWarp primarily.

If the consensus is to have TurboWarp automatically reopen after a refresh, then it should do so automatically by remembering if you were using the TurboWarp player before the refresh, so we wouldn't need to make a setting for it.

If the reason is just because people like TurboWarp, then sure, I guess that would help someone, but it would be an edge case. Maybe it would be possible to only auto-open TurboWarp if the project contains [TurboWarp configuration data](https://github.com/TurboWarp/scratch-vm/commit/8c365062bfaeb044c2999fe4628420be8420a831), which is used by TurboWarp to automatically set certain settings like frame rate. Another alternative would be to update `60fps` to automatically use the frame rate configured in the project.

I would suggest we create/locate an issue or issues related to what we're trying to resolve here and discuss the idea and alternatives before proceeding or taking a step back to focus on resolving the [linked issue](https://github.com/ScratchAddons/ScratchAddons/issues/5480) (we could also create a new PR for that).

addons/turbowarp-player/addon.json
Samq64: I would say just don't show the toggle when the new tab mode is selected.

BroJac5246:
@DNin01 aren't you kinda overthinking it? Like, if you don't want this to happen, just don't turn on the setting 🤷

I'm sure people would use this setting, and I don't see any harm in adding it.

DNin01:
If we really wanted to, this could potentially work out with the "replace player" option - people would always benefit from the improved performance, and if it was set to "open new tab", it could open a tab for TurboWarp after clicking on the project thumbnail, which would have some sort of icon or "Open in TurboWarp" label on it.

Another thing to note is this would theoretically prevent Scratch from counting your views of projects you play, since you're always playing them on TurboWarp.

BroJac5246:
> If we really wanted to, this could potentially work out with the "replace player" option - people would always benefit from the improved performance, and if it was set to "open new tab", it could open a tab for TurboWarp after clicking on the project thumbnail, which would have some sort of icon or "Open in TurboWarp" label on it.

Sounds great!
> Another thing to note is this would theoretically prevent Scratch from counting your views of projects you play, since you're always playing them on TurboWarp.

How does Scratch count views? Is it only if you click the green flag? Does it matter to us?

DNin01:
> > Another thing to note is this would theoretically prevent Scratch from counting your views of projects you play, since you're always playing them on TurboWarp.
> 
> How does Scratch count views? Is it only if you click the green flag? Does it matter to us?

Yeah, I think so. Any user that uses this automatic TurboWarp opening and never the Scratch player wouldn't increase the view counter on the projects they view. This might be worth addressing if possible, but I don't know.

addons/turbowarp-player/userscript.js
DNin01, line 1: This shouldn't automatically open tabs of TurboWarp. Instead, we could display an icon or label on the stage to play the project on TurboWarp, or we could hide the setting and not auto-click the button if the button's behavior is set to "open new tab".

DNin01: Sorry, [duplicate conversation](https://github.com/ScratchAddons/ScratchAddons/pull/3985/files#r1332034783).

addons/turbowarp-player/addon.json
DNin01: > I would say just don't show the toggle when the new tab mode is selected.

Good idea. We could also display an icon or label-button on the stage to play the project on TurboWarp, though, that kind of defeats the purpose of it being automatic.

iqnite:
Does this addon pause the Scratch project download when clicking on the TurboWarp button before the project finishes loading? Because having both Scratch and TurboWarp loading the same project at the same time is kind of... inefficient.

PR #4002, opened 2021-12-29
Hide signatures

### LINKED ISSUE #3951: Hide forum signatures addon
This is kind of a no-brainer so I'm surprised it apparently hasn't been suggested yet :P

Simple addon which hides user signatures on forum posts when it's enabled. The "Full areas" addon expands signatures when active; there should be an opposite addon for those who don't care for signatures, want a break from banners and the like, or have any other reasons for hiding signatures!

Another option is to replace signatures with a (quite subtle) "expand signature" button that un-hides the sig when clicked (collapsing by default). This would make it so users don't have to disable the addon if they just want to have a quick look at one person's signature, for example.

FunctionalMetatable:
Resolves #3951

### Changes

Adds a hide signatures addon, which allows you to hide a signature

### Tests

Tested on Firefox


hello-smile6:
Looks interesting! I wanna test this.

addons/hide-signatures/addon.json
```diff
@@ -0,0 +1,37 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["topics"],
+      "runAtComplete": true
+    }
+  ],
+  "settings": [
+    {
+      "name": "Hide signatures by default",
```
RedGuy12, line 19: When would you not want to hide them by default if you have this addon?

addons/hide-signatures/userscript.js
```diff
@@ -0,0 +1,42 @@
+export default async function ({ addon, global, console, msg }) {
+  // get signatures
+  const posts = document.querySelectorAll(".blockpost");
```
RedGuy12, line 3: WaitForElement

WorldLanguages: I'm not sure if there's consensus on whether waitForElement should be used for forum posts 🤔

addons/hide-signatures/addon.json
FunctionalMetatable: Idk

addons/hide-signatures/userscript.js
FunctionalMetatable: The script loads and runs when the page is loaded. The posts will already be loaded as they are not added by JS. Addons which touches posts already use `document.querySelector`, not `addon.tab.waitForElement`

addons/hide-signatures/userscript.js
```diff
@@ -0,0 +1,42 @@
...
+
+  for (const post of posts) {
+    const actionsWrapper = post.querySelector(".postfootright");
+    const signature = post.querySelector(".postsignature");
```
WorldLanguages, line 7: `signature` might be `null`!

RedGuy12: ts be like

PR #4018, opened 2022-01-02
New addon: compact editor

lisa-wolfgang:
### Changes

image
image
image
image
image

### Reason for changes

An addon that makes the editor look like 2.0 is a very popular request. While this isn't an exact implementation, it hits many of the major changes that 3.0 made while still serving as a 3.0-style compact theme for those who want it.

### Tests

Tested on browser w/ Chromium 96.
Since this theme encompasses the entirety of the Scratch editor, it will need thorough, extensive testing before merge.

To-do:
- [x] Fix flyout in RTL layout
- [x] Test on Firefox

We have a lot of editor addons with UI that this addon doesn't touch, but keep in mind that not everything needs to be super-compacted (some things are okay-sized right now, and UI components that advanced users likely won't touch, such as direction popup in sprite info, should be fine too)


Secret-chest:
Nice

mxmou:
What will happen to this addon when a 2.0 or 1.4 theme is added - will they be separate (so one would need to enable both to get a full 2.0 design) or would the user have to choose one or the other (so a 2.0 theme without compact sizing wouldn't exist)?

lisa-wolfgang:
I do think there's value in having this addon be a separate customization option for users wanting a more compact interface but not a full 2.0 revamp.

That being said, when we eventually implement superpresets and separate themes, this would either go under the Scratch 2.0 theme and/or the Scratch 2.0 superpreset.

lisa-wolfgang:
Since this addon doesn't necessarily need to encompass everything, I'm going to mark it as ready for review. We can start by focusing on the core editor and address individual addons as separate PRs later. I think most people are more concerned about things like the sprites panel than the debugger panel, which can be hidden and dragged anywhere on the screen.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,422 @@
+/* Full page */
```
TheColaber, line 1: addon should remove labels in the sounds tab as it does for the costumes tab
image

TheColaber, line 1: the tutorials and logs popup are not compact (they should have the same classes?)
image

TheColaber (dismissed):

addons/editor-compact/userstyle.css
TheColaber, line 1: bug: when hide-flyout is enabled, the scrollbars shows when the flyout is hidden:
image

WorldLanguages:
Would be nice to have @mxmou's current opinion on this. 10 months have happened since this addon was last discussed.

addons/editor-compact/userstyle.css
lisa-wolfgang: Can't repro.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+  height: 20px;
+}
+#s3devHelpContent {
+  height: calc(100% - 2rem);
+}
+
+/* Code tab */
+[iscolumns="true"] .blocklyToolboxDiv,
+.blocklyFlyout {
+  width: 210px;
```
mxmou, line 201: Changing the width of the toolbox probably isn't what most users will expect from a compact UI addon. A resizable toolbox addon would be great though.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+.blocklyScrollbarVertical.blocklyFlyoutScrollbar {
+  transform: translate(calc(196.5px + 60px), 2.5px) !important;
+}
+[iscolumns="true"] .blocklyScrollbarVertical.blocklyFlyoutScrollbar {
+  transform: translate(196.5px, 172.5px) !important;
+}
+.scratchCategoryItemBubble,
+.scratchCategoryItemIcon {
+  width: 0.75rem;
+  height: 0.75rem;
```
mxmou, line 218: This makes icons really tiny. Reducing padding might be a better way to make the category menu compact.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+[class*="sound-editor_button_"] img,
+[class*="sound-editor_tool-button"] img {
+  width: 20px;
+  height: 20px;
+}
+[class*="asset-panel_detail-area"] [class*="label_input-label"],
+[class*="asset-panel_detail-area"] [class*="labeled-icon-button_edit-field-title"],
+[class*="icon-button_title_"] {
+  display: none;
+}
```
mxmou, line 307: Buttons in the costume editor have tooltips so it's possible to identify them even if the label is hidden. This isn't the case in the sound editor. I think hiding the labels makes sense but the userscript should add tooltips with the same text.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+[class*="asset-panel_detail-area"] [class*="button_button"],
+[class*="sound-editor_button_"] {
+  padding: calc(0.125rem - 1px);
+}
+[class*="asset-panel_detail-area"] [class*="dashed-border"] {
+  border-right: none;
```
mxmou, line 313: This doesn't work in RTL.
```suggestion
  border: none;
```

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+  height: 1.5rem;
+}
+[class*="filter_filter-icon"],
+[class*="filter_x-icon_"] {
+  width: 0.75rem;
+  margin: 0.25rem 0.5rem !important;
+}
+[class*="filter_filter-input"] {
+  padding-left: 1.75rem !important;
+}
```
mxmou, line 441: Unnecessary `!important`.
```suggestion
[dir="ltr"] [class*="filter_filter-icon"],
[dir="rtl"] [class*="filter_filter-icon"],
[class*="filter_x-icon_"] {
  width: 0.75rem;
  margin: 0.25rem 0.5rem;
}
[dir="ltr"] [class*="filter_filter-input"] {
  padding-left: 1.75rem;
}
[dir="rtl"] [class*="filter_filter-input"] {
  padding-right: 1.75rem;
}
```

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+[class*="library_tag-wrapper"] {
+  height: auto;
+}
+[class*="library_library-scroll-grid"] {
+  height: calc(100% - 2rem - 2rem) !important;
+}
+[class*="modal_header"] + [class*="library_library-scroll-grid"] {
+  height: calc(100% - 2rem) !important;
+}
```
mxmou, line 457: Unnecessary `!important`.
```suggestion
[class*="library_library-scroll-grid"] {
  height: calc(100% - 2rem);
}
[class*="library_library-scroll-grid"][class*="library_withFilterBar"] {
  height: calc(100% - 4rem);
}
```

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+
+[class*="prompt_body"],
+[class*="custom-procedures_body"] {
+  padding: 1rem;
+}
+[class*="modal_modal-content"] [class*="box_box"] button {
+  padding: 0.5rem 0.75rem !important;
```
mxmou, line 464: Unnecessary `!important`

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+}
+[class*="selector_list-area"] img[class*="sprite-selector-item_sprite-image"] {
+  max-width: 52px;
+  max-height: 44px;
+}
+#react-tabs-5 [class*="selector_list-area"] > div > div {
+  height: 2rem;
+}
+#react-tabs-5 [class*="sprite-selector-item_sprite-image-outer"] {
+  display: none;
```
mxmou, line 264: Scratch 2.0 did have these icons so I don't think this addon should remove them.

mxmou (commented):
An important part of this addon's design is how it will integrate with other addons. The way I understand it is that it will become a part of a "Scratch 2" "superpreset" in the future. There will also be a separate and independent addon that will change the visual style to match Scratch 2.

If the actual intention is different, it should probably be explained in the PR description. In the rest of my review, I'll assume that it isn't and that my description above is correct.

I described things I would change in comments below. The general idea is that this addon shouldn't make unnecessary visual changes and that it doesn't have to make things more compact than they were in Scratch 2.

I also noticed some things that aren't changed but probably should be:
* 4px box shadows on selected or focused items look really big in this compact layout. Some also touch the edge of the container, which looks weird.
image
image
A 2px shadow might look better:
image
image
* The top edges of the workspace and the stage aren't aligned:
image
This could be fixed by making stage controls smaller:
image
* Icons next to the tool size input in the costume editor are too big.
image
Switching to a tool with a size input makes the toolbar higher which causes a layout shift.

> UI components that advanced users likely won't touch, such as direction popup in sprite info, should be fine

Not everyone who prefers the 2.0 design is an advanced user.

It's also hard to predict which features people will use - I consider myself an advanced user and I do use that popup to set the rotation mode.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+  width: 8px;
+  height: 8px;
+}
+body::-webkit-scrollbar {
+  background-color: var(--editorDarkMode-page, #e5f0ff);
+}
+::-webkit-scrollbar-thumb {
+  background-color: var(--editorDarkMode-workspace-scrollbar, #cecdce);
+  border-radius: 4px;
+}
```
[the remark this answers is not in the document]
lisa-wolfgang: On Windows, scrollbars take up an uncomfortable amount of space. This is most noticeable on list monitors. Changes were made to the appearance to make the scrollbars consistently compact.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+[class*="delete-button_delete-button-visible"] {
+  width: 1rem;
+  height: 1rem;
+}
+[class*="delete-button_delete-icon"] {
+  width: 2rem;
+  height: 2rem;
+  bottom: 0.25rem;
+}
```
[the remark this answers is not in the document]
lisa-wolfgang: One consideration when I styled the button was how much it overlaps with 4:3 rectangular sprites. The early 3.0 design would do this a bit. I do think that it makes sense to change the button to look like this, though.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+[class*="sprite-selector-item_sprite-image-inner"] {
+  margin: 0;
+}
+img[class*="sprite-selector-item_sprite-image"] {
+  max-width: 56px;
```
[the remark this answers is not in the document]
lisa-wolfgang: The design philosophy I followed was to minimize the size of the UI to give the user-generated content the most possible space. It doesn't make much sense to shrink everything down and then do nothing with the extra space.

addons/editor-compact/userstyle.css
lisa-wolfgang: Does the toolbox not count as a UI element?

addons/editor-compact/userstyle.css
lisa-wolfgang: In addition to adding tooltips when the buttons are hidden, it would make sense to have a setting for whether or not to hide the labels.

lisa-wolfgang:
Thanks for leaving such detailed feedback, mxmou. I replied as best I could, but I'd first like to open up the addon on my laptop (it's been a while) before addressing the remaining comments.

addons/editor-compact/userstyle.css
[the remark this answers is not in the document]
mxmou: Makes sense - I checked how the thumbnails are sized in 2.0 and they can indeed get quite wide.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+  display: none;
+}
+
+[class*="gui_menu-bar-position"][class*="menu-bar_menu-bar"][class*="box_box"],
+[class*="menu-bar_menu-bar-item"] {
+  height: 2rem;
+}
+[class*="gui_menu-bar-position"][class*="menu-bar_menu-bar"][class*="box_box"] {
+  padding: 0;
+}
```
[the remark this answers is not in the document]
lisa-wolfgang: I'll remove it.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,492 @@
...
+[class*="sprite-info_sprite-info"] [class*="input_input-form"],
+[class*="gui_tabs"] [class*="input_input-form"] {
+  height: 1.5rem;
+  padding: 0 0.5rem;
+}
+[class*="sprite-selector-item_is-selected"] [class*="sprite-selector-item_sprite-info"] {
+  background: none;
+  color: inherit;
+}
```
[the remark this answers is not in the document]
lisa-wolfgang: The intention was to allow the sprite costume preview to be a bit larger without looking strange due to a lack of padding.

addons/editor-compact/userstyle.css
lisa-wolfgang: It does make the sounds list more compact, but perhaps it would be better to be consistent with the costumes tab.

addons/editor-compact/userstyle.css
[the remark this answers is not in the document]
mxmou: I think there's enough padding even if the background is there:
image

The addon allows the preview to get wider but not taller.

addons/editor-compact/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Compact editor",
+  "description": "Decreases the size of the menu bars and buttons in the editor, similar to the appearance of Scratch 2.0.",
```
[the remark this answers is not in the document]
mxmou: "Menu bars" are still mentioned in the description. That sounds weird to me - isn't there only one menu bar?

addons/editor-compact/userstyle.css
mxmou: From a user's perspective it's a section/area, not a single element. Changing it to the 2.0 size would only make sense if blocks were also made smaller by the same addon.

WorldLanguages:
Should mxmou re-review this?

addons/columns/userscript.js
```diff
@@ -148,11 +148,22 @@ export default async function ({ addon, msg, console }) {
...
+
+    addAttribute();
+  }
+
+  function addAttribute() {
+    // Add attribute to allow editor-compact to handle this addon
+    if (addon.tab.editorMode === "editor") {
+      if (addon.self.disabled) document.querySelector("[class*='gui_tab-panel']").setAttribute("iscolumns", "false");
+      else document.querySelector("[class*='gui_tab-panel']").setAttribute("iscolumns", "true");
+    }
```
WorldLanguages, line 160: I'd suggest you change this to `data-sa-columns`:

```js
element.dataset.saColumns = "";
```

And I don't think it's necessary to set it to "false". Just remove the attribute with removeAttribute(). You'll avoid having to type `="true"` in the CSS.

addons/editor-compact/addon.json
```diff
@@ -0,0 +1,31 @@
...
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.30.0",
```
WorldLanguages, line 29: 1.31.0

addons/editor-compact/userscript.js
```diff
@@ -0,0 +1,15 @@
+export default async function ({ addon, global, console }) {
+  // The workspace needs to be manually resized via a window resize event
+  // whenever the addon modifies or stops modifying UI elements
+  resizeWorkspace();
+
+  let resizeObserver = new ResizeObserver(resizeWorkspace);
+  let menuBar = await addon.tab.waitForElement(
+    '[class*="gui_menu-bar-position"][class*="menu-bar_menu-bar"][class*="box_box"]'
+  );
+  resizeObserver.observe(menuBar);
```
WorldLanguages, line 10: Isn't `while(true)` necessary here? Isn't this element re-created every time the user enters the editor?

addons/editor-compact/userstyle.css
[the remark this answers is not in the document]
WorldLanguages: Is this resolved?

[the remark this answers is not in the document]
mxmou: Not yet - the CSS variable is still used incorrectly.

WorldLanguages:
Inspired by #5866, we should consider adding a setting to show only the icons for the code/costumes/sounds tabs:
image

addons/editor-compact/addon.json
TylerMS887, line 29:
```suggestion
  "versionAdded": "1.32.0",
```
1.32 has been released some time ago (March 11<sup>th</sup>, 2023) so this might not work for older SA versions.

Platformer444: > this might not for older SA versions

The addon would work for older SA versions. The `versionAdded` property just decides whether the addon should be shown in the `New and updated addons` category. Just clarifying your statement.

TylerMS887: > The addon would work for older SA versions. The `versionAdded` property just decides whether the addon should be shown in the `New and updated addons` category.

Thanks.

GrahamSH-LLK:
Seems like there are a few small changes waiting to be made, should we assign somebody to implement them?

WorldLanguages:
@mxmou Want to make some changes to this PR? (not necessarily all of the pending ones)

mxmou:
> @mxmou Want to make some changes to this PR? (not necessarily all of the pending ones)

I think I'll do the ones I suggested.

WorldLanguages:
I assume this PR will need some changes after the upcoming Scratch editor update. Might not be worth it to try to get it merged before then.

WorldLanguages:
> I assume this PR will need some changes after the upcoming Scratch editor update. Might not be worth it to try to get it merged before then.

Ping! @mxmou We can continue.

mxmou (approved):

addons/editor-compact/userscript.js
```diff
@@ -0,0 +1,39 @@
...
+      else icon.removeAttribute("title");
+    }
+  };
+  updateTooltips();
+  addon.settings.addEventListener("change", updateTooltips);
+  addon.self.addEventListener("disabled", updateTooltips);
+  addon.self.addEventListener("reenabled", updateTooltips);
+  while (true) {
+    await addon.tab.waitForElement("[class*='sound-editor_editor-container_']", {
+      markAsSeen: true,
```
WorldLanguages, line 34: Borrowed from editor-dark-mode. Would probably work fine here (untested)

```suggestion
      markAsSeen: true,
      reduxEvents: [
        "scratch-gui/navigation/ACTIVATE_TAB",
        "scratch-gui/mode/SET_PLAYER",
        "fontsLoaded/SET_FONTS_LOADED",
        "scratch-gui/locales/SELECT_LOCALE",
        "scratch-gui/targets/UPDATE_TARGET_LIST",
      ],
```

addons/columns/userscript.js
```diff
@@ -148,11 +148,22 @@ export default async function ({ addon, msg, console }) {
...
+    if (addon.tab.editorMode === "editor") {
+      if (addon.self.disabled) document.querySelector("[class*='gui_tab-panel']").classList.toggle("sa-columns", false);
+      else document.querySelector("[class*='gui_tab-panel']").classList.toggle("sa-columns", true);
+    }
   }
 
   updateToolbox();
   addon.self.addEventListener("disabled", updateToolbox);
   addon.self.addEventListener("reenabled", updateToolbox);
+  addon.tab.addEventListener("urlChange", addClass);
```
WorldLanguages, line 166: Is this enough? This won't react to actions such as changing the language of the editor.

WorldLanguages:
Looks like going into fullscreen (from the project page) is glitchy: (should this addon actually affect this view? I feel this discussion applies to multiple addons, we've discussed it quite a few times but I never remember what we concluded)

image

addons/editor-compact/hide-labels.css
```diff
@@ -0,0 +1,63 @@
...
+[dir="rtl"] [class*="paint-editor_mod-dashed-border_"] {
+  border: none;
+}
+[class*="mode-tools_mod-labeled-icon-height"],
+[class*="paint-editor_mod-labeled-icon-height"] {
+  height: auto;
+}
+[class*="paint-editor_row_"] > [class*="tools_row_"],
+[class*="paint-editor_mod-mode-tools_"] > [class*="mode-tools_mode-tools_"] {
+  gap: calc(0.25rem);
```
WorldLanguages, line 29: @mxmou Have we been avoiding `gap` prior to this? It was rarely used in editor addon CSS before this.
Using it is perfectly fine, I'm just wondering if there's places where we have been avoiding it and it could result on cleaner CSS.

addons/editor-compact/userscript.js
```diff
@@ -0,0 +1,39 @@
+export default async function ({ addon, global, console }) {
+  // The workspace needs to be manually resized via a window resize event
+  // whenever the addon modifies or stops modifying UI elements
+  resizeWorkspace();
+
+  let resizeObserver = new ResizeObserver(resizeWorkspace);
+  (async () => {
+    while (true) {
+      let menuBar = await addon.tab.waitForElement('[class*="menu-bar_menu-bar"]', { markAsSeen: true });
```
WorldLanguages, line 9: Self TODO: add reduxEvents and reduxCondition

WorldLanguages:
Amazing work from Lisa and mxmou!

mxmou:
> Looks like going into fullscreen (from the project page) is glitchy: (should this addon actually affect this view? I feel this discussion applies to multiple addons, we've discussed it quite a few times but I never remember what we concluded)

This addon probably shouldn't affect full screen at all. It didn't before this commit: 3bd97f6

Only changing editor full screen mode could be confusing because it looks identical to website full screen in vanilla Scratch.

WorldLanguages:
Yeah, I agree, this addon shouldn't affect fullscreen mode.

WorldLanguages:
Tiny list of things left to do:

- Do not affect fullscreen
- Add some optimizations to waitForElement calls
- Fix `columns` addon not calling addClass() when changing editor language

WorldLanguages:
> Do not affect fullscreen

@mxmou What's the best way to achieve this?

mxmou:
> > Do not affect fullscreen
> 
> @mxmou What's the best way to achieve this?

Adding `:not([class*="stage-wrapper_full-screen"] *) to stage button selectors would probably work.

> (Related: how does mouse-pos distinguish editor full screen vs project page full screen? I can't find the code that does that)

It checks if `editorMode === "editor"` in the `waitForElement` loop. This condition would always be false in full screen mode, but it isn't actually checked when entering or exiting full screen.

addons/editor-compact/userstyle.css
```diff
@@ -0,0 +1,490 @@
...
+.sa-body-editor [class*="green-flag_green-flag_"],
+.sa-body-editor .pause-btn {
+  margin-inline-end: 0.125rem;
+}
+.sa-stage-hidden [class*="stage-header_stage-size-row"] {
+  height: 2rem;
+}
+[class*="gui_stage-and-target-wrapper_"] [class*="stage-wrapper_full-screen_"] {
+  top: 2rem;
+}
```
WorldLanguages, line 224: This selector intentionally targets fullscreen.

WorldLanguages: Commit https://github.com/ScratchAddons/ScratchAddons/commit/3bd97f62f2e2f850e089fe709fbe90f78173d315

WorldLanguages: I removed this CSS rule.

WorldLanguages:
Removed some CSS rules that applied on fullscreen mode.

As far as I can see, right now only the scrollbar differs with this addon enabled vs. disabled, in project player fullscreen mode.
Note that usually, going to fullscreen from inside the editor does not result in a vertical scrollbar, but going fullscreen from the project page does.

WorldLanguages (approved):
Looks like we're good to go 🎉 

This addon still affects the `<body>` scrollbar in fullscreen mode, but we can fix it later.

In the future, I would also like to provide a setting to achieve this (which is even more compact):
image

CST1229:
> In the future, I would also like to provide a setting to achieve this (which is even more compact):

I don't really see any use for that (since there isn't any overflow on the tab bar even with find bar and variable manager enabled on the lowest possible editor width).

Samq64:
This PR was the 3,000th commit!

WorldLanguages:
Issue #6930 suggests adding the ability to toggle the compactness of different part of the editor when using this addon.

PR #4019, opened 2022-01-02
New addon: Compact block category menu

CST1229:
### Changes

Adds the `compact-categories` addon, which hides the labels on the block category menu.

### Reason for changes

Some people might have smaller screens, and this allows them to see more categories and also have slightly more code area space.

### Tests

Works.

`columns` compatibility is still a WIP and currently not really working.


lisa-wolfgang:
My opinion is that `columns` and this addon need to be mutually exclusive, and the only way to do that is to have them as a `select` setting within the same addon.

Samq64:
I think the Add Extension button should be a circle with the label removed and put next to the other loaded extensions when columns is enabled.

CST1229:
> I think the Add Extension button should be a circle with the label removed and put next to the other loaded extensions when columns is enabled.

That's pretty hard to do without a userscript.

JaydenDev:
An option for a category menu like this would be cool:
image

Source: ClipCC

CST1229:
> An option for a category menu like this would be cool: -snip-
> 
> Source: ClipCC

Outside the scope of this addon; you know it is supposed to be compact?

JaydenDev:
> > An option for a category menu like this would be cool: -snip-
> > Source: ClipCC
> 
> Outside the scope of this addon; you know it is supposed to be compact?

This is a very compact category menu what do you mean?

CST1229:
> This is a very compact category menu what do you mean?

It's not compact when it's literally bigger than the vanilla category menu.

JaydenDev:
It was just an example screenshot, it wasn't a mockup or anything.

RedGuy12:
I vote to have this in the same addon as #4018

mxmou:
> I vote to have this in the same addon as #4018

#4018 is for reducing empty space, not removing labels. I actually think it shouldn't remove icons from tabs either.

I agree with combining addons that modify the layout of the category menu into a single addon with a select setting.

CST1229:
This PR has been inactive and drafted for a while, and also I'd probably start from scratch if I wanted to develop this addon further, so I'm gonna close this.

WorldLanguages:
There wasn't an issue for this, so I created one: #5973

PR #2789, opened 2021-06-15
Bold block text addon

Secret-chest:
Resolves None

### Changes

Added bold text on blocks addon

### Reason for changes

Readability

### Tests

No, too simple. I tested it in Stylus

Issue #2645, opened 2021-05-30
Bolder / larger symbols in math operator blocks


astrosticks:
### Is your request related to a problem? Please describe.

The way that Scratch displays individual symbols in the math operator blocks can be pretty hard to read. For reference, here's what they look like:

![image](https://user-images.githubusercontent.com/32315760/120116955-6c182c00-c150-11eb-9b4d-320c97600f95.png)

### Describe the solution you'd like

It would be cool to have an extension that replaces these symbols with larger, bolder symbols. Certain ones, like the multiplication and division symbols, could be replaced with better substitutes **×** and **÷**.

Cool-Scratcher:
Making the operators bigger and bolder would make sense, but wouldn't that look inconsistent when compared to the other blocks? Also replacing the * and / symbols wouldn't look any better in my opinion. The × and ÷ symbols are even harder to read than * and /.

22yeetz22:
> Making the operators bigger and bolder would make sense, but wouldn't that look inconsistent when compared to the other blocks? Also replacing the * and / symbols wouldn't look any better in my opinion. The × and ÷ symbols are even harder to read than * and /.

Then leave it as a option

Time-dragon:
> The × and ÷ symbols are even harder to read than * and /.

Wrong.

Cool-Scratcher:
> > The × and ÷ symbols are even harder to read than * and /.
> 
> Wrong.

Nobody's wrong lol, I'm allowed to have an opinion.  Honestly, × and ÷ are just way too tiny and unreadable for me to actually envision them on the blocks.

Cool-Scratcher:
> > Making the operators bigger and bolder would make sense, but wouldn't that look inconsistent when compared to the other blocks? Also replacing the * and / symbols wouldn't look any better in my opinion. The × and ÷ symbols are even harder to read than * and /.
> 
> Then leave it as a option

Well, if I were to enable this, I think it would look inconsistent with the other blocks.

towerofnix:
I don't think consistency needs to be a big concern here. If someone doesn't like the look of it next to other blocks, that's okay: they don't need to enable this addon!

For those who do struggle to recognize the symbols as they currently are, bolder ones would be a big help, so I think this suggestion could be a really nice accessibility addon.

[...]

PR #4031, opened 2022-01-06
New addon: customizable block text style

lisa-wolfgang:
Moved from #2789

### Changes

Like the previous PR, this new addon increases the thickness of block text.
However, it is now an option alongside adding a shadow to block text. Moving this option from `editor-dark-mode` was suggested in the previous PR.

image

image

### Reason for changes

The original PR was made to enhance block readability.
As for the reason this PR exists, the author of the old PR deleted the branch that contained the proposed changes. Since changes needed to be made, I created my own branch to avoid the messiness of working off of a non-existent branch.

### Tests

Tested on browser w/ Chromium 96 and Firefox 95.0.2.
Works by itself and compatible with `editor-theme3` and `custom-block-shape`.
One thing I noticed is that the text shadow offset does not scale when the code editor zoom level is modified. If we want to fix this, it would likely require a Blockly-based approach, which is probably a better implementation for this addon anyway. Not a big deal, though, since this has been happening since the very beginnings of `editor-dark-mode`. (If we do decide to rewrite in Blockly, might as well implement #2645 too...)


WorldLanguages:
@mxmou What do you think about moving block text shadow to this addon? I think it makes sense.

background/get-addon-settings.js
```diff
@@ -46,6 +46,15 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
         addonsEnabled["block-count"] = true;
       }
       if (addonId === "editor-dark-mode") {
+        if (settings.textShadow && addonsEnabled["custom-block-text"] === undefined) {
+          // Transition v1.23 to v1.24
+          // Moved text shadow option to the custom-block-text addon
+          madeChangesToAddon = madeAnyChanges = true;
+          delete settings.textShadow;
+          addonsEnabled["custom-block-text"] = true;
```
WorldLanguages, line 54: This new addon should only be enabled if dark-www was enabled at the time of the transition.

background/get-addon-settings.js
```diff
@@ -46,6 +46,15 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
         addonsEnabled["block-count"] = true;
       }
       if (addonId === "editor-dark-mode") {
+        if (settings.textShadow && addonsEnabled["custom-block-text"] === undefined) {
```
WorldLanguages, line 49: `custom-block-text` comes before `dark-www` in addons.json, so there's no way it will be undefined at this point.

background/get-addon-settings.js
lisa-wolfgang: You mean `editor-theme-3` or `editor-dark-mode`?

WorldLanguages: I meant editor dark mode, whoops

WorldLanguages:
@mxmou You're fine with this, right? Bold text setting will move to this addon.

addons/custom-block-text/addon.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "Customizable block text style",
+  "description": "Changes the thickness of the text on Scratch blocks and optionally adds a text shadow.",
```
mxmou, line 3: Is "Scratch" necessary in this description? I don't think there are any other "blocks" this could possibly refer to, and other addons just call them "blocks".

WorldLanguages: Agreed

mxmou:
> @mxmou You're fine with this, right? Bold text setting will move to this addon.

Yes, I like this. Editor dark mode doesn't control anything else about blocks, so this is certainly a better place for the text shadow setting.

background/get-addon-settings.js
```diff
@@ -46,6 +46,15 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
...
+          // Transition v1.23 to v1.24
+          // Moved text shadow option to the custom-block-text addon
+          madeChangesToAddon = madeAnyChanges = true;
+          delete settings.textShadow;
+          addonsEnabled["custom-block-text"] = true;
+          addonSettings["custom-block-text"].textShadow = true;
```
WorldLanguages, line 55: The setting is `shadow`, not `textShadow`!

WorldLanguages:
Migration code tested

Mbrick2:
Co-authored-by: RedGuy12 <paul@reid-family.org>
Co-authored-by: World_Languages <mundofinky@gmail.com>

PR #4047, opened 2022-01-10
Manifest: new credits.note property, intended to be localized

### LINKED ISSUE #2172: Localize addon credits

Some addon credits should be localized, for example *Boomer001 (Userstyle)* in `forum-quote-code-beautifier` and *Maximouse (addon, "Experimental" Dark)* in `editor-dark-mode`.

I think the best solution is to add another property to credits, separate from `name` (maybe `contribution`), that would be localizable. For example:
```json
{
  "name": "Maximouse",
  "contribution": "addon, \"Experimental\" Dark",
  "link": "..."
}
```

Alternatively, the `name` property could be localized.

WorldLanguages:
Resolves #2172

Also, moves the Silk credits from forum-toolbar from the description to the addon credits.

### DIFF
--- a/addons/forum-quote-code-beautifier/addon.json
+++ b/addons/forum-quote-code-beautifier/addon.json
@@ -9,7 +9,7 @@
     },
     {
       "name": "Boomer001",
-      "note": "userstyle",
+      "note": "original userstyle",
       "id": "userstyle"
     }
   ],
--- a/addons/forum-quote-code-beautifier/addon.json
+++ b/addons/forum-quote-code-beautifier/addon.json
@@ -3,10 +3,14 @@
   "description": "Shows nicer code and quote blocks on the forums.",
   "credits": [
     {
-      "name": "GrahamSH (Port)"
+      "name": "GrahamSH",
+      "note": "addon",
+      "id": "addon"
     },
     {
-      "name": "Boomer001 (Userstyle)"
+      "name": "Boomer001",
+      "note": "userstyle",
+      "id": "userstyle"
     }
   ],
   "settings": [
--- a/addons/forum-quote-code-beautifier/addon.json
+++ b/addons/forum-quote-code-beautifier/addon.json
@@ -4,8 +4,8 @@
   "credits": [
     {
       "name": "GrahamSH",
-      "note": "addon",
-      "id": "addon"
+      "note": "addon version",
+      "id": "addonVersion"
     },
     {
       "name": "Boomer001",
--- a/addons/scratch-notifier/addon.json
+++ b/addons/scratch-notifier/addon.json
@@ -9,7 +9,9 @@
       "name": "NitroCipher/ZenithRogue"
     },
     {
-      "name": "notificationsounds.com (Scratch Addons ping)"
+      "name": "notificationsounds.com",
+      "note": "Scratch Addons ping",
+      "id": "Scratch Addons ping"
     }
   ],
   "permissions": ["notifications"],
--- a/background/load-addon-manifests.js
+++ b/background/load-addon-manifests.js
@@ -21,6 +21,11 @@
           info.text = scratchAddons.l10n.get(`${folderName}/@info-${info.id}`, {}, info.text);
         }
       }
+      if (manifest.credits) {
+        for (const credit of manifest.credits || []) {
+          credit.note = scratchAddons.l10n.get(`${folderName}/@credits-${credit.id}`, {}, credit.note);
+        }
+      }
       if (manifest.popup) {
         manifest.popup.name = scratchAddons.l10n.get(`${folderName}/@popup-name`, {}, manifest.popup.name);
       }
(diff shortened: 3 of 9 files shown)


mxmou:
Related: #2172

WorldLanguages:
@mxmou I think it would be easy to do that, so let's do it now! I'll convert this PR to a fix for that.

WorldLanguages:
Hmm, credits would need to have an ID if they have a note.

RedGuy12:
(Note to self to add credit tooltips to scradd when this is done)

WorldLanguages:
TODOs:

- [ ] Manifest schema
- [ ] Update localization script

RedGuy12:
@WorldLanguages update fuse settings so this can be searchable?

WorldLanguages:
@RedGuy12 Doesn't make sense IMO.

RedGuy12:
Well we already let credit.name be searchable and this info was in credit.name before. so removing it could break some old behaviour as well but that's really my main reason tbh

addons/forum-quote-code-beautifier/addon.json
```diff
@@ -3,10 +3,14 @@
...
     {
-      "name": "GrahamSH (Port)"
+      "name": "GrahamSH",
+      "note": "addon",
+      "id": "addon"
     },
     {
-      "name": "Boomer001 (Userstyle)"
+      "name": "Boomer001",
+      "note": "userstyle",
```
RedGuy12, line 12: while we're at it, change this to "original userstyle"?

addons/forum-quote-code-beautifier/addon.json
```diff
@@ -3,10 +3,14 @@
   "description": "Shows nicer code and quote blocks on the forums.",
   "credits": [
     {
-      "name": "GrahamSH (Port)"
+      "name": "GrahamSH",
+      "note": "addon",
```
RedGuy12, line 7: "port" sounds better imo

RedGuy12:
live-featured-project, material-forum-editor-buttons, scratch-notifier still have unlocalized notes

addons/forum-quote-code-beautifier/addon.json
WorldLanguages: port, where ships load? tcp ports?

RedGuy12: hmm i see your point. maybe `port to addon`?

mxmou: "Addon version"?

apple502j:
l10n: ScratchAddons/l10n-script#10
Manifest: ScratchAddons/manifest-schema#55

apple502j (approved):

addons/forum-quote-code-beautifier/addon.json
WorldLanguages: I'm going with "addon version"

addons/forum-quote-code-beautifier/addon.json
WorldLanguages: Okay

background/load-addon-manifests.js
```diff
@@ -21,6 +21,11 @@
           info.text = scratchAddons.l10n.get(`${folderName}/@info-${info.id}`, {}, info.text);
         }
       }
+      if (manifest.credits) {
+        for (const credit of manifest.credits || []) {
+          credit.note = scratchAddons.l10n.get(`${folderName}/@credits-${credit.id}`, {}, credit.note);
```
RedGuy12, line 26: This does not guarantee credit.id is set. If it is not set, credit.note ends up addonName/@credits-undefined.

apple502j: @RedGuy12 Such errors should be caught by the schema validator.

addons/scratch-notifier/addon.json
```diff
@@ -9,7 +9,9 @@
       "name": "NitroCipher/ZenithRogue"
     },
     {
-      "name": "notificationsounds.com (Scratch Addons ping)"
+      "name": "notificationsounds.com",
+      "note": "Scratch Addons ping",
+      "id": "Scratch Addons ping"
```
RedGuy12, line 14: does this work with non-alphanumeric characters?

background/load-addon-manifests.js
RedGuy12: either both `id` and `note` is set, or neither is set. if both are set, credit.note gets reassigned to the message `addonName/@credits-creditId`. If neither are set, **this line still runs**. We forgot to add any conditional here. it tries to fetch the string with id `addonName/@credits-undefined`.  No string is found so it falls back to the key, `"addonName/@credits-undefined"`.

addons/scratch-notifier/addon.json
WorldLanguages: Yeah this was unintentional

PR #4051, opened 2022-01-11
New Addon: mouse-cursor

Samq64:
Resolves https://github.com/ScratchAddons/ScratchAddons/pull/3515#issuecomment-928536581

### Changes

Adds a new addon that hides the cursor when hovering over the stage.

### Reason for changes

Projects with custom cursors would look better without the real cursor and others like animations don't need a cursor at all. This might get closed because there's no way to toggle it, but Scratchers who mainly look at animations and other projects that don't need a cursor might still enable it.

### Tests

Tested on Edge 97.


WorldLanguages:
Since there's no real way to know if a project has a cursor sprite that would replace the real OS cursor, I don't think a global toggle would make sense. A per-project toggle would be better but may be too much (not sure, just a thought). Also, any extension that lets you hide the cursor when it's above a canvas would do the job, even if the author did not develop it for Scratch specifically.

PR #4104, opened 2022-01-22
Change updated message to "New features"

apple502j:
Some addons can update without adding new options.


WorldLanguages:
"Options" is a really weird too, we usually use "settings".

PR #4121, opened 2022-01-24
Make avatar outlines added by 2.0 → 3.0 customizable

mxmou:
### Changes

Makes profile picture outlines added by the Scratch 2.0 → 3.0 addon affected by the customize-avatar-border addon.

`addonEnabled` doesn't auto-refresh, which means that the page has to be reloaded for this to work after enabling scratchr2. Could that be fixed?


WorldLanguages:
> `addonEnabled` doesn't auto-refresh, which means that the page has to be reloaded for this to work after enabling scratchr2. Could that be fixed?

`addonEnabled` was initially added just for using in settings, but if you find this pattern useful for making userstyles compatible with each other, we can make it recalculate after addons are disabled/enabled.

mxmou:
> > `addonEnabled` doesn't auto-refresh, which means that the page has to be reloaded for this to work after enabling scratchr2. Could that be fixed?
> 
> `addonEnabled` was initially added just for using in settings, but if you find this pattern useful for making userstyles compatible with each other, we can make it recalculate after addons are disabled/enabled.

This is now the second addon that uses it that way (dark-www already does).

WorldLanguages:
> This is now the second addon that uses it that way (dark-www already does).

Whoops, and does it not suffer from this problem (no autorefresh)? Or does the specified addon not have dynamicDisable/Enable?

WorldLanguages:
Huh, it's also scratchr2. So why didn't bring this up earlier, is there something I'm missing?

mxmou:
> Huh, it's also scratchr2. So why didn't bring this up earlier, is there something I'm missing?

I thought I already mentioned this on #3710 - looks like I didn't.

WorldLanguages:
@mxmou So should we fix this before releasing? I haven't tested, but it looks like we'll have bugs in some pages if scratchr2 is dynamically enabled/disabled with dark-www on.

mxmou:
> @mxmou So should we fix this before releasing? I haven't tested, but it looks like we'll have bugs in some pages if scratchr2 is dynamically enabled/disabled with dark-www on.

I think the only problem it currently causes is this:
image
image

WorldLanguages:
@mxmou That seems low prio so we will release without fixing this.

WorldLanguages:
> `addonEnabled` doesn't auto-refresh, which means that the page has to be reloaded for this to work after enabling scratchr2. Could that be fixed?

Fixed (#4289)

WorldLanguages:
@mxmou Can you confirm if dynamic per-userstyle dynamic/disable is working correctly? If so, you can merge.

mxmou:
> @mxmou Can you confirm if dynamic per-userstyle dynamic/disable is working correctly? If so, you can merge.

It seems to work.

PR #4156, opened 2022-02-01
Attempt to fix broken profile layout on macOS

### LINKED ISSUE #4154: Profile areas have no spacing
### Description

Profiles look.. broken..
For example here's Chiroyce's profile:
image

### Repoduction steps

1. go to any profile
2. see the big problem

### Expected behaviour

Should look similar to before with a few changes

### Actual behaviour

It has no spacing at all and the About Me doesn't have a line break

### Screenshots

image

### Environment

- Browser: Firefox 96.0.3
- Scratch Addons version: 1.24.0
- Operating system: macOS Monterey 12.2

### Additional context

I just saw the new update and it's cool but this is confusing
It might be my addon configuration?
I tested with my userstyle and adblocker off too

mxmou:
Resolves #4154

### Changes

This should fix an issue on profile pages that happens consistently on macOS and on specific zoom levels on Windows when both Scratch 2.0 → 3.0 and profile page banner are enabled.

### Tests

Tested in Edge on Windows. Someone who uses macOS would need to check if this fixes the problem there as well.

### DIFF
--- a/addons/better-featured-project/userstyle.css
+++ b/addons/better-featured-project/userstyle.css
@@ -131,6 +131,10 @@
 #lfp-change-featured {
   display: none !important;
 }
+#bio,
+#bio-readonly {
+  margin-bottom: 0;
+}
 #bio-readonly,
 #status-readonly {
   width: 285px;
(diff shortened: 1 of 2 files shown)


Chiroyce1:
I'll test it now :)

Chiroyce1:
image

Seems to work fine now.

WorldLanguages:
Is it worth releasing v1.24.1 to fix this?

mxmou:
@WorldLanguages I think so.

PR #4159, opened 2022-02-01
New "Hide studio icon" addon

devxan:
Resolves issue of icon covering studio thumbnails.

### Changes
Hides the icon that's in front of studio thumbnails. 

### Reason for changes
Many people want to see the whole studio thumbnail without clicking on it to go to the studio page. The icon is not needed as studio thumbnails have a different aspect ratio from project thumbnails. 

### Tests
When testing this addon, it worked as intended on Google Chrome. 

### Suggestion
Make this on by default because it's such a small addon that it would make sense to just have it enabled already.

### Screenshot 
hide-studio-icons


WorldLanguages:
Might be too small for an addon.

devxan:
> @WorldLanguages 
> Might be too small for an addon.

### Some bullet points of debate
- It's a large enough issue that I think people would use it. 
- Scratch Addons has some other addons that do small things like the block count addon.
- There aren't similar addons that could be merged into one.

retronbv:
> It's a large enough issue that I think people would use it.

it removes a small icon…?

> Scratch Addons has some other addons that do small things like the block count addon.

it brought back a feature from 2.0

> imo this'd be better in 2.0->3.0

agreed… its way to small. Maybe an option in 2.0 to 3.0?

devxan:
> **@RedGuy12:** imo this'd be better in 2.0->3.0
> **@retronbv:** agreed… its way to small. Maybe an option in 2.0 to 3.0?

Great idea. This does have an issue where it hides scratch project statistics when enabled on project pages, so instead of hiding the span `stats`, an option in 2.0 > 3.0 should hide `.gallery.thumb .stats`/`.thumb span.stats, .thumb span.date, .thumb span.owner`. This might work: 
```css
:is(.box .stats):not(#view > .preview > .inner *) {
    display: none !important;
}
```

retronbv:
> This does have an issue where it hides scratch project statistics when enabled on project pages

maybe you should have said that in the pr….

devxan:
> **@retronbv:** maybe you should have said that in the pr….

The addon doesn't run on project pages so it isn't a problem. That's just using a userstyle that works on the whole scratch domain.

RedGuy12:
> The addon doesn't run on project pages so it isn't a problem for Scratch Addons. That's just using a userstyle that works on the whole scratch domain.

Not true. Scratch 2.0->3.0 is split into many files, each for a different part of the website. We could either use the different selector or make it its own file. And if it is added with an option, it'd need its own file anyways.

devxan:
> **@RedGuy12:** Scratch 2.0->3.0 is split into many files, each for a different part of the website. We could either use **_a_** different selector or make it its own file. And if it is added with an option, it'd need its own file anyways.

Are there any other Scratch 2.0 pages that studio thumbnails are shown on besides Profiles, profile sub-pages, and the project studios pages? Those are the only pages that this PR's addon works on. (It adjusts Scratch 2.0 pages, the studio icon isn't shown on Scratch 3.0 pages.)

RedGuy12:
You reminded the code...

PR #4178, opened 2022-02-04
debugger: fix go to block

GarboMuffin:
This makes debugger's go to block feature work again. Previously it would silently error.


apple502j:
@TheColaber bruh

WorldLanguages:
I suggest v1.24.2 with this PR merged as well as a fix yet to be found for bug #4184 (latter does not affect turbowarp)

WorldLanguages:
I find `this` PR funny

PR #4183, opened 2022-02-04
Optimize graphs #4180

### LINKED ISSUE #4180: Laggy Debbuger
https://user-images.githubusercontent.com/90691415/152486591-0afb5464-e29d-45b3-9ca3-292589947f37.mov

Norbiros:
Resolves #4180

### Changes

Create new option, hide fill and animation for smaller lags.

### Reason for changes

Some people have lags viewing graphs.

### Tests

I tested it in browser :)

### DIFF
--- a/addons/debugger/addon.json
+++ b/addons/debugger/addon.json
@@ -73,13 +73,19 @@
       "id": "log_invalid_cloud_data",
       "type": "boolean",
       "default": false
+    },
+    {
+      "name": "Faster graph rendering",
+      "id": "graphs_optimization",
+      "type": "boolean",
+      "default": false
     }
   ],
   "tags": ["editor", "featured"],
   "enabledByDefault": false,
   "versionAdded": "1.16.0",
   "latestUpdate": {
-    "version": "1.24.0",
+    "version": "1.25.0",
     "isMajor": true,
     "temporaryNotice": "New \"Threads\" and \"Performance\" tabs in the debugger window.",
     "newSettings": [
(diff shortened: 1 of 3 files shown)


addons/debugger/addon.json
```diff
@@ -73,23 +73,23 @@
...
     }
   ],
   "tags": ["editor", "featured"],
   "enabledByDefault": false,
   "versionAdded": "1.16.0",
   "latestUpdate": {
-    "version": "1.24.0",
-    "isMajor": true,
-    "temporaryNotice": "New \"Threads\" and \"Performance\" tabs in the debugger window.",
+    "version": "1.24.1",
```
RedGuy12, line 88: 1.25.0

addons/debugger/addon.json
```diff
@@ -73,23 +73,23 @@
...
+    "isMajor": false,
+    "temporaryNotice": "New option to make graphs not laggy!",
```
RedGuy12, line 90: not needed

WorldLanguages: Agreed, the change is too small to warrant a lastUpdated change.

apple502j:
Why have optimization a setting? Why not turn it on always?

Norbiros:
Because it looks no so pretty... It is line without any animation. We can change it... but default in previous version was with fill etc, so i think let's keep as it is. 
Also i found some bugs with dark mode... i will fix it tomorrow?

**Can someone tell me if in line optimization mode, must be gray, maybe change it to blue and add some thickness?**

apple502j:
@Norbiros Then the setting should be renamed to "Use fancy graph" (default on) or "Faster graph rendering" (default off). "Optimization" sounds like there is no compromise.

Norbiros:
About this error. Can someone check it again? I think I fixed error with `Format code using Prettier`.

PR #3950, opened 2021-12-18
Change banner hue based on highlight color

mxmou:
### Changes

Makes blue colored areas change based on the highlight color setting.
image

Adds a `"map"` type for custom CSS variables.

Issue #1092, opened 2020-12-23
Option to change the green color of the forums


leahcimto:
**Describe the new addon**
There would be an option to customize the green color on the 3.0 forums that is at the header and at the top of all posts. Its the part that has the search and says "Discuss Scratch".

Marc92020:
Good idea! This could be intresting!

Example:
![bluediscuss](https://user-images.githubusercontent.com/75702334/103045318-faf59c80-4551-11eb-8db4-df29fb29f6da.png)

j5155:
Cool idea! Would be nice with a hex code input option as well, for people who want a really specific color or something.

RedGuy12:
YES!!! I already use stylus to change it to a dark blue!
```css
#brdheader, #brdheader > .block-head, .blockpost .block-head, .blockpost div.block {
    background-color: #082c5e!important;
    border-top-color: #3020b17!important
```

mxmou:
I think this can remain open - the color is no longer used for posts, but the header is still green, which means users might still want to customize it. There could even be an addon to change all the banner colors, including Scratch's own ones.

WorldLanguages:
I think someone once asked if they could change the color of the orange "Explore" header.

RedGuy12:
> I think this can remain open - the color is no longer used for posts, but the header is still green, which means users might still want to customize it. There could even be an addon to change all the banner colors, including Scratch's own ones.

#3950?



> I think someone once asked if they could change the color of the orange "Explore" header.

That was Aidan in discord

PR #4187, opened 2022-02-04
New "customizable banner colors" addon

MystPi:
### Changes
Allows you to customize the color of the discuss, explore, and My Stuff banner. However, you can only change the discuss and My Stuff banners if you have the Scratch 2 to 3 addon enabled.

image

image

image

### Reason for changes

Some people (like me) dislike the default green color. Now you'll be able to change it.

> This is my first addon, so please forgive me if I've done anything wrong.


WorldLanguages:
@MystPi This is a great first contribution!

The main issue with this is that this is probably too small to be a whole addon.
@mxmou said [here](https://github.com/ScratchAddons/ScratchAddons/issues/1092#issuecomment-1029124658):
> There could even be an addon to change all the banner colors, including Scratch's own ones.

So if this was extended to change other colors, like the Explore header, it would surely be added.
Let's see what mxmou can say about the current state of this PR and then we can consider extending it.

MystPi:
> @MystPi This is a great first contribution!
> 
> The main issue with this is that this is probably too small to be a whole addon. @mxmou said [here](https://github.com/ScratchAddons/ScratchAddons/issues/1092#issuecomment-1029124658):
> 
> > There could even be an addon to change all the banner colors, including Scratch's own ones.
> 
> So if this was extended to change other colors, like the Explore header, it would surely be added. Let's see what mxmou can say about the current state of this PR and then we can consider extending it.

Thanks, and great idea! I could implement an option to change the explore header too.

RedGuy12:
does #3950 not work?

WorldLanguages:
> does #3950 not work?

That only affects variations of blue.

MystPi:
I changed the following:
- Renamed it to "customizable banner colors"
- Added the option to change the explore page's banner

lisa-wolfgang:
I'm thinking that this would fit better as a new option in the "website dark mode and customizable colors" addon. This doesn't mean that we can't reuse what's been created so far, though.

mxmou:
> I'm thinking that this would fit better as a new option in the "website dark mode and customizable colors" addon. This doesn't mean that we can't reuse what's been created so far, though.

I think a separate addon is better, but the "colored areas" setting in dark-www should remain. This means that
* we don't need to add four (or more) new settings to dark-www;
* website dark mode presets can still have a preferred banner color scheme, but it will be ignored if this addon is enabled;
* this addon can have its own presets independent from website dark mode (although the "normal", "darker" and "darker and desaturated" options should probably be duplicated here);
* colors that only appear on one page, for example pink on the WeDo page and teal on the 2020 annual report, would be affected by dark-www and don't need separate settings.

It could have the following settings:
* blue banners
* dark blue banners (could also be generated based on the blue color)
* green banners
* yellow banners
* purple banners

"Areas" might be a better word for these because the colors aren't only used for banners. All of them should also have automatically generated text colors.

The selectors from `addons/dark-www/desaturated_banners.css` could be used in this addon's code.

Samq64:
If most of this addon's code and presets are being duplicated from dark-www why not just remove the option from dark-www?

MystPi:
> If most of this addon's code and presets are being duplicated from dark-www why not just remove the option from dark-www?

I didn't take any code from dark-www, and the presets are from the default Scratch light mode.

mxmou:
> > If most of this addon's code and presets are being duplicated from dark-www why not just remove the option from dark-www?
> 
> I didn't take any code from dark-www, and the presets are from the default Scratch light mode.

I think @Samq64 was replying to my post, where I said that the code and presets from dark-www should be used. The reason for not removing the option from there is to have good defaults for each dark mode preset.

MystPi:
Bump.

WorldLanguages:
@mxmou Could you remind me your stance on this PR as it is currently? Do you think this should be merged, and if not, what changes would you make, or should this not be a separate addon at all?

mxmou:
> @mxmou Could you remind me your stance on this PR as it is currently? Do you think this should be merged, and if not, what changes would you make, or should this not be a separate addon at all?

It currently only affects three different banners - I think it should be extended to all of them using the stylesheet from dark-www.

MystPi:
I'm finished with this addon, and it is ready to merge. It can customize the colors of the explore, ideas, and my stuff page banners.

MystPi:
Any updates on this?

MystPi:
Okay, thanks for the info. 😸

WorldLanguages:
@mxmou Do you think this should be reopened? Or should we add a feature like this in the future?

Secret-chest:
Any updates? Also, I don't think dark blue should be generated, because they have a different role, you don't necessarily want to have them darker.

PR #4219, opened 2022-02-06
Error with not showing loading message

### LINKED ISSUE #4215: No "Loading messages..."
### Description

After the messaging refactor, the "Loading messages..." text no longer appears when opening the messaging popup, which makes loading seem much slower.

### Repoduction steps

1. Enable Scratch Messaging.
2. Open the popup.

### Expected behaviour

While the messages are loading, a message that says "Loading messages..." would be shown as before.

### Actual behaviour

There is no message. The bottom bar is also hidden while loading, making the popup entirely empty.

Another new problem is that there is too much padding above "Loading comments...".

### Environment

- Browser: Edge 97
- Scratch Addons version: 1.24.1
- Operating system: Windows

Norbiros:
Resolves #4215

### Changes

`After the messaging refactor, the "Loading messages..." text no longer appears when opening the messaging popup, which makes loading seem much slower. `
So now it works 😄

### DIFF
--- a/popups/scratch-messaging/popup.js
+++ b/popups/scratch-messaging/popup.js
@@ -226,7 +226,7 @@ export default async ({ addon, msg, safeMsg }) => {
       stMessages: [],
       messages: [],
       comments: {},
-      error: "loadingPage",
+      error: "notReady",
       hasCustomError: false,
 
       username: null,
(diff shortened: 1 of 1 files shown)


Norbiros:
**I created 2 changes, but only one is important (Second override first)**

Issue #4188, opened 2022-02-04
Manifest if.addonEnabled autorefresh when addon is enabled/disabled


WorldLanguages:
See #4121

WorldLanguages:
Blocked PRs until this is fixed:
#4121
#4226

apple502j:
I'll add this soon for userstyles. Userscript support is considered but will not be added soon due to its interaction with addon API for dynamic disable/enable.

PR #4226, opened 2022-02-07
Horizontal My Stuff tabs addon

### LINKED ISSUE #4196: Option to change scratchr2 navigation layout

Not everyone likes the my stuff or settings navigation layout after the `scratchr2` addon update.

Adding a toggle to `scratchr2` to move the navigation to the top of stuff or settings like it was before 1.24.

Removing the new navigation completely or making it a new addon.

mxmou:
Resolves #4196

### Changes

Adds an addon that makes the tab bar in My Stuff horizontal.
Without scratchr2 addon
With scratchr2 addon

Enabling or disabling 2.0 → 3.0 currently breaks it (#4188).


Norbiros:
❤️  I love that addon!

WorldLanguages:
> Enabling or disabling 2.0 → 3.0 currently breaks it (#4188).

Fixed by #4289

addons/horizontal-mystuff-tabs/addon.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Horizontal tabs on My Stuff",
+  "description": "Moves the My Stuff navigation buttons (\"All Projects\", \"My Studios\", \"Trash\") to the top of the page, instead of keeping them on the left.",
+  "tags": ["community"],
```
WorldLanguages, line 4: @mxmou Any reason this shouldn't be a theme?

Samq64:
Should this addon also affect Account settings?

WorldLanguages:
> Should this addon also affect Account settings?

See https://github.com/ScratchAddons/ScratchAddons/issues/4196#issuecomment-1030779787

RedGuy12:
I think it would be good even though the pills are on the other side

addons/horizontal-mystuff-tabs/addon.json
mxmou: It could be - editor-stage-left has a similar purpose and is a theme.

mxmou:
> Should this addon also affect Account settings?

This addon is mainly intended for people who are used to the previous behavior of the Scratch 2.0 → 3.0 addon, which moved My Stuff tabs to the top and Account Settings tabs to the right. There is no good reason to put tabs on the right, so this addon only reverts the My Stuff change. It could also provide an option to make the Account Settings tabs horizontal, but it should probably be disabled by default.

PR #4233, opened 2022-02-07
Improve install from source for Firefox

Chiroyce1:
### Changes
Changes the wording for step 1 a bit, and changes `right-hand` to `left-hand` in step 2.
The instructions for Firefox in the [README](https://github.com/ScratchAddons/ScratchAddons#from-source) tells users to click a button the the right-hand menu, but it actually is in the left now. 
image

### Reason for changes
Improving instructions to install the addon from source.

### Tests
This doesn't change any file except `README.md`, so I haven't tested it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ After downloading the extension to your computer, just load it by following thes
   3. Click the `Load unpacked` button on the top and select the extension directory which has the `manifest.json` file in it. 
 
 - Mozilla Firefox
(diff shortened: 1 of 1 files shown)


WorldLanguages:
https://scratchaddons.com/docs/getting-started/installing/ should also be updated if necessary.

Chiroyce1:
> https://scratchaddons.com/docs/getting-started/installing/ should also be updated if necessary.

I don't think so, it seems to correct.
image

PR #4234, opened 2022-02-07
scratch-messaging: hide Delete when inapplicable

apple502j:
Hides Delete button for comments in others' profile/project. Not doing this for studio comments, since that'll require an additional API request just for that.


WorldLanguages:
There's a bug, you now cannot delete comments you just posted a second ago within Scratch Messaging, even if it's inside your own profile/project. I'll fix it.

Issue #331, opened 2020-09-26
Scratch Messaging: user might be logged out if it tries to load comments in unshared project


WorldLanguages:
This is because of a bug in Scratch, click this link and you'll be logged out: https://scratch.mit.edu/site-api/comments/user/kaj/

We can avoid this by using the `chrome.webRequest` API, and remove or edit the `Set-Cookie` header. I'm not sure if browsers allow to avoid cookie sets this way, tho.

Issue #3615, opened 2021-10-06
Chrome: use declarativeNetRequestWithHostAccess instead of webRequestBlocking


WorldLanguages:
EDIT: outdated

The new `declarativeNetRequestWithHostAccess` permission will ship in Chrome stable on November 16 2021.
Around three months later (February 2022), we'll remove the `webRequestBlocking` permission from our Chrome manifest.json and start using `declarativeNetRequestWithHostAccess` instead.
Users running Chrome 95 and below will still be able to use Scratch Addons by then, except some scratch-messaging and scratch-notifier features (more detail below).
Firefox will continue to use `webRequestBlocking`, possibly forever.

We currently use blocking webRequest for:
- Fetching `?sareferer` URLs (used to delete/post comments and clear message count from background/popup contexts)
- ~~Removing `Set-Cookie` headers from `https://scratch.mit.edu/site-api/comments/*` responses which unexpectedly log users out.~~ **Removed in v1.24.3 through #4237 and no bug reports received about unexpected logouts yet.**

Moving away from webRequestBlocking is a good first step to be ready for manifest v3, and may also possibly speed up our Chrome Web Store review times.

WorldLanguages:
> (added by quoter: this probably means that profile comments won't work)

No, they'll still work. All we did to those URLs is prevent them from logging the user out.
URL param `sareferer` is used for:
- Marking messages as read (scratch messaging and notifier)
- Posting comments, deleting comments, clearing alerts (scratch messaging)

RedGuy12:
Maybe we can change webRequestBlocking to an optional permission and only request it if declarativeNetRequestWithHostAccess isn't supported


________________________________
From: World_Languages ***@***.***>
Sent: Wednesday, October 6, 2021 5:53:23 PM
To: ScratchAddons/ScratchAddons ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [ScratchAddons/ScratchAddons] Chrome: use declarativeNetRequestWithHostAccess instead of webRequestBlocking (#3615)


The new declarativeNetRequestWithHostAccess permission will ship in Chrome stable on November 16 2021.
Around three months later (February 2022), we'll remove the webRequestBlocking permission from our Chrome manifest.json and start using declarativeNetRequestWithHostAccess instead.
Users running Chrome 95 and below will still be able to use Scratch Addons by then, except some scratch-messaging and scratch-notifier features (more detail below).
[...]

PR #4237, opened 2022-02-07
Remove unnecessary background/handle-cookie.js

WorldLanguages:
This file exists because of #331
PR #3625 stopped sending credentials, so the Set-Cookie header is now ignored.

Keeping the file means we fix a bug in Scratch (users get logged out if they visit https://scratch.mit.edu/site-api/comments/user/kaj/) but it's not worth it. We only added the fix because we caused the login, but that won't happen anymore.


WorldLanguages:
Releasing inside v1.24.2 so that we are ready for #3615

WorldLanguages:
Let's hope we don't need to release v1.24.3 because of this 🙏

PR #4244, opened 2022-02-08
Compute addon.tab.clientVersion only when the addon needs it

### LINKED ISSUE #4212: 'Cannot access Blockly on this page' Error causing addons not to load
Sometimes, when switching from the project page to the editor, many addons will fail to work properly and the error below will be spammed in the console. I have seen this happen on at least three different occasions but, am unable to reliably reproduce it.

image

Firefox 96.0 64-bit
Ubuntu 20.04.3 (LTS)

WorldLanguages:
Mitigates #4212

Note: commit name is inaccurate, every addon will compute clientVersion when requested for the first time, and it will be cached only for the addon that requested it.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -23,14 +23,18 @@ export default class Tab extends Listenable {
   constructor(info) {
     super();
     this._addonId = info.id;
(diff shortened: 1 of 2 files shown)


GrahamSH-LLK:
I'd say keep the issue open

apple502j (approved):

PR #4256, opened 2022-02-09
New "orange" color preset for msg count badge

devxan:
Resolves the issue of the message count badge looking wrong. 

### Changes
Added "Orange" preset to message count badge

### Reason for changes
- The badge doesn't match the extension logo 
	- Other extensions with badges on my device use colors similar to their logos(/icons).
- This adds an SA-color orange preset to make the logo and badge look good together.

### Tests
It works. ☑️

### DIFF
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -40,6 +40,14 @@
       "values": {
         "color": "#000000"
       }
+    },
+    {
+      "name": "Orange",
+      "id": "orange",
+      "description": "The same color as the Scratch Addons logo.",
+      "values": {
+        "color": "#FF7B26"
+      }
     }
   ],
   "credits": [
@@ -53,6 +61,9 @@
     {
       "name": "lisa_wolfgang",
       "link": "https://scratch.mit.edu/users/lisa_wolfgang"
-    }
+    },
+    {
+      "name": "Xan"
+      "link": "https://scratch.mit.edu/users/-Xanimation-"
   ]
 }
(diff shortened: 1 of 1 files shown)


addons/msg-count-badge/addon.json
```diff
@@ -53,6 +61,10 @@
     {
       "name": "lisa_wolfgang",
       "link": "https://scratch.mit.edu/users/lisa_wolfgang"
+    },
+    {
+      "name": "Xan"
```
RedGuy12, line 66: you did nothing

WorldLanguages: lol

devxan: sad 😢

GrahamSH-LLK:
> > Message counts usually need to stand out - other extensions probably don't show very important information in badges, so using the same color as the logo makes sense. This could be added, but I don't think it's very useful.
> 
> Other extensions know what they're doing with their badges. This is such a small change (and keep in mind that it's entirely optional like everything else in the extension besides the "Addons" tab in the extension popup) that it wouldn't really matter if it was in there for many people. But for people such as myself who like consistency, this is awesome.

The thing is, people can add it themselves

devxan:
> The thing is, people can add it themselves

This just makes it easier. Most people won't think to look at [the Scratch Addons docs](https://scratchaddons.com/docs/reference/design/) for the color, and this orange is mentioned on that page, not any greens or other stuff currently in the presets.

WorldLanguages:
I think this is good enough

devxan:
Awesome

PR #4264, opened 2022-02-12
Fix word-wrapping for Ocular Status in forums

### LINKED ISSUE #4262: my-ocular+scratchr2: ocular statuses can overflow horizontally
Likely regressed by #4227

Only happens if the ocular status has a very long word without any spaces.

image

Chiroyce1:
Resolves #4262 

### Changes

In the [addons/my-ocular/discuss.js](https://github.com/Chiroyce1/ScratchAddons/commit/97685481256ca1c2a61d0c9b056c581879632772#diff-822fae7015766686e1951e78da4e29c345e0ea347c2c567afaa3a7aba0def7ec) file, it adds a class to the ocular status and in [addons/my-ocular/discuss.js](https://github.com/Chiroyce1/ScratchAddons/commit/97685481256ca1c2a61d0c9b056c581879632772#diff-822fae7015766686e1951e78da4e29c345e0ea347c2c567afaa3a7aba0def7ec) it adds `word-wrap: break-word;` for the `ocular-status` class.

### Reason for changes

Resolve word wrapping issue

### Tests

Tested on Firefox, seems to work.
image

### DIFF
--- a/addons/scratchr2/forums.css
+++ b/addons/scratchr2/forums.css
@@ -375,3 +375,7 @@ body > #pagewrapper {
   color: white;
   line-height: 32px;
 }
+
+.ocular-status {
+  word-wrap: break-word;
+}
--- a/addons/my-ocular/discuss.js
+++ b/addons/my-ocular/discuss.js
@@ -13,6 +13,7 @@ export default async function ({ addon, global, console, msg }) {
       let br = document.createElement("br");
       addon.tab.displayNoneWhileDisabled(br);
       let status = document.createElement("i");
+      status.setAttribute("class", "ocular-status")
       addon.tab.displayNoneWhileDisabled(status);
       status.title = msg("status-hover");
       status.innerText = userStatus;


mxmou:
The bug could theoretically also affect the group and post count - could you add the `word-wrap` [here](https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/scratchr2/forums.css#L251) instead?

addons/scratchr2/forums.css
```diff
@@ -375,3 +375,7 @@ body > #pagewrapper {
   color: white;
   line-height: 32px;
 }
+
+.ocular-status {
+  word-wrap: break-word;
```
mxmou, line 380: The standard name for this property is `overflow-wrap`. It works in all browser versions supported by SA.

Chiroyce1:
@mxmou I made those changes in a new branch called [`overflow-wrap`](https://github.com/Chiroyce1/ScratchAddons/tree/overflow-wrap), can you test it out? I'll merge it with [`master`](https://github.com/Chiroyce1/ScratchAddons) after that.

addons/scratchr2/forums.css
Chiroyce1: Oops,

mxmou (approved):

PR #4267, opened 2022-02-13
New addon: default-costume-editor-color

GarboMuffin:
It changes the default colors and outlline size used by the paint editor

### DIFF
--- a/addons/default-costume-editor-color/addon.json
+++ b/addons/default-costume-editor-color/addon.json
@@ -34,7 +34,7 @@
       "default": 4
     },
     {
-      "name": "Use previous color instead of default color when resetting color after switching tools",
+      "name": "Use previous color instead of default colors when switching tools",
       "id": "persistence",
       "type": "boolean",
       "default": true
--- a/addons/default-costume-editor-color/addon.json
+++ b/addons/default-costume-editor-color/addon.json
@@ -1,0 +1,42 @@
+{
+  "name": "Customize default costume editor colors",
+  "description": "Changes the default colors and outline size used by the costume editor.",
+  "credits": [
+    {
+      "name": "GarboMuffin"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Default fill color",
+      "id": "fill",
+      "type": "color",
+      "default": "#9966FF"
+    },
+    {
+      "name": "Default outline color",
+      "id": "stroke",
+      "type": "color",
+      "default": "#000000"
+    },
+    {
+      "name": "Default outline size",
+      "id": "strokeSize",
+      "type": "integer",
+      "min": 0,
+      "max": 100,
+      "default": 4
+    }
+  ],
+  "versionAdded": "1.25.0",
+  "tags": ["editor", "costumeEditor"],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "enabledByDefault": false
+}
--- a/addons/default-costume-editor-color/addon.json
+++ b/addons/default-costume-editor-color/addon.json
@@ -32,6 +32,12 @@
       "min": 0,
       "max": 100,
       "default": 4
+    },
+    {
+      "name": "Use previous color instead of default color when resetting color after switching tools",
+      "id": "persistence",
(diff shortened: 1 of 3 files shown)


GarboMuffin:
The primary reason this is marked as a draft, aside from lack of thorough testing, is because I'm considering another option/feature.

When you make a selection that causes your current color to become "mixed" then switch back to a color-based tool like the paint brush, wouldn't it make more sense to apply the previous color you had selected instead of a single specific default color?

Mbrick2:
Maybe defaults for each tool and options to reset to default after changing tool

GarboMuffin:
I'm guessing the addon and setting names etc. will be changed, but otherwise I believe this is complete

addons/default-costume-editor-color/addon.json
```diff
@@ -0,0 +1,48 @@
...
+      "name": "Default outline size",
+      "id": "strokeSize",
+      "type": "integer",
+      "min": 0,
+      "max": 100,
+      "default": 4
+    },
+    {
+      "name": "Use previous color instead of default color when resetting color after switching tools",
+      "id": "persistence",
```
WorldLanguages, line 38: The persistence only applies to the fill color, not to outline and stroke, correct?

addons/default-costume-editor-color/userscript.js
```diff
@@ -0,0 +1,135 @@
...
+    if (!addon.self.disabled) {
+      applyFillColor();
+      applyStrokeColor();
+      applyStrokeWidth();
+    }
+  };
+  addon.settings.addEventListener("change", () => {
+    setDefaultColorsToSettings();
+    applyEverything();
+  });
```
WorldLanguages, line 74: Not sure if this is intentional or not, but changing the default fill color also changes the currently selected fill color in all open editors, even if the default color isn't currently selected.

addons/default-costume-editor-color/addon.json
GarboMuffin: It's currently supposed to apply to fill, stroke color, and stroke width.

WorldLanguages: I think I'm confused since I don't know under what situations it's supposed to return to the previous colors and width. How are you testing this?

GarboMuffin:
Due to very significant time constraints, this addon with its current feature set is unlikely to be ready by Saturday

If "Use previous color instead of default color when resetting color after switching tools" is removed (it's a bit of scope for this addon if we're being honest, and it's quite buggy and hard to explain), then I can get it done very quickly (basically just reverting to 6fa887687d54b5536e76a5ba14155179d5ed2503) Of course, that option could still be added as its own addon or added as an option in a later update

thoughts?

WorldLanguages:
We can wait until v1.26 I think

addons/default-costume-editor-color/addon.json
GarboMuffin: For fill:

1. Make a new project and open the costume editor on the cat
2. Change the default fill color to something that isn't the default
3. Do Ctrl+A
4. Notice how Scratch has reset your fill color to a generic "mixed color" icon
5. Switch to the paint brush tool. Scratch would normally reset your color to the default purple, but with the option enabled, it will instead use the color you had selected before step 3.

For stroke/outline: Same as above but using vector rectangle tool instead of paint brush tool.

It's buggy right now, but that's what it's supposed to do.

GarboMuffin:
Some changes have been made to how the `persistence` option works. It supports gradients and should be more intuitive. For example, if you have both fill and stroke on the rectangle tool, make a selection, and then go back to the rectangle tool, you won't lose the stroke (previously it would only restore the fill).

I think this addon is basically done now

WorldLanguages:
@GarboMuffin I couldn't get it to restore the stroke width in a specific situation:
1. Go to https://scratch.mit.edu/projects/104/editor/ (girl costume)
2. Convert the costume to bitmap
3. Create a square
4. Select the girl bitmap (now an object inside the SVG)
5. Click the square tool again. The fill and stroke colors are restored, but not the stroke width. The stroke width stays at 0.

I now see that you used the Scratch cat costume to test this and I used a bitmap object inside the SVG. That kinda explains why I was confused previously.
I don't think what I mentioned above is 100% a bug but we should be aware of it at least.

PR #4228, opened 2022-02-07
 Featured Button obscures ocular statuses (ScratchAddons#2464) 

Norbiros:
Resolves #2464

### Changes

The text wrap around the button, not be behind it.

### Reason for changes

The "change featured project" button obscures parts of long ocular statuses on profile pages if my-ocular and better-featured-project are enabled.

PR #4269, opened 2022-02-14
Fixed some problems with ocular status (#2464)

### LINKED ISSUE #2464: better-featured-project obscures long ocular statuses
**Describe the bug**

The "change featured project" button obscures parts of long ocular statuses on profile pages if my-ocular and better-featured-project are enabled.

**To Reproduce**

1. Enable better-featured-project and my-ocular
2. Set your ocular status to something very long
3. Go to your profile page
4. See that part of the status is behind the button

**Expected behavior**

The text should wrap around the button, not be behind it.

**Screenshots**

overlap

- OS: Windows 10
- Browser: Firefox
- Version: 88.0

(My browser / operating system: Windows NT 10.0, Firefox 88.0, No Flash version detected)

Norbiros:
Resolves #2464

More info in https://github.com/ScratchAddons/ScratchAddons/pull/4228

**I added changes in to banner addon, because ocular works normally without it.**

### DIFF
--- a/addons/better-featured-project/userstyle.css
+++ b/addons/better-featured-project/userstyle.css
@@ -144,6 +144,6 @@
 }
 .location {
   word-break: break-all;
-  width: 660px;
+  width: 600px;
   margin-left: 15px;
 }
--- a/addons/better-featured-project/userstyle.css
+++ b/addons/better-featured-project/userstyle.css
@@ -142,3 +142,8 @@
 .sa-user-id {
   color: white;
 }
+.location {
+  word-break: break-all;
+  width: 660px;
+  margin-left: 15px;
+}
--- a/addons/better-featured-project/userstyle.css
+++ b/addons/better-featured-project/userstyle.css
@@ -145,5 +145,4 @@
 .location {
   word-break: break-all;
   width: 600px;
-  margin-left: 15px;
 }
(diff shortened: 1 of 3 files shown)


addons/better-featured-project/userstyle.css
```diff
@@ -142,3 +142,8 @@
 .sa-user-id {
   color: white;
 }
+.location {
+  word-break: break-all;
+  width: 660px;
```
WorldLanguages, line 147: This width would need to change depending on the width of the `Change featured project` button (doing this dynamically with JavaScript would be overkill, let's not do that). Some languages might use a longer string for this. So I suggest using a smaller widththat will work for the vast majority of languages, maybe 600px.

WorldLanguages:
@mxmou What do you think about this solution?

addons/better-featured-project/userstyle.css
```diff
@@ -142,3 +142,8 @@
 .sa-user-id {
   color: white;
 }
+.location {
+  word-break: break-all;
+  width: 600px;
+  margin-left: 15px;
```
WorldLanguages, line 148: The `margin-left` makes it look wrong if scratchr2 is not enabled:

image

Norbiros: Okay, so I will create another CSS files that works when this addon is enabled...

Norbiros:
@WorldLanguages 
I fixed that problem with margin. PR is ready to merge...

PR #4288, opened 2022-02-19
Not intengers labels on graph

### LINKED ISSUE #4270: Non-integer labels on follower graph
### Description

On profiles which have gained or lost very few followers, the labels on their follower graphs include non-integer numbers.

### Reproduction steps

1. Go to someone's profile (for example, [@Hello_world](https://scratch.mit.edu/users/Hello_world/))
2. View the follower graph
3. Read the number labels

### Expected behavior

Numerical labels on follower graphs should be integers only.

### Actual behavior

Non-integer labels are displayed.

### Screenshots

Screenshot 2022-02-14 at 13-39-07 Scratch - Imagine, Program, Share

### Environment

- Browser: Firefox 97
- Scratch Addons version: 1.24.3
- Operating system: Windows 10

Norbiros:
Resolves #4270

### Changes

It hides not integer labels, also I parsed numbers (number with comma) for all American people out there...

### Tests

Downloaded and tested on Firefox.

### DIFF
--- a/addons/scratchstats/userscript.js
+++ b/addons/scratchstats/userscript.js
@@ -114,7 +114,14 @@ export default async function ({ addon, msg, console }) {
                   },
                 },
                 y: {
-                  stepSize,
+                  precision: 0,
+                  ticks: {
+                    callback: function(stepSize) {
+                       if (!(stepSize%1)) {
+                         return new Intl.NumberFormat().format(Number(stepSize).toFixed(0));
+                       }
+                    }
+                  }
                 },
               },
               plugins: {
--- a/addons/scratchstats/userscript.js
+++ b/addons/scratchstats/userscript.js
@@ -116,12 +116,12 @@ export default async function ({ addon, msg, console }) {
                 y: {
                   precision: 0,
                   ticks: {
-                    callback: function(stepSize) {
-                       if (!(stepSize%1)) {
-                         return new Intl.NumberFormat().format(Number(stepSize).toFixed(0));
-                       }
-                    }
-                  }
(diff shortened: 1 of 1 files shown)


addons/scratchstats/userscript.js
```diff
@@ -114,7 +114,14 @@ export default async function ({ addon, msg, console }) {
                   },
                 },
                 y: {
-                  stepSize,
+                  precision: 0,
+                  ticks: {
+                    callback: function (stepSize) {
```
WorldLanguages, line 119: The `stepSize` variable defined above (not the function argument) is now unused. Can you remove line 91?

WorldLanguages:
> I parsed numbers (number with comma) for all American people out there

Could you explain the change in more detail? (explain in what situations the numbers will be different before vs now)

apple502j:
I think there is a simpler solution, closing

Norbiros:
Okay, I will, sorry but I forgot about that.

@apple502j 
Can you explain? What solution? I looked on a lot of issues about that... I think only better solution is in some kind optimizing it?

Norbiros:
> > I parsed numbers (number with comma) for all American people out there
> 
> Could you explain the change in more detail? (explain in what situations the numbers will be different before vs now)

Before we had `100, 000` for example, and this code `Intl.NumberFormat().format()` make it looks the same as before.

addons/scratchstats/userscript.js
Norbiros: Done

WorldLanguages:
This PR is replaced by #4314

PR #4289, opened 2022-02-19
Add partial dynamicEnable/dynamicDisable for userstyles

### LINKED ISSUE #4188: Manifest if.addonEnabled autorefresh when addon is enabled/disabled

apple502j:
Resolves #4188

Took 3 hours to implement and 30 minutes to write this.

# Major Changes
- `scratchAddons.dependents`: `object<Set<string>>` - set of addon IDs that declare the addon as `addonEnabled` dependency for userstyles.
- `getAddonData` and other places that handle an array of userstyles now receive an array of objects instead of strings. The object always has `href` (absolute URL to the style) and `index` (index in manifest, does not change if userstyle gets enabled/disabled), and additional `text` for `injectAsStyleElt` addons.
- Due to internal code changes, there's a maximum limit of userstyles per addon. Exceeding this won't throw error, but may cause unreliable style priority on some situations. Currently this is 100. If this is ever reached, a warning will be logged to the page console.
- `disabledDynamicAddons` and `everLoadedAddons` are now sets of addon IDs, which should fix issues with duplicate IDs. This should also make has/delete calls faster.
- And the big one, introducing partial dynamicEnable/dynamicDisable! This is like dynamicEnable (PDE)/dynamicDisable (PDD) but for part of the addon (specifically userstyles). See below

## PDE
PDE occurs when a dynamicEnable addon (A)'s userstyle depends on addon B, and B gets enabled while A is enabled. PDE doesn't emit reenabled event on addons. Note that PDE only occurs when enabling such addons caused a new userstyle to load. (For example, if A had two styles, X and Y, and X is always enabled while Y depends on (both disabled) addon B and C - if B is enabled, PDE attempt occurs at `change-addon-state.js`, but since C is disabled the style Y is still disabled. Thus, there are no new styles (since X is already enabled) and no PDE occurs. This is `get-userscripts.js` L39)

## PDD
PDD is a type of autism... wait no, not that.

PDD is the opposite of PDE, that occurs when a dynamicDisable addon (A)'s userstyle depends on addon B, and B gets disabled while A is enabled. PDD doesn't emit events. If there are remaining styles or loaded scripts, it just disables affected styles, removes them from the array, and early returns. (`cs.js` L448) However, if there are no scripts or styles left after disabling the affected styles, the addon is considered entirely disabled and is removed from `addonsWithUserstyles`, etc.

Normal dynamicDisable may follow PDD if someone disabled B and then A. To not "double disable" an addon, we first check if it's currently entirely disabled via PDD, and early-returns if so. (`cs.js` L431)

### DIFF
--- a/background/declare-scratchaddons-object.js
+++ b/background/declare-scratchaddons-object.js
@@ -9,7 +9,7 @@ scratchAddons.localEvents = new EventTarget();
 
 // Load manifests into memory
 scratchAddons.manifests = [];
-// addonId to array of addon IDs that has the addon inside userstyle's settings.if.addonEnabled
+// addonId to set of addon IDs that has the addon inside userstyle's settings.if.addonEnabled
 scratchAddons.dependents = {};
 
 // Other files may add their own global methods here so that addon-api files can access them
--- a/background/declare-scratchaddons-object.js
+++ b/background/declare-scratchaddons-object.js
@@ -9,6 +9,8 @@ scratchAddons.localEvents = new EventTarget();
 
 // Load manifests into memory
 scratchAddons.manifests = [];
+// addonId to array of addon IDs that has the addon inside userstyle's settings.if.addonEnabled
+scratchAddons.dependents = {};
 
 // Other files may add their own global methods here so that addon-api files can access them
 scratchAddons.methods = {};
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -30,7 +30,7 @@ scratchAddons.localEvents.addEventListener("addonDynamicEnable", ({ detail }) =>
               // Handle partial dynamic enable (PDE)
               // Userscripts currently cannot be PDEd
               // Note: this can still result in userstyles being empty
-              // if a userstyle depends on multple addons
+              // if a userstyle depends on multiple addons
               // If no running userstyles depend on the provided dependency,
               // this means no new userstyles are loaded, thus early return.
               // We still need to send the whole array.
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -18,7 +18,7 @@ chrome.runtime.onMessage.addListener(function (request, sender, sendResponse) {
 });
 
 scratchAddons.localEvents.addEventListener("addonDynamicEnable", ({ detail }) => {
-  const { addonId, manifest } = detail;
+  const { addonId, manifest, partialDynamicEnableBy } = detail;
   chrome.tabs.query({}, (tabs) =>
     tabs.forEach((tab) => {
       if (tab.url || (!tab.url && typeof browser !== "undefined")) {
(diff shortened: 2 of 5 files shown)


background/get-userscripts.js
```diff
@@ -110,25 +134,34 @@ async function getAddonData({ addonId, manifest, url }) {
...
+                href: styleHref,
+                text,
+                index: i,
+              };
             })
         );
       } else {
-        userstyles.push(chrome.runtime.getURL(`/addons/${addonId}/${style.url}`));
+        userstyles.push({
+          href: chrome.runtime.getURL(`/addons/${addonId}/${style.url}`),
```
RedGuy12, line 162: Move `styleHref` out of the `if` and use it here.

background/declare-scratchaddons-object.js
```diff
@@ -9,6 +9,8 @@ scratchAddons.localEvents = new EventTarget();
 
 // Load manifests into memory
 scratchAddons.manifests = [];
+// addonId to set of addon IDs that has the addon inside userstyle's settings.if.addonEnabled
```
RedGuy12, line 12: Can we not auto-generate this?

RedGuy12: Oh wait we do nvm

WorldLanguages:
@mxmou After this is merged, do you think we should move addon dark mode styles to the addons themselves instead of having the styles inside dark-www? Before the feature this PR adds, that was the only alternative, but that won't be the case anymore.

mxmou:
> @mxmou After this is merged, do you think we should move addon dark mode styles to the addons themselves instead of having the styles inside dark-www? Before the feature this PR adds, that was the only alternative, but that won't be the case anymore.

The CSS variables already allow us to do that, at least for some addons.

content-scripts/cs.js
```diff
@@ -366,33 +383,79 @@ async function onInfoAvailable({ globalState: globalStateMsg, addonsWithUserscri
...
+      if (disabledDynamicAddons.has(addonId)) return;
+      const scriptIndex = addonsWithUserscripts.findIndex((a) => a.addonId === addonId);
+      const styleIndex = addonsWithUserstyles.findIndex((a) => a.addonId === addonId);
+      if (partialDynamicDisabledStyles) {
+        // Userstyles are partially disabled.
+        // This should not result in other parts being disabled,
+        // unless that means no scripts/styles are running on this page.
+        removeAddonStylesPartial(addonId, partialDynamicDisabledStyles);
+        if (styleIndex > -1) {
+          // This should exist... right? Safeguarding anyway
```
WorldLanguages, line 440: I'm reaching this if condition.

content-scripts/cs.js
```diff
@@ -366,33 +383,79 @@ async function onInfoAvailable({ globalState: globalStateMsg, addonsWithUserscri
...
+          const userstylesEntry = addonsWithUserscripts[styleIndex];
```
WorldLanguages, line 441: It should be `addonsWithUserstyles`.

content-scripts/cs.js
```diff
@@ -366,33 +383,79 @@ async function onInfoAvailable({ globalState: globalStateMsg, addonsWithUserscri
...
+      disabledDynamicAddons.delete(addonId);
       addStyle({ styles: userstyles, addonId, injectAsStyleElt, index });
-      if (everLoadedAddons.find((addon) => addon.addonId === addonId)) {
-        if (!dynamicDisable) return;
-        // Addon was reenabled
-        _page_.fireEvent({ name: "reenabled", addonId, target: "self" });
+      if (partial) {
+        // Partial: part of userstyle was (re-)enabled.
+        // No need to deal with userscripts here.
+        const addonsWithUserstylesEntry = addonsWithUserscripts.find((entry) => entry.addonId === addonId);
```
WorldLanguages, line 402: You probably also meant `addonsWithUserstyles` here.

WorldLanguages:
PDE doesn't seem to work, even by replacing `addonsWithUserscripts` with `addonsWithUserstyles` (see 2 code comments above).

Here's my repro:
1. Go to dark-www addon.json, add `"addonEnabled": ["pause"]` to scrollbar.css
2. Go to Scratch homepage with pause addon enabled
3. Disable pause addon - scrollbar.css is correctly removed, PDD appears to work
4. Enable pause addon - no changes are made, scrollbar.css should be reinjected

content-scripts/cs.js
WorldLanguages: See repro in issue comment (not code comment) below to see how to reach this if case.

apple502j:
Hmm, not sure if I made it work with addons inapplicable for the page. Checking.

PR #4318, opened 2022-02-23
fix too much padding between credits

RedGuy12:
it looks too ugly before this

### DIFF
--- a/webpages/settings/components/addon-body.html
+++ b/webpages/settings/components/addon-body.html
@@ -40,14 +40,11 @@
         </div>
       </div>
       <div class="addon-credits" v-if="addon.credits">
-        <span>{{ msg("creditTo") }}</span>
-        <span v-for="credit of addon.credits">
-          <span v-if="credit.link">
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Screenshots before and after?

apple502j:
Before:
image

After:
image

Hans5958:
I suppose it's need a little more, it looks quite cramped.

RedGuy12:
@Hans5958 how it is now is just leaving it to to the browser...

mxmou:
> @Hans5958 how it is now is just leaving it to to the browser...

Which isn't enough - there should be more than one space between each pair of items to make the more readable. About half the current padding would be better.

lisa-wolfgang:
Personally, I think a vertical list would look nicer than just sprawling them across the screen.
It could be a left sidebar on wider displays.

Hans5958:
Probably space considerations. Doing it vertically would waste more space.

lisa-wolfgang:
The settings page needs a redesign at some point regardless. I don't think there's much point in scrutinizing over the old design when there are much larger issues such as addon discoverability and theme separation.

mxmou:
The current design is better.

PR #4320, opened 2022-02-23
add cmd key support

### LINKED ISSUE #4299: [mute-project]: On Mac, command key doesn't work
I can't reproduce this since I don't have a mac, but we would just need to modify this line https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/mute-project/userscript.js#L9

Dhuls-Scratch:
Resolves #4299

### Changes

basically just used what was in /addons/ctrl-enter-post/forums.js

### Reason for changes

So Mac users can use the CMD key.

### Tests
can't test since I don't have a Mac

### DIFF
--- a/addons/mute-project/userscript.js
+++ b/addons/mute-project/userscript.js
@@ -6,7 +6,7 @@ export default async function ({ addon, global, console }) {
   icon.loading = "lazy";
   icon.style.display = "none";
   const toggleMute = (e) => {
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Well, this has the side effect of also affecting the windows key, but I don't think we should care

Chiroyce1:
Tested on Firefox on macOS, works.

PR #4321, opened 2022-02-23
Change regex for comments that mention "scratch addons"

WorldLanguages:
This message now appears when linking to the Chrome Web Store link for scratch addons, now that the regex is more strict:

image


RedGuy12:
https://chrome.google.com/webstore/detail/fbe... also redirects to the install

WorldLanguages:
@RedGuy12 We should be censoring comments the same way as Scratch does (we are telling the user we avoided a mute), and the URL without `scratch-addons` isn't censored by Scratch.

PR #4322, opened 2022-02-23
Add some JSDoc

RedGuy12:
Alright, let's try this again. This time, I only made changes to comments and other required things for this to work. No `?.` everywhere.

I know that WL doesn't want tons of files in the root directory (that's why he keeps changing his mind about whether he likes #3968), and this adds `jsconfig.json`. So I moved `.eslintignore` into `.eslintrc` so the total number of files doesn't go up.

`jsconfig.json` isn't actually required for this to work but if it wasn't there the types couldn't be global and you'd have to import them in each file which WL said he didn't like. Also, it could be shorter, yes, but I added all the options so if someone (aka me) wants to make an addon that has type-checking via JSDoc they can by adding `// @ts-check` to the top of the file. This is opt-in so addon creators don't have to worry about types unless they want to.

Also, if we wanted, we could use JSDoc to dynamically generate docs. That way we don't have outdated stuff in there all the time.


apple502j:
Can you split this PR into addon-api docs, TypeScript stuff, and addon changes? This PR is too big and hard to review, and addon-api is more important than other stuff. I don't think we have to add that code to all addons now, either - we could add it along with other changes to the addon. No need to edit the file just for that.

RedGuy12:
I'll make new pr "at some point"

PR #4335, opened 2022-02-26
Change minimum supported Firefox version, remove unused code

### LINKED ISSUE #4334: Change minimum Firefox required version to 86 (previously 74)
Current minimum FF version is 74 (March 2020).
Unlike chromebook users, an outstanding majority of Firefox users can get the latest version of it easily. So I suggest changing the minimum to Firefox 86 (February 2021).

- `tabs` permission not required when host permissions are already present. We can now remove many `typeof browser` checks in core code ([Firefox 86+](https://blog.mozilla.org/addons/2021/02/09/extensions-in-firefox-86/))
- CSS `min()`, `max()` and `clamp()` can now be used (Firefox 75+)
- `Intl.DateTimeFormat` dateStyle and timeStyle parameter support, already in use by some addons (Firefox 79+)
- `<link rel="preload">` support: already in use by our settings page to load faster (Firefox 85+)

WorldLanguages:
Resolves #4334 

Firefox 86 gives us access to the URL of tabs we already have a host permission for, so we can avoid sending messages to all tabs in Firefox, and instead do the same thing we've always done in Chrome, which is only messaging scratch.mit.edu tabs.

### DIFF
--- a/addon-api/popup/Popup.js
+++ b/addon-api/popup/Popup.js
@@ -23,10 +23,7 @@ export default class Popup {
     return new Promise((resolve) => {
       chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
         if (tabs.length === 0) return resolve(null);
-        chrome.tabs.sendMessage(tabs[0].id, "getLocationHref", { frameId: 0 }, (url) => {
-          if (chrome.runtime.lastError) return resolve(null);
-          resolve(url);
-        });
+        return resolve(tabs[0]?.url || null);
       });
     });
   }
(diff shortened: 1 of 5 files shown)


Samq64:
> this ff version will literally be 10 days old when we release the next sa version...pls wait until 1.26.0 for this

He said 86. 87 is the newest version.

apple502j (approved):
LGTM, we could remove other code that queries url using messages (I think we have a couple) but this change looks fine.

WorldLanguages:
> we could remove other code that queries url using messages

Which ones are you referring to? I only found `getLocationHref` which is only used once. `getInitialUrl` gets the **initial** URL of the tab so it cannot be replaced by chrome.tabs API.

apple502j:
@WorldLanguages Hm, think it's the only one. (Also we need to rework cloud-games, yeah)

PR #4339, opened 2022-02-27
Settings page - CSS tweaks - Clickable controls now use pointer cursor

Ascor8522:
## Reason for changes

It adds more consistency and better accessibility to the controls used in the settings page.
All the controls (buttons, reset input buttons, open color picker button, etc.) now show a pointer mouse cursor (pointing hand) when hovered.
It's now clearer that those controls are clickable.
It is basically the point of that cursor, so we might as well use it.

Suggested those changes back in August this year.
Didn't get a lot of feedback, but was mainly positive overall.
Even if this was never really requested, it's still a nice quality of life addition and doesn't hurt.

### DIFF
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -74,7 +74,6 @@ body {
   font-size: 12px;
   padding-inline-start: 10.5px;
   padding-inline-end: 12px;
-  cursor: pointer;
   background-color: var(--button-background);
   border: 1px solid var(--control-border);
   border-bottom: none;
@@ -83,6 +82,9 @@ body {
 }
 .popup-name:hover {
   background-color: var(--button-hover-background);
+}
+.popup-name:not(.sel):hover {
+  cursor: pointer;
 }
 .popup-name:not(:last-child) {
   border-inline-end: none;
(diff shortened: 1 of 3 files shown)


mxmou:
I like this. It would be nice if selected `.filter-option`s were consistent with selected tabs in the popup - both are technically clickable, but clicking them doesn't do anything, so they should have the same cursor (either default or pointer).

Ascor8522:
> both are technically clickable, but clicking them doesn't do anything

I thought about that, but I think you shouldn't be "allowed to click" on an element that's already enabled.

mxmou:
> > both are technically clickable, but clicking them doesn't do anything
> 
> I thought about that, but I think you shouldn't be "allowed to click" on an element that's already enabled.

Could the cursor of selected popup tabs be changed to default then?

WorldLanguages:
Should the color picker be changed too?

image

WorldLanguages:
> Could the cursor of selected popup tabs be changed to default then?

Agreed

WorldLanguages:
We forgot to change the cursor of `scratch-messaging` popup buttons

image

mxmou:
I think it would make sense to move the CSS for elements that are used in many places, including buttons, into `/webpages/styles`.

WorldLanguages:
@mxmou Should I create an issue for this so we don't forget?

PR #4340, opened 2022-02-27
Settings: fix display when there are no results to show

apple502j:
- Show "No results found. Perhaps try different keywords?" when there is no search result
- Show "Explore all 130+ addons" when all addons are disabled in popup

"No results found. Perhaps try different keywords?" in settings
"No results found. Perhaps try different keywords?" in popup settings
"Explore all 130+ addons" in popup settings


RedGuy12:
@shock59 had a graphic for this page they were going to pr lol

shock59:
Tbh it might look better without a graphic

PR #4341, opened 2022-02-27
Settings page - Allows closing an addon category by re-clicking on it.

Ascor8522:
When you click the the accordion category that is already opened, the category is closed and the main / default category is then opened / selected.

### Reason for changes

I know this has been debated when I originally submitted the proposal, but hear me out.

Many accordion menus behave this way.
This commit just adds a behavior that is to be expected from such kind of menu.

It doesn't change anything for the users used to the menu's current behavior,
but it's still a nice quality of life improvement for those expecting it to behave that way.

### DIFF
--- a/webpages/settings/components/category-selector.js
+++ b/webpages/settings/components/category-selector.js
@@ -19,11 +19,11 @@ export default async function ({ template }) {
     methods: {
       onClick(event) {
         event.stopPropagation();
-        this.$root.selectedCategory = (
-          this.selectedCategory === this.category.id && !this.category.parent
-            ? this.$root.categories.find(category => category.id === "all")
-            : this.category
-        ).id;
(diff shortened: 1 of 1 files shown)


WorldLanguages:
@Ascor8522 I changed it so that two clicks under 500ms don't trigger the category reset. Maybe some users double click the categories, and it shouldn't affect your use case. What do you think?

Ascor8522:
Sounds ok, tho I would probably set the timeout to a slightly lower value. 500ms is quite slow for a double click imo.

WorldLanguages:
Okay, I'll change it to 250ms

WorldLanguages:
> to a slightly lower value

Oh, didn't read "slightly", and I chopped it in half :P

350ms?

Ascor8522:
350ms should be about right. 👍

PR #4346, opened 2022-03-01
Scratcher promotion dark mode

mxmou:
### Changes

Adds dark mode styles for the upcoming Scratcher promotion page.

### Tests

Tested by running scratch-www locally.


WorldLanguages:
/become-a-scratcher is not going to be an iframe, it will open in a new tab, right?

mxmou:
> /become-a-scratcher is not going to be an iframe, it will open in a new tab, right?

Yes, it's a full page. Links to `/users/*/#promote` will now lead to `/become-a-scratcher` instead.

PR #4376, opened 2022-03-08
"View settings" button on settings page

### LINKED ISSUE #4360: Get raw json from export file.

My friend and I are collaborating on a very messy scratch project. We use scratch addons to keep it organized. The problem is my computer (managed by my school) doesn't let extensions download files.

I would love a way to get the raw json of the config, so I can email it to my friend or something like that.

A way to save configs to the scratch addons website, so you can easily share a config with a friend.

Chiroyce1:
Resolves #4360

### Changes

Adds a button to view the raw settings file in the Settings page.

### Reason for changes

#4360

### Tests

Like this

https://user-images.githubusercontent.com/97374054/157168220-b648ad40-b6ff-4cb1-9710-55684647f64a.mov

### Drawbacks?
Well this is currently using `window.open()` so it counts as a pop-up, according to the user's browser settings, they may or may not see this message when they click the button

image

In vanilla JS, under a user event, pop-ups are allowed to be made (like when a button is clicked), but since this is using Vue, I don't think it works that way.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -290,7 +290,7 @@
   "import": {
     "message": "Import Settings"
   },
-  "view":{
+  "view": {
     "message": "View settings file"
   },
   "fileNotSelected": {
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -290,6 +290,9 @@
   "import": {
     "message": "Import Settings"
   },
+  "view":{
+    "message": "View settings file"
+  },
   "fileNotSelected": {
     "message": "Please select a settings file."
   },
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -290,7 +290,7 @@
   "import": {
     "message": "Import Settings"
   },
-  "view": {
+  "viewSettings": {
     "message": "View settings file"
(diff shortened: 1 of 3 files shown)


Chiroyce1:
Apologies since this is a bit off-topic, but I can see that the checks have failed, is it because I haven't run them on my branch?

WorldLanguages:
@Chiroyce1 Don't worry about Prettier, I'll fix it before merging

Chiroyce1:
@WorldLanguages was this supposed to be released with v1.25?

WorldLanguages:
@Chiroyce1 It was not. It will be released with v1.26.

WorldLanguages:
> In vanilla JS, under a user event, pop-ups are allowed to be made (like when a button is clicked), but since this is using Vue, I don't think it works that way.

It's not Vue. The problem is that `serializeSettings()` is async.
I think we could open a tab that says "loading" and then we could use `openedWindow.location.replace()`.

WorldLanguages:
@Chiroyce1 Looks good now?

Chiroyce1:
Yep, thanks!

PR #4381, opened 2022-03-09
Exclude 500 and 503 errors from isNotScratchWWW

### LINKED ISSUE #4364: 503 error is considered a scratchr2 page
### Description

The [maintenance](https://scratch.mit.edu/cdn/maintenance.html) and [trouble](https://scratch.mit.edu/cdn/trouble.html) pages appear when Scratch isn't available, with a 503 error code. When that happens, userstyles for scratchr2 pages run. Ideally, neither `scratchWWWNoProject` nor `isNotScratchWWW` would match those pages because they're static pages.

### Reproduction steps

Open the pages directly or open Scratch when it's down.

### Expected behavior

URLs under `/cdn/` (for testing) and the 503 error code would be excluded from `isNotScratchWWW`.

### Actual behavior

`isNotScratchWWW` matches these pages.

apple502j:
Resolves #4364

Does not contain any addon changes. Addons that do want to run on these pages need to add the page URL explicitly.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -348,12 +348,10 @@ const WELL_KNOWN_MATCHERS = {
     const { projects, projectEmbeds, scratchWWWNoProject } = WELL_KNOWN_PATTERNS;
     // Server errors are neither r2 nor www
     return !(
-      (
-        projects.test(match) ||
-        projectEmbeds.test(match) ||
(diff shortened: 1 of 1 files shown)


mxmou:
Could `/cdn/*` also be excluded for easier testing?

apple502j:
@mxmou Done.

mxmou (approved):

PR #4388, opened 2022-03-10
Move eslint ignores into .eslintrc

RedGuy12:
Resolves lots of files in root

### Changes

updates the eslint ignore patterns, move them out of root, add root: true

### Reason for changes

too many files in root
### Tests

it works

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,6 @@
 {
   "root": true,
-  "ignorePatterns": ["!**", ".git/", "node_modules/", "libraries/thirdparty/*", "*.ts"],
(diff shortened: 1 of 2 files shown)


WorldLanguages:
VS Code now has file nesting. By adding a `.vscode` folder, we can now achieve this:

image

It looks like this collapsed:
image

.eslintrc.json
```diff
@@ -1,4 +1,6 @@
 {
+  "root": true,
```
RedGuy12, line 2: this is best practice so when linting it doesn't inherit any personal configs

RedGuy12:
@WorldLanguages can i/you merge this pls
this is tiny and has no downsides

.eslintrc.json
WorldLanguages: Good idea!

PR #4391, opened 2022-03-11
Twelve hour time

### LINKED ISSUE #4021: AM/PM (12 hour) time option for forums
I... being american. Am only able to read 12 hour time or AM/PM. 24 hour time... yuck.
Toggle/Option for AM/PM time for forums
Doing unnecessary math for time format conversion.
### Additional context
24:00 Bad, 12 AM (good)

webdev03:
Resolves #4021

### Changes

Adds options to `forum-time-zones`, mainly twelve hour time, but there's also a new relative date option.

### Reason for changes

Some people don't like 24 hour time.

### Tests

Successful test through Firefox Developer Edition 99
image


addons/twelve-hour-time/addon.json
```diff
@@ -0,0 +1,35 @@
...
+    },
+    {
+      "name": "towerofnix",
+      "link": "https://github.com/towerofnix"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "changer.js",
+      "matches": ["https://scratch.mit.edu/discuss/topic/*"]
```
RedGuy12, line 23: use word matcher

addons/twelve-hour-time/addon.json
```diff
@@ -0,0 +1,35 @@
...
+    }
+  ],
+  "settings": [
+    {
+      "name": "Use relative dates",
+      "id": "relative-dates",
+      "description": "Use dates like \"Today\" instead of \"1 March\"",
```
RedGuy12, line 30: description does not exist...

addons/twelve-hour-time/addon.json
```diff
@@ -0,0 +1,35 @@
...
+  "enabledByDefault": false,
+  "credits": [
+    {
+      "name": "god286/webdev03",
+      "link": "https://github.com/webdev03"
+    },
+    {
+      "name": "JaydenDev",
+      "link": "https://github.com/JaydenDev"
+    },
```
RedGuy12, line 14: the credits section is for people who wrote code. not the person who just got the idea.

RedGuy12 (changes requested):
I'd put this in `forum-time-zones` tbh

JaydenDev:
Yes, @webdev03, it would be better to not have me in the credits for just giving the idea.

Chiroyce1:
> I'd put this in `forum-time-zones` tbh

Agreed, but I think `forum-time-zones` is a bit confusing, maybe renaming the entire addon to something like `Customizable Forum Timestamps` or something similar is better...

webdev03:
I'm converthing this to the forum time zones addon. But I have some questions:
1. There doesn't seem to be any mention of ScratchDB even though it requests from ScratchDB.
2. Would I add the credits from twelve-hour-time or just remove them entirely, because nobody added credits for that addon?
3. Should the relative date option still exist?

RedGuy12:
> 1. There doesn't seem to be any mention of ScratchDB even though it requests from ScratchDB.

there should be

> 2. Would I add the credits from twelve-hour-time or just remove them entirely, because nobody added credits for that addon?

yes

> 3. Should the relative date option still exist?

yes

webdev03:
For \#3

> > 1. There doesn't seem to be any mention of ScratchDB even though it requests from ScratchDB.
> 
> there should be
> 
> > 2. Would I add the credits from twelve-hour-time or just remove them entirely, because nobody added credits for that addon?
> 
> yes
> 
> > 3. Should the relative date option still exist?
> 
> yes

OK, I will add it after my guitar class.

webdev03:
Stuck on
Screen Shot 2022-03-12 at 11 04 40 AM
nothing in the console

webdev03:
It should be in forum-time-zones now

addons/twelve-hour-time/addon.json
webdev03: Oh sorry, I didn't realise that was a thing.
(I changed the URL name.. LOL..)

webdev03:
So the relative date option won't run unless either one is on. So should I make another JS file to handle that if both of those options are off?

GrahamSH-LLK:
Is this just waiting on reviews?

webdev03:
> Is this just waiting on reviews?

Maybe, but I think that there should be some changes to the variable names. It has been a long time since I made this PR. Also it looks like there is a conflict in the code that can't be fixed online.

addons/forum-time-zones/matcher.js
```diff
@@ -0,0 +1,66 @@
+export default async function ({ addon, global, console }) {
+  // check if setting enabled
```
webdev03, line 2: Should it be "check if twelve-hour-time is enabled"?

addons/forum-time-zones/matcher.js
```diff
@@ -0,0 +1,66 @@
...
+
+    let formattedDate = formatter.format(date);
+    if (isRelative && addon.settings.get("relative-dates") === true) {
+      const resetFormatter = new Intl.DateTimeFormat("default", {
+        hour12: true,
+        dateStyle: "long",
+      });
+      formattedDate = formattedDate.replace(resetFormatter.format(date), day);
+    }
+
```
webdev03, line 63: Should it be if/else instead?
```js
if (isRelative && addon.settings.get("relative-dates") === true) {
  const resetFormatter = new Intl.DateTimeFormat("default", {
    hour12: true,
    dateStyle: "long",
  });
  return formatter.format(date).replace(resetFormatter.format(date), day);
} else {
  return formatter.format(date);
}
```

addons/forum-time-zones/matcher.js
WorldLanguages: This comment is probably unnecessary, so don't worry about it.

addons/forum-time-zones/addon.json
```diff
@@ -1,16 +1,58 @@
...
+      "id": "relative-dates",
+      "type": "boolean",
+      "default": true
+    }
+  ],
+  "info": [
+    {
+      "type": "notice",
+      "text": "The forum timezone changer connects to ScratchDB.",
+      "id": "scratchDBNotice"
```
WorldLanguages, line 54: I don't think this is the case anymore, right?

webdev03: https://github.com/ScratchAddons/ScratchAddons/commit/386201165fd4f01bb57231f45d4fac30ea536057 changed it and I didn't notice

addons/forum-time-zones/addon.json
```diff
@@ -1,16 +1,58 @@
...
+      "id": "twelve-hour-time",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "name": "Use relative dates",
+      "id": "relative-dates",
+      "type": "boolean",
+      "default": true
+    }
```
webdev03, line 48: This only works if twelve-hour-time is enabled as well

WorldLanguages: Do relative dates also use relative time? For example, is "8 hours ago" an expected relative date? What about "yesterday 08:00 AM"?

webdev03: No, relative dates are only for the date currently. With all options on, "Yesterday 00:50:07" becomes "Yesterday at 12:50:07 am"

webdev03: This setting "relative-dates" only works when 12 hour time is on. I need to fix that soon...

WorldLanguages: Setting this conversation as resolved then.

addons/forum-time-zones/matcher.js
```diff
@@ -0,0 +1,64 @@
...
+      default:
+        date = new Date(day);
+    }
+
+    date.setHours(hours, minutes, seconds, 0);
+
+    if (isNaN(date.getTime())) {
+      console.warn(`Failed to calculate date/time for forum timestamp: ${text} - regex timeParts:`, timeParts);
+      return;
+    }
```
webdev03, line 51: This causes problems in other languages

jdev082: Its also not indented.

WorldLanguages: > This causes problems in other languages

Good catch. Maybe we should reconsider whether it's actually necessary to add an option for relative dates? The forums already use relative dates for "today" and "yesterday".

webdev03: This relative dates option doesn't add anything other than "today" and "yesterday"! The main function for that option is to turn it off which is confusing

WorldLanguages: Ohh, I see.

WorldLanguages: Here's the solution: compare `addon.tab.scratchMessage("Yesterday")` instead of `"Yesterday"`. Same with `Today`.
We need to check if some language puts the time before the date.

addons/forum-time-zones/addon.json
```diff
@@ -1,16 +1,46 @@
 {
-  "name": "Local timezone on forums",
-  "description": "Uses your local time zone for displaying post dates and times on the forums.",
+  "name": "Forum time changer",
+  "description": "Changes the times for posts in the forums.",
```
Samq64, line 3:
```suggestion
  "description": "Changes the way the date and time is displayed on the forums.",
```

addons/forum-time-zones/timezones.js
```diff
@@ -0,0 +1,144 @@
...
+      const localTimeRepr = new Date(instantTimeRepr.toLocaleString("en-US", { timeZone }));
+      // Due to daytime saving, diff between two Date is inaccurate unless we both use timezoned Date
+      // Math.min makes sure nobody gets posts from tomorrow
+      const localDateDiff = Math.min(Math.floor(localTimeRepr / 8.64e7) - Math.floor(localCurrentTimeRepr / 8.64e7), 0);
+      const timePart = localTimeRepr.toLocaleTimeString("en-GB");
+      switch (localDateDiff) {
+        case 0:
+        case -1: {
+          if (addon.settings.get("relative-dates") === true) {
+            const relativeFormatter = new Intl.RelativeTimeFormat(locale, {
```
Samq64, line 24:
```
Uncaught (in promise) TypeError: can't convert null to object
```

webdev03: For line 24?

Samq64: I got that error message on that line when going to a forum topic on Firefox.

Samq64: I'll look into it a bit later.

Samq64:
This addon has been broken ever since the mobile forums were removed a year ago unrelated to this PR.

PR #4396, opened 2022-03-12
dark-www: Add styles for various info and error messages

mxmou:
### Changes

Adds dark mode support to:
* validation errors (only if "colored areas" is set to desaturated)
* unsupported browser error
* alerts
* studio invitations
* mute message and modal
* remix credit
* the message that appears if comments are disabled globally
* welcoming committee link in "What's Happening"

Adds `/components` to the `scratchWWWNoProject` regex for easier development.

### Tests

Some screenshots:
Check out some Scratchers you might like to follow
Unsupported Browser


CST1229:
Shouldn't the unsupported browser error be part of editor-dark-mode since it's part of the editor?

mxmou:
> Shouldn't the unsupported browser error be part of editor-dark-mode since it's part of the editor? (Unless it also appears in scratch-www pages)

It appears on the project page.

Secret-chest:
But actually, nobody with the extension will be able to see the unsupported browser warning.

WorldLanguages:
> But actually, nobody with the extension will be able to see the unsupported browser warning.

if some Chromium fork removes `Chrom` from its user agent string, they might.
And other extensions can modify the user agent string too.

PR #4404, opened 2022-03-13
Fix normal stuff

apple502j:
- Editor addons run on https://llk.github.io/scratch-gui/hotfix/totally-normal-2022/ (for testing, will be reverted)
- Fix crash in addon api
- Addons that modify the Blockly internals will break when enabling "Show me cats" on Mode. data-category-tweaks-v2 crashes.

Question: should we spend this week making addons work correctly for just a couple of days, or should we just disable editor addons during the totally normal period?

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -27,6 +27,7 @@ export default class Tab extends Listenable {
     this._waitForElementSet = new WeakSet();
   }
   get clientVersion() {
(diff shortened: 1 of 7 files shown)


Norbiros:
IMAO, we should disable for these weeks (now 3 days only)...

lisa-wolfgang:
Another option is to disable all addons before using any of the totally-normal features. If someone doesn't want to use any of them, then they can continue to use our extension as usual.

PR #4427, opened 2022-03-19
Remove use of InstallTrigger for UA detection

apple502j:
An intent-to-deprecate was issued, replacing them with general User-Agent detection.


RedGuy12:
Why can't we use `typeof browser` like we usually do?

WorldLanguages:
> Why can't we use `typeof browser` like we usually do?

This code runs on scratch.mit.edu context.

PR #4436, opened 2022-03-24
Scratch Avocados (April Fools)

WorldLanguages:
Our April Fools, from March 31st 12:00:00 UTC+0 to April 2nd 12:00:00 UTC+0.

Took avocado hex color from [Wikipedia](https://en.wikipedia.org/wiki/List_of_colors:_A%E2%80%93F).

Popup changes: brand color, extension name (note that real extension name does not change, only this header). "Scratch Avocados" is not localized.
image

Settings page changes: brand color, avocado emoji on header next to "settings"
image

### DIFF
--- a/libraries/common/global-theme.js
+++ b/libraries/common/global-theme.js
@@ -1,6 +1,7 @@
 export default function () {
   const now = Date.now() / 1000;
-  if (now < 1648911600 && now > 1648738800) document.documentElement.style.setProperty("--brand-orange", "#568203");
+  if (now < 1648911600 && now > 1648738800 && ["settings", "popup"].includes(location.pathname.split("/")[2]))
+    document.documentElement.style.setProperty("--brand-orange", "#568203");
 
   const lightThemeLink = document.createElement("link");
   lightThemeLink.setAttribute("rel", "stylesheet");
--- a/libraries/common/global-theme.js
+++ b/libraries/common/global-theme.js
@@ -1,4 +1,7 @@
 export default function () {
+  const now = Date.now() / 1000;
+  if (now < 1648911600 && now > 1648738800) document.documentElement.style.setProperty("--brand-orange", "#568203");
+
   const lightThemeLink = document.createElement("link");
   lightThemeLink.setAttribute("rel", "stylesheet");
   lightThemeLink.setAttribute("href", chrome.runtime.getURL("/webpages/styles/colors-light.css"));
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -26,6 +26,8 @@ const vue = new Vue({
   },
   methods: {
     msg(message, ...params) {
+      const now = Date.now() / 1000;
+      if (message === "extensionName" && now < 1648911600 && now > 1648738800) return "Scratch Avocados";
       return chrome.i18n.getMessage(message, ...params);
     },
     direction() {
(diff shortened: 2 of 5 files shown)


RedGuy12:
Why avacado tho

Norbiros:
> Why avacado tho

I think it is something like:
Scratch Addons >> Scratch Avocados
- Almost the same amount of letter's so we don't need to fix word wrapping problems
-  the same first letter
-   W_L love sabotaging potato so he created avocado)

**Also, I want just say that a lot of user will don't see, because how often do you change settings when you are not dev/tester etc. I think we should add it somewhere to scratch page.**

So maybe we can change scratch logo to something like that:
savocado

Also, maybe we can change favicon to: 🥑
image

Also, new avocado style in scratch website and customize colors named "Avocado Theme" (that will work as easter egg for long time)?

In editor, when you select "new sprite", it automaticity creates avocado?

mxmou:
> **Also, I want just say that a lot of user will don't see, because how often do you change settings when you are not dev/tester etc. I think we should add it somewhere to scratch page.**

Anyone who uses Scratch Messages will open the popup frequently.

Chiroyce1:
> Anyone who uses Scratch Messaging will open the popup frequently.

Yeah, I use it a lot, but still I feel like this is too little, maybe an easter egg in the main website along with this would be a nice idea since more users would see it (and more frequently).

WorldLanguages:
An Easter egg in the main website could easily be confused with a joke from the Scratch Team itself. I was close to skipping April Fools this year, so this feels better than doing nothing at all.
I guess we could go one step further and hardcode the msg-count-badge color to avocado color during these 2 days, but figuring out a good way to let users opt out from avocado color on badge is not that easy, and changing the badge color against the user's will is pretty bad.

RedGuy12:
we could replace instances of "avocado" with the emoji, and a tooltip that says it's sa

Norbiros:
Maybe changes on scratchaddons.com page?
When you change logo there, no one will think that is from ST.

If it is "Scratch Avocados", change only addon, but a lot of thing inside in it. For example: new addon logo (it is possible wth overriding image for that time using js)

WorldLanguages:
> Maybe changes on scratchaddons.com page?

The extension popup probably gets more attention than our website because of Scratch Messaging.

rgantzosonscratch:
> > Why avacado tho
> 
> 
> 
> I think it is something like:
> 
> Scratch Addons >> Scratch Avocados
> 
> - Almost the same amount of letter's so we don't need to fix word wrapping problems
> 
> -  the same first letter
> 
> -   W_L love sabotaging potato so he created avocado)
> 
> 
> 
> **Also, I want just say that a lot of user will don't see, because how often do you change settings when you are not dev/tester etc. I think we should add it somewhere to scratch page.**
> 
> 
> 
> So maybe we can change scratch logo to something like that:
> 
> savocado
> 
> 
> 
> Also, maybe we can change favicon to: 🥑
> 
> image
> 
> 
> 
> Also, new avocado style in scratch website and customize colors named "Avocado Theme" (that will work as easter egg for long time)?
> 
> 
> 
> In editor, when you select "new sprite", it automaticity creates avocado?
> 
> 
> 
> 
> 
> 
> 
> 

It's not only settings, it's also messaging.

PR #4437, opened 2022-03-24
Remove dango-rain addon

WorldLanguages:
TODO: remove addon-l10n/*/dango-rain.json


Secret-chest:
Wait why? It's hidden.

WorldLanguages:
Some users still have it enabled (judging by my Scratch profile comments).
And it's been a year already since 2021 April Fools.

RedGuy12:
No. It's still a fun addon.

WorldLanguages:
It can be kept on Scratch Addons Nightly.

RedGuy12:
Many people don't use Nightly. And we don't want the Easter Eggs section to be completely empty. And there's no harm in keeping it.

WorldLanguages:
We can keep it but it should be duplicated into `dango-rain-v2` and be disabled by default.

RedGuy12:
People had to manually enable the `force` setting anyway so the people who still have it on are the people who want the addon.

WorldLanguages:
> Some users still have it enabled (judging by my Scratch profile comments).

I meant users that didn't intentionally enable it. Our migration code probably didn't run for a fraction of our users. I had comments asking why ice cream was falling out from my profile.

Chiroyce1:
> I had comments asking why ice cream was falling out from my profile.

I guess it makes sense, but maybe just disable it automatically instead of totally removing it? (I have it on 😛)

WorldLanguages:
> I guess it makes sense, but maybe just disable it automatically

That's what we did, yet some users still have it enabled because the migration did not apply to them for some reason, such as not updating Scratch Addons for a few months until we removed the migration.

CST1229:
> > I guess it makes sense, but maybe just disable it automatically
> 
> That's what we did, yet some users still have it enabled because the migration did not apply to them for some reason, such as not updating Scratch Addons for a few months until we removed the migration.

I guess then

> We can keep it but it should be duplicated into `dango-rain-v2` and be disabled by default.

WorldLanguages:
Okay, let's discuss this again. Tagging everyone that downvoted. @CST1229 @Secret-chest @Norbiros @RedGuy12 @Hans5958 @Dhuls-Scratch @Oren-Lindsey @RIMOPA

I experienced on my Scratch profile (that has the "dango" word on it) multiple users that had `dango-rain` enabled but didn't know what it was, saying stuff such as "ice cream is falling off your profile". That means these users didn't manually enable it by typing in the konami code on the settings page. Instead, this addon was enabled for them when we set it to `enabledByDefault: true` a year ago for April Fools, and for some unknown reason, our migration never disabled the addon for this fraction of users.

At this point we have no way to know whether a user intentionally enabled the addon or not. So instead of removing the addon completely like I initially considered, I suggest we duplicate it into `dango-rain-v2` and change the addon name and description if necessary. Any users that intentionally meant to enable the feature will need to type in the konami code and enable it again, but everyone who enabled it will be able to remember the dangos are added by Scratch Addons and won't be confused.

Chiroyce1:
> I suggest we duplicate it into `dango-rain-v2` and change the addon name and description if necessary. Any users that intentionally meant to enable the feature will need to type in the konami code and enable it again, but everyone who enabled it will be able to remember the dangos are added by Scratch Addons and won't be confused.

+1 and as long as dango rain isn't entirely going away - I see this as the best solution.

Hans5958:
I'd rather suggest that we should just keep the migration function for as long as we can. I don't know how it is implemented, but another temporary variable that stores if it has been disabled automatically for at least 3-5 minor versions for everyone.

WorldLanguages:
> I'd rather suggest that we should just keep the migration function for as long as we can.

It's too late to do that. We have no way of knowing if dango-rain was manually enabled or not. `addonsEnabled` does not distinguish between an addon that was enabled by default and an addon that was enabled manually.

Hans5958:
I know. I was telling you that if this is a real concern, you should just turn it off of everyone, ever. In fact, changing the addon ID would produce the same effect. 

I know you can't differentiate it, which is why you should add a secret variable that checks if it has disabled it automatically.

Secret-chest:
> PS. Please don't change the name into some bullcrap long shit such as this thing called \"Switch variables between 'For all sprites' and 'For this sprite only'\" I thought we have discussed this on a issue that I forgot the number is. I absolutely despise these kind of names, in addition of the sentence case.

I agree except for the sentence case. The addons use sentence case to describe functions that they add, and that is correct.

Norbiros:
When we are creating `v-2`, maybe we can add new features? Like potato rain?
**Just suggestion, i know that is stupid idea, but POTATO IS WAY OF LIFE!**

RedGuy12:
If the transition code didn't run, wouldn't the `force` setting still be on addonSettings?

WorldLanguages:
> you should just turn it off of everyone, ever. In fact, changing the addon ID would produce the same effect.

Yes, but changing the ID is the least complex way to disable it for everyone.

WorldLanguages:
> If the transition code didn't run, wouldn't the `force` setting still be on addonSettings?

You may be right - I'd have to double check. But that would be a solution.

TheColaber:
@WorldLanguages is this still a problem or can we close this?

WorldLanguages:
> If the transition code didn't run, wouldn't the `force` setting still be on addonSettings?

Recent changes now delete unused settings AFAIK, so this isn't possible anymore

WorldLanguages:
Closing this PR. An issue should be opened to discuss whether it's worth it to disable dango-rain for everyone, and clone it into dango-rain-v2. I have been getting less Scratch comments about "ice cream" lately.

PR #4455, opened 2022-04-01
Fixed z-index property

Norbiros:
Resolves: Problem with search bar showing in front of new easter egg layer with for example, torch:
Screenshot from 2022-04-01 20-41-36

### Changes

Just simple added `z-index`

### Tests

Yep. It looks nice, and everything works normally:
Screenshot from 2022-04-01 20-42-06

### Info

Just today we have this torch, but when SA or ST will add new things in front layer, we will need to fix it. So why just not fix it now?

### DIFF
--- a/addons/editor-devtools/userscript.css
+++ b/addons/editor-devtools/userscript.css
@@ -2,6 +2,7 @@
   display: flex;
   white-space: nowrap;
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  z-index: 1;
 }
 
 .s3devLabel {


mxmou:
Not a bug. Context menus work the same way:
image

Norbiros:
Yes, but Scratch's things have a lot of problems with z-index, for example inputs. ST should fix their problems (for example with context menu). We can fix our own elements like dev-tools addon, so why don't do that?

If you will find any SA element, that has the same problem - okay.

mxmou:
> But i think context menu even should be on top layer, because you need to click elements in it and torch would cover it. In find bar you just need to type, and youse keyboard...

The results are clickable.

WorldLanguages:
We won't be fixing issues with the Scratch April Fools modes because they'll be gone in a few hours.

Norbiros:
> We won't be fixing issues with the Scratch April Fools modes because they'll be gone in a few hours.
But maybe we will create modes addon: https://github.com/ScratchAddons/ScratchAddons/discussions/4456

WorldLanguages:
@Norbiros Then this bug should be taken into consideration after that addon is created.

Norbiros:
Okay 😄

PR #4461, opened 2022-04-03
Revert "Scratch Avocados (April Fools)"

apple502j:
Reverts #4436


WorldLanguages:
#4404 should be reverted too

WorldLanguages:
#4404 should be reverted too

RIMOPA:
> #4404 should be reverted too

I think It would be useful for the easter egg addon that returns the modes (if it is added)

mxmou:
> > #4404 should be reverted too
> 
> I think It would be useful for the easter egg addon that returns the modes (if it is added)

An addon would be very different from Scratch's implementation.

PR #4475, opened 2022-04-07
New Addon - Hide backpack

### LINKED ISSUE #4199: Hiding the backpack

The backpack area, even if small, gets some space out of the UI. I don't know how many users use it. I never use it. 

Hiding the backpack area

None

### Additional context

None

Norbiros:
Resolves #4199

### Changes

Hides backpack. Works as the same addon in Turbowarp.

### Tests

Everything works fine!

PS: I don't know if I need to move it to "theme" section in `addons.json`... But everything works fine now, so IDK.


WorldLanguages:
Other addons that hide stuff aren't "themes", so I will remove the theme tag from this.
As discussed in #4199, this addon will have the dangerous tag.

WorldLanguages:
The addon might be enabled by the user while the backpack is open, resulting in this bug. Not sure if this can be fixed with CSS alone, we might need a userscript.

image

Norbiros:
I think it will work. 
```js
//run this code on enable
let event = new CustomEvent("resize");
window.dispatchEvent(event);
```

lisa-wolfgang:
Since it's easy to forget that such an addon is enabled (remember Ctrl+Click to run scripts?) I think it would be better to provide a more compact way to open it. Perhaps an auto-hiding button added in this general area:

Hide Backpack Concept

Norbiros:
> Since it's easy to forget that such an addon is enabled (remember Ctrl+Click to run scripts?) I think it would be better to provide a more compact way to open it instead of completely removing it. Perhaps an auto-hiding button added in this general area:
> 
> Hide Backpack Concept

Okay, but there is problem with space where you want to put this button.... Backpack is not relative to workspace, but to workspace + block palette, so we don't know how long block palette is... we can do this by % (bugs), maybe with js (huge amount of work) [also what with auto hiding block palete]... So i think that we can put this button on right side of screen (but we need find good place for it, because it can overlap with buttons to change size)

mxmou:
> Perhaps an auto-hiding button added in this general area:

There would need to be a way to open the backpack from the Costumes and Sounds tabs as well. It would be weird if the button was in different places on different tabs.

> we don't know how long block palette is

It's always 310px.

WorldLanguages:
Removed this from the v1.26..0 milestone - as discussed above, we'll wait until we find a way to let users show the backpack without having to interact with the settings page. A setting that's disabled by default could restore the current behavior, where there's no way to show the backpack without interacting with Scratch Addons. This also means we could hide the "dangerous" tag if the button to show the backpack is intuitive enough.

RIMOPA:
> Removed this from the v1.26..0 milestone - as discussed above, we'll wait until we find a way to let users show the backpack without having to interact with the settings page. A setting that's disabled by default could restore the current behavior, where there's no way to show the backpack without interacting with Scratch Addons. This also means we could hide the "dangerous" tag if the button to show the backpack is intuitive enough.

I think We can make a tooltip that shows up when you mouse is over the backpack place, it would say something like 'The backpack has been hidden because you have the "Hive backpack" addon enabled. Click here twice to show the backpack again!'

Norbiros:
Okay, so this hide backpack icon should be in place that you can access from all tabs, maybe here?
awful-image-lol
*Top left corner*

mxmou:
Putting anything in the top left corner would require moving the tabs. The top right corner might be better but it's very far from the backpack so it should probably be replaced by a popup.

CST1229:
Maybe a small button at the bottom right corner of the left panel (which [the left panel] contains the tab contents)?

mxmou:
> Maybe a small button at the bottom right corner of the left panel (which [the left panel] contains the tab contents)?

It could cover other things such as the zoom buttons in code and paint editors.

Norbiros:
> Putting anything in the top left corner would require moving the tabs. The top right corner might be better but it's very far from the backpack so it should probably be replaced by a popup.

Okay, I think it is best idea. What with icon? This backpack icon is good?

lisa-wolfgang:
> Since it's easy to forget that such an addon is enabled (remember Ctrl+Click to run scripts?) I think it would be better to provide a more compact way to open it instead of completely removing it. Perhaps an auto-hiding button added in this general area:
> 
> 
> 
> Hide Backpack Concept
> 
> 

We could put it here for the code tab and put it in a similar location on the costume and sound tabs.

lisa-wolfgang:
For the costume and sound tabs, it might make more sense to put it next to the "add costume/sound" button since that's where users go when they want to add any other sort of costume. We could also have an "add to backpack" context menu item for convenience, although that's probably outside the scope of this addon.

Similarly, for the code editor, the button could go on the inside-right of the block palette. I think a consistent solution for all three tabs is unfeasible because of how different the UI between the code and costume/sound tabs is.

Norbiros:
Something like that?
Costume and sound editor:
Screenshot from 2022-07-22 16-22-28
Code editor:
Screenshot from 2022-07-22 16-27-05

Samq64:
Doesn't adding a button to show it defeat the purpose of the addon? (Not that I would enable this.)

mxmou:
> Doesn't adding a button to show it defeat the purpose of the addon? (Not that I would enable this.)

I believe the purpose of the addon is to increase the amount of usable space.

Secret-chest:
I don't think this makes sense. It's around 16 poxels of vertical space.

Norbiros:
I worked on that, and now it looks like this:
Screenshot from 2022-07-25 20-25-23
Screenshot from 2022-07-25 20-25-13

To Do:
- tests, tests, tests
- Maybe change icon? Make it look more like button? (@mxmou can you give opinion)

lisa-wolfgang:
Didn't we decide that the button should go here? https://github.com/ScratchAddons/ScratchAddons/pull/4475#issuecomment-1192629133

Norbiros:
> Didn't we decide that the button should go here? [#4475 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/4475#issuecomment-1192629133)

It was hard to implement, and it changed location (between sprite and costume editor), but now it is in the same location...

Norbiros:
I think that everything is fixed. Also, I changed icon:
Screenshot from 2022-09-08 17-02-51

addons/hide-backpack/userscript.css
```diff
@@ -0,0 +1,16 @@
+[class^="backpack_backpack-container"] {
+  display: none;
+}
+
+[class^="button-group_button-group_"] {
```
mxmou, line 5: This rule should be in a separate CSS file that's only injected if "show small backpack icon" is enabled.

```suggestion
[class^="paint-editor_zoom-controls_"] {
```
`button-group_button-group_` will select every button group in the paint editor, including undo/redo:
image

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,93 @@
+export default async function ({ addon, global }) {
+  document.querySelector("[class^=backpack_backpack-header_]").style.display = "none";
```
mxmou, line 2: This crashes if the backpack doesn't exist yet. Use `waitForElement`.

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,93 @@
...
+  if (addon.settings.get("showButton")) createBackpackButton(addon);
+
+  // Fix automatic enable bug
+  window.dispatchEvent(new Event("resize"));
```
mxmou, line 6: This needs to be called every time the addon is enabled or disabled.

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,93 @@
...
+
+  // Event listeners that add dynamic enable/disable + setting change
+  addon.settings.addEventListener("change", (e) => changeBackpackVisibility(addon));
+
+  addon.self.addEventListener("reenabled", (e) => changeBackpackVisibility(addon));
+  addon.self.addEventListener("disabled", (e) => {
+    moveResizeButtons(40);
```
mxmou, line 13:
```suggestion
    moveResizeButtons(0);
    window.dispatchEvent(new Event("resize"));
```

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,93 @@
+export default async function ({ addon, global }) {
+  document.querySelector("[class^=backpack_backpack-header_]").style.display = "none";
+  if (addon.settings.get("showButton")) createBackpackButton(addon);
+
+  // Fix automatic enable bug
```
mxmou, line 5:
```suggestion
  // Fix dynamic enable bug
```

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,93 @@
...
+  backpackButton.style.backgroundImage = `url('${addon.self.dir}/backpack.png')`;
+  backpackButton.classList.add("sa-backpack-button");
+  backpackButton.addEventListener("click", toggleBackpack);
+  document.querySelector(".injectionDiv").appendChild(backpackButton);
+
+  moveResizeButtons(35);
+
+  // Add backpack button to costume and sound editor
+  addSoundEditorButton(addon, backpackButton);
+  addCostumeEditorButton(addon, backpackButton);
```
mxmou, line 49: Just make one button and append it to `[class*="gui_tabs_"]`.

mxmou (changes requested):

mxmou:
The design of the icon could still be improved. I can do that myself once the more important problems are fixed.

Norbiros:
Thank you very much @mxmou! I tested that and everything works great.

*Improved backpack button, screenshots:*
Screenshot from 2022-09-12 18-11-18
Screenshot from 2022-09-12 18-11-34

WorldLanguages:
Just tested this, it appears to work well.

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,80 @@
+export default async function ({ addon }) {
+  let originalBackpack = await addon.tab.waitForElement("[class^=backpack_backpack-header_]", {
+    markAsSeen: true,
+  });
+
+  changeBackpackVisibility(addon);
```
WorldLanguages, line 6: This code is already here below. No need to have it here too.

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,80 @@
...
+    moveResizeButtons(0);
+    originalBackpack.style.display = "block";
+    window.dispatchEvent(new Event("resize"));
+    document.querySelector(".sa-backpack-button").style.display = "none";
+  });
+
+  while (true) {
+    originalBackpack = await addon.tab.waitForElement("[class^=backpack_backpack-header_]", {
+      markAsSeen: true,
+    });
```
WorldLanguages, line 21: Possible optimization (needs to be tested)
`reduxCondition: (state) => state.scratchGui.editorTab.activeTabIndex === 0 && !state.scratchGui.mode.isPlayerOnly,`

Norbiros: It doesn't work because of `state.scratchGui.editorTab.activeTabIndex === 0` (normally you open project in first page, but if you click see inside it doesn't work), but we can just add `reduxCondition: (state) => !state.scratchGui.mode.isPlayerOnly,`

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,80 @@
...
+  window.dispatchEvent(new Event("resize"));
+}
+
+// Move resize buttons to top
+function moveResizeButtons(distance) {
+  const resizeElements = document.querySelectorAll(".blocklyZoom > image");
+  resizeElements[0].setAttribute("y", (44 - distance).toString());
+  resizeElements[1].setAttribute("y", (0 - distance).toString());
+  resizeElements[2].setAttribute("y", (88 - distance).toString());
+}
```
WorldLanguages, line 80: Scratch has a bug where it sometimes puts these buttons lower than they should be.
Is this considered here?
I'm trying to find more information about that bug

mxmou: The bug happens because the workspace dimensions are incorrect. It's unrelated to these values.

It's probably possible to calculate them from constants defined in `scratch-blocks/core/zoom_controls.js` instead of hardcoding them though.

WorldLanguages:
This PR lacks an assignee. @mxmou Would you like to be assigned?

mxmou:
Bug: the button is in the "closed" state if the addon is enabled while the backpack is open.

mxmou:
Going to the project page and back while the backpack is open closes the backpack but the button state doesn't change.

WorldLanguages:
@mxmou Do you think the PR is mergeable? Should I re-review?

WorldLanguages:
It might be a good idea to release this backpack addon at the same time as the other backpack addon, #5120

mxmou:
> @mxmou Do you think the PR is mergeable? Should I re-review?

Ideally the bugs I mentioned above would be fixed before merging.

WorldLanguages:
Oh, I quickly scanned the page and thought the commits below were bug fixes

mxmou (approved):
The bugs I found are now fixed. I think the description and setting name could be improved, but I don't have specific suggestions.

Norbiros:
Thanks for working on this PR ❤️

Samq64:
The compact editor causes the button to overflow slightly into the paint canvas:
image

addons/hide-backpack/button.css
```diff
@@ -0,0 +1,46 @@
...
+.sa-backpack-button {
+  position: absolute;
+  bottom: 13px;
+  width: 36px;
+  height: 36px;
+  display: flex !important;
+  justify-content: center;
+  align-items: center;
+  padding-bottom: 2px; /* vertical position of the icon */
+  background-color: #575e75;
```
Samq64, line 23: I think the button and icon should be the same colour as the others. The green can stay though.

Samq64 (commented):
The icon on

DNin01 (commented):
I think people who just want a different backpack button or one that takes up less space might miss this addon if we call it "Hide backpack button". We might want to rename this addon to something like "Customizable backpack button" to more accurately reflect its functionality and purpose.

Another idea that comes to mind is using a select setting. Both of the current settings, on and off, change completely different things - setting it to on moves the backpack button to a different location and makes it smaller, while setting it to off removes access to the backpack feature entirely. The switch should be replaced with a selection menu with two choices, one to hide the button and one to make it smaller, for clarity. However, if you think of it like the main backpack button is always removed with this addon on, turning the "show small backpack button" setting on adds onto the addon, so maybe it's fine.

Or, alternatively, we could just remove the option to hide the backpack button and only let users make it smaller. That would solve almost all the issues.

But is this addon worth it when there are these UX issues and most people don't care about the current position anyway? I don't think it takes up much space.

addons/hide-backpack/button.css
mxmou: I chose a different color because the button behaves differently - it's a toggle button.

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,86 @@
...
+    window.dispatchEvent(new Event("resize"));
+    document.querySelector(".sa-backpack-button").style.display = "none";
+  });
+
+  while (true) {
+    originalBackpack = await getBackpackElement();
+    if (!addon.self.disabled) changeBackpackVisibility();
+  }
+
+  async function getBackpackElement() {
```
Samq64, line 20: Does this need to be its own function? I don't think the call at the top is necessary.

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,86 @@
...
+    window.dispatchEvent(new Event("resize"));
+  }
+}
+
+function isBackpackOpen() {
+  return !!document.querySelector("[class^=backpack_backpack-list_]");
+}
+
+// Create default backpack button
+function createBackpackButton(addon) {
```
Samq64, line 50: Is there a reason `addon` is being passed as a parameter?

Samq64 (commented):
I also agree that the addon being called "hide backpack" doesn't make much sense any more.

addons/hide-backpack/userscript.js
mxmou: This function is defined outside the main function. It needs `addon` because it calls `addon.tab.scratchMessage()`.

WorldLanguages:
I think the addon description should mention the default behavior of the addon. The current description is quite confusing.

We could also change the addon name so that it can be better understood that the backpack is still accessible, but through a smaller button. Any ideas?

WorldLanguages:
Some proposals:

- Addon name: `Backpack button` -- or -- `Access backpack through button` -- or -- `Backpack button instead of bar`
- Addon description: `Replaces the backpack tray with a backpack button, below the zoom controls which lets you toggle whether the backpack is open. Optionally, hide the backpack completely.` -- (if you can think of a better word other than "tray", go ahead - what's the best way to name the horizontal bar?)
- Setting name: `Hide backpack completely` (defaults to off)
- Info warning: `If you enable the "hide backpack completely" setting, the backpack will be inaccessible.`

WorldLanguages:
Whoops, I just read Dnin's comment here: https://github.com/ScratchAddons/ScratchAddons/pull/4475#pullrequestreview-1846493086

addons/hide-backpack/addon.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "Hide backpack",
```
Samq64, line 2:
```suggestion
  "name": "Replace backpack bar with button",
```
Can someone come up with something better?

addons/hide-backpack/addon.json
```diff
@@ -0,0 +1,54 @@
...
+      "name": "Maximouse",
+      "link": "https://scratch.mit.edu/users/Maximouse/"
+    }
+  ],
+  "settings": [
+    {
+      "name": "Show small backpack icon",
+      "id": "showButton",
+      "type": "boolean",
+      "default": true
```
Samq64, line 26:
```suggestion
      "default": false
```

addons/hide-backpack/addon.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "Hide backpack",
+  "description": "Removes the backpack from the bottom of the editor or adds a small button to open it.",
```
Samq64, line 3:
```suggestion
  "description": "Replaces the backpack bar along the bottom of the editor with a button next to the zoom controls or removes it completely",
```

mxmou:
> Setting name: `Hide backpack completely` (defaults to off)

Should this even be an option? I don't think it's very useful.

WorldLanguages:
> > Setting name: `Hide backpack completely` (defaults to off)
> 
> Should this even be an option? I don't think it's very useful.

I guess it's useful if you **never** use the backpack...? I don't remember anyone asking to hide the backpack completely

mxmou:
> I guess it's useful if you **never** use the backpack...? I don't remember anyone asking to hide the backpack completely
> 
> EDIT: nevermind, the issue this intends to resolve (#4199) asked for exactly that.

Even if someone never uses the backpack, they will enable this addon to save space, not to make the backpack inaccessible.

WorldLanguages:
I think it is a good idea to remove the setting, then. We can always add it back in the future.

CST1229:
> I guess it's useful if you **never** use the backpack...? I don't remember anyone asking to hide the backpack completely

For reference, TurboWarp has an addon that does this:
image

addons/hide-backpack/userscript.js
```diff
@@ -0,0 +1,75 @@
...
+    originalBackpack = await addon.tab.waitForElement("[class^=backpack_backpack-header_]", {
+      markAsSeen: true,
+      reduxCondition: (state) => !state.scratchGui.mode.isPlayerOnly,
+    });
+    if (!addon.self.disabled) changeBackpackVisibility();
+  }
+
+  function changeBackpackVisibility() {
+    originalBackpack.style.display = "none";
+    let backpackEl = document.querySelector(".sa-backpack-button");
```
WorldLanguages, line 22:
```suggestion
    const backpackEl = document.querySelector(".sa-backpack-button");
```

WorldLanguages (approved):
Tested in Chromium :)

WorldLanguages:
> For reference, TurboWarp has an addon that does this:

It makes sense that TurboWarp users would want to hide the backpack, as they cannot log in.

Norbiros:
Wow. Thanks for finishing this PR!

Secret-chest:
2022...

PR #4478, opened 2022-04-08
Use declarativeNetRequestWithHostAccess when possible

apple502j:
Progress on #3615

### Changes
Uses declarativeNetRequestWithHostAccess when possible. Users running Chrome 95 and below, plus Firefox users, will still use webRequestBlocking. Packer script now automatically removes DNR permission from Firefox build.

Note that `domains` condition was renamed to `initiatorDomains`, but this is not shipped yet.

Tested that requests using sareferer still works in Chrome and Firefox.

Next step would be to drop all Chrome versions without DNR support and implement a way to ask for non-WHA DNR permission in browsers that require those. Not sure how to test that though.

### DIFF
--- a/background/handle-fetch.js
+++ b/background/handle-fetch.js
@@ -1,59 +1,62 @@
-const extraInfoSpec = ["blocking", "requestHeaders"];
-if (Object.prototype.hasOwnProperty.call(chrome.webRequest.OnBeforeSendHeadersOptions, "EXTRA_HEADERS"))
-  extraInfoSpec.push("extraHeaders");
+if (chrome.declarativeNetRequest) {
+  // DNR available, use this instead
+  chrome.declarativeNetRequest.updateDynamicRules({
+    removeRuleIds: [1],
+    addRules: [
+      {
+        id: 1,
+        priority: 1,
+        action: {
+          type: "modifyHeaders",
+          requestHeaders: [
+            {
+              header: "Referer",
+              operation: "set",
+              value: "https://scratch.mit.edu/",
+            },
+          ],
+        },
+        condition: {
+          domains: [chrome.runtime.id],
+          regexFilter: "^https:\\/\\/(api\\.|clouddata\\.|)scratch\\.mit\\.edu\\/.*(\\?|\\&)sareferer",
+          resourceTypes: ["xmlhttprequest"],
+        },
+      },
+    ],
+  });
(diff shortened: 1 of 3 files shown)


Hans5958:
idk why i asked but #givewebsiteaccess idk #558

apple502j:
@Hans5958 Required? nope. Optional? maybe.
You can PR yourself

Hans5958:
awww

RedGuy12:
So we're okay with users having to re-opt-in to the extension?

apple502j:
@RedGuy12 It shouldn't cause that I believe

Hans5958:
Moving to MV3, even with same permissions (but with different keys), would result into a re-request. That's something we proved on #558.

Or atleast at that time. I expect Google does nothing about it tho.

RedGuy12:
> Moving to MV3, even with same permissions (but with different keys), would result into a re-request. 

Well, this PR doesn't move to MV3 so we still should avoid rer-equests.

background/handle-fetch.js
```diff
@@ -1,59 +1,62 @@
-const extraInfoSpec = ["blocking", "requestHeaders"];
-if (Object.prototype.hasOwnProperty.call(chrome.webRequest.OnBeforeSendHeadersOptions, "EXTRA_HEADERS"))
-  extraInfoSpec.push("extraHeaders");
+if (chrome.declarativeNetRequest) {
```
Weredime, line 1:
```suggestion
if (typeof chrome.declarativeNetRequest !== "undefined") {
```

apple502j: ?

Norbiros: Why do you want to check if this function returns string with value "undefined", not for undefined type (!== checks for type too)?

mxmou: > Why do you want to check if this function returns string with value "undefined", not for undefined type (!== checks for type too)?

`typeof` returns a string.

RedGuy12 (dismissed):

Weredime (approved):

WorldLanguages (approved):
Tested in Chrome latest stable
Tested in Chrome 80

WorldLanguages:
We need to update https://scratchaddons.com/docs/policies/chrome-web-store-privacy-information/

Issue #3222, opened 2021-08-03
A better way to see unshared projects


Hans5958:
Alright, this one might be controversial and is going to make someone angry (and I bet Apple will close this because Apple being Apple), but hear me out.

Here are the reasons.

1. It is still not fixed, even after numerous of calls to the Scratch Team.  
I couldn't find one, but I think it's safe to assume so.
2. Other players (such as TurboWarp) can run it just with the ID.
3. It is probably a open secret for now.  
There are some people are sharing their unshared project using TurboWarp. If they didn't know, I think it is the best interest to make them aware of it, to fight the false sense of privacy.
4. It is already proposed multiple times.  
At least twice, long time ago, at #70 and #91.

My idea is to have a button on not found project page. No need to check if it is exist (if you do, sure). Just a button that said "you can try use...." You can, say, implement a player to run it, but for now, let's just keep it simple.

WorldLanguages:
*Probably* fine. Users can already use TurboWarp, or even the default-project addon.

WorldLanguages:
Apparently when I posted the previous comment, I forgot about a conversation I had with griffpatch on September 2020, before Scratch Addons was actually released to the masses. All messages by griffpatch, comments between square brackets by me:

> Not sure we should have an extension to view deleted comments except perhaps on your own profile.

> Still not sure about the deleted comment thing...

> Lol - sadly [view unshared projects] is probably a good idea to keep off too

> And yes - that update cloud variables or pretend to be a different user one is right out there

WorldLanguages:
Unshared projects will be actually private eventually

https://docs.turbowarp.org/unshared-projects

Issue #510, opened 2020-10-16
New addon: Make project shared


Explosion-Scratch:
**Describe the new addon**

As the title says. Unshared projects become shared. Magic.
**Link to the new addon**

https://github.com/forkphorus/cat-plus/blob/master/make-projects-shared.user.js

mxmou:
Considering that the deleted comments addon was removed, I don't think this will be added.

WorldLanguages:
Yeah. In the same way I agree deleted comments should actually be deleted, unshared projects should stay unshared. It's sad that this is possible.

Hans5958:
I wonder why is this closed. Considering that people can just use Forkphorus/Leopard/TurboWarp to get the same effect, I don't see why this exact effect should be excluded here.

WorldLanguages:
@Hans5958 For the same reason we removed the "see deleted comments" addon.

mxmou:
@WorldLanguages Could the deleted comments addon be re-added if it only affected your own profile?

Hans5958:
> 
> 
> @Hans5958 For the same reason we removed the "see deleted comments" addon.

Yeah, but that is compared to this suggestion, which method is more widely known, compared to the "see deleted comments" one.

Explosion-Scratch:
True...

JeffaloBob:
we need to make a fork where its not on the webstore and we just go out and make all of the crazy things

YOYITsM3M8:
> we need to make a fork where its not on the webstore and we just go out and make all of the crazy things

+ 100m support

Explosion-Scratch:
I sorta suggested that https://github.com/ScratchAddons/ScratchAddons/discussions/555

apple502j:
Please no.

PR #4479, opened 2022-04-09
TurboWarp redirect

illogicalapple:
### Changes
Added turbowarp redirect addon

### Reason for changes
Redirects unshared projects to Turbowarp

### Tests
Everything works!
I have tested it by installing a copy of Scratch Addons

### DIFF
--- a/addons/turbowarp-redirect/addon.json
+++ b/addons/turbowarp-redirect/addon.json
@@ -34,5 +34,5 @@
   ],
   "versionAdded": "1.26.0",
   "dynamicEnable": true,
-  "enabledByDefault": true
+  "enabledByDefault": false
 }
--- a/addons/turbowarp-redirect/addon.json
+++ b/addons/turbowarp-redirect/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Turbowarp redirect",
+  "name": "TurboWarp redirect",
   "description": "Redirects unshared projects to Turbowarp.",
   "tags": ["community"],
   "credits": [
--- a/addons/turbowarp-redirect/addon.json
+++ b/addons/turbowarp-redirect/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "Turbowarp Redirect",
-  "description": "Redirects unshared projects to turbowarp.",
+  "name": "Turbowarp redirect",
+  "description": "Redirects unshared projects to Turbowarp.",
   "tags": ["community"],
   "credits": [
     {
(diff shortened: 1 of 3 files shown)


addons/turbowarp-redirect/addon.json
```diff
@@ -0,0 +1,38 @@
...
+        {
+          "id": "//forkphorus.github.io/#",
+          "name": "Forkphorus"
+        }
+      ]
+    }
+  ],
+  "versionAdded": "1.26.0",
+  "dynamicEnable": true,
+  "enabledByDefault": true
```
Norbiros, line 37: I think it should be `false`.

addons/turbowarp-redirect/addon.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "Turbowarp redirect",
```
CST1229, line 2:
```suggestion
  "name": "TurboWarp redirect",
```
TurboWarp is a name.

Norbiros (changes requested):

addons/turbowarp-redirect/addon.json
```diff
@@ -0,0 +1,38 @@
...
+  "settings": [
+  	{
+      "name": "Redirect to",
+      "id": "redirectUrl",
+      "type": "select",
+      "default": "//turbowarp.org/",
+      "potentialValues": [
+        {
+          "id": "//turbowarp.org/",
+          "name": "Turbowarp"
```
CST1229, line 26:
```suggestion
  "description": "Redirects unshared projects to TurboWarp.",
  "tags": ["community"],
  "credits": [
    {
      "name": "illogicalapple",
      "link": "https://scratch.mit.edu/users/illogicalapple/"
    }
  ],
  "userscripts": [
    {
      "url": "userscript.js",
      "matches": ["https://scratch.mit.edu/projects/*"]
    }
  ],
  "settings": [
  	{
      "name": "Redirect to",
      "id": "redirectUrl",
      "type": "select",
      "default": "//turbowarp.org/",
      "potentialValues": [
        {
          "id": "//turbowarp.org/",
          "name": "TurboWarp"
```
(also in these places)

pufferfish101007:
Pretty sure this has been rejected for ethical reasons.

Norbiros:
> Pretty sure this has been rejected for ethical reasons.

I agree!

Screenshot from 2022-04-09 17-36-29

I think it is in some kind "exploiting bug", scratch wanted to share projects, so you can hide them. They didn't block api, but when we are creating simple way to see unshared projects, it is not nice :-/. But as always, the last word belongs to W_L.

mxmou:
> Pretty sure this has been rejected for ethical reasons.

#510 is very similar to this PR and was rejected. #3222 suggests adding a link to an alternative player instead and is apparently not rejected.

addons/turbowarp-redirect/addon.json
illogicalapple: Yes, I forgot about that, thanks! :)

illogicalapple:
> Pretty sure this has been rejected for ethical reasons.

https://docs.turbowarp.org/unshared-projects
Besides, there's [probably] no way you would be able to get a random project ID of a random unshared project without someone giving it to you.

Chiroyce1:
Maybe some text on screen that says `Redirecting you to TurboWarp...` while it redirects? It will only be shown for <= 1 second(s) but still...

Norbiros:
Screenshot from 2022-04-09 17-33-29
I think that we can display scratch tool bar to, if you want to go back etc...

WorldLanguages:
"See deleted comments" and an addon like this were rejected by griffpatch back in 2020.

Norbiros:
@WorldLanguages so please, close also #3222. If it is rejected  other issues like 3222 should be rejected too

WorldLanguages:
@Norbiros Thanks!

See #3222, I just posted more details about what addons griffpatch rejected in 2020.

PR #4493, opened 2022-04-13
New Addon: `remaining-replies`

BarelySmooth:
### Changes
New addon that adds the number of remaining replies in a studio comment thread.
[Image demonstrating the addons function](https://scratch.mit.edu/studios/30131729/comments/#comments-181718673)
(ignore the typo in the screenshot)

~~The placement of the text should probably be reconsidered (it's a little confusing), but that was the best place I could think of.~~

### Tests
Tested on Edge (because firefox makes it really hard to test local extensions)


mxmou:
> The placement of the text should probably be reconsidered (it's a little confusing), but that was the best place I could think of.

Could it be on the reply button of each comment in the thread? For example "reply (5 left)".

WorldLanguages:
@BarelySmooth Note that mxmou liked your message (GitHub doesn't notify reactions)
But yeah it's repetitive, maybe it could only be shown after the thread has 10 replies? @mxmou What do you think?

mxmou:
It could also be a tooltip on the reply button.

BarelySmooth:
> But yeah it's repetitive, maybe it could only be shown after the thread has 10 replies?

Could this be an option within settings?

Norbiros:
Please add developer comments to it, because translators can translate it as "left site" (in polish "lewo"), or as " 5 left' (in polish " zostało 5")...

BarelySmooth:
> Please add developer comments to it, because translators can translate it as "left site" (in polish "lewo"), or as " 5 left' (in polish " zostało 5")...

How can I do that? (Adding it to the userscript probably wont work because the translation is done on Transifex)


https://github.com/ScratchAddons/ScratchAddons/blob/f1fddc7409085b7d63fd2990028fbacef7092b84/addons-l10n/en/remaining-replies.json#L3
I used the word "remaining" instead of "left" in the json to prevent confusion

Norbiros:
Yesterday @apple502j added this feature... Maybe he know?

Okay. Thats better!

BarelySmooth:
> But yeah it's repetitive, maybe it could only be shown after the thread has 10 replies?

@WorldLanguages I've added an option to the addon settings, is that fine?

screenshot
screenshot

DNin01 (approved):
Tried it, works well! The counter even updates after you reply 😃

Perhaps the addon could also gray out/dim the reply button when the comment thread has reached its limit.

Endermaster3690:
> screenshot

is there a way to make it custom?

BarelySmooth:
> Tried it, works well! The counter even updates after you reply 😃
>
> Perhaps the addon could also gray out/dim the reply button when the comment thread has reached its limit.

Thanks! 
Also, Greying out the reply button is possible, but that isn't the purpose of the addon (imo).

lisa-wolfgang:
In my opinion, a better, less repetitive location for this is next to the Post/Cancel button area.

image

BroJac5246:
> Another option would be to keep the number in the reply button, but only show the remaining replies text while it's being hovered.

I think that would be best.


> > Also, Greying out the reply button is possible, but that isn't the purpose of the addon (imo).
> 
> It seems like a good related feature to bundle in the same addon.

I agree.

addons-l10n/en/remaining-replies.json
```diff
@@ -0,0 +1,4 @@
+{
+  "remaining-replies/reply": "reply",
+  "remaining-replies/remaining": "left"
```
Samq64, line 3: Merging the two messages into
```suggestion
  "remaining-replies/reply": "reply ({replies} left)"
```
should make things simpler in the JS file and possibly for translators.

Samq64 (commented):
Could you add dynamic enable/disable support?

addons/remaining-replies/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Remaining replies for studio comments",
```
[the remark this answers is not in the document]
DNin01: This sounds more like replies to a studio, not replies to comments in studios. Maybe "remaining reply counter" and specify in the description that this is for studios? We don't need the name to 100% describe the addon.

addons/remaining-replies/addon.json
```diff
@@ -0,0 +1,27 @@
...
+  "description": "Show number of remaining replies for studio comment threads.",
```
DNin01, line 3: I have a few description ideas:
- Shows you how many more times a comment thread in a studio can receive a reply.
- Shows you how many more replies can be posted on each comment thread in a studio.
- Shows you each studio comment thread's number of remaining replies.
- Shows you the number of remaining replies for each comment thread in a studio.
- Shows you how many more times you can reply to each comment thread in a studio.

Which one do you think is the best?

BroJac5246: > I have a few description ideas:
> 
> * Shows you how many more times a comment thread in a studio can receive a reply.
> * Shows you how many more replies can be posted on each comment thread in a studio.
> * Shows you each studio comment thread's number of remaining replies.
> * Shows you the number of remaining replies for each comment thread in a studio.
> * Shows you how many more times you can reply to each comment thread in a studio.
> 
> Which one do you think is the best?

I prefer "Shows you how many more replies can be posted on each comment thread in a studio."

Samq64:
@BarelySmooth I committed the suggestions and made the addon dynamic.

addons/remaining-replies/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+      // substring in order to remove the "comments-" prefix
+      (comment_in_list) => comment_in_list.id == parentCommentID.substring(9)
+    )[0];
+
+    const remainingReplies = 25 - parentCommentData?.reply_count;
+
+    const span = document.createElement("span");
+    span.classList.add("sa-replies-remaining");
+    if (remainingReplies > 10) span.classList.add("sa-replies-remaining-hide");
+    span.style.marginLeft = "0.25em";
```
Samq64, line 23: Would it be better to include a leading space in the message?

Samq64 (approved):

addons/remaining-replies/userscript.js
WorldLanguages: Probably not, as the leading space may get ignored by translators.

WorldLanguages:
Comment sections on studios don't support RTL, right?

mxmou:
> Comment sections on studios don't support RTL, right?

RTL support for studios was implemented but then reverted for some reason: https://github.com/scratchfoundation/scratch-www/commit/604238d

addons/remaining-replies/addon.json
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  "settings": [
+    {
+      "name": "Only show if less than 10 replies remain",
+      "id": "show_only_if_less_than_10",
+      "type": "boolean",
+      "default": true
+    }
+  ],
+  "versionAdded": "1.27.0",
```
WorldLanguages, line 31:
```suggestion
  "versionAdded": "1.36.0",
```

addons/remaining-replies/addon.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "Remaining reply counter",
+  "description": "Shows you how many more replies can be posted on each comment thread in a studio.",
```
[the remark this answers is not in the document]
DNin01: That's a good idea. I would personally think to say that in the description.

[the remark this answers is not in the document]
Samq64:
```suggestion
  "description": "Shows you how many more replies can be posted on each comment thread in a studio before reaching the limit of 25.",
```

WorldLanguages:
It's not clear to me why comments need to be observed through Redux and "waitForElement" at the same time, but it appears to work.

addons/remaining-replies/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+
+    const parentCommentID = parentComment.getAttribute("id");
+    const parentCommentData = addon.tab.redux.state.comments.comments.filter(
+      // substring in order to remove the "comments-" prefix
+      (comment_in_list) => comment_in_list.id == parentCommentID.substring(9)
+    )[0];
+
+    const remainingReplies = 25 - parentCommentData?.reply_count;
+
+    const span = document.createElement("span");
```
WorldLanguages, line 20: Shouldn't it check first if such span already exists in the comment?
This function "addRemainingReplyCount" can be called more than once on the same comment. It appears to be intentional.
This causes a bug

WorldLanguages (changes requested):
Still buggy

Samq64:
Should the reply threshold be customizable?

Samq64:
> Shouldn't it check first if such span already exists in the comment? This function "addRemainingReplyCount" can be called more than once on the same comment. It appears to be intentional. This causes a bug image

I fixed that along with a few other edge cases. The only one left that I can think of is deleting a reply.

WorldLanguages:
> Should the reply threshold be customizable?

@Samq64 Sorry, what do we mean by "reply threshold"?

TheColaber (approved):

TheColaber:
@WorldLanguages waiting for your approval to merge

addons/remaining-replies/addon.json
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  "settings": [
+    {
+      "name": "Only show if less than 10 replies remain",
+      "id": "show_only_if_less_than_10",
+      "type": "boolean",
+      "default": true
+    }
+  ],
+  "versionAdded": "1.36.0",
```
WorldLanguages, line 31:
```suggestion
  "versionAdded": "1.39.0",
```

WorldLanguages:
Nice work
I'm not going to test it carefully again, it's a studio feature which is disabled by default.

Issue #3933, opened 2021-12-15
"See also" section in addons


lisa-wolfgang:
This would be an optional addon manifest property that would display links to other addons.
_Originally suggested here: https://github.com/ScratchAddons/ScratchAddons/issues/2031#issuecomment-811075546_

Here's an idea for what this could look like:
```
{
  "seeAlso": [
    {
      "addon": "custom-block-shape",
      "description": "To change the shape of blocks."
    },
    {
      "addon": "editor-dark-mode",
      "description": "To change the colors of the editor."
    }
  ]
}
```
The `addon` field would be the ID of the related addon, and the `description` field would specify how the addon is related.
_Originally suggested here: https://github.com/ScratchAddons/ScratchAddons/issues/3444#issuecomment-995184740_

Implementation details to consider:
1. Where should these related addons be displayed? Ideas:
    - At the bottom of the addon description (below the presets)
    - Below the addon description and notices (above the settings)
    - As a context menu shown by enabling the addon
    - On the right side of the addon description (below the main toggle)
    - Don't show them at all; only use them for implementing #3444
2. How do we fit descriptions into the section? Ideas:
    - Display them next to the links (the links would be laid out vertically)
    - Have them as tooltips (less discoverable, but it wouldn't be as bad if we implemented a custom tooltip that would appear instantly)
    - Don't show them at all; they would only be included for the developers' convenience
3. We would also need to figure out how to get the user to the addon. We don't currently have direct URLs to addons (although that would be nice, especially to show/share addons) so we would need to implement that somehow. Ideas:
    - Scroll the page to that addon (and possibly switch to the "All" page since the addon might not be in the currently selected category)
    - Switch to the "All" page and programmatically enter that addon's name in the search bar
    - Do something to the effect of the previous one, except have a specialized view for it with some kind of "Back to addon list" button
    - Have a mini-popup show with just the linked addon
    - Don't do anything and make the user search for the addon themselves
4. Sometimes it would be better to link directly to an addon *setting* instead of the whole addon. For example, `fullscreen` would benefit from linking to the full screen background color setting in `editor-dark-mode` instead of just saying, "Go find the full screen background color setting buried inside the dozens of other settings in the addon." However, this would make the implementation a bit more complicated -- is it worth it?

Samq64:
My opinions on this:

Under themes.
> How do we fit descriptions into the section?

Next to the links.
[...]

PR #4494, opened 2022-04-13
Better addon linking in settings page

### LINKED ISSUE #3934: Better addon linking

The current addon linking system could be improved.

- Adding a button next to addons that opens them as a direct link. This would make addon linking more discoverable.
- Automatically Refreshing the page when the hash changes
- Switching to the `all` category and expanding collapsed categories
- Expanding the addon, even if it's not enabled or collapsed by default
- Hiding all other addons?
- Sort of related: Allow linking to different categories or the more settings page if that doesn't already exist.

Using the current addon linking system which works pretty well apart from needing to reload the page.

### Additional context

This would make implementing  #3933 easier.

Norbiros:
Resolves #3934

### Changes

Now when you add `addon-{addon-id}` as hash to your URL, it will change category to all, find specific addon, open category and open addon's settings.

### Reason for changes

This would make implementing https://github.com/ScratchAddons/ScratchAddons/issues/3933 easier.

### Tests

Everything works fine (Firefox).


RedGuy12:
Could you make the linking work with easterEggs too?

Norbiros:
I don't think it would be great idea. If I did something like that, the user wouldn't have to type _secret code_. Just add `#addon-dango-rain` at end of link... But it is possible, NP

lisa-wolfgang:
If someone wants to send a link to an easter egg addon but can't, they could easily tell the person how to show easter eggs. For this reason, I don't think it makes sense to not allow easter egg addons to be linked to.

Weredime (commented):
Now we can link dango-rain like [this](https://scratch.mit.edu/scratch-addons-extension/settings#addon-dango-rain)!
Maybe add a little paperclip icon that copies the link to the clipboard, that you can share with other SA users?

Norbiros:
This is great idea, but:
- This link is different for other computers/browsers: for moz-extension://3b232903-3739-45b6-9498-3c27c198a6ba/webpages/settings/index.html (mox-extension and id can change). 
- Paperclip addon can take some part of name, maybe Shift + Click or something like that?

Weredime:
> This link is different for other computers/browsers: for moz-extension://3b232903-3739-45b6-9498-3c27c198a6ba/webpages/settings/index.html (mox-extension and id can change).

Doesn't https://scratch.mit.edu/scratch-addons-extension/settings redirect to the settings page, regardless of which computer/browser you are on? The redirect saves hash & query params, which is also a good thing

Norbiros:
Oh, I didn't know that, thanks :) So where this paper clip should be? Near undo button?

Weredime:
> So where this paper clip should be?

Like this (next to undo button)?
image
Also, for the correct icon, use https://icon-sets.iconify.design/uil/link/

webpages/settings/components/addon-body.js
```diff
@@ -65,6 +65,11 @@ export default async function ({ template }) {
           console.log(`Loaded default values for ${this.addon._addonId}`);
         }
       },
+      copyAddonLink() {
+        navigator.clipboard.writeText(
+          `https://scratch.mit.edu/scratch-addons-extension/settings#addon-${this.addon._addonId}`
+        );
+      },
```
Weredime, line 72: We need some sort of user input feedback, so the user knows that it was successfully copied. Maybe some sort of speech bubble on top?

Weredime (changes requested):

webpages/settings/components/addon-body.js
lisa-wolfgang: A green "Copied" label to the left of the button would be easier and just as effective.

mxmou: Replacing the button with a checkmark for a short interval might be better if the window is narrow and adding a label would cause the addon name to wrap at a different point.

Norbiros: I really like your idea. Can you give me icon link/svg file? Maybe something like in read.svg?
read

Norbiros: Now I'm using https://icon-sets.iconify.design/charm/tick/

Weredime:
@Norbiros Could you disable clicking on the icon when the checkmark is visible?

webpages/settings/components/addon-body.js
mxmou: Use the existing `check.svg`.

Weredime (dismissed):

Norbiros:
So there is anything to do in this PR?

WorldLanguages:
I don't really agree with this suggestion on the issue this PR resolves:

> Adding a button next to addons that opens them as a direct link. This would make addon linking more discoverable.

I don't think the average user will understand what the button does. And it's another button all users have to mentally process when looking at the settings page.

Norbiros:
Maybe we can add new setting for that?

lisa-wolfgang:
Perhaps a simpler alternative would be to make opening the addon settings view change the URL in the address bar. This would also have a side benefit of allowing individual addons to live in a user's browser history.

WorldLanguages:
> Perhaps a simpler alternative would be to make opening the addon settings view change the URL in the address bar. This would also have a side benefit of allowing individual addons to live in a user's browser history.

This could work - however it wouldn't be perfect, and I would need to try out the implementation before knowing if it's good enough.

WorldLanguages:
> Maybe we can add new setting for that?

We won't be adding settings to change the UI/UX of our own interfaces.

mxmou:
> Perhaps a simpler alternative would be to make opening the addon settings view change the URL in the address bar. This would also have a side benefit of allowing individual addons to live in a user's browser history.

I would not expect expanding an addon to add it to the history - it would change the behavior of the back button too.

Norbiros:
WorldLanguages said:

> I'd say remove the button
> we can find a different way in the future with another PR

apple502j (approved):

Weredime (approved):

apple502j (changes requested):
Blocking merge, will push some fix

apple502j (approved):

apple502j:
I'll let LGTM run, then maybe ask for third review (since I rewrote the entire code), then merge?

Weredime (approved):

Issue #2670, opened 2021-06-04
Let users choose language for addon names/descriptions independent from browser UI language


WorldLanguages:
Using the browser UI language for localizing things like the extension name, description, tag names (e.g. "recommended") is a good idea - and it's reasonable for these to be consistent with the browser UI language.
However, because the addon localization system is not dependent on `chrome.i18n`, we can use any language for addon properties.
We could add a setting to "more settings" that lets users choose a language for addon manifests. The extension would need a restart after changing the language.
The text in the settings UI would stay in the same language, but the addon names, descriptions, setting names, etc. would be in whatever language the user chose (probably English).
It could also be technically possible to move everything into our own system but I don't think it's worth exploring. `chrome.i18n`'s non-async nature is useful since we can't use top level await yet. Loading manifests is async anyway, so reading `chrome.storage` first is not a problem.
We could store the language choice in `chrome.storage.local` or `chrome.storage.sync`, I vote local. And I'm not sure if it should be exported when using the import/export function.
Whether the UI is in RTL/LTR probably shouldn't be affected depending on the chosen language, since the main settings UI text will continue to be in the browser UI language.

RedGuy12:
Re: moving away from chrome.i18n to our own system

I think it IS worth exploring. If we move, we could use 'Structured JSON', which is a cross between key-value json and Chrome i18n json. It enables support for plurals and translator context simultaneously in l10n files.

<https://docs.transifex.com/formats/json/structured-json>https://docs.transifex.com/formats/json/structured-json
________________________________
From: World_Languages ***@***.***>
Sent: Friday, June 4, 2021 5:39:42 PM
To: ScratchAddons/ScratchAddons ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [ScratchAddons/ScratchAddons] Let users choose language for addon names/descriptions independent from browser UI language (#2670)


Using the browser UI language for localizing things like the extension name, description, tag names (e.g. "recommended") is a good idea - and it's reasonable for these to be consistent with the browser UI language.
However, because the addon localization system is not dependent on chrome.i18n, we can use any language for addon properties.
We could add a setting to "more settings" that lets users choose a language for addon manifests. The extension would need a restart after changing the language.
The text in the settings UI would stay in the same language, but the addon names, descriptions, setting names, etc. would be in whatever language the user chose (probably English).
[...]

PR #4498, opened 2022-04-15
Support loading Structured JSON translations

apple502j:
Support loading Structured JSON translations. Note that this PR does not introduce the new format. Structured JSON is specified at https://docs.transifex.com/formats/json/structured-json.

Future plans:

- `addons-l10n/en` will use a mix of key-value json and structured json; current syntax works fine, but you can also specify an object that meets the structured json spec.
- Since key-value json is not a valid structure json syntax, the script will convert that to actual structured json.
- After pulling the translated strings as structured json, the script will convert them to key-value json to save some bytes. Translations on this repository do not need developer comments, etc.

### DIFF
--- a/libraries/common/cs/l10n.js
+++ b/libraries/common/cs/l10n.js
@@ -26,7 +26,9 @@ export default class LocalizationProvider extends EventTarget {
   _get(key, placeholders, messageHandler, fallback) {
     messageHandler = messageHandler || ((m) => m);
     if (Object.prototype.hasOwnProperty.call(this.messages, key)) {
-      const message = messageHandler(this.messages[key]);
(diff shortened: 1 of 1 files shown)


RedGuy12:
Related:

> Re: moving away from chrome.i18n to our own system
> 
> I think it IS worth exploring. If we move, we could use 'Structured JSON', which is a cross between key-value json and Chrome i18n json. It enables support for plurals and translator context simultaneously in l10n files.
> 
> https://docs.transifex.com/formats/json/structured-json

_Originally posted by @RedGuy12 in https://github.com/ScratchAddons/ScratchAddons/issues/2670#issuecomment-855153020_

apple502j:
Merging

PR #4506, opened 2022-04-16
hide-flyout changes

### LINKED ISSUE #1613: Slider not Working properly when block palette is hidden.
**Slider not Working properly when block palette is hidden.**

When we hide the block pallet, users need more space to code! But in spite of hiding the palette, the slider is not sliding any way more.

**To Reproduce**

1. Go to the scratch editor.
2. Hide the block palette.
3. try to slide the button slider left-wards.
4. The slider is getting stopped where the block pallet should have been and sliding no more.

**Expected behaviour**

The slider is getting stopped here, and sliding no more.

**Screenshots**

Screenshot 2021-02-17 170816
The slider is getting stopped here, and sliding no more.

- Windows 10
- chrome
- 87

### LINKED ISSUE #1877: Palette is initially visible when auto-hide block palette is enabled

### LINKED ISSUE #2015: Auto-hide block palette causes dragged blocks to be hidden behind the sprite selector
**Describe the bug**

When the `hide-flyout` addon is enabled, blocks being dragged over the sprite selector are invisible.

**To Reproduce**

1. Enable "auto-hide block palette"
2. Open the editor
3. Drag a block to another sprite
4. Note that the block is invisible when dragged over the sprite selector because the code editor has `overflow: hideen`

**Expected behavior**

They would be visible. A different solution is necessary to prevent the hidden palette from overlaying the stage in `editor-stage-left` (the reason why `overflow: hidden` is there).

### LINKED ISSUE #3996: Lock shows in front of stage
### Description

In the auto hide block palette, the lock somehow if appearing in the front of the stage.?
### Repoduction steps
Enable auto hide block palette. make it hide.

### Expected behaviour
The lock will hide.

### Actual behaviour

it doesn't.

### Screenshots
Screenshot 2021-12-28 9 17 56 AM

### Environment

- Browser: Chrome 96.0.4664.111
- Scratch Addons version: 1.23.2
- Operating system: Chrome OS

mxmou:
Resolves #1877
Resolves #2015
Resolves #3996

### Changes

* Makes auto-hiding block palette work correctly in RTL
* Hides the palette when opening the editor
* Makes the lock icon visible and functional in both palette area hover and category hover modes
* Makes the flyout remain open if the user started dragging inside it (mainly intended to make dragging the scrollbar easier)
* Extends the horizontal scrollbar into the additional space in category click mode (doesn't work in RTL)
* Fixes dragged blocks being hidden behind the sprite selector
* Makes all categories appear as unselected when the palette is closed in category click mode
* Removes the scrolling animation when opening the flyout (but not switching between categories) in category click mode


WorldLanguages:
@mxmou Excellent, just one thing.

In "category click mode", if you open the palette, then switch to another tab (Costumes, Sounds, etc), then come back to the Code tab, the palette is now closed.
Is this intentional? If not, I feel it should remember the opened state before switching tabs.
Currently on master, it defaults to opening the palette even if it was previously closed. You could see it as the same issue backwards, but I think changing the palette to be hidden by default makes it more noticeable and inconvinient.

WorldLanguages:
TODO: add an "updated" tag to this addon. This is the least used recommended addon (see #4100), and that might be caused by the bugs. There's 100+ addons so reminding people this exists should be a good idea.

mxmou:
> @mxmou Excellent, just one thing.
> 
> In "category click mode", if you open the palette, then switch to another tab (Costumes, Sounds, etc), then come back to the Code tab, the palette is now closed. Is this intentional? If not, I feel it should remember the opened state before switching tabs.

I agree - remembering the state of the palette is the correct behavior. I'll try to fix this next week when I get home.

WorldLanguages:
This can be moved to the next release if there's not enough time to develop or test

mxmou:
Switching tabs in category click mode now works as expected. I also fixed a few other bugs and made the difference between locked and unlocked more visible. I noticed that the lock icon has no alt text - adding a tooltip would be a good idea but I won't include it in this PR because it would require new strings.

WorldLanguages:
Found a bug, so moving this to the next milestone, so we have more time to test the changes (and we could actually advertise the addon again now that it's less buggy). So the tooltip text for the lock icon can be added inside this PR.

WorldLanguages:
Bug on "category click" mode:
1. Go to https://scratch.mit.edu/projects/104/editor/
2. Show the flyout (click on any category)
3. See project page
4. See inside
5. Show the flyout - it's not shown

apple502j (approved):

WorldLanguages:
I get this weird visual artifact in Chromium (look at the left border, after I move the yellow block, there's a yellow line):

brave_5GcTzM61QU

After refreshing the same project, it becomes 2 blue lines ???
brave_c9cIHNDZCV

WorldLanguages:
> we could actually advertise the addon again now that it's less buggy

I think we could create another PR later that addresses this, instead of reusing this one, so I created the #4668 issue to discuss what we could do.

WorldLanguages:
This is good to merge as-is but I wonder if those weird lines can be fixed by CSS, or if it's weird Chrome renderer behavior?

Samq64:
Seems to be a element shifting issue.

https://user-images.githubusercontent.com/81489795/173258529-acf8f121-5f0f-48ac-9fcb-e43b572cc75d.mp4

WorldLanguages:
@Samq64 Oh, editor-stage-left (you have it enabled) helps to visualize what is going on! interesting

WorldLanguages:
@mxmou Was that last commit supposed to fix the issue? I still experience it

mxmou:
It can still occur depending on the exact `devicePixelRatio` but I don't think it's possible to avoid that entirely.

WorldLanguages:
Found a bug (default addon settings):
1. Go to https://scratch.mit.edu/projects/104/editor/
2. Go to costumes tab
3. See project page
4. See inside

image

addons/hide-flyout/style.css
```diff
@@ -100,7 +100,9 @@
   right: calc(var(--sa-toolbox-width) - 15px - 32px);
 }
 
-.gui_stage-and-target-wrapper_69KBf {
+/* Setting these styles on gui_stage-and-target-wrapper breaks full screen */
+[class*="stage-wrapper_stage-wrapper_"],
+[class*="gui_target-wrapper_"] {
```
WorldLanguages, line 105: These styles affected player mode too:
image

PR #1136, opened 2020-12-30
Export and Import Settings

apple502j:
Resolves #1132 

Rewrote exsisting:tm: PR (#1134). Uses storage API directly instead of using DOM, and exports light theme status.

PR #4510, opened 2022-04-17
Add a modal component

mxmou:
### Changes

Moves the implementation of the More Settings modal into a separate component. This doesn't change anything, but it allows other modals to be added, specifically for preset switching.

### DIFF
--- a/webpages/settings/components/modal.js
+++ b/webpages/settings/components/modal.js
@@ -1,0 +1,35 @@
+export default async function ({ template }) {
+  const Modal = Vue.extend({
+    props: ["isOpen", "title"],
+    template,
+    data() {
+      return {
+        canCloseOutside: false,
+      };
+    },
+    watch: {
+      isOpen(value) {
+        if (value) {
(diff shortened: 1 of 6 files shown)


webpages/settings/components/modal.js
```diff
@@ -0,0 +1,35 @@
...
+      },
+    },
+    methods: {
+      msg(...params) {
+        return this.$root.msg(...params);
+      },
+    },
+    events: {
+      clickOutside(e) {
+        if (this.isOpen && this.canCloseOutside && e.isTrusted) {
```
apple502j, line 28: Do we need isTrusted check?

WorldLanguages: This is probably code from stackoverflow?

mxmou: It was added in #1136 to ignore the click event from the hidden file input.

apple502j: ah, makes sense

PR #4541, opened 2022-05-01
Update mystuff.css

MWB6763:
Bold these buttons


mxmou:
The current design matches the Scratch 3.0 style, which is the purpose of this addon.

PR #4553, opened 2022-05-10
`discuss-button`: new option - compact account drop down

retronbv:
Resolves None

### Changes

Adds a compact navbar addon that removes the username from the navbar and adds it in the dropdown.
image
image

### Reason for changes

Makes the Scratch navbar UI look cleaner.

### Tests

I have tested in scratch-www and in scratchr2. 

### Issues

With the `scratchr2` addon there is a slight margin error (see image below).

scratchr2 margin error

There is also a problem with the `account-settings-capitalize` addon. I presume this could be fixed via checking if this addon is enabled with `addon.self.getEnabledAddons()` in the account settings capitalization addon.

I would appreciate it if someone would take the liberty to PR into my fork with fixes to the above issues.


addons/compact-nav/addon.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Compact Navbar",
```
RedGuy12, line 2: This name is a bit broad for such a narrow addon

addons/compact-nav/addon.json
```diff
@@ -0,0 +1,24 @@
...
+  "description": "Makes a more compact navbar.",
```
RedGuy12, line 3: Elaborate?

RedGuy12 (changes requested):
Also enable actions on your fork

Chiroyce1:
The navbar is the whole part
image

So I think navbar drop-down or something similar is a better name.

addons/compact-nav/addon.json
RIMOPA: Moves the username of the navbar pop-up to inside the pop-up, making a more compact navbar.

retronbv:
> The navbar is the whole part
> 
> image
> 
> 
> 
> So I think navbar drop-down or something similar is a better name.

Yes, i agree. I am actually re implementing this into the customizable navbar addon.

Samq64:
If this will be part of custom navbar then maybe an option to remove the duplicate my stuff button from the navbar could be useful.

mxmou:
> There is also a problem with the `account-settings-capitalize` addon. I presume this could be fixed via checking if this addon is enabled with `addon.self.getEnabledAddons()` in the account settings capitalization addon.

You could avoid this by replacing the "Profile" text with the username instead of adding a new item.

retronbv:
> > There is also a problem with the `account-settings-capitalize` addon. I presume this could be fixed via checking if this addon is enabled with `addon.self.getEnabledAddons()` in the account settings capitalization addon.
> 
> You could avoid this by replacing the "Profile" text with the username instead of adding a new item.

I feel people would ask where the profile button went after they enable this... Maybe `[USERNAME] (Profile)` would be better text. I will get working on this

Norbiros:
> > > There is also a problem with the `account-settings-capitalize` addon. I presume this could be fixed via checking if this addon is enabled with `addon.self.getEnabledAddons()` in the account settings capitalization addon.
> > 
> > 
> > You could avoid this by replacing the "Profile" text with the username instead of adding a new item.
> 
> I feel people would ask where the profile button went after they enable this... Maybe `[USERNAME] (Profile)` would be better text. I will get working on this

What if user have very long nickname? `(Profile)` wouldn't be visible.

mxmou:
> Why not say `<username>'s profile`?

It could be too long. It also wouldn't clearly indicate that it's the username of the user who is currently logged in.

WorldLanguages:
> Why not say `<username>'s profile`?

This is also harder to localize into other languages.

Secret-chest:
> > Why not say `<username>'s profile`?
> 
> This is also harder to localize into other languages.

We can use placeholders.

Secret-chest:
> > Why not say `<username>'s profile`?
> 
> It could be too long. It also wouldn't clearly indicate that it's the username of the user who is currently logged in.

Have it as an option?

apple502j:
@Secret-chest Doesn't seem worth it.

Secret-chest:
> @Secret-chest Doesn't seem worth it.

Seriously? Are you GNOME?

mxmou:
> > > Why not say `<username>'s profile`?
> > 
> > 
> > This is also harder to localize into other languages.
> 
> We can use placeholders.

Possessive forms aren't as simple as adding `'s` in every language. For example, the correct form might depend on the user's gender. This means that it would have to be translated as "profile of user &lt;username&gt;" or something similar, which would be unnecessarily long.

> > It could be too long. It also wouldn't clearly indicate that it's the username of the user who is currently logged in.
> 
> Have it as an option?

It wouldn't be very useful because it doesn't change the addon's functionality in any way.

WorldLanguages (dismissed):
Needs Prettier lint before merging

WorldLanguages:
> You could avoid this by replacing the "Profile" text with the username instead of adding a new item.

@mxmou Do you want to implement this?

retronbv:
If mxmou could that would be great, the way I tried was very sucky and didn't look good.

retronbv:
ill run prettier

TheColaber (approved):

Secret-chest:
Having a multiline item here is very strange. Why just not use Profile (username)?

addons/discuss-button/addon.json
```diff
@@ -59,13 +75,19 @@
...
+      "type": "boolean",
+      "default": false
     }
   ],
   "versionAdded": "1.0.0",
   "latestUpdate": {
-    "version": "1.24.0",
-    "newSettings": ["items"],
-    "isMajor": true
+    "version": "1.26.0",
```
apple502j, line 88:
```suggestion
    "version": "1.28.0",
```

retronbv:
pog

PR #4559, opened 2022-05-12
Don't send message if fallback is set

Norbiros:
### Changes

Send message "key missing", when fallback is not set. 

### Reason for changes

Scratch has a lot of spam in console. We don't want to make it worse.
*I am talking about debbuger's name*

### Tests

Firefox 95, everything works great.

### DIFF
--- a/libraries/common/cs/l10n.js
+++ b/libraries/common/cs/l10n.js
@@ -31,7 +31,8 @@ export default class LocalizationProvider extends EventTarget {
       const message = messageHandler(rawMessage.string || rawMessage);
       return this.formatter.format(message, placeholders);
     }
-    console.warn("Key missing:", key);
+    if (fallback == "")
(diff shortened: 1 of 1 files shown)


apple502j:
This is fine

Norbiros:
> This is fine

We have a lot of spam in scratch's console. If you are trying to develop something, you don't know if that error is made by you, or by scratch, or by SA. `debbuger/@name` spams when you are changing sprites, so you have a lot of this in your console. 
Okay, this solution might not be the best (why debbuger's name is not translated to other languages), but when you know that you need to use fallback, why it should print warning in console.

apple502j:
Okay, if it can appear multiple times this is probably worth fixing

apple502j:
Closing still, this does not function as intended, PRing myself

PR #4573, opened 2022-05-15
Sound duration: Fix selecting nothing showing 0 seconds

TheColaber:
Fix selecting nothing showing 0 seconds

And also, this pr changes the display to minutes when sound is longer than 5 seconds. We can play with this number since it's probably better to raise it to a number like 10 or 20.


apple502j:
30 seems like the good threshold.

mxmou:
It might be better to always show something like `0:00.00`.

TheColaber:
Yeah, your right. Done.

PR #4575, opened 2022-05-15
dark-www: fix stage monitor colors

### LINKED ISSUE #4560: Website dark theme affects stage
### Description

Website dark mode affects the stage

### Reproduction steps

1. Enable dark-www (disable editor-dark-mode) with either of the Dark presets
2. Go to a project with variable monitors shown (on the project page only, not the editor)
3. See bug

### Expected behavior

The stage is unaffected
### Actual behavior

The stage is affected

### Screenshots

image
### Environment

not applicable
### Additional context
Reported by Cubester on Discord

apple502j:
Resolves #4560 
Sadly there is no CSS keyword that reverts to previously defined non-inherited style. Also this is not a prio-2.

### DIFF
--- a/addons/dark-www/experimental_scratchwww.css
+++ b/addons/dark-www/experimental_scratchwww.css
@@ -9,13 +9,6 @@ body {
   background-color: var(--darkWww-page);
   color: var(--darkWww-page-text);
 }
-div[class*="stage-wrapper_"] {
-  /*
-    Revert the color specified above. Editor dark mode provides its own CSS.
-    See also: #4560 and LLK/scratch-gui#5797
-  */
-  color: #575e75;
-}
 .subactions .share-date,
 .conf2019-description {
   color: var(--darkWww-page-text);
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Maybe it would be better to stay consistent with how editor-dark-mode fixes this (or change both):

```css
.monitor_label_ci1ok,
.monitor_list-empty_1UKsB {
  color: #575e75;
}
```

https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/editor-dark-mode/experimental_editor.css#L582

mxmou:
@WorldLanguages I think this is better than specifically styling monitors because other elements could be added to the stage in the future. Editor dark mode could be changed to use the same selector.

For consistency, I suggest moving it next to `.project-info-alert` near the bottom of `expermiental_scratchwww.css`.

WorldLanguages:
Okay, so for consistency, let's change editor-dark-mode too, as @mxmou said.

PR #4592, opened 2022-05-19
add traditional chinese

foxcat-test:
Resolves #

### Changes

add ZH_TW(traditional chinese) 

### Reason for changes

make it better

### Tests

just add it:)


RedGuy12:
https://www.transifex.com/scratch-addons/scratch-addons-extension/

Issue #1719, opened 2021-03-03
Make the setting JSON file look cleaner


scratchusernamemrtbts:
**Is your feature request related to a problem? Please describe.**

I want to be able to edit the setting by using the JSON file but it is hard to read
So I can edit only the JSON file and import to every browser I use
**Describe the solution you'd like**

Make the SA save the setting file in a more cleaner file
**Describe alternatives you've considered**

 N/A

**Additional context**

N/A

TheColaber:
By "cleaner", do you just mean formatted?

scratchusernamemrtbts:
Yes

scratchusernamemrtbts:
So it will be easy to read

WorldLanguages:
JSON is for machines, not humans. If you really want to edit it, get a "code editor" (json isn't code but you get the point lol) that will format it for you.

scratchusernamemrtbts:
> JSON is for machines, not humans. If you really want to edit it, get a "code editor" (json isn't code but you get the point lol) that will format it for you.

Thanks

PR #4600, opened 2022-05-22
Beautify settings on the "View settings file" page

Explosion-Scratch:
This annoys me so much

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -301,7 +301,7 @@ let fuse;
       viewSettings() {
         const openedWindow = window.open("about:blank");
         serializeSettings().then((serialized) => {
-          const blob = new Blob([serialized], { type: "text/plain" });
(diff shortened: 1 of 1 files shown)


Chiroyce1:
Well, it was mainly intended for being copy/pasted and not directly readable.

RIMOPA:
> Well, it was mainly intended for being copy/pasted and not directly readable.

If it is intended for being copy/pasted we must add a copy button like this

scratchusernamemrtbts:
uhh #1719

WorldLanguages:
The "view settings file" feature was implemented so that users can copy the JSON. It's not intended to be readable.

Samq64:
I agree that adding a copy button, styles, etc. is unnecessary, but i think changing one line of code to format the output is a good fine.

WorldLanguages:
@Samq64 It makes the JSON harder to copy if the user isn't aware of the Ctrl+A hotkey.

PR #4608, opened 2022-05-25
Improve addon update notices

### LINKED ISSUE #4606: Only show update notices from the latest version

Addon update notices are still displayed for addons that were updated in previous versions. After enough time, most addons will have notices and it will be harder to notice the ones that were actually updated.

Hide notices from previous versions.

mxmou:
Resolves #4606

### Changes

Hides updates from previous versions and makes their appearance more similar to normal notices.

### DIFF
--- a/webpages/settings/components/addon-body.js
+++ b/webpages/settings/components/addon-body.js
@@ -30,6 +30,12 @@ export default async function ({ template }) {
       addonSettings() {
         return this.$root.addonSettings[this.addon._addonId];
       },
+      showUpdateNotice() {
+        if (!this.addon.latestUpdate || !this.addon.latestUpdate.temporaryNotice) return false;
+        const [extMajor, extMinor, _] = this.$root.version.split(".");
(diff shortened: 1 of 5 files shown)


webpages/settings/components/addon-body.js
```diff
@@ -30,6 +30,12 @@ export default async function ({ template }) {
       addonSettings() {
         return this.$root.addonSettings[this.addon._addonId];
       },
+      showUpdateNotice() {
+        if (!this.addon.latestUpdate || !this.addon.latestUpdate.temporaryNotice) return false;
+        const [extMajor, extMinor, _] = this.$root.version.split(".");
```
RedGuy12, line 35: Why not just not destructure the unused variables?

mxmou: The same code is already used in `index.js` for the "New" tag and in another function in the same file for the "New settings" tag. I think it would make sense to move it into a module and make any changes there.

apple502j (approved):

PR #4609, opened 2022-05-25
Hide stage addon

### LINKED ISSUE #1453: Hide the stage/sprites view
**Describe the new addon**

By clicking a button located at the right of the editor area, the element with class `gui_stage-and-target-wrapper_69KBf` would glide out of view.
When combined with the auto-hide block palette addon, there would be a HUGE amount of space:

image

mxmou:
Resolves #1453

### Changes

Adds an addon that allows the stage and sprite pane to be hidden completely.
image


Norbiros:
This works correctly with `editor-stage-left`?

Samq64:
We're already running low on space above the stage, so could this addon and an option to remove the small stage button?

mxmou:
> > We're already running low on space above the stage, so could this addon and an option to remove the small stage button? (and maybe even an option to remove all stage size buttons altogether for those who never use small stage?)
> 
> Possibly turn the stage size button into a cycling toggle.

A dropdown menu would be easier to use. It could look like this:
image

This would require an extra click to change the size, so it should probably be optional.

WorldLanguages:
@mxmou Hmm, toggling the hide stage mode while on the costume tab can cause weird aspect ratio issues on the costumes tab:

image

It's fixed by switching tabs again (costumes to code, then back to costumes)

WorldLanguages:
Feedback: 

> In regards to the new "hide stage and sprite panel" extension, I have a suggestion: as soon as I enabled it, I thought "hey, this could be really good for costume editing!" And then I realised it only stretched the editing panel. Something I would like to see is the editing panel resized so that either all the tools are on one side, or the tools usually on the top move to the side when the stage and sprite panel are hidden.

PR #4627, opened 2022-05-31
More dynamic addons

TheColaber:
#1673

Read commits and their descriptions for information.

If you guys don't merge this yet, I'll add some more addons, but this is what I got done.

### DIFF
--- a/addons/exact-count/addon.json
+++ b/addons/exact-count/addon.json
@@ -58,6 +58,8 @@
       "runAtComplete": false
     }
   ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
   "versionAdded": "1.2.0",
   "tags": ["community", "recommended", "profiles", "forums", "studios"],
   "enabledByDefault": false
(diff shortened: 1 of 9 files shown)


Chameleon-Coder (commented):
Looks awesome!

addons/exact-count/addon.json
```diff
@@ -58,6 +58,8 @@
       "runAtComplete": false
     }
   ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
```
mxmou, line 62: `dynamicDisable` should be `false` here because the userscript doesn't actually support disabling.

apple502j (approved):

mxmou (changes requested):

addons/blocks2image/userscript.js
```diff
@@ -44,71 +44,75 @@ export default async function ({ addon, global, console, msg }) {
...
-
-      items.splice(
-        insertBeforeIndex,
-        0,
-        {
-          enabled: true,
-          text: msg("export_selected_to_SVG"),
-          callback: () => {
-            exportBlock(false, block);
+      if (!addon.self.disabled) {
```
RedGuy12, line 86: use early returns so the diff is cleaner pls

Chameleon-Coder (commented):
awesome work

apple502j (approved):

PR #4632, opened 2022-06-02
Forum post snip addon

### LINKED ISSUE #4538: "Collapse nested forum quotes" addon

Sometimes people do this...

Short forum quote inside a short forum quote inside a short forum quote...

...and it's difficult for your gosh darn brain to parse, especially if you're reading chronologically (oldest posts to newest, one by one).

An addon which automatically collapses nested forum quotes! I'm thinking it could look something like this:

Thrice nested forum quotes, but the deepest one is just "WOFFAN123 wrote:" with an "expand" link/button

Ideally it would let you decide how many levels of nesting you want to show before collapsing. (I feel it should also auto-collapse all quotes *under* that, so you can expand further levels one at a time, but maybe others would disagree.)

This could also have an option to show a "collapse" button on *all* quotes, so that when someone quotes the entire first post (or similar), you can choose to manually hide that quote yourself.

Another option would be to just, like, completely erase the quote and replace it with "snip" or something. "Automatic forum snipper" is a funny addon name :P But it's less useful/interactive so it's probably not as good of an addon, even though it'd be a bit easier to code.

### Additional context

This is the quintessential userstyle/userscript, how has this never been suggested for Scratch Addons before. :package:

Weredime:
Resolves #4538

### Changes

Adds collapse forum quotes addon

### Tests

Firefox Developer Edition

### DIFF
--- a/addons/forum-post-snip/addon.json
+++ b/addons/forum-post-snip/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Nested forum post collapser",
-  "description": "Collapses forum quotes to a limit.",
+  "description": "When there are multiple forum quotes nested inside of each other, only shows a certain amount of them.",
   "credits": [
     {
       "name": "Weredime (9gr)",


addons/forum-post-snip/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/dist/schema.json",
+  "name": "Collapse nested forum quotes",
+  "description": "Collapses forum quotes",
```
apple502j, line 4: Too short and does not describe the functionality well.

addons/forum-post-snip/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/dist/schema.json",
+  "name": "Collapse nested forum quotes",
```
apple502j, line 3: Avoid verbs where possible

addons/forum-post-snip/addon.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["topics"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Post collapse",
```
apple502j, line 19: Not very descriptive

apple502j (dismissed):
Also prettier is failing.

addons/forum-post-snip/addon.json
Weredime: that's todo

addons/forum-post-snip/addon.json
Weredime: to be honest I'm not sure what to name settings and stuff....

Chameleon-Coder (commented):
The name's kinda good. I am very, _very_ confused with what @apple502j said - why can't addon creators use verbs?

WorldLanguages:
@Chameleon-Coder It only applies to addon names

Chameleon-Coder:
@WorldLanguages okay.

WorldLanguages:
We will be using the `status: needs review` for pull requests that require a second approval. Removing it from this PR since it lacks both.

addons/forum-post-snip/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Nested forum post collapser",
+  "description": "Collapses forum quotes to a limit.",
```
lisa-wolfgang, line 3: The purpose of this addon can be confusing initially, so we need to be careful about how we word this. I'd suggest something like:
```suggestion
  "description": "When there are multiple forum quotes nested inside of each other, only shows a certain amount of them.",
```

lisa-wolfgang (changes requested):

GrahamSH-LLK:
Is this just waiting on review?

PR #4636, opened 2022-06-04
New Addon: Self Referencing in Dropdowns ("myself" option for sensing blocks)

### LINKED ISSUE #3396: New feature "Touching myself"
Hi!

I propose a new feature in the "Block switching" addon (or "Developper tools", or an other...).

Let's have a look on this picture :

image

You see? If I make clones of this sprite, they cannot detect that they are touching each other... And that's a shame... 😥

Sombode:
Resolves #3396

### Changes

Allows certain sensing block menu dropdowns ("touching []", "distance to []", and "[] of []") to select its own sprite, similar to the "myself" option for creating a clone.

image

### Reason for changes

Useful when working with clones of a sprite, especially for detecting collisions between clones.

### Tests

Tested on Chromium 102, with and without the searchable dropdowns addon. No apparent issues.

### DIFF
--- a/addons/self-referencing/addon.json
+++ b/addons/self-referencing/addon.json
@@ -1,0 +1,20 @@
+{
+  "name": "Self referencing",
+  "description": "Adds the \"myself\" option to certain block dropdowns in the sensing category. Useful when working with clones.",
+  "credits": [
+    {
+      "name": "smalllevy"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.26.1",
+  "tags": ["editor"],
+  "enabledByDefault": false
+}
--- a/addons/self-referencing/addon.json
+++ b/addons/self-referencing/addon.json
@@ -14,7 +14,7 @@
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "versionAdded": "1.26.1",
+  "versionAdded": "1.27.0",
   "tags": ["editor"],
   "enabledByDefault": false
 }
--- a/addons/self-referencing/addon.json
+++ b/addons/self-referencing/addon.json
@@ -12,8 +12,6 @@
       "matches": ["projects"]
     }
   ],
-  "dynamicEnable": true,
-  "dynamicDisable": true,
   "versionAdded": "1.27.0",
   "tags": ["editor"],
   "enabledByDefault": false
--- a/addons/self-referencing/userscript.js
+++ b/addons/self-referencing/userscript.js
@@ -1,0 +1,18 @@
+export default async function ({ addon, global, console, msg }) {
+  const Blockly = await addon.tab.traps.getBlockly();
+  const vm = addon.tab.traps.vm;
+  //Most of this code was derived from GarboMuffin's "searchable dropdowns" addon
+  const oldFieldDropdownGetOptions = Blockly.FieldDropdown.prototype.getOptions;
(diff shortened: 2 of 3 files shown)


Norbiros (dismissed):
Great addition! Can you please just change version added from `1.26.1` to `1.27.0`?

Samq64:
Could you also add point towards and go to?

mxmou:
> Could you also add point towards and go to? (Probbably more confusing to understand though)

Point towards myself wouldn't make sense and go to myself would do nothing.

Secret-chest:
> > Could you also add point towards and go to? (Probbably more confusing to understand though)
> 
> Point towards myself wouldn't make sense and go to myself would do nothing.

For clones they would make sense.

TheColaber:
The addon says it supports "dynamicDisable" but you do not return on addon.self.disabled. Please make sure to do that \:)

TheColaber:
I do recommend adding dynamic enable and disable, it should be fairly simple for this addon. Before line 16, just add
```js
if (!addon.self.disabled) return options;
```
And then add back what you removed from the addon manifest.

addons/self-referencing/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Self referencing",
```
lisa-wolfgang, line 2: This needs to be more specific -- something like "self-referencing blocks" or "sprite self-referencing".

addons/self-referencing/addon.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "Adds the \"myself\" option to certain block dropdowns in the sensing category. Useful when working with clones.",
```
lisa-wolfgang, line 3: To avoid confusing the user, we should find a different way to word "the myself option". Perhaps something like "Adds the option to get information from the selected sprite (in addition to other sprites) for certain blocks."

addons/self-referencing/addon.json
```diff
@@ -0,0 +1,18 @@
...
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.27.0",
+  "tags": ["editor"],
```
lisa-wolfgang, line 16:
```suggestion
  "tags": ["editor", "codeEditor"],
```

addons/self-referencing/userscript.js
```diff
@@ -0,0 +1,26 @@
+export default async function ({ addon, global, console, msg }) {
+  const Blockly = await addon.tab.traps.getBlockly();
+  const vm = addon.tab.traps.vm;
```
lisa-wolfgang, line 3:
```suggestion
  const vm = await addon.tab.traps.vm;
```
You want to `await` the VM -- otherwise users will experience race-condition-related errors.

lisa-wolfgang (changes requested):
Other than a couple of nitpicks, this looks good to me. You'll definitely want Garbo to give this a look-through as well.

addons/self-referencing/userscript.js
WorldLanguages: @lisa-wolfgang Whoops, this is not correct. VM is synchronous in SA.
You won't find any SA API where you have to await a getter. SA APIs only return promises when calling some functions

GrahamSH-LLK:
Is this still under development?

TSFBCE24RhythmHeaveners:
(classical: Put your head on my-) (kids music: -Shoulders, knees and toes, knees and toes,)

Samq64:
#8884 is a similar PR and this one is no longer compatible with the current editor.

TSFBCE24RhythmHeaveners:
f█▌k scratchtools' "Select Self"

Issue #1880, opened 2021-03-15
Simplify `Message by "addon name" from Scratch Addons`


Hans5958:
Can we shorten it into something like `Scratch Addons (Addon name):`?

Also, can we capitalize the phrase after it? Pretty sure `Save changes and leave editor?` is the proper grammar and looks nice than `save changes and leave editor?`

`Message by "disable auto-save" addon from Scratch Addons: save changes and leave editor?`  
`Message by "disable auto-save" addon from Scratch Addons: Save changes and leave editor?`  
`Scratch Addons (Disable auto-save): save changes and leave editor?`  
`Scratch Addons (Disable auto-save): Save changes and leave editor?`  
`SA (Disable auto-save): save changes and leave editor?`  
`SA (Disable auto-save): Save changes and leave editor?`

mxmou:
I think "Scratch Addons (Disable auto-save): Save changes and leave editor?" is the best one.

Hans5958:
cc. @WorldLanguages @apple502j

mxmou:
For consistency, messages from the devtools addon that start with "Griffpatch:" should probably be changed too.

CST1229:
> 
> 
> I think "Scratch Addons (Disable auto-save): Save changes and leave editor?" is the best one.

Maybe "Disable auto-save (Scratch Addons): Save changes and leave editor?"

Hans5958:
> > I think "Scratch Addons (Disable auto-save): Save changes and leave editor?" is the best one.
> 
> Maybe "Disable auto-save (Scratch Addons): Save changes and leave editor?"

IIRC the primary focus is Scratch Addons

Secret-chest:
> > 
> > 
> > I think "Scratch Addons (Disable auto-save): Save changes and leave editor?" is the best one.
> 
> Maybe "Disable auto-save (Scratch Addons): Save changes and leave editor?"

"Disable auto-save from Scratch Addons: save chabges and leave editor?" (no capitalization seems better)

mxmou:
> > > I think "Scratch Addons (Disable auto-save): Save changes and leave editor?" is the best one.
> > 
> > 
> > Maybe "Disable auto-save (Scratch Addons): Save changes and leave editor?"
> 
> "Disable auto-save from Scratch Addons: save changes and leave editor?" (no capitalization seems better)

A good thing about capitalizing the start of the actual message is that it's easy to see where it begins.

Hans5958:
[...]

PR #4639, opened 2022-06-04
confirm-actions: Confirm cancelling comments

### LINKED ISSUE #4637: Confirm canceling comments
Sometimes people accidentally hit the Cancel button after they write a comment, completely erasing it. A new option for the `confirm-actions` addon might help with this.

With a "confirm canceling comments" option, you would be asked if you really wanted to cancel your comment.

An alternative could be to just hide the button instead (carefully, though - it should still appear when replying).

Weredime:
Resolves #4637

### Changes

Adds confirm cancelling comments options

### Tests

firefox developer

### DIFF
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -53,6 +53,7 @@ export default async function (
           .confirm(title, cancelMessage, {
             okButtonLabel: msg("yes"),
             cancelButtonLabel: msg("no"),
+            useEditorClasses: addon.tab.editorMode == "editor",
           })
           .then((confirmed) => {
             if (confirmed) {
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -1,30 +1,33 @@
-export default async function ({ addon, msg }) {
+export default async function ({ addon, console, msg }) {
   let override = false;
+
   document.addEventListener(
     "click",
     (e) => {
       if (override) {
-        override = null;
+        override = false;
         return;
       }
+
       let title = null;
       let cancelMessage = null;
+
       if (
         addon.settings.get("projectsharing") &&
         e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-button")
       ) {
-        title = addon.tab.scratchMessage("project.share.shareButton");
+        title = addon.tab.scratchMessage("project.share.shareButton"); // "Share"
         cancelMessage = msg("share");
       } else if (addon.settings.get("projectunsharing") && e.target.closest(".media-stats a.unshare")) {
-        title = e.target.closest(".media-stats a.unshare").textContent;
+        title = e.target.closest(".media-stats a.unshare").textContent; // "Unshare"
         cancelMessage = msg("unshare");
       } else if (addon.settings.get("followinguser") && e.target.closest("#profile-data .follow-button")) {
         const button = e.target.closest("#profile-data .follow-button");
         if (button.classList.contains("notfollowing")) {
-          title = button.querySelector("span.follow").textContent;
+          title = button.querySelector("span.follow").textContent; // "Follow"
           cancelMessage = msg("follow");
         } else {
-          title = button.querySelector("span.unfollow").textContent;
+          title = button.querySelector("span.unfollow").textContent; // "Unfollow"
           cancelMessage = msg("unfollow");
         }
       } else if (
@@ -46,6 +49,7 @@ export default async function ({ addon, msg }) {
         title = msg("cancelcomment-title");
         cancelMessage = msg("cancelcomment");
       }
+
       if (cancelMessage !== null) {
         e.preventDefault();
         e.stopPropagation();
@@ -53,7 +57,7 @@ export default async function ({ addon, msg }) {
           .confirm(title, cancelMessage, {
             okButtonLabel: msg("yes"),
             cancelButtonLabel: msg("no"),
-            useEditorClasses: addon.tab.editorMode == "editor",
+            useEditorClasses: addon.tab.editorMode === "editor",
           })
           .then((confirmed) => {
             if (confirmed) {
@@ -63,6 +67,6 @@ export default async function ({ addon, msg }) {
           });
       }
     },
-    true
+    { capture: true }
   );
 }
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -41,6 +41,8 @@ export default async function ({ addon, msg }) {
         addon.settings.get("cancelcomment") &&
         e.target.closest("div[data-control='cancel'] > a, .compose-cancel")
       ) {
+        // Do not ask to confirm cancelling empty comments
+        if (e.target.closest("form").querySelector("textarea").value === "") return;
         title = msg("cancelcomment-title");
         cancelMessage = msg("cancelcomment");
       }
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -20,6 +20,8 @@ export default async function ({ addon, console, msg }) {
         cancelMessage = msg("joinstudio");
       } else if (addon.settings.get("closingtopic") && e.target.closest("dd form button")) {
         cancelMessage = msg("closetopic");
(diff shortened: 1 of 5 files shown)


Chameleon-Coder (approved):
Lookin' good!

mxmou:
This has the exact problem @Chiroyce1 mentioned in the issue: having to press "Cancel" to not cancel is very confusing. This addon should be rewritten to use `addon.tab.confirm`.

Norbiros:
I added this feature, but @Weredime was faster, so I gonna make just suggestion how you can add this feature, so:
```js
  let override = null;
  document.addEventListener(
    "click",
    (e) => {
      if (override == true) {
        override = null;
        return;
      }
      let cancelMessage = null;
      if (
        ... // I just didn't paste this big if part
      }
      if (cancelMessage !== null) {
        e.preventDefault();
        e.stopPropagation();
        addon.tab.confirm(msg("confirmaction"), cancelMessage).then((result) => {
          if (!result) return;
          override = true;
          e.target.click();
        });
      }
    },
    true
  );
}
```
You need to remove `ScratchAddons\n:`  from `addons-l10n` and you need to add new entry `confirmaction: "Confirm Action"`
You can also add:
```json
"latestUpdate": {
  "version": "1.27.0",
  "isMajor": false,
  "newSettings": [
    "cancelingmessage"
  ]
}
```
to `addon.json` at the bottom.
I can make own PR with those changes if you want...

mxmou:
> ```js
>         addon.tab.confirm(msg("confirmaction"), cancelMessage).then((result) => {

The title of the modal should be the action that caused it to appear, for example "Share" or "Follow".

Norbiros:
Okay, so what should appear in description?
Maybe something like that:
Title: Confirm Action - Follow
Description: Are you sure ...?

mxmou:
> Okay, so what should appear in description?
> Maybe something like that:
> Title: Confirm Action - Follow
> Description: Are you sure ...?

Follow
Are you sure you want to follow this user?

Weredime:
> PS: Can you please enable actions on your fork, that you can run Prettier?

Done

Weredime:
and everything else is done @mxmou @Norbiros

mxmou (approved):

Norbiros:
This code works correctly with `editor-unshare-button`?

addons/confirm-actions/userscript.js
```diff
@@ -1,31 +1,65 @@
...
+      }
+      let title = null;
       let cancelMessage = null;
       if (
         addon.settings.get("projectsharing") &&
         e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-button")
       ) {
+        title = addon.tab.scratchMessage("project.share.shareButton");
         cancelMessage = msg("share");
       } else if (addon.settings.get("projectunsharing") && e.target.closest(".media-stats a.unshare")) {
```
Norbiros, line 20:
```suggestion
      } else if (addon.settings.get("projectunsharing") && e.target.closest(".media-stats a.unshare, .sa-unshare-button")) {
```

addons/confirm-actions/userscript.js
```diff
@@ -1,31 +1,65 @@
...
+        title = e.target.closest(".media-stats a.unshare").textContent;
```
Norbiros, line 21:
```suggestion
        title = e.target.closest(".media-stats a.unshare, .sa-unshare-button").textContent;
```

addons/confirm-actions/userscript.js
```diff
@@ -1,31 +1,65 @@
...
-        if (!confirm(cancelMessage)) {
-          e.preventDefault();
-          e.stopPropagation();
-        }
+        e.preventDefault();
+        e.stopPropagation();
+        addon.tab
+          .confirm(title, cancelMessage, {
+            okButtonLabel: msg("yes"),
+            cancelButtonLabel: msg("no"),
```
Norbiros, line 55:
```suggestion
            cancelButtonLabel: msg("no"),
            useEditorClasses: addon.tab.editorMode == "editor",
```

Norbiros (changes requested):
Okay, so to fix that problem, you need to remove this:
```js
if (!(await addon.tab.confirm(msg("unshare-button"), msg("unshare-msg"), { useEditorClasses: true }))) return;
```
From line 12 in file `editor-unshare-button/userscript.js`, and add these suggestions. 
*We are moving confirm from editor-unshare to this addon*

addons/confirm-actions/userscript.js
Norbiros: When you are sharing in editor it looks correctly (with editor classes)

mxmou:
@Norbiros Unsharing is a destructive action and should always have a confirmation, even if this addon is not enabled. This addon should be changed ignore the unshare button.

Norbiros:
Cancelling comments, closing forums topics, publishing/unsharing projects from  mystuff and project page too.
Because of that this addon exists. But okay, we can do this in other way...

Norbiros:
So if we don't want to remove confirmation from that addon, you can add:
```js
if (!e.target.closest(".sa-unshare-button")) return;
```
*Scratch changes button element to share first, but this code is still executed, so we need to stop it*
At line 12 in `editor-ushare-addon/userscript.js`.

Weredime:
> So if we don't want to remove confirmation from that addon, you can add:
> 
> ```js
> if (!e.target.closest(".sa-unshare-button")) return;
> ```
> 
> _Scratch changes button element to share first, but this code is still executed, so we need to stop it_ At line 12 in `editor-ushare-addon/userscript.js` (after `addEventListener`).

what
the addon works

WorldLanguages:
Now that `addon.tab.confirm` exists, I like how you removed `Scratch Addons:` from the beginning of strings. In the future, we could change the implementation of `addon.tab.confirm` so that it automatically shows the name of the addon that triggered the alert (in a smaller font). What do you think @mxmou? This is related to #1880

WorldLanguages:
I think cancelling comments and losing all of its content causes frustration for many Scratchers, so the setting will be enabled by default for all Scratch Addons users with `Confirmations` enabled already (the addon itself is not enabled by default).

addons/editor-unshare-button/userscript.js
```diff
@@ -9,6 +9,8 @@ export default async function ({ addon, msg, global, console }) {
     button.classList.add("sa-unshare-button");
     button.querySelector("span").innerText = msg("unshare-button");
     button.addEventListener("click", async (e) => {
+      // Prevent the listener if unshared
+      if (!e.target.closest("sa-unshare-button")) return;
```
WorldLanguages, line 13: The condition is always evaluating to `false`. Did you forget a dot before `sa-` to refer to the class name?
And if you fix that, won't the condition be met every time?
I don't understand why this line was added.

WorldLanguages: @Norbiros Could you explain?

Norbiros: There was bug with 2 confirmations in unshare button in editor (one from editor-unshare and from confirm-actions).

> Did you forget a dot before sa- to refer to the class name?

Yes! I forgot about it...

> And if you fix that, won't the condition be met every time?

Nope, when you are using it without confirmations it always returns true. But look at this example:
User clicks unshare button in editor >> it shows 2 popups >> when you click first one it unshares project (and changes button from sa-unshare, to normal share button) >> but when you click second one it executes .click on the same button as first one, so it shares project
So I added that IF to check if button didn't change.

WorldLanguages: @Norbiros Wow, that's not what I expected would be the reason for adding that one line, and I think it would be better to fix that issue in a different way. Or the code comment above that line needs to explain this in detail.

mxmou: It would be easier to understand if a similar condition was added to the `confirmations` code instead.

Norbiros: Confirm actions always runs first (to fix other bugs), so we can't fix that (using js). Maybe we can look for classes, and add `:not`?

WorldLanguages: editor-unshare-button is buggy even if confirmations is disabled.

WorldLanguages: I'll be fixing this in a separate PR, since the bug isn't really specific to confirm-actions.

PR #4647, opened 2022-06-07
turbowarp-player: send project token

### LINKED ISSUE #4603: [Meta] Changes to projects.scratch.mit.edu
https://github.com/LLK/scratch-www/pull/6773

We should check what addons could be affected by this.

apple502j:
Resolves #4603

See also: https://discord.com/channels/751206349614088204/751207339545198682/981968933819154463

### DIFF
--- a/addons/turbowarp-player/userscript.js
+++ b/addons/turbowarp-player/userscript.js
@@ -27,6 +27,13 @@ export default async function ({ addon, console, msg }) {
   }
 
   button.onclick = async () => {
+    let search = "";
+    if (
+      addon.tab.redux.state?.preview?.projectInfo?.public === false &&
+      addon.tab.redux.state.preview.projectInfo.project_token
+    ) {
+      search = `#?token=${addon.tab.redux.state.preview.projectInfo.project_token}`;
+    }
     if (action === "player") {
       playerToggled = !playerToggled;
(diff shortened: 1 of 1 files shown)


Chiroyce1:
Maybe a privacy warning since users might be paranoid that their Scratch assets token will be processed by TurboWarp? (won't be sent to the server since it's a fragment and not a query param)

WorldLanguages:
I think this is good enough until we can find a better fix. Of course, sending tokens anywhere in the URL is unideal (URL hash will remain in places like browser history) but it's slightly better than using query params.

apple502j:
TurboWarp's postMessage based API is experimental and currently requires sending the whole sb3, which is not ideal

GarboMuffin:
The experimental postMessage API currently only exists in the packager. I'm planning to add something similar to the main site this weekend or earlier. It should be able to accept just a project.json when that happens.

With ?token= I am concerned that the tokens are expected to expire after an unknown (and possibly short) amount of time.

apple502j:
Token expiration should not be a problem; we can fetch ourselves, and I believe TurboWarp does not fetch the token twice unless reloaded (which is unlikely; they'll probably load from saved files or click TurboWarp button again.)

@WorldLanguages Merge this and get this to 1.26.1?

PR #4650, opened 2022-06-08
scratch-messaging: hide read messages on initial load

apple502j:


apple502j:
Self merging, confirmed working by W_L on DM

Issue #4617, opened 2022-05-29
Four settings for theme colours


Secret-chest:
There should be four settings for theme colours, namely Background, Surface Background, Primary (orange), Secondary (blue).

I'd like changing them to match my dark-www.

apple502j:
Is this a suggestion for editor themes?

mxmou:
> Is this a suggestion for editor themes?

Probably for customizing SA's own UI.

Secret-chest:
> > Is this a suggestion for editor themes?
> 
> Probably for customizing SA's own UI.

Yes.

WorldLanguages:
@Secret-chest This is about changing Scratch Addons' UI colors, right? Are these 4 settings you specify already available as CSS variables?

Secret-chest:
@WorldLanguages Yes

WorldLanguages:
So you're suggesting that we add a feature where these 4 colors are customizable and retrieved from chrome.storage on each webpage load, I guess.

Secret-chest:
Exactly that. They're CSS variables already. Also, I added the link colour.

Secret-chest:
This could have a button to adopt `dark-www` colours

Secret-chest:
This could have a button to adopt `dark-www` colours

Chameleon-Coder:
Maybe some presets could be Scratch Addons Orange and Scratch Haddocks Blue?

Secret-chest:
> Maybe some presets could be Scratch Addons Orange and Scratch Haddocks Blue?

Scratch Avocados?

CST1229:
> > Maybe some presets could be Scratch Addons Orange and Scratch Haddocks Blue?
> 
> Scratch Avocados?

Developer Blue?

PR #4656, opened 2022-06-10
Change UI font

### LINKED ISSUE #4614: Replace Sora font with a different one

I hate the sora font

Some way to use `system-ui` (no Montserrat or anything, that isn't a neutral font!)

None

mxmou:
Resolves #4614

### Changes

Replaces the Sora font with Roboto for most text on the settings page and in the popup. The exceptions are:
* The Scratch Addons header in the popup
* The Settings/Permissions/Licenses header
* The title of the More Settings modal
* The title and Enable button on the Permissions page
* The title of each license on the Licenses page

Also replaces the browser's default font for text inputs because using two similar fonts next to each other doesn't look very good.

I wanted to use `system-ui` but it would be difficult to make everything look good with any system font. For example, Segoe UI isn't exactly in the center when centered vertically next to an icon. It could also be difficult to debug issues caused by specific fonts.

Sora is quite wide so I increased the font size to 14px. It looks a bit larger but I think the previous size was too small, even for Sora. The set of weights is also different: I used 500 in some places and 700 in others where 600 was previously used.


Norbiros:
Maybe we can add setting to `more settings` tab, where you can change font between roboto and sora. In future we can add new settings for customization like #4617.

mxmou:
@Norbiros We should choose one good font and use that.

lisa-wolfgang:
I agree that this merits more discussion in the linked issue before we go ahead with this.

Hans5958:
I would argue that it is still possible for going with the generic system-ui group of fonts, but I'm not sure how it is going to appear on the settings,

WorldLanguages:
Could we consider removing `text-transform: uppercase;` from setting names?

Before:
image

After: (font weight was removed too)
image

RedGuy12:
That looks ugly to me, the kerning looks too small and the characters too thin

mxmou:
> Could we consider removing `text-transform: uppercase;` from setting names?

Yes, but I think the weight should stay. If we change that, we need to make sure the way setting names are capitalized is consistent.

penguinmoose:
I don't mind getting rid of `text-transform: uppercase` in the setting names, but I think they need to stand out more. Given that the text is gray, I think it should be bold.

mxmou:
> I don't mind getting rid of `text-transform: uppercase` in the setting names, but I think they need to stand out more. Given that the text is gray, I think it should be bold.

I agree - a thin gray label usually means the feature is disabled.

Alternatively, we could try making them white instead of gray:
image

WorldLanguages:
I agree on making setting names white, instead of gray.

@mxmou Could you send the picture above again, but with `text-transform: uppercase` applied to it?

mxmou:
image

BroJac5246:
> image

I think it looks better without `text-transform: uppercase`.

Secret-chest:
Let's not get offtopic though.

WorldLanguages:
> This is a draft for now because it's a big change and there wasn't a lot of discussion about it yet

I think it would now be worth it to undraft this. The consensus is to change Sora, and no one has made objections about Roboto (yet).

Secret-chest:
I'm in favour of changing setting labels to not be uppercase and be white. Maybe they should use weight `300` for some contrast.

Secret-chest:
I would also like a 16px text size. It's the standard.

WorldLanguages:
> I would also like a 16px text size. It's the standard.

What font size are we currently using? Is this PR proposing to change it?
Also, who defines "the standard"?

lisa-wolfgang:
16px is "standard" in that it is the default text size defined by browsers.

BroJac5246:
> I would also like a 16px text size. It's the standard.

Just because it's the default in most browsers doesn't mean we should switch to it.

Samq64:
The heart and star icons in the popup are a little off-centered.
image

WorldLanguages:
@mxmou Sure, we could get this merged for v1.41

WorldLanguages (approved):

libraries/license-info.json
```diff
@@ -56,5 +56,8 @@
     "license": "ISC",
     "year": "2016,",
     "fullname": "Jake Archibald <jaffathecake@gmail.com>"
+  },
+  "Roboto": {
+    "license": "Apache-2.0"
```
Samq64, line 61: Should the extra license info be filled out?

Samq64 (approved):

libraries/license-info.json
mxmou: I'll add it. I thought it wasn't necessary because a few other libraries use the same license and they don't include the info, but there's indeed a placeholder near the end of the license text:
image

WorldLanguages:
No feedback from end-users whatsoever after changing the font.

DNin01:
> No feedback from end-users whatsoever after changing the font.

It's possible few will even notice!

For me, I honestly like either font.

Secret-chest:
> No feedback from end-users whatsoever after changing the font.

Here's one from me: I like it

PR #4660, opened 2022-06-11
`move-to-top-bottom`: add `costumeEditor` tag

lisa-wolfgang:
### Changes

Self-explanatory.

### Reason for changes

A user browsing the settings page for costume editor addons (via the "Costume Editor Features" subcategory) should see this addon in the category listing.

### Tests

No, but I would be very surprised if this actually broke anything.

### DIFF
--- a/addons/move-to-top-bottom/addon.json
+++ b/addons/move-to-top-bottom/addon.json
@@ -1,7 +1,7 @@
 {
   "name": "Move costume to top or bottom",
   "description": "Adds a right click context menu item to move a costume or a sound to the top or the bottom of the list. Previously part of the developer tools.",
-  "tags": ["editor", "featured"],
+  "tags": ["editor", "costumeEditor", "featured"],
   "userscripts": [
     {
       "url": "userscript.js",


lisa-wolfgang:
The only reasons I'm waiting on merge are lack of testing and a discrepancy between the docs and actual practice.

The docs say that the `costumeEditor` tag is for addons that primarily affect the costume editor. However, in the extension, the developer tools addon is marked with the tag, even though most of its features are for the code editor. Similarly, this addon affects both the costume and sound tabs.

I don't agree with what the docs say the tag should be restricted to, since the user would expect to see **all** addons that affect the costume editor, not just the ones that **only** affect the costume editor. I wanted to get some more opinions about it first, though.

If we decide to merge this PR, the other addon tags in the docs with similar wording should be changed as well.

(Also, I suppose there could be an opinion that this technically doesn't affect the costume editor but the asset ribbon in the costume/sound tabs. However, my opinion is that this addon is part of the general costume editing experience and is therefore deserving of the tag.)

WorldLanguages:
> I don't agree with what the docs say the tag should be restricted to, since the user would expect to see **all** addons that affect the costume editor, not just the ones that **only** affect the costume editor. I wanted to get some more opinions about it first, though.

This happens because currently, an addon can only be in a single subcategory

> Also, I suppose there could be an opinion that this technically doesn't affect the costume editor but the asset ribbon in the costume/sound tabs

Thinking that way is overkill, I think the average user would expect this addon to be listed under costume editor addons

PR #4671, opened 2022-06-13
Handle settings change effectively changing addon state

### LINKED ISSUE #4665: Enabling a setting that adds a userstyle does not dynamically enable the addon
### Description

CSS variables for addon settings are only set if any of the addon's userstyles are active on the page. If those userstyles depend on a setting, enabling that setting doesn't set the CSS variables. This affects `editor-theme3`.

### Reproduction steps

1. Enable "customizable block colors" and disable "change the colors of blocks in the forums".
2. Go to https://scratch.mit.edu/discuss/topic/574329/
3. Enable "change the colors of blocks in the forums". The blocks will not have correct colors because the CSS variables are missing.

### Environment

- Browser: Edge 102
- Scratch Addons version: tested on 1.26.0 and current master
- Operating system: Windows

apple502j:
Resolves #4665

Addon state (whether addon was enabled or disabled) was previously only changed by enabling or disabling an addon directly. #4289 changed this by introducing partial dynamic enable and disable. However, there existed another way for userstyle addons to get enabled, involving a setting condition. Due to implementation quirks this actually added the style previously as well, but other effects normally achieved by (dynamically-)enabling addons weren't applied; this included addonsWithUserstyles array and custom CSS variables. This also applied to disabling.

The change consists of four parts:

- Warn if updateUserstylesOnSettingsChange was used without either dynamicEnable or dynamicDisable, and if a userstyle had a settings condition. This causes the described bug.
- If dynamicEnable is available and changing the setting causes a new userstyle to load, perform the steps done for normal dynamic enable, except for userscript related steps as they are inapplicable due to conditional userscript being unsupported.
- If dynamicDisable is available and changing the setting causes an existing userstyle to be disabled, perform the steps done for normal dynamic disable, except for userscript related steps.
- Small changes to background to pass necessary information.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -91,7 +91,7 @@ scratchAddons.localEvents.addEventListener("addonDynamicDisable", ({ detail }) =
   );
 });
 scratchAddons.localEvents.addEventListener("updateUserstylesSettingsChange", ({ detail }) => {
-  const { addonId, manifest } = detail;
+  const { addonId, manifest, newSettings } = detail;
   chrome.tabs.query({}, (tabs) =>
     tabs.forEach((tab) => {
       if (tab.url) {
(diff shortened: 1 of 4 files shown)


apple502j:
Self-merging with approval from World_Languages

Issue #4281, opened 2022-02-18
Debugger breakpoints don't work if there is no yield between them


Tacodude7729:
![image](https://user-images.githubusercontent.com/27910867/154765856-1ba77485-5b3a-423a-9996-29f21e219bd1.png)

If there is no yield between two breakpoints, as in the script above, the ones after the first one will not work. When the above script is run, a breakpoints is hit, but when play is pressed 1, 2 and 3 are all logged, than the script is paused.

The breakpoints after the logging of 1 and 2 should work.

Issue #4210, opened 2022-02-05
Debugger single stepping changes behaviour of broadcasts


Tacodude7729:
![image](https://user-images.githubusercontent.com/27910867/152661119-7215272c-5b8e-4089-aad1-47126239b0e0.png)
_(See https://scratch.mit.edu/projects/639623384/ for code)_ 

If you just unpause after the breakpoint, the following logs are produced:

![image](https://user-images.githubusercontent.com/27910867/152661138-6161519c-3164-4294-aaca-3e4c067c9ff3.png)

If you single step the rest of the way through the code after the breakpoint, the following logs are produced:

![image](https://user-images.githubusercontent.com/27910867/152661154-2f4f7672-1dc1-4a24-b787-abdeb0a84ce5.png)

I believe this issue comes from single stepping not handling the thread with the second broadcast restarting correctly.

Samq64:
New addon?

apple502j:
@TheColaber bruh

PR #4420, opened 2022-03-17
Optimize graphs - REUPLOAD

Norbiros:
Resolves #4180

### Changes

New settings that hides animation, and fill function.

### Tests

Yep.

**I removed previous branch, so I couldn't fixed some bugs, but now everything works!**

PR #4674, opened 2022-06-14
Fix some bugs in the pause and debugger shared module

GarboMuffin:
- Fix performance regression where pausing a project will make Scratch's sequencer do 25ms of unnecessary work every frame.
 - Fix some bugs where breakpoints and single stepping could affect script execution order
 - Fix error when you pause project with no threads -> start new thread -> try to single step

I've tested this on quite a few projects and haven't found any regressions so far (but have found many bug fixes), but with how complex the debugger is I wouldn't be surprised if there's something. I have verified that #4210 and #4281 are still fixed.

### DIFF
--- a/addons/debugger/module.js
+++ b/addons/debugger/module.js
@@ -25,23 +25,26 @@ const pauseThread = (thread) => {
     return;
   }
 
-  const pausedState = {
+  const pauseState = {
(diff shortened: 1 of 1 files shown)


Norbiros:
We can also merge this PR #4420 (adds setting to optimize graphs on slower computers).

PR #4679, opened 2022-06-16
Don't group addons based on tag when sorting

### LINKED ISSUE #3012: Addons are sorted weirdly on the settings page.
### Description

On the settings page, addons aren't sorted the way one would expect.

### Expected behaviour

Within each group, addons should either
- be sorted alphabetically;
- be split based on the category (website, editor, popup) if that's clearly indicated.

### Actual behaviour

In each group, editor addons are shown first, then website addons and finally popups, but that isn't obvious to the user.

mxmou:
Resolves #3012

### Changes

Sorts addons on the settings page alphabetically without checking their tags (except "beta" and "danger").

### Reason for changes

To make it easier to find a specific addon in the list without using the search feature.

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -580,7 +580,8 @@ let fuse;
       group.addonIds = group.addonIds
         .map((id) => vue.manifestsById[id])
         .sort((manifestA, manifestB) => {
-          for (const tag of [...(group.customOrder || order), null]) { // null means "any tag"
+          for (const tag of [...(group.customOrder || order), null]) {
+            // null means "any tag"
             const val = checkTag(tag, manifestA, manifestB);
             if (val !== null) return val;
           }
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -557,22 +557,30 @@ let fuse;
     fuse = new Fuse(cleanManifests, fuseOptions);
 
     const checkTag = (tagOrTags, manifestA, manifestB) => {
-      const tags = Array.isArray(tagOrTags) ? tagOrTags : [tagOrTags];
-      const aHasTag = tags.some((tag) => manifestA.tags.includes(tag));
-      const bHasTag = tags.some((tag) => manifestB.tags.includes(tag));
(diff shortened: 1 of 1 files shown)


WorldLanguages:
As I said at #3012, the current order is intentional.
Rather than putting website and editor addons on the same bag, I think it makes more sense to include a separator that tells the user that the addons inside the category they're currently on are grouped by website/editor/popup.

Honestly, finding a good way to group addons both by website/editor and recommended/featured/others at the same time has been a challenge, and I'm no UX expert. I'm not convinced what the best way is, but I'm pretty sure it makes sense to group by both. Users could always filter by website/editor, but I'm not sure if that means we should stop grouping by website/editor on "all" mode.

mxmou:
> Rather than putting website and editor addons on the same bag, I think it makes more sense to include a separator that tells the user that the addons inside the category they're currently on are grouped by website/editor/popup.

Something like this?
image

WorldLanguages:
@mxmou Yes, something like that

WorldLanguages:
A simple separator like that is easier said than done with our vue1 spaghetti but yeah...

Secret-chest:
> > Rather than putting website and editor addons on the same bag, I think it makes more sense to include a separator that tells the user that the addons inside the category they're currently on are grouped by website/editor/popup.
> 
> Something like this?
> image

That makes no sense... why are dangerous addons listed separately

mxmou:
> That makes no sense... why are dangerous addons listed separately

#2407

mxmou:
Closing, will try to implement W_L's suggestion instead.

WorldLanguages:
> That makes no sense... why are dangerous addons listed separately

They are intentionally moved to the top of the "enabled" category to remind the user that those are enabled.

mxmou:
> > > That makes no sense... why are dangerous addons listed separately
> > 
> > 
> > They are intentionally moved to the top of the "enabled" category to remind the user that those are enabled.
> 
> Why tho?

Because they're dangerous

PR #4680, opened 2022-06-16
New "info" type info for less important notices and warnings

### LINKED ISSUE #3015: Less notices and warnings

Notices are meant to be important instructions, but some addons use them for things that should probably be part of the description instead. Others use warnings when notices would be more appropriate.

Move some notices to addon descriptions:
- non-draggable sprites
- developer tools
- pause button
- the second notice ("this feature works ...") in sprite folders

Change the custom block shape warning to a notice.

mxmou:
Resolves #3015

### Changes

Moves some notices to the addon description and changes some warnings to notices.


addons/old-studio-layout/addon.json
```diff
@@ -1,16 +1,11 @@
 {
   "name": "Old studio layout",
-  "description": "Reverts the layout of the studio page back to the one used before the studio update or the one previously available as part of the Scratch 2.0 \u2192 3.0 addon.",
+  "description": "Reverts the layout of the studio page back to the one used before the studio update or the one previously available as part of the Scratch 2.0 \u2192 3.0 addon. The size of most elements is not changed. If you want to see more projects and curators at once, use the \"more items per row\" addon.",
```
WorldLanguages, line 3: I think this one should stay as a notice. We don't have "see this other addon" functionality yet, but if we do at some point, the list of related addons won't be listed on the description, but probably somewhere more visible. A notice is closer to that than simply appending to the description.

addons/editor-devtools/addon.json
```diff
@@ -1,13 +1,6 @@
...
   "name": "Developer tools",
-  "description": "Adds a find box and new menu options to the editor: copy/paste blocks, better clean up, go to custom block definition, and more!",
-  "info": [
-    {
-      "type": "notice",
-      "text": "Press \"Ctrl+F\" to focus the find box, \"Ctrl+Space\" to bring up the \"add block by name\" popup, \"Ctrl+Left/Right\" to navigate find history.",
-      "id": "shortcuts"
-    }
-  ],
+  "description": "Adds a find box and new menu options to the editor: copy/paste blocks, better clean up, go to custom block definition, and more! Press \"Ctrl+F\" to focus the find box, \"Ctrl+Space\" to bring up the \"add block by name\" popup, \"Ctrl+Left/Right\" to navigate find history.",
```
WorldLanguages, line 3: If we change this, the quotes between hotkeys (e.g. `"Ctrl+F"` instead of `Ctrl+F`) can be removed to stay consistent with other addon's descriptions.

addons/old-studio-layout/addon.json
WorldLanguages: Good point - maybe we should add an "info" type message that doesn't stand as much? The color shouldn't show off too much, and the info text should be the same size as the description text. What do you think?

mxmou: Like this?
image
image

lisa-wolfgang: Perhaps the border in dark mode should be a little brighter to be consistent with the dark mode notice border and the light mode info message?

I agree that this is a good transient solution, though.

mxmou:
Info messages now look like this:
image

WorldLanguages:
@mxmou Excellent. I think we can now use info messages for stuff like `On macOS, use the Cmd key instead of the Ctrl key`.

WorldLanguages:
Maybe we should rename the type from `info` to `note`, to avoid having an info of type info.

image

mxmou:
> Maybe we should rename the type from `info` to `note`, to avoid having an info of type info.

That might be too similar to `notice`. Could info be the default type to make writing info twice unnecessary?

WorldLanguages:
> Could info be the default type to make writing info twice unnecessary?

Good idea, let's do that

PR #4684, opened 2022-06-17
Hide rows

### LINKED ISSUE #1758: Customize front page
**Describe the new addon**

This would allow you to hide certain sections of the front page, such as What's Happening or What the Community is Loving.
This could also add new rows like "Recently Favorited Projects" or "Featured Projects from scratchstats.com".

Also, it would be nice if the "Projects by/Loved by Scratchers I'm Following" rows were fixed or just completely replaced with new versions.

Platformer444:
Resolves a part of #1758

### Changes
It helps to hide the rows of the Scratch Front Page.

### Tests
Everything Works Fine.


Norbiros:
1. It has bug - when user is logged out you have just 6 rows.
2. I think we should add way to change order of rows (this idea had issue, but I can't find it). I created something like that, but it had bug that shows Scratch error when user logs in. I can make PR with that feature...

addons/hide-rows/userscript.js
```diff
@@ -0,0 +1,52 @@
+export default async function ({ addon, global, console }) {
+  // Updates the page if any settings are changed
+  addon.settings.addEventListener("change", Change());
+
+  // Main Function
+  function Change(argument) {
+    // What's Happening
+    if (addon.settings.get('whs_ha') === true) {
```
Norbiros, line 8:
```suggestion
    if (addon.settings.get('whs_ha')) {
```
You don't need to check if `true == true`.

addons/hide-rows/addon.json
```diff
@@ -0,0 +1,69 @@
...
+    {
+      "name" : "What the Community is Loving",
+      "id" : "com_lov",
+      "type" : "boolean",
+      "default" : false
+    }
+  ],
+  "versionAdded" : "1.26.1"
+  "tags": ["community"],
+  "enabledByDefault": false
```
Norbiros, line 68:
```suggestion
  "enabledByDefault": false,
  "dynamicDisable": true,
  "dynamicEnable": true
```

addons/hide-rows/userscript.js
```diff
@@ -0,0 +1,52 @@
+export default async function ({ addon, global, console }) {
+  // Updates the page if any settings are changed
+  addon.settings.addEventListener("change", Change());
+
```
Norbiros, line 4:
```suggestion

  addon.self.addEventListener("disabled", () =>
    document.querySelectorAll(".box").forEach(el => el.style.display = "inline-block")
  );
```

Norbiros (changes requested):

Norbiros:
BTW, if you want to fix Prettier error, enable actions on your fork.

addons/hide-rows/userscript.js
Norbiros: So now you added automatic disable, but you didn't implement it.

PR #4696, opened 2022-06-23
New Addon: `player-thumb`

### LINKED ISSUE #589: Persistent thumbnail
**Describe the new addon**

Shows the project's thumbnail in the player area of the project page until the project is started, helping with the inconsistency of Scratch thumbnails showing.

### LINKED ISSUE #6461: Replace project loading screen with the projects thumbnail
### Why this would be helpful

simple and small addon that makes the loading screen a little more interesting.
the thumbnail is usually rarely seen, only when first seeing the project on the user page and front page , this lets the user see the thumbnail a little more

### How this addon works

this addon would move the current loading screen into the corner and fill the extra space with the projects thumbnail

### Possible alternatives

_No response_

### Additional context

_No response_

Samq64:
Resolves #589
Resolves #6461

### Changes

Always shows the thumbnail under the green flag overlay. There's also now an option to use the thumbnail as the loading screen background.

### Reason for changes

It fixes some inconsistencies and is better than the gray rectangle that sometimes appears. Some users would also like to have something more interesting as their loading screen.

### Tests

Tested on Chromium 115.


addons/persistent-thumb/addon.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Always show thumbnail",
+  "description": "Always shows thumbnails under the big green flag before projects are started.",
```
lisa-wolfgang, line 3:
```suggestion
  "description": "Fixes a bug in the Scratch project player where thumbnails don't always load before a project is started.",
```

mxmou: It isn't a bug. What you see before the project starts isn't a thumbnail, it's the state of the project when it was saved. That's why it's blank for pen projects - Scratch 3 doesn't save the pen layer.

Norbiros: We can remove part about bug, but I think this part `Scratch project player where thumbnails don't always load ` should stay.

Norbiros: When I was testing this addon I didnt undetstood how it works, until mxmou told about pen extension,  so maybe we can add info about it? Maybe
`This addon always shows the project thumbnail before a project is started (even if project uses pen extension).`

apple502j:
Could you rename the addon ID? persistent-thumb is already used im different context.

addons/player-thumb/addon.json
```diff
@@ -0,0 +1,28 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"],
+      "runAtComplete": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.27.0",
```
apple502j, line 25:
```suggestion
  "versionAdded": "1.28.0",
```

addons/player-thumb/userscript.js
```diff
@@ -0,0 +1,16 @@
+export default async function ({ addon, global, console }) {
+  const el = await addon.tab.waitForElement("div.stage_green-flag-overlay-wrapper_2hUi_.box_box_2jjDp", {
```
apple502j, line 2:
```suggestion
  const el = await addon.tab.waitForElement('div[class*="stage_green-flag-overlay-wrapper_"]', {
```

apple502j (approved):

WorldLanguages (dismissed):
This addon is hard to explain, so I'm blocking merge until I can take a deeper look at the addon name and description.

CST1229:
> This addon is hard to explain, so I'm blocking merge until I can take a deeper look at the addon name and description.

I suggest something like:
"Show project thumbnail under green flag overlay"
"Shows the project's actual thumbnail instead of its initial state before starting the project."

Samq64:
Sounds a bit long, but I guess it'll work.

addons/player-thumb/addon.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "url": "userscript.js",
+      "matches": ["projects"],
+      "runAtComplete": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.28.0",
+  "tags": ["community", "theme"],
```
lisa-wolfgang, line 26:
```suggestion
  "tags": ["community", "projectPage"],
```
In my opinion, this is more of a behavior change than a theme. There's only one thing that's having its appearance changed, and the content is dynamic.

addons/player-thumb/addon.json
```diff
@@ -0,0 +1,28 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"],
+      "runAtComplete": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.28.0",
```
lisa-wolfgang, line 25: Reminder to change this before merge

lisa-wolfgang (changes requested):

addons/player-thumb/userscript.js
```diff
@@ -0,0 +1,16 @@
+export default async function ({ addon, global, console }) {
+  const el = await addon.tab.waitForElement('div[class*="stage_green-flag-overlay-wrapper_"]', {
+    markAsSeen: true,
+  });
+
+  const projectId = window.location.pathname.split("/")[2];
+  const thumb = `https://uploads.scratch.mit.edu/get_image/project/${projectId}_480x360.png`;
```
lisa-wolfgang, line 7: I think compatibility with the "high resolution thumbnails" addon should be considered.

addons/player-thumb/userstyle.css
```diff
@@ -0,0 +1,6 @@
+[class*="stage_green-flag-overlay-wrapper"] {
+  background-repeat: no-repeat;
+  background-position: center center;
+  background-size: cover;
+  background-color: white;
+}
```
lisa-wolfgang, line 6: I'm not sure if it's within the scope of the addon to remove the tint from the green flag overlay.

addons/player-thumb/userscript.js
Samq64: I don't think thumbnails are stored at higher resolutions that, but I could be wrong.

BroJac5246: Does http://uploads.scratch.mit.edu/get_image/project/1_960x720.png (for example) work if the image is a higher resolution then 480x360?

Samq64: Yes it does.

Samq64: I'm always using the highest resolution regardless if high resolution thumbnails is enabled or not now.

mxmou:
> Everything works except the thumbnail isn't darkened anymore.

That could be fixed by creating an `img` element and inserting it into the DOM between the stage and the green flag overlay. Making it disappear when the project is started shouldn't be too difficult - I think you can listen for the Redux action `scratch-gui/vm-status/SET_STARTED_STATE`. I think this change would also make the addon compatible with `hi-res-thumbnails`.

Samq64:
@mxmou I have a z-index issue with the img element. Could you help me with it?

Samq64:
Thanks @mxmou!

addons/player-thumb/userstyle.css
```diff
@@ -0,0 +1,10 @@
+.sa-project-thumb {
+  position: absolute;
+  top: 1px;
+  bottom: 1px;
+  left: 1px;
+  right: 1px;
+  width: calc(100% - 1px);
+  height: calc(100% - 1px);
```
mxmou, line 8: Subtracting 1 pixel isn't enough - the thumbnail still covers the bottom and right boder.
```suggestion
  width: calc(100% - 2px);
  height: calc(100% - 2px);
```

mxmou (changes requested):

mxmou (approved):

WorldLanguages (changes requested):
Bug:
Go to https://scratch.mit.edu/create (while logged in)
Broken image with URL `https://uploads.scratch.mit.edu/get_image/project/editor_480x360.png` will be added above the stage, inside the editor.

Samq64:
> Bug: Go to https://scratch.mit.edu/create (while logged in) Broken image with URL `https://uploads.scratch.mit.edu/get_image/project/editor_480x360.png` will be added above the stage, inside the editor.

I can't reproduce on Chromium.

WorldLanguages:
Related discussion: #6461

addons/player-thumb/addon.json
```diff
@@ -0,0 +1,42 @@
...
+    }
+  ],
+  "settings": [
+    {
+      "name": "Show while loading projects",
+      "id": "loading",
+      "type": "boolean",
+      "default": true
+    }
+  ],
```
mxmou, line 37: The description should mention this option, especially if it's enabled by default.

addons/player-thumb/userscript.js
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  if (addon.settings.get("loading")) {
+    const LoaderBackground = stageWrapper.querySelector('[class*="loader_background_"]');
+    stageWrapper.insertBefore(thumb, LoaderBackground);
+  }
+
+  addon.tab.redux.initialize();
+  addon.tab.redux.addEventListener("statechanged", (e) => {
+    console.log(e.detail.action.type);
```
mxmou, line 22:
```suggestion
```

addons/player-thumb/userscript.js
```diff
@@ -0,0 +1,45 @@
...
+      moveThumb();
+      // TODO: Remove listener
+    }
+    if (e.detail.action.type == "scratch-gui/vm-status/SET_STARTED_STATE") {
+      document.getElementById("sa-project-thumb").remove();
+    }
+  });
+
+  function moveThumb() {
+    console.log("moveThumb activated");
```
mxmou, line 33:
```suggestion
```

addons/player-thumb/userstyle.css
```diff
@@ -0,0 +1,10 @@
+.sa-project-thumb {
+  position: absolute;
+  top: 1px;
+  bottom: 1px;
+  left: 1px;
+  right: 1px;
+  width: calc(100% - 2px);
+  height: calc(100% - 2px);
+  border-radius: calc(0.5rem - 1px);
+}
```
mxmou, line 10: The thumbnail is stretched while the loading screen is visible and it changes size when the project loads. This could be fixed by adding a "loading" class and some CSS:
```suggestion
}

.sa-project-thumb-loading {
  top: calc(2.75rem + 1px);
  height: calc(100% - 2.75rem - 2px);
}
```

mxmou (changes requested):

Joeclinton1:
I wonder if once this addon is made you could take it a step further, by having the ability to put custom text in the loading screen, which could be stored in a comment. Then you could have funny/useful messages just for scratch addon users.

Samq64:
Is there a way to remove the block animation transparency? Other than that I think this is ready.

TheColaber (changes requested):
seems like it doesnt work well in full screen 
image

TheColaber:
same issue occurs... but once the project loads it fixes itself

addons/player-thumb/loading.css
```diff
@@ -0,0 +1,9 @@
+body:not(.sa-body-editor) div[class*="loader_background_"] {
+  background-color: rgba(0, 0, 0, 0.25);
```
mxmou, line 2:
```suggestion
  background-color: rgba(0, 0, 0, 0.25);
  border-radius: calc(0.5rem - 1px);
```

addons/player-thumb/loading.css
mxmou, line 1: > Is there a way to remove the block animation transparency?

I think the simplest way is to use a filter:
```css
body:not(.sa-body-editor) [class*="loader_block-animation_"] {
  filter: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg'><filter id='opaque'><feComponentTransfer><feFuncA type='linear' slope='3.5'></feFuncA></feComponentTransfer></filter></svg>#opaque");
}
```

mxmou: It could be in a SVG file in the addon directory instead of a data URI.

addons/player-thumb/addon.json
```diff
@@ -0,0 +1,42 @@
...
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.37.0",
```
mxmou, line 40: Needs to be updated before merging

mxmou (approved):

Samq64:
Is the loading setting necessary? The toggle doesn't currently do anything outside the project page and removing it would make the addon easier to explain.

Samq64:
I added fullscreen and editor support locally and removed the setting. Should I push the changes or wait for this to be merged first?

mxmou:
I think it makes sense to remove the setting - as you said, it would be easier to explain the purpose of the addon. Showing the thumbnail while loading is a more visible change than showing it under the green flag overlay, so it should probably be the main feature mentioned in the name and description.

mxmou (approved):

addons/player-thumb/addon.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "Show project thumbnail while loading",
+  "description": "Shows the project's thumbnail instead of its initial state before starting the project and while it loads.",
```
mxmou, line 3: The description could probably be better - it can currently be read as
`Shows the project's thumbnail instead of its initial state (before starting the project and while it loads)`
instead of
`Shows the project's thumbnail (instead of its initial state before starting the project) and (while it loads)`

WorldLanguages:
```suggestion
  "description": "Shows the project's thumbnail before starting the project and while it loads.",
```

WorldLanguages:
```suggestion
  "description": "Shows the project's thumbnail as a placeholder before the project is started and while it loads.",
```

addons/player-thumb/addon.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "Show project thumbnail while loading",
+  "description": "Shows the project's thumbnail as a placeholder before the project is started and while it loads.",
```
WorldLanguages, line 3: Leaving this comment open for a few days...

PR #4697, opened 2022-06-23
Fix a performance regression in the pause and debugger shared module

GarboMuffin:
Further testing of https://github.com/ScratchAddons/ScratchAddons/pull/4674 reveals that it introduced a performance regression in some projects where some internal Scratch functions will be permanently slower (until reload) after pausing once, seemingly related to JS JITs not being able to optimize the getter/setter on thread.status very well.

https://scratch.mit.edu/users/GarboMuffin/#comments-212097554

### DIFF
--- a/addons/debugger/module.js
+++ b/addons/debugger/module.js
@@ -224,6 +224,11 @@ const singleStepThread = (thread) => {
       enumerable: true,
       writable: true,
     });
+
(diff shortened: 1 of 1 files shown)


Weredime:
Double GarboMuffin in https://github.com/ScratchAddons/ScratchAddons/commit/7d9c2cfb9b42f0835771da2064bdda0fcb384ba8

PR #4704, opened 2022-06-25
Fix custom block arg switching bug

pufferfish101007:
Fixes https://discord.com/channels/806602307750985799/938807989874360330/990211808981823528

### Changes

Fixes a bug where any argument could be switched, even if it didn't belong to the root custom block...
eg https://media.discordapp.net/attachments/990211808981823528/990212037680447518/unknown.png

### Reason for changes

It's unintended behaviour IMO; I didn't intend for this when originally implementing it.

### Tests

Tested on chromium 101

### DIFF
--- a/addons/block-switching/userscript.js
+++ b/addons/block-switching/userscript.js
@@ -733,6 +733,7 @@ export default async function ({ addon, global, console, msg }) {
           !block.isShadow()
         ) {
           const customBlocks = getCustomBlocks();
+          const currentValue = block.getFieldValue("VALUE");
           if (customArgsMode === "all") {
             switch (type) {
               case "argument_reporter_string_number":
@@ -758,8 +759,8 @@ export default async function ({ addon, global, console, msg }) {
                 switches = customBlockObj.boolArgs;
                 break;
             }
+            if (!switches.includes(currentValue)) return items;
           }
-          const currentValue = block.getFieldValue("VALUE");
           switches = uniques(switches).map((i) => ({
             isNoop: i === currentValue,
             fieldValue: i,


GarboMuffin:
I don't think this is a bug

If I copy a block from this script

image

To another script like this

image

I don't see why I shouldn't be able to convert `i` to `index` with block switching. If anything, if the custom block doesn't have an argument with the reporter's name, wouldn't it be more likely that I would want to switch it to something that exists?

pufferfish101007:
Hmm I suppose that does make sense - I hadn't considered the copy/paste scenario.
The person who reported the bug on discord originally suggested this as a option but I said that I considered it a bug - maybe another setting? Or maybe just leave it?

RedGuy12:
```js  
{
  "name": "Shown custom block arguments options",
  "id": "customargsmode",
  "type": "select",
  "default": "defOnly",
  "potentialValues": [
    {
      "id": "defOnly",
      "name": "Arguments in own custom block"
    },
    {
      "id": "all",
      "name": "Arguments in all custom blocks in sprite"
    }
}
``` this exists right?

GarboMuffin:
@RedGuy12 This PR changes how the "defOnly" mode works. In the images in my previous comment, right clicking on "i" in "block name" with this PR will show no switches (because the custom block doesn't have an argument called "i") while on master it shows an option to switch it to "index".

I personally think that what master does is the correct behavior. Having an argument reporter that doesn't exist in the current custom block is usually a mistake and you can always just ignore the options if you don't want to use them.

GarboMuffin:
I read the bug report on Discord and it sounds like the bug is that if you dragged a new "i" argument reporter directly into "block name" and right clicked on the block, it would show switches for "block name" instead of "aaaaa". I think that's the correct behavior.

WorldLanguages:
This appears to be intentional, so @pufferfish101007 please create an issue to discuss whether the current behavior should be changed. This PR may be reopened later after discussion.

PR #4716, opened 2022-06-27
New addon: Forum post save

### LINKED ISSUE #1989: Automatically Save Draft to your Storage/Cloud (forums)
I think there should be an addon to automatically save a draft to your computer storage or the cloud if extensions can do that. The setting could allow you to change the interval with default as 15.

Weredime:
Resolves #1989

### Changes
Adds a new addon: Forum post save.
It saves the draft post at an interval.

### Tests
Tested on Firefox


Samq64:
I think saving every 250ms is a little excessive. I think it should probably be at least 1-3 seconds.

DNin01:
> I think saving every 250ms is a little excessive. I think it should probably be at least 1-3 seconds.

Maybe even slower, like every 10 seconds, or after a moment of not typing, and if the user is leaving the page.

Might also want to limit how many of these are stored - from looking it up, I believe local storage can store up to 5 MB unless the user gives the extension permission to store as much as it wants. Maybe this addon could keep only the 5 most recent draft posts?

WorldLanguages:
> I believe local storage can store up to 5 MB unless the user gives the extension permission to store as much as it wants

Note that the code within this PR uses localStorage in non-extension context, so we get as much storage as Scratch itself gets. "unlimitedStorage" wouldn't change this

addons/forum-post-save/userscript.js
```diff
@@ -0,0 +1,59 @@
...
+  if (madeAnyChanges) {
+    localStorage.setItem("sa-forum-post-save", JSON.stringify(cache)); // dont use updateCache here since it won't remove deleted items
+  }
+  if (typeof cache[topicId]?.cache === "string") {
+    box.value = cache[topicId].cache;
+  }
+
+  let lastSaved = 0;
+  box.addEventListener("input", (e) => {
+    if (!addon.self.disabled && Date.now() - lastSaved >= 250) {
```
DNin01, line 26:
```suggestion
    if (!addon.self.disabled && Date.now() - lastSaved >= 5000) {
```
As Samq64 said, I don't think that we really need to save it four times per second. Even saving every five seconds, there's not much to lose if you close the page.

The only problem with that is that, with this if-then condition, the last part of someone's message won't get saved if they wrote it in the middle of the cooldown time. Imagine the following scenario:
- User writes part of their post
- Draft saves
- User spends another fraction of a second writing
- (Notice that the last part of the post doesn't save because during that last part of writing, the cooldown time wasn't over, therefore if-then condition was false, meaning it didn't trigger the save)

<sub>(Hope that makes sense.)</sub>
Might need to do some sort of "wait until" thing. (Also, don't expect to replicate the above because... 👇)

Also, even with this code, it's somehow actually saving _every single keystroke_ as if the cooldown wasn't even there (I added a `console.log` action to test it), when it's expected to only save if it's been 250 ms, as demonstrated by the code, so it seems like the cooldown doesn't work for some reason. That issue is something to work on too.

addons/forum-post-save/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Save draft posts",
+  "description": "Saves posts that you haven't sent yet at an interval.",
```
DNin01, line 3: Here's my description:
```suggestion
  "description": "Saves forum posts that you haven't submitted.",
```

DNin01 (changes requested):

DNin01:
I wonder how it would work if there was a "save draft" button instead of autosaving. 🤔

Secret-chest:
> Might also want to limit how many of these are stored - from looking it up, I believe local storage can store up to 5 MB unless the user gives the extension permission to store as much as it wants. Maybe this addon could keep only the 5 most recent draft posts?

1 byte = 1 character. No way a post could take 1 MB, that's 1 million chars.

addons/forum-post-save/userscript.js
mxmou: Apparently `lastSaved` is initialized once (to 0) and then never updated. That causes the second problem.

Samq64: I don't think it should even check for keystrokes, just keep looping and add a timeout to the function.

Samq64:
I'll test later since the forums are down right now.

OneShot-Niko:
> I'll test later since the forums are down right now.

As side note - According to the ST's twitter, it should [reopen about early next week](https://twitter.com/scratch/status/1750611762540347555).

Samq64:
Right now the post is only saved when a key is pressed. Would it make sense to save after the user stops typing for a few seconds instead?

DNin01:
> Right now the post is only saved when a key is pressed. Would it make sense to save after the user stops typing for a few seconds instead?

I was going to say that. We should definitely fix it.

I think what you said is a pretty good solution. Maybe save after 2 seconds of inactivity (and maybe also when the user unfocuses the reply box)? That should be enough.

Samq64 (approved):

addons/forum-post-save/userscript.js
```diff
@@ -0,0 +1,59 @@
...
+    timeout = setTimeout(updateCache, 2000, topicId);
+  });
+
+  box.addEventListener("blur", () => {
+    updateCache(topicId);
+  });
+
+  addon.self.addEventListener("reenabled", () => {
+    updateCache(topicId);
+  });
```
DNin01, line 37: Some comments might be nice:
```suggestion
  // Save the post contents after two seconds of no typing
  let timeout;
  box.addEventListener("input", () => {
    if (addon.self.disabled) return;
    clearTimeout(timeout);
    timeout = setTimeout(updateCache, 2000, topicId);
  });
  // Also save if the text field is unfocused
  box.addEventListener("blur", () => {
    updateCache(topicId);
  });
  // or if the addon is dynamically enabled
  addon.self.addEventListener("reenabled", () => {
    updateCache(topicId);
  });
```

DNin01 (dismissed):
Forum post drafts should be deleted after posting them.
Scratch forum thread with a post in the reply compose box that had already been posted

It's also not a bad idea to remove empty draft posts from local storage. That's these guys:
Picture of data - object containing two numbered keys with objects as values, each containing cache keys with empty strings

Samq64:
Just a minute I forgot about posting.

addons/forum-post-save/userscript.js
```diff
@@ -0,0 +1,70 @@
...
+  box.addEventListener("blur", () => {
+    updateCache(topicId);
+  });
+  // or if the addon is dynamically enabled
+  addon.self.addEventListener("reenabled", () => {
+    updateCache(topicId);
+  });
+
+  document.querySelector("[name=AddPostForm]")?.addEventListener("click", (e) => {
+    if (!document.querySelector(".errorlist")) {
```
DNin01, line 41: Post errors are only displayed after a page reload, so checking if there's an error message immediately after the Submit button is clicked won't work.

[`auto-follow-topics`](https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/auto-follow-topics/userscript.js) uses a different solution.

Samq64: Yeah I'll just copy that here tomorrow.

addons/forum-post-save/addon.json
DNin01, line 1: Not sure if this is necessary, but we might want to add a notice saying that draft posts are only saved temporarily, just so users don't rely on this too much.

addons/forum-post-save/userscript.js
DNin01, line 1: Wonder if we should add support for the "new topic" page.

addons/forum-post-save/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Save draft posts",
+  "description": "Saves forum posts that you haven't submitted.",
```
DNin01, line 3: It may not be clear how this addon works. (_"How do you access drafts?"_) Here's a description suggestion:
```suggestion
  "description": "Saves forum posts that you haven't submitted and restores them when you come back to where you wrote them.",
```
There might also be a better way of putting this. Do you like this, @WorldLanguages?

DNin01 (commented):
Seems to be working now. Tested successful post and error message.

addons/forum-post-save/addon.json
Samq64: Putting it in the description should be fine I think.

addons/forum-post-save/userscript.js
Samq64: There's also post edits which could be useful for long OPs.

addons/forum-post-save/addon.json
WorldLanguages: Samq's looks okay

DNin01: It's anyone's guess how long "temporary" is and what it really means here, though.

But most people will probably revisit their drafts no later than the same session or the next day so it's probably okay to say. Having an uncertain/imprecise expiration date will probably discourage users from relying on letting their unposted drafts sit there for days, which is a good thing because eventual local storage data loss is inevitable.

So, sure.

Joeclinton1:
This should be extended to comments too. Probably best to do both in the same PR

Samq64:
> This should be extended to comments too. Probably best to do both in the same PR

Maybe other forum text fields but not comments, that should be separate.

Joeclinton1:
> > This should be extended to comments too. Probably best to do both in the same PR
> 
> Maybe other forum text fields but not comments, that should be separate.

Why would there be one addon titled autosave forum posts and another titled autosave comments. They are clearly achieving the same goal.

Samq64:
> Why would there be one addon titled autosave forum posts and another titled autosave comments. They are clearly achieving the same goal.

I meant a separate PR.

Samq64:
Is there still any interest on getting this merged?

addons/forum-post-save/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Save draft posts",
+  "description": "Temporarily saves forum posts that you haven't submitted and restores them when you come back to where you wrote them.",
```
mxmou:
```suggestion
  "description": "Temporarily saves forum posts that you haven't submitted and restores them when you come back to the same topic.",
```

addons/forum-post-save/userscript.js
```diff
@@ -0,0 +1,82 @@
+export default async function ({ addon, console }) {
+  const topicId = location.pathname.split("/")[3];
+  const box = document.querySelector(".markItUpContainer .markItUpEditor");
+  if (!box) return;
+
+  const getTopics = () => JSON.parse(sessionStorage.getItem("sa-forum-post-save-topics") || "[]");
```
mxmou: This seems to store which drafts have already been submitted, but that isn't clear from the name of the key and its related functions. I think it needs a better name and possibly a comment.

addons/forum-post-save/userscript.js
```diff
@@ -0,0 +1,82 @@
...
+
+  function updateCache(topic) {
+    const update = {};
+    update[topic] = {
+      cache: box.value,
+      stamp: Date.now(),
+    };
+    const stored = _getAllCache();
+    const cache = Object.assign({ ...stored }, update);
+    if (cache === stored) return; // if no diff, return
```
mxmou: Comment repeats the code

mxmou (commented):
Should the addon show some kind of message when a draft is saved?

Samq64:
> Should the addon show some kind of message when a draft is saved?

Yes, where do you think it should display? In the toolbar or next to the submit button? It would also be nice if the browser's unsaved changes prompt was disbaled after a successful draft save.

mxmou:
> > Should the addon show some kind of message when a draft is saved?
> 
> Yes, where do you think it should display? In the toolbar or next to the submit button?

I think it would make sense to put it next to the Post button.

PR #4733, opened 2022-06-29
Migrate to uploads.scratch.mit.edu in all addons

### LINKED ISSUE #4732: Migrate to uploads.scratch.mit.edu in most addons
Scratch uses uploads now, instead of cdn2. cdn2 urls redirect directly to uploads so we should save the user a redirect.

Weredime:
Resolves #4732
### Changes
Replaces all usage of cdn2 to uploads.

### Reason for changes
To save network usage

### Tests
Firefox developer edition


apple502j:
Hm wait, does this mean hi-res-thumbnail no longer works? @towerofnix cc

CST1229:
> Hm wait, does this mean hi-res-thumbnail no longer works? cc

Doesn't cdn2 redirect to uploads already?
Also, thumbnails are still scaled with uploads..

mxmou:
> I'm not sure what "uses uploads now" in the issue description (#4732) means. Project pages (scratch-www) and profile pages (scratchr2) both still reference cdn2 URLs, so I don't know what this is referring to?

All URLs on cdn2.scratch.mit.edu now redirect to uploads.scratch.mit.edu.

Samq64:
> All URLs on cdn2.scratch.mit.edu now redirect to uploads.scratch.mit.edu.

I thought that was temporary.

towerofnix:
> All URLs on cdn2.scratch.mit.edu now redirect to uploads.scratch.mit.edu.

Oh cool. Like @CST1229 wrote, since uploads URLs still support scaling dimensions (and cdn2 redirects include those), all should keep working—but if they take down support for cdn2 altogether (which I think is unlikely / a long ways off) we should change the "output" URL to follow the uploads format of hi-res-thumbnails instead of cdn2.

Weredime:
> Edit: @Weredime Did you test hi-res-thumbnails and ensure it still works across both `uploads` and `cdn2` URLs? It looks like you just subtracted a portion of the code in the addon which is essential (ironically) for handling `uploads` URLs, i.e. converting them to `cdn2` format so that the regular expression even works on them (if the regex doesn't match anything the image is skipped and not upscaled, which generally isn't what we want).

The `uploads` server uses the exactly same format as `cdn2`.

towerofnix:
> The `uploads` server uses the exactly same format as `cdn2`.

That's not always true! On the homepage, "What's Happening?" activity images use the format `//uploads.scratch.mit.edu/users/avatars/36977.png`. Project thumbnails use `//uploads.scratch.mit.edu/projects/thumbnails/628146090.png`. Note both of these lack dimensions. I know the `get_image` format is *supported* on uploads, but not all uploads URLs use it. The removed code was written to convert these URLs into the cdn2 format which, most importantly, does include dimensions.

That said, the existing (pre-PR) code does drop (discard) dimensions which are present in a cdn2-style `uploads` URL. Since cdn2-style URLs match the cdn2 regex anyway, we could skip the reformatting if it already initially matches—something like the following code should work:

```js
  const cdn2Regex = /(.*\/get_image\/.*?\/[0-9]+?_)([0-9]+?)x([0-9]+?)(\.[a-z]+)/);
  const uploadsRegex = /^https?:\/\/uploads\.scratch\.mit\.edu\//;

  main: while (true) {
    // ...

    // If the image is from uploads.scratch.mit.edu, but doesn't match the
    // dimensions-including cdn2 style, reformat src so it does.
    if (!cdn2Regex.test(src) && uploadsRegex.test(src)) {
      const id = src.match(/[0-9]+/);
      if (src.includes("projects")) {
        // Project thumbnails are always 480x360.
        src = `//uploads.scratch.mit.edu/get_image/project/${id}_480x360.png`;
      } else if (src.includes("users")) {
        // Max user avatar size is 500x500.
        src = `//uploads.scratch.mit.edu/get_image/user/${id}_500x500.png`;
      } else if (src.includes("galleries")) {
        // Max studio thumbnail size is 500x500.
        src = `//uploads.scratch.mit.edu/get_image/gallery/${id}_500x500.png`;
      }
    }

    let width, height, newSrc;

    const cdn2 = src.match(cdn2Regex);
    // ...
```

Weredime:
I didn't mean that I supported old image urls. I misunderstood you originally, sorry!

Weredime:
@towerofnix Is this ok now?

towerofnix:
I tested locally and it looks as though it's working to me! Thanks for your patience and time working that out :)

apple502j (approved):

PR #4734, opened 2022-06-30
Volume slider addon

### LINKED ISSUE #2755: [Mute project player] Adjust volume
> Volume slider on projects, next to the stop sign (this would basically be an update for the project muting addon)

### LINKED ISSUE #4422: Volume slider in project and project editor

When a project is started in Scratch, the **sound setting** of the **computer** of the person **who created the project** may not be the same as the **sound of the computer of the person playing the project**. This means that some projects may be quieter than normal, while some projects may have a louder sound. _That's the problem_.

-----------------------------------------------------------

I think there should be a **slider** for **adjusting the volume** in **both the project and the editor**. In this way, users will be able to **change only the sound of the project**, instead of **changing the sound of their computer and changing the sound** of all their applications.

-----------------------------------------------------------

**Volume slider** => To adjust the volume of the volume (both in the project player and in the editor)

**Click on speaker icon** => To mute/unmute sound

**Alt+Up Arrow/Down Arrow** => To set the volume in multiples of 10

-----------------------------------------------------------

### Additional context

This plugin may **conflict** with the **"Muted project player mode"** plugin. Maybe they should be **combined**.

_I tried to make the **images look like the original**. **I hope it was good :)**_

_(Also ignore the LIST-BASED KAHOOT project. That project was one of the last projects I worked on and I discovered the sound problem in this project. I used this as an example. My purpose is not to advertise.)_

-----------------------------------------------------------

**Examples:**

-----------------------------------------------------------
Adsız4
-----------------------------------------------------------
Adsız3

### LINKED ISSUE #4957: A feature to adjust the volume of all projects' sounds
### What this feature/function does

Some people who visit lots of projects know that some can have sound effects that are relatively loud. Some projects' music can be over twice as loud as most YouTube videos' commentary. You might leave your volume at a good level for YouTube, but then when you go to Scratch, the projects' music is _blasting_. That's why I'm proposing a feature here to allow you to turn down the master volume of all projects on Scratch so you don't have to think about it.

### How it works

This could probably be included in the ["Muted project player mode"](https://github.com/ScratchAddons/ScratchAddons/tree/master/addons/mute-project) addon, which would now be called "Volume control and muted mode" to fit the new function. There would be a new number input setting to change the overall volume of all Scratch projects. This will default to 100, which means no change, but you can change it to any number between 0 and 100. It is basically the master volume %. For example, at 50, all projects' audio will be only half as loud as they are normally.

SA could also change this addon to add a button and slider on project pages to adjust the volume or mute the sound (like you can on YouTube with its volume slider). That's what #2755 suggests. We could also combine these two ideas, so the option I was talking about would just be the default and you could change the volume with the slider.

Image of an addon's settings with a new setting to change the project volume

Samq64:
Resolves #2755
Resolves #4422
Resolves #4957

### Changes

Adds a volume slider next to the stop sign.

image

### Reason for changes

It's useful to be able to adjust the volume of a Scratch project without having to use volume mixer or change the system volume.

### Tests

Tested on Chromium 104 and Firefox 103. `mute-project` is compatible, but there's some cases where the volume isn't set to what's expected like when dynamically enabling `vol-slider` while the project is muted by `mute-project`.


addons/mute-project/addon.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "Muted project player mode",
-  "description": "Ctrl+Click the green flag to mute/unmute the project.",
+  "name": "Project player volume control",
+  "description": "Ctrl+Click the green flag to mute/unmute the project, or use the slider",
```
apple502j, line 3: The description should end with a period. Also, just calling it "slider" is unclear.

Samq64:
Screenshot from 2022-06-30 18-49-21

Anyone know how to style the slider, for example, make the knob smaller and turn the slider pink?

Cool-Scratcher:
> Screenshot from 2022-06-30 18-49-21
> 
> Anyone know how to style the slider, for example, make the knob smaller and turn the slider pink? As far as I can tell normal css doesn't work.

Make a new element with functionality like a slider or something. I don't think you can style a built in html component like a slider...

addons/mute-project/userscript.js
```diff
@@ -1,37 +1,65 @@
...
+    addon.tab.displayNoneWhileDisabled(container, { display: "inline-block" });
+    addon.tab.appendToSharedSpace({ space: "afterStopButton", element: container, order: 0 });
+    container.appendChild(icon);
+    container.appendChild(slider);
     button.addEventListener("click", toggleMute);
     button.addEventListener("contextmenu", toggleMute);
+    setVol(1);
+    slider.addEventListener("change", function (e) {
+      setVol(this.value);
+    });
```
WorldLanguages, line 59: Is it possible to update the volume live while the user moves the slider, instead of waiting until mouseup (releasing click)?

WorldLanguages:
IMO this should be a separate addon. Ctrl+Click to mute is a 2.0 feature. Volume slider is an enhancement that some users want.
Having 2 separate addons makes it a bit harder to make them compatible with each other, but I think it makes the most sense.

Samq64:
Should I open a new PR and close this one then?

WorldLanguages:
If we decide to have 2 separate addons, mute-project will need to be made compatible anyways, so I don't see any reason to open a new PR. This one can be changed.

Samq64:
> IMO this should be a separate addon.

I'm working on that locally.

DNin01:
It might be nice to either have a default volume option in the addon's settings, or have it remember the volume you set it to previously.

Samq64:
> It might be nice to either have a default volume option in the addon's settings as I suggested in #4957 or have it remember the volume you set it to previously.

Done.

addons/mute-project/userscript.js
```diff
@@ -30,7 +30,9 @@ export default async function ({ addon, global, console }) {
       markAsSeen: true,
       reduxEvents: ["scratch-gui/mode/SET_PLAYER", "fontsLoaded/SET_FONTS_LOADED", "scratch-gui/locales/SELECT_LOCALE"],
     });
-    addon.tab.appendToSharedSpace({ space: "afterStopButton", element: icon, order: 0 });
+    if (!addon.self.getEnabledAddons("vol-sldier")) {
+      addon.tab.appendToSharedSpace({ space: "afterStopButton", element: icon, order: 0 });
+    }
```
WorldLanguages, line 35: This is not the proper way. We should use CSS to hide it

addons/vol-slider/userscript.js
```diff
@@ -0,0 +1,51 @@
...
+  addon.self.addEventListener("disabled", () => {
+    vm.runtime.audioEngine.inputNode.gain.value = 1;
+  });
+
+  addon.self.addEventListener("reenabled", () => {
+    setVol(defVol);
+  });
+
+  while (true) {
+    let button = await addon.tab.waitForElement("[class^='green-flag_green-flag']", {
```
apple502j, line 36: Unused variable

addons/vol-slider/userscript.js
```diff
@@ -0,0 +1,45 @@
+import { setup, setVol, isMuted, getDefVol, setDefVol } from "./module.js";
```
apple502j, line 1:
```suggestion
import { setup, setVol, getDefVol, setDefVol } from "./module.js";
```

apple502j (approved):

TheColaber:
Is it intended that reenabling the addon doesn't restore the old value in the slider? For example, set slider to muted, disable and reenable addon, and notice that the slider is not muted.

Samq64:
Yes, but I could probably make it restore the last value. Right now it sets it to the default volume in settings.

mxmou:
> Is it intended that reenabling the addon doesn't restore the old value in the slider? For example, set slider to muted, disable and reenable addon, and notice that the slider is not muted.

`mute-project` doesn't remember the previous muted status when reenabling. I think a similar behavior would make sense here - dynamically enabling the addon would set the volume to 100%.

WorldLanguages (dismissed):
The slider needs to be hidden while the stage is in small mode:
image

WorldLanguages:
Note: however, if the project is muted, the "muted" icon needs to stay, even in small stage mode. mute-project already does this

Hans5958:
> IMO this should be a separate addon. Ctrl+Click to mute is a 2.0 feature. Volume slider is an enhancement that some users want. Having 2 separate addons makes it a bit harder to make them compatible with each other, but I think it makes the most sense.

How come making it practically 0% and anything between 0% and 100% are different features? That is foolish.

Samq64:
whether this should be a new addon is still under discussion on Discord.

Samq64:
> Note: however, if the project is muted, the "muted" icon needs to stay, even in small stage mode. mute-project already does this

Now it only does if vol-slider isn't enabled (because of `display: none !important` on the `mute-project` icon). I can fix that, but the easiest way would probably be to check if the other addon is enabled from the userscript, but apple rejected it.

WorldLanguages:
Could we agree on adding this as a separate addon now, and "merging" it later, when we have dedicated hotkey support?

I can't think of any way to add this as part of `mute-project` in a way that isn't confusing.
1. We don't want the volume slider UI to be enabled for old `mute-project` users
2. Potential solution to 1 would be confusing - if the `volume slider` addon had a "show slider" setting that defaulted to off for all existing thousands of users, they would ask "why is the volume slider not showing, if the volume slider addon is enabled?"

If we release this as a separate addon, users that want this would just enable this new addon, and that's all.

After we add a dedicated hotkey section to the settings page, "Ctrl+Click to mute" would disappear from the addon list completely, and look like a normal input in a hotkey customizer. And we will probably have a way for addons to include "related hotkeys" inside their settings, so if we do that, these addons would be "merged" in the future, but without taking two rows in the addon list.

Opinions? @Hans5958 @mxmou

TheColaber:
i was just about to say that this problem won't exist in the future when the hotkey system comes up. this solution lgtm

Samq64:
Priority 4 and pending labels then?

WorldLanguages:
Waiting for hans thumbs up!

Hans5958:
To be honest, I don't want to do any compromises looking at past experiences. The extension is already quite slow on it's development, and people are really hesitant on changes, specifically this kind of changes that would need migrations and such. If everyone here is fine on doing these settings migration each update, then I would say it is fine, but this isn't the case.

The problem about hotkeys shouldn't be related to the design of having two separate addons or not.

The fact that you think that users will get confused because of the feature turned off tells me that you may doing this wrong, either with the update itself, or the way of telling users about the update.

I mean, just look at YouTube. Why don't we do this instead?

firefox_2022-09-13_07-33-44

Hans5958:
For the note, in case this is unclear: I would like to have this feature added in the `mute-project` addon, instead of having a separate addon.

Hans5958:
I think I misintepret the `mute-project` addon. I have to say that looking at images isn't enough to know what the addon does.

The `mute-project` only does one thing: <kbd>CTRL</kbd> + <kbd>Left Click</kbd>. It doesn't add a volume button and such, and so with this kind of addition I think it's fine? I still wish to have the fully-fledged volume controls just like YouTube.

WorldLanguages:
So, we will be releasing this as a separate new addon.

I'll soon check compatibility with `mute-project` (there's some edge cases we might have not covered), and it would be great to offer a setting (potential name: `only show slider on hover`) that makes it work like YouTube's volume selector, like Hans suggested.

addons/vol-slider/userstyle.css
```diff
@@ -0,0 +1,29 @@
+.sa-volume > * {
+  vertical-align: middle;
+  margin-top: 6px;
+}
```
WorldLanguages, line 4: I think we could add a `margin-left` here, so the volume icon isn't so close to the stop button.

Also, we might need to support RTL here, and put the margin in the other side. Not sure if other addons that add their UI elements above the stage take RTL into account for margins. If they do, this addon should too.

WorldLanguages:
I would also like to take a deeper look at the dynamicEnable/dynamicDisable logic.
For example, it looks like enabling volume-slider dynamically unmutes the project in all situations, even if the project was previously muted through mute-project. I think this topic was discussed above, but I'm not sure if this specific situation was.

@mxmou If you'd like, you could take a look into changing the UI slightly (more margin? see my review above), RTL support (see my review above too). And potentially having a setting that makes the slider hide unless the volume icon is hovered, like the gif from the YouTube UI that Hans showed above, I'd work on it but my CSS transition skills aren't good enough.

addons/vol-slider/userstyle.css
Samq64: I forgot about RTL...

WorldLanguages: We'd need to check if other similar addons take it into account.
If not, we can proceed. But we need to stay consistent with other addons that use the editor stage head

WorldLanguages:
I will merge this PR tomorrow, regardless of whether I make any changes, in order to get the addon name and description into the translation process.
We have a few more days to then check whether everything works as intended, functionality-wise.

WorldLanguages:
> I will merge this PR ~tomorrow~ in 2 days, regardless of whether I make any changes, in order to get the addon name and description into the translation process. We have a few more days to then check whether everything works as intended, functionality-wise.

See #5200

PR #4738, opened 2022-06-30
Cloud Games - highlight currently opened project.

Norbiros:
Resolves:

> TODO: add currently opened game to projects array. Sort function should put it on top

### Changes

Adds blue background to project where user is online. Also puts it on top of list.

### Tests

Everything works on Gentoo, Firefox Dev.


mxmou:
Orange border means danger. I think a blue background (`--blue`) and border (`--blue-variant`) would be better. The list of usernames could remain gray:
image

popups/cloud-games/popup.js
```diff
@@ -92,8 +92,13 @@ export default async ({ addon, msg, safeMsg }) => {
     computed: {
       projectsSorted() {
         return this.projects.sort((b, a) => {
-          if (a.amt !== b.amt) return a.amt - b.amt;
-          return a.timestamp - b.timestamp;
+          if (b.online == true) return -99;
```
WorldLanguages, line 95: `===`

popups/cloud-games/popup.js
```diff
@@ -145,8 +151,10 @@ export default async ({ addon, msg, safeMsg }) => {
             }
             const dateNow = Date.now();
             const usersSet = new Set();
+            projectObject.online = false;
             for (const varChange of json) {
               if (dateNow - varChange.timestamp > 60000) break;
+              if (varChange.user == username) projectObject.online = true;
```
WorldLanguages, line 157: `===`

WorldLanguages:
The current project is moved to the top correctly, but it looks like the rest of the projects aren't sorted by amount of users anymore.

WorldLanguages:
Also, this fails to consider a user might be online in more than one project at once. Why not stick to the currently opened project? It's already stored as `selectedTabUrl`.

Norbiros:
I can change that in few days (maybe 5).

popups/cloud-games/popup.css
```diff
@@ -105,3 +105,11 @@ body {
   text-align: center;
   padding-top: 0.5rem;
 }
+
+.online {
+  border: 1px solid var(--blue-variant);
+}
+
+.online .title {
+  background: var(--blue);
```
mxmou, line 114: Could you change the border color of this element too?

popups/cloud-games/popup.js
GrahamSH-LLK: you don't need === or ==....
remember, that just outputs a boolean that is the exact same as before the equality check

WorldLanguages: @GrahamSH-LLK Our ESLint check will complain if we use `==`

apple502j: @WorldLanguages Graham prob is saying `if (b.online)` is enough

GrahamSH-LLK: I was

Norbiros:
> Also, this fails to consider a user might be online in more than one project at once. Why not stick to the currently opened project? It's already stored as `selectedTabUrl`.

User can be online in for example project that is opened in other tab.

mxmou:
> > Also, this fails to consider a user might be online in more than one project at once. Why not stick to the currently opened project? It's already stored as `selectedTabUrl`.
> 
> User can be online in for example project that is opened in other tab.

I think it would make sense to
* pin all projects where the user is online to the top and
* highlight the project that's in the currently selected tab and move it above any other project where the user is online.

mxmou:
Please don't set the order using CSS: it's bad for accessibility and for keyboard users.

Norbiros:
A lot of things changed, now it should work fine

popups/cloud-games/popup.js
```diff
@@ -103,7 +103,7 @@ export default async ({ addon, msg, safeMsg }) => {
         if (this.projects.length === 0 && this.error !== "no-projects") return null;
         if (this.projectsChecked !== this.projects.length) return null;
         const { id, type } = extractUrl(this.selectedTabUrl);
-        if (!id || gameSet.has(`${type}/${id}`)) return null;
+        if (!id || this.projects.some((el) => el.id == id)) return null;
```
Norbiros, line 106: Now if you have studio, it doesn't ask to add game from that studio (if you had griffpatch games studio it asked for Slither.io even if that game was in studio)

Norbiros:
> Please don't set the order using CSS: it's bad for accessibility and for keyboard users.

I tried, but my solution had bugs - sometimes it putted random project on top, i tried it for some time and i came up with this solution, and it works 100% correct. If you know how to fix that, please help me

WorldLanguages:
> I tried, but my solution had bugs - sometimes it putted random project on top, i tried it for some time and i came up with this solution, and it works 100% correct. If you know how to fix that, please help me

Sorting functions are weird sometimes but ordering with CSS is a worse idea.

Norbiros:
> > I tried, but my solution had bugs - sometimes it putted random project on top, i tried it for some time and i came up with this solution, and it works 100% correct. If you know how to fix that, please help me
> 
> Sorting functions are weird sometimes but ordering with CSS is a worse idea.

When I was working on rows order (on front page), you told me:
> Ideally the order should be changed with css only but it might not be practical
But okay. I tried working with sorting function and it didn't work, so if someone know how to fix it, please tell me 😄

WorldLanguages:
> When I was working on rows order (on front page), you told me:
> 
> > Ideally the order should be changed with css only but it might not be practical

I think I said that because Scratch's front page uses React, and a state update can cause it to rerender, and I'm not sure what would happen if we switched the order of the elements in a way React hasn't expected could be possible. I haven't done much research, but I've seen crashes if we call `element.remove()` on elements controlled by React, so we hide them instead.
In this case, we control the whole popup environment, so none of these problems apply, so the best solution is to sort from JS before telling Vue what to render.

mxmou (approved):

WorldLanguages (approved):

PR #4748, opened 2022-07-03
New addon: Delete button in editor

### LINKED ISSUE #4723: Addon Request: Delete Button Inside of Editor
I often wish to delete a project while I'm inside of its editor for many different reasons.

I'd love an addon that adds a delete button inside of the editor:
image
After clicking it, the project would be deleted and the user would be redirected back to My Stuff.

Obviously you could just delete the project from My Stuff, but that is somewhat annoying to do.

### Additional context
N/A

Weredime:
Resolves #4723

### Changes
Adds a delete button in the editor

### Tests
Tested on Firefox Dev

### DIFF
--- a/addons-l10n/en/editor-delete-button.json
+++ b/addons-l10n/en/editor-delete-button.json
@@ -1,7 +1,7 @@
 {
-    "editor-delete-button/button-text": "Delete",
-    "editor-delete-button/modal-msg": "Are you sure you want to delete this project?",
-    "editor-delete-button/yes": "Yes",
-    "editor-delete-button/no": "No",
-    "editor-delete-button/modal-title": "Deleting project"
+  "editor-delete-button/button-text": "Delete",
+  "editor-delete-button/modal-msg": "Are you sure you want to delete this project?",
+  "editor-delete-button/yes": "Yes",
+  "editor-delete-button/no": "No",
+  "editor-delete-button/modal-title": "Deleting project"
 }
--- a/addons-l10n/en/editor-delete-button.json
+++ b/addons-l10n/en/editor-delete-button.json
@@ -1,0 +1,7 @@
+{
+    "editor-delete-button/button-text": "Delete",
+    "editor-delete-button/modal-msg": "Are you sure you want to delete this project?",
+    "editor-delete-button/yes": "Yes",
+    "editor-delete-button/no": "No",
+    "editor-delete-button/modal-title": "Deleting project"
+}
--- a/addons/editor-delete-button/addon.json
+++ b/addons/editor-delete-button/addon.json
@@ -1,0 +1,19 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/dist/schema.json",
+  "name": "Delete button in editor",
+  "description": "Adds a delete button to the file menu that allows you to delete your project.",
+  "credits": [
+    {
+      "name": "Weredime (9gr)",
+      "link": "https://scratch.mit.edu/users/9gr/"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "tags": ["editor", "editorMenuBar"],
+  "versionAdded": "1.28.0"
+}
--- a/addons/editor-delete-button/userscript.js
+++ b/addons/editor-delete-button/userscript.js
@@ -1,0 +1,45 @@
+export default async ( /** @type {import("../../addon-api/content-script/typedef").UserscriptUtilities} */{ addon, console, msg }) => {
+    const { redux } = addon.tab;
+    while (true) {
+        const loadItem = await addon.tab.waitForElement("div[class^='menu-bar_file-group'] > :nth-child(3) ul > :nth-child(4):not(.sa-editor-delete-button)", {
+            markAsSeen: true,
+            reduxCondition: (state) => !state.scratchGui.mode.isPlayerOnly && typeof state.session?.session?.user === "object" && !state.preview.visibilityInfo.deleted,
+        });
+        
+        const dropdownItem = document.createElement("li");
+        dropdownItem.className = addon.tab.scratchClass("menu_menu-item", "menu_hoverable", { others: ["sa-editor-delete-button"] });
+        dropdownItem.innerText = msg("button-text");
+        addon.tab.displayNoneWhileDisabled(dropdownItem, { display: "block" });
+
+        dropdownItem.addEventListener("click", async (e) => {
+            if (!await addon.tab.confirm(msg("modal-title"), msg("modal-msg"), { okButtonLabel: msg("yes"), cancelButtonLabel: msg("no"), useEditorClasses: true })) return;
+            const res = await fetch(`/site-api/projects/all/${redux.state.preview.projectInfo.id}/`, {
+                headers: {
+                    "X-CSRFToken": addon.auth.csrfToken,
+                },
+                method: "PUT",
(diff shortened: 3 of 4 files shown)


addons-l10n/en/editor-delete-button.json
```diff
@@ -0,0 +1,7 @@
+{
+  "editor-delete-button/button-text": "Delete",
```
apple502j, line 2: This should say "Delete Project".

mxmou: I don't think "project" needs to be capitalized.

addons-l10n/en/editor-delete-button.json
```diff
@@ -0,0 +1,7 @@
...
+  "editor-delete-button/modal-msg": "Are you sure you want to delete this project?",
+  "editor-delete-button/yes": "Yes",
+  "editor-delete-button/no": "No",
+  "editor-delete-button/modal-title": "Deleting project"
```
mxmou, line 6: *This* should say "Delete Project".

lisa-wolfgang:
Does this "delete" the project or just move it to trash? I think that should be clarified somewhere.

Weredime:
> Does this "delete" the project or just move it to trash? I think that should be clarified somewhere.

There is no way to permanently "delete" a project; The best thing you can do is move it to the trash.

lisa-wolfgang:
There are actually three plausible interpretations of "delete":
- The project is simply moved to the trash
- The project is deleted from the trash and is rendered invisible to even the project author by normal means
- The project is replaced with the default new project file and then deleted from the trash, which in effect permanently deletes the original project

addons-l10n/en/editor-delete-button.json
```diff
@@ -0,0 +1,7 @@
+{
+  "editor-delete-button/button-text": "Delete",
+  "editor-delete-button/modal-msg": "Are you sure you want to delete this project?",
+  "editor-delete-button/yes": "Yes",
+  "editor-delete-button/no": "No",
+  "editor-delete-button/modal-title": "Delete project"
```
mxmou, line 6: This should be capitalized to be consistent with Scratch's modals such as "Make a Block".

addons/editor-delete-button/userscript.js
```diff
@@ -0,0 +1,56 @@
...
+        });
+        // TODO: Maybe go to mystuff instead?
+        redux.dispatch({
+          type: "scratch-gui/mode/SET_PLAYER",
+          isPlayerOnly: true,
+        });
+      }
+    });
+
+    loadItem.insertAdjacentElement("beforebegin", dropdownItem);
```
mxmou, line 54: A better position for this option would be at the bottom of the menu, with a separator above:
image

addons/editor-delete-button/userscript.js
```diff
@@ -0,0 +1,56 @@
...
+        reduxCondition: (state) => !state.scratchGui.mode.isPlayerOnly && !state.preview.visibilityInfo.deleted,
+      }
+    );
+
+    const dropdownItem = document.createElement("li");
+    dropdownItem.className = addon.tab.scratchClass("menu_menu-item", "menu_hoverable", {
+      others: ["sa-editor-delete-button"],
+    });
+    dropdownItem.innerText = msg("button-text");
+    addon.tab.displayNoneWhileDisabled(dropdownItem, { display: "block" });
```
mxmou, line 19: This has no effect if you don't add `"dynamicEnable": true` and `"dynamicDisable": true` to `addon.json`.

mxmou (approved):

addons/editor-delete-button/userscript.js
```diff
@@ -0,0 +1,66 @@
...
+      const confirmed = await addon.tab.confirm(msg("modal-title"), msg("modal-msg"), {
+        okButtonLabel: msg("yes"),
+        cancelButtonLabel: msg("no"),
+        useEditorClasses: true,
+      });
+
+      if (addon.settings.get("resetProject")) {
+        const defaultProject = await defaultProjectPromise;
+        vm.loadProject(defaultProject);
+      }
```
mxmou, line 35: If you keep the setting, move this below `if (!confirmed) return`.

addons/editor-delete-button/addon.json
```diff
@@ -0,0 +1,29 @@
...
+    }
+  ],
+  "settings": [
+    {
+      "name": "Reset project before trashing",
+      "type": "boolean",
+      "default": true,
+      "id": "resetProject"
+    }
+  ],
```
mxmou, line 24: This setting doesn't work because it doesn't save the project.

Even if it did what it says, I'm not sure if it's useful: all it does is make the project impossible to restore. If you want to keep it, it should be disabled by default because it's dangerous.

mxmou (changes requested):

lisa-wolfgang:
Again, because there are [three different types of deletions](https://github.com/ScratchAddons/ScratchAddons/pull/4748#issuecomment-1173128755), the addon needs to use clarifying language to describe exactly what will happen to the project. Deletion (or lack thereof) is serious business, so it's important that the user understands exactly what they're doing.

apple502j:
Question: can this delete shared projects? (it shouldn't be able to, if it is)

WorldLanguages:
Let's be careful and be careful when merging this it will be the first addon that simplifies deleting an asset. I know deleted projects go to the trash first, but still.
I think the addon manifest should have a note indicating that deleted projects go to the trash.

Weredime:
Implemented in [`cbab8a1` (#4748)](https://github.com/ScratchAddons/ScratchAddons/pull/4748/commits/cbab8a13330363124fdbc0bbefe3171cbfbd8126)

apple502j:
@Weredime Bug: logged-out editor shows Delete button, which it shouldn't.

addons/editor-delete-button/userscript.js
```diff
@@ -0,0 +1,114 @@
...
+        useEditorClasses: true,
+      });
+      if (!confirmed) return;
+      if (addon.settings.get("resetProject")) {
+        const defaultProject = await defaultProjectPromise;
+        vm.loadProject(defaultProject);
+        const safe = await new Promise((resolve, reject) => {
+          const xhrOpen = XMLHttpRequest.prototype.open;
+          XMLHttpRequest.prototype.open = function (method, url, ...args) {
+            if (method === "put" && String(url).startsWith("https://projects.scratch.mit.edu")) {
```
apple502j, line 67:
```suggestion
            if (method === "put" && String(url).startsWith("https://projects.scratch.mit.edu/")) {
```

apple502j (changes requested):
Fix the bug with logged-out editor. Also we need to better handle Enter key on the modal. It should treat Enter for dangerous operation like this (keyboard users need to tab on the OK button and press Enter) but I do believe this is out-of-scope.

mxmou:
Could the reset option be per project (a checkbox in the confirmation modal)? Even if someone always wants to delete things permanently, this would require confirming it explicitly each time, making the feature less dangerous.

MystPi:
I still think this addon would be useful. Maybe the button could be reworded from 'delete' to 'move to trash'? It would make it clear that the project would not be deleted permanently.

DNin01:
> Maybe the button could be reworded from 'delete' to 'move to trash'? It would make it clear that the project would not be deleted permanently.

That makes sense.

I still feel like this addon is kind of out of scope, though.

Secret-chest:
Put it in the file menu?

apricotwtf:
> Put it in the file menu?

The button is already in the file dropdown.
image

PR #4757, opened 2022-07-04
new addon: `fave-tab`

### LINKED ISSUE #4057: Favourites Tab

Scratch's favourite system doesn't make it very easy to find a specific favourited project if there are a lot of them.

An extension pop-up tab that lists all of the user's favourite projects with a search bar for easily finding one.

Using Cloud Games once it's customizable, but it takes up more space because it shows the online players and isn't searchable.

### Additional context

Maybe there could also be a toggle to show someone else's favourites when visiting his profile.

Samq64:
Resolves #4057

### Changes

Adds a new extension popup addon which shows the currently logged user's favorite projects.

### Reason for changes

It makes opening favorite projects easier.

### Tests

Tested on Chromium 103. Still WIP, but it works.

### DIFF
--- a/popups/fave-tab/popup.js
+++ b/popups/fave-tab/popup.js
@@ -11,8 +11,8 @@ const fetchFaves = async (user) => {
   if (res.status >= 400) {
     throw HTTPError.fromResponse("Error when fetching favorites", res);
   }
-  return res.json()
-}
+  return res.json();
+};
 
 export default async ({ addon, msg, safeMsg }) => {
   window.vue = new Vue({
@@ -26,9 +26,9 @@ export default async ({ addon, msg, safeMsg }) => {
       },
     },
     async created() {
-      if (!await addon.auth.fetchIsLoggedIn()) {
+      if (!(await addon.auth.fetchIsLoggedIn())) {
         this.error = msg("login");
-        return
+        return;
       }
       const username = await addon.auth.fetchUsername();
       let projects;
@@ -36,7 +36,7 @@ export default async ({ addon, msg, safeMsg }) => {
         projects = await fetchFaves(username);
       } catch (e) {
         if (e instanceof HTTPError) {
-          this.error = msg((e.status >= 500) ? "server-error" : "general-error");
+          this.error = msg(e.status >= 500 ? "server-error" : "general-error");
         }
       }
       this.projects = projects.map((project) => ({ title: project.title, id: project.id }));
--- a/popups/fave-tab/popup.js
+++ b/popups/fave-tab/popup.js
@@ -3,14 +3,14 @@ export default async ({ addon, msg, safeMsg }) => {
     el: "body",
     data: {
       projects: [],
-      loaded: false,
+      projectsVisable: false,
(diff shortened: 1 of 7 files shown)


popups/fave-tab/popup.js
```diff
@@ -0,0 +1,16 @@
+export default async ({ addon, msg, safeMsg }) => {
+  window.vue = new Vue({
+    el: "body",
+    data: {
+      projects: [],
+      projectsVisable: false,
+    },
+    async created() {
+      const username = await addon.auth.fetchUsername();
+      const res = await fetch(`https://api.scratch.mit.edu/users/${username}/favorites/?limit=40`);
```
Samq64, line 10: Is there a way to fetch more than 40 projects? I want to add a search bar to and be able to search all favorites.

lisa-wolfgang: I believe that there's a `page` parameter for fetching the next set of 40, but I haven't checked if that's implemented for favorites.

WorldLanguages:
I think some languages don't have enough space for 4 tabs on the popup 🤔

lisa-wolfgang:
Can the popup be dynamically made wider?

lisa-wolfgang:
Another option would be to collapse unopened tabs by hiding the text.

popups/fave-tab/popup.js
CST1229: There's an `offset` parameter that offsets individual items, so you can set it to multiples of 40 to show different pages.

mxmou:
> Another option would be to collapse unopened tabs by hiding the text.

I think we should do that, but only when there isn't enough space to display them normally.

WorldLanguages:
We're not ready yet to support a third popup addon, so we should fix that if we consider this popup is worth adding.

@Samq64 Can you explain the motivation for making this a popup addon? And I just searched over sent feedback, and a common suggestion is to add a search bar to favorite projects, but it could also be added to /users/x/favorites instead of inside this new popup.

Samq64:
> @Samq64 Can you explain the motivation for making this a popup addon? And I just searched over sent feedback, and a common suggestion is to add a search bar to favorite projects, but it could also be added to /users/x/favorites instead of inside this new popup.

You're right, that would work just as well. I guess I'll do that instead, but won't make a PR just yet.

Secret-chest:
Why? The popup tab had easy access! @WorldLanguages

Samq64:
I'm going to re-open this and if a search bar is added on the favorites page, this pr can me linked to be closed.

lisa-wolfgang:
I think W_L is concerned that regular browser bookmarks achieve the same purpose that a popup tab would. However, I would disagree -- realistically, most users don't have all of their favorites bookmarked. In addition, some users may have thousands of favorited projects. This would make it convenient for the user to access their favorite projects without navigating to their profile's favorites page every time, even if they have that specific page bookmarked. There could even be another feature added to justify the tab's existence beyond convenience, like a very simple recommendation system that cycles through the list randomly.

PR #4761, opened 2022-07-06
Use the same color for debugger window, category icon and blocks

### LINKED ISSUE #3005: Debugger icon
the debugger icon needs to match the block colour when in the editor

mxmou:
Resolves #3005

### Changes

Changes the icon of the debugger block category and the green color used in the debugger window to match the block color.
image


apple502j (approved):

WorldLanguages:
At some point, other addons might add their own blocks, so the category would be named "addons" instead of "debugger". But until then, this is good to merge, right?

mxmou:
> At some point, other addons might add their own blocks, so the category would be named "addons" instead of "debugger". But until then, this is good to merge, right?

There could also be separate categories for each addon, depending on how many blocks they add. But for now I think this icon is good.

PR #4776, opened 2022-07-09
Scratch front page tools

Platformer444:
Resolves a part of #1758

### Changes
This addon helps to customize the Scratch Front Page by editing row names and hiding rows.

### Tasks
- [x] Switch to Settings Page for row editing

### Tests
Tested on Chrome and works successfully


Weredime:
Please merge the code into a single javascript file, and not use `var` for variable declarations. Use `let` instead.

addons/scratch-front-page-tools/addon.json
```diff
@@ -0,0 +1,47 @@
...
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "rows",
+      "id": "rowData",
+      "type": "table",
```
lisa-wolfgang, line 14: I don't think a table setting is the right idea for this addon unless you're considering expanding its functionality. Table settings allow the user to reorder rows, but the addon doesn't seem to support this. Additional rows can also be added in a table setting, but again, the addon doesn't seem to support the addition of new rows on the front page, so the user is likely to be confused by this perceived lack of functionality.

lisa-wolfgang (changes requested):

addons/scratch-front-page-tools/addon.json
WorldLanguages: I thought rows were reorder-able

Platformer444: I tried to hard to Reorder rows but it didn't work.

Platformer444: @WorldLanguages if you want I could add those buttons back because I think the settings page lack functionality.

WorldLanguages: I don't understand, can't the rows be reordered?

TheColaber:
closing unless @Platformer444 wants to continue to work on this.

PR #4778, opened 2022-07-09
Add latestUpdate to hide-flyout, change default mode

### LINKED ISSUE #4668: [v1.27.0] Add "new" tag to hide-flyout, consider other changes
#4506 fixes a lot of bugs, and this addon is the least used recommended addon (#4100), so we should consider showcasing it so more users can enjoy it, even tho it doesn't have any new settings.

We could also consider rephrasing the current notice. It currently goes:
> "Palette area hover" mode only extends the viewing area. If you want to be able to drag blocks into that area without them getting trashed, use one of the other modes.

We could also consider
1. Changing the default mode
2. ~~Adding a notice reminding users to reload the page when switching modes (there's no setting autorefresh, and most addons support this, so it might be confusing)~~ Fixed by #4752

WorldLanguages:
Resolves #4668

- Adds latestUpdate to addon:
image
- Credit Maximouse for the addon
- Change default mode to category hover. There's no migration, so it will only affect new Scratch Addons users - old SA users that had this enabled will still have category hover as the default. For this reason, I'm keeping the info hoverExplanation.

### DIFF
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -83,7 +83,7 @@
           "name": "Category click"
         }
       ],
-      "default": "hover"
+      "default": "cathover"
     },
     {
       "name": "Animation duration",
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -5,6 +5,10 @@
     {
       "name": "TheColaber",
       "link": "https://scratch.mit.edu/users/TheColaber"
+    },
+    {
+      "name": "Maximouse",
+      "link": "https://scratch.mit.edu/users/Maximouse/"
     }
   ],
   "info": [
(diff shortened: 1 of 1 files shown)


addons/hide-flyout/addon.json
```diff
@@ -106,5 +110,10 @@
       "default": "default"
     }
   ],
-  "tags": ["editor", "codeEditor", "recommended"]
+  "tags": ["editor", "codeEditor", "recommended"],
+  "latestUpdate": {
+    "version": "1.27.0",
+    "temporaryNotice": "This addon was revisited and many bugs were fixed.",
```
apple502j, line 116:
```suggestion
    "temporaryNotice": "This addon was revised and many bugs were fixed.",
```

mxmou (approved):

PR #4791, opened 2022-07-10
initialise-sprite-position: run on sprite duplication

### LINKED ISSUE #4786: Customizable new sprite position does not work when you duplicate sprite
### Description

When you use Customizable new sprite position it won't work with duplicates.
### Reproduction steps

1 Create a sprite
2 Right click it and click Duplicate.

### Expected behavior

It should go to 0 0(the location I set) when I duplicate it.

### Actual behavior

It went to a place around 0 0

### Environment

- Browser: google chrome
- Scratch Addons version: 1.26.1
- Operating system: Mac

That is it

pufferfish101007:
Resolves #4786

### Changes

- adds a setting for how to handle duplication: either randomise (scratch's default), go to custom x/y values, or stay in the same position as the original sprite; setting & option names & orders should probably be reconsidered
- uses `const` instead of `let` in a few places

### Reason for changes

- it's a bug, and users should choose the behaviour they want
- `const` is love, `const` is life

### Tests

tested on chromium (kiwi) 101 with all 3 settings + dynamic enable/disable

### DIFF
--- a/addons/initialise-sprite-position/addon.json
+++ b/addons/initialise-sprite-position/addon.json
@@ -42,11 +42,11 @@
       "id": "duplicate",
       "name": "Behaviour when duplicating sprites",
       "type": "select",
-      "default": "keep",
+      "default": "randomize",
       "potentialValues": [
         {
           "id": "custom",
-          "name": "Send to custom x/y values"
+          "name": "Send to specified x/y values"
         },
         {
           "id": "keep",
--- a/addons/initialise-sprite-position/addon.json
+++ b/addons/initialise-sprite-position/addon.json
@@ -60,6 +60,10 @@
     }
   ],
   "versionAdded": "1.13.0",
+  "latestUpdate": {
+    "version": "1.28.0",
+    "newSettings": ["duplicate"]
+  },
   "tags": ["editor", "featured"],
   "enabledByDefault": false
 }
--- a/addons/initialise-sprite-position/addon.json
+++ b/addons/initialise-sprite-position/addon.json
@@ -37,6 +37,26 @@
       "name": "Randomize the position of library sprites",
       "type": "boolean",
       "default": false
+    },
+    {
+      "id": "duplicate",
+      "name": "Behaviour when duplicating sprites",
+      "type": "select",
+      "default": "keep",
+      "potentialValues": [
+        {
+          "id": "custom",
+          "name": "Send to custom x/y values"
+        },
+        {
+          "id": "keep",
+          "name": "Keep the same as the original sprite"
+        },
+        {
+          "id": "randomize",
+          "name": "Randomize"
+        }
+      ]
     }
   ],
   "versionAdded": "1.13.0",
--- a/addons/initialise-sprite-position/addon.json
+++ b/addons/initialise-sprite-position/addon.json
@@ -40,7 +40,7 @@
     },
     {
       "id": "duplicate",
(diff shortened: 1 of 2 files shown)


TheColaber (approved):

RedGuy12 (approved):

Samq64:
I think there should also be an option to shift the new sprite a little, like the paint editor.

mxmou:
> I think there should also be an option to shift the new sprite a little, like the paint editor.

I don't think that behavior is useful - its purpose is to make the duplicate visible on the stage but hopefully people who enabled this addon won't expect that. Randomize might be a better default though.

addons/initialise-sprite-position/addon.json
```diff
@@ -37,6 +37,26 @@
       "name": "Randomize the position of library sprites",
       "type": "boolean",
       "default": false
+    },
+    {
+      "id": "duplicate",
+      "name": "Behaviour when duplicating sprites",
+      "type": "select",
+      "default": "keep",
```
WorldLanguages, line 45: I'm not sure if the default should be "keep". This setting did not exist, but the default has been "randomize" for 15+ months or so.

WorldLanguages (changes requested):

addons/initialise-sprite-position/addon.json
```diff
@@ -37,6 +37,26 @@
...
+      "potentialValues": [
+        {
+          "id": "custom",
+          "name": "Send to custom x/y values"
```
WorldLanguages, line 49: When I chose this option while testing, I expected two new settings to pop up. (using `setting.if`). Of course, then I noticed this referred to the same x/y values as the default for new sprites.
Maybe it could be a bit more clear that it's referring to the x/y values. `specified` instead of `custom`?

pufferfish101007: Soumds good to me 👍

addons/initialise-sprite-position/addon.json
```diff
@@ -37,6 +37,26 @@
       "name": "Randomize the position of library sprites",
       "type": "boolean",
       "default": false
+    },
+    {
+      "id": "duplicate",
+      "name": "Behaviour when duplicating sprites",
```
apple502j, line 43:
```suggestion
      "name": "Behavior when duplicating sprites",
```

apple502j (changes requested):

addons/initialise-sprite-position/userscript.js
```diff
@@ -19,4 +20,29 @@ export default async function ({ addon }) {
...
+        }
+        return newSprite;
+      });
+    };
+  };
+
+  if (vm.runtime.getTargetForStage()) {
+    registerDupPrototype();
+  } else {
+    vm.runtime.on(vm.runtime.constructor.PROJECT_LOADED, registerDupPrototype);
```
WorldLanguages, line 46: Any reason not to use `vm.runtime.on("PROJECT_LOADED", registerDupPrototype)` like other addons?

WorldLanguages (approved):

addons/initialise-sprite-position/userscript.js
WorldLanguages: It could also be `.once` instead of `.on`.

WorldLanguages: Solved my commit

apple502j (approved):

PR #4796, opened 2022-07-12
Fix editor-unshare-button bug

WorldLanguages:
Resolves the bug that @Norbiros commented over here: https://github.com/ScratchAddons/ScratchAddons/pull/4639#discussion_r906731593

To repro the bug on master:
1. Go to a shared project you own
2. Click the unshare button
3. Confirm the action
4. Click the share button
5. The editor-unshare-button addon thinks the "unshare" button was clicked, and asks to confirm
6. If the confirmations addon is set up to confirm sharing projects, it will also ask to confirm.

Bug was fixed by removing the confirmation described on number 5 - this addon shouldn't do anything if the "share" button was clicked after using the "unshared" button.

### DIFF
--- a/addons/editor-unshare-button/userscript.js
+++ b/addons/editor-unshare-button/userscript.js
@@ -8,11 +8,7 @@ export default async function ({ addon, msg, global, console }) {
 
     button.classList.add("sa-unshare-button");
     button.querySelector("span").innerText = msg("unshare-button");
-    button.addEventListener("click", async (e) => {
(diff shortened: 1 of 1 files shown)


TheColaber:
I feel like you could have come up with a better name than `thisFunction` :P

WorldLanguages:
> I feel like you could have come up with a better name than `thisFunction` :P

`thisFunction` is already used in 2 other files :P

PR #4803, opened 2022-07-13
Refresh setting for forum-live-preview

### LINKED ISSUE #1442: WYSIWYG forum editor
(I understand this would be rather hard! Just putting the idea out there!)

Some people might not understand Bbcode so a WYSIWYG (what you see is what you get) editor for the forums world be nice.

### LINKED ISSUE #4782: Customizable `forum-live-preview` preview box refresh delay
My suggestion is a feature that allows you to specify how long you need to pause your typing for the preview to refresh. Currently, it always has a 1 second delay, but with an option, you could customize it to be faster or slower (maybe 0.5 - 3 seconds).

The delay before the preview refreshes can be adjusted by changing this part of the code:
https://github.com/ScratchAddons/ScratchAddons/blob/fd0877e1f997818542d63199b8925091ff5304f5/addons/forum-live-preview/userscript.js#L26

Most of us probably won't change this, but it's a nice touch for someone who wants to.

Samq64:
Resolves #4782

### Changes

Adds a setting to `forum-live-preview` for adjusting the refresh rate.

### Reason for changes

Some people might want to configure the refresh rate.

### Tests

Tested on Chromium 103.

### DIFF
--- a/addons/forum-live-preview/addon.json
+++ b/addons/forum-live-preview/addon.json
@@ -28,20 +28,20 @@
       "id": "refresh",
       "type": "select",
       "potentialValues": [
-       {
-           "name": "fast",
-           "id": "250"
-         },
-         {
-           "name": "normal",
-           "id": "1000"
-         },
-         {
-           "name": "slow",
-           "id": "3000"
-         }
-       ],
-       "default": "normal"
+        {
+          "name": "fast",
+          "id": "250"
+        },
+        {
+          "name": "normal",
+          "id": "1000"
+        },
+        {
+          "name": "slow",
+          "id": "3000"
+        }
+      ],
+      "default": "normal"
     }
   ],
   "tags": ["community", "forums"],
--- a/addons/forum-live-preview/addon.json
+++ b/addons/forum-live-preview/addon.json
@@ -27,9 +27,9 @@
       "name": "Refresh rate (ms)",
       "id": "refresh",
       "type": "integer",
-      "min": "250",
-      "max": "10000",
-      "default": "1000"
+      "min": 250,
+      "max": 10000,
+      "default": 1000
     }
   ],
   "tags": ["community", "forums"],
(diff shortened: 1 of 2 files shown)


addons/forum-live-preview/userscript.js
```diff
@@ -23,7 +23,7 @@ export default async function ({ addon, console }) {
   let timeout;
   textarea.addEventListener("input", () => {
     if (timeout !== undefined) clearTimeout(timeout);
-    timeout = setTimeout(updatePreview, 1000);
+    timeout = setTimeout(updatePreview, addon.settings.get("refresh"));
```
WorldLanguages, line 26: Please declare a const instead of inlining the `addon.settings.get()` call

apple502j:
JSON validation is failing because the min/max/default are string, not integer. Since it does not match [the definition](https://github.com/ScratchAddons/manifest-schema/blob/master/1/1.18.json#L530) of `tableNestableSettings` it assumes it is a non-table-nestable setting, which is the table setting, which requires rows to be defined.

apple502j:
Also, WYSIWYG is NOT the live preview. For that you should ideally be able to click on the preview, type on the preview, then the textbox should be changed to reflect it - no need to type BBCode at all.

RedGuy12 (approved):

addons/forum-live-preview/addon.json
```diff
@@ -22,6 +22,16 @@
...
   ],
+  "settings": [
+    {
+      "name": "Refresh rate (ms)",
+      "id": "refresh",
+      "type": "integer",
+      "min": 250,
+      "max": 10000,
+      "default": 1000
+    }
```
WorldLanguages, line 33: A ms setting is pretty precise. Maybe we could offer three or four options only? Just like `Auto-hiding block palette`'s "animation duration" setting

Samq64: I'll do that, but it might be a good idea for settings to support floats eventually.

WorldLanguages (approved):

PR #4816, opened 2022-07-15
Compress PNGs

Samq64:
Resolves none

### Changes

Compresses PNGs.

### Reason for changes

Smaller files sizes are always better if the quality is the same.

### Tests

Not tested.


apple502j:
Please don't compress the logo; otherwise probably ok

Samq64:
> Please don't compress the logo; otherwise probably ok

Why? it's the only one that saves more than 10KB.

Samq64:
@apple502j Which logo are you talking about? sAll of them?

apple502j:
> Why? it's the only one that saves more than 20KB.

Other places use the logo too, and we should use the same file everywhere.

Weredime (approved):

WorldLanguages:
Does it make sense to compress Scratch emojis?

And many images only save ~100 bytes when compressed, is it worth it to compress those?

Weredime:
> Does it make sense to compress Scratch emojis?
> 
> And many images only save ~100 bytes when compressed, is it worth it to compress those?

It's still better than nothing - We save file space in the bundle size, thus faster downloading and updating. Every little thing counts

WorldLanguages:
I don't think it makes sense to cut 100 bytes in images when we could avoid shipping all language locales within the extension. I think we should only compress some images where it makes sense. If we really want to invest efforts into shrinking extension size (which I'm not convinced we should), that's where we would look first.

apple502j (approved):

Samq64:
> I don't think it makes sense to cut 100 bytes in images when we could avoid shipping all language locales within the extension. I think we should only compress some images where it makes sense. If we really want to invest efforts into shrinking extension size (which I'm not convinced we should), that's where we would look first.

OK, I'll remove the ones that save less than 500 bytes. I wanted to compress the logo, which actually saves a reasonable amount (25KB) but apple doesn't want me to.

Weredime:
> I don't think it makes sense to cut 100 bytes in images when we could avoid shipping all language locales within the extension. I think we should only compress some images where it makes sense. If we really want to invest efforts into shrinking extension size (which I'm not convinced we should), that's where we would look first.

<sup>build step pls</sup>

WorldLanguages (changes requested):
Blocking merge for now

Issue #3700, opened 2021-10-23
Split editor-devtools into more specific addons


WorldLanguages:
This is a quick, first draft, to spin up discussion on how this should be done.

Move costume to top/bottom was already removed from devtools and is now an independent addon.
Ctrl+Left and Ctrl+Right in costume editor to navigate to previous/next costume has been broken for a while and no one noticed until I asked when writing this issue :P We should not readd that until we can decide how to better handle hotkeys. Hopefully, we can add a dedicated hotkey system for the editor at some point, that lets the user map hotkeys to addon functions or just vanilla Scratch methods, like going to next costume, pausing the project, and anything we could imagine.

Random comment: this addon is currently, how do I say it, too advanced? Griffpatch initially did it for just himself, but it now has many times the users it used to have, so we need to focus on making every feature understandable to the average user.

Okay, so this is how I would split up the addon right now. To do this, I opened the "help" modal of devtools and did my best to imagine what should be separate. Addon names will probably change.
- **Find bar:** on the code tab, quickly find and jump to any custom block, variable, event, or hat block. On the costumes tab, search costumes by name. Currently searching for sounds is not supported, but that could be added. Shift+Click or middle mouse button on a variable opens it on the find bar. Shift+Click or middle mouse button on a custom block definition (only definition) opens it on the find bar (see addon below for middle click on custom blocks callers).
- **Jump to block definition:** using the middle mouse button or Shift + Click on a custom block (but not its definition) allows you to jump to its definition.

Both addons above use Ctrl+LeftArrow and Ctrl+RightArrow, which is a problem. What they do is *Navigate to previous / next visited position in the code area (after using the navigate to block or find bar). This allows you to middle click a custom block to go to its definition, then press ctrl + Left to go back to where you were before.*
AFAIK, since devtools is a single addon, these hotkeys follow a single history entry list. So I'm not sure how that should behave now that you can jump to a custom block definition without enabling the find bar.

- **Cut/copy/paste blocks:** right click a block and 3 new options are available to Copy All, Copy Block, and Cut Block. Right click the workspace to paste the copied block to cursor position (Scratch pastes where you copied it by default). This addon could possibly be refactored to use actual Scratch VM or Blockly methods, instead of emulating the user pressing Ctrl+C and Ctrl+V, which has caused a bug in the past (#3513)
[...]

PR #4817, opened 2022-07-16
Extract find bar and jump to definition from devtools to separate addons

TheColaber:
Related to https://github.com/ScratchAddons/ScratchAddons/issues/3700

### Changes

- New addon: find-bar
- New addon: jump-to-definition
- Removed help popup
- Dynamic Enable and disable for both new addon
- Search for sounds in find bar
- Action when find-bar is enabled but jump-to-definition is not.
- Transition code
- Bug fixes for find-bar.

### Devtools extension
I spoke to WL and he said that we will do one of two things.
1. Remove more and more features from the extension so that users can transition to SA
2. Bundle up this addon and the devtools addon together

Feel free to review if I've missed anything.

### Tests
I need people to test this.


Samq64:
The only problem I've noticed so far is that dark mode styles are broken. Other than that it looks pretty good.

image

Samq64:
Also, the help button doesn't work anymore.

TheColaber:
> Also, the help button doesn't work anymore.

Yeah, I'm going to probably remove it since the addon is separate from devtools now.

Weredime:
the folder `blockly` should be put into `libraries` because multiple addons use it

Samq64:
Is the find label next to the textbox necessary? The textbox placeholder already says that.

addons-l10n/en/editor-devtools.json
```diff
@@ -1,40 +1,8 @@
...
   "editor-devtools/unused-list": "{count, plural, one {Developer tools: Delete 1 unused local list? Here it is:\n} other {Developer tools: Delete # unused local lists? Here they are:\n} }",
   "editor-devtools/var-not-exist": "That variable does not exist...",
-  "editor-devtools/find": "Find",
-  "editor-devtools/find-placeholder": "Find (Ctrl+F)",
-  "editor-devtools/deep": "Deep",
-  "editor-devtools/help": "Help",
-  "editor-devtools/tutorials": "Tutorials",
   "editor-devtools/replace": "Developer tools: Switch all {name} in this sprite for the variable named:",
   "editor-devtools/top": "send to top",
   "editor-devtools/bottom": "send to bottom",
```
Samq64, line 8: Wasn't this removed from devtools a while ago?

addons-l10n/en/find-bar.json
```diff
@@ -0,0 +1,6 @@
+{
+  "find-bar/find": "Find",
+  "find-bar/find-placeholder": "Find (Ctrl+F)",
+  "find-bar/help": "Help",
+  "find-bar/tutorials": "Tutorials"
```
Samq64, line 5: There are also unused now?

TheColaber:
> Is the find label next to the textbox necessary? The textbox placeholder already says that.

That can be changed in a later pr, this pr should keep everything relative the same as before.

OregSamSas:
~~I think it doesn't work if language is set to a language other than English. I don't know if it's because it needs translations now, If I remember right `devtools` didn't need them, and it worked fine for me even in Hungarian.~~
(My Chrome is in Hungarian, and because loading extensions in are not just temporarily in Chrome, I used that one and found this, maybe luckily)

> ---
**Ohh, UPDATE:
Only for Hungarian doesn't it work.**
I had a look at on what about other languages, and the result was it was fine for them. I don't what's going on with HU.
> ---

 I changed the language (at first I didn't need to, because it was automatically in Hungarian), then middle clicked one of the message hat blocks, and this was the result:
image

I found out then that, there was only one event listed in the dropdown whose name was _"üzenet érkezésekor"_ which is the translation of "when I receive ..."
image

OregSamSas:
And it's not that easy. 
It works if I middle click on a stack event block, but middle clicking jumps to the "üzenet érkezésekor" event, but doesn't give any search results for it.

<html>
<details>
<summary>
screenshots
</summary>
<p><code>#1</code>:<br>

<br>
<p><code>#2</code>:<br>

</details>
<details>
<summary>
Also  the &quot;event <i>&uuml;zenet &eacute;rkez&eacute;sekor</i>&quot; always shows up in the dropdown list
</summary>

</details>
</html>

TheColaber:
@OregSamSas Fixed! This bug existed before this PR was created. Thanks for reporting.

Samq64:
I have a few suggestions:

* It would look better if the menu faded or had another animation when opening and closing it.
* I don't like the colour of highlighted items in the costumes and sounds tabs.
* When using the keyboard to scroll through items, the selected item can go offscreen. It should scroll the menu.
* When the find bar is enabled, but jump to block definition isn't, middle-clicking custom blocks should open it in the find bar like variables.

Secret-chest:
> It would look better if the menu faded or had another animation when opening and closing it.

It should roll down.

mxmou:
> * When the find bar is enabled, but jump to block definition isn't, middle-clicking custom blocks should open it in the find bar like variables.

This was already suggested here: https://github.com/ScratchAddons/ScratchAddons/issues/3700#issuecomment-1188246800

I also think this is something this PR should change.

> * It would look better if the menu faded or had another animation when opening and closing it.
> * I don't like the colour of highlighted items in the costumes and sounds tabs. Maybe something like Scratch blue in the costume editor, and pink in the sound editor would be better.
> * When using the keyboard to scroll through items, the selected item can go offscreen. It should scroll the menu.

Could you open issues for these?

> > It would look better if the menu faded or had another animation when opening and closing it.
> 
> It should roll down.

That would look weird because the popup opens behind the input, not under it.

TheColaber:
I implemented this table:
(thx lisa)
|   | **Middle-clicking non-definitions** | **Middle-clicking definitions**
| --- | --- | --- |
| **Find bar only** | Open in find bar | Open in find bar |
| **Jump to definition only** | Jump to definition | Do nothing |
| **Both** | Jump to definition | Open in find bar |

by polluting blockly which means this addon needs even more testing.

Weredime:
@TheColaber Since both `jump-to-def` and `find-bar` uses the `blockly` folder, why not put it in `libraries/cs`?

TheColaber:
No need to do so, it's fine for an addon to import other addon code like we do for the pause button and debugger.

Weredime (approved):
Tested and it works

OregSamSas:
I have a suggestion, what I tried to make, but I just wasn't enough smart to succeed. I found that the find bar stays there on the forth tab too (if the `variable-manager` is enabled), so if we could hide its own search bar and use `find-bar`'s one to jump to specific variables, this would be awesome. Because Find bar can be accessed by Ctrl+F, and it looks more good.

Below it can be checked, what I've managed to do. I couldn't complete it (the main part was very hard), because it would have needed the communication between two addon, and I couldn't use that one properly.

The special features that needs to be added that on clicking an item in the dropdown would jump to it's value input field, and on shift click it would jump into the name field, and focus on it automatically.

Maybe I should make an issue for this, but I feel it's just too late for that. I rather finish this comment.
<html>
<details>
<summary>concept</summary>

</html>

mxmou:
@OregSamSas I don't think a separate addon should be required to search the list of variables.

OregSamSas:
Ok then just hide the find bar on the variable manager tab.

TheColaber:
I think the best thing to do would be to leave the list of items in the find bar empty.

mxmou:
> I think the best thing to do would be to leave the list of items in the find bar empty.

Why does it need to be visible if you can't use it to find anything?

OregSamSas:
> @OregSamSas I don't think a separate addon should be required to search the list of variables.

It wouldn't be required, it would add just some cool features for the searching on the `variable-manager` addon, IF both addon were enabled.
Reasons: it doesn't look good that you have multiple ways to search for an item in the variable+lists list, with a find-bar which technically wouldn't work (because it would just list nothing, according to the current state of this addon declared by Colaber⁽¹⁾), and a search bar, which doesn't jump to searched items, but hides the items which didn't match with the key. I'm not totally satisfied with that, I would prefer the find bar instead.

So basically the current search bar for variable-manager would stay the same in my suggestion, just a new (new, only for that tab) find-bar would be added to options. (If  searching for variables on the 4th tab was working, there should be a setting called: "Replace variable manager's built-in search bar", which would then hide that search bar, to use the find-bar, which has/would have more conveniences.)

If there's anything I'm wrong about, just let me know, because I could imagine, I've misunderstood your response.
And so sorry for not speaking understandably, it does really annoy me too, just I can't change it sadly :( .

⁽¹⁾ Here's the quote:
> I think the best thing to do would be to leave the list of items in the find bar empty.

mxmou:
> > Why does it need to be visible if you can't use it to find anything?
> 
> Following that logic, shouldn't Scratch hide the stop button icon if the project isn't running? I think it would be distracting if the find bar was hidden each time you switch to the variables tab.

It could also be disabled, but I don't think the reason why it's disabled would be obvious.

apple502j:
> Following that logic, shouldn't Scratch hide the stop button icon if the project isn't running?

FYI, stop button isn't just for stopping threads.

> I think it would be distracting if the find bar was hidden each time you switch to the variables tab.

I think it's fine.

BroJac5246:
> I think it would be distracting if the find bar was hidden each time you switch to the variables tab.

I personally disagree, although that's just me.


>  Following that logic, shouldn't Scratch hide the stop button icon if the project isn't running?

That should be an addon [Goes to suggest it] ;)

WorldLanguages:
Note that we need griffpatch approval before merging this

Weredime:
> Note that we need griffpatch approval before merging this

@griffpatch

griffpatch:
Sure thing... Make it so.

addons/find-bar/addon.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "Editor find bar",
+  "description": "Quickly find and jump to scripts, costumes and sounds in a sprite by clicking on the new find bar located to the right of sounds tab. Use the input to filter and your mouse or arrow keys to select items. Use Ctrl + Left and Ctrl + Right in the scripting area to navigate to previous / next visited position in the code area after using the navigate to block or find bar.",
```
apple502j, line 3:
```suggestion
  "description": "Adds a find bar to find and jump to scripts, costumes and sounds next to the sounds tab. Use Ctrl + Left and Ctrl + Right in the code area to navigate to previous or next visited position after using the find bar.",
```

addons/jump-to-def/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Jump to custom block definition",
+  "description": "Using the middle mouse button or Shift + Click on a custom block allows you to jump to its definition.",
```
apple502j, line 3:
```suggestion
  "description": "Allows you to jump to a custom block's definition using the middle mouse button or Shift + Click on the block.",
```

apple502j (changes requested):

apple502j:
Migration code is broken (version setter removed, see #4916 and also some code is messed up)

apple502j (approved):

apple502j:
:shipit:

addons/editor-devtools/DevTools.js
```diff
@@ -2207,105 +1264,13 @@ export default class DevTools {
     this.costTab = guiTabs[1];
     this.costTabBody = document.querySelector("div[aria-labelledby=" + this.costTab.id + "]");
 
-    if (!document.getElementById("s3devFind")) {
-      // noinspection JSUnresolvedVariable
-      root.insertAdjacentHTML(
-        "beforeend",
-        `
-                <div id="s3devToolBar">
```
WorldLanguages, line 2215: I think you removed this without checking if other files reference this id.

This id should have been kept, it's referenced on Tab.js

PR #4827, opened 2022-07-16
discuss-button: Support dynamic enabling/disabling

Weredime:


addons/discuss-button/userscript.js
```diff
@@ -42,7 +43,14 @@ export default async function ({ addon, global, console }) {
     }
     function init() {
       removeAllItems();
-      let items = addon.settings.get("items");
+      let items = addon.self.disabled
+        ? [
+            { name: "Create", url: "/projects/editor/", extraClass: "create" },
+            { name: "Explore", url: "/explore/projects/all", extraClass: "explore" },
+            { name: "Ideas", url: "/ideas", extraClass: "ideas" },
+            { name: "About", url: "/about", extraClass: "about" },
```
TheColaber, line 51: this would need to be localized. regardless, i think the best thing to do would be to store scratch's old locales

Weredime: Resolved in https://github.com/ScratchAddons/ScratchAddons/pull/4827/commits/c9f61a5f30e577de55f6d0d0a261d522838731ab

Weredime:
Closed in favor of #4828

PR #4828, opened 2022-07-16
Even more dynamics

Weredime:
See #1673


addons/comments-linebreaks/userscript.js
```diff
@@ -1,9 +1,31 @@
...
 export default async function ({ addon, global, console }) {
+  const cache = Object.create(null);
+  addon.self.addEventListener("disabled", () => {
+    for (const comment of Object.keys(cache)) {
+      const cached = cache[comment];
+      cached.comment.innerHTML = "";
+      cached.comment.style.whiteSpace = "nowrap";
+
+      for (const elm of cached.content.childNodes) {
+        const cloned = elm.cloneNode();
```
apple502j, line 12: Is this intentionally not deep?

apple502j (changes requested):

addons/comments-linebreaks/userscript.js
Weredime: I'll fix that

apple502j (approved):

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -37,4 +37,11 @@ export default async function ({ addon, global, console }) {
     };
     return realFileReader;
   };
+  window.FileReader = pollutedFileReader;
+  addon.self.addEventListener("disabled", () => {
+    window.FileReader = originalFileReader;
+  });
+  addon.self.addEventListener("reenabled", () => {
+    window.FileReader = pollutedFileReader;
+  });
```
WorldLanguages, line 46: This is a pattern that should always be avoided.
Functions should never be unpolluted, because two addons can pollute the same object in an unpredictable order. What would happen if one of them decides to unpollute the object?
This is the only addon that pollutes `FileReader`, but it's still a bad idea.
Use an if statement in the polluted function instead.

WorldLanguages (changes requested):

TheColaber (approved):

PR #4840, opened 2022-07-17
exact-count: refactor users.js and dynamic enable/disable

Weredime:
dynamicEnable/disable for all three
users.js refactor (it was messy)

### DIFF
--- a/addons/exact-count/forum.js
+++ b/addons/exact-count/forum.js
@@ -18,7 +18,6 @@ export default async function ({ addon, global, console }) {
     const userbox = await addon.tab.waitForElement(".postleft dl", { markAsSeen: true });
     if (userbox.childNodes[6] && userbox.childNodes[6].nodeValue.includes("+")) {
       const postCountReal = Number(/(\d+)\+? posts?/.exec(userbox.childNodes[6].nodeValue)[1]);
-
       fetch("https://scratchdb.lefty.one/v3/forum/user/info/" + userbox.querySelector("a").innerText)
         .then((response) => response.json())
         .then((data) => {
--- a/addons/exact-count/forum.js
+++ b/addons/exact-count/forum.js
@@ -1,12 +1,34 @@
 export default async function ({ addon, global, console }) {
+  const cache = (window.cache = Object.create(null));
+  window.scratchAddons.CACHE = cache;
+  addon.self.addEventListener("disabled", () => {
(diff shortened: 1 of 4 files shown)


apple502j (approved):

addons/exact-count/studio.js
```diff
@@ -22,7 +22,21 @@ export default async function ({ addon }) {
     "https://api.scratch.mit.edu/studios/" + /[0-9]+/.exec(location.pathname)[0] + "/projects/?limit=40&offset=";
   const initialDelta = 100;
   const countElement = await addon.tab.waitForElement(".studio-tab-nav > a:first-child .tab-count");
+  const originalText = countElement.innerText;
+  let counted = false;
   countProjects(apiUrlPrefix, 0, initialDelta, function (count) {
-    countElement.innerText = `(${count})`;
+    counted = count;
```
WorldLanguages, line 28: Could this code be converted from callbacks to promises, since we're refactoring anyway?

Weredime: Done

addons/exact-count/forum.js
```diff
@@ -1,12 +1,32 @@
...
   while (true) {
-    let userbox = await addon.tab.waitForElement(".postleft dl", { markAsSeen: true });
+    const userbox = await addon.tab.waitForElement(".postleft dl", { markAsSeen: true });
     if (userbox.childNodes[6] && userbox.childNodes[6].nodeValue.includes("+")) {
       const postCountReal = Number(/(\d+)\+? posts?/.exec(userbox.childNodes[6].nodeValue)[1]);
       fetch("https://scratchdb.lefty.one/v3/forum/user/info/" + userbox.querySelector("a").innerText)
         .then((response) => response.json())
         .then((data) => {
           if (data.counts && data.counts.total && data.counts.total.count > postCountReal) {
+            cache[userbox.parentElement.parentElement.parentElement.parentElement.id] = {
```
WorldLanguages, line 25: Could we avoid `parentElement.parentElement`?

`userbox.closest(".blockpost")`?

WorldLanguages (approved):
LGTM
We need to be careful, if any addon also adds their own text to these elements, dynamically disabling exact-count could possibly remove text added by other addons.

addons/exact-count/forum.js
Weredime: Addressed

PR #4845, opened 2022-07-17
Clickable forum mentions

### LINKED ISSUE #1445: More more links

It's always helpful to have clickable links!

Have the more links addon linkify links in new scratchers' posts and @mentions on the forum

A separate addon, but i think its related enough

Samq64:
Resolves #1445

### Changes

Makes mentions on the forums clickable when the more links addon is enbled.

### Reason for changes

Clickable links are always nice.

### Tests

Tested on Chromium 103. Right now it only linkifies mentions that aren't in BBCode.

### DIFF
--- a/addons/more-links/addon.json
+++ b/addons/more-links/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "More links",
-  "description": "Adds links for URLs pointing outside scratch.mit.edu.",
+  "description": "Adds links for URLs pointing outside scratch.mit.edu and makes most mentions on the forums clickable.",
   "credits": [
     {
       "name": "Hans5958"
@@ -12,7 +12,8 @@
       "matches": [
         "https://scratch.mit.edu/projects/*",
         "https://scratch.mit.edu/studios/*",
-        "https://scratch.mit.edu/users/*"
+        "https://scratch.mit.edu/users/*",
+        "https://scratch.mit.edu/discuss/*"
       ]
     }
   ],
--- a/addons/more-links/userscript.js
+++ b/addons/more-links/userscript.js
@@ -1,4 +1,4 @@
-import { linkifyTextNode, linkifyTag } from "../../libraries/common/cs/fast-linkify.js";
+import { linkifyTextNode, pingifyTextNode, linkifyTag } from "../../libraries/common/cs/fast-linkify.js";
 
 export default async function ({ addon, console }) {
   const pageType = document.location.pathname.substr(1).split("/")[0];
@@ -41,10 +41,17 @@ export default async function ({ addon, console }) {
 
   (async () => {
     if (addon.tab.clientVersion === "scratchr2") {
-      while (true) {
-        let comment = await addon.tab.waitForElement(".comment .content", { markAsSeen: true });
-        // scratchr2 comment is a simple linkifyTextNode.
-        linkifyTextNode(comment);
+      if (pageType === "discuss") {
+        while (true) {
+          let post = await addon.tab.waitForElement(".post_body_html", { markAsSeen: true });
+          pingifyTextNode(post);
+        }
+      } else {
+        while (true) {
+          let comment = await addon.tab.waitForElement(".comment .content", { markAsSeen: true });
+          // scratchr2 comment is a simple linkifyTextNode.
+          linkifyTextNode(comment);
+        }
       }
     } else {
       while (true) {


RedGuy12 (approved):

Secret-chest:
This should be a separate addon, more links is for website

mxmou:
> This should be a separate addon, more links is for website

Forums are a part of the website.

Weredime (approved):

DNin01:
> This should be a separate addon, more links is for website

This new feature does seem to have a pretty different purpose from the addon it's being added to here, plus, [`curator-link`](https://github.com/ScratchAddons/ScratchAddons/tree/66f9072dd7968929a87c30d2ed8b1b430033b689/addons/curator-link) is a separate addon that also makes a username a link. I kind of agree.

DNin01 (commented):
[I see this addon differently](https://github.com/ScratchAddons/ScratchAddons/issues/4882#issue-1315478456) and I'm not sure it makes sense to add this feature to this addon. This feature might be better off in a new one instead...? This all depends on what we want this addon to be, though. Maybe all of this _is_ better when it's packed into one more general addon, but I'm not sure what to do.

WorldLanguages (dismissed):
Blocking merge

WorldLanguages:
I think this could be a separate addon.
Any objections? @mxmou @Hans5958 @RedGuy12 @Weredime

Hans5958:
I'm with Samq64's side and thinks that it is better of to keep it as one addon of `mini-htmls`.

Hans5958:
In retrospect then we should merge `curator-links` and put it as a checkbox inside `more-links`.

WorldLanguages:
> In retrospect then we should merge `curator-links` and put it as a checkbox inside `more-links`.

Yeah, if we merge these 3 addons into one, then `more-links` would need to have 3 checkboxes.

Hans5958:
Still better in my opinion due to how relatively small the addons are

Secret-chest:
Remember the `better-quoter` and `forum-quote-beautifier` merge idea? They did completely different things to quotes. This does completely different things to links.

Hans5958:
No?

`better-quoter` changes how the "Quote" button works, `forum-quote-code-beautifier` changes how code blocks are styled.

All `more-links`, this expansion, and `curator-links` have something in common: creating links from text that before are not linked, and I fail to see how it "does completely different things to links"

Samq64:
> I think this could be a separate addon.

Should I push that then?

addons/forum-mentions/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["topics"]
+    }
+  ],
+  "versionAdded": "1.18.0",
+  "tags": ["community", "forums"],
+  "enabledByDefault": false
```
Weredime, line 18:
```suggestion
  "versionAdded": "1.28.0",
  "tags": ["community", "forums"],
  "enabledByDefault": false
```

Weredime (dismissed):

addons/forum-mentions/addon.json
Samq64: 🤦‍♂️

DNin01 (dismissed):
That's perfect!

I do realize why some of us felt like it fit better in the other existing addon, but I'm feeling pretty good about having this new addon at the moment (and we can always change our minds later). Soon, we could merge [`curator-link`](https://github.com/ScratchAddons/ScratchAddons/tree/66f9072dd7968929a87c30d2ed8b1b430033b689/addons/curator-link) with this addon, too.

Hans5958:
That kind of changes (including ID renames) rarely happen as people really worried about the setting migration from version to version. That's why I wanted it to be one addon sooner than later.

RedGuy12:
I agree with hans

WorldLanguages:
Thumbs up reaction: separate addon
Thumbs down reaction: merge this and `curator-link` into `more-links`, and have 3 separate settings in that addon

A problem I see with merging them is that while `more-links` means some users might be "a click away from dangerous websites", the other 2 can only link to Scratch profiles. If we decide not to merge this with `more-links`, the 2 remaining addons could still be merged (forum mentions with `curator-link`)

BroJac5246:
> Thumbs up reaction: separate addon
> Thumbs down reaction: merge this and `curator-link` into `more-links`, and have 3 separate settings in that addon
> 
> A problem I see with merging them is that while `more-links` means some users might be "a click away from dangerous websites", the other 2 can only link to Scratch profiles. If we decide not to merge this with `more-links`, the 2 remaining addons could still be merged (forum mentions with `curator-link`)

I believe that more-links should remain as-is. It is currently exclusively for links to external websites, and I agree with DNin01's stance, expressed in #4882, that more-links should keep its function and be renamed. Linked my forum mentions should be its own addon, although, if we want, we could create a more-links counterpart for internal links, which would be created by merging this with curator-link and renaming the resulting addon.

DNin01:
I think separating the features into their own addons makes them easier to find and use.

RedGuy12:
We have a lot of addons currently, and if addons could be combined, they should be. These are very similar addons and I see no reason to not have them together.

DNin01:
I don't think we need to combine `more-links` now, but we might want to combine clickable forum mentions with `curator-link`... maybe?

DNin01:
Have we decided whether we should merge this addon with another one(s) or keep them separate, @WorldLanguages?

DNin01:
Alright, I _do_ think we should combine this and `curator-link` somehow. `more-links`... it's different. It makes _URLs_ clickable. Both `curator-link` and this addon make _usernames_ clickable. I think separating those features makes sense.

`curator-link` could turn into a general clickable username addon with two settings, one to choose whether it adds links to non-mention usernames around the website (like the front page curator and Cloud variable activity monitors), and another to choose whether it adds links to @mentions. (Or maybe just skip the settings.)

BroJac5246:
> Alright, I _do_ think we should combine this and `curator-link` somehow. `more-links`... it's different. It makes _URLs_ clickable. Both `curator-link` and this addon make _usernames_ clickable. I think separating those features makes sense.
> 
> `curator-link` could turn into a general clickable username addon with two settings, one to choose whether it adds links to non-mention usernames around the website (like the front page curator and Cloud variable activity monitors), and another to choose whether it adds links to @mentions. (Or maybe just skip the settings.)

I agree; we should merge `curator-link`and this into one addon that _does_ have settings (since, while this will be different from `more-links`, some people might only want one feature).

WorldLanguages:
> Alright, I _do_ think we should combine this and `curator-link` somehow.

I don't think anyone would be against that.
So go ahead, this "new addon" should instead be a new setting. It could be enabled by default or not. The addon should also be renamed.

At the moment, I wouldn't be able to say there's consensus in favor of merging `more-links` with other addons. So we should discuss this again in a few months.

Samq64:
curator-link id broken on master. Should I fix that on this PR?

WorldLanguages:
> `curator-link` is broken on master. Should I fix that on this PR?

Hmm, that's unfortunate.
I think you could fix it on this PR, but make sure to create an issue named "curator-link addon is broken" or something like that, so we can add it to the 1.29 milestone.

addons/curator-link/addon.json
```diff
@@ -1,16 +1,30 @@
...
+      "if": {
+        "settings": { "curator": true }
+      }
+    },
+    {
+      "url": "mentions.js",
+      "matches": ["topics"],
+      "if": {
+        "settings": { "mentions": true }
+      }
```
WorldLanguages, line 27: See #5272

Have a single userscript that checks the settings when first ran. No need to support dynamically changing the settings - just 2 separate `if()` statements and that's all.

addons/curator-link/addon.json
```diff
@@ -23,16 +37,35 @@
...
+    },
     {
       "id": "styleAsNormalText",
       "name": "Style as normal text",
       "type": "boolean",
-      "default": false
+      "default": false,
+      "if": {
+        "settings": { "curator": true }
+      }
```
WorldLanguages, line 59: Note: this usage of "if" is valid - you shouldn't change it.
Using "if" with userscripts is the only problematic situation

Samq64:
It looks like something broke in one of the commits

Samq64:
I can't read my own addon notices lol

addons/curator-link/addon.json
```diff
@@ -23,16 +34,35 @@
...
+    },
+    {
+      "id": "curator",
+      "name": "Make the cuator's username clickable",
+      "type": "boolean",
+      "default": true
+    },
     {
       "id": "styleAsNormalText",
       "name": "Style as normal text",
```
DNin01, line 51: Users might think that this means to style both the FPC's username and mentions as normal text. While this setting goes away when you have the `curator` setting disabled, that setting is enabled by default, so unless users change it, they won't know the `styleAsNormalText` setting has to do with the curator link, so we might want to have that `curator` setting disabled by default.

DNin01: Also, "cu***r***ator" is typoed.

DNin01 (commented):
This addon can probably support dynamic settings changing since it already supports dynamic enable/disable.

Also, this could apply to [Cloud variable monitor pages](https://scratch.mit.edu/cloudmonitor/409786642/).

Samq64:
Oops, thanks for fixing the typo in my typo fix.

> This addon can probably support dynamic settings changing since it already supports dynamic enable/disable.

World Languages said I didn't need to, but it wouldn't be hard to add, so I'll probably do it.

> Also, this could apply to [Cloud variable monitor pages](https://scratch.mit.edu/cloudmonitor/409786642/).

I think that could be done separately because the function I'm using relies on the @ symbol being in front of the username.

DNin01:
> > Also, this could apply to [Cloud variable monitor pages](https://scratch.mit.edu/cloudmonitor/409786642/).
> 
> I could do that but the function I'm using only linkifies usernames with the @ in front of it.

You could probably make a new one just for the Cloud variable monitors.

Samq64:
@DNin01 All settings should be fully dynamic now.

For the cloud monitor page, could that be done separately?

DNin01:
> @DNin01 All settings should be fully dynamic now.
> 
> 
> 
> For the cloud monitor page, could that be done separately?

By "separately", do you mean a separate setting, a separate script, or something else?

Samq64:
> By "separately", do you mean a separate setting, a separate script, or something else?

I mean a separate PR, so hopfuly this could get merged faster than if I started changing it even more.

DNin01 (approved):
Anyway, I tested it and it looks good to me. Doing it for the Cloud variable logs doesn't really matter too much anyway, we usually don't visit users from there, but if someone really wants that feature, maybe an issue will be opened for it someday.

WorldLanguages:
I believe this PR has some discussion on it about whether this feature should be merged with other addons.

We're probably going to add this as a separate addon, and use the upcoming "related addons"/"see also" functionality so that all addons link to each other.

DNin01 (approved):
> We're probably going to add this as a separate addon, and use the upcoming "related addons"/"see also" functionality so that all addons link to each other.

I'm torn on whether to add this to `more-links` or its own addon, but if we're linking them together, I guess this is fine.

Actually, I think I'm leaning towards a new addon, which is where we are now, but I haven't come up with a final opinion yet. These addons have good reasons to be combined and separate.

BroJac5246:
> > We're probably going to add this as a separate addon, and use the upcoming "related addons"/"see also" functionality so that all addons link to each other.
> 
> I'm torn on whether to add this to `more-links` or its own addon, but if we're linking them together, I guess this is fine.

The point of `more-links` is making more things links, which is what this does. So doesn't it make more sense for it to be part of that?

DNin01:
> > > We're probably going to add this as a separate addon, and use the upcoming "related addons"/"see also" functionality so that all addons link to each other.
> > 
> > 
> > I'm torn on whether to add this to `more-links` or its own addon, but if we're linking them together, I guess this is fine.
> 
> The point of `more-links` is making more things links, which is what this does. So doesn't it make more sense for it to be part of that?

It's just `more-links` and this addon have somewhat different use cases, and are the difference between going to an external site and going somewhere else in Scratch.

BroJac5246:
> > > > We're probably going to add this as a separate addon, and use the upcoming "related addons"/"see also" functionality so that all addons link to each other.
> > > 
> > > 
> > > I'm torn on whether to add this to `more-links` or its own addon, but if we're linking them together, I guess this is fine.
> > 
> > The point of `more-links` is making more things links, which is what this does. So doesn't it make more sense for it to be part of that?
> 
> It's just `more-links` and this addon have somewhat different use cases, and are the difference between going to an external site and going somewhere else in Scratch.

We need a solution for stuff like this because it has come up in the past...

Good point, though, I agree now.

Joeclinton1:
Is there a reason you identified bbcode as an edge case then just didn't solve it? Also perhaps this addon could be made larger into an addon about making all mentions on all parts clickable, following #7301 and #7364.

Samq64:
> Is there a reason you identified bbcode as an edge case then just didn't solve it?

The library doesn't support it. I'll look into it.

> Also perhaps this addon could be made larger into an addon about making all mentions on all parts clickable, following https://github.com/ScratchAddons/ScratchAddons/issues/7301 and https://github.com/ScratchAddons/ScratchAddons/pull/7364.

I agree. External links should be one addon and usernames another, but there there's never been a conclusion on that.

Samq64:
I've added BBCode support but dynamic re-enable isn't handled very well and running `querySelectorAll`twice isn't ideal.

WorldLanguages:
I strongly suggest we consider #7364 to be part of this addon as they both make UI elements clickable and they also only affect the forums.

addons/linkify-mentions/addon.json
```diff
@@ -0,0 +1,44 @@
...
+      "if": {
+        "settings": {
+          "plainText": true
+        }
+      }
+    }
+  ],
+  "settings": [
+    {
+      "name": "Style quote authors as normal text",
```
Samq64, line 33: Should this also apply to mentions?

rickbot-dot: > Should this also apply to mentions?

Nah, it would be even more confusing whether it was text or a link if it was part of an actual sentence.

Samq64:
@WorldLanguages Do you are still need to be blocking this?

PR #4849, opened 2022-07-18
Remove `displayNoneWhileDisabled`

mxmou:
See #4825

### Changes

Removes `addon.tab.displayNoneWhileDisabled` and replaces it with a userstyle-based solution. Also makes ocular react to setting changes and makes the search-profile tab grayed out instead of red when disabled.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -234,18 +234,6 @@ export default class Tab extends Listenable {
   }
 
   /**
-   * Hides an element when the addon is disabled.
-   * @param {HTMLElement} el - the element.
-   * @param {object=} opts - the options.
-   * @param {string=} opts.display - the fallback value for CSS display.
-   */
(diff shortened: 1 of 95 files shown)


WorldLanguages:
I'll extend discussion a bit more (#4825)

WorldLanguages:
Closing since we decided to use the idea described on https://github.com/ScratchAddons/ScratchAddons/discussions/4825#discussioncomment-3176220 instead.

@mxmou please say whether you want to send a PR implementation of the new idea or if anyone else should take it

mxmou:
> Closing since we decided to use the idea described on [#4825 (reply in thread)](https://github.com/ScratchAddons/ScratchAddons/discussions/4825#discussioncomment-3176220) instead.
> 
> @mxmou please say whether you want to send a PR implementation of the new idea or if anyone else should take it

I probably won't work on that.

PR #4851, opened 2022-07-18
Fix missing `reduxEvents`

### LINKED ISSUE #4829: Paint editor looses darkmode when language changes
> Really good, the only bad thing is that when I switch languages the costume editor background becomes white

Go to project
Turn on editor-dark
Go to the paint tab
Change language
Notice bug

cc @mxmou

mxmou:
Resolves #4829

### Changes

Fixes two bugs that would happen when changing the language:
* the costume editor would lose its dark background if dark mode was enabled
* the onion skin settings would disappear


TheColaber (approved):

WorldLanguages (changes requested):
We should also consider File>Load (after getting `Replace contents of the current project?` prompt)

WorldLanguages (approved):

Weredime (approved):

PR #4854, opened 2022-07-18
`editor-buttons-reverse-order`: add `theme` tag

lisa-wolfgang:
### Changes

Adds the theme tag to the `editor-buttons-reverse-order` (reverse order of project controls) addon.

### Reason for changes

The addon only makes visual changes and doesn't provide in-editor functionality to toggle or change them; therefore, it is a theme and should be labelled as such.

### Tests

Works on browsers with Chromium 103.

image


lisa-wolfgang:
@WorldLanguages Do you have a specific uncertainty you'd like @mxmou to address?

Issue #4857, opened 2022-07-19
Hide stage when running bug


Coolguy236gg:
-->

### Description
So, when I click the green flag and then hide the stage (using the new addon ofc) it freezes the entire webpage.
**NOTE:** This bug only happens when you start a big project containing loads of code, the **results may vary according to the power of your device.**

### Reproduction steps

1. Click the green flag of a project (**make sure it is a big project like paper Minecraft or something** or you will not have enough time to reproduce the bug.) The better your device is, you'll have to choose bigger projects.
2. Hide the stage using the new button obtained by enabling the "hide stage" addon.
3. If you have followed these steps correctly, your tab should crash.

### Expected behavior
The project should be **stopped** and then only it should close the stage.

### Actual behavior
The project **does not stop**, and the stage hides. I guess it probably uses up a lot of resources of the computer, which is why it freezes the tab.

### Screenshots
_Screenshots will not make sense_, so I will send a video
The video is a bit corrupted (idk y) so it has a black screen for a few seconds. Also, it says veed.io and it is editing software. I used it to compress the video, not edit it.

https://user-images.githubusercontent.com/109452958/179736237-a2ef8186-2836-4306-8e53-4bfdb52c90ad.mp4

### Environment

- Browser: Brave v1.41.96 (based on chromium)
- Scratch Addons version: v1.27.1
- Operating system: Windows 11 (_with extra tweaks to improve performance_)

### Additional context

I don't have anything to say here. It is kind of _self-explanatory._

TheColaber:
So apparently putting `display: none` on `stage-wrapper_stage-canvas-wrapper` causes the page to crash. Either there is a CSS fix or we should stop the project when this happens.

Coolguy236gg:
_Idk about actual coding_, (I code only from scratch). I hope it is easy to fix!

Samq64:
I can't reproduce the bug. I tested on Firefox with Paper Minecraft while in the processor sprite which has 4000+ blocks and hiding the stage lagged a tiny bit, but less than dragging a stack of blocks in that sprite.

GarboMuffin:
> So apparently putting display: none on stage-wrapper_stage-canvas-wrapper causes the page to crash. Either there is a CSS fix or we should stop the project when this happens.

Have you reproduced this or is this just a guess?

If you have reproduced it, use your browser's profiler to figure out which function it's getting stuck in.

lisa-wolfgang:
[...]

PR #4861, opened 2022-07-19
hide-stage: Fix renderer crashes

GarboMuffin:
A hidden stage is interpreted as 0x0 in some parts of the renderer which can cause infinite loops or crashes.

 - Fixes https://github.com/ScratchAddons/ScratchAddons/issues/4857 (infinite loop in "touching mouse pointer" block)
 - Hides code editor eye droppers when the stage is hidden because the user has no way to hover over the stage anyways. Previously attempting to use these would crash the editor.

### DIFF
--- a/addons/hide-stage/style.css
+++ b/addons/hide-stage/style.css
@@ -40,7 +40,7 @@
   align-items: center;
 }
 
-[dir="rtl"] .sa-stage-hidden [class*="stage-header_stage-size-row"] {
+.sa-stage-hidden [dir="rtl"] [class*="stage-header_stage-size-row"] {
   right: auto;
   left: 0.5rem;
 }
--- a/addons/hide-stage/style.css
+++ b/addons/hide-stage/style.css
@@ -52,3 +52,7 @@
   /* makes small and large stage buttons appear unselected */
   filter: var(--editorDarkMode-accent-desaturateFilter, saturate(0));
 }
+
+.sa-stage-hidden .scratchEyedropper {
+  display: none;
+}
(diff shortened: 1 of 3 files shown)


Weredime (approved):
Tested and perfectly works!

addons/hide-stage/style.css
```diff
@@ -52,3 +52,7 @@
   /* makes small and large stage buttons appear unselected */
   filter: var(--editorDarkMode-accent-desaturateFilter, saturate(0));
 }
+
+.sa-stage-hidden .scratchEyedropper {
+  display: none;
+}
```
lisa-wolfgang, line 58: We don't want to repeat our mistakes here. Scratch doesn't assume that the eyedropper is showing, does it?

WorldLanguages: This bug happens because Scratch uses the stage's width and height for calculations. I don't think Scratch ever calls `clientWidth` on the eye dropper element.

addons/hide-stage/userscript.js
```diff
@@ -1,34 +1,76 @@
...
-  let bodyWrapper;
   let smallStageButton;
   let largeStageButton;
   let hideStageButton;
 
   function hideStage() {
     stageHidden = true;
-    if (!bodyWrapper) return;
-    bodyWrapper.classList.add("sa-stage-hidden");
+    document.documentElement.classList.add("sa-stage-hidden");
```
WorldLanguages, line 16: We could use the body instead AFAIK. Not necessary to change it tho, root element is fine too

addons/hide-stage/style.css
GarboMuffin: Shouldn't be an issue

WorldLanguages (approved):

Issue #4866, opened 2022-07-19
Bug fix on v1.27.2 hides the stage on player mode


WorldLanguages:
Regressed by #4861 
However, the bug fixed by that PR was a crash, and this is lower prio. So I still think it was worth it to release that fix ASAP.
Unlike the bug fixed by the PR that introduced this bug, this bug does not lead to loss of unsaved data.

Repro:
1. Go inside any project
2. Hide the stage with the "hide the stage and sprite pane" addon
3. See project page, the stage is hidden

Workaround: go inside the project, disable "hide stage" mode, then go to the project page again

PR #4867, opened 2022-07-20
hide-stage: Fix hidden stage on project page regression and other errors

GarboMuffin:
- Fixes https://github.com/ScratchAddons/ScratchAddons/issues/4866. sa-stage-hidden is now once again applied to the body wrapper, not the root element. A different class is now applied to the root element to style something that isn't inside the body wrapper.
 - Fixes an editor crash when rendering a text bubble after entering and leaving fullscreen with the stage hidden

### DIFF
--- a/addons/editor-stage-left/stageleft.css
+++ b/addons/editor-stage-left/stageleft.css
@@ -66,7 +66,7 @@
   justify-content: flex-end;
 }
 
-.sa-stage-hidden [dir="rtl"] [class*="stage-header_stage-size-row"] {
+[dir="rtl"] .sa-stage-hidden [class*="stage-header_stage-size-row"] {
   left: auto;
   right: 0.5rem;
 }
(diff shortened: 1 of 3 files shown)


mxmou (approved):

GarboMuffin:
Found another crash I believe related to this addon when leaving fullscreen. Will look at that in the morning.

Weredime (approved):

GarboMuffin:
So there is another way to do this that is actually probably better

Instead of all these renderer traps, the `display: none` can be replaced with `position: absolute; z-index: -9999999` which hides it in a way that clientWidth and clientHeight aren't 0.

WorldLanguages:
Requesting review again because this PR now uses a different method to hide the stage.

WorldLanguages (approved):
Tested

WorldLanguages:
Here's a diff between v1.27.1 and this PR: https://github.com/ScratchAddons/ScratchAddons/compare/v1.27.1..a94a371#diff-e9250ac7772b00c94ceb300c763f7664acfa95559f6dbfe8c1094729cde182f6

Weredime (approved):

PR #4873, opened 2022-07-20
New addon: Show password button

valmontechno:
I added an addon that Add a button in the login form that displays the password.


apple502j:
This is more like a generic improvement applicable for all websites; therefore, rejected.

PR #4881, opened 2022-07-22
New addon: Alternating nested block colors

### LINKED ISSUE #2742: Zebra Striping nested blocks from the same category

CST1229:
Resolves #2742

After, like, 4 months of the branch existing, it's here.

### Changes

Adds the Alternating nested block colors (`zebra-striping`) addon, which makes nested blocks alternate between their normal colors and a different shade, like zebra striping in Snap<i>!</i>. The intensity of the striping is adjustable, from -100% (black) to 0% (no change) to 100% (basically white), with the default being 20%.

Also, this is not #3143 again, ***I*** (:P) (cst edit: colaber basically rewrote the whole addon lol) wrote this from scratch and it should work with other addons.

### Reason for changes

Some projects can have large C block or reporter stacks that aren't very navigatable when everything blends together. This addon might be able to help.

### Tests - outdated - needs to be retested

Mostly tested on Firefox, with very light basic functionality testing on Chrome.

Seems to be compatible with editor-theme3 and custom-block-shape, as well as both Scratch's and editor-stepping's block glows.

Seeing inside and switching tabs and sprites has been tested.


Samq64:
I can't seem to get it working on Chromium...

WorldLanguages:
Let's keep in mind the problems we found previously on #3143

WorldLanguages:
> I can't seem to get it working on Chromium. The console tells me `editingTarget` is undefined, so you probably just need to wait for the workspace to load before zebra-striping anything.

This is an easy fix, many addons need to wait until vm.editingTarget exists already.
Just search for `if (vm.editingTarget) return resolve();` on other addons and copypaste into the top of the file.

CST1229:
> > I can't seem to get it working on Chromium. The console tells me `editingTarget` is undefined, so you probably just need to wait for the workspace to load before zebra-striping anything.
> 
> This is an easy fix, many addons need to wait until vm.editingTarget exists already. Just search for `if (vm.editingTarget) return resolve();` on other addons and copypaste into the top of the file.

I have a (currently unpushed) fix where I instead enable runAtComplete on the userscript and check for vm and vm.editingTarget at the initial striping.

WorldLanguages:
@CST1229 Any updates?

CST1229:
> @CST1229 Any updates?

yes

WorldLanguages:
@CST1229 Any known bugs? Why is this a draft?

CST1229:
> @CST1229 Any known bugs? Why is this a draft?

I'll let this GIF explain:

TheColaber:
Bugs fixed, read the original post, this should be ready to merge soon.

addons/zebra-striping/userscript.js
```diff
@@ -0,0 +1,46 @@
+export default async function ({ addon, msg, global, console }) {
+  const vm = addon.tab.traps.vm;
+  const scratchBlocks = await addon.tab.traps.getBlockly();
+  const _render = scratchBlocks.BlockSvg.prototype.render;
+  scratchBlocks.BlockSvg.prototype.render = function (opt_bubble) {
+    if (!this.isInFlyout && !this.isShadow_) {
+      let block = this;
+      while (block.getSurroundParent() && block.getSurroundParent().category_ === block.category_) {
```
CST1229, line 8:
```suggestion
      while (block.getSurroundParent() && block.getSurroundParent().getCategory() === block.getCategory()) {
```

addons/zebra-striping/userscript.js
```diff
@@ -0,0 +1,46 @@
+export default async function ({ addon, msg, global, console }) {
+  const vm = addon.tab.traps.vm;
+  const scratchBlocks = await addon.tab.traps.getBlockly();
+  const _render = scratchBlocks.BlockSvg.prototype.render;
+  scratchBlocks.BlockSvg.prototype.render = function (opt_bubble) {
+    if (!this.isInFlyout && !this.isShadow_) {
```
CST1229, line 6:
```suggestion
    if (!this.isInFlyout && !this.isShadow()) {
```

CST1229: actually nvm i'll just do it myself

addons/zebra-striping/addon.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/dist/schema.json",
+  "name": "Alternating nested block colors",
+  "description": "Makes blocks of the same category alternate between lighter and darker shades when nested inside eachother.",
```
mxmou, line 4: "Zebra stripes" should be mentioned somewhere in the addon name or description because people will probably search for that.

OregSamSas: @mxmou I think it already finds it at searching for "zebra stripes", because of the addon ID. 
(I'm not sure though that you meant this.)
image

addons/zebra-striping/addon.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "Alternating nested block colors",
+  "description": "Makes blocks of the same category alternate between lighter and darker shades when nested inside each other.",
+  "credits": [
+    {
+      "name": "CST1229",
+      "link": "https://github.com/CST1229"
```
GarboMuffin, line 7: Just so you're aware, in TurboWarp will be replaced with a link to your Scratch profile instead

GarboMuffin (approved):

addons/zebra-striping/addon.json
mxmou: The addon ID is a technical detail. Yes, it's possible to search by ID, but that's mainly intended for developers.

OregSamSas: > Yes, it's possible to search by ID,...

Is it just possible or the only option as well? Because I think the search engine always look at the addon ID too in any search cases.

mxmou: It does but I don't think it should have to just to show relevant results.

OregSamSas: > It does but I don't think it should have to just to show relevant results.

Why do you think that? I don't see any issues (mostly meaning confusions) it could cause.

OregSamSas: Sorry for debating, I'm just curious about your reasons.

mxmou: Addon IDs are not translated and (in the stable version) never shown to the user.

OregSamSas: Oh ok, you are totally right. But I don't think that someone will search for zebra-stripes in another language because it's not translated anywhere. So in this very case it doesn't matter that much, as I think, but in general it would be logical.

TheColaber (approved):

addons/zebra-striping/addon.json
```diff
@@ -0,0 +1,44 @@
...
+    {
+      "name": "GarboMuffin",
+      "link": "https://scratch.mit.edu/users/GarboMuffin/"
+    },
+    {
+      "name": "TheColaber",
+      "link": "https://scratch.mit.edu/users/thecolaber"
+    }
+  ],
+  "tags": ["editor", "codeEditor"],
```
WorldLanguages, line 18: Add `featured` tag

addons/zebra-striping/addon.json
```diff
@@ -0,0 +1,44 @@
...
+  "name": "Alternating nested block colors",
+  "description": "Makes blocks of the same category alternate between lighter and darker shades when nested inside each other. This is also known as zebra striping.",
+  "credits": [
+    {
+      "name": "CST1229",
+      "link": "https://github.com/CST1229"
+    },
+    {
+      "name": "GarboMuffin",
+      "link": "https://scratch.mit.edu/users/GarboMuffin/"
```
WorldLanguages, line 11: Do not link to Scratch profile, be consistent with other addons
No link

addons/zebra-striping/addon.json
TheColaber:
```suggestion
  "tags": ["editor", "codeEditor", "featured"],
```

WorldLanguages (changes requested):
We'll split the "intensity" setting into 2, to avoid negative numbers and offer better UX
I'll work on this

Secret-chest:
> We'll split the "intensity" setting into 2, to avoid negative numbers and offer better UX
> I'll work on this

Why do you always want to remove customizability?

GarboMuffin:
The proposal does not remove customizability, rather it allows the same level of control in a less confusing interface.

Secret-chest:
> The proposal does not remove customizability, rather it allows the same level of control in a less confusing interface.

How?

mxmou:
> > The proposal does not remove customizability, rather it allows the same level of control in a less confusing interface.
> 
> How?

Probably one setting to choose between brighter and darker and another to choose the amount.

addons/zebra-striping/addon.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "Alternating nested block colors",
+  "description": "Makes blocks of the same category alternate between lighter and darker shades when nested inside each other. This is also known as zebra striping.",
```
OregSamSas, line 3: > This is also known as zebra striping.

Is "this is" neccessary? Because I feel in a short description, it sounds better to leave that out, it's obvious what does it refer to. 
((For real, this is a very minor thing, just wanted to ask in the purpose of learning and hearing other opinions.))

WorldLanguages: It could be considered unnecessary, but I don't think it matters much.

PR #4887, opened 2022-07-23
New addon: Bottom navigation bar

valmontechno:
Positions the navigation bar at the bottom of the page.
I fixed the bugs from the previous pull request.

### DIFF
--- a/addons/bottom-navigation-bar/addon.json
+++ b/addons/bottom-navigation-bar/addon.json
@@ -1,0 +1,27 @@
+{
+  "name": "Bottom navigation bar",
+  "description": "Positions the navigation bar at the bottom of the page.",
+  "tags": ["community", "theme"],
+  "versionAdded": "1.27.2",
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["https://scratch.mit.edu/*"]
+    },
+    {
+      "url": "infinite-scroll.css",
+      "matches": ["https://scratch.mit.edu/*"],
+      "if": {
+        "addonEnabled": ["infinite-scroll"]
+      }
+    }
+  ],
+  "dynamicEnable": true,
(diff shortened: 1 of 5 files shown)


addons/bottom-navigation-bar/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Bottom navigation bar",
+  "description": "Positions the navigation bar at the bottom of the page.",
+  "tags": ["community", "theme"],
+  "versionAdded": "1.27.2",
```
Norbiros, line 5:
```suggestion
  "versionAdded": "1.28.0",

```
We are on 1.28.0-pre.

addons/bottom-navigation-bar/addon.json
```diff
@@ -0,0 +1,27 @@
...
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["https://scratch.mit.edu/*"]
+    },
+    {
+      "url": "infinite-scroll.css",
+      "matches": ["https://scratch.mit.edu/*"],
+      "if": {
+        "addonEnabled": ["infinite-scroll"]
```
Norbiros, line 15: It doesn't work when you have `infinite-scroll` addon enabled, but you unselected fix footer setting.

Norbiros: Also, you have this ugly blue bar at bottom of page
Screenshot from 2022-07-23 20-21-09

Norbiros (changes requested):
1. Without scratchr2, but with horizontal my stuff:
Screenshot from 2022-07-23 20-23-08
2. My Stuff without scratchr2
bug
These are bugs that I found in 5 minutes of research, so probably this addon has way more...

PR #4892, opened 2022-07-24
forum-time-zones: use mobile forums instead of scratchdb

Weredime:
### Changes
Leaves the ScratchDB deprecated v2 APIs and get UTC dates using the mobile forums
Also fixes bug when you're on a big topic and you're on the first page (we don't specify a "page" to scratchdb)
And dynamic enable/disable

### Reason for changes
Less bytes sent to client

### Tests
Firefox

### DIFF
--- a/addons/forum-time-zones/userscript.js
+++ b/addons/forum-time-zones/userscript.js
@@ -34,7 +34,7 @@ export default async function ({ addon, global, console }) {
       }
     }
   };
-  const pageNumber = /page=(\d+)/.exec(location.search)?.[1];
+  const pageNumber = new URLSearchParams(location.search).get("page");
 
   addon.self.addEventListener("disabled", () => {
     for (const c of cache) {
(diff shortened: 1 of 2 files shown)


WorldLanguages (changes requested):
Doesn't work for me, it's showing UTC+0 (my timezone is UTC-3)
Did you forget to convert it to the user's time zone?

addons/forum-time-zones/userscript.js
```diff
@@ -34,21 +34,43 @@ export default async function ({ addon, _global, _console }) {
       }
     }
   };
-  window
-    .fetch(`https://scratchdb.lefty.one/v2/forum/topic/${forum_topic_id}`)
+  const pageNumber = /page=(\d+)/.exec(location.search)?.[1];
```
WorldLanguages, line 37: `new URLSearchParams(location.search).get("page")` would be cleaner

WorldLanguages (approved):

apple502j (approved):

PR #4909, opened 2022-07-27
Fix incorrect add extension button corner when using editor-stage-left and columns

GarboMuffin:
When both editor-stage-left and columns were enabled, the add extension button's bottom left corner (or bottom right in RTL) would be more rounded than the other corners.

Before

image

After

image

### DIFF
--- a/addons/columns/style.css
+++ b/addons/columns/style.css
@@ -49,7 +49,8 @@
   height: calc(1.75rem - 2px);
   background-color: transparent;
   border-color: var(--editorDarkMode-border, rgba(0, 0, 0, 0.15));
-  border-radius: 0.25rem;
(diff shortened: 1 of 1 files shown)


mxmou (commented):
Since the styles this overrides don't have `!important`, I think it's better to use more specific selectors:

```css
[class*="gui_extension-button-container_"] {
  ...
}

/* [dir="ltr"] is for specificity to override editor-stage-left */
[dir="ltr"] [class*="gui_extension-button-container_"],
[dir="rtl"] [class*="gui_extension-button-container_"] {
  border-radius: 0.25rem;
}
```

mxmou (approved):

WorldLanguages (approved):

PR #4916, opened 2022-07-28
Customizable block palette opacity

mxmou:
### Changes

Allows the opacity of the block palette to be changed using editor dark mode. All presets use the default 80%. Migration code will be needed to add transparency to existing settings.

### Reason for changes

I don't think there is a good reason to not make it customizable.


WorldLanguages:
Good idea! This should be customizable

WorldLanguages:
@apple502j Please review my changes to get-addon-settings.js
Is it safe to remove the old transition code? Or do we still need to set `addonSettings["editor-dark-mode"]._version = 2` if it's not yet defined?

background/get-addon-settings.js
```diff
@@ -59,57 +59,11 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
...
-            palette: "#333333",
-            fullscreen: "#282828",
-            stageHeader: "#333333",
-            border: "#444444",
-          },
-          2
-        );
+        // Transition v1.27 to v1.28
+        if (settings.palette !== undefined && settings.palette.length === 7) {
+          settings.palette += "cc";
```
TheColaber, line 64: Could we have a comment to describe this, because appending "cc" to the string isn't very clear here? it makes sense in the context of this pr but not for someone looking at this file by itself.

apple502j:
@WorldLanguages Read the comment on that file

apple502j (changes requested):
Migration needs comments + version

apple502j:
> Or do we still need to set `addonSettings["editor-dark-mode"]._version = 2` if it's not yet defined?

Yes

apple502j:
#4817 changed the transition code to better handle no-op transition.

apple502j (approved):

TheColaber (approved):

PR #4930, opened 2022-07-31
Fix issues from the pre-release color PR

apple502j:
plus the toggle thingy, it can stay

Also, re-adds favicon to fix Firefox missing favicons. Also fixed favicon on fullscreen messaging


mxmou:
There is stil an issue with focused inputs:
image

Samq64:
Oops, I should have tested on FF.

> There is stil an issue with focused inputs: image

I'll fix that.

PR #4810, opened 2022-07-14
Different primary colour for pre-release versions

Samq64:
Resolves #4805

### Changes

Adds `-pre` to the popup version (there wasn't enough space for the full version name) and automatically turns the logo and primary color blue on pre-release versions. The logo shown of the extensions page has to be manually updated.

### Reason for changes

It makes it easier to identify pre-release versions of the extension.

### Tests

Tested on Chromium 103.

Screenshot from 2022-07-20 08-06-26

PR #4934, opened 2022-08-01
Fix a couple bugs from #4810

Samq64:
Fixes https://github.com/ScratchAddons/ScratchAddons/pull/4930#issuecomment-1200446217 and incorrect icon preload path. Tested on Chromium 103.

### DIFF
--- a/libraries/common/global-theme.js
+++ b/libraries/common/global-theme.js
@@ -3,7 +3,6 @@ export default function () {
   if (prerelease) {
     const blue = getComputedStyle(document.documentElement).getPropertyValue("--blue");
     document.documentElement.style.setProperty("--brand-orange", blue);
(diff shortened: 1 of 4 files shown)


mxmou:
Ideally the solution would be consistent with buttons, which #4930 changed back to orange.

Samq64:
> Ideally the solution would be consistent with buttons, which #4930 changed back to orange.

They're now consistently orange.

OregSamSas:
I think it's better to have it as blue in beta versions, but then it seems I'm the only one who's thinking like this.

apple502j (approved):

mxmou:
> I think it's better to have it as blue in beta versions, but then it seems I'm the only one who's thinking like this.

I don't think the exact color is very important, it just has to be the same for inputs and buttons.

mxmou (approved):

PR #4945, opened 2022-08-03
variable-manager: Validate variable names

### LINKED ISSUE #4941: variable-manager doesn't validate variable names properly
Repro: variable names can be set to empty, see video https://youtu.be/rLZ41UbCgqk?t=27

Weredime:
Resolves #4941

### Changes
Validate cloud variables names

### DIFF
--- a/addons/variable-manager/userscript.js
+++ b/addons/variable-manager/userscript.js
@@ -145,26 +145,37 @@ export default async function ({ addon, global, console, msg }) {
       const onLabelOut = (e) => {
         e.preventDefault();
         const workspace = Blockly.getMainWorkspace();
-        const existingVariableWithNewName = workspace.getVariable(label.value, this.scratchVariable.type);
-        if (existingVariableWithNewName) {
+
+        let newName = label.value;
+        if (newName === this.scratchVariable.name) {
+          // If the name is unchanged before we make sure the cloud prefix exists, there's nothing to do.
+          return;
+        }
+
+        const CLOUD_SYMBOL = "☁";
+        const CLOUD_PREFIX = CLOUD_SYMBOL + " ";
+        if (this.scratchVariable.isCloud) {
+          if (newName.startsWith(CLOUD_SYMBOL)) {
+            if (!newName.startsWith(CLOUD_PREFIX)) {
+              // There isn't a space between the cloud symbol and the name, so add one.
+              newName = newName.substring(0, 1) + " " + newName.substring(1);
+            }
+          } else {
+            newName = CLOUD_PREFIX + newName;
+          }
+        }
+
+        const isEmpty = !newName.trim();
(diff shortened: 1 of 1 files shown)


Chameleon-Coder (approved):
Lookin' good!

addons/variable-manager/userscript.js
```diff
@@ -148,6 +149,19 @@ export default async function ({ addon, global, console, msg }) {
         if (existingVariableWithNewName) {
           label.value = this.scratchVariable.name;
         } else {
+          if (
+            !Blockly.Variables.nameValidator_.call(
+              null,
```
apple502j, line 154: Hmm, what's the null here?

GarboMuffin (changes requested):
While digging into nameValidator_ and other testing I found you can remove the cloud prefix from cloud variables. Here's my proposal that fixes this issue and the one in the PR since they're in the same place:

```js
      const onLabelOut = (e) => {
        e.preventDefault();
        const workspace = Blockly.getMainWorkspace();

        let newName = label.value;
        if (newName === this.scratchVariable.name) {
          // If the name is unchanged before we make sure the cloud prefix exists, there's nothing to do.
          return;
        }

        const CLOUD_PREFIX = '☁ ';
        if (this.scratchVariable.isCloud && !newName.startsWith(CLOUD_PREFIX)) {
          newName = CLOUD_PREFIX + newName;
        }

        const isEmpty = !newName.trim();
        const nameAlreadyUsed = !!workspace.getVariable(newName, this.scratchVariable.type);
        if (isEmpty || nameAlreadyUsed) {
          label.value = this.scratchVariable.name;
        } else {
          workspace.renameVariableById(this.scratchVariable.id, newName);
          // Only update the input's value when we need to to avoid resetting undo history.
          if (label.value !== newName) {
            label.value = newName;
          }
        }
      };
```

Lightly tested.

I don't see a convincing reason to use getBlockly() here. It'll just be more complicated and might display incorrect prompts about variables already existing in a few edge cases involving cloud variables without some roundabout solutions.

addons/variable-manager/userscript.js
Weredime: See https://github.com/LLK/scratch-blocks/blob/893c7e7ad5bfb416eaed75d9a1c93bdce84e36ab/core/variables.js#L297

apple502j (approved):

GarboMuffin (approved):

addons/variable-manager/userscript.js
```diff
@@ -1,5 +1,6 @@
 export default async function ({ addon, global, console, msg }) {
   const vm = addon.tab.traps.vm;
+  const Blockly = await addon.tab.traps.getBlockly();
```
GarboMuffin, line 3: not needed anymore

PR #4949, opened 2022-08-04
Show error message if an addon.json is invalid

### LINKED ISSUE #4947: Better error messages

If there is an syntax error, for example, a missing comma in the `addon.json`, the popup and settings pages don't load and it only lets the user know in the background page console.

Throw better error messages and maybe display something in the popup too.

Skip addons with invalid manifests.

Weredime:
Resolves #4947

### Changes
Stops crashing the manifest loader if an addon manifest is invalid, and error in background page

### Tests
Firefox

### DIFF
--- a/background/load-addon-manifests.js
+++ b/background/load-addon-manifests.js
@@ -42,7 +42,13 @@ const localizeSettings = (addonId, setting, tableId) => {
   const useDefault = scratchAddons.l10n.locale.startsWith("en");
   for (const addonId of addonIds) {
     if (addonId.startsWith("//")) continue;
-    const manifest = await (await fetch(`/addons/${addonId}/addon.json`)).json();
+    let manifest;
+    try {
+      manifest = await (await fetch(`/addons/${addonId}/addon.json`)).json();
+    } catch (ex) {
+      console.error(`Failed to load addon manifest for ${addonId}, skipping to next addon:`, ex);
+      continue;
+    }
     let potentiallyNeedsMissingDynamicWarning =
       manifest.updateUserstylesOnSettingsChange && !(manifest.dynamicEnable && manifest.dynamicDisable);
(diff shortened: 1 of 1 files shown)


apple502j:
I believe addon.json being invalid should crash. Crash handling is bad, I can say, but it is still a crash and no proper recovery is expected.

BroJac5246:
This works on Chrome 103. However, it presents a problem:
You have to manually check for errors.

With the method SA uses, the addons not loading is the error report in itself. It's annoying, but much faster than your implementation.
What I would _personally_ prefer is if the settings & settings popup showed an error message, either as an HTML element, or with `alert()` to let the user (who, in such a case will also be a developer), know that the addon being loaded had an error in the manifest, and will be skipped.

Example:
`alert("Addon \"" + addonId + "\" has an error in its manifest, this addon will not be loaded.");`

WorldLanguages:
IMO it should throw and run
```chrome.tabs.create(`data:text/plain,Scratch Addons crashed: invalid addon.json for addon with id ${addonId}`)```

BroJac5246:
> IMO it should throw and run
> ```chrome.tabs.create(`data:text/plain,Scratch Addons crashed: invalid addon.json for addon with id ${addonId}`)```

That would work too.

RedGuy12:
I think it should catch the manifest error then rethrow with the addon id added to the message

TheColaber (approved):

TheColaber:
Actually, the tab didn't open when I tested it in Firefox. @WorldLanguages Know why?

apple502j (changes requested):
Blocking, will investigate Firefox

apple502j (approved):
It's fine to make this Chrome-only, I suppose. Firefox is for the devs. (MDN cites "security reasons").

PR #4950, opened 2022-08-04
Don't reload already-loaded scripts

Weredime:
## Changes
Stop duplicate scripts from running, e.g color-picker and 2d-color-picker use tinycolor, and if you use both at the same time it loads the same script two times

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -61,8 +61,12 @@ export default class Tab extends Listenable {
         } else if (obj.error === null) {
           // Script has been appended to document.head, but not loaded yet.
           obj.script.addEventListener("load", (e) => {
-            obj.loaded = true;
-            resolve(e);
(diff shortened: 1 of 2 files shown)


addon-api/content-script/Tab.js
```diff
@@ -53,11 +53,40 @@ export default class Tab extends Listenable {
...
-      script.onload = resolve;
-      script.onerror = reject;
+      if (scratchAddons.loadedScripts[url]) {
+        const obj = scratchAddons.loadedScripts[url];
+        if (obj.loaded) {
+          // Script has been already loaded
+          resolve();
+        } else if (obj.error === null) {
+          // Script has been appended to document.head, but not loaded yet.
+          obj.script.addEventListener("load", (e) => {
```
apple502j, line 63: Won't this wait indefinitely if this errors?

Weredime: Fixed

apple502j (approved):

WorldLanguages (approved):

PR #4965, opened 2022-08-08
Get textContent instead of innerHTML (forum-search)

WorldLanguages:
Resolves bug from feedback:
> There's a new forum now ( https://scratch.mit.edu/discuss/60/ ) called "Project Save & Level Codes", which has an ampersand in it.
When the forum search addon is enabled and you visit this page, then the placeholder text says "Project Save \&amp; Level Codes" instead of the expected "Project Save & Level Codes".

### DIFF
--- a/addons/forum-search/userscript.js
+++ b/addons/forum-search/userscript.js
@@ -260,7 +260,7 @@ export default async function ({ addon, global, console, msg }) {
         break;
       }
       case 4: {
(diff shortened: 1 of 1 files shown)


apple502j:
Guess we should ban reading innerHTML?

apple502j (approved):

TheColaber (approved):

PR #4966, opened 2022-08-09
studio-tools: update redux instead of reload take two

Weredime:
### Changes
Update redux instead of reloading whole page

### Reason for changes
Why need a whole reload
### Tests
Firefox 105.
I've checked thoroughly for crashes, and there seems to be none present (I fixed the bug from #3230, [bug](https://scratch.mit.edu/discuss/topic/543336/))


WorldLanguages:
@Weredime Could you re-review your own PR? It's been a while., maybe you can find some issues with your own code. And we can try to merge this for 1.30

PR #4972, opened 2022-08-13
Fix one in a million race condition

TheColaber:
So apparently an if statement was too long. Previously, nothing would happen if you tried to enable, disable, or change settings. But this is really hard to repro. To repro, you can add a timeout promise before the listeners are made.


WorldLanguages:
Can you point out what are the changes compared to the code before #4828 was merged?

WorldLanguages:
Edit: I meant #4627

TheColaber:
Before #4627, there was a bug where dynamically enabling, disabling, or changing settings of an addon would appear as if the addon had been changed. However the listeners in the cs file had not been created yet, so no changes would occur. That's what #4627 fixed. That pr left an if statement rapped around it that shouldn't have. That inevitably lead to this pr. This pr fixed the 1 in a million case where \_page_ had somehow loaded really fast. I think this makes sense hopefully.

apple502j (approved):

Weredime (approved):

PR #4974, opened 2022-08-13
New Addon: Necropost Highlighter

### LINKED ISSUE #4973: New Addon: Necropost Highlighter

**- _pull request en route_**

I'm always frustrated when... I waste time on an old forum topic that has been brought back to life by some evil necromancer (or possibly just some new scratcher who found it in a google search.)

Necroposts are especially evil because you can't tell they're old in the topic list (and they're just infrequent enough that you may not check the first post date until you've spent time trying to figure out how you can help someone.)

I'm submitting a pull request for an addon that adds a highlight color to the topic cell when it is substantially older than the other topics listed on the same forum page.

So that I don't have to load linked pages to inspect actual date stamps, I use the topic ID as a proxy for age. I've inspected old posts and the rate at which the topic ID grows on Scratch over time is pretty stable (definitely enough for the low precision this needs).

I've identified the forums this makes sense for and have defaults (and custom forum selection) in settings.

Tested on Chrome with all forum addons enabled. Works with Standard and Mobile. Plays nicely with "Website dark mode and customizable colors" and "Infinite Scrolling" addons.

I considered chasing links to determine actual dates posts were made. That extra precision doesn't seem like it will be a significant improvement. I've used the Necropost Highlighter locally for a week or so and also scanned many cases. There are occasional false positives, but typically these are also cases that would be difficult to weed out with precise dates of the full message log.

I considered an icon or a thick border instead of a highlighted cell. I have thoughts about other addons that might be more appropriate for that screen real estate.

I considered highlighting the entire row instead of the topic cell. I think the single cell is less obtrusive, but still entirely noticeable, to prevent going to an old topic without being forewarned.

### Additional context

This is what it looks like with the default color choice. _(You will not get the cool gigantic red arrow in your topic list. That's just to point out the cell being highlighted by this addon.)_

Necropost Highlighter Thumb - old topics - with arrow

-- RT_Borg

RT-Borg:
Resolves #4973

More details available in Issue #4973

### Changes

New Addon that adds a highlight color to the topic cell when it is substantially older than the other topics listed on the same forum page.

### Reason for changes

Necroposts are evil. They're less evil when you can see them in the topic list.

### Tests

Tested on Chrome with all forum addons enabled. Works with Standard and Mobile. Plays nicely with "Website dark mode and customizable colors" and "Infinite Scrolling" addons.

### Everyone likes pictures...

This is what it looks like with the default color choice. _(You will not get the cool gigantic red arrow in your topic list. That's just to point out the cell being highlighted by this addon.)_

Necropost Highlighter Thumb - old topics - with arrow

### Feedback note

I'm a seasoned programmer, but this is my first brush with javascript. So if I've done something wrong or broken style conventions, please explain in a way that a noob can understand.


addons/addons.json
```diff
@@ -1,4 +1,5 @@
 [
+  "necropost-highlighter",
```
Weredime, line 2: Should generally be at the end where new userscripts are supposed to be

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,294 @@
...
+ * this ends up working very well.
+ *
+ * Inspects topic ids of all posts on the page and highlights any topics significantly
+ * older than surrounding posts.
+ * 
+ * Works with:
+ *   - Addon: Website dark mode and customizable colors
+ *   - Addon: Infinite scrolling
+ *   - Standard and Mobile websites
+ */
```
Weredime, line 20: Please remove this large comment - It makes the userscript load slower, since network requests are a thing

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,294 @@
...
+	
+	addon.settings.addEventListener("change", onAddonSettingsChange);
+	addon.self.addEventListener("disabled", onDisabled);
+	addon.self.addEventListener("reenabled", onReenabled);
+	
+	// Support for infinite scrolling (only available on Standard not Mobile)
+	const tableBodyNode = document.querySelector("tbody"); // null on mobile
+	const config = {childList: true};
+	const tableBodyMutationObserver = new MutationObserver(onRowsChanged);
+	attachMutationObserverOnStandardSite();
```
Weredime, line 59: Please don't attach `MutationObserver`s to the page, it's not a good practice. Please use [`addon.tab.waitForElement`](https://scratchaddons.com/docs/reference/addon-api/addon.tab/#using-addontabwaitforelement) instead.

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,294 @@
...
+	
+	/**
+	 * How many topics since the earliest on the same page to be considered a "necropost"
+	 */
+	function determineTopicInterval() {
+		if (!addon.settings.get("chooseCustomTopicInterval")) {
+			return defaultTopicInterval;
+		}
+		return addon.settings.get("customTopicInterval");
+	}
```
Weredime, line 168: `addon.settings.get` will **not** return a nullish object.

addons/necropost-highlighter/addon.json
```diff
@@ -0,0 +1,198 @@
...
+				"https://scratch.mit.edu/discuss/11/*",
+				"https://scratch.mit.edu/discuss/m/11/*",
+				"https://scratch.mit.edu/discuss/4/*",
+				"https://scratch.mit.edu/discuss/m/4/*",
+				"https://scratch.mit.edu/discuss/1/*",
+				"https://scratch.mit.edu/discuss/m/1/*",
+				"https://scratch.mit.edu/discuss/3/*",
+				"https://scratch.mit.edu/discuss/m/3/*",
+				"https://scratch.mit.edu/discuss/49/*",
+				"https://scratch.mit.edu/discuss/m/49/*"
```
Weredime, line 45:
```suggestion
				"https://scratch.mit.edu/discuss/*"
```

Weredime (changes requested):
Nice first PR! If you can, could you please enable [actions](https://github.com/RT-Borg/ScratchAddons-NecropostHighlighter/actions) on your fork?

addons/addons.json
RT-Borg: This would be a good clarification for Step 7 of [Getting Started](https://scratchaddons.com/docs/develop/getting-started/creating-an-addon/).

addons/necropost-highlighter/userscript.js
mxmou: `chooseCustomTopicInterval` is a Boolean setting.

addons/necropost-highlighter/addon.json
mxmou: This addon only runs in specific forums so a regex match could be used.

addons/necropost-highlighter/userscript.js
RT-Borg: Thanks. As I said, I'm brand new to javascript. I haven't worked in a language before where comments are a detriment. I guess I just imagined a pipeline that reduced files before transit. I stripped everything but "Uses topic id as a proxy for age, which doesn't require loading other pages to read actual dates." If someone other than me needs to alter this in the future, that much seems necessary to orient them to what's going on.

addons/necropost-highlighter/userscript.js
RT-Borg: I stumbled across mention of addon.tab.waitForElement as I was figuring things out, but I couldn't find any good documentation about it. The mutation observer, in conjunction with the way Infinite Scrolling works, delivers a single event batching together all the new rows for an entire page at a time. Does waitForElement do anything like that? If not, I'm not exactly sure how to wait for the complete set of rows before I consider all the topic IDs together and redetermine highlights.

If I were to replace the coarse mechanism now, which is basically "reset and recomputer" (to support table manipulation beyond what Infinite Scrolling does) I'd want each topic to determine its relative age based on a window of posts around it _in both directions_, which feels even less tractable using waitForElement, unless there are some associated options or events I haven't seen yet.

I also have in mind another addon, which Necropost Highlighter would need to remain compatible with, that could remove rows from a table. Is there a way waitForElement handles that?

With the infrequency of changes to table rows (only via Infinite Scrolling at this point) is this really an issue? Or is there an extreme cost just to have a MutationObserver attached to a forum page?

addons/necropost-highlighter/userscript.js
RT-Borg: I think what you're saying is on 164 "!" can't be used to determine a "null" return.

That's not what I'm doing here. chooseCustomTopicInterval is a boolean setting that acts as an on/off switch for whether to use a default interval or a custom user value.

I'm using "!" as a _logical not_ to say "if the user **does not** enable the option to choose a custom interval", return the default. If you get past this guard, the user has switched "choose custom" to true, so use their custom interval.

addons/necropost-highlighter/addon.json
RT-Borg: The addon really shouldn't run on most discuss pages. Not on the top level directory, not on any of the actual message threads, and not on the many forums where necroposts aren't really an issue (at least not one that can be solved like this, because topics are expected to continue over long periods of time.)

Is there a heavy cost to matching accurately?

addons/necropost-highlighter/userscript.js
apple502j: @Weredime Actually that's not true, userscripts are stored on disk, load time does not matter. However, @RT-Borg I do believe that info is better served on inline comments or just some PR comment (for addon testing data).

addons/necropost-highlighter/addon.json
apple502j: You can use regex if you want - just prefix it with `^` like any other head-matching regexes.

RT-Borg:
I made 2 of the requested changes, pushed, and seem to have introduced a merge conflict in addons. json.  Sorry, I'm really holding onto a lot of new things (new language, new version control, new merge format) with barely a fingernail grasp. I'll have to look at it tomorrow. Any hints or help will be greatly appreciated.

addons/necropost-highlighter/addon.json
RT-Borg: Thank you. Unless someone has a magical incantation they can share with me that gets the job done, I'll look up "head-matching regexes" tomorrow.

RT-Borg: After some false starts and learning json doesn't escape regex special characters in strings, this seems to be the magic:
"^https://scratch.mit.edu/discuss/(?:m/)?(?:search|1|3|4|6|7|9|11|49)/.*"
Lightly tested and pushed, but can someone with deep regex fu please verify.

DNin01:
Maybe we could have (New Posts) replaced with (Recent Necropost) instead of having a blue background? Potentially, changing or adding some text would make it more intuitive than if it changed the background color of the topic in the list to blue. I don't know.

Anyway, nice work! It's pretty cool that you made it work for the Mobile view too.

RT-Borg:
> Maybe we could have (New Posts) replaced with (Recent Necropost) instead of having a blue background? Potentially, changing or adding some text would make it more intuitive than if it changed the background color of the topic in the list to blue. I don't know.

Nice @DNin01! I like the idea of using text. I'd resist eliminating the highlight because a small text change is easy to miss, and in my own use I almost never follow the "New Posts" link. But if the text works out, I believe I can support a transparent color instead of requiring the user to color-pick whatever background they're using (which may change with the dark theme addon).

I think I'll try (Necropost?) instead of (New Posts). That also raises the possibility that it could potentially be a legitimate old thread.

Thanks for the great idea.

RT-Borg:
Thanks again for the suggestion, @DNin01, it looks great! The top necro has the new link text (Necropost?) instead of (New Posts). The bottom necro is already read, so has no link to substitute, so I appended (Necropost?) to the "byuser".

Necro with text affordance

Your idea makes mobile looks better too!

Necro with text affordance-mobile

addons/necropost-highlighter/userscript.js
RT-Borg: Hi @Weredime, I think this and someone experienced with regex putting eyes on my URL match in addon.json are the only outstanding items.

I just pushed a change to address the efficiency of updating row changes coming from the Infinite Scrolling addon.

I did a fair bit of research about mutation observers.  (It generally comports with the accepted answer to https://stackoverflow.com/questions/31659567/performance-of-mutationobserver-to-detect-nodes-in-entire-dom which has been kept updated over time and is still often cited). It seems like the kind of Mutation Observer I'm using _should_ be pretty efficient (I'm targeting a single parent, observing infrequent changes to direct children, subtree false, and only observe childList changes.)

However I wasn't treating it like a service interrupt, and the operation I performed in my mutation callback was a bit expensive. Based on a couple sources including the one above, I altered the callback. It now defers the work and immediately return control.

```
  async function onRowsChanged(mutationList, mutationObserver) {
	requestAnimationFrame(rebuild);
  }

  async function rebuild() {
	await removeAnyPriorHighlights();
	highlightNecropostsIfOnSelectedForumPage();
  }
```

Even before the change, the page fetch and parse from Infinite Scrolling dominates everything else so strongly that this work is imperceivable to me.

I profiled in chrome. My computer was fast when I bought it, but that was 12 years ago, so it's pretty low end by today's standards. The animation frame above is taking < 2 ms (as do subsequent layout and paint). There's a layout shift (< 17 ms) associated with Infinite Scrolling that seems unaffected by the presence or absense of Necropost Highlighter.

(I only learned there were profiling tools in a browser today, so hopefully I'm focusing on the right aspects.)

=================================================

I also updated the code that stores original cell clones, in a topic, and restores them later. I now only store/restore the cells that are actually altered. I don't feel a difference when trying it out, but it still feels like a good change.

=================================================

Since Tabs.waitForElement doesn't seem well documented, I read the implementation in Tab.js (and SharedObserver in module.js).

I could rewrite the addon to use waitForElement, but it would be more than a simple drop in replacement for the MutationObserver.

The first design problem I don't have a good solution for is understanding when all rows from the initial page load have been returned by waitForElement. I really can't proceed a row at a time from the beginning, because it won't be until after the rows are available that I can determine what's "old" on the page and begin highlighting.

Unless there's a mechanism I don't see, I'd probably still have to do the initial page load exactly the way I do now. Then I'd have to start a Tabs.waitForElement loop and do bookkeeping to throw away all the rows it returns that match the cells I've already processed from the initial page load.

That would be necessary because the matches have to be in _waitForElementSet later, in order to use markAsSeen to process the post-load changes that Infinite Scrolling cause.

Then, instead of a single batch update of a page-worth of rows (as it is now with my Mutation Observer) it seems the update would look like:

Necropost Highlighter: await addon.tab.waitForElement (in a loop)

Infinite Scrolling: adds a page-worth of rows (let's say 24)

SharedObserver: has a MutationObserver watching the entire document, and will be active, as I'm waiting. For each of the 24 rows, its callback has to do a document.querySelectorAll on my row query and throw away the prior page(s) items as they match the contents of _waitForElementSet, until it finally hits the first new row (match) and resolves it.

Necropost Highlighter: gets the single resolved row, does bookkeeping to make sure it's unseen on this side, and processes the single row. It would be fine to judge the age based on only prior loaded rows at this point, since we have at least a page of prior cellIds. It's not optimal for the interacting with future addons, but it's as good as the current implementation.

This SharedObserver/Necropost Highlighter interplay happens 24 times (with the full document.querySelectorAll in SharedObserver on each of them).

It feels like Tabs.waitForElement and its SharedObserver might be a very valuable optimization for some cases. Especially with many addons in an editor tab, waiting to respond to isolated element changes (and especially to replace many addon observers that all want to service those requests simultaneously).

But Tabs.waitForElement feels out of place for my particular use, processing a batch update on a discussion page.

And as I said up top, the Infinite Scrolling fetch and parse is so time consuming relative to any of this, it's hard to understand the value proposition.

I could well be misunderstanding something, since I'm brand new to javascript. There could be something I'm not seeing and didn't come across in my research.

Is it truly required to rewrite Necropost Highlighter using Tabs.waitForElement, in order to support Infinite Scrolling?

Weredime:
Should we really need to support mobile? cc @WorldLanguages

RT-Borg:
> Should we really need to support mobile? cc @WorldLanguages

@Weredime, "Need to?" Probably not. But necroposts are just as inconvenient there as on the standard site.

TR_Borg (my wife) uses mobile to browse the forums, and I do occasionally. Kiwi Browser runs chrome extensions on mobile and Scratch Addons works just fine there in the forum. (you do have to browse to a /m/ link, or it takes you to the desktop version). I've got the mobile Help with Scripts page open right in now in Kiwi, with the Dark website addon active. So for my own use, I'd like to see it retained if that's not a problem.

Of the 5-6 people who have commented on this addon, one said how happy he was that I supported mobile, so I guess I'm not alone. Nothing I did for the inclusion of mobile in this addon should slow anything down on the standard site.

RT-Borg:
I just loaded up my development version of Scratch Addons + Necropost Highlighter in Kiwi Browser on my Android phone. The highlighter works exactly as expected on mobile pages there.

I also tested it on Desktop Firefox (windows). Also no surprises, just like Chrome, where I've done my prior testing.

RT-Borg:
Addressed 2 minor bugs I found while testing more extensively with Infinite Scrolling. (Full details in the commit description.)

The Infinite Scrolling addon has a couple problems of its own (one very significant). I made repairs locally and can contribute bug fixes for Infinite Scrolling, but I'll have to learn how. (Git only allows one fork, so I think I either have to wait for this PR to be accepted, or learn how to juggle multiple PRs from my fork.)

Norbiros:
You can create new branch. `git checkout -b NAME`

WorldLanguages:
> Should we really need to support mobile? cc @WorldLanguages

I don't think that's a problem. We don't officially support mobile, but supporting mobile here doesn't affect desktop users as far as I can see.

RT-Borg:
> We don't officially support mobile, but supporting mobile here doesn't affect desktop users as far as I can see.

Thanks @WorldLanguages :-) 

I see I got bounced out of the upcoming release, to [v1.29.0](https://github.com/ScratchAddons/ScratchAddons/milestone/38). Is that just about reviewer availability?

If I'm in a different release, is it my responsibility to update versionAdded in addons.json, or do you guys do that?

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,339 @@
...
+      .filter((topic) => topic.topicId < lowestNewTopicId)
+      .filter((topic) => activatedForums.includes(topic.forum))
+      .forEach(highlightSingle);
+
+    function highlightSingle(topic) {
+      if (!topic.restoreCell) {
+        topic.restoreCell = topic.topicCell.cloneNode(true);
+      }
+      const necropostMessage = "(Necropost?)";
+      const highlightColor = addon.settings.get("highlightColor");
```
Weredime, line 275: Should be translated

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,339 @@
...
+    applyToQuestionsAboutScratch: "Questions about Scratch",
+    applyToBugsAndGlitches: "Bugs and Glitches",
+    applyToProjectIdeas: "Project Ideas",
+    applyToRequests: "Requests",
+    applyToSuggestions: "Suggestions",
+    applyToOpenSourceProjects: "Open Source Projects",
+  };
+  const searchResultsPageName = "Search Results"; // recent posts and unanswered posts
+  const isMobileSite = await determineIsMobileSite();
+  // tableBodyNode will be null on mobile and non-table search page:
```
Weredime, line 33: We don't need to use a function for something we only call once!
```suggestion
  const searchResultsPageName = "Search Results"; // recent posts and unanswered posts
  const isMobileSite = /^\/discuss\/m\//.test(location.pathname)
  // tableBodyNode will be null on mobile and non-table search page:
```

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,339 @@
...
+
+  /**
+   * How many topics since the earliest on the same page to be considered a "necropost"
+   */
+  function determineTopicInterval() {
+    if (!addon.settings.get("chooseCustomTopicInterval")) {
+      return defaultTopicInterval;
+    }
+    return addon.settings.get("customTopicInterval");
+  }
```
Weredime, line 187: You should just return the `customTopicInterval` setting, since it's not nullish?

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,339 @@
...
+    // Standard format: "/discuss/topic/622368/"
+    // Mobile format (read): "/discuss/m/topic/622368/"
+    // Mobile format (unread): "/discuss/m/topic/622368/unread/"
+    // In each case, we want to return the int value 622368
+    const standardTopicPrefix = "https://scratch.mit.edu/discuss/topic/";
+    const mobileTopicPrefix = "https://scratch.mit.edu/discuss/m/topic/";
+    const typicalSuffix = "/"; // all Standard and read mobile messages
+    const unreadSuffix = "unread/"; // only on mobile
+
+    let a = await topicCell.querySelector("a");
```
Weredime, line 324: `await` is not needed here

addons/necropost-highlighter/userscript.js
mxmou: Using a function can make the code more readable and organized. In this case the variable name is clear enough though.

addons/necropost-highlighter/userscript.js
RT-Borg: chooseXXX is a boolean switch in the configuration. If the user has the switch off (false) we use the defaultTopicInterval. If chooseXXX is true, we use the value supplied by the user in customTopicInterval.

addons/necropost-highlighter/userscript.js
RT-Borg: Thanks :)

addons/necropost-highlighter/userscript.js
RT-Borg: Pushed. I used the translation key "(Necropost?)". It works, but please verify that conforms with the style allowed in this codebase.

WorldLanguages:
> If I'm in a different release, is it my responsibility to update versionAdded in addons.json, or do you guys do that?

We'll take care about details like that.

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,333 @@
...
+    // check each forum we explicitly know about, so might be highlighting
+    for (let setting in customForumSettingToForum) {
+      let forum = customForumSettingToForum[setting];
+      if (titleText.includes(forum)) {
+        return forum;
+      }
+    }
+    // a page we don't know how to highlight (and probably shouldn't have run on)
+    return "";
+  }
```
Weredime, line 238: No need for async here!
```suggestion
  function determineForumOrSearchPageName() {
    const title = document.querySelector("title");
    if (!title) {
      return "";
    }
    const titleText = title.innerText;
    if (titleText.includes(searchResultsPageName)) {
      return searchResultsPageName;
    }
    // check each forum we explicitly know about, so might be highlighting
    for (let setting in customForumSettingToForum) {
      let forum = customForumSettingToForum[setting];
      if (titleText.includes(forum)) {
        return forum;
      }
    }
    // a page we don't know how to highlight (and probably shouldn't have run on)
    return "";
  }
```

addons/necropost-highlighter/userscript.js
Weredime: Why have a switch though? We could just use the bare setting (keep in mind people can reset back to default settings easily)

addons/necropost-highlighter/addon.json
```diff
@@ -0,0 +1,185 @@
...
+  ],
+  "dynamicDisable": true,
+  "dynamicEnable": true,
+  "enabledByDefault": false,
+  "tags": ["community", "forums"],
+  "versionAdded": "1.28.0",
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["^https://scratch.mit.edu/discuss/(?:m/)?(?:search|1|3|4|6|7|9|11|49)/.*"]
```
Weredime, line 25: We don't need wildcard at the end? Query params are stripped from the end of the URL
```suggestion
      "matches": ["^https://scratch.mit.edu/discuss/(?:m/|)(?:search|1|3|4|6|7|9|11|49)/"]
```

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,333 @@
...
+  // tableBodyNode will be null on mobile and non-table search page:
+  // https://scratch.mit.edu/discuss/search/?action=show_user&show_as=posts
+  const tableBodyNode = document.querySelector("tbody"); // null on mobile
+
+  if (!tableBodyNode && !isMobileSite) {
+    // Standard site page without a table. Don't run on this page.
+    return;
+  }
+
+  const forumOrSearchPageName = await determineForumOrSearchPageName();
```
Weredime, line 43:
```suggestion
  const forumOrSearchPageName = determineForumOrSearchPageName();
```

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,333 @@
...
+        highestSoFar = topics[i].topicId;
+      }
+    }
+    return highestSoFar;
+  }
+
+  /**
+   * Returns int topicId, or 0 if no topic was found.
+   */
+  async function extractTopicIdFrom(topicCell) {
```
Weredime, line 308: Most of the functions in this addon really do not need async.
```suggestion
  function extractTopicIdFrom(topicCell) {
```

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,333 @@
...
+   *    - restoreCell  Only defined if the cell is modified. A clone of the initial state
+   *                   of the cell Node, to revert for disable addon reload and settings changes
+   */
+  let topics = [];
+
+  highlightNecropostsIfOnSelectedForumPage();
+
+  addon.settings.addEventListener("change", onAddonSettingsChange);
+  addon.self.addEventListener("disabled", onDisabled);
+  addon.self.addEventListener("reenabled", onReenabled);
```
Weredime, line 60: Could the referenced functions here be used as arrow functions?

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,333 @@
...
+
+  // Support for infinite scrolling (only available on Standard not Mobile)
```
Weredime, line 62: Do we need to create a `MutationObserver` on mobile pages, if Infinite Scrolling doesn't work?

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,333 @@
...
+   * Restores topic cells to the initial state they had before any highlighting on this page.
+   * Clears any previously gathered topics.
+   * Does not conflict with Addon: Website dark mode and customizable colors.
+   */
+  async function removeAnyPriorHighlights() {
+    topics.forEach((topic) => {
+      if (topic.restoreCell) {
+        topic.topicCell.replaceWith(topic.restoreCell);
+      }
+    });
```
Weredime, line 200: Maybe implement colors as a style? We shouldn't waste memory on something that can be implemented as a userstyle

addons/necropost-highlighter/userscript.js
RT-Borg: I think you're suggesting to declare tableBodyMutationObserver here, but assign it inside the !isMobileSite guard of attach (and free it in detach). I'll make a commit for this.

addons/necropost-highlighter/userscript.js
RT-Borg: I didn't want to put the strange detail of how age is determined in front of the 99% of users who won't care. But it'll probably look fine. I'll try it out, see how it feels, and probably commit your suggestion.

addons/necropost-highlighter/userscript.js
RT-Borg: If I understand the references I'm reading about => correctly, the different closures wouldn't be relevant in these cases. (I'd have to test it.) To me, this code looks easy to understand, but maybe that's because I'm used to idioms from other languages.

Are you pointing out a correctness issue, or raising a style preference?

Working on bugs in the infinite-scroll addon, forumScroll.js, I see that instead of 
```
function update () {...}
window.addEventListener("scroll", update, {passive: true});
```
the code is
https://github.com/ScratchAddons/ScratchAddons/blob/d1a7bbf61ec2cfb8ca470138dab528311452227b/addons/infinite-scroll/forumScroll.js#L19
https://github.com/ScratchAddons/ScratchAddons/blob/d1a7bbf61ec2cfb8ca470138dab528311452227b/addons/infinite-scroll/forumScroll.js#L66

Is that what you're suggesting? It looks awkward to me, but if that's idiom Scratch Addons uses, I can conform. Please let me know if I need to make these changes.

addons/necropost-highlighter/userscript.js
RT-Borg: I initially just added a style background property to highlight cells and removed all such properties to deactivate. When I added the feature to replace link text or append "Necropost?", I instead started capturing the state before my change in a "restoreCell" to revert all changes at once.

Is the memory used to store copies of a couple cells on the page really significant?

addons/necropost-highlighter/addon.json
RT-Borg: @Weredime suggested an improvement in review. Thank you.

addons/necropost-highlighter/userscript.js
RT-Borg: I think this is resolved? (since @Weredime suggested a change to how the MutationObserver get constructed, in a review).

RT-Borg:
@Weredime thanks for the great reviews! This is the first javascript I've written (kinda a long "Hello World!") so I appreciate the explanations (and the regex suggestions!!)

Early into this, I didn't realize querySelector calls are synchronous, so sorry for some stray awaits that held on to the bitter end.

For review question, _"Could the referenced functions here be used as arrow functions?"_ and _"Maybe implement colors as a style? We shouldn't waste memory on something that can be implemented as a userstyle"_, I wasn't sure if I was supposed to click the [Start a review] button. I just added my replies as a single comment and left them "unresolved". Please let me know if I'm using the review tool wrong.

I think I've addressed everything else. All the changes are lightly tested, but I'll exercise it diligently to be sure.

RedGuy12:
> I see I got bounced out of the upcoming release, to [v1.29.0](https://github.com/ScratchAddons/ScratchAddons/milestone/38). Is that just about reviewer availability?

Reviewer availability is one reason, yes, but also v1.28 is being released this week and we want to give translators time to translate new strings before we send the update to the Chrome Web Store.

I don't think Requests should be "suspect". If a topic in there dies, either shop workers are no longer active or the request is no longer needed.

addons/necropost-highlighter/addon.json
```diff
@@ -0,0 +1,174 @@
+{
+  "name": "Necropost highlighter",
```
RedGuy12, line 2:
```suggestion
  "name": "Forums necropost highlighter",
```

addons/necropost-highlighter/addon.json
```diff
@@ -0,0 +1,174 @@
...
+  "settings": [
+    {
+      "name": "Highlight color",
+      "id": "highlightColor",
+      "type": "color",
+      "allowTransparency": true,
+      "default": "#D1E9F2"
+    },
+    {
+      "name": "How many topics is 'old'? (7,500 is about a month)",
```
RedGuy12, line 37: Maybe ask "How many months is 'old'?" and multiply that by 7500 in the code.

RT-Borg:
> Reviewer availability is one reason, yes, but also v1.28 is being released this week and we want to give translators time to translate new strings before we send the update to the Chrome Web Store.

Thanks @RedGuy12 :-)

> I don't think Requests should be "suspect". If a topic in there dies, either shop workers are no longer active or the request is no longer needed.

I've never really been happy with the terms "advised/marginal/suspect" and now I see they caused a misunderstanding. These terms are supposed to be about how recommended it is to turn on the highlighter in a forum, because it may have too many false positives.

This was a good prompt to come up with better phrases. I've changed the notes on forum names in Settings to be "usually accurate", "mostly accurate", "less accurate". Those really capture what the decision is about. Commit incoming. Thanks again, @RedGuy12.

addons/necropost-highlighter/addon.json
RT-Borg: I only see the term Necropost used about posts on the Scratch site, so "forum" is implied.

Most other addons that apply to the forums don't begin with the word Forum (some don't use it at all). If the word "forum" is required, I'd prefer:

Necropost highlighter (forums)

or the less scannable

Necropost highlighter on forums

addons/necropost-highlighter/addon.json
RT-Borg: I think that should look good, if I say "Roughly how many months is 'old'?"

"Roughly" should avoid users thinking it will be precisely a month (it could be more like 27 or 33 days, depending on recent forum traffic).

Putting the 7500 topic interval back in the code also gives us more flexibility to update the per-month interval if eg Scratch forums get significantly more traffic next year.

Thanks @RedGuy12. Commit incoming.

RT-Borg:
I'm not sure I'm doing this right. It was over 2 weeks without hearing anything, and since I made changes from my last review. I'd left a couple questions in the review for @Weredime and hadn't received a reply. I just clicked something I think called "request re-review" thinking that would be the way to make sure @Weredime had seen my questions. I see above that it removed a request for @WorldLanguages to review. I just want to make sure I've done what I need to do (or discover what else is needed) so I can have this addon included in the next release. 

Apologies if I've used this review tool wrong.

Samq64:
Don't worry, reviews just take a while sometimes.

TheColaber:
@RT-Borg expect this pr to be merged for this version.

RT-Borg:
Thanks @Samq64 and @TheColaber! It's 3 weeks since I got bumped from the last release,  so I was concerned time is growing short.

I completely understand the limitations and bottlenecks on a volunteer project. When I saw my action remove @WorldLanguages as a reviewer, I wanted to make sure I wasn't gumming up the works.

Weredime:
@RT-Borg Sorry for not replying sooner, I was busy with school. I'll review it soon

mxmou:
If possible, addons should use the colors from the "website dark mode and customizable colors" addon instead of having a separate setting for the color. This one could use the "blue background". There could also be a setting to enable or disable color highlighting.

I don't think the age setting needs to be hidden behind a switch.

RT-Borg:
> @RT-Borg ...This looks good to me right now, but we probably need to test edge cases

Great, thanks @Weredime. I really appreciate all your input on the PR. I've tested it quite a bit (and have used it daily for over a month) but always happy to have more eyes on it watching for bugs.

RT-Borg:
Thanks @mxmou

> If possible, addons should use the colors from the "website dark mode and customizable colors" addon instead of having a separate setting for the color. This one could use the "blue background"

Oh, I really liked this idea when I read it, but I just tried the 3 presets in "website dark mode and customizable colors" and manually changed the Necropost Highlighter color to the hex values listed in "blue background".

With the "Scratch defaults" theme and "Experimental Dark" theme, a cell with "blue background" isn't very distinguishable from the other cells on my monitor. Certainly not a difference I'd call "highlighted". I'd easily miss them scanning the screen. Dark WWW is a little more distinct.

Setting "blue background" to an obvious color and looking around a bit, I see "blue background" is used as the background color for studio pages. I'm not sure where else. **Is "blue background" used to call attention to cells like this anywhere else?** Because the default color choices for "blue background" don't seem chosen with cell highlighting in mind (and anyone who has customized it for the look of their studio pages wouldn't necessarily pick a great cell highlight color either).

I also gave presets for blue and red options because some users might want it to blend in with the standard colors more, while others might consider a necropost more like a warning or error.

> There could also be a setting to enable or disable color highlighting.

The "No Cell Color" preset disables highlighting. Added in the[ August 13 commit](https://github.com/ScratchAddons/ScratchAddons/pull/4974/commits/c4c0ceca606a4940687c7765fe63cdb923092024), based on @DNin01's suggestion: https://github.com/ScratchAddons/ScratchAddons/pull/4974#issuecomment-1214297564

> I don't think the age setting needs to be hidden behind a switch.

That was changed in the [August 23 commit](https://github.com/ScratchAddons/ScratchAddons/pull/4974/commits/99881a1640f19a2cf53b0c3334dc71d996a58722). Age is no longer behind a switch (it now asks directly "Roughly how many months is 'old'? with a default of 1). The calculation is done internally, which also gives us the nice backdoor to change the base value for a month down the line if forum usage changes significantly.

mxmou:
> **Is "blue background" used to call attention to cells like this anywhere else?** Because the default color choices for "blue background" don't seem chosen with cell highlighting in mind (and anyone who has customized it for the look of their studio pages wouldn't necessarily pick a great cell highlight color either).

The blue background color is used in a few different ways:
* to separate an element from the background (example: project instructions)
* as a background for a section or an entire page (examples: project comments, studios)
* to highlight unread messages on the Messages page

The last one is the most similar to what this addon needs. It's also quite different from the other two. Would using a transparent version of the "highlight color" work better for things like that?

RT-Borg:
TL;DR adopting @mxmou's suggestion to use a partially transparent "Highlight color" from "website dark mode and customizable colors". I'll add a toggle setting for whether cells should be colored, as the color selection is removed.

> The blue background color is used in a few different ways:

> - to highlight unread messages on the Messages page

> The last one is the most similar to what this addon needs. It's also quite different from the other two. Would using a transparent version of the "highlight color" work better for things like that?

@mxmou, trying partially transparent versions of "Highlight color" was a creative solution!

I played with different transparency levels on each of the 3 default themes. To my eye, both dark themes and the standard Scratch theme all highlight nicely with "Highlight color" at 20% opacity (Hex 0x33). 

Just to be super clear, so you can have a look too, since all three use the same highlight color, #4d97ff, that means I'm proposing #4d97ff33 for this type of highlight, on all 3 default themes.

While I think this addon in isolation would be slightly more friendly if it allowed users to select more of an "error" color if they wanted (as I first envisioned it) I think Scratch Addons overall would benefit more from keeping the custom colors in one place. I support @mxmou's suggestion to remove the color selection from Necropost Highlighter.

Unless there's pushback on 20% opacity, I'll go ahead and put #4d97ff33 in for the default when "website dark mode and customizable colors" is disabled, and append "33" onto whatever "Highlight color" that addon reports when it is enabled.  I believe I saw there's a way to detect when another addon is active, and retrieve a setting from it, but I'll have to dive back into the docs. A commit for this might not be ready until tomorrow.

With color selection going away, I'll also add a settings toggle to turn cell coloring on/off (since that's currently handled in color selection).

@mxmou, thanks for sticking with this and making the idea work. I think the same 20% opacity of "Highlight color" would look good for new messages.

-- RT_Borg

CST1229:
> 2. Use getComputedStyle(_element_).getPropertyValue("--darkWww-highlight-transparent20") to retrieve the value. For the standard site I'll use tableBodyNode for the element. I'll find an appropriate analog on the mobile site.

Wouldn't it be possible to apply a class to the part you need and then use a userstyle?

RT-Borg:
> > 2. Use getComputedStyle(_element_).getPropertyValue("--darkWww-highlight-transparent20") to retrieve the value. For the standard site I'll use tableBodyNode for the element. I'll find an appropriate analog on the mobile site.
> 
> Wouldn't it be possible to apply a class to the part you need and then use a userstyle?

Thanks @CST1229. I implemented the getComputedStyle approach and committed before seeing your message. It seemed pretty straightforward, and I only need to retrieve the value once, but if there's a better way, I'm open to revising. (This is my time using javascript, css, and many other things, so I'm still learning.)

Are you suggesting to use classList.add() on the highlighted cells. Would the argument I pass be "--darkWww-highlight-transparent20"? I think I'd still need some kind of check to find out whether I need to use the default value instead (because dark-www may not be enabled, in which case that variable isn't set). It seems like a little more work, but would it be better?

I think it might properly update if the user happens to modify their highlight color in dark-www, but I don't think it would behave well if the user disabled dark-www.

I'd like to respond to dark-www being disabled/enabled and to highlight color changes in dark-www. The way I think I can do that is using a second Mutation Observer with

`observe(html, { attributes: true, attributeFilter: ["style"] })`

and then calling my existing onAddonSettingsChange() if "--darkWww-highlight-transparent20" differs from the current highlight color (unless it's falsy and the highlight color is already the default color, which is very likely.)

I'll try the Mutation Observer tomorrow (need sleep now). But I know there were concerns before over the efficiency of a Mutation Observer, so if anyone has a better suggestion to witness the value from dark-www being available/unavailable when that addon is enabled/disabled, please let me know.

CST1229:
> Are you suggesting to use classList.add() on the highlighted cells. Would the argument I pass be "--darkWww-highlight-transparent20"?

I'm suggesting to use classList.add() on the highlighted cells to add a class (something like `highlighted-necropost`), and then use a userstyle to do something like `.highlighted-necropost {background-color: var(--darkWww-highlight-transparent20) /* and also other things if needed */}`.

RT-Borg:
Oh, I didn't know you could do that. Ok, in the morning I'll make a .css like that, reference it in my addon.json, and in userscripts.js add the class to cells instead of directly setting the background color.

Thanks so much, @CST1229. That's a really big help.

#### REVIEWERS please note:
I'll be changing the code from commit 32a064e based on this new advice.

RT-Borg:
I made the changes suggested by @CST1229 and it works great (thanks!)

**I haven't committed because I'm running into a problem with git (github desktop). Can someone with git experience please help me?**

Something went wrong with the previous commit (the action running Prettier) and I just have no idea what the error messages mean or how to recover.

The way I've been interacting with git (though git desktop) is to pull from origin first (in case there are any conflict) and then commit and push.

Attempting to pull, I get an error message:

"Unable to pull when changes are presenet on your branch. The following files would be overwritten: addons\necropost-highlighter\userscript.js"


Investigating, I see the commit I pushed yesterday is marked as 

"This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

On github, Actions, it lists a workflow run error:

Format code using Prettier #8926: Pull request #4974 synchronize by RT-Borg

The linked page is:

https://github.com/ScratchAddons/ScratchAddons/actions/runs/3060584804

On that page it says:

Annotations
2 errors
Format
Process completed with exit code 128.
Format
Error: Input required and not supplied: app_id

I also see I received mail saying "Format code using Prettier: All jobs have failed"

I don't know anything about github workflows or actions or what it means that no app_id is supplied.

mxmou:
@RT-Borg Pulling from origin can fail if there are uncommited changes. You can select Branch → Stash all changes, then pull, restore from stash, and commit.

mxmou:
> I got [the same error](https://github.com/ScratchAddons/ScratchAddons/actions/runs/3066376042) on Prettier again. Any idea what's causing it (and is it a problem I need to worry about)? I did get a nice green checkmark next to the commit this time [7736398](https://github.com/ScratchAddons/ScratchAddons/pull/4974/commits/77363983585735f4e13e49c0fc99cdbf126144ff).

Those errors always appear when Prettier has to make formatting changes.

> Also, I don't seem to have all the latest from Scratch Addons after pulling. E.g., from manifest.json I still have "version_name": "1.28.0-prerelease", but looking on github, I see manifest.json has been updated to "version_name": "1.29.0-prerelease". Do I have something configured wrong?

Your fork is the remote repository used by push and pull by default. To get the latest changes from Scratch Addons, you can merge "upstream/master" into your branch. This is usually only necessary if there are merge conflicts.

RT-Borg:
> Those errors always appear when Prettier has to make formatting changes.

Ah, thanks. 

> > Also, I don't seem to have all the latest from Scratch Addons...
> Your fork is the remote repository used by push and pull by default. To get the latest changes from Scratch Addons, you can merge "upstream/master" into your branch. This is usually only necessary if there are merge conflicts.

I realized I haven't gotten to use the cool new _Alternating nested blocks_ addon, which released in 1.28 . But I think I'll hold off on anything that might rock the boat, since this last change to base the highlight color off _dark-www_ just made it under the wire for the 1.29 milestone due date.

RT-Borg:
**Bumped AGAIN? To yet another release?? Not even an explanation?** Not a word? _I just don't know how to express how disheartening this is._

WorldLanguages:
Unfortunately, this PR didn't receive any approvals as of October 14th, so it can't be merged for 1.29
Keep in mind that approvals from members outside of the ScratchAddons org may also count, in some scenarios.

Thanks for the feedback, we'll think of ways to speed up the review process of new addon submissions, which are usually the most complex. We org members slowly started to (unintentionally) specialize in certain types of addons, but there's no written rules on who is more suitable to review a new addon for the forums (other types of addons usually do have an unwritten rule where a specific contributor(s) will review it). I do agree on that we should work on making the review process more predictable.

RT-Borg:
@WorldLanguages 

I asked time and time again, **for months**, whether there was anything else I needed to do.

Two months ago I was bounced from v1.28 _(also silently, without any notice.)_ When I asked, I was told at the time the reason I didn't get in v1.28 was "Reviewer availability is one reason, yes, but also v1.28 is being released this week and we want to give translators time to translate new strings". https://github.com/ScratchAddons/ScratchAddons/pull/4974#issuecomment-1224894980 (For this addon, there's a single word to be localized.)

I addressed all @Weredime's review issues months ago. Then after not getting confirmation for a couple weeks, I asked (on September 8), "I just want to make sure I've done what I need to do (or discover what else is needed) so I can have this addon included in the next release." https://github.com/ScratchAddons/ScratchAddons/pull/4974#issuecomment-1240523088


On Sep 8, @TheColaber replied "expect this pr to be merged for this version." https://github.com/ScratchAddons/ScratchAddons/pull/4974#issuecomment-1240697426


On Sep 9, the reviewer @Weredime finally replied, "**_This looks good to me right now_**, but we probably need to test edge cases" https://github.com/ScratchAddons/ScratchAddons/pull/4974#issuecomment-1242293716 (to which I assured him I'd tested extensively.)

The final change (using colors from "website dark mode") was requested and completed over a month ago.

And the very last comment on the thread, a full month before getting silently bumped out of v1.29, I wrote that I didn't want to even upgrade my local version to take advantage of "Alternating nested blocks" because I'd met the 1.29 deadline (even with the last minute color request) and was code freezing myself for the pr.

Since then, the "deadline" for v.1.29 was (again silently) pushed back an entire month, and somehow I've still been bumped out of it?

> Keep in mind that approvals from members outside of the ScratchAddons org may also count, in some scenarios.

Was I supposed to find other programmers I know, who have never worked on ScratchAddons, and recruit them to come look at my code and give a review? Nothing in the instructions suggest this, and the many people commenting on this PR never told me to bring in an external reviewer.

> Unfortunately, this PR didn't receive any approvals as of October 14th, so it can't be merged for 1.29

This pr has had 2 "pending reviewers" listed for months. I already explained that @Weredime wrote "looks good to me" over a month ago. I don't know why he didn't press some kind of "approve" button to go along with that. @WorldLanguages, you're the other "pending reviewer" on the project, and have been "pending" for months. In fact, on Sep 8 when I tapped "request re-review" to get @Weredime's attention after weeks, and the tool removed you, I raised a concern and @Weredime re-added you the next day.

Relative to many other addons I've browsed in the repo, this is really straightforward. It doesn't do anything complex or risky, and there's nothing tricky about the implementation. Any time a request was made, I submitted code or an explanation immediately. (Usually within hours, but at least by the following day. I even stayed up working _very_ late into the night/morning more than once, to keep the turn-around tight.) I asked so many times, on this page and reviews, whether I was doing everything necessary that I felt on the border of being a pest about it. Now I just have sadness, and I look foolish to the people I told to expect my addon. And worst of all, the addon is still unavailable to those who would benefit.

RedGuy12:
This addon just isn't an addon that we see being requested a lot, so it's not a high priority to be merged.

> For this addon, there's a single word to be localized.

The entire manifest as well

addons/necropost-highlighter/userscript.js
Weredime: > Is that what you're suggesting? It looks awkward to me, but if that's the idiom Scratch Addons uses, I can conform. Please let me know if I need to make these changes.

I'm suggesting something like this:
```javascript
addon.self.addEventListener("disabled", () => {
	// ...
});
```

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,307 @@
...
+
+  highlightNecropostsIfOnSelectedForumPage();
+
+  addon.settings.addEventListener("change", onAddonSettingsChange);
+  addon.self.addEventListener("disabled", onDisabled);
+  addon.self.addEventListener("reenabled", onReenabled);
+
+  // Support for infinite scrolling (only available on Standard not Mobile)
+  let tableBodyMutationObserver;
+  const config = { childList: true };
```
Weredime, line 64: Put this with the object initializer instead of referencing config there

addons/necropost-highlighter/userscript.js
Weredime: It's not significant, but it could help multi-tasking

Weredime (approved):
Other than the above, LGTM

lisa-wolfgang:
@RT-Borg Please keep in mind that this is a hobby project. The org members (myself included) have commitments outside of Scratch Addons. Even as an org member, my PRs get delayed release after release simply because of necessary prioritization of other issues. The proper reviewing of new addons takes time, even for simple ones, and we simply were not able to devote enough time to this project over the past couple of months to pass this one through the approval process. As you mentioned, we even needed to postpone v1.29 by a full month because of this.

WorldLanguages:
> And worst of all, the addon is still unavailable to those who would benefit.

I want to add: I really understand this feeling.
That feeling was, in a nutshell, the main motivation for Scratch Addons to be created. At the time, the "Scratch 2.0→3.0" userstyle made by mxmou wasn't widely used, even tho it was very useful. It was unavailable for many Scratchers, because they were unaware of it.
However, after 2 years of Scratch Addons having thousands of users, it becomes impossible for frequent contributors/reviewers to preserve that aura. This is a hobby project, and as SA gets bigger, the responsibility gets larger and complexity increases. Our productivity would actually go down, if we felt pressured because users aren't getting the best potential version of SA. It would be mentally unhealthy to approach the project that way, week after week, so that's why it may look like we lack some emotions - we would experience burn-out.
However, we're still better off from the old days, where there was no centralized place for Scratch features. But we do need to improve our communications with new contributors, specially the ones that haven't joined our Discord server. Nothing we can do about things that already happened, tho.

WorldLanguages:
@apple502j I remember you once mentioned that you believe this addon should is not suitable for the SA addon library.

If that's the case, could you briefly elaborate why?

Samq64 (approved):

addons/necropost-highlighter/addon.json
```diff
@@ -0,0 +1,142 @@
...
+      "default": true
+    },
+    {
+      "name": "Choose custom forums?",
+      "id": "chooseCustomForums",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "name": "New Scratchers (usually&nbsp;accurate)",
```
Samq64, line 54: Is `&nbsp;` necessary? It makes the code a little harder to read.

addons/necropost-highlighter/addon.json
```diff
@@ -0,0 +1,142 @@
...
+      "url": "style.css",
+      "matches": ["^https://scratch.mit.edu/discuss/(?:m/|)(?:search|1|3|4|6|7|9|11|49)/"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Roughly how many months is 'old'?",
+      "id": "monthCountConsideredOld",
+      "type": "positive_integer",
+      "default": 1
```
Samq64, line 39: I think this setting should have minimum and maximum values, 1 and 24 seem good to me.

addons/necropost-highlighter/userscript.js
Samq64: I agree with Weredime. It doesn't necessarily have to be an arrow function but could you put the dynamic enable/disable code here instead of using separate functions?

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,307 @@
...
+  // mobile urls look like: https://scratch.mit.edu/discuss/m/6/*
+  const isMobileSite = /^\/discuss\/m\//.test(location.pathname);
+  // tableBodyNode will be null on mobile and non-table search page:
+  // https://scratch.mit.edu/discuss/search/?action=show_user&show_as=posts
+  const tableBodyNode = document.querySelector("tbody"); // null on mobile
+
+  if (!tableBodyNode && !isMobileSite) {
+    // Standard site page without a table. Don't run on this page.
+    return;
+  }
```
Samq64, line 41: Not important, but conditions like this can be condensed into a single line.
```suggestion
  // Standard site page without a table. Don't run on this page.
  if (!tableBodyNode && !isMobileSite) return;
```

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,307 @@
...
+  function gatherTopics() {
+    topics = [];
+    let possibleTopicCells; // verified before forming a topic
+    if (isMobileSite) {
+      possibleTopicCells = document.querySelectorAll(".topic.list>li");
+    } else {
+      possibleTopicCells = document.querySelectorAll(".tcl");
+    }
+    let theForum = forumOrSearchPageName; // usually, but overridden per cell if on a search page
+    for (let i = 0; i < possibleTopicCells.length; i++) {
```
Samq64, line 155: Could a `for... of` loop could be used here and elsewhere where possible?

Samq64 (commented):
These are all just minor things, the addon works great!

addons/necropost-highlighter/addon.json
RT-Borg: It's been quite a few months, but as I recall, I added the non-breaking space after recognizing that the output looked really bad when "(usually" formats on one line and "accurate)" overflows to the next line. I suggest keeping the non-breaking space unless the review team wants to test it the other way and disagrees about how qualifiers like "(usually accurate)" look when split.

RT-Borg:
Thanks @Samq64.

On October 21 2022, @WorldLanguages wrote:
> If some small changes need to be done, you don't need to worry @RT-Borg, we can directly push to your branch

As WorldLanguages suggested, please consider the code donated and feel free to make changes deemed necessary by the review team.

Samq64 (approved):

addons/necropost-highlighter/addon.json
```diff
@@ -0,0 +1,143 @@
+{
+  "name": "Necropost highlighter",
+  "description": "Spot old forum topics brought back to life, before opening the message. Topics started much earlier than others listed on the same page are highlighted (in selected forums). If you've wasted time reading zombie messages or a help/question/request that was actually asked years ago, this is for you!",
```
Samq64, line 3:
```suggestion
  "description": "In the selected forums, highlights topics started much earlier than others on the same page, so you don't waste time reading zombie messages or replies to help, question or request topics that were actually asked years ago.",
```

addons/necropost-highlighter/userscript.js
```diff
@@ -0,0 +1,307 @@
...
+ * @author RT_Borg
+ *
+ * Uses topic id as a proxy for age, which doesn't require loading other pages
+ * to read actual dates.
+ */
+
+export default async function ({ addon, global, console, msg }) {
+  // Default is a little over a month's worth of topics. Reasonably stable through 2022.
+  // There's a lot of tolerance. Half or twice as many topics filter pretty much the same.
+  const TOPICS_PER_MONTH = 7500;
```
Samq64, line 11: Would replacing this with a topic count setting (maybe divided by 1000) work so we don't need to maintain this?

RT-Borg: Back when I started this, I looked at the pace topics increased over the last couple years and found that averaged over months they were surprisingly consistent (IIRC, a small upward drift from one year to the next, but not so much that this constant would need to have changed). I was a surprised there weren't spikes associated with summer or the school year in some way, but I didn't see a dramatic change. It probably deserves a quick yearly spot check, to be on the safe side (or a check if something significantly increases or decreases Scratch use overall).

I'm not sure what alternate age mechanism you're suggesting might be used?

addons/necropost-highlighter/addon.json
RT-Borg: Thanks Sam! I support however you think the description should be worded.  Thinking a bit more about the actual implementation, the last phrase should probably be "... actually asked **long** ago" rather than "... actually asked years ago" since the default is ~1 month, not years.

addons/necropost-highlighter/userscript.js
Samq64: I was thinking of replacing the "Roughly how many months is 'old'?" setting with a setting to input the number of topics directly, but if `TOPICS_PER_MONTH` stays relatively stable over time it probably isn't necessary.

addons/necropost-highlighter/addon.json
Samq64: Feel free to tweak my description.

addons/necropost-highlighter/userscript.js
RT-Borg: Funny. That's how I initially implemented it, and earlier reviewers suggested this instead. I think this actually turns out to be much better because if the rate of topics significantly changes, a SA maintainer can make a single change to update it, rather than expecting every user to be knowledgeable enough to adjust their settings.

addons/necropost-highlighter/addon.json
RT-Borg: I'm happy with your description, Sam. "**long** ago" is my only suggestion. Thanks again.

PR #4981, opened 2022-08-14
Change spelling: "cancelling" to "canceling"

DNin01:
Does not resolve any open issues.

### Changes

Changed "cancelling" to "canceling" in the `confirm-actions` addon.

### Reason for changes

"Canceling" is usually favored by Americans, whereas "cancelling" is usually used in other dialects. Source: https://www.grammarly.com/blog/canceled-vs-cancelled

Nobody will notice this, but I think this change is still worth it.

### Tests

Didn't test, but the changes should work just fine. It's just language.

### DIFF
--- a/addons/confirm-actions/addon.json
+++ b/addons/confirm-actions/addon.json
@@ -37,7 +37,7 @@
       "default": true
     },
     {
-      "name": "Confirm cancelling comments",
+      "name": "Confirm canceling comments",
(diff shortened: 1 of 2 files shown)


Secret-chest:
In British English it's cancelling

apple502j (approved):

WorldLanguages (approved):

PR #4952, opened 2022-08-04
More Blues + Less Classical

OregSamSas:
Resolves #4935

### Changes

No. 1: Complete Blues on Beta (aka. Full De Daba Dei)
No. 2: Remove blue toggles from stable (aka. what the title of the linked issue sez)
I've replaced most of the orange colours when it's a prerelease version (No. 1), and also removed the blue colour of the setting options' toggles(No. 2). Now it's a less saturated type of that colour which is set to the main. (i. e. the "--brand-orange" variable)
   
<details>
<summary>Screenshot1</summary>

</details>
<details>
<summary>Screenshot2</summary>

</details>
<details>
<summary>Screenshot3</summary>

</details>
<details>
<summary>Screenshot4</summary>

</details>
<details>
<summary>Screenshots 5, 6</summary>

</details>
<details>
<summary>Screenshots 7, 8</summary>

</details>

### Reason for changes

Changing helps distinguish whether the enabled addon is a beta version or not, and also looks better. 
### Tests

I've tested all of the changes on all platforms (Chrome 103.0, Waterfox  G4.1.3.2, Edge 103.0), and sadly it didn't work somehow on Microsoft Edge with prerelease versions, but I open the pull request anyway (considering that not many people will probably use Edge for beta-testing)

### Related PR
#4934 But in my version, the colour of input borders on focusing are altering on the type of the SA version (beta=>blue/stable=>orange)

<details>
<summary>See image...</summary>

</details>

For the reason I prefer this way, but I can change it to Samq64's ideas if that will have been merged, and not this. 

My previous proposal can be found here: #4936

PR #4989, opened 2022-08-16
Color changes

### LINKED ISSUE #4983: Use colors in our UIs in a more meaningful way
#4810 replaced the main orange color with blue in pre-release versions and nobody seems to agree on what should stay orange and what should become blue. My suggestion is to improve the way different colors are used in general:
* Icons and headers would be orange in the stable version and blue in the development version.
image
image
* Orange would indicate elements that were selected, focused, or enabled by the user:
  * tabs and categories:
  image
  image
  * buttons and inputs:
  image
  image
  image
  * `"type": "select"` settings:
  image
  * addon and setting switches:
  image
  * if orange indicates something important or dangerous, it would be changed to yellow:
  image
  image
* Blue would be used for other elements that need to stand out:
  * the main button in a button row:
  image
  * currently opened game in cloud games list:
  image
  * links (always or when hovered):
  image
  image

Badges could also be improved - the amount of different colors makes them look a bit messy.
* I don't think "Forums", "For website" and "For editor" are important enough to have unique colors. Something like this would look fine:
image
* Changing "Beta" to yellow would allow orange to be used for updates, making the additional purple color unnecessary:
image

mxmou:
Resolves #4983
Closes #4952

Also replaces some icons with inline SVGs for easier styling. Marking as draft for discussion.


webpages/common/components/inline-icon.html
```diff
@@ -0,0 +1,28 @@
...
+    viewBox="0 0 24 24"
+    fill="currentcolor"
+    style="transform: rotate(360deg)"
+  >
+    <path
+      v-if="icon === 'cloud'"
+      d="M18.42 9.21a7 7 0 0 0-13.36 1.9A4 4 0 0 0 6 19h11a5 5 0 0 0 1.42-9.79zM17 17H6a2 2 0 0 1 0-4 1 1 0 0 0 1-1 5 5 0 0 1 9.73-1.61 1 1 0 0 0 .78.66A3 3 0 0 1 17 17z"
+    />
+    <path
+      v-if="icon === 'evenlope'"
```
TheColaber, line 16: You spelled envelope wrong here and here

https://github.com/ScratchAddons/ScratchAddons/blob/ba8c3357b28a99bc39c1f20fe58d59b21123f847/addons/scratch-messaging/addon.json#L13

WorldLanguages:
Can you send screenshots of the visual changes?

Or is the discussion about the creation of an `<inline-icon>` component? We should be careful and check if settings page performance is affected by that. Vue will probably need to handle many more JS objects in memory and there may be more DOM nodes.

mxmou:
> Can you send screenshots of the visual changes?

image
image
image
image
image
image

Hans5958:
This will require changes on the docs, specifically on [the Design page](https://scratchaddons.com/docs/reference/design).

mxmou:
This has been open for almost a year and no progress has been made. I'll probably split it into smaller PRs.

PR #4996, opened 2022-08-17
`addon.tab.loadScript` is redundant

nexensys:
### Changes

Removed `Tab.loadScript` and `Self.lib` from the addon API and all userscripts. All libraries that do not specify exports should be used with
```js
import "../../libraries/path/to/lib.js";
```

### Reason for changes

`addon.tab.loadScript` is completely unnecessary since global imports can be used.

### Tests

All changed addons work exactly the same.


apple502j:
Actually, this causes the files to be parsed under strict mode. Which I assume is fine, but there is a difference.

WorldLanguages (changes requested):
Blocking merge until we are 100% sure this won't create any problems

ErrorGamer2000:
https://gomakethings.com/side-effects-in-es-modules-with-vanilla-js/

That aside, angular actually uses this same method instead of having the developers add a script tag. And I tested all of the addons that I changed before making this PR.

addons/debugger/performance.js
```diff
@@ -1,10 +1,9 @@
+import "../../libraries/thirdparty/cs/chart.min.js";
```
GarboMuffin, line 1: It may be better for this to be a dynamic import(). A lot of the time, this module is imported but never used, so statically importing chart.js (which takes a non-trivial amount of time) is unnecessary.

Samq64:
While this should be done at some point the PR has many conflicts and there are likely more uses added since.

PR #5025, opened 2022-08-23
Fix editor-devtools not styling properly in darkmode

TheColaber:
I know this isn't the place to put this, but at some point, we should not have dark mode be styling elements created by other addons like this. Instead, we would have a scratch class that you would add to elements that just set their background to #fff or similar and the darkmode addon would change that class' background. The code would be cleaner and shorter in the long run. This would work for addons such as devtools (middle click menu), find-bar, hide-flyout, gamepad (which currently does not have darkmode support in the editor, but I think that was on purpose?), onion skinning, and a lot more but I won't list them all.


apple502j (approved):

mxmou:
> I know this isn't the place to put this, but at some point, we should not have dark mode be styling elements created by other addons like this. Instead, we would

reference the dark mode CSS variables directly in the addon's userstyle.

WorldLanguages (approved):

PR #5026, opened 2022-08-23
README changes

Weredime:


mxmou:
I think the few people who use Firefox forks know that they can install Firefox extensions.

PR #5028, opened 2022-08-23
studio-followers: fixes

Weredime:
### Changes
Actually close properly when disabled
Don't let text be highlighted, because it looks bad

### DIFF
--- a/addons/studio-followers/userscript.js
+++ b/addons/studio-followers/userscript.js
@@ -92,7 +92,7 @@ export default async function ({ addon, global, console, msg }) {
     { passive: true }
   );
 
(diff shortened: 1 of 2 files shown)


addons/studio-followers/userscript.js
```diff
@@ -92,7 +92,7 @@ export default async function ({ addon, global, console, msg }) {
     { passive: true }
   );
 
-  addon.self.addEventListener("disabled", () => (modal.style.display = "none"));
+  addon.self.addEventListener("disabled", () => (backdrop.style.display = "none"));
```
mxmou, line 95: You can just call the `close()` function provided by the modal API.

mxmou (approved):

apple502j (approved):

PR #5049, opened 2022-08-29
Add dynamic enable and disable to popup addons

TheColaber:
It resolves nothing, it's just good to know that these are dynamic enable and disable already since they have persistent scripts/popups

### DIFF
--- a/addons/cloud-games/addon.json
+++ b/addons/cloud-games/addon.json
@@ -81,5 +81,7 @@
     "isMajor": true
   },
   "versionAdded": "1.1.0",
-  "enabledByDefault": true
+  "enabledByDefault": true,
+  "dynamicEnable": true,
+  "dynamicDisable": true
 }
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -62,5 +62,7 @@
       "name": "lisa_wolfgang",
       "link": "https://scratch.mit.edu/users/lisa_wolfgang/"
     }
-  ]
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true
 }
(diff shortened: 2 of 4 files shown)


RedGuy12 (approved):

apple502j (approved):

WorldLanguages:
Well, this kinda made dynamicEnable and dynamicDisable lose their meaning 🤔

TheColaber:
> Well, this kinda made dynamicEnable and dynamicDisable lose their meaning 🤔 

Doesn't it mean that these addons can be dynamically enabled and disabled?

WorldLanguages:
Dynamic enabling/disabling only makes sense for userscripts and userstyles.
Addons that don't have any of these two don't have any dinamicity we can toggle. Addon popups are dynamic by design, and can't be made non-dynamic.
So adding these properties to these addons is weird, but ok, I guess it's fine...

Samq64:
> Dynamic enabling/disabling only makes sense for userscripts and userstyles. Addons that don't have any of these two don't have any dinamicity we can toggle. Addon popups are dynamic by design, and can't be made non-dynamic. So adding these properties to these addons is weird, but ok, I guess it's fine...

I think they should be true by default and tell new contributors to set them to false for their first addon, that way we can get rid of 2 lines from almost every addon.json.

WorldLanguages:
@Samq64 The default can't be `true`, since userscripts need extra code for dynamic disable to work. Addons need to opt-in into dynamic disable

Weredime:
Bug: When you open `scratch-messaging` in a new tab and disable it, it still works.

TheColaber:
I don't think it should stop working there, that would be weird. As long as you can't access in the pop up, it isn't considered disabled.

PR #5063, opened 2022-08-31
Update gitignore to include DS_Store

### LINKED ISSUE #5045: Add `.DS_Store` to gitignore
https://scratch.mit.edu/users/World_Languages/#comments-221456612
I am suggesting to add `.DS_Store` to the gitignore file. Maybe even add more things to the gitignore if necessary.
https://en.wikipedia.org/wiki/.DS_Store

This file is automatically generated by the Finder application on macOS computers. These files can sometimes come in git repositories when Mac users contribute. They are probably unnecessary for SA and also the Finder application doesn't even show these files, even when the option to show hidden files is on. The only way to see it is to use a different application like VS Code or `git status`. According to Wikipedia those files are also frequently changed so it would add more things to the commit log.

Do you think mac users should just add it to the global gitignore file on their device instead? Or do you think some people who are new to git would benefit from it being in the SA gitignore file?

webdev03:
Resolves #5045

### Reason for changes
Discussed in #5045

### Tests

It should work if git is working

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 # Files created during packaging
 .manifest.js.bak
 .dist
+# Automatically generated file by macOS
+.DS_Store


Chiroyce1 (approved):

apple502j:
Will make a new PR with more files myself

Norbiros:
Maybe you can add `.idea` too?

WorldLanguages:
> Will make a new PR with more files myself

Was the PR created already did I miss it? Or is it still pending?

Hans5958:
For the note: Superseeded with #5096.

PR #5065, opened 2022-08-31
New addon: Disable warning not save (2)

valmontechno:
I formatted the code with Prettier compared to my last pull request.

Disables warning when exiting or refreshing an unsaved project.

### DIFF
--- a/addons/disable-warning-not-save/addon.json
+++ b/addons/disable-warning-not-save/addon.json
@@ -1,0 +1,25 @@
+{
+  "name": "Disable warning not save",
+  "description": "Disables warning when exiting or refreshing an unsaved project.",
+  "tags": ["editor", "danger"],
+  "versionAdded": "1.19.0",
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "credits": [
+    {
+      "name": "valmontechno",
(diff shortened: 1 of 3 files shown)


TheColaber:
When would this be useful?

valmontechno:
This is useful with the addon that disables autosave to test simple code without saving it and to use only the costume editor.

addons/disable-warning-not-save/addon.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "Disable warning not save",
```
Norbiros, line 2: Maybe we can change name a little bit?
For example `Disable warning about closing not saved project` or something like that.

Samq64: `Disable leave editor warning`?

Norbiros:
@valmontechno Can you enable actions on your fork [here](https://github.com/valmontechno/ScratchAddons/actions)?

apple502j:
Too dangerous IMO.

apple502j:
@WorldLanguages Can you say "rejected"

BroJac5246:
It could be marked "dangerous," but who would use it?

WorldLanguages:
Rejected until we decide otherwise on an issue.

WorldLanguages:
@valmontechno If you really want this to be added, create an issue and explain the motivation in detail

PR #5067, opened 2022-08-31
Update issue forms

### LINKED ISSUE #5061: Hide "I understand" checkbox
### READ BEFORE CREATING!

- [X] I understand

The "I understand" checkbox with the "READ BEFORE CREATING!" header and no other text clutters the issue and looks little silly to anyone viewing the issue after it is created.

### What this change does

Hiding the checkbox after the issue is created. (This can already be done through editing since the template only appears the first time)

### Possible alternatives

Removing it.

image

Samq64:
Resolves  #5061

### Changes

Updates the issue forms:

- Removes the "I understand" checkboxes
- Removes `required: false`
- Adjusts a couple little things
- ~Deletes the new addon form, since it was pretty much exactly the same as the enhancement one~

### Tests

Untested


DNin01 (approved):
Thanks for correcting the checkbox problem. Now I know about that for next time.

And I don't mind the removal of the "new addon" form too much. While for me, it felt useful to have new addon ideas be automatically labeled, triaging them all is not that bad and that was one almost identical form, now out of the way.

apple502j (approved):

Samq64:
While I'm at, do you want me to add a spot to add enabled addons? This would make more scene after the user has an easier way to get that, but it can be done now.

DNin01:
> While I'm at it, do you want me to add a spot to add enabled addons? This would make more scene after the user has an easier way to get that, but it can be done now.

Maybe. I like when we get to see those just in case.

TheColaber:
> Deletes the new addon form, since it was pretty much exactly the same as the enhancement one

it was added so github could automatically add the "new addon" tag. Please add it back.

Samq64:
> it was added so github could automatically add the "new addon" tag. Please add it back.

@TheColaber Done.

TheColaber (approved):

PR #5079, opened 2022-09-03
Separate notification dot from debugger icon

mxmou:
### Changes

Makes the notification dot on the debugger icon a pseudo-element instead of using a different icon. This prevents it from becoming white in dark mode:
image

It also changes the color depending on the settings:
image (1.x preset)

### Tests

Tested with different dark mode presets.


apple502j (approved):

Hans5958 (approved):
LGTM

PR #5081, opened 2022-09-04
Clarify some strings

### LINKED ISSUE #5074: Improve the setting name "Use previous color instead of default when switching tools"
This is a text change.

### What I'm talking about

[**"Use previous color instead of default when switching tools"**](https://github.com/ScratchAddons/ScratchAddons/blob/1a40713a6414f13df7e65f3b0584a4b06a540edf/addons/default-costume-editor-color/addon.json#L37) from the `default-costume-editor-color` addon's settings. If you're wondering what that means, if the fill color is set to transparent and you select the brush tool, Scratch, by default, resets the color to purple so you have a color to draw with (and something similar happens with the outline color). This addon option makes it so that instead of being switched back to the default color, it changes back to the previous fill or outline color you used.

This option may take some time for some people to understand, but it can be pretty hard to choose a concise name for that option and have everyone understand it. Should we try to find a better name?

### Possible changes
My ideas are:
- Recall previous color instead of resetting after switching tools
- Don't reset to default color after switching tools

This is worth discussing _thoroughly_ - I'd love if we could work together to find _the_ most easy-to-understand but still concise description possible for that option, especially since Scratch Addons doesn't have extended option descriptions (yet).

### Don't understand what I mean? Let me show you

https://user-images.githubusercontent.com/106490990/188237242-68f52189-816e-46c2-932f-030bab24adba.mp4

DNin01:
Resolves #5074

### Changes

Changes the setting name "Use previous color instead of default when switching tools" → "Recall previous color instead of resetting after switching tools", plus a few other edits ranging from corrections to clarifications.

### Reason for changes

To make it clearer what that option does.

### Tests

No, but it should work - it's just text.

### DIFF
--- a/addons/default-costume-editor-color/addon.json
+++ b/addons/default-costume-editor-color/addon.json
@@ -34,7 +34,7 @@
       "default": 4
     },
     {
-      "name": "Recall previous color instead of resetting after switching tools",
+      "name": "Use previous color instead of resetting after switching tools",
       "id": "persistence",
       "type": "boolean",
       "default": true
--- a/addons/default-costume-editor-color/addon.json
+++ b/addons/default-costume-editor-color/addon.json
@@ -34,7 +34,7 @@
       "default": 4
     },
     {
-      "name": "Use previous color instead of default when switching tools",
+      "name": "Recall previous color instead of resetting after switching tools",
       "id": "persistence",
       "type": "boolean",
       "default": true
--- a/addons/editor-colored-context-menus/addon.json
+++ b/addons/editor-colored-context-menus/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Colored context menus",
-  "description": "Makes block right click context menus colorful.",
+  "description": "Makes the background of block right click menus colorful.",
   "credits": [
     {
       "name": "GarboMuffin"
--- a/addons/editor-colored-context-menus/addon.json
+++ b/addons/editor-colored-context-menus/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Colored context menus",
-  "description": "Makes the background of block right click context menus colorful.",
+  "description": "Makes block right click context menus colorful.",
   "credits": [
     {
       "name": "GarboMuffin"
--- a/addons/editor-colored-context-menus/addon.json
+++ b/addons/editor-colored-context-menus/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Colored context menus",
-  "description": "Makes the background of block right click menus colorful.",
+  "description": "Makes the background of block right click context menus colorful.",
   "credits": [
     {
       "name": "GarboMuffin"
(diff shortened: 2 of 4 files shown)


addons/move-to-top-bottom/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Move costume to top or bottom",
-  "description": "Adds a right click context menu item to move a costume or a sound to the top or the bottom of the list.",
+  "description": "Adds buttons to the right click context menu for costumes and sounds to move them to the top or the bottom of the list.",
```
TheColaber, line 3: which do we use more? "right click context menu" or "context menu"?

DNin01: I see [1](https://github.com/ScratchAddons/ScratchAddons/blob/4d70ad6f65f32c47585a29060b6d20bc4d2837e3/addons/hide-delete-button/addon.json#L3), [2](https://github.com/ScratchAddons/ScratchAddons/blob/a3c035b087962391eef6747fc4d91c9ae2d6025e/addons/editor-colored-context-menus/addon.json#L3) other uses of "right click context menu", and I don't see a use of "context menu" alone. IMO, we can leave this as it is, and changing it might work best in a new PR.

RedGuy12 (approved):

addons/move-to-top-bottom/addon.json
mxmou, line 3: I think we should call menu items "options", not "buttons".

addons/editor-colored-context-menus/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Colored context menus",
-  "description": "Makes block right click context menus colorful.",
+  "description": "Makes the background of block right click context menus colorful.",
```
mxmou, line 3: If the "black" preset for "customizable block colors" is selected, this addon changes the border and text color, not the background.

DNin01: Should I revert it, then?

apple502j, line 3:
```suggestion
  "description": "Makes block right click context menus colorful.",
```

addons/default-costume-editor-color/addon.json
```diff
@@ -34,7 +34,7 @@
       "default": 4
     },
     {
-      "name": "Use previous color instead of default when switching tools",
```
apple502j, line 37: "Recall" is not usually used in this context.

addons/default-costume-editor-color/addon.json
```diff
@@ -34,7 +34,7 @@
...
+      "name": "Recall previous color instead of resetting after switching tools",
```
apple502j, line 37:
```suggestion
      "name": "Use previous color instead of resetting after switching tools",
```

apple502j (approved):

apple502j:
Requesting one more approval.

mxmou (approved):

Issue #1621, opened 2021-02-18
Addons that use the key property


TheColaber:
We have had the ability to make addons that use hotkeys but have not yet made any. We would need:

- [ ] APIs for these things
- [ ] Addons to use these

PR #5086, opened 2022-09-05
[New addon] Player controls keyboard shortcuts

ego-lay-atman-bay:
### Changes

I've created a new addon. Project controls keyboard shortcuts (probably need a better name).

I've put in the combo customization in the settings. Not sure if that's going to stay.

I made sure that the shortcuts block the browser shortcuts, that way, whenever you press ctrl+P, it doesn't show the print dialog, if that's set to a shortcut. If it's not, it'll show the print dialog.

### Reason for changes

It was suggested, and I had already created my own custom snippet to do it.

### Tests

I've only tested it on Chrome (Brave) on windows 10.
I also made sure that it will not break when the pause button addon is off.

### Notes
The way I did the keyboard combo refresh from the settings, might not be the best way. I just couldn't find any documentation on the addon api, so I opted to use probably an unconventional method.

I only made the keys customizable because I couldn't think of good keyboard combos.

And also, the description needs work. I also don't know if we want to stick to just using the ctrl key, or allow other keys.

This obviously needs work before it's ready.


ego-lay-atman-bay:
wait, I just found the documentation. I'm going to read through it.

ego-lay-atman-bay:
Alright, I cleaned up the code a bit, and fixed `addon.json`. The only problem that I could see in the Prettier format check was this
```
No node_modules/ folder.
No package-lock.json file.
[detached HEAD 1d5841b] Format code
 Author: ego-lay-atman-bay <ego-lay-atman-bay@users.noreply.github.com>
 2 files changed, 84 insertions(+), 87 deletions(-)
fatal: You are not currently on a branch.
To push the history leading to the current (detached HEAD)
state now, use

    git push origin HEAD:<name-of-remote-branch>

Error: Process completed with exit code 128.
```
I don't know how to fix this.

Norbiros:
You need to go to your fork and enable actions (https://github.com/ego-lay-atman-bay/ScratchAddons/actions here)

ego-lay-atman-bay:
> You need to go to your fork and enable actions (https://github.com/ego-lay-atman-bay/ScratchAddons/actions here)

Thanks!

mxmou:
In Scratch 1.4, Enter was a shortcut for the green flag. I think it should be possible to choose it in this addon.

addons/project-shortcuts/userscript.js
```diff
@@ -0,0 +1,42 @@
+export default async function ({ addon, global, console }) {
+  var green_flag = addon.settings.get("greenflag").toLowerCase();
```
Norbiros, line 2: Please use `let`.

TheColaber:
We are planning to add an entire section to the settings page for this, or maybe not, depends how it goes. But the reason why I'm closing this is that we don't have a current system put in place for hotkeys https://github.com/ScratchAddons/ScratchAddons/issues/1621 that is crucial. The plan is for the settings page to have a bunch of settings to add hotkeys for almost anything in the editor, costume editor, etc, and possibly for some non-project pages. Maybe we'll look at this again after we start some sort of work on this kind of system.

ego-lay-atman-bay:
Ah, ok. I'm excited for the new hotkey system.

WorldLanguages:
Discussion for hotkeys/keybind system is happening at #5145

PR #5087, opened 2022-09-05
Fix block-cherry-picking and block-duplicate race

GarboMuffin:
Fixes https://discord.com/channels/806602307750985799/1015981428472811520/1016169173514211421

> Has anyone noticed the Ctrl + Alt drag to duplicate a single block is bugged? It is currently leaving the original block detached from the script after duplicating
> https://cdn.discordapp.com/attachments/1015981428472811520/1015982258844356679/2022-09-04_14-50-06.mp4

The addons trapped the same method and would break if they ran in the wrong order (regression from https://github.com/ScratchAddons/ScratchAddons/pull/4682). They now use a shared module to ensure consistent order.

### DIFF
--- a/addons/block-cherry-picking/userscript.js
+++ b/addons/block-cherry-picking/userscript.js
@@ -1,12 +1,12 @@
-import * as sharedModule from '../block-duplicate/module.js';
+import * as sharedModule from "../block-duplicate/module.js";
 
 export default async function ({ addon, global, console }) {
   const update = () => {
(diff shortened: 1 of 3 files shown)


WorldLanguages:
Should we add a bit of intentional randomness to userscript order? (only for runAtComplete:false userscripts)
I think would allow us to catch bugs like this faster.

GarboMuffin:
I've only tested this on a Mac because that's the only computer I have access to at the moment. It *probably* stills works on other systems

WorldLanguages (approved):
I don't think I was able to reproduce the bug. But it continues to work for me

PR #5093, opened 2022-09-10
Project folders

Platformer444:
Resolves #

### Changes
It adds project folders to organize your projects

### Reason for changes
It makes finding your projects easier

### Tests
Tested on Edge 105 and it works fine


Norbiros:
It works with search bar in my stuff?

Platformer444:
> It works with search bar in my stuff?

Didn't check that

WorldLanguages:
Storing the project folders in localStorage is unideal. Users might not be aware that clearing cookies for scratch.mit.edu will reset their project folders (or switching to another device - or using incognito mode).
An addon like this will need proper discussion first, before we should try to implement it.

Platformer444:
> Storing the project folders in localStorage is unideal. Users might not be aware that clearing cookies for scratch.mit.edu will reset their project folders (or switching to another device - or using incognito mode). An addon like this will need proper discussion first, before we should try to implement it.

I first thought to store folder data in a json file situated in the addon folder itself. I thought it as a bad idea so went with localStorage.

WorldLanguages:
I think it would be better to remove the concept of "folders" completely. I don't think users might be able to conclude that the folders are stored in the extensions, and not in their Scratch account. For example, the "name scripts before placing in backpack" addon could also work with extension or local storage, but I think we only added the feature because it's stored in the account.

If we instead allowed users to "filter by hashtag", the data would be stored in the project instructions/notes, and wouldn't be lost if SA is uninstalled. We could provide buttons that allow the user to edit the project description from the mystuff page in some way, so that they can avoid manually adding the hashtags on each project page.

I think this needs more discussion in an issue, instead of this PR.

Chiroyce1:
Agreed with @WorldLanguages, sprite folders are practical because they are stored in the project data itself, and other devices and SA users can also access it, whereas over here it is all gone if the user clears cookies or uses another device/browser.

WorldLanguages:
Closing - this needs discussion first.

RedGuy12:
> I first thought to store folder data in a json file situated in the addon folder itself. I thought it as a bad idea so went with localStorage.

litterally impossible

PR #5104, opened 2022-09-13
Make logo larger, stretch group divider to align with box

Hans5958:
Primarily, this PR does two things:
1. Make the logo larger to make it more prominent. The size is now as large as the website.
2. Stretch the addon group divider to align with the box and other items, not the inside contents. This is an old issue and I wanted to push this again.

The PR also does some small styling changes on other parts. (prefer flex, prefer rem, etc).

Screen Shot 2022-09-13 at 13 43 38
Screen Shot 2022-09-13 at 12 52 11

### DIFF
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -19,6 +19,7 @@ body {
 #header {
   display: flex;
   height: 60px;
+  padding: 0 1rem;
 }
 #title-text,
 #settings {
@@ -29,15 +30,14 @@ body {
   flex-grow: 1;
   display: flex;
   align-items: center;
-  padding: 0 20px;
 }
 #logo {
   height: 30px;
-  margin-inline-end: 20px;
+  margin-inline-end: 0.75rem;
+  transform: scale(1.75);
   vertical-align: middle;
 }
 #settings {
-  padding: 0 20px;
   cursor: pointer;
   display: flex;
   align-items: center;
--- a/webpages/settings/components/addon-group-header.html
+++ b/webpages/settings/components/addon-group-header.html
@@ -15,7 +15,6 @@
     align-items: center;
     cursor: pointer;
     margin: 0px 11px;
-    padding: 0px 10px;
   }
 
   .addon-group > img {
@@ -37,7 +36,6 @@
     border-top: 1px solid var(--content-separator);
     width: 100px;
     margin-inline-start: 20px;
-    margin-inline-end: 5px;
     flex: 1;
   }
 </style>
--- a/webpages/settings/index.html
+++ b/webpages/settings/index.html
@@ -122,7 +122,7 @@
       <div id="title">
         <img :src="switchPath" class="toggle" @click="sidebarToggle()" v-cloak v-show="smallMode" alt="Logo" />
         <img src="../../images/icon-transparent.svg" class="logo" alt="Logo" />
-        <h1 v-cloak>{{ msg("settings") }}</h1>  
+        <h1 v-cloak>{{ msg("settings") }}</h1>
       </div>
       <img v-cloak @click="setTheme(!theme)" class="theme-switch" :src="themePath" />
     </div>
--- a/webpages/settings/index.html
+++ b/webpages/settings/index.html
@@ -119,9 +119,11 @@
   </head>
   <body v-cloak :class="{light: theme}">
     <div class="navbar">
-      <img :src="switchPath" class="toggle" @click="sidebarToggle()" v-cloak v-show="smallMode" alt="Logo" />
-      <img src="../../images/icon-transparent.svg" class="logo" alt="Logo" />
-      <h1 v-cloak>{{ msg("settings") }}</h1>
+      <div id="title">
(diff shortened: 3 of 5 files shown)


Samq64:
Could you just crop the transparent logo instead? I was lazy on #4810 and just removed the background leaving whitespace instead of just resizing the image correctly.

Hans5958:
To be honest, resizing it up to the image borders wouldn't be big enough to my liking. I would still need to use CSS' `scale`. I think this is fine for now.

Samq64:
Really? After cropping the whitespace from [icon-transparent.svg](https://GitHub.com/ScratchAddons/ScratchAddons/images/icon-transparent.svg) with Inkscape on master it looks pretty much the same as yours:
Screenshot_20220913_125454

GrahamSH-LLK:
👎
It doesn't need to be that big. That just looks weird. There's no point in changing this just to change it.

WorldLanguages:
I think changing the logo size could be moved to a separate PR, so we can discuss that separately and merge obvious fixes first.

> Stretch the addon group divider to align with the box and other items, not the inside contents.

Wow, nice catch. Maybe this was intentional, but I can't unsee it now.

Hans5958:
> Really? After cropping the whitespace from [icon-transparent.svg](https://github.com/ScratchAddons/ScratchAddons/images/icon-transparent.svg) with Inkscape on master it looks like this:

Huh. I thought you were thinking to not scale it altogether,

> It doesn't need to be that big. That just looks weird. There's no point in changing this just to change it.

I can see why it is quite shocking, looking that it is been how it is for a while. I too also feel weird at first. Keep in mind that it is how it's been on the website. This is how the website looks with the extension-styled theme, but the size is the same on the normal theme.

![](https://user-images.githubusercontent.com/11584103/190098364-7743283b-f088-462e-81eb-9a19ecae5b39.png

The mentality of "changing [this] just to change it" seems like a bad mentality, especially that this is an open source project, and this PR suggestion which can be discussed.

> Wow, nice catch. Maybe this was intentional, but I can't unsee it now.

This is actually a choice of mxmou, which aligns the divider with the inner contents of the addon box, not the whole box itself. I tried to find the issue but I haven't found it yet.

Hans5958:
> The extension UI doesn't contain large text or images which makes the larger logo look unbalanced - it feels like the logo wants to be the most important thing on the screen. 

I disagree. I think these kind of brand logo, especially if placed on the navigation bar, should be (slightly) bigger in comparison with the sizes of other elements (e.g. font size). The logo is already on the prominent place, being in the top-left corner of the website, and a simple fact that it is an image, and it is related to the brand, so undermining it by making it the smaller is not good. I also think being "most important thing on the screen" is an overstatement.

With the current state of the settings, I don't feel a sense of "balance", but that's just me. Inconsistent shadowing, lengthy page, etc.

Here's a comparison. The size of the website's logo is based on Bootstrap 4's website, so it is included for reference. For the note that I would size it based on the visible boundary, so the inner part (the S with the addon block) size on the dark theme would be smaller than the size on the light theme, because I base the logo with the outer boundary (the orange circle) on the dark theme, and the inner part on the light theme.



There are some navbar that has different sizes of logo. Typescript has a slim kind of navbar, but compared to the website and the extension which is big enough. The Scratch logo on its website has 30px as its weight, but that's compared to the smaller links on its side.

> I don't think we should switch to rem units because pixels are currently used everywhere in the code - if something is consistent and doesn't cause problems it should probably stay.

Good idea. I just realised that lots of the font sizes would have different values than the website. Sora somehow has a different size than the usual fonts, so last time I implement the extension-styled theme, I use `font-size-adjust` instead.

> Was the change to the position of the small mode menu icon intentional?

I would take a look at this again. I changed it so it doesn't necessarily align with something, but I will take another look.

GrahamSH-LLK:
My thoughts are this:
It doesn't need to be that big. It looks huge and out of place in the current settings page design. We don't need exact design parity with the site, nobody really cares anyway. It doesn't fit with the current text size (right now the scratch S is about the same size as the S in settings). The biggest one just looks awful in the design and mxmou's looks equal to current, but not necessarily better.

Hans5958:
> It doesn't fit with the current text size (right now the scratch S is about the same size as the S in settings). 

Again, IMO it shouldn't be the same, since it is an image, or a branding, or both. Read the first paragraph on the previous message.

> In addition, there is a large difference between the use of the logo on the site and the use here. On the site, it's a click target and brings you home. It makes sense to make it large. However, on the extension, it does nothing. It's simply showing branding.

I made it bigger with the reason of the logo is there as a branding, not what it functions.

WorldLanguages:
> I think changing the logo size could be moved to a separate PR, so we can discuss that separately and merge obvious fixes first.

I said this and as far as I could see, no one replied (maybe I missed it).
I think we should create a bullet list for each change in this PR, and approve/unvote each one separately.

DNin01:
I like how it looks when slightly bigger in the popup. I'm not sure about other places.

WorldLanguages:
Related: #7485

PR #5114, opened 2022-09-18
Rename a few addon IDs

Samq64:
Resolves nothing

### Changes
Renames these addon IDs:

- `comments-linebreaks` --> `comment-linebreaks`
- `editor-theme3` --> `custom-block-color`
- `discuss-button` --> `navbar`

Unfortunately, two of the files weren't counted as a rename, but the rest are fine.

### Reason for changes

The grammar mistake in `comments-linereaks` bugs me and the other two don't describe when the addon does at all. I was going to rename a few more, but decided to just stick to the ones hat really needed it.

I know this might be rejected, but I thought I would give it a try.

### Tests

Tested on Chromium 105.


RedGuy12:
This would reset users' settings.

Samq64:
I could do some migration code, but yeah, it's probably not worth it.

Hans5958:
The arch nemesis of Scratch Addons. I would also like to change some IDs but people really didn't want to invest to migration.

However, I would say that this is a good idea for the future major version (SA2), so you can mention it again on #5103 for another discussion.

PR #5115, opened 2022-09-19
New Addon: `project-folders`

Platformer444:
Resolves #

Changes
It adds project folders to organize your projects

Reason for changes
It makes finding your projects easier

Tests
Tested on Chrome 105 and it works fine


addons/project-folders/userscript.js
```diff
@@ -0,0 +1,191 @@
...
+          "x-token": token,
+        },
+      })
+    ).json();
+
+    return projectDetails;
+  }
+
+  async function load() {
+    const foldersJSON = { folders: [] };
```
GrahamSH-LLK, line 38: Why this structure? Couldn't this just be an array instead of being nested in an object?

addons/project-folders/userscript.js
```diff
@@ -0,0 +1,191 @@
...
+    noFolderSpan.textContent = msg("loading");
+    folderDiv.appendChild(noFolderSpan);
+
+    const tempBR = document.createElement("br");
+    folderDiv.appendChild(tempBR);
+
+    const loader = document.createElement("progress");
+    loader.setAttribute("value", "0");
+    loader.setAttribute("max", "100");
+    folderDiv.appendChild(loader);
```
GrahamSH-LLK, line 52: Does this load slow enough to require a progress bar? If so it should probably be made more performant

addons/project-folders/userstyle.css
```diff
@@ -0,0 +1,51 @@
...
+  width: 100px !important;
+  border: 1px black solid !important;
+  text-align: center !important;
+  border-radius: 5px !important;
+  height: 90px !important;
+  padding: 5px !important;
+  display: inline-block !important;
+  margin: 10px !important;
+  background-color: #fcfcfc !important;
+  color: black !important;
```
GrahamSH-LLK, line 15: Why is the extra specificity required? Aren't you creating this element yourself?

GrahamSH-LLK (changes requested):

addons/project-folders/userscript.js
Platformer444: I added a progress bar to tell the use the amount of time it is taking. But if the user has less projects then it will take less time but I don't think the performance can be made any better.

addons/project-folders/userstyle.css
Platformer444: I did that because I also added a `.button` class `.folder`. But I will check if `!important` is still required or not

addons/project-folders/userscript.js
Platformer444: I have changed it into an array in the latest commit.

addons/project-folders/userstyle.css
Platformer444: It is fixed in the latest commit

addons/project-folders/userscript.js
```diff
@@ -0,0 +1,190 @@
...
+  const projectTab = document.querySelectorAll("li.first")[0];
+  projectTab.childNodes[0].onclick = async () => {
+    projectColumns = await addon.tab.waitForElement("ul.media-list", {
+      markAsSeen: true,
+    });
+    createFolderAreaAndButton();
+    foldersArray = await load();
+  };
+
+  if (document.querySelectorAll("div.folders-container")[0] === undefined) {
```
Samq64, line 17:
```suggestion
  if (!document.querySelectorAll("div.folders-container")[0]) {
```
The same can be applied elsewhere.

addons/project-folders/userscript.js
```diff
@@ -0,0 +1,190 @@
...
+
+    const loader = document.createElement("progress");
+    loader.setAttribute("value", "0");
+    loader.setAttribute("max", "100");
+    folderDiv.appendChild(loader);
+
+    for (let i = 0; i < projectColumns.childNodes.length - 2; i++) {
+      loader.setAttribute("value", (i / (projectColumns.childNodes.length - 2)) * 100);
+
+      const link = projectColumns.childNodes[i + 2].childNodes[1].childNodes[3].childNodes[1].childNodes[0].href;
```
Samq64, line 57:
```suggestion
      const link = projectColumns.childNodes[i + 2].childNodes[1].childNodes[1].childNodes[1].href;
```
This uses the project link from the thumbnail which should be slightly faster to access.

addons/project-folders/userscript.js
```diff
@@ -0,0 +1,190 @@
...
+      const projectID = link.replace("https://scratch.mit.edu/projects/", "").replace("/", "");
+      const projectDetails = await getProjectDetails(projectID);
+      let instructions = projectDetails.instructions;
+
+      if (instructions.includes("#_")) {
```
Samq64, line 62:
```suggestion
      if (!instructions.includes("#_")) continue;
```
Just remember to remove the closing brace.

addons/project-folders/userscript.js
```diff
@@ -0,0 +1,190 @@
...
+        instructions = instructions.split("\n");
+
+        for (let j = 0; j < instructions.length; j++) {
```
Samq64, line 65: Most of the loops can be replaced with for of loops.

Samq64:
Actually, there is a an API for my stuff, so that should probably be used: https://scratch.mit.edu/site-api/projects/all/

Samq64:
I'm not sure, but I don't think so.

WorldLanguages:
Unfortunately, I'm going to have to move this to the next milestone.

addons/project-folders/userscript.js
```diff
@@ -0,0 +1,138 @@
+export default async function ({ addon, global, console, msg }) {
+  const projects = await (await fetch("https://scratch.mit.edu/site-api/projects/all")).json();
+
+  async function loadFolders() {
+    const folders = [];
+    for (let i = 0; i < projects.length; i++) {
+      const projectID = projects[i].pk;
+      const token = await addon.auth.fetchXToken();
```
Samq64, line 8: Could this be moved above the loop?

addons/project-folders/userscript.js
```diff
@@ -0,0 +1,138 @@
...
+      });
+      if (folder.length === 0) continue;
+
+      for (let j = 0; j < folder.length; j++) {
+        const matches = folders.filter((element) => {
+          if (element.name === folder[j].replace("#_", "")) {
+            return true;
+          } else {
+            return false;
+          }
```
Samq64, line 32:
```suggestion
          return element.name === folder[j].replace("#_", "");
```

Samq64 (commented):
This addon is still pretty slow, so could some kind of loading indicator be re-added?

addons/project-folders/addon.json
```diff
@@ -0,0 +1,32 @@
...
+      "matches": ["https://scratch.mit.edu/mystuff"]
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["https://scratch.mit.edu/mystuff"]
+    }
+  ],
+  "info": [
```
Samq64, line 22: Could a performance notice be added since the addon still takes a while to make all the requests?

GrahamSH-LLK:
I think I forgot to link it here, but https://github.com/Platformer444/ScratchAddons/pull/1

WorldLanguages:
I haven't considered that maybe we should try to reuse code from `search-my-stuff` in this addon, for filtering folders on the My Stuff page.

Joeclinton1:
I tried it myself, and I think this addon should not be accepted and the PR closed. Here's why:

1. **It's unneccessary.** With studios which let you group your projects (albeit quite publicly) and the search-my-stuff addon you can already quite quickly find any project you're interested in.
2. **It's extrememly slow to load.** The loading time is just far too slow, and there's not much you can do about that if you really must have the data for the project folders being stored in project descriptions. 

Now problem #2 can (and should) be fixed by storing data in the addon itself. But that doesn't excuse the main problem that the idea is really un-neccessary given that we can search for projects with search-my-stuff.

GrahamSH-LLK:
> I tried it myself, and I think this addon should not be accepted and the PR closed. Here's why:
> 
> 1. **It's unneccessary.** With studios which let you group your projects (albeit quite publicly) and the search-my-stuff addon you can already quite quickly find any project you're interested in.
> 2. **It's extremely slow to load.** The loading time is just far too slow, and there's not much you can do about that if you really must have the data for the project folders being stored in project descriptions. Also every time you reload the page you have to wait for it to load again. It's so slow, that you could easily find your project just by scrolling in the time it takes to load.
> 
> Now problem #2 can (and should) be fixed by storing data in the addon itself. But that doesn't excuse the main problem that the idea is really un-neccessary given that we can search for projects with search-my-stuff. It's a big, complex addon that will waste resources, for little actual additional functionality.

https://github.com/Platformer444/ScratchAddons/pull/1 fixes the slow loading

Joeclinton1:
@GrahamSH-LLK Wow, that is alot faster. Well, then it comes down to the question of whether the added utility is worth mantaining this addon. I guess with so much work having already gone into it, it wouldn't hurt to see it come to frutition.

Samq64:
@Platformer444 Could you merge https://github.com/Platformer444/ScratchAddons/pull/1? We can take over from there if you're not interested in maintaining this.

TheColaber:
closing due to inactive author. @Samq64 @GrahamSH-LLK feel free to take over

Issue #1095, opened 2020-12-24
Better backpack


j5155:
**Describe the new addon**

The backpack seems pretty clunky right now, it would be nice if it were moved to something like a new blocks tab, such as below the My Blocks tab. Mockup; 
![Screenshot 2020-12-23 184621](https://user-images.githubusercontent.com/54331556/103059643-29737780-4552-11eb-9f59-1b92062258b5.png)
For scripts, having the entire script there and easy to see and use would be nice. For things that aren't scripts, I guess having them similar to the way they are now, but with the name of the costume/sprite next to them, would also work.

Marc92020:
Cool idea! but I don't know how costumes and complete scripts could fit in a category.

scratchusernamemrtbts:
we can have a right click to open a menu that will let us select the complete script from the backpack
and if there is a sprite in the backpack add a new button on the create sprite pop-up menu

j5155:
> Cool idea! but I don't know how costumes and complete scripts could fit in a category
How would scripts not fit in a category? The category might get a little big if you have a lot of things in your backpack, but you could just scroll down.

For costumes, sprites, and audio, just having a thumbnail and the name of the item should be fine, as you won't be putting them directly into the script area.

Marc92020:
> > Cool idea! but I don't know how costumes and complete scripts could fit in a category
> > How would scripts not fit in a category? The category might get a little big if you have a lot of things in your backpack, but you could just scroll down.
> 
> For costumes, sprites, and audio, just having a thumbnail and the name of the item should be fine, as you won't be putting them directly into the script area.

OK it just sounded confusing at first

Explosion-Scratch:
This would be COOL!! Total support! What about for sprites and sounds though? And how would it handle loading backpack assets? I can see that being a potential problem with loading speeds. Also, this could really clutter up the block pallet if each script was represented as a script there, what about simply moving the back there and representing everything as a backpack tile, ready to drag onto sprites?

j5155:
> This would be COOL!! Total support! What about for sprites and sounds though? And how would it handle loading backpack assets? I can see that being a potential problem with loading speeds. Also, this could really clutter up the block pallet if each script was represented as a script there, what about simply moving the back there and representing everything as a backpack tile, ready to drag onto sprites?

For sprite and sounds, I was thinking of leaving them how they are in the backpack, and just moving their location. 

[...]

PR #5120, opened 2022-09-21
New addon: Multiple rows in backpack

### LINKED ISSUE #5056: `editor-dark-mode` incorrectly colors certain text in light themes
### READ BEFORE CREATING!

- [X] I understand

### Description

The Editor dark mode and customizable colors addon does not correctly set the text color for Code tab context menus and selected costumes/sounds to what their color would be on the highlight color.

### Steps to reproduce

1. Enable Editor dark mode and customizable colors
2. Enable the Scratch 2.0 or Scratch's default colors presets
3. Right click in the Code tab or select a costume/sound in their tabs
4. See bug

### Environment

Browser: Firefox 105.0b4
Scratch Addons version: v1.28.0
Operating system: Windows 10

pumpkinhasapatch:
Resolves #5056 and [many](https://discord.com/channels/806602307750985799/806609030935740477/1019298147039723532) [old](https://discord.com/channels/806602307750985799/806609030935740477/1018233224138334208) [suggestions](https://discord.com/channels/806602307750985799/806609030935740477/1023716288121556993) from the SA community Discord server.

### Changes

The main purpose of this addon is to reinvent the layout and behavior of the Backpack menu in the Scratch project editor. It allows multiple rows of items in the backpack to be displayed in a neat grid, (similar to the Costume library) and changes the horizontal scrollbar to a mouse-friendly vertical one for easy scroll wheel use. It does this through simple CSS tweaks to [this file](https://github.com/LLK/scratch-gui/blob/develop/src/components/backpack/backpack.css) loaded by the scratch-gui.

__Video preview:__ [recording-2022-09-21_14.03.46.webm](https://user-images.githubusercontent.com/113677241/191446313-e8cf7d65-b98f-415d-b22d-f6be02383ead.webm)

It can also improve the look of backpack thumbnails. By default, Scratch saves these icons as 96x96 pixel jpegs of the costume/script with solid white backgrounds, but resizes them down to 32x32 pixels when displayed. The addon remove this limit, allowing them to span up to the edges of the button and appear larger without losing quality.

unknown

Another feature of the addon is that it can fill the background around thumbnails to hide the unpleasant white artifacts caused by the jpeg conversion of costumes and script previews with transparent backgrounds. This is most noticeable when the `Editor Dark mode and Customizable colors` addon is enabled, as it drastically changes the thumbnail background color. The default Scratch theme does this seamlessly by using a very light gray background, close to the solid white used in the thumbnails, and [mix-blend-mode](https://www.w3schools.com/cssref/pr_mix-blend-mode.asp) multiplying to do the rest, but using black or other background colors through a custom theme ruins the effect.

Dark mode without Expanded backpack addon:
Screenshot_2022-09-20_17-52-12
Larger previews and background fill enabled:
Screenshot_2022-09-20_17-39-36

### Reason for changes

The backpack's design has been left untouched by Scratch developers since it's introduction during the Scratch 2.0 era, and is in need of some changes to improve user accessibility. Scratch Addons also lacks significant modifications to the backpack, only including the `Name scripts before placing in Backpack` addon that prompts the user to rename code snippets before uploading them to the backpack server.

Expanded backpack provides much more customization and accessibility than the regular Scratch backpack. Increasing the size and number of backpack items that can be displayed at once would be very benificial for those who use the backpack a lot, or have a lot of content in their backpack and are trying to find a specific item they're looking for. The user can configure or disable each part of the addon, such as changing the backpack height by selecting between 1-3 rows of items to be displayed at once, using default or large size thumbnails and toggling the white background fill behind such images.

image

### Tests

The proposed addon has been thoroughly tested on both Chromium 105 and Firefox 100 using the SA v1.27.3 release source code, and has been revised to match the current v1.29.0-pre master branch. Feel free to [download my branch](https://github.com/pumpkinhasapatch/ScratchAddons/archive/refs/heads/expanded-backpack.zip) and try the addon out yourself, or suggest an edit to the script if you find a bug or need to tweak the addon.json.


addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,73 @@
...
+  "name": "Expanded backpack",
+  "description": "Adds multiple rows, better previews and vertical scrolling to the editor backpack to make it easier to find things.",
+  "credits": [
+    {
+      "name": "pumpkinhasapatch",
+      "link": "https://github.com/pumpkinhasapatch"
+    }
+  ],
+  "versionAdded": "1.27.3",
+  "tags": ["editor"],
```
RedGuy12, line 11: "theme" tag?

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "Expanded backpack",
+  "description": "Adds multiple rows, better previews and vertical scrolling to the editor backpack to make it easier to find things.",
+  "credits": [
+    {
+      "name": "pumpkinhasapatch",
+      "link": "https://github.com/pumpkinhasapatch"
+    }
+  ],
+  "versionAdded": "1.27.3",
```
Norbiros, line 10:
```suggestion
  "versionAdded": "1.29.0",
```

addons/expanded-backpack/addon.json
mxmou: The purpose of this addon is to improve usability of the backpack, not the appearance.

addons/expanded-backpack/backpack.css
```diff
@@ -1,21 +1,21 @@
...
 .backpack_backpack-list-inner_10a2A {
     /* move backpack items to the top left instead of vertically centering them */
-    -webkit-align-items: start !important;
-        -ms-align-items: start !important;
+    -webkit-align-items: start;
+        -ms-align-items: start;
     /* force horizontal wrapping and enable vertical scrolling */
-    flex-wrap: flex;
-    overflow-x: hidden !important;
+    flex-wrap: wrap;
```
pumpkinhasapatch, line 11: fixed invalid value causing the buttons to not wrap at all

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "Expanded backpack",
```
apple502j, line 2: Please rename this addon.

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,64 @@
...
+  "description": "Adds multiple rows, better previews and vertical scrolling to the editor backpack to make it easier to find things.",
```
apple502j, line 3:
```suggestion
  "description": "Adds multiple rows, better previews and vertical scrolling to the editor backpack.",
```

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,64 @@
...
+      "name": "pumpkinhasapatch",
+      "link": "https://github.com/pumpkinhasapatch"
+    }
+  ],
+  "versionAdded": "1.29.0",
+  "tags": ["editor"],
+  "enabledByDefault": false,
+  "settings": [
+    {
+      "name": "Number of rows to display in backpack",
```
apple502j, line 15:
```suggestion
      "name": "Number of displayed rows",
```

apple502j (changes requested):
Opening a backpack is usually called "expanding" (vs collapsing). The current name sounds like it opens the backpack by default.

addons/expanded-backpack/addon.json
pumpkinhasapatch: What name were you suggesting? Extended backpack, Better backpack?

CubesterYT: I'd personally say "Advanced Backpack" or "Expandable Backpack".

apple502j: Don't use an adjective. Describe the functionality.

WorldLanguages:
@pumpkinhasapatch Nice first PR! Thanks for the detailed PR body.

addons/expanded-backpack/addon.json
fractioneater: What about "Larger Backpack" or "Increased Backpack Size"?

pumpkinhasapatch: then again that might confuse users into thinking that it increases the amount of items you can store in the backpack, rather than making it easier to find stuff in it
it has other features like improving the look of the costume and script icons which don't really fit the title description
i chose that name because its a general addon that "expands" the functionality of the backpack by adding a few different useful features but am not exactly sure what we should call it

was also planning on having basic sort/filter buttons above the items inside the backpack (like the costume library) to show only costumes, sounds, scripts or sprites and not the others, but that would require quite a bit of javascript to show/hide each item depending on what kind it is and that im not familiar enough with js to write such a thing yet

pumpkinhasapatch:
whoops didnt mean to do that

pumpkinhasapatch (commented):
Renamed addon to "Multiple rows in backpack" to better describe functionality

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "Multiple rows in backpack",
+  "description": "Display the contents of the editor backpack in an expandable two-way grid similar to the costume library and increase the size of thumbnails to improve ease of use.",
+  "credits": [
+    {
+      "name": "pumpkinhasapatch",
+      "link": "https://github.com/pumpkinhasapatch"
+    }
+  ],
+  "versionAdded": "1.29.0",
```
WorldLanguages, line 10: Let's change this to 1.30.0 and try to get this merged soon :D

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "backpack.css",
+      "matches": ["projects"]
+    }
+  ],
+  "updateUserstylesOnSettingsChange": true,
```
WorldLanguages, line 61: I don't think `updateUserstylesOnSettingsChange` is necessary for this addon. This property only affects userstyles that use the "if" property inside userstyle definitions. Custom CSS variables are always updated when settings change, and that can't be turned on nor off.

addons/expanded-backpack/backpack.css
```diff
@@ -0,0 +1,31 @@
+.backpack_backpack-list_MOp5X {
+    /* set height of the backpack picker, 5.5rem is the height of 1 row so multiply it by the rows setting */
+    min-height: calc(var(--expandedBackpack-rows) * 5.5rem);
+}
+
+.backpack_backpack-list-inner_10a2A {
+    /* move backpack items to the top left instead of vertically centering them */
+    -webkit-align-items: start;
+        -ms-align-items: start;
```
WorldLanguages, line 9: There's an unprefixed CSS property name supported in all modern browsers: `align-items`

addons/expanded-backpack/backpack.css
```diff
@@ -0,0 +1,31 @@
...
+    /* force horizontal wrapping and enable vertical scrolling */
+    flex-wrap: wrap;
+    overflow-x: hidden;
+    overflow-y: auto;
+}
+
+.backpack_backpack-item_hwqzQ {
+    /* add vertical margin to make stacked buttons look nicer */
+    margin: 0.25rem 0.25rem;
```
WorldLanguages, line 18: Nitpick, but you may replace this with simpler `margin: 0.25rem`.

addons/expanded-backpack/backpack.css
```diff
@@ -0,0 +1,31 @@
+.backpack_backpack-list_MOp5X {
```
WorldLanguages, line 1: Change this to `[class^="backpack_backpack-list"]`
Do this for every CSS selector

By the way, @mxmou When is it suitable to target full hashed class names? Addons such as editor-dark-mode intentionally do not, as far as I know. At some point we should write about this on the developer docs.

mxmou: The hashes don't work in TurboWarp Desktop so they should never be included in the selector. This isn't a problem for editor dark mode because it doesn't exist in TurboWarp.

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,64 @@
...
+    {
+      "name": "Number of displayed rows",
+      "id": "rows",
+      "type": "integer",
+      "min": 1,
+      "max": 4,
+      "default": 2
+    },
+    {
+      "name": "Improved costume and script previews",
```
mxmou, line 23: I prefer the previous name - it explained how it improves the previews.

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,64 @@
...
+    {
+      "name": "bgcolor",
+      "value": {
+        "type": "ternary",
+        "source": {
+          "type": "settingValue",
+          "settingId": "upscale"
+        },
+        "true": "white",
+        "false": "none"
```
mxmou, line 51: `none` is a valid value for `background` but not for `background-color`. I think this should be `transparent`.

mxmou (dismissed):
Enabling or disabling the addon or changing the `rows` setting while the backpack is open can make the editor look like this:
image

This can be fixed by adding a userscript that triggers a `resize` event on the `window` when the addon is enabled or disabled or a setting is changed.

addons/expanded-backpack/addon.json
pumpkinhasapatch: Thanks, leaving this open to discuss the buggy resizing issue above.

I've also noticed that the vertical gap between rows changes depending on the height of the backpack element.
This will definitely need to be fixed before merging, and is especially noticeable on large screens or when using rows=4.
Not sure what the exact cause of it is though and might need to change how resizing the height works.

https://user-images.githubusercontent.com/113677241/205540655-621fc40d-f1f4-4bcf-bc64-00138bbd0c56.mp4

mxmou: I didn't have the latest changes. The spacing issue is caused by this selector: `[class^="backpack_backpack-item"]`

It should be `[class*="backpack_backpack-item"]` because that class doesn't appear at the beginning of the attribute value.

WorldLanguages: > that class doesn't appear at the beginning of the attribute value

My bad, sorry!

WorldLanguages: Is this fixed?

pumpkinhasapatch: > This can be fixed by adding a userscript that triggers a `resize` event on the `window` when the addon is enabled or disabled or a setting is changed.

did some digging through the addon api docs and the hide-stage addon userscript (which seems to have a similar issue) and i think i've found a solution

mxmou:
This bug still exists:
> there's a lot of empty space if there aren't enough visible items to fill three or four rows. image
> 
> This could probably be fixed by adding the following properties to the `backpack_backpack-list-inner`:
> 
> ```css
> height: auto;
> max-height: 100%;
> ```

Everything else looks good now.

WorldLanguages:
@mxmou @pumpkinhasapatch Think there are any possible drawbacks to fixing the bug mentioned above by adding the listed properties to `backpack_backpack-list-inner` ?

WorldLanguages:
Other than that, looks good to merge.

pumpkinhasapatch:
Sorry for the slow reply, I've spent hours trying to find a solution to the bug and overthinking things way too much when the answer was right here all along.

> This could probably be fixed by adding the following properties to the `backpack_backpack-list-inner`:
> ```css
> height: auto;
> max-height: 100%;
> ```

I tried this when you originally sent it and it had no effect on the backpack. Maybe I forgot to reload the extension or some other code being tested was messing with it. Added it back again and it works like a charm :smile: 

All the previously mentioned bugs have been cleaned up and the addon seems much more stable and ready for release.

addons/expanded-backpack/addon.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "Multiple rows in backpack",
+  "description": "Display the contents of the editor backpack in an expandable two-way grid similar to the costume library and increase the size of thumbnails to improve ease of use.",
```
mxmou, line 3:
```suggestion
  "description": "Displays the contents of the editor backpack in an expandable two-way grid similar to the costume library and increases the size of thumbnails to improve ease of use.",
```

mxmou (commented):
I'm not sure if this PR should close #1095. It does improve the backpack, but the way it does that is very different from what the issue is suggesting.

BroJac5246:
> I'm not sure if this PR should close #1095. It does improve the backpack, but the way it does that is very different from what the issue is suggesting.

I don't think so, since it's suggesting something different that could still be applicable.

pumpkinhasapatch:
A lot of the existing backpack suggestions that move everything into the block palette or costume library would be too confusing to new users and clutter up the GUI. Instead, I want to _treat the backpack more like the library_, like giving it larger icons to see what you're about to add into your project instead of squinting at 256 pixels. These will all hide away in the backpack menu when you don't need them and don't interfere with other parts of Scratch.

There's a few more things involving JS in the works that I feel the backpack's been deserving for a long time, like filtering tabs to only show one type of items in the backpack at a time (sprites, costumes, scripts or sounds) for users who have a ton of stuff in their backpack and are trying to find just one asset inside it. Here's a mockup of what it might look like:

image

Samq64 (approved):

WorldLanguages:
@pumpkinhasapatch Thanks! By the way, changes to existing addons are usually merged faster, and often the creator of the addon implicitly has some extra decision power. So if you have any more ideas on enhancing the backpack, go ahead!

PR #5121, opened 2022-09-21
Add keybinds for pause button

rgantzos:
This is a requested feature from the Scratch Addons community server. It allows you to press ctrl+p to pause/resume the current project. I find this very helpful, and it's nice and easy for platformers and other fast-paced games.

### Changes

I added an event listener for when ctrl+p is pressed, and then it simulates the button click. It's really simple, but has a large impact.

### Reason for changes

Highly requested suggestion marked as a Good Idea in the community server.

### Tests

I have tested this, in the editor, on project pages, and in fullscreen. It worked every time, including when switching in between the pages.


jeffalo:
ctrl p is the default keybind to print a page. is it used rarely enough on scratch to justify overwriting the keybind?

CST1229:
> ctrl p is the default keybind to print a page. is it used rarely enough on scratch to justify overwriting the keybind?

Did anyone ever even print a Scratch page?

rgantzosonscratch:
> > ctrl p is the default keybind to print a page. is it used rarely enough on scratch to justify overwriting the keybind?
> 
> 
> 
> Did anyone ever print a Scratch page at all?

Even if they wanted to, all the elements get jumbled up when you try.

WorldLanguages:
I would suggest something more reachable. And no conflicts with browser hotkeys such as printing
Alt+X for example

rgantzosonscratch:
> I would suggest something more reachable. And no conflicts with browser hotkeys such as printing
> Alt+X for example

Good idea, you should be able to use it with one hand.

rgantzosonscratch:
I've changed the keybind to alt+x, and added an option to the settings to enable/disable the keybinds. This is also dynamic.

addons/pause/userscript.js
```diff
@@ -16,6 +16,16 @@ export default async function ({ addon, global, console, msg }) {
   onPauseChanged(setSrc);
 
   while (true) {
+    if (!document.querySelector(".pause-btn")) {
```
apple502j, line 19: This adds multiple listeners. Move this up.

addons/pause/userscript.js
```diff
@@ -16,6 +16,16 @@ export default async function ({ addon, global, console, msg }) {
...
+      document.addEventListener("keydown", function (e) {
+        if (addon.settings.get("keybindEnabled")) {
+          if (e.altKey && e.key === "≈") {
```
apple502j, line 22: What is this key?

addons/pause/userscript.js
```diff
@@ -16,6 +16,16 @@ export default async function ({ addon, global, console, msg }) {
...
+            e.preventDefault();
+            img.click();
```
apple502j, line 24:
```suggestion
            setPaused(!isPaused())
```

addons/pause/addon.json
```diff
@@ -25,5 +25,18 @@
     }
   ],
   "tags": ["editor", "projectPlayer", "recommended"],
-  "enabledByDefault": false
+  "enabledByDefault": false,
+  "settings": [
+    {
+      "name": "Alt+X keybind",
```
apple502j, line 31: Better terms?

apple502j (dismissed):

addons/pause/userscript.js
rgantzosonscratch: > What is this key?

It's the key for x when alt is pressed at the same time.

addons/pause/userscript.js
rgantzosonscratch: > 

Would that allow the pause button image to change as well?

addons/pause/addon.json
rgantzosonscratch: > Better terms?

Any ideas? How about "Alt+X to pause/resume"

addons/pause/userscript.js
mxmou: On which keyboard layout?

rgantzosonscratch: Macbook Air, just normal keyboard. At first when I tried checking if the key was "x", it wouldn't work. When I checked what the key was, it had "≈".

mxmou: That's probably specific to Mac keyboards. Try using `e.code` instead.

addons/pause/addon.json
mxmou: I don't think this needs to be optional.

rgantzosonscratch: Well, it prevents any default actions so... maybe it should just be enabled by default?

addons/pause/userscript.js
apple502j: It should, I believe.

WorldLanguages:
Note: this keybind is being used as an example to discuss #5145, a better way to let users customize keybinds.

addons/pause/userscript.js
Samq64: The listener should go after the button.

sfederici:
This issue seems old enough (sept/oct 2022). When will the addon be added to ScratchAddons? I find it extremely useful (and important for my teaching).

WorldLanguages:
Consensus appears to be in favor of making this keybind not-togglable, at least for now.

WorldLanguages:
@Samq64 If we remove the setting and add a note to the addon, we're all set, right?

Samq64:
> @Samq64 If we remove the setting and add a note to the addon, we're all set, right?

Yes. I forgot I self-assigned this.

Samq64:
@rgantzos I removed the keybind toggle. That's fine, right?

addons/pause/addon.json
```diff
@@ -1,6 +1,12 @@
 {
   "name": "Pause button",
   "description": "Adds a pause button next to the green flag.",
+  "info": [
+    {
+      "text": "The project can also be paused with Alt+X.",
```
WorldLanguages, line 6: Had to Google whether Macs have a key named "Alt", apparently not.

```suggestion
      "text": "The project can also be paused with Alt+X (Option+X on macOS).",
```

addons/pause/userscript.js
WorldLanguages: Is there a reason why this needs to be inside the while loop...?

WorldLanguages: > After it adds the listener, it immediately adds the button. That means that it won't ever add a listener again.

What about going inside then back to the project page?

Samq64 (approved):

addons/pause/userscript.js
```diff
@@ -16,6 +16,14 @@ export default async function ({ addon, console, msg }) {
   onPauseChanged(setSrc);
 
   while (true) {
+    if (!document.querySelector(".pause-btn")) {
```
WorldLanguages, line 19: Can we move this outside the while loop? I don't see why it needs to be here. This may create multiple event listeners in the same tab session.

WorldLanguages (approved):

PR #5133, opened 2022-09-26
confirmation

aaa464:
it keeps you safe from hackers
Resolves #

### Changes
it makes sure you are you when doing stuff in scratch, such as sharing projects

### Reason for changes
scratch hacks

to keep you safe from hackers
### Tests
 scratch account


Platformer444:
What is this PR about?
@aaa464 Pls elaborate

PR #5136, opened 2022-09-27
scratch messaging: Ctrl + Enter to post message

TheColaber:
Suggested by micahlt on discord
https://discord.com/channels/806602307750985799/1024110625070272613

> At the moment, you can use Ctrl+Enter to submit comments on users, studios, and projects without clicking the Post button.  I'd love to see the same feature in SA's Scratch Messaging addon that appears in the extension popup.

### DIFF
--- a/popups/scratch-messaging/popup.html
+++ b/popups/scratch-messaging/popup.html
@@ -68,7 +68,12 @@
           </div>
         </bdo>
         <div class="reply-box-comment" v-show="replying">
-          <textarea class="reply-textarea" maxlength="500" v-model="replyBoxValue"></textarea>
+          <textarea
+            class="reply-textarea"
+            maxlength="500"
+            v-model="replyBoxValue"
+            @keyup.enter="postComment($event.shiftKey)"
+          ></textarea>
           <div class="reply-box-buttons">
             <button @click="postComment()" class="large-button post-button" :disabled="postingComment">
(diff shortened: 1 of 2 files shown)


popups/scratch-messaging/popup.html
```diff
@@ -68,7 +68,12 @@
           </div>
         </bdo>
         <div class="reply-box-comment" v-show="replying">
-          <textarea class="reply-textarea" maxlength="500" v-model="replyBoxValue"></textarea>
+          <textarea
+            class="reply-textarea"
+            maxlength="500"
+            v-model="replyBoxValue"
+            @keyup.enter="postComment($event.shiftKey)"
```
apple502j, line 75: Does look ugly a bit IMO, also I'm pretty sure you broke the button

apple502j (changes requested):
Also why Shift instead of Ctrl?

popups/scratch-messaging/popup.html
TheColaber: > Does look ugly a bit IMO

Blame vue1 not supporting `.ctrl` modifier

apple502j: Can't you `&&` in the HTML?

TheColaber: that would look uglier but sure

Weredime (approved):

apple502j (approved):

PR #5142, opened 2022-09-28
Isonline :eyes:

GrahamSH-LLK:
Resolves #


jeffalo:
FYI the original isOnline was stylized as isOnline, not IsOnline

WorldLanguages:
Public note: this is a proof of concept made by GrahamSH, and this is in no way planned to be added to the addon library.

Using Scratch servers with the intention of indicating online status is rejected "by default". Deep discussion is needed if we ever, somehow, consider it to be a good idea (won't happen to be honest)

Norbiros:
I was talking with WL and he told me that we need to change add-on policy to do that, or we can create totally new extension with is online.

PR #5146, opened 2022-09-29
Add dark comment migration

apple502j:
Adds migration code for #3862.

### Changes
- Add migration code for 3862. If dark comment is turned off, the vanilla color is always used. If dark comment is turned on, the default for the preset is used, if applicable. Use the attached JSONs for testing: [testcases.zip](https://github.com/ScratchAddons/ScratchAddons/files/9675771/testcases.zip)
- Fix bug with importing old, non-versioned settings files when the current version uses versioned settings - such as loading the testcase in this branch in current version. They previously skipped the preset transition code. (This does not affect things currently, because the only broken code, v1.23 editor-dark-mode transition code, was removed in 1.28.)

### Reason for changes
The now-merged pull request moved a setting between addons, which requires transition code.

### Tests
Tested on Firefox 105. Use the attached testcase to test easily.

### DIFF
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -5,6 +5,7 @@ import minifySettings from "../libraries/common/minify-settings.js";
  the versions separately. Current versions:
 
  - editor-dark-mode 2 (bumped in v1.23 twice)
+ - editor-theme3 4 (last bumped in v1.29)
  */
 
 const updatePresetIfMatching = (settings, version, oldPreset = null, preset = null) => {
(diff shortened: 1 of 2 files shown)


WorldLanguages (approved):

background/get-addon-settings.js
```diff
@@ -138,6 +144,58 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
...
+              "input-color": "#202020",
+              text: "white",
+            },
+            manifest.presets.find((p) => p.id === "dark")
+          );
+
+          if (addonSettings["editor-dark-mode"]?.darkComments === false) {
+            // Transition v1.28 to v1.29
+            // Override the preset color if dark comments are not enabled
+            addonSettings["comment-color"] = "#FEF49C";
```
DNin01, line 196: Pretty sure this should have been...
```suggestion
            addonSettings["editor-theme3"]["comment-color"] = "#FEF49C";
```
But it's too late to fix now.

<sub>Not unless anyone still has a settings export from before v1.29 saved, lol.</sub>

WorldLanguages: Whoops.

PR #5153, opened 2022-10-01
Look at random direction

emir4169:
n/a
### Changes

Adds a "random position" option to the "point towards ()" block that makes the sprite point in a random direction. This will work for everyone using the project.

### Reason for changes

To lower the amount of blocks used for projects that use a extra block in the "point towards ()" block to make it point to a random direction

### Tests

tests i did
1. Does it show up
2. Does it work
both were true
albeit the first one took way too much commits to my own fork


addons/addons.json
```diff
@@ -149,5 +149,6 @@
   "editor-stage-left",
   "editor-buttons-reverse-order",
   "columns",
-  "editor-theme3"
+  "editor-theme3",
+  "lookatrandom"
```
BroJac5246, line 153: This needs to be above the line that says "NEW ADDONS ABOVE THIS," since this isn't a theme.

addons/lookatrandom/addon.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "Point towards random direction",
```
BroJac5246, line 2: This needs a better name.

addons/lookatrandom/addon.json
```diff
@@ -0,0 +1,13 @@
...
+  "description": "Adds a \"random position\" option to the \"point towards ()\" block that makes the sprite point in a random direction. This will work for everyone using the project.",
+  "userscripts": [
+    {
+      "url": "code.js",
+      "matches": ["https://scratch.mit.edu/*"]
+    }
+  ],
+  "versionAdded": "1.29.0",
+  "tags": ["community"],
+  "enabledByDefault": false
```
pufferfish101007, line 12:
```suggestion
  + "dynamicEnable": true,
  + "dynamicDisable: true,
  + "credits": [ // maybe add some credits - you and also me seeing as I wrote the code :) ]
```

addons/lookatrandom/addon.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "Point towards random direction",
+  "description": "Adds a \"random position\" option to the \"point towards ()\" block that makes the sprite point in a random direction. This will work for everyone using the project.",
+  "userscripts": [
+    {
+      "url": "code.js",
+      "matches": ["https://scratch.mit.edu/*"]
+    }
+  ],
+  "versionAdded": "1.29.0",
```
pufferfish101007, line 10: This should not be "community"

addons/lookatrandom/code.js
```diff
@@ -0,0 +1,22 @@
+export default async function ({ addon, global, console }) {
+  const ScratchBlocks = await addon.tab.traps.getBlockly();
+  ScratchBlocks.Blocks.motion_pointtowards_menu.init = function () {
```
pufferfish101007, line 3:
```suggestion
  const oldInit = ScratchBlocks.Blocks.motion_pointtowards_menu.init
  ScratchBlocks.Blocks.motion_pointtowards_menu.init = function () {
    if (!addon.self.disabled) {
      // ...
    } else {
      oldInit.call(this);
    }
 }
```

pufferfish101007 (changes requested):

addons/lookatrandom/addon.json
```diff
@@ -0,0 +1,13 @@
...
+  "tags": ["community"],
```
BroJac5246, line 11:
```suggestion
  "tags": ["editor"],
```

addons/lookatrandom/addon.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "Point towards random direction",
+  "description": "Adds a \"random position\" option to the \"point towards ()\" block that makes the sprite point in a random direction. This will work for everyone using the project.",
+  "userscripts": [
+    {
+      "url": "code.js",
+      "matches": ["https://scratch.mit.edu/*"]
```
BroJac5246, line 7:
```suggestion
      "matches": ["projects"]
```
I _think_ I got that right.

BroJac5246 (commented):
This is my first review, so let be know if I did something wrong :)

emir4169:
please remember to fact read your code before asking me to commit that
now my codes broken and i have to manually find it and fix it

BroJac5246:
> please remember to fact read your code before asking me to commit that now my codes broken and i have to manually find it and fix it

`console.log` is working, so it's something in `code.js`. I don't know what, though.

addons/lookatrandom/addon.json
```diff
@@ -0,0 +1,16 @@
...
+      "url": "code.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.29.0",
+  "tags": ["editor"],
+  "enabledByDefault": false,
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "credits": ["pufferfish101007#2024: wrote the code", "me (winwinlmfao#8430): made the addon"]
```
BroJac5246, line 15:
```suggestion
  "credits": [
    {
      "name": "pufferfish101007",
      "link": "https://scratch.mit.edu/users/pufferfish101007/"
    },
    {
      "name": "winwinlmfao"
    }
  ]
```
This way the credits will actually display.

And yes, I tested it. It should work.

BroJac5246: Remember to mark this as resolved :)

BroJac5246:
The addon prevents you from selecting a sprite in the dropdown. That's quite problematic, so it needs to be fixed.
The issue occurs with both `code.js` and `oldcode.js`.

pumpkinhasapatch:
Scratch contains unused code needed to show the _random position_ option in the point towards block [as seen here](https://github.com/LLK/scratch-blocks/blob/4e10e1f3c45e75c434009a280cd1691b0fe6eac2/blocks_vertical/motion.js#L137), but for some reason it doesn't show to the end-user when changing the dropdown in the editor.

There must be another function somewhere that removes _random position_ from the dropdown and adds the sprite list into it instead. This addon makes it load the block definition again, resetting the block input with the _mouse position_ and _random position_ options but not adding the sprites back.

If we can find this function and have the addon use it to load the sprites but not remove the _random position_ it would likely fix what @BroJac5246 mentioned above while still accomplishing what the addon is meant to do.

pufferfish101007:
The function is overwritten at https://github.com/LLK/scratch-gui/blob/ba76db7350bd43b79119cac2701bc10f6c511f0c/src/lib/blocks.js#L178. https://github.com/LLK/scratch-gui/blob/ba76db7350bd43b79119cac2701bc10f6c511f0c/src/lib/blocks.js#L10-29 makes the json, and you can see that it concatenates the given mous option with https://github.com/LLK/scratch-gui/blob/ba76db7350bd43b79119cac2701bc10f6c511f0c/src/lib/blocks.js#L116-130
See if you can give that a go yourself - copy the `spriteMenu` function and then call & add it to the list of menu options.

emir4169:
> The function is overwritten at https://github.com/LLK/scratch-gui/blob/ba76db7350bd43b79119cac2701bc10f6c511f0c/src/lib/blocks.js#L178. https://github.com/LLK/scratch-gui/blob/ba76db7350bd43b79119cac2701bc10f6c511f0c/src/lib/blocks.js#L10-29 makes the json, and you can see that it concatenates the given mous option with https://github.com/LLK/scratch-gui/blob/ba76db7350bd43b79119cac2701bc10f6c511f0c/src/lib/blocks.js#L116-130 See if you can give that a go yourself - copy the `spriteMenu` function and then call & add it to the list of menu options.

yea uh finding the end of sprite menu is the hard part of all that

emir4169:
THERES A BIG COMPATIBLITY ISSUE WITH ONE ADDON THAT MAKES THE RANDOM DIRECTION THING NOT FUNCTION
IT SHOWS BUT DOESNT DO ANYTHIN

RedGuy12:
bro chill it's not even close to merging
this is why we have pull requests and reviews and testing

BroJac5246:
> THERES A BIG COMPATIBLITY ISSUE WITH ONE ADDON THAT MAKES THE RANDOM DIRECTION THING NOT FUNCTION
> IT SHOWS BUT DOESNT DO ANYTHIN

What is that addon?

emir4169:
> > THERES A BIG COMPATIBLITY ISSUE WITH ONE ADDON THAT MAKES THE RANDOM DIRECTION THING NOT FUNCTION
> > IT SHOWS BUT DOESNT DO ANYTHIN
> 
> Which addon?

I DONT KNOW BUT WE HAVE A SUSPECT ON WHICH IT IS
AND THE SUSPECT IS THE EXTRA KEYS

emir4169:
> > > > THERES A BIG COMPATIBLITY ISSUE WITH ONE ADDON THAT MAKES THE RANDOM DIRECTION THING NOT FUNCTION
> > > > IT SHOWS BUT DOESNT DO ANYTHIN
> > > 
> > > 
> > > Which addon?
> > 
> > 
> > I DONT KNOW BUT WE HAVE A SUSPECT ON WHICH IT IS
> > AND THE SUSPECT IS THE EXTRA KEYS
> 
> Okay, first of all, please turn off caps lock, and second, does the error still occur with all other addons off?
> 
> There should be no 'suspect,' since all you need to do is turn off the addon that you think is causing the problem. If the problem is still there, then you know what's causing it, if it's not then you know that that certain addon is not the problem.

oh crap the problem is still occouring cause of a freaking undefined function (which i litterally defined in a build of my fork not yet put in github. same error.) at this point im going crazy trying to define this ONE function thats really important.

emir4169:
im putting my local version (which is newer than the current github one but is still broken)

addons/lookatrandom/addon.json
CubesterYT: We could say "Randomize Direction Option", or something similar.

BroJac5246: Maybe "Random direction option in point towards block"?

Scratch Addons is somewhat picky about addon names, so I'm really not sure.

Cool-Scratcher: I'm just randomly hopping into this discussion, but I think the original name is the best (or maybe add "block" to the end of it)

Tacodude7729:
Maybe have a look at the code for the extra key support addon [here](https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/editor-extra-keys/userscript.js). You really shouldn't be completely overwriting the existing menu, rather appending to it.

WorldLanguages:
Superseded by #5479 because PR author went inactive.
I'm also to blame here, because I was requested review and said anything! Sorry, I was very busy during October.

WorldLanguages:
@emir4169 Can we credit you with your github username `emir4169`"? Or do you have strong reasons to prefer to be credited as `winwinlmfao`?

For context, see https://github.com/ScratchAddons/ScratchAddons/pull/5479#discussion_r1316143517

emir4169:
> @emir4169 Can we credit you with your github username `emir4169`? Or do you have strong reasons to prefer to be credited as `winwinlmfao`?
> 
> For context, see [#5479 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/5479#discussion_r1316143517)

id like to be credited as emir4169

CubesterYT:
HE RESPONDED YOOOOO

PR #5158, opened 2022-10-03
New addon: collapsing sprite properties

### LINKED ISSUE #4970: Button to hide stage and expand sprite area

The sprite area is rather small in Scratch, and switching to different sprites can be annoying especially when a project has many sprites in it.

A button could be added next to the stage size buttons (just like the "Hide stage and sprite pane" addon) that hides the stage, and stretches the sprite pane up to fill the space that the stage normally occupies.

lisa-wolfgang:
Resolves #4970 (more or less)

### Changes

Users on small screens rejoice: you're about to get double the sprite grid space!

A screenshot of the sprites area with the addon enabled. The sprite properties panel has been hidden, leaving more room for the sprites grid. A blue button displaying the letter "i" is present in the upper-left corner of the selected sprite tile.

When this addon is enabled, the sprite properties panel is hidden by default. The panel can be shown again by clicking a 2.0-like info button on the currently selected sprite or double-clicking a sprite tile. The panel can then be hidden by clicking a new collapse button on the properties panel or double-clicking a sprite tile again.

A screenshot of the sprites area with the addon enabled. The sprite properties panel is now showing. The blue button displaying the letter "i" from earlier is no longer present. A button with an arrow pointing up has been added to the bottom-right corner of the sprite properties panel.

When the "don't collapse panel on load" option is enabled, when the page is loaded or when the addon is enabled, the sprite properties panel will stay showing until it is closed manually.

When the "automatically collapse when mouse leaves sprite panel" option is enabled, the sprite properties panel will automatically hide once the mouse leaves the area encompassing both the properties panel and the sprites grid.

https://user-images.githubusercontent.com/43426138/196075309-a9e6d612-4877-4a87-a076-4a837ba648a7.mp4

### Reason for changes

Users with small screens currently have difficulty accessing the sprites grid, especially when there are many sprites per project. This addon attempts to alleviate that by increasing the available sprites grid space by over a factor of 2.

### Tests

Tested to work on Chromium 106 and Firefox 105.0.3.


Samq64:
What about people on tablets or chromebooks with bad touchpads? They probably don't have problems double-clicking, but there's still probably a better way.

Samq64:
What about people on tablets or chromebooks with bad touchpads? They probably don't have problems double-clicking, but there's still probably a better way.

Samq64:
What about people on tablets or chromebooks with bad touchpads? They probably don't have problems double-clicking, but there's still probably a better way.

BroJac5246:
> What about people on tablets or chromebooks? They might not have problems double-clicking, but there's probably still a better way.

Good point. This could get awkward for tablets or Chromebooks in tablet mode. I wonder if some sort of "show sprite properties panel" button could be beneficial. It would expand the sprite properties panel when clicked or hovered over.

lisa-wolfgang:
I think it would be best to avoid adding such a button, since it would have the considerable tradeoff of subtracting space from the sprites grid area, which would reduce the efficacy of this addon. I did consider adding a "properties" button to each sprite's context menu, but double-tapping is much easier to perform for tablet users than long-holding (the current way to access context menus on touchscreens) and then tapping the option. If the concern is about accidentally hitting the delete button, touchscreen users will have likely already enabled the deletion confirmation and/or delete button hiding addons for when they need to select sprites normally.

lisa-wolfgang:
I'm actually not sure to what extent Scratch Addons supports (or should support) touchscreen devices. As mentioned earlier, touchscreen Chromebooks certainly do exist, and we support Scratch Addons on Chrome. I'm going to create an issue for this, but I don't think that it's a merge-blocker.

lisa-wolfgang:
On a somewhat related note, because of this addon's usefulness to users with small screen sizes and its classification as a 2.0-like feature, I would like to propose that this addon receive the `featured` tag.

lisa-wolfgang:
Thinking about it some more, it might be best to include an option that keeps the name bar showing and have this enabled by default. There is otherwise no UI-indicated way to edit a sprite's name.

BroJac5246:
> I think it would be best to avoid adding such a button, since it would have the considerable tradeoff of subtracting space from the sprites grid area, which would reduce the efficacy of this addon.

I was thinking it could be like the backpack's button: short and wide. You could choose if the button shows through the addon settings, and it would help compensate for the disappearance of the sprite properties panel while causing minimal additional disturbance.

Secret-chest:
> Thinking about it some more, it might be best to include an option that keeps the name bar showing and have this enabled by default. There is otherwise no UI-indicated way to edit a sprite's name, unless an info button is added like in Scratch 2.0.

Why not allow editing it from the name displayed below the thumbnail? When clicked it would turn into an input.

BroJac5246:
> > Thinking about it some more, it might be best to include an option that keeps the name bar showing and have this enabled by default. There is otherwise no UI-indicated way to edit a sprite's name, unless an info button is added like in Scratch 2.0.
> 
> Why not allow editing it from the name displayed below the thumbnail? When clicked it would turn into an input.

Probably double clicked, but good idea. 👍

lisa-wolfgang:
> Why not allow editing it from the name displayed below the thumbnail? When clicked it would turn into an input.

That's a possible option, but keep  in mind that there is very little horizontal space inside a sprite tile.

Secret-chest:
> > Why not allow editing it from the name displayed below the thumbnail? When clicked it would turn into an input.
> 
> That's a possible option, but keep  in mind that there is very little horizontal space inside a sprite tile.

Well, the input could be minimal, like the Windows Explorer double-click to rename.

joshpinto6:
This is a great feature!

Secret-chest:
Also, can you make an animation?

lisa-wolfgang:
Given that the input doesn't necessarily need to be confined to the dimensions of the sprite tile, I do like the idea of double-clicking a sprite's name to quickly edit it. However, this feature would likely fit better in a new addon. Most users wouldn't expect to find this specific feature inside of an addon whose primary feature is to collapse the properties panel.

lisa-wolfgang:
I'd still like to add an "info" button on the currently selected sprite that shows the properties panel, as well as a "close" button on the properties panel itself, to make this addon more user-friendly. These added UI elements will not be configurable.

BroJac5246:
> However, this feature would likely fit better in a new addon. Most users wouldn't expect to find this specific feature inside of an addon whose primary feature is to collapse the properties panel.

I agree. Perhaps that addon could also have an option to remove the sprite name input- that way, if someone doesn't quite want this addon, they could at least make the sprite properties panel smaller.

GarboMuffin (changes requested):
The animation does not work in Firefox

And in all browsers, if I have something like this:

image

double clicking in the area under Sprite1 where there isn't any sprite does not work to show the menu

lisa-wolfgang:
@GarboMuffin I don't think I've touched anything relevant since you reviewed, but the animation works fine in Firefox from my testing on the latest commit -- could you reconfirm that it doesn't work for you?

lisa-wolfgang:
Besides fixes to any additional bugs discovered, I no longer have any changes I wish to make in this PR.

GarboMuffin:
Animation is working fine now, isn't that funny

Addon currently breaks when you leave the editor and go back inside

WorldLanguages:
@GarboMuffin Do you think this PR could be mergeable as-is? (regarding functionality and bugs)

GarboMuffin (approved):

addons/sprite-properties/addon.json
```diff
@@ -0,0 +1,83 @@
...
+          "short": "0.15s",
+          "default": "0.25s",
+          "long": "0.5s"
+        }
+      }
+    }
+  ],
+  "settings": [
+    {
+      "name": "Collapse panel on load",
```
WorldLanguages, line 43: I think we could use a better term than "on load" here

WorldLanguages: onion-skinning uses `Enable by default` for pretty much the same setting meaning.
We could use that or maybe `Collapse panel by default`

WorldLanguages:
> On a somewhat related note, because of this addon's usefulness to users with small screen sizes and its classification as a 2.0-like feature, I would like to propose that this addon receive the `featured` tag.

I agree!

addons/sprite-properties/addon.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "Collapsing sprite properties",
+  "description": "Hides the sprite properties panel by default, like in Scratch 2.0. Use the info button on the currently selected sprite or double-click a sprite to show the properties panel again. To re-hide it, use the collapse button in the properties panel or double-click a sprite.",
+  "tags": ["editor"],
```
WorldLanguages, line 4: TODO: add featured tag

addons/sprite-properties/userscript.js
```diff
@@ -0,0 +1,106 @@
+export default async function ({ addon, global, console, msg }) {
+  const showPropsClass = "sa-show-sprite-properties";
+  const hidePropsClass = "sa-hide-sprite-properties";
+  const propsBtnClass = "sa-sprite-properties-btn";
+  const propsCloseBtnClass = "sa-sprite-properties-close-btn";
```
WorldLanguages, line 5: We could use `SNAKE_CASE` for these constants

addons/sprite-properties/addon.json
```diff
@@ -0,0 +1,83 @@
...
+          "id": "long",
+          "name": "Long"
+        }
+      ],
+      "default": "default"
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.29.0",
```
WorldLanguages, line 81: 1.30.0

WorldLanguages (dismissed):
There's a bug (setting "COLLAPSE PANEL BY DEFAULT" enabled):
Go to scratch.mit.edu/create and wait for the project to load. The "i" icon will appear when the panel is expanded, instead of when it's collapsed, which is the opposite of what normally happens.

lisa-wolfgang:
Could not repro in browser with Chromium 107.

mxmou:
The arrow for closing it isn't very noticeable. Could it be centered below the bottom row of properties?

WorldLanguages:
Merging: PR author is an org member + 1 approval

WorldLanguages:
> The arrow for closing it isn't very noticeable. Could it be centered below the bottom row of properties?

You may create an issue or a PR.

WorldLanguages:
Low prio bug I dismissed was moved to issue #5443

WorldLanguages:
> The arrow for closing it isn't very noticeable. Could it be centered below the bottom row of properties?

@mxmou Reminder!

PR #5163, opened 2022-10-06
Show More button on Scratch News

themirrazz:
n/a 

### Changes
This addon allows you to view all the Scratch News entries without having to open the discussion forum.
The "View All" button is replaced with a "Show More" button. When expanding the news section, a "View in Forums" button will open the discussion the "View All" button opened. The best part: it opens it in a new tab, instead of redirecting the current tab.
There's also a really satisfying sliding animation, that if you feel like it, you can disable.

### Reason for changes
Scratch despite having an API for all news, only showed the first 3. As far as I know, there's no place on Scratch that allows you to view all of the news entry. And the "View All" button would redirect the current tab, instead of opening it up in a new one. Plus, Scratch kind of lacks cool-looking animations :)

### Tests
This addon has been tested on Brave 1.42, which is based off of Chromium 104.
It should work on the latest versions of Chrome, Edge, and Opera.


BroJac5246:
Two things:
1. The addon is crashing due to an error in the manifest.
2. Should this be part of the Infinite scrolling addon? I haven't tried this out yet (because of the manifest error), but it seems like this might be a good addition to that addon.

themirrazz:
> Two things:
> 
> 1. The addon is crashing due to an error in the manifest.
> 2. Should this be part of the Infinite scrolling addon? I haven't tried this out yet (because of the manifest error), but it seems like this might be a good addition to that addon.

No, this should not be part of the infinite scrolling addon, as it does not allow infinite scrolling.
And unlike the load more addon, it allows you to Show More and Show Less.

I'll check the manifest right now.

themirrazz:
oh I'm so stupid

themirrazz:
I named it manifest.json not addon.json

themirrazz:
Fixed it

BroJac5246:
> > Two things:
> > 
> > 1. The addon is crashing due to an error in the manifest.
> > 2. Should this be part of the Infinite scrolling addon? I haven't tried this out yet (because of the manifest error), but it seems like this might be a good addition to that addon.
> 
> No, this should not be part of the infinite scrolling addon, as it does not allow infinite scrolling. And unlike the load more addon, it allows you to Show More and Show Less.

I meant Full areas, sorry 🤦. I think it might go well with that addon since it has similar functionality.

Also, it loads a lot of news at once, which is overwhelming. Could you have some sort of button similar to what Full areas uses? It might make sense to let both a show more and show less button display at the same time.

themirrazz:
Oh, okay

themirrazz:
Gonna update when I get home

CST1229:
Maybe it could add a scrollbar instead of expanding the container a lot.

TheColaber:
needs to be rewritten

PR #5201, opened 2022-10-15
[GitHub meta] Change CODEOWNERS entries

WorldLanguages:
Since we allow any org member to merge PRs after 2 approvals, the current CODEOWNERS file makes little sense.

@Hans5958 Do you want to keep your entry, what do you think?

Does any org member want CODEOWNER of a specific folder/addon? It's a good moment to add it
Personally, I'd like to self-request review when appropiate, instead of getting it automatically in every single PR.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,7 @@
 /.github/ @Hans5958
+/addons/addons.json @WorldLanguages  # New addons
+/addon-api @WorldLanguages @apple502j
+/background @WorldLanguages @apple502j
+/content-scripts @WorldLanguages @apple502j
+/libraries @WorldLanguages @apple502j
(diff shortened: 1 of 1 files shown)


Hans5958:
I think at this point the CODEOWNERS file should be made just in case anyone wanted to be mentioned every time their addon (or certain parts) has something to be changed. I would also like to add a little note so anyone can also add then to it

PS.

> Personally, I'd like to self-request review when appropiate, instead of getting it automatically in every single PR.

At that point you shouldn't need to self-request

Hans5958:
Ah, didn't know that you would utilize it like that.

WorldLanguages:
Hans: your codeowners rule has stayed unchanged
@apple502j Please approve/reject the new rules

apple502j (approved):

PR #5202, opened 2022-10-15
New setting: "display addon names in English"

### LINKED ISSUE #5186: New SA setting: "display addon names in English"

Until we can solve #2670 directly, we should allow users to see addon names, description, setting names, etc. in English, even if their browser language is set to a different language that we support.
Tags such as "Recommended" and other strings would remain in that language, but many users would still appreciate the feature.

### What this change does

A new setting available on "More Settings", which could be named "display addon names in English".
We could add a small "beta" tag next to it, to indicate that it's meant to be temporal. In the future, we should let the user set the whole SA UI to any of the supported languages, and it would affect everything except the extension description.

The value of this setting would be stored in `chrome.storage.local`, it would not be exportable through the JSON file, and it would set `useDefault` to `true` on this line: https://github.com/ScratchAddons/ScratchAddons/blob/master/background/load-addon-manifests.js#L42

Changing this setting would require reloading the extension, with exactly the same logic as importing a JSON settings file.

### Possible alternatives

Waiting until we can provide an actual language selector, and allow users to choose other languages, not just English.
But I've seen too many suggestions about this. I even saw users asking how to change the language on Reddit (r/scratch).

_No response_

WorldLanguages:
Resolves #5186

brave_oudFb3KiC2

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -308,6 +308,12 @@
   "confirmImport": {
     "message": "Confirm"
   },
+  "language": {
+    "message": "Language"
+  },
+  "applySettings": {
+    "message": "Apply settings"
+  },
   "quickSettings": {
     "message": "Addons"
   },
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -312,7 +312,7 @@
     "message": "Language"
   },
   "applySettings": {
-    "message": "Apply settings"
+    "message": "Apply changes"
   },
   "quickSettings": {
     "message": "Addons"
--- a/webpages/settings/index.html
+++ b/webpages/settings/index.html
@@ -264,18 +264,29 @@
         </div>
         <div class="addon-body">
           <div class="addon-topbar">
-            <span class="addon-name"><img src="../../images/icons/translate.svg" class="icon-type" />{{ msg("language") }}
+            <span class="addon-name"
+              ><img src="../../images/icons/translate.svg" class="icon-type" />{{ msg("language") }}
             </span>
           </div>
           <div class="addon-settings">
-            <div class="addon-setting" style="margin-top:0">
-              <input type="checkbox" class="setting-input check" v-model="forceEnglishSetting" 
-                style="margin-left: 0; margin-right: 8px">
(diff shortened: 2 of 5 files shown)


_locales/en/messages.json
```diff
@@ -308,6 +308,12 @@
   "confirmImport": {
     "message": "Confirm"
   },
+  "language": {
+    "message": "Language"
+  },
+  "applySettings": {
+    "message": "Apply settings"
```
lisa-wolfgang, line 315:
```suggestion
    "message": "Apply changes"
```

apple502j (approved):

webpages/settings/index.html
```diff
@@ -262,6 +262,34 @@ <h1 v-cloak>{{ msg("settings") }}</h1>
...
+          <div class="addon-settings">
+            <div class="addon-setting" style="margin-top: 0">
+              <input
+                type="checkbox"
+                class="setting-input check"
+                v-model="forceEnglishSetting"
+                style="margin-left: 0; margin-right: 8px"
+              />
+              <span>Show addon names and descriptions in English</span>
+              <div class="badge red">{{ "Beta" }}</div>
```
WorldLanguages, line 280: TODO: localize "beta" (string already exists)

WorldLanguages:
Note: the name for the setting is intentionally not localized, at the moment

mxmou:
The switch should probably be a checkbox because it's in front of the label.

WorldLanguages:
> The switch should probably be a checkbox because it's in front of the label.

Could you submit a PR? I'm pretty sure my design is not ideal, but I think it's worth it, because some users really wanted this feature.

lisa-wolfgang:
Do we have checkboxes anywhere else in the extension anymore? I'm not sure why this specific scenario warrants an inconsistent design component.

Samq64:
Or the switch could be moved to the right to be consistent with everything else.

WorldLanguages:
Note: enabling this new setting has the side-effect of converting these strings to English too (popup names):

image

I wouldn't say this was intentional, but it's not a bug either.

Issue #5206, opened 2022-10-16
rename-broadcasts considerations before v1.29 release


WorldLanguages:
- [x] GarboMuffin review
- [x] Attempting to rename a broadcast to the name of a broadcast that already exists silently fails (see comment https://github.com/ScratchAddons/ScratchAddons/pull/5178#issuecomment-1279840943)

PR #5215, opened 2022-10-16
Fix a lot of issues in rename-broadcast

GarboMuffin:
- The addon actually works now
 - Undo/redo works
 - When a broadcast with the new name already exists, the old broadcast is "merged" into the new one instead of silently doing nothing
 - ~~I changed some strings to use the same words that Scratch itself uses. If it's too late for string updates then just revert 8424f49696072bcaadbcf4bf55d5d8a059f41a76 before merging.~~ ~~Still think this change is correct but it seems to be too late for string updates~~ moved to https://github.com/ScratchAddons/ScratchAddons/pull/5231
 - editor-searchable-dropdowns now treats the rename message option the same as it does the rename variable option (hide during search)

I've been able to open up Appel and rename broadcasts while the project is running without issues, and those new names are saved when the project is saved and loaded again

This is a draft because it needs more actual testing.

### DIFF
--- a/addons-l10n/en/rename-broadcasts.json
+++ b/addons-l10n/en/rename-broadcasts.json
@@ -1,5 +1,5 @@
 {
-  "rename-broadcasts/RENAME_BROADCAST": "Rename message",
-  "rename-broadcasts/RENAME_BROADCAST_TITLE": "Rename all \"{name}\" messages to:",
-  "rename-broadcasts/RENAME_BROADCAST_MODAL_TITLE": "Rename Message"
+  "rename-broadcasts/RENAME_BROADCAST": "Rename broadcast",
+  "rename-broadcasts/RENAME_BROADCAST_TITLE": "Rename all \"{name}\" broadcasts to:",
+  "rename-broadcasts/RENAME_BROADCAST_MODAL_TITLE": "Rename Broadcast"
 }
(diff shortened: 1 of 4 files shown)


addons-l10n/en/rename-broadcasts.json
```diff
@@ -1,5 +1,5 @@
 {
-  "rename-broadcasts/RENAME_BROADCAST": "Rename broadcast",
-  "rename-broadcasts/RENAME_BROADCAST_TITLE": "Rename all \"{name}\" broadcasts to:",
-  "rename-broadcasts/RENAME_BROADCAST_MODAL_TITLE": "Rename Broadcast"
+  "rename-broadcasts/RENAME_BROADCAST": "Rename message",
+  "rename-broadcasts/RENAME_BROADCAST_TITLE": "Rename all \"{name}\" messages to:",
+  "rename-broadcasts/RENAME_BROADCAST_MODAL_TITLE": "Rename Message"
```
apple502j, line 4: Please revert these.

GarboMuffin: Removed for this PR, will worry about them again later.

RedGuy12:
fixes #5206

WorldLanguages:
> Still think this change is correct but it seems to be too late for string updates

I think we can change the strings from `addon-l10n` anyway.

apple502j (approved):

WorldLanguages:
> When a broadcast with the new name already exists, the old broadcast is "merged" into the new one instead of silently doing nothing

Could we emit a workspace update after doing this? The old broadcast message, which was merged into the new one, still appears in the menu, until I switch sprites. Or is this intentional because it's not that easy to update the context menu?

WorldLanguages (approved):
LGTM

PR #5217, opened 2022-10-17
Project player theme

### LINKED ISSUE #5210: Standalone project player theme
From https://github.com/ScratchAddons/ScratchAddons/discussions/4863#discussioncomment-3783366:

> Based on the discussion so far, I think the best course of action would be to create a separate theme for the project player and treat full screen mode as an extension of the project player -- so we would provide options for things like full screen background color, for example, in this new addon. This is the most logical connection because full screen mode is always directly accessed through the project player, and for this reason, I would assume that most users (with little technical knowledge about Scratch) also think of full screen mode as an extension of the project player. It would be best if we also [moved the project player addons to a top-level category](https://github.com/ScratchAddons/ScratchAddons/pull/3955) in order to increase this theme's visibility.
>
> Theming when the addon is disabled is a little trickier. Implementation complexity and how much sense the fallback(s) make both need to be considered here. Since the project player is logically an extension of the project player, the most logical fallback would be the context that the project player is in. This would require some additional style overrides for the project page, but it would otherwise not require much additional effort. For full screen mode, if we're following the same logic as when the addon is enabled, it should be themed after the context that the project player was opened from.

List of options to include in this new addon:
- Full screen header background color
- Full screen background color
- Monitor color, which is extended in [this PR](https://github.com/ScratchAddons/ScratchAddons/pull/3912)

Samq64:
Resolves #5210

### Changes

Right now it just splits off project player stuff from `editor-dark-mode` into a new addon. I'm planning to include more in this PR like monitor colour customization though.

### Reason for changes

`editor-dark-mode` should only effect the editor, not the the project page or embeds.

### Tests

Tested briefly on Chromium 106.


Samq64:
Maybe I should've waited until this was more complete before submitting it.

Samq64:
Too many conflicting styles.

Issue #5219, opened 2022-10-17
New addon: Snap to grid in paint editor


Endermaster3690:
### Why this would be helpful/solve a problem

I want to make a game and have these things snap to the grid without converting the image to bitmap.

### What this addon does

When the addon is turned on, there is an option in the paint editor that has a number that decides how the image will snap to the grid, for example 1 would make it snap to 1 bitmap pixel on the 480x360 canvas, 1.5 would make it snap to 1.5 pixels, 2 snaps to 2 pixels and so on. 0 would turn off snapping.

In the addons list, like the place you turn off and on addons, you can set a default that it is set to.

Make the addon either snap the top right or center of the selection to the grid. I'd prefer the center, but being able to switch through these would be nice to the users.

UtterDonkey:
You should close this now since it's been added.

mxmou:
> You should close this now since it's been added.

The snapping addon doesn't have an option to snap to a grid yet.

UtterDonkey:
Ah. I understand now.

UtterDonkey:
Would the options be like "snap to pixels:", "Offset X:" and "Offset Y"?

Endermaster3690:
> Would the options be like "snap to pixels:", "Offset X:" and "Offset Y"?

yes sounds good
oh my god you posted that on jan 8 i didnt reply until feb 19 i am so done with life

PR #5226, opened 2022-10-18
Addon Storage API

### LINKED ISSUE #1950: addon.settings.set
Some addons have setting menus that they create when you are in the scratch editor (ex. onion skinning). Could we add a setter to the addon object to save changes users may have made in those setting menus? I ask this mainly because it is a feature that I would like to be able to use for my keyboard editor addon.

example usage:
```js
addon.settings.set("option name", "value")
```

may also need something to get possible options of a setting:
```js
addon.settings.getOptions("option name")
//returns options array or type of option (ex. number or string)
```

nexensys:
Resolves #1950
*although it is still a somewhat different thing*

### Changes

Added the addon storage api:
`addon.storage`
  |- `local`
  |     |- `get: Promise<any>`
  |     |- `set: Promise<void>`
  |- `sync` 
  |     |- `get: Promise<any>`
  |     |- `set: Promise<void>`
`addon.storage`, `addon.storage.local`, and `addon.storage.sync` all extend Listenable and have change events fired
`addon.storage` catches change events from both `addon.storage.local` and `addon.storage.sync`.

This API is essentially a reorganized version of the `chrome.storage` api, but addons are unable to interfere with each other.

### Reason for changes

There are multiple addons that currently use `localStorage` to store preferences and other data, and would benefit from the `addon.storage.local` API:
- `animated-thumb`
- `dango-rain`
- `message-filters`

Addons that create a UI element could also benefit from this API, allowing setting changes to sync across tabs (via the `change` events) or even across browsers using `addon.storage.sync`, preventing users from needing to start from scratch with these settings which switching browsers or computers (The current sync features only store changes to settings in the manifest, not those in the addon's separate UI).

Not only does this provide more opportunities for addon developers, it can also reduce the messy aspect of string-based storage using `localStorage`.

This could also be merged with the `addon.settings.set` feature that has been discussed previously for similar effect.

@WorldLanguages Pinging you so that you see. Ping whoever you want to discuss with.


WorldLanguages:
> Can we try to rush this so that I can merge #5219 with `paint-snap`?

I don't think it's a good idea to intentionally block a new addon PR, and make it depend on a new SA API that doesn't exist yet. Couldn't this new addon work just like onion-skinning in the meantime (default settings on SA settings, and allow them to be changed per editor session)

ErrorGamer2000:
> > Can we try to rush this so that I can merge #5219 with `paint-snap`?
> 
> I don't think it's a good idea to intentionally block a new addon PR, and make it depend on a new SA API that doesn't exist yet. Couldn't this new addon work just like onion-skinning in the meantime (default settings on SA settings, and allow them to be changed per editor session)

@WorldLanguages It could, it does, and I'm not actually blocking the PR. I am simply saying that it would be nice to be able to just put them both together, rather than make a second pr to add it later. I was only saying that the grid aspect should wait.

WorldLanguages:
@ErrorGamer2000 I haven't looked deeply into the paint-snap PR, I assume "grid aspect" means a specific feature from the addon, independent from other features, which could be delayed until we add the storage API (but not all other features from the addon)? Is that what you mean?

ErrorGamer2000:
@WorldLanguages Yes, exactly.

PR #5231, opened 2022-10-19
rename-broadcasts: Update some strings

GarboMuffin:
Extracted from https://github.com/ScratchAddons/ScratchAddons/pull/5215

Scratch uses strings like "New Message" or "New message name:", so the addon should too

Maybe the addon's name should also be changed. I have no particularly strong feelings on that.

Took this out to a separate one because I don't know what the plan is for merging these this close to release


RedGuy12 (approved):

DNin01 (approved):

Hans5958 (approved):
LGTM

PR #5240, opened 2022-10-22
`better-quoter`: fix compatibility issues

pufferfish101007:
Resolves https://discord.com/channels/806602307750985799/1033021298181419109

### Changes

- moves the majority of `better-quoter`'s logic and some of `forum-id`'s to `addons/better-quoter/module.js`; this includes the pollution of `window.copy_paste`, automatically adding post links to quotes, generating post links, and generating bbcode from selected text
- `better-quoter`: do not add an author to the quote tag if the only part of the author selected is ` wrote:` (or part of that)
- `better-quoter`: ignore 'copy code' buttons added by `forums-copy-code`
- `better-quoter`: pollute `window.copy_paste` rather than modifying event handlers
- change various `let`s and `var`s to `const`
- `forum-id`: early return on closed topics

### Reason for changes

- fixes a bug where quoting selected text with `better-quoter` enabled would not automatically a post link, even when `forum-id` and its `auto_add` setting were enabled; this solution is a bit messy but it does work and is a bit more understandable (IMO) than other solutions
- fixes a bug where selecting text beginning within ` wrote:`, then continuing to outside the quote, then quoting, would result in a malformed quote author eg `rote: wrote: ...`
- fixes a bug where enabling `forums-copy-code`, then selecting text beginning before a code block and continuing through the code block, then quoting, would add sone text saying 'copy code' to the quote
- fixes a bug where `forum-id` would add the 'quote post number' button on closed topics

### Tests

tested on chromium 107 with various combinations of addons and dynamic enanle/disable


WorldLanguages:
@RedGuy12 I'd suggest you self-assign this PR, possibly 🤔

addons/better-quoter/module.js
```diff
@@ -0,0 +1,203 @@
+let forumIdAddon = null;
+let betterQuoterAddon = null;
+let isSetup = false;
```
WorldLanguages, line 3: Is `isSetup` ever set to true?

addons/forum-id/userscript.js
```diff
@@ -1,5 +1,6 @@
-/* global $, paste */
```
WorldLanguages, line 1: You should move this line into module.js

addons/better-quoter/module.js
pufferfish101007: Oops, looks like it isn't.
TBH it probably wouldn't cause any problems if it ran `setup()` twice, but I'll go fix that real quick anyway.

RedGuy12:
> @RedGuy12 I'd suggest you self-assign this PR, possibly 🤔

misunderstood the difference at first

pufferfish101007:
I appear to have slightly broken something with git here, apologies.

WorldLanguages:
Looking for an assignee

pufferfish101007:
Merge conflicts have been resolved. Tested on chromium 107 and everything works as expected as far as I can tell.

WorldLanguages:
Merging. PR author is interested in forum addons and most contributors are not. Also, these are bug fixes!

PR #5241, opened 2022-10-22
Auto snip images

### LINKED ISSUE #5095: Auto-snip Images on the Forums

Images get repeatedly quoted on the forums, taking up a lot of vertical space and network data.

### What this addon does

This addon would automatically convert quoted images into links.

For example, this:
```
Look at my great new banner!
[img]https://assets.scratch.mit.edu/get_image/.%2E/3318aae4d8ae240cc2eaca6c6d70cc85.png[/img]
```
Would get converted into this after quoting the post:
```
Look at my great new banner!
https://assets.scratch.mit.edu/get_image/.%2E/3318aae4d8ae240cc2eaca6c6d70cc85.png
```

### Possible alternatives

You could remove the [img] tags by hand, but I feel like an addon would be perfect for doing it automatically.

_No response_

Chiroyce1:
Resolves #5095 

### Changes

Creates a new addon called `auto-snip-images` to resolve issue, but this still doesn't implement the part where it also works for non SA users (when someone quotes a post that contains images, convert them to links)

### Tests

Tested on a separate Firefox profile, works 👌

### Considerations before merging
- The images flash for a split second, in between the page load and the addon run, so if there was a way to prevent that, it would be nice.
- Improve addon name/description (since I'm bad at that)

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -130,7 +130,7 @@
   "pick-colors-from-stage",
   "rename-broadcasts",
   "auto-snip-images",
-  
+
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
   "// because the first in the list gets their CSS injected first,",
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -129,7 +129,8 @@
   "vol-slider",
   "pick-colors-from-stage",
   "rename-broadcasts",
-
+  "auto-snip-images",
+  
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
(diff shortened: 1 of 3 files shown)


pufferfish101007:
Once #5240 is merged, `addons/better-quoter/module.js` will need to be changed in order to insert the `[img]` tag correctly I think.
If this pr is merged before that one then I'll update my pr.

WorldLanguages:
> The images flash for a split second, in between the page load and the addon run, so if there was a way to prevent that, it would be nice.

You could try `"runAtComplete": false`, and using `addon.tab.waitForElement` instead of `document.querySelectorAll`, but I'm not sure if it will make a big difference in the forums, because they are server-side rendered.

MystPi:
I still think this would be a useful addon to have!

PR #5245, opened 2022-10-23
`hide-flyout` animation speed changes

DNin01:
Resolves no open issues.

### Changes

- I changed the setting title from "Animation duration" to "Animation speed".
- I changed the option names from _None/Short/Default/Long_ to _Instant/Quick/Default/Slow_.
- I made each setting faster.

### Reason for changes

1. The Long setting was honestly _too_ slow.
2. Even on the Short setting, it still felt a little slower than I'd like.

### Tests

Changes tested in Edge 106. No problems found.

### DIFF
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -86,17 +86,17 @@
       "default": "cathover"
     },
     {
-      "name": "Animation duration",
+      "name": "Animation speed",
       "id": "speed",
       "type": "select",
       "potentialValues": [
         {
           "id": "none",
-          "name": "None"
+          "name": "Instant"
         },
         {
           "id": "short",
-          "name": "Short"
+          "name": "Quick"
         },
         {
           "id": "default",
@@ -104,7 +104,7 @@
         },
         {
           "id": "long",
-          "name": "Long"
+          "name": "Slow"
         }
       ],
       "default": "default"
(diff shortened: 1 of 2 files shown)


WorldLanguages:
I don't think it would be ok to change the value of the current settings (except "long", I don't think anyone uses that one)

I'd suggest:
1. Add a new setting for `0.19` (why exactly 19 and not 20, btw?)
2. Add a new setting for `0.31` (why not 30?)
3. Keep `0.5` instead of changing it to `0.46`
4. Rename the settings, considering the changes, in a way that makes sense

DNin01:
> 1. Add a new setting for `0.19` (why exactly 19 and not 20, btw?)
> 2. Add a new setting for `0.31` (why not 30?)
> 3. Keep `0.5` instead of changing it to `0.46`

Okay, I changed the settings so that we have the following options: `0`; `0.2`; `0.3`; `0.5`. I removed the `1`.

> 4. Rename the settings, considering the changes, in a way that makes sense

@WorldLanguages What would you prefer for the settings `0.2`; `0.3`; `0.5`? I'm thinking of _Faster; fast; normal_. And should the "Instant" option be renamed too?

WorldLanguages:
Actually, I think Instant/Quick/Default/Slow is fine

mxmou (approved):

RedGuy12 (approved):

WorldLanguages:
@DNin01 Could you open this PR again please?

DNin01:
> @DNin01 Could you open this PR again please? (Edit: I meant creating another one, but use the same branch)

Okay.

PR #5257, opened 2022-10-24
Changes to theme3 related to the new block colors

mxmou:
### Changes

Changes the black text mode in the `editor-theme3` addon to make it possible to recreate the new block colors (#5192):
Example screenshot

* The color of block and input text and some icons is now `#000`.
* Automatically generated variants of the primary colors (used for block borders and some input fields) are now very similar to those in the Scratch Lab.
* Text color of items in the find bar dropdown now has enough contrast even if the block colors are bright.
* Text color of hovered items in the find bar dropdown is now black again (it was accidentally changed to white in #4817).
* Text color of forum blocks with a color specified as a hex code is no longer changed to black because that often made the text unreadable. This is unrelated to the new colors but it's an improvement to the black text mode.

### Reason for changes

When Scratch implements the new colors, our defaults will need to match Scratch's as closely as possible. These improvements will also make the experience better for people who already use the black text mode.

### Tests

Tested different values of the "text color" setting.


WorldLanguages:
@mxmou You may self-assign this PR, if you'd like.
1.30 milestone

DNin01 (commented):
I downloaded the modified addon to test it out, and it's pretty darn close!

Comparison image

Noticed a few differences:
- Extension icons (for blocks from extensions like Music) are slightly different, but Scratch will update those. Might have to do a few changes to help make them work in both light and dark color palettes.
- The block borders are still a little different. Maybe we can get a little closer to Scratch's colors?
- A couple icons (like the counterclockwise arrow on the turn _n_ degrees block) still have an outline, but the ones on Scratch Lab don't have an outline.

CST1229:
> Comparison image

Which is Scratch Lab and which is SA?

mxmou:
> Which is Scratch Lab and which is SA?

SA is left, Scratch Lab is right.

mxmou:
> The block borders are still a little different. Maybe we can get a little closer to Scratch's colors?

Increasing the saturation would make them look more similar to Scratch's but I'm not sure if it would work well for colors set by the user.

> A couple icons (like the counterclockwise arrow on the turn _n_ degrees block) still have an outline, but the ones on Scratch Lab don't have an outline.

The addon could replace the icons with the new ones but that probably isn't necessary - the difference is small and will disappear when Scratch updates the icons.

DNin01:
> > The block borders are still a little different. Maybe we can get a little closer to Scratch's colors?
> 
> Increasing the saturation would make them look more similar to Scratch's but I'm not sure if it would work well for colors set by the user.

I realize that... Maybe there could be an option for border contrast (being able to choose between two contrast options for the block's borders)?

WorldLanguages:
Should we add a ScratchLab preset?

DNin01:
> > Should we add a ScratchLab preset?
> 
> 
> 
> I think that should be a separate PR because it's technically independent of these changes.

Yeah, that makes sense. I'm going to consider making a PR for this. I've already recreated the colors and made my own new dark variant.

WorldLanguages:
I don't think this can be reasonably merged in time for v1.30.0

addons/editor-theme3/theme3.js
```diff
@@ -260,6 +272,13 @@ export default async function ({ addon, console }) {
     return oldFieldMatrixCreateButton.call(this, fill);
   };
 
+  const oldFieldVerticalSeparatorInit = Blockly.FieldVerticalSeparator.prototype.init;
+  Blockly.FieldVerticalSeparator.prototype.init = function () {
+    // Vertical line between extension icon and block label
+    oldFieldVerticalSeparatorInit.call(this);
+    if (textMode === "black") this.lineElement_.setAttribute("stroke", this.sourceBlock_.getColourTertiary());
+  };
+
```
WorldLanguages, line 281: I think you forgot dynamicDisable support here.

dynamicDisable is coded weirdly on this addon. The `addon.self.disabled` checks aren't found where you expect in other traps.

addons/editor-theme3/black_text.css
```diff
@@ -1,21 +1,36 @@
+@namespace xlink url("http://www.w3.org/1999/xlink");
```
WorldLanguages, line 1: I didn't know CSS had a @namespace flag :P

WorldLanguages (approved):

Samq64 (approved):

WorldLanguages:
Can be merged!

Issue #5254, opened 2022-10-24
Arabic language close to being added (RTL)


WorldLanguages:
General translation for Arabic on Transifex is at 87.5% translated
I think the GitHub workflow will automatically add Arabic to /_locales at some point if translators keep pushing new strings, so we should pay close attention into enabling RTL support in our own webpages.

mxmou:
We also need to make sure that everything still looks as intended in RTL. An easy way to test this is adding `"en"` to `webpages/rtl-list.js`.

WorldLanguages:
The new https://scratchaddons.com/translate page now explicitly mentions Arabic and Hebrew are supported, but no one has asked to join for those languages yet.

The Arabic translator got pretty close to the 80% threshold, but it's not ready yet, and there was no activity in Arabic for a few months.

PR #5267, opened 2022-10-25
Fix some RTL bugs in webpages

mxmou:
### Changes

Fixes incorrect RTL styles.

### Reason for changes

The Arabic translation might be added soon. (#5254)

### Tests

Tested by adding English to `rtl-list.js`.

### DIFF
--- a/popups/scratch-messaging/popup.css
+++ b/popups/scratch-messaging/popup.css
@@ -327,7 +327,7 @@ a.message-type-title-text {
 .popout {
   height: 10px;
   vertical-align: baseline;
-  margin-left: 5px;
+  margin-inline-start: 5px;
   opacity: 0.6;
   filter: var(--content-icon-filter);
 }
--- a/webpages/settings/components/addon-setting.html
+++ b/webpages/settings/components/addon-setting.html
@@ -210,6 +210,9 @@
     border-radius: 5px;
     transition: all 0.25s ease;
   }
+  [dir="rtl"] .setting-input.check::before {
+    left: 25px;
+  }
 
   .setting-input.check:checked {
     background-color: var(--blue);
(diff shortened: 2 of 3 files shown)


apple502j (approved):

WorldLanguages:
Note that, technically, we shouldn't be using `margin-inline-start` because it won't work in Chrome 80-87.
But supporting RTL in modern Chrome is more important than inexistent margins in old Chrome, which few people use.

Here's how the settings page (rtl-fixes branch) looks in Chrome 80, in case you're curious (the checkboxes look good in Chrome 84+ however):
image

WorldLanguages (approved):

RedGuy12:
Related: `document.body.dir` is set without regard to `rtl-list.js`'s values:
https://github.com/ScratchAddons/ScratchAddons/blob/48106c8599fca407ded70aeb073085cd7b6b655a/webpages/set-lang.js#L4

WorldLanguages:
@RedGuy12 False alarm - look at the implementation for getDirection. It works as intended.

mxmou:
> Note that, technically, we shouldn't be using `margin-inline-start` because it won't work in Chrome 80-87. But supporting RTL in modern Chrome is more important than inexistent margins in old Chrome, which few people use.
> 
> Here's how the settings page (rtl-fixes branch) looks in Chrome 80, in case you're curious (the checkboxes look good in Chrome 84+ however)

Would adding prefixed versions of the properties (`-webkit-margin-start` and `-webkit-margin-end` for margins, `-webkit-appearance` for checkboxes) fix that?

WorldLanguages:
> Would adding prefixed versions of the properties (`-webkit-margin-start` and `-webkit-margin-end` for margins, `-webkit-appearance` for checkboxes) fix that?

Good idea, forgot those existed.
Not worth it, because we'll need to drop many Chrome versions after updating to MV3 anyway.

WorldLanguages:
This can be merged, right?

PR #5279, opened 2022-10-27
Remove `global` parameter from addon default functions

### LINKED ISSUE #5276: Remove useless "global" parameter from userscript default functions (entry points)
```js
export default async function ({ addon, global, console, msg}) {
```
You've seen this countless times, `global` is pointless. It's supposed to be a dictionary which is shared by all userscripts from the same addon running on the same tab, but that makes no sense. I added this back in [July 2020](https://github.com/ScratchAddons/ScratchAddons/pull/8/files#diff-2fecbd9e597c1f23b582774c2171c526cc81723344d5624ed737f0aa16c0a753R6), and no addon uses it. There's no point on running more than 1 userscript in the same page.
Popup entry points also have the `global` object, but there it's even more pointless lol, so it should be removed too.

The meaning of this parameter can be confusing to new contributors, so I'd suggest:
1. Stop creating an object which is not used by any userscript
2. Remove "global, " from the head of the default function of every userscript that still has it

RedGuy12:
Resolves #5276 

### Changes

Removes the `global` object from all userscripts, as well as an unused and undefined `cons` destructure from `editor-extra-keys`.
### Reason for changes

It's unused


Hans5958:
Would there be cases of where global can be used, but then left unused in favor of other methods. Is there a potential where `global` can be used?

apple502j (approved):

WorldLanguages (approved):

PR #5283, opened 2022-10-28
pick-colors-from-stage: only process user-initiated clicks

GarboMuffin:
Resolves https://discord.com/channels/806602307750985799/1035494962193772604

> When I set the opacity in the sprite editor for some reason when I hover over stage it thinks that I am going to select a colour to use for sprite editing.

Probably also affects the other addons that use the color picker.


GarboMuffin:
opened this a bit early it seems, having some trouble reproducing both the original bug and fix now. will figure that out at some point

apple502j (approved):

Samq64 (approved):
I can't reproduce the bug either, but this PR can't make anything worse.

PR #5285, opened 2022-10-29
block-duplicate: respect block-cherry-picking's invert setting 

GarboMuffin:
Resolves https://discord.com/channels/806602307750985799/1034089671984169030

Previously, if you had block-cherry-picking enabled in invert mode, normal block drags would only drag one block but alt+dragging would duplicate the entire stack. Now alt+dragging will just duplicate the one block unless ctrl is being held. This is what users would expect to happen.


WorldLanguages:
This is a 1.29.0 regression I suppose?

GarboMuffin:
Not a regression

apple502j (approved):

Samq64 (approved):

Issue #275, opened 2020-09-13
Better way to provide dark mode for every addon


mxmou:
**Is your feature request related to a problem? Please describe.**

@WorldLanguages recently ported some editor dark themes. Currently the addon overrides CSS from other addons, which is not good for multiple reasons:

- The order in which addons are loaded is important, which might cause problems with circular dependencies (which are avoidable and not a good practice anyway but it's better if addon developers don't need to care about that).

- Addons should be independent of each other, i.e. when making an addon one shouldn't need to change other addons. Currently the dark mode addon needs to be changed every time another addon wants to support dark mode.

**Describe the solution you'd like**

Something similar to `setting_match` that would test if another addon is enabled, maybe like this:

```json
{
  "userstyles": [
    {
      "url": "dark.css",
      "matches": ["..."],
      "if_addon_enabled": "editor-dark-mode"
    }
  ]
}
```

**Describe alternatives you've considered**

Of course it could stay the way it is, but then addon development would be more complicated. Or there could be a function in the addon API that would load a userstyle and another function that would check if an addon is enabled so it would be possible to inject userstyles dynamically just like you can `import()` userscripts.

**Additional context**

I will try to implement this myself if @WorldLanguages agrees.

mxmou:
Note that the solution I suggested doesn't allow addons to provide multiple different dark mode userstyles.

WorldLanguages:
I'm not fully convinced. Maybe because I just ported editor dark modes (and not website ones), where there's a lot more styles, but less addons adding UI to it.
I fail to see how this helps, here's an example: the colorful context menus addon.
We of course don't want the context menus to be dark if both the colorful context menus, and editor dark mode addons are enabled. So I simply made colorful context menus more important by putting it at the bottom of the addons.json list.
Using this, how would that work? The colorful context menus doesn't care if dark mode is enabled or not, it doesn't change anything for it.  So now, the dark mode addon needs to check whether colorful context menus is enabled, identify those context menus (not all context menus are colored, only right-click block ones), and not apply dark mode styles to them. To detect those context menus, the dark mode addon would need a userscript. Doesn't sound good to me.
Another point: this way of doing dark mode also means dark mode is applied to third party extensions like Scratch Developer Tools by griffpatch, even if it's loaded as an extension, and not as an addon. If that wasn't the case, there would be a white rectangle in the middle of the dark editor.

mxmou:
This is now possible using CSS variables.

[...]

PR #5295, opened 2022-10-30
New Addon: Share through My Stuff

Steve0Greatness:
From [Share Projects through MyStuff](https://discord.com/channels/806602307750985799/1035719064347234415) on the Discord.

### Changes

On the My Stuff page, there's now a button on unshared projects next to delete that lets you share your project.

### Reason for changes

It's annoying to have to open up the project page or editor just to share a project.

### Tests

Tested on:

* Chrome
* Firefox

Both with and without many other addons enabled.


addons/share-through-mystuff/addon.json
```diff
@@ -0,0 +1,27 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/mystuff"],
+      "runAtComplete": true
+    }
+  ],
+  "dynamicDisable": true,
+  "dynamicEnable": true,
+  "tags": ["community"],
```
DNin01, line 25: I think you need to add `"versionAdded": "<VERSION>"` to the addon manifest in order for the check to pass.
```suggestion
  "tags": ["community"],
  "versionAdded": "1.30.0",
```

DNin01:
And I think the last failed check is because you might need to turn on workflows [here](https://github.com/Steve0Greatness/ScratchAddons/actions). Probably don't need to worry that it failed here, but you can still turn on workflows so that check on your future PRs pass.

addons/share-through-mystuff/userscript.js
```diff
@@ -0,0 +1,35 @@
...
+    await addon.tab.waitForElement(".media-item-content.not-shared");
+    let elements = document.querySelectorAll(".media-item-content.not-shared").length;
+    next_button = document.querySelector(next_button_selector);
+    next_button.addEventListener("click", event)
+    addon.tab.addEventListener("urlChange", event);
+    run(0, elements);
+  }
+  function event(wait = 500) {
+    console.log("Event Triggered")
+    setTimeout(basics, wait); // allows for time to load
```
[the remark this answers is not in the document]
Steve0Greatness: I'll try that when I get back from breakfast.

Steve0Greatness:
Probably be good as a settings option enable by default.

addons/share-through-mystuff/addon.json
```diff
@@ -20,6 +20,14 @@
       "runAtComplete": true
     }
   ],
+  "settings": [
+    {
+      "name": "Ask if the project should be shared",
```
DNin01, line 25: `confirm-actions` already has a confirmation setting for sharing projects.
https://github.com/ScratchAddons/ScratchAddons/blob/48106c8599fca407ded70aeb073085cd7b6b655a/addons/confirm-actions/addon.json#L22
This addon _could_ use that setting instead (so we have just one setting, easy to use), but then users may want to not/only have a confirmation for sharing **from the My Stuff page**. Worth discussing.

WorldLanguages: IMO, this new addon shouldn't have any settings. It should always ask to confirm.
If you want to share many projects rapidly without confirmation, just open 1 tab per project.

addons/share-through-mystuff/userscript.js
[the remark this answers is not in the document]
Steve0Greatness: Didn't work.

addons/share-through-mystuff/addon.json
Steve0Greatness: One tab per project gets easily cluttered--if it's more than ~2 and makes the process of sharing a project 10% slower when you have a slower connection.

WorldLanguages: I don't think we should take into account the potential need to share multiple projects at once. This is Scratch Addons, not Scratch Macros.

Steve0Greatness: fine.

Steve0Greatness:
I removed the setting. Now it always has a confirmation box.

addons/share-through-mystuff/userscript.js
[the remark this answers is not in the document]
WorldLanguages: @Steve0Greatness We can't accept addons that use this pattern of timeouts and intervals to wait until certain elements exist.
You should use `addon.tab.waitForElement`. If you find any bugs, or need help on how to using, please tell

WorldLanguages:
@Steve0Greatness Could you tell us why you closed this PR?

Steve0Greatness:
I've fully remade the addon. I got frustrated last time I worked on this.

Samq64:
The share button isn't lined up right when disabling `scratchr2`.
image

Steve0Greatness:
huh

Steve0Greatness:
@Samq64 try it now, it should be fixed.

addons/share-through-mystuff/userscript.js
```diff
@@ -0,0 +1,25 @@
+export default async function ({ addon, global }) {
+  const share_function = document.createElement("a");
```
Samq64, line 2: Using the word function in an object name is a little confusing. Could you use a different variable name and use `camelCase`?

addons/share-through-mystuff/userscript.js
```diff
@@ -0,0 +1,25 @@
...
+  share_function.dataset.control = "share";
+  share_function.href = "#";
+
+  const share_button = document.createElement("a");
+  function share_confirmation(event) {
+    let confirmation = confirm("Would you like to share this project?");
+    if (confirmation) event.target.parentElement.querySelector(".media-share").click();
+  }
+  share_button.href = "#";
+  share_button.classList.add("__share_button__");
```
Samq64, line 13:
```suggestion
  share_button.classList.add("sa-share-button");
```

addons/share-through-mystuff/userstyle.css
```diff
@@ -0,0 +1,16 @@
+.media-item-content.not-shared .media-action {
+  top: 20px !important;
+}
+
+.__share_button__ {
```
Samq64, line 5:
```suggestion
.sa-share-button {
```

addons/share-through-mystuff/userstyle.css
```diff
@@ -0,0 +1,16 @@
+.media-item-content.not-shared .media-action {
+  top: 20px !important;
```
Samq64, line 2: `top` is set to `0px` lower down in the file, so maybe that's why you used `!important` here? If so could you remove this and set it to 20px down there?

Steve0Greatness: Honestly, I have no idea why that's there. After a bit of testing it seems that all it does is merge the delete and share buttons, making them impossible to read, let alone click. Here's what it looks like for me if I remove both of them:
Screenshot from 2023-01-20 19-39-52

Steve0Greatness: I can't get it to work without both(as in both buttons are layered on top of each other without both). Not sure what's up with that, or if I'm doing something wrong.

Samq64:
Could you briefly explain why you're creating 2 `a` elements?

Steve0Greatness:
`shareFunction` is there so that I don't have to deal with making a request to an internal Scratch API, `shareButton` is there so that a prompt can ask the user if they meant to share the project.

addons/share-through-mystuff/userstyle.css
Samq64: I still don't think `!important` is necessary, but I could be wrong.

Samq64:
Could you use `displayNoneWhileDisabled` on your button?

addons/share-through-mystuff/userstyle.css
Steve0Greatness: I think you're right, I'll remove `!important`.

Samq64 (approved):

addons/share-through-mystuff/userscript.js
```diff
@@ -0,0 +1,28 @@
+export default async function ({ addon, global }) {
+  const shareFunction = document.createElement("a");
+  shareFunction.classList.add("media-share");
+  shareFunction.dataset.control = "share";
+  shareFunction.href = "#";
+
+  const shareButton = document.createElement("a");
+  function shareConfirmation(event) {
+    let confirmation = confirm("Would you like to share this project?");
```
WorldLanguages, line 9: I'll soon change this to use `addon.tab.confirm()` instead

DNin01: Steve0Greatness has updated the code and it now uses `addon.tab.confirm()`.

Steve0Greatness:
The last commit made the styling work better with Scratch 2 -> Scratch 3(id `scratchr2`). Before, with both enabled the share and delete buttons would be all the way to the edge of the container, which just looked bad. Now they should be lined up with the unshare button in shared projects' containers.

CST1229:
Sharing a project while on the Unshared Projects tab seems to do nothing (I don't have the Confirmations addon enabled, by the way).

https://user-images.githubusercontent.com/68464103/216142243-21618778-97c3-40ff-a9c4-48fcfcb9a0cc.mp4

Steve0Greatness:
I now know why that was happening, using `#` in the link would redirect back to All Projects. You can see that happening in the video.

Also, sorry for the long response time.

Samq64 (commented):
Could one of the links be removed and instead use `e.preventDefault()` if the user cancels the confirmation?

Steve0Greatness:
`e.preventDefault()` likely wouldn't have an effect. `preventDefault` only prevents out-of-the-box events, such as opening a link or submitting a form, that are built into Chromium(and other web browsers). As for the project sharing after `shareFunction` is "clicked\:" it's not built into the web browser as part of a zero-JavaScript environment.

addons/share-through-mystuff/userscript.js
```diff
@@ -2,26 +2,27 @@ export default async function ({ addon }) {
...
 
   while (true) {
     const project = await addon.tab.waitForElement("div.media-item-content.not-shared", {
       markAsSeen: true,
     });
-    project.querySelector(".media-action div").appendChild(shareFunction.cloneNode());
+    let localShareFunction = shareFunction.cloneNode();
     let localShare = shareButton.cloneNode(true);
+    localShareFunction.href = `${location.hash}`;
+    localShare.href = localShareFunction.href;
```
Samq64, line 24: This seems to cause the page to reload before the confirmation pops up.

Samq64 (commented):
You're right about `e.preventDefault`.

On the unshared tab when sharing a project all of the projects on the tab disappear.

addons/share-through-mystuff/userscript.js
Steve0Greatness: `location.hash` has a `#` at the beginning of the string it returns (as long as there is one to return). A `#` at the beginning of a `href` makes it _not_ reload the page. So I a don't think this is the problem.

Samq64: There are two separate bugs: the first is when the all projects tab is selected clicking the share link reloads the page instantly without sharing the project. The second is when the unshared projects tab is selected sharing works as expected and doesn't reload the page, but after sharing the project all the project on that tab disappear untul I reload it.

Steve0Greatness: Try sharing a project with the share project tab open.

Steve0Greatness: Also I think I fixed the other bug.

Samq64 (approved):

addons/share-through-mystuff/scratchr2.css
```diff
@@ -0,0 +1,3 @@
+.media-item-content.not-shared .media-action {
+  width: 75px;
+}
```
CST1229, line 3: Somewhat nitpicky, but I think the general rule of thumb for addon elements affected by other addons (like dark mode and 2.0->3.0) is to have the code that affects the addon elements in the addon that affects them, not in the addon that adds the element.

Steve0Greatness: STMS-Scratchr2css-moved-confirmation
That?

Steve0Greatness: So rename `scratchr2.css`(probably to something like `share-through-mystuff.css`), move it to `/addons/scratchr2/`(and also edit the `addon.json`s to reflect these changes)?

CST1229: > So rename `scratchr2.css`(probably to something like `share-through-mystuff.css`), move it to `/addons/scratchr2/`(and also edit the `addon.json`s to reflect these changes)?

I mean merging scratchr2.css into `addons/scratchr2/mystuff.css` (without any `if`s).
`

WorldLanguages: I don't think there's consensus about this specific case. There's #275 but specifically about dark mode.

addons/share-through-mystuff/userscript.js
```diff
@@ -0,0 +1,36 @@
+export default async function ({ addon, console }) {
+  const shareFunction = document.createElement("a");
+  shareFunction.classList.add("media-share");
+  shareFunction.dataset.control = "share";
+
+  const shareButton = document.createElement("a");
+  shareButton.classList.add("sa-share-button");
+  shareButton.innerText = "Share";
```
WorldLanguages, line 8: https://scratchaddons.com/docs/develop/userscripts/#localizing-userscripts

addons/share-through-mystuff/userscript.js
```diff
@@ -0,0 +1,36 @@
...
+  addon.tab.displayNoneWhileDisabled(shareButton, {
+    display: "block",
+  });
+  async function shareConfirmation(event) {
+    event.preventDefault();
+    let confirmation = await addon.tab.confirm("Share this project?", "Are you sure you want to share this project?");
+    if (confirmation) {
+      if (location.hash == "#unshared") {
+        let container = event.target.parentElement.parentElement.parentElement.parentElement;
```
WorldLanguages, line 17: Try `event.target.closest(".media-list > li")`

WorldLanguages (approved):
LGTM. Waiting for someone else to check my changes

Samq64 (approved):

WorldLanguages:
@mxmou You might want to check the CSS with `"if": { "addonEnabled": ["scratchr2"] }`

addons/share-through-mystuff/scratchr2.css
mxmou: Things like this can usually be done without any `if`s. I made a PR that rewrites this addon's CSS: #5816

PR #5302, opened 2022-11-01
Fix unreadable text in jQueryUI modals in dark mode

### LINKED ISSUE #5298: Unreadable text in "report profile" modal
### Description

image
The profile report screen goes blank when using dark mode.

### Steps to reproduce

using it
image

_No response_

### Environment

Browser: google chrome
Operating system: Windows 10

mxmou:
Resolves #5298

### Changes

Sets `color-scheme` on iframes to light. The explanation is in a code comment.

### Reason for changes

To make text in certain modals readable.

### Tests

The profile report modal is now readable:
Screenshot of the modal

### DIFF
--- a/addons/dark-www/experimental_scratchr2.css
+++ b/addons/dark-www/experimental_scratchr2.css
@@ -664,3 +664,10 @@ input.link.black {
   background-image: url("%addon-self-dir%/assets/meow.png");
   background-size: 20px 15px;
 }
+
+/* If the color scheme of an iframe element is different from that
+   of the document inside, it gets an opaque backgound. The .iframeshim
+   is meant to be entirely transparent. */
+.iframeshim {
+  color-scheme: light;
+}


apple502j (approved):

WorldLanguages:
I believe this could be cherry-picked into the v1.29.x branch. We could release v1.29.1 in a few days, to avoid having an unreadable report modal for ~50% of SA users for a whole month. What do you think?
Some other commits could also be cherry-picked.

WorldLanguages (approved):
Tested (Chrome)

WorldLanguages:
@mxmou Have you noticed that, on v1.29.0 stable, this bug does not happen if you dynamically enable dark-www? I can only reproduce if the website dark mode addon is enabled prior to loading the profile. Any idea why?
This is not important, but it could mean there's some bug in the code that handles userstyles or something related?

mxmou:
> @mxmou Have you noticed that, on v1.29.0 stable, this bug does not happen if you dynamically enable dark-www? I can only reproduce if the website dark mode addon is enabled prior to loading the profile. Any idea why? This is not important, but it could mean there's some bug in the code that handles userstyles or something related?

The `color-scheme` is set correctly so that might be a Chromium bug. On Firefox the iframe always has a white background, even if the addon is enabled dynamically.

PR #5006, opened 2022-08-19
Addon from devtools: Middle click block dropdown

TheColaber:
Related: https://github.com/ScratchAddons/ScratchAddons/issues/3700

### Changes

Rewrite/Class-ify parts of the devtools feature ctrl space dropdown, to a new addon.

### Tests

todo

### Todo
- [x] getting blocks from the dropdown currently don't go to mouse pointer (@Tacodude7729 epic legend )
- [x] using enter to create block currently does not have the mouse drag it ~~and it probably never will~~
- [x] Look at CSS file and rename classes to have `sa-` prefix
- [x] darkmode
- [x] other stuff probably
- [x] use class names with dashes instead of camel casing

PR #5305, opened 2022-11-01
Transition code for middle-click-popup

TheColaber:
Consistent with the transition code above. #5006


WorldLanguages:
Reminder: this new addon has `enabledByDefault: true`, so we need to make sure it stays disabled if the user had `editor-devtools` disabled.

TheColaber:
That is exactly what this transition code does.

apple502j (approved):

WorldLanguages (approved):
Tested

PR #5308, opened 2022-11-02
Add two new presets to `editor-theme3`

DNin01:
### Changes

- Added two presets.
  - The first preset is Scratch's updated block colors as lab.scratch.mit.edu/contrast-blocks
  - The second preset was made by me - it's called "Dark"
- The existing "Dark" preset was renamed to "Darker"
- Some presets were renamed
- The preset buttons now have previews

### Reason for changes

To give users the option to use the new block colors early, and to prepare the addon for the block colors update

### Tests

Tested on Edge 109 (with the latest changes from upstream).


apple502j:
Waiting for #5257

DNin01:
I suppose we should also add an update notification.

DNin01:
#5257 is merged. Ready for review.

addons/editor-theme3/addon.json
```diff
@@ -367,32 +436,10 @@
...
-        "comment-color": "#423f24",
-        "input-color": "#202020",
-        "text": "white"
-      }
-    },
     {
       "name": "Black",
       "id": "black",
-      "description": "Makes block backgrounds black",
+      "description": "Makes block backgrounds black and borders/text colored",
```
DNin01, line 442: Could we call this preset "Neon"?

WorldLanguages: The chemical element?

DNin01: > The chemical element?

Yeah. They just look reminiscent of neon lights to me.

WorldLanguages: It might be a bit more confusing. But at least it's not hard to translate into other languages, since it's just the name of a chemical element.

DNin01: Both names self-explain different aspects, so it's hard to choose which one to use.

WorldLanguages: Why not "neon black"? :P

DNin01: Oh, that's a good option.

mxmou: I'm not sure if that's a very good name - the main purpose of the preset is to make the background darker, not the text brighter.

addons/editor-theme3/addon.json
```diff
@@ -324,9 +327,75 @@
...
+        "sounds-color": "#E19DE1",
+        "events-color": "#FFD966",
+        "control-color": "#FFBE4C",
+        "sensing-color": "#85C4E0",
+        "operators-color": "#7ECE7E",
+        "data-color": "#FFA54C",
+        "data-lists-color": "#FF9966",
+        "custom-color": "#FF99AA",
+        "Pen-color": "#13ECAF",
+        "sa-color": "#12EDD5",
```
mxmou, line 345: I think `#57DBD7` might be better - it's less saturated and closer to the default color for these blocks.

addons/editor-theme3/addon.json
```diff
@@ -324,9 +327,75 @@
...
+        "events-color": "#AE8500",
+        "control-color": "#B07700",
+        "sensing-color": "#468CAF",
+        "operators-color": "#438C43",
+        "data-color": "#B46313",
+        "data-lists-color": "#A84210",
+        "custom-color": "#973545",
+        "Pen-color": "#2A7468",
+        "sa-color": "#2A797C",
+        "comment-color": "#242422",
```
mxmou, line 368: I don't think this preset should make comments dark gray, especially if it's meant to be less dark than the other "dark" one.

addons/editor-theme3/addon.json
```diff
@@ -324,9 +327,75 @@
...
+    {
+      "name": "Dark",
+      "id": "new-dark",
+      "description": "Dark versions of the default colors that look good in dark themes",
+      "values": {
+        "motion-color": "#2461B7",
+        "looks-color": "#633EAC",
+        "sounds-color": "#863286",
+        "events-color": "#AE8500",
+        "control-color": "#B07700",
```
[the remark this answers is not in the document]
DNin01: This color already only passes AA18px for the WCAG guidelines, and this proposed color doesn't pass any of them. I think we should prioritize text readability over ease of color differentiation. Users can always change it if they want to.

I'll change the control color slightly, though.

<sub>Information source: https://contrastchecker.com/</sub>

mxmou (approved):

addons/editor-theme3/addon.json
Samq64, line 442:
```suggestion
      "description": "Makes block backgrounds black and the borders and text colored",
```

addons/editor-theme3/addon.json
```diff
@@ -411,5 +458,9 @@
         "text": "colorOnBlack"
       }
     }
-  ]
+  ],
+  "presetPreview": {
+    "type": "palette",
+    "colors": ["motion-color", "looks-color", "sounds-color", "events-color", "control-color", "sensing-color"]
```
Samq64, line 464: Maybe variables could be added too?

Samq64 (approved):

mxmou:
Could "high contrast" be moved after the dark and black presets? It will probably be used by less people, and it looks like it won't actually become a new default.

addons/editor-theme3/addon.json
WorldLanguages: Indifferent about this. We aren't that strict with preset descriptions. For example, we don't end the sentences with a period within preset descriptions

WorldLanguages:
Can this be merged? Are Samq's concerns solved?

Samq64:
> Are Samq's concerns solved?

No, but ir doesn't really matter.

PR #5323, opened 2022-11-05
Fix comment replies on deleted profiles cause messaging to fail

### LINKED ISSUE #5318: Comment replies on deleted profiles cause messaging to fail
### Description

If you open up the Scratch Messaging popup tab, it loads messages. As soon as it reaches each comment or reply, it fetches the comments from the page where the message is from. If the profile is banned, it returns a 403 error, as shown below:

Screenshot 2022-11-03 at 5 56 42 PM

The popup seems to be prepared for a 404, but not a 403 error (according to a background message I saw, and then found the source for [here](https://github.com/ScratchAddons/ScratchAddons/blob/master/popups/scratch-messaging/api.js#L164-L167).

After a 403 error is returned, the popup immediately fails and shows this:

The solution is to not only stop if there's a 404 error but also if there's a 403 error.

### Steps to reproduce

1. Comment on a profile and get replied to
2. Wait for it to get banned
3. Open the mmessages popup and wait for it to try loading comments from profile

_No response_

### Environment

Browser: Chrome 106
Scratch Addons version: v1.29 (tested on multiple other beta versions though)
Operating system: MacOS

rgantzos:
Resolves #5318

### Changes

This changes `popups/scratch-messaging/api.js` to be prepared for 403 errors when fetching comments or replies, as well as 404 errors.

### Reason for changes

The Scratch Messaging popup would error whenever it tried fetching comments/replies from banned profiles. It's only ready for 404 errors when fetching, not 403.

### Tests

I did test:
Browser: Chrome 106
Scratch Addons version: v1.29 (tested on multiple other beta versions though)
Operating system: MacOS

### DIFF
--- a/popups/scratch-messaging/api.js
+++ b/popups/scratch-messaging/api.js
@@ -197,7 +197,7 @@ export async function fetchMigratedComments(
     const getReplies = async (offset) => {
       const repliesRes = await fetch(getRepliesUrl(parentId, offset));
       if (!repliesRes.ok) {
-        if (repliesRes.status === 404 || repliesRes === 403) return;
+        if (repliesRes.status === 404 || repliesRes.status === 403) return;
         throw HTTPError.fromResponse(`Ignoring comment ${resourceType}/${commentId}`, repliesRes);
       }
       const repliesJson = await repliesRes.json();
--- a/popups/scratch-messaging/api.js
+++ b/popups/scratch-messaging/api.js
@@ -162,7 +162,7 @@ export async function fetchMigratedComments(
     const res = await fetch(getCommentUrl(commentId));
 
     if (!res.ok) {
-      if (res.status === 404) continue;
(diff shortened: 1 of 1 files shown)


popups/scratch-messaging/api.js
```diff
@@ -197,7 +197,7 @@ export async function fetchMigratedComments(
     const getReplies = async (offset) => {
       const repliesRes = await fetch(getRepliesUrl(parentId, offset));
       if (!repliesRes.ok) {
-        if (repliesRes.status === 404) return;
+        if (repliesRes.status === 404 || repliesRes === 403) return;
```
apple502j, line 200:
```suggestion
        if (repliesRes.status === 404 || repliesRes.status === 403) return;
```

apple502j (approved):

WorldLanguages (approved):

PR #5333, opened 2022-11-07
Remove add border setting from `block-switching` addon and just always add the border

### LINKED ISSUE #5326: Remove "add border" setting from `block-switching`

The "add border" option in the `block-switching` addon right here:
https://github.com/ScratchAddons/ScratchAddons/blob/c4818311c48ef3223bd6ee8f766d1e8272cc7a8b/addons/block-switching/addon.json#L31
...isn't really that important of a setting. When this is on, there's a border/line separating the block right click actions from the block switch options. When it's off, the border isn't in-between those sets of options.

I'm sure nobody really cares if the border is there or not. Removing the option and just _always_ showing the border might make the addon easier to use, since the setting is kind of hard to understand. (_What's "border"? "Add" where?_)

### What this change does

I suggest we remove the option and make it so that the border separating the block actions from the block switch options are _always_ there. Could be called "Show separation line"? Might make a bit more sense that way, even if the user doesn't figure it out immediately.

### Possible alternatives

We could also maybe change the setting name.

_No response_

DNin01:
Resolves #5326

### Changes

- The "Add border" setting has been removed. The border separating the right click context menu actions and block options will now always be there.
- Settings have been rearranged - the option to change the block to itself is at the bottom.

### Reason for changes

It honestly doesn't really matter for that part of the addon to be customizable.

### Tests

Tested on Edge 107. Looks like everything works.


Samq64 (approved):

WorldLanguages:
I hope to get **more than 2 approvals** here, because we are removing an existing setting.

mxmou (approved):

RedGuy12 (approved):

CST1229 (approved):

PR #5260, opened 2022-10-24
`hide-flyout` animation speed changes

DNin01:
Resolves no open issues.

### Changes

- I changed the setting title from "Animation duration" to "Animation speed".
- I changed the option names from _None/Short/Default/Long_ to _Instant/Quick/Default/Slow_.
- I made each setting faster.

### Reason for changes

1. The Long setting was honestly _too_ slow.
2. Even on the Short setting, it still felt a little slower than I'd like.

### Tests

Changes tested in Edge 106. No problems found.

PR #5338, opened 2022-11-08
Change `custom-zoom` animation speeds

DNin01:
Resolves no open issues. Related to #5260.

### Changes

Applies the same animation speed changes I made to `hide-flyout` last month (PR linked above) to `custom-zoom`'s autohide zoom controls feature.

### Reason for changes

1. I don't think anyone was going to use the Long setting.
2. Even on the Short setting, it still felt just a _little_ slower than I'd like.

### Tests

Tested on Edge 107. Everything seems fine.

### DIFF
--- a/addons/custom-zoom/addon.json
+++ b/addons/custom-zoom/addon.json
@@ -62,17 +62,17 @@
       "default": false
     },
     {
-      "name": "Autohide Animation duration",
+      "name": "Autohide Animation Speed",
       "id": "speed",
       "type": "select",
       "potentialValues": [
         {
           "id": "none",
-          "name": "None"
+          "name": "Instant"
         },
         {
           "id": "short",
-          "name": "Short"
+          "name": "Quick"
         },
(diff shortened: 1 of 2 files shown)


RedGuy12 (approved):

mxmou (approved):

DNin01:
Now I wonder, what if we just replace the speed settings for this addon with an on/off switch? I'm not sure how many people would want or need to change the speed.

What do you think of that?

I think we should keep the speed settings on `hide-flyout`, though.

BroJac5246:
> Now I wonder, what if we just replace the speed settings for this addon with an on/off switch? I'm not sure how many people would want or need to change the speed.
> 
> What do you think of that?
> 
> I think we should keep the speed settings on `hide-flyout`, though.

You're overthinking this, it's fine as it is.

DNin01:
Ready to merge, @WorldLanguages?

Issue #3059, opened 2021-07-18
[New Addon] The Scratch Font


CodeGuy92:
I'm really sorry but i am in a really big rush right now, so i cannot fill in the correct form. I'll edit it later.

I want an addon that will enable the Scratch Font in the editor.

lisa-wolfgang:
I believe that the devs are planning on pulling support for the font in the near future, even from projects that already have it.

This should be part of the more general-purpose suggestion to add custom fonts: #2050

mxmou:
> I believe that the devs are planning on pulling support for the font in the near future, even from projects that already have it.

This is correct. See LLK/scratch-render-fonts#23

Issue #2050, opened 2021-04-02
Ability to Upload Fonts to Editor


Scratchfangs:
I think you should be able to add more fonts. There could be a button under all the fonts that says "upload new font". And once you upload the font, it would be saved to your storage for use in different projects. Why can't you just use Google Drawings or something? When you try editing that in Scratch, it would go back to default Sans Serif.

![image](https://user-images.githubusercontent.com/71499307/113422918-85dd0300-939b-11eb-8835-4aab9dadbe14.png)

apple502j:
We don't have storage to store uploaded fonts, rejected.

Scratchfangs:
> We don't have storage to store uploaded fonts, rejected.

I said that it could be saved to the computer storage.

mxmou:
> > We don't have storage to store uploaded fonts, rejected.
> 
> I said that it could be saved to the computer storage.

Other users would not see it.

TheColaber:
You can store a link to the font in the svg? I think I've done this with Sora before...

mxmou:
@TheColaber This is only possible with fonts that are available online, not with uploaded fonts.

TheColaber:
True, so why not do that? Why not use some API or something, so the user can input "Sora" and returns a URL to it's styles

Hans5958:
Reopening. Classic Apple closing stuff for reading only a portion of the text. Alternatives may exist in various form.

Hans5958:
> @TheColaber This is only possible with fonts that are available online, not with uploaded fonts.

Well, he said "the link," not "the file".

I don't know, is it possible to embed fonts to `.svg`s?

WorldLanguages:
We could provide more fonts by converting SVG text into path (but somehow storing metadata that will still allow the user to edit the path as if it was text)

TheColaber:
> We could provide more fonts by converting SVG text into path (but somehow storing metadata that will still allow the user to edit the path as if it was text)

In that case, there might as well be a scratch tool website to do this. Enter text, choose point, download svg. Then import to scratch

Scratchfangs:
Downloading is a little too much, I have more than 100 files just from doing that from Google Drawings

CubeyTheCube:
> > > We don't have storage to store uploaded fonts, rejected.
> > 
> > 
> > I said that it could be saved to the computer storage.
> 
> Other users would not see it.

[...]

PR #5344, opened 2022-11-11
New addon: More costume fonts

valmontechno:
An add-on that adds the following fonts to the costume editor.
Scratch, Arial, Helvetica, Verdana, Impact and Monospace.

### DIFF
--- a/addons/more-costume-fonts/addon.json
+++ b/addons/more-costume-fonts/addon.json
@@ -1,0 +1,40 @@
+{
+	"name": "More costume fonts",
+	"description": "Added new fonts for the costume editor.",
+	"tags": ["editor", "costumeEditor"],
+	"info": [
+		{
+			"type": "notice",
+			"text": "Some fonts may not be compatible with all devices.",
+			"id": "font-compatibilities"
+		}
+	],
+	"credits": [
+		{
+			"name": "Valmontechno",
+			"link": "https://scratch.mit.edu/users/valmontechno/"
+		},
+		{
+			"name": "CST1229",
+			"link": "https://github.com/CST1229/"
+		},
+		{
+			"name": "rgantzos",
+			"link": "https://github.com/rgantzos"
+		}
+	],
+	"userscripts": [
+		{
+			"url": "userscript.js",
+			"matches": ["projects"]
+		}
+	],
+	"userstyles": [
+		{
+			"url": "style.css",
+			"matches": ["projects"]
+		}
+	],
+	"enabledByDefault": false,
+	"versionAdded": "1.30.0"
+}
--- a/addons/more-costume-fonts/userscript.js
+++ b/addons/more-costume-fonts/userscript.js
@@ -1,0 +1,63 @@
+export default async function ({ addon, console }) {
+	
+	const paper = await addon.tab.traps.getPaper();
+	const updateImage = paper.tool.onUpdateImage.bind('paper.tool');
+	const fonts = [
+		{ 'name': 'Sans Serif', 'family': 'Sans Serif' },
+		{ 'name': 'Serif', 'family': 'Serif' },
+		{ 'name': 'Handwriting', 'family': 'Handwriting' },
+		{ 'name': 'Marker', 'family': 'Marker' },
+		{ 'name': 'Curly', 'family': 'Curly' },
+		{ 'name': 'Pixel', 'family': 'Pixel' },
+		{ 'name': '中文', 'family': '"Microsoft YaHei", "微软雅黑", STXihei, "华文细黑"' },
+		{ 'name': '日本語', 'family': '"ヒラギノ角ゴ Pro W3", "Hiragino Kaku Gothic Pro", Osaka, "メイリオ", Meiryo, "ＭＳ Ｐゴシック", "MS PGothic"' },
+		{ 'name': '한국어', 'family': 'Malgun Gothic' },
+		{ 'name': 'Scratch', 'family': 'Scratch' },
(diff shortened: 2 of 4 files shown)


BroJac5246:
Related:
#2050 (so this is _probably_ rejected)
#3059

addons/more-costume-fonts/userscript.js
```diff
@@ -0,0 +1,63 @@
...
+		{ 'name': 'Monospace', 'family': 'Monospace' }
+	]
+	
+	function setFont(font, update) {
+		paper.tool.changeFont(font.font);
+		document.querySelector("[class*='font-dropdown_mod-unselect']").click();
+		updateImage();
+	}
+	
+	function addNewFont(font, previousFont, modContextMenu) {
```
TheColaber, line 29: Can you just add a check for `if (!addon.self.disabled) return` at the top of this function and then add `dynamicEnable` and `dynamicDisable` in the addon.json, so the addon can be turned on and off dynamically?

addons/more-costume-fonts/userscript.js
```diff
@@ -0,0 +1,63 @@
...
+		const button = document.createElement('span');
+		button.className = addon.tab.scratchClass('button_button') + ' ' + addon.tab.scratchClass('font-dropdown_mod-menu-item');
```
TheColaber, line 31: I think you can just use `addon.tab.scratchClass('button_button', 'font-dropdown_mod-menu-item')`

addons/more-costume-fonts/userscript.js
```diff
@@ -0,0 +1,63 @@
...
+		button.addEventListener('mouseover', function(){ paper.tool.changeFont(font.family); }, false);
+		button.addEventListener('mouseout', function(){ paper.tool.changeFont(previousFont); }, false);
+	}
+
+	while (true) {
+		const elem = await addon.tab.waitForElement("[class*='font-dropdown_mod-context-menu'], [class*='font-dropdown_displayed-font-name']", {
+			markAsSeen: true,
+		});
+		
+		if (elem.classList.contains(addon.tab.scratchClass('font-dropdown_mod-context-menu')) && elem.classList.contains(addon.tab.scratchClass('input-group_input-group'))) {
```
TheColaber, line 47: Won't the first part of this if statement always be true? `font-dropdown_mod-context-menu` is already in the waitForElement, and waitForElement should guarantee that class will exist on the element.

addons/more-costume-fonts/addon.json
```diff
@@ -0,0 +1,40 @@
+{
+	"name": "More costume fonts",
+	"description": "Added new fonts for the costume editor.",
+	"tags": ["editor", "costumeEditor"],
+	"info": [
+		{
+			"type": "notice",
+			"text": "Some fonts may not be compatible with all devices.",
```
TheColaber, line 8: Could you tell us what they will and won't work with and why?

addons/more-costume-fonts/userscript.js
```diff
@@ -0,0 +1,63 @@
...
+			elem.textContent = '';
```
TheColaber, line 48: Is it really necessary to clear the element's previous items and then re-add them in?

addons/more-costume-fonts/userscript.js
```diff
@@ -0,0 +1,63 @@
...
+		{ 'name': 'Pixel', 'family': 'Pixel' },
+		{ 'name': '中文', 'family': '"Microsoft YaHei", "微软雅黑", STXihei, "华文细黑"' },
+		{ 'name': '日本語', 'family': '"ヒラギノ角ゴ Pro W3", "Hiragino Kaku Gothic Pro", Osaka, "メイリオ", Meiryo, "ＭＳ Ｐゴシック", "MS PGothic"' },
+		{ 'name': '한국어', 'family': 'Malgun Gothic' },
+		{ 'name': 'Scratch', 'family': 'Scratch' },
+		{ 'name': 'Arial', 'family': 'Arial' },
+		{ 'name': 'Helvetica', 'family': 'Helvetica' },
+		{ 'name': 'Verdana', 'family': 'Verdana' },
+		{ 'name': 'Impact', 'family': 'Impact' },
+		{ 'name': 'Monospace', 'family': 'Monospace' }
```
TheColaber, line 20: Are there any other fonts that work natively in the scratch editor? How can we find out?

TheColaber (changes requested):

Secret-chest:
@valmontechno @BroJac5246 @TheColaber Remember not all devices are guaranteed to have these fonts (example Linux)!

Samq64:
> Remember not all devices are guaranteed to have these fonts (example Linux)!

Aa a linux user I think it's fine as long as there's a notice like we alreay do for the more keys addon.

Secret-chest:
> > Remember not all devices are guaranteed to have these fonts (example Linux)!
> 
> As a linux user I think it's fine as long as there's a notice like we alreay have for the more keys addon.

I don't want to have to see text that is cut off!

Secret-chest:
Also Scratch will get removed.

mxmou:
This doesn't solve any of the problems previous suggestions for custom fonts had. These fonts (except Scratch) only work if they're installed on the user's device - most of them are only available on Windows and macOS. As @Secret-chest pointed out, a fallback font will be used on other platforms, which sometimes causes text to be cut off or otherwise broken.

Monospace is a generic font name that can refer to any monospace font, which causes similar issues.

WorldLanguages:
This is probably rejected, so I'm closing this PR to avoid any misunderstandings.

PR #5352, opened 2022-11-17
Rename the setting “Maximum number of lines” in Line breaks in comments

DNin01:
Resolves no open issues.

### Changes

Changed “Maximum number of lines” to “Maximum number of lines visible at once”.

### Reason for changes

- To make the addon easier to understand

### Tests

Didn’t test the changes, but there weren’t any code changes.

### DIFF
--- a/addons/comments-linebreaks/addon.json
+++ b/addons/comments-linebreaks/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Line breaks in comments",
-  "description": "Makes line breaks (pressing Enter) in comments on the website visible instead of converting them to spaces.",
+  "description": "Allows adding line breaks (by pressing Enter/Return) in comments you're writing.",
   "credits": [
     {
       "name": "TheColaber",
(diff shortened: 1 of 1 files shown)


mxmou:
I agree with the setting name change, but I think the previous description was more accurate. The addon doesn't make line breaks in comments you write visible to others - it makes line breaks in all comments (including those written by people who don't use the addon) visible to you.

The scrollbar setting is on by default because comments with line breaks visible can be really tall.

DNin01:
> I agree with the setting name change, but I think the previous description was more accurate. The addon doesn't make line breaks in comments you write visible to others - it makes line breaks in all comments (including those written by people who don't use the addon) visible to you.
> 
> 
> 
> The scrollbar setting is on by default because comments with line breaks visible can be really tall.

Okay. I didn't fully understand how the addon worked. I'll change back everything but the setting name.

WorldLanguages (approved):

WorldLanguages:
@mxmou Do you want to suggest any more changes? If not, you can merge this

mxmou:
Adding a notice explaining that people without the addon won't see line breaks would probably be a good idea, but that can be a separate PR.

WorldLanguages:
> Adding a notice explaining that people without the addon won't see line breaks would probably be a good idea, but that can be a separate PR.

See #5421

PR #5362, opened 2022-11-21
Dark mode support for new Scratch Link page

mxmou:
### Changes

Makes the Scratch Link page that was recently implemented (but not yet released) compatible with website dark mode.

### Reason for changes

We don't want to break the page when it becomes public.

### Tests

Tested by building `scratch-www` locally.


apple502j (approved):

WorldLanguages (approved):
I won't build scratch-www to test this, and I haven't checked if the regex is valid

Approving so that PR assignee can merge without admin permissions (PR author is org member so the 2 approval rule is followed)

WorldLanguages:
Note that admins can now be pinged through `@ScratchAddons/admins`

mxmou:
The page is now live: https://scratch.mit.edu/download/scratch-link

PR #5368, opened 2022-11-22
Fix image uploader Material icon

### LINKED ISSUE #5337: Make "Better forum post editor buttons" Addon Also Affect "Forums image uploader" Addon

The `Better forum post editor buttons` addon turns all of the buttons in the forums post editor's toolbar black and makes them look consistent. However, this does not apply to the button added by the `Forums image uploader` addon, which remains colorful. This looks strange and defies the point of `Better forum post editor buttons`.

### What this change does

The `Better forum post editor buttons` will also affect the button added by `Forums image uploader`.

It's interesting to note that `More forum toolbar buttons` _is_ compatible with `Better forum post editor buttons`. I don't know why, but that's worth noting.

### Possible alternatives

We could also change the icon that `Forums image uploader` uses to one that naturally fits in with those that `Better forum post editor buttons` uses.

**Note: In all screenshots, I have `More forum toolbar buttons` enabled as well.**

Normal icons (neither addon being used):
Screenshot 2022-11-08 12 08 07

Just `Better forum post editor buttons` being used:
Screenshot 2022-11-08 12 03 21

Just `Forums image uploader` being used:
Screenshot 2022-11-08 12 05 45

Both addons being used:
Screenshot 2022-11-08 12 06 31
Notice how the icon from `Forums image uploader` stands out.

mxmou:
Resolves #5337

### Changes

Adds the new class name for the image uploader button to the `material-forum-editor-buttons` userstyle. I didn't remove the old one because some people might still be using Jeffalo's extension.

Is the `#ibbBtn` selector still needed? It was apparently added for compatibility with an ImgBB userscript made by easrng. Event if that userscript still exists, it probably doesn't work because the Wikimedia open redirect bug was fixed.

### Reason for changes

To fix a bug.

### Tests

Before:
image
image

After:
image
image

### DIFF
--- a/addons/material-forum-editor-buttons/icons.css
+++ b/addons/material-forum-editor-buttons/icons.css
@@ -68,7 +68,8 @@
   background-image: url(done.svg) !important;
 }
 #ibbBtn,
-#uploadButton {
+#uploadButton,
+.sa-image-upload-button {
   background-image: url(uploader.svg) !important;
 }
 


apple502j (approved):

Hans5958 (approved):
LGTM

PR #5369, opened 2022-11-22
`cat-blocks`: Change info type from warning to notice

DNin01:
### Changes

Previously, the message on the `cat-blocks` addon that the "Watch mouse cursor" option may affect performance was a red warning. I changed it to a yellow notice.

### Reason for changes

The warning message is pretty bright for a minor warning that only applies if you turn on a setting that I don't think everyone will use.

### Tests

Change tested.

### DIFF
--- a/addons/cat-blocks/addon.json
+++ b/addons/cat-blocks/addon.json
@@ -18,7 +18,7 @@
   ],
   "info": [
     {
-      "type": "warning",
+      "type": "notice",
       "text": "The \"Watch mouse cursor\" option may impact performance when the editor is open.",
       "id": "watch"
     }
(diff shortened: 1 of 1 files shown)


RedGuy12 (approved):

WorldLanguages:
Could we change the wording too?

1. It should say "setting", not "option"
2. The "w" in "watch" wouldn't be capitalized there, IMO.

Also, we need a better way to handle "dangerous settings". I think the user should be asked to confirm after trying to enable one of these settings.

DNin01:
> 1. It should say "setting", not "option"
> 2. The "w" in "watch" wouldn't be capitalized there, IMO.

Done.

WorldLanguages (approved):

Issue #4387, opened 2022-03-10
Search addons by highlighted words


RIMOPA:
### Is your request related to a problem? Please describe.

Sometimes it's hard to find addons, and I think some highlited words per addon is the best solution.

### Describe the solution you'd like

Every addon, after its tags, has a array with a few highited words
Example:
` "name": "2D color picker",`
`  "description": "Replaces saturation and brightness sliders in the costume editor with a 2D color picker. Hold Shift while dragging the cursor to change the values on a single axis.",`
`  "tags": ["editor", "costumeEditor"],`
`  "highwords": ["pick", "color", "2D", "select", "saturation", "brightness"]`
`  "credits": [...`

### Additional context

![image](https://user-images.githubusercontent.com/97112832/157722340-9377b0e0-fd62-4cae-a0e0-9463709f77b1.png)

mxmou:
This would probably be difficult to translate.

RedGuy12:
@RIMOPA it would be difficult to translate because there would be no context for any of the strings

A keywords field was also discussed in #2799. Remember the effectiveness of this decreases when extended search is turned on.

I don't think extended search is needed tbh. With it off, I get much more relevant results. We also don't say anywhere that these filters are available.

Chiroyce1:
I agree with @RedGuy12, search is already good enough now. You can find this specific addon by simply searching for `color` and scrolling down.

RedGuy12:
> I agree with @RedGuy12, search is already good enough now. You can find this specific addon by simply searching for `color` and scrolling down.

I was saying extended search, which was enabled in #2799, is not needed. This issue is definitely a good idea. It would be hard to translate but I think it would be worth it.

WorldLanguages:
I think search is good enough as-is.

WorldLanguages:
There has been some discussion about this feature on #5375

PR #5375, opened 2022-11-27
Rename the "more links" addon

### LINKED ISSUE #4882: Rename the "more links" addon
Currently, the more links addon "adds links for URLs pointing outside scratch.mit.edu", but it's a little unclear if you don't know what problem it fixes. The title also doesn't self-explain the addon right now. I have a new name and description:

**Addon name:** Clickable external links
**Description:** Makes URLs pointing outside Scratch from descriptions and comments clickable links.

However, if #4845 were to be merged and/or if [`curator-link`](https://github.com/ScratchAddons/ScratchAddons/tree/66f9072dd7968929a87c30d2ed8b1b430033b689/addons/curator-link) were to be combined with this addon, it might make more sense to call it something else.

DNin01:
Resolves #4882

### Changes

Changed the name of the "More links" addon to "Clickable external links" and edited the description.

### Reason for changes

The new name makes more sense.

### Tests

Changes should work fine.

### DIFF
--- a/addons/more-links/addon.json
+++ b/addons/more-links/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "More links",
-  "description": "Adds links for URLs pointing outside scratch.mit.edu.",
+  "name": "Clickable external links",
+  "description": "Makes URLs pointing outside scratch.mit.edu clickable.",
   "credits": [
     {
       "name": "Hans5958"
--- a/addons/more-links/addon.json
+++ b/addons/more-links/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Clickable external links",
+  "name": "More links",
   "description": "Makes URLs pointing outside scratch.mit.edu clickable.",
   "credits": [
     {


RedGuy12 (approved):

WorldLanguages:
I think @Hans5958 would like to be pinged here

Hans5958:
This is an artifact when names can be abstract and doesn't have to be descriptive. I disagree with this kind of handling (Names can, and should be able to, be abstract like "Scratch Messaging"), so I'm going to neither approve nor disapprove this PR.

mxmou (changes requested):
Searching for "more links" still shows this addon because the ID is `more-links`. That won't work in other languages, so I think the keywords "more" and "link" should be included in the description.

Another option is to leave the old name and only change the description.

DNin01:
Alright, I'll change the name back but keep the new description.

mxmou (approved):

RedGuy12:
> This is an artifact when names can be abstract and doesn't have to be descriptive.

Personally, I like the descriptive names. In any case, most addons have descriptive names so this addon should have one too.



> Searching for "more links" still shows this addon because the ID is `more-links`. That won't work in other languages, so I think the keywords "more" and "link" should be included in the description.

I think that we should add a `keywords` field to addon manifests for search purposes. This has been discussed in the past but never in depth.



> Alright, I'll change the name back but keep the new description.

Just because someone says "an *option* is xyz," does NOT mean you have to immediately change it! It's better to first discuss all the available options and agree on one together before implementing any.

WorldLanguages:
> I think that we should add a `keywords` field to addon manifests for search purposes. This has been discussed in the past but never in depth.

Translating single words, without context, is not an easy task. The "keywords" field is an obvious missing feature until you consider the perspective of translators.
Do we have an issue number for this feature? Or should we create one?

CST1229:
> Translating single words, without context, is not an easy task. The "keywords" field is an obvious missing feature until you consider the perspective of translators.

That's why message descriptions exist.

WorldLanguages:
> That's why message descriptions exist.

Yes, we can add message descriptions now. But this should be discussed in a separate issue, not here.

WorldLanguages (approved):

WorldLanguages:
See #4387 for discussion regarding keywords.

DNin01:
> Searching for "more links" still shows this addon because the ID is `more-links`. That won't work in other languages, so I think the keywords "more" and "link" should be included in the description.

Coming back to this, why is this important?

Although I undid the name change to resolve this problem, I still think it would be a good idea to change the name as I don't think it's as self-explanatory as it should be. I mean, technically, it fits because you have clickable links for more URLs, but there are other things the current name could mean.

WorldLanguages:
> Coming back to this, why is this important?

I think the idea is to avoid the pain of searching for the old addon name, and not getting any results. Users would expect that after enabling an addon, they can search for the addon name as they remember it, right? We sometimes fix this by including the keywords from the old name on the addon description

DNin01:
> > Coming back to this, why is this important?
> 
> I think the idea is to avoid the pain of searching for the old addon name, and not getting any results. Users would expect that after enabling an addon, they can search for the addon name as they remember it, right? We sometimes fix this by including the keywords from the old name on the addon description

It'd still have one of the two original words, though. The only thing that would happen in that case is that it wouldn't be as high in the search results.

mxmou:
> > I think the idea is to avoid the pain of searching for the old addon name, and not getting any results. Users would expect that after enabling an addon, they can search for the addon name as they remember it, right? We sometimes fix this by including the keywords from the old name on the addon description
> 
> It'd still have one of the two original words, though. The only thing that would happen in that case is that it wouldn't be as high in the search results.

Searching for two keywords only shows addons that have both in their title or description.

Secret-chest:
An alternative would be to add a `keywords` JSON field.

DNin01:
> An alternative would be to add a `keywords` JSON field.

Maybe we could start adding keywords for just a few addons in exceptions like these.

BroJac5246:
> An alternative would be to add a `keywords` JSON field.

I've been thinking about that as well.

WorldLanguages:
Yeah, that is #4387

PR #5386, opened 2022-11-30
resizable debugger window

### LINKED ISSUE #5222: What if there was a Re sizable debug window?

When I want to code something and check on the fps at the same time, I can't because the huge screen blocks the code making it difficult to work on my project
Untitled

### What this change does

It would allow the debug window to resize

### Possible alternatives

It could also be a bar at the bottom that you can click on and it will appear on the code

Untitled
Or
Untitled

Platformer444:
Resolves #5222 

### Changes
This makes the debugger window resizable

### Tests
Works fine


addons/debugger/style.css
```diff
@@ -35,6 +35,9 @@
   background-color: white;
   width: 565px;
   height: 25rem;
+  resize: both;
+  min-width: 565px;
+  min-height: 25rem;
```
RedGuy12, line 40: What if you want to make it smaller? I think that's the main want currently, it's quite big

Platformer444: I think I should make this thing a setting and those values customizable

WorldLanguages: It doesn't make sense to customize the minimum dimensions.
The minimum dimensions should be the lowest ones where the buttons still fit and our CSS doesn't break.

Platformer444: So **I** have to manually put that values

BroJac5246: > So **I** have to manually put that values

Correct.

Platformer444: @WorldLanguages Is it a good size (height: 335px and width: 483px)?

image

WorldLanguages: @Platformer444 Personally I'd round the width to 450px and height to 300px

BroJac5246: > @Platformer444 Personally I'd round the width to 450px and height to 300px

Does that work? You're rounding down. (Just making sure before something theoretically breaks.)

Also, why does rounding matter? The minimum dimensions should be the minimum dimensions possible, not some number evenly divisible by 50 just for the sake of being evenly divisible by 50. The difference between what's allowed and what's possible might matter to some people.

WorldLanguages: > You're rounding down.

Sorry, you're right. The min width can't be 450px

Platformer444: @WorldLanguages Is this good now(height: 300px and width: 483px)?
image

mxmou: Does the minimum width work in languages with longer button labels?

Platformer444: > Does the minimum width work in languages with longer button labels?

Oh didn't see that

Samq64: The debugger window doesn't seem to be translated.

mxmou: @Samq64 It is - you need to reload the page after changing the language.
image

WorldLanguages:
Related: I believe the **current** (store release) version of debugger fails to account for longer strings in other languages.

mxmou:
> Related: I believe the **current** (store release) version of debugger fails to account for longer strings in other languages.

The current width (which is also the minimum width in this PR) is wide enough in all languages except Portuguese.
image

The minimum width proposed in the discussion above is not - it overflows in German, for example:
image

A possible solution is to calculate the minimum width in the userscript based on the width of the labels. Alternatively, the tabs could become scrollable if the window is too narrow.

WorldLanguages:
@mxmou I think we should make it scrollable for now

TheColaber:
closing due to inactive author. if anyone wants to implement wl's above comment, feel free to pr.

WorldLanguages:
+1 from feedback, see: https://github.com/ScratchAddons/ScratchAddons/issues/5222#issuecomment-2303266162

PR #5391, opened 2022-12-02
Adding blocks by name improvement

### LINKED ISSUE #5342: Add text translations for icons which are embedded in blocks
In SA right now there are two addons which need to convert blocks into text; dev tools for the middle click menu and the debugger for the thread window. Some blocks, however use icons in their text, namely the `when [green flag] clicked` and `turn [anti/clockwise] () degrees` blocks. Right now both of the addons handle these icons poorly, with the debugger awkwardly treating the icons as inputs and devtools keeping the word 'flag' untranslated or completely omitting the direction indicator for the turn blocks.

_Devtools leaving 'flag' untranslated_
image
_Devtools omitting the clockwise or anticlockwise icons, meaning the two blocks are indistinguishable_
image
_The debugger omitting the flag and doing weird things to the direction arrow_
image

**SA should have translations for these icons that addons can insert in their place when a text representation of blocks is needed.**

I'm not sure where these translations should go, given they need to be used in multiple addons.

Tacodiva:
Current Status: 
- Code quality reviews are helpful (my comments are filled with typos lmao)
- UI improvements are welcome
- Mergeable

TODO List:

- [x] Resolve #5342 
  - [x] Fix this issue in the middle click popup addon
  - [x] Fix this issue in the debugger addon
- [x] Fix 'when backdrop switches to' being black
image
- [x] Fix 'stop other scripts in this sprite' from spawning without being able to stack blocks under it
image
- [x] Fix blocks you can't stack blocks under from rendering wrong
image
- [x] Fix the middle click menu going off the screen
- [x] Clean up `userscript.js`
- [x] Re-add a way click the suggestions using the mouse or direction keys
- [x] Fix pasting rich text / images into the search bar
- [x] Don't suggest block with strings who's contents can be parsed as blocks
- [x] Fix required inputs sometimes being rendered as empty
image
- [x] Don't regress #5512 
- [x] Don't regress #5522
- [x] Make the menu delete blocks dragged into it
- [x] Don't regress #5523
- [x] Fix autocomplete (i know what i mean)
- [x] Make the block renderer render default inputs instead of blank ones
- [x] Fix https://discord.com/channels/751206349614088204/751207339545198682/1066592461536907354
- [x] Differentiate between round and square dropdowns, and allow round blocks in round ones
- [x] Change 'green flag' to 'flag'
- [x] Fix autocomplete deselecting in chromium
- [x] Make pressing enter start a block drag
- [x] Make the menu change size based on search results and fix it being too big on computers with scaling
- [x] Look into autocompleting booleans like they where round blocks
- [x] Fix browser history suggestions sometimes poppuping up
- [x] Add block priority for frequently used blocks
- [x]  Improve the algorithm for discarding string literal tokens
- [x] Fix autocomplete adding score
- [x] Add horizontal end block shape to renderer
- [x] Add quotes for forcing a string
- [x] Fix whatever this is
image
- ~~There seems to be some kind of a problem with the last letter of some tokens being able to be ignored causing weirdness~~ Wont fix (for now)
Query here is `no ` (notice space)
image
image
image
adding a space after `s i t`
image
- [x] Having nothing in the search bar should list all the blocks, not none of them.
- [x] Investigate block height addon setting
- [x] Better handle hitting the token limit
- [x] Remove element IDs and turn them back into classes


WorldLanguages:
Hopefully users can get this by v1.31 (January 2023)

Samq64:
When I try to middle-click I can't stop dragging the workspace.

GarboMuffin:
I am not familiar with how this PR works but you'll want to make sure it doesn't regress https://github.com/ScratchAddons/ScratchAddons/pull/5512

Tacodude7729:
> I am not familiar with how this PR works

me neither lmao

GarboMuffin:
Also check that it hasn't regressed: https://github.com/ScratchAddons/ScratchAddons/pull/5523

Samq64:
Please don't delete the brandh just yet, just in case you want to come back to this one day.

Tacodiva:
Wait no I didn't mean to I just renamed the branch and apparently github can't handle that... whoops guess it's stuck like that forever lmao

addons/middle-click-popup/WorkspaceQuerier.js
```diff
@@ -1095,6 +1103,7 @@ export default class WorkspaceQuerier {
     for (const option of this.tokenGroupBlocks.parseTokens(query, 0)) {
       if (option.end >= queryStr.length) {
         if (option.isLegal) {
+          option.score += WorkspaceQuerier.SCORE_BUMP[option.type.block.id];
```
CST1229, line 1108: won't this make the score of all non-bumped blocks NaN?

Tacodiva: good eye, good eye. I was just testing you :rofl:

Samq64:
Are there any plans to support `custom-block-shape`? If not that's fine.

Also, the way the popup is resized based on browser window width is a bit strange. 
Maximized (1920 pixels wide):
Screenshot from 2023-01-22 10-06-10
Maximized with browser devtools open (1448 pixels wide):
Screenshot from 2023-01-22 10-06-32

Secret-chest:
> If not that's fine.

How is it fine? Especially with this addon, the blocks are already obese enough and I would like the menu to be compact.

Samq64:
Of course it will be done in a follow-up PR, but if we want this to be merged for 1.31 it might be a good idea not to add any more complexity to this PR.

Tacodiva:
> Are there any plans to support custom-block-shape
Unfortunately, this difficult to fix, and essentially involves reprogramming that addon inside of this one. A good compromise solution might be to include a block height setting on this addon or something to that effect.
> Also, the way the popup is resized based on the browser window width is a bit strange.
Yeah I forgot to make the height of the blocks dependent on their scale. I'll fix this now.

mxmou:
> > Are there any plans to support custom-block-shape
> 
> Unfortunately, this difficult to fix, and essentially involves reprogramming that addon inside of this one. A good compromise solution might be to include a block height setting on this addon or something to that effect.

Would it be possible to use Blockly's functions for rendering block SVGs in the popup?

Tacodiva:
> Would it be possible to use Blockly's functions for rendering block SVGs in the popup?

@mxmou 

Probably, but it would probably be a substantial slow down and very painful. I believe in order to make that work I would have to create a Blockly block for every block that needs to be rendered, then somehow modify the Blockly renderer to render the blocks horizontally like they are in my renderer, then actually render the SVGs, then delete the Blockly form of the blocks so it doesnt mess anything up. I decided all that was a lot more work than the 200-something lines it was to just render them myself. If you can make it work with reasonable speed, go ahead, but I really don't want to.

Secret-chest:
Then just don't do it. I really could not use that menu if the blocks were this obese.

Tacodiva:
> Then just don't do it. I really could not use that menu if the blocks were this obese.

@Secret-chest I don't understand, don't do what?

mxmou:
@Secret-chest So you want them to be displayed as text? I don't think that makes sense - it's important to be able to see what the expression will look like.

Secret-chest:
> @Secret-chest So you want them to be displayed as text? I don't think that makes sense - it's important to be able to see what the expression will look like.

Yeah, then just support custom block shape.

Tacodiva:
> > @Secret-chest So you want them to be displayed as text? I don't think that makes sense - it's important to be able to see what the expression will look like.
> 
> Yeah, then just support custom block shape.

WorldLanguages:
We want to release this as part of v1.31.0, in a few days
We should all try our best to find any small enhancements or bugs that we could fix before release.
I sent this to griffpatch so that he can try it out

Samq64:
I tested on both browsers and there weren't any console errors. On Firefox there's a number in the input every time I open it. The setting should also probably have larger minimum values.

Samq64:
The bug I described above is just an issue on my end where middle-clicking pastes the contents of my clipboard and is unrelated to this PR.

Tacodiva:
As for the minimum settings being bigger, I don't really see why. It is very tiny with them all set to 1 but that number is a percentage of the window's width and height so on ultrawide displays having the ability to make it really small could be handy.

WorldLanguages:
Merging. Expect follow-up PRs soon to fix code comment typos, bugs, etc. Thanks tacodiva for the enormous effort put into this.

WorldLanguages:
We should probably change the addon description before release.

WorldLanguages:
My bad, actually versionAdded should probably be kept as 1.30.0 and we instead should add a `latestUpdate` property. It's pretty crazy to think that this PR is technically not a new addon.

DNin01:
> My bad, actually versionAdded should probably be kept as 1.30.0 and we instead should add a `latestUpdate` property. It's pretty crazy to think that this PR is technically not a new addon.

Oh, that's right. This isn't a new addon, just an overhaul.

PR #5392, opened 2022-12-02
Fix Compact User Dropdown

Tacodude7729:
Fixed two issues retron pointed out with the compact user dropdown setting in Customizable navigation bar while I was streaming on Discord.

The compactification not working on some pages, like the 'my suff' page.
image

Extra space after the profile icon on some pages that isn't present on the homepage.
image


retronbv:
ty taco

apple502j (approved):

addons/discuss-button/r2.js
```diff
@@ -5,8 +5,9 @@ export default async function ({ addon }) {
   const add = async () => {
     if (!span && (await addon.auth.fetchIsLoggedIn())) {
       const username = await addon.auth.fetchUsername();
-      const container = document.querySelector(".dropdown");
-      const dropdown = container.querySelector(".dropdown-menu .user-nav");
+      const container = await addon.tab.waitForElement(".dropdown");
+      const dropdown = await addon.tab.waitForElement(".dropdown-menu .user-nav");
+      (await addon.tab.waitForElement(".user-icon")).style = "margin-right: 0px;";
```
Samq64, line 10: Is there a reason for the extra set of parenthacies?

Samq64 (commented):
Dynamically disabling the compact user dropdown removes the margin between the username and the icon.

addons/discuss-button/r2.js
mxmou: I think the browser would read it as
```js
(await addon.tab.waitForElement(".user-icon").style) = "margin-right: 0px;";
```
without them.

WorldLanguages (dismissed):
Bug reported by Samq64 above

Tacodude7729:
@WorldLanguages Resolved the issue above

Samq64 (approved):

PR #5395, opened 2022-12-03
Fix addons running on iframes without src (Firefox)

### LINKED ISSUE #4380: Firefox runs content scripts in about:blank iframes (dark mode bugs)
### Reproduction steps

1. Make sure the `Website dark mode and customizable colors` addon is enabled.
2. Go [here](https://scratch.mit.edu/contact-us/).
3. Scroll all the way down, and click on `Get Help`.
4. Click on `Contact Us` in the pop-up that appears on the right.

### Expected behavior

Textboxes in the pop-up that appears on the right should be consistent.

### Actual behavior

The pop-up that appears on the right should is not dark themed, textboxes are light the text is dark, except for one textbox (the description).

### Screenshots

image

### Environment

Firefox 98, Scratch Addons 1.24.3

WorldLanguages:
Resolves #4380

### Changes

Throws at the top of `cs.js` if being ran on an iframe when `iframe.getAttribute("src") === null`
Other content script files will still run on Firefox, but I don't think that is causing any trouble.

### Tests

Tested on Firefox 106
Checked website dark mode should not apply to the contact us form
Checked addons should still run in iframes (pause+live-featured-project in profile)

### DIFF
--- a/content-scripts/cs.js
+++ b/content-scripts/cs.js
@@ -1,5 +1,6 @@
 try {
   if (window.parent.location.origin !== "https://scratch.mit.edu") throw "Scratch Addons: not first party iframe";
+  if (window.frameElement && window.frameElement.getAttribute("src") === null) throw "Ignored iframe without src attribute";
 } catch {
   throw "Scratch Addons: not first party iframe";
 }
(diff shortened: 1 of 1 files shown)


mxmou (approved):
I think the error message will be "Scratch Addons: not first party iframe", but that isn't really a problem.

apple502j (approved):

WorldLanguages:
> I think the error message will be "Scratch Addons: not first party iframe", but that isn't really a problem.

Good catch, doesn't matter for now

PR #5397, opened 2022-12-04
Remove beta tag from opacity-slider

WorldLanguages:
No serious bugs reported regarding opacity slider, so we can remove the beta tag

### DIFF
--- a/addons/opacity-slider/addon.json
+++ b/addons/opacity-slider/addon.json
@@ -1,7 +1,7 @@
 {
   "name": "Opacity slider",
   "description": "Adds an opacity slider to the costume editor color picker.",
-  "tags": ["editor", "costumeEditor", "beta", "featured"],
+  "tags": ["editor", "costumeEditor", "featured"],
   "info": [
     {
       "type": "info",


apple502j (approved):

mxmou:
It still has some major issues, even if they aren't very serious.

BroJac5246:
> It still has some major issues, even if they aren't very serious.

I agree and think that the beta tag should stay until they're fixed. Otherwise we'll have people wondering why the experience is so inconsistent.

WorldLanguages:
Let's fix some bugs before we merge this

TheColaber:
let's close this since it doesn't seem that any opacity slider bug fixes prs have been made.

WorldLanguages:
> let's close this since it doesn't seem that any opacity slider bug fixes prs have been made.

Why close this? We want to remove the beta tag at some point.

WorldLanguages:
A bug was fixed here: #5625

BroJac5246:
Do we intend for this to ever come out of beta?

WorldLanguages:
Probably not until more bugs are fixed.

PR #5399, opened 2022-12-04
New addon: number-pad

GarboMuffin:
https://github.com/ScratchAddons/ScratchAddons/discussions/5396

Apparently on touchscreen devices, Scratch shows this menu on number inputs:

image

This addon makes it always appear on  all devices.

### DIFF
--- a/addons/number-pad/addon.json
+++ b/addons/number-pad/addon.json
@@ -1,0 +1,19 @@
+{
+  "name": "Always show number pad",
+  "description": "Show Scratch's number pad input when editing number fields on all devices instead of only touchscreen devices.",
+  "credits": [
+    {
+      "name": "GarboMuffin"
+    }
+  ],
+  "tags": ["editor", "codeEditor"],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.28.0"
+}
--- a/addons/number-pad/addon.json
+++ b/addons/number-pad/addon.json
@@ -15,5 +15,5 @@
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "versionAdded": "1.28.0"
+  "versionAdded": "1.31.0"
 }
--- a/addons/number-pad/addon.json
+++ b/addons/number-pad/addon.json
@@ -6,6 +6,12 @@
       "name": "GarboMuffin"
     }
   ],
+  "info": [
+    {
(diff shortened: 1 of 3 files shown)


CST1229:
Does this work for other types of inputs that show a touchscreen number pad, too?

Samq64 (approved):
I tested about 10 blocks. Some nuberpads can input negative values when they shouldn't be able to (e.g. item () of ()) but that's an upstream issue.

Samq64:
> Some nuberpads can input negative values when they shouldn't be able to (e.g. item () of ())

See https://github.com/LLK/scratch-blocks/issues/2916

WorldLanguages:
> Does this work for other types of inputs that show a touchscreen number pad, too?

Such as?

BroJac5246:
Has anyone tested this on a tablet or other device that automatically uses the number pad?

Samq64:
> Has anyone tested this on a tablet or other device that automatically uses the number pad?

I just did and it works as expected. (the addon does nothing)

addons/number-pad/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Always show number pad",
+  "description": "Show Scratch's number pad input when editing number fields on all devices instead of only touchscreen devices.",
```
WorldLanguages, line 3: I'm not sure if users are aware of the number pad. So I don't think it makes sense to use words such as "always".
It may make more sense to say things such as that the number pad input appears below the block, and that the idea is to remove the need to use the keyboard (either physical or virtual)

WorldLanguages: Apparently, laptops with touch screen already support this feature (vanilla Scratch), so this addon would only benefit users that want to use a physical mouse to enter the numbers - and users without a physical keyboard would have no reason to enable this addon.

BroJac5246: > Apparently, laptops with touch screen already support this feature (vanilla Scratch), so this addon would only benefit users that want to use a physical mouse to enter the numbers - and users without a physical keyboard would have no reason to enable this addon.

That's the entire point of the number pad– to show on touchscreen devices.

WorldLanguages (approved):

WorldLanguages:
This can be merged if the addon name and description are OK.
Send a comment if you think it could be enhanced @mxmou @Samq64
I feel this addon will be confusing for some users who have never seen the touch pad.

WorldLanguages:
I added the following note:

image

Samq64:
I think the word block could be added to the addon name.

WorldLanguages:
`Always show block number pads`? (added the word "block", made "pads" plural)

WorldLanguages:
Merging as-is, since there's no consensus over any potential better addon names.

PR #5400, opened 2022-12-04
New addon: Skewing in costume editor

CST1229:
Resolves [**#skewing**](https://discord.com/channels/806602307750985799/1033659563188166716)

It's here, after probably too much time.

### Changes

Adds the Skewing in costume editor (`paint-skew`) addon, which allows skewing shapes in the costume editor by holding Ctrl while resizing a shape on its sides.

### Reason for changes

Sometimes you want to skew a shape in the costume editor, and this addon makes it easy without having to use external image editors.

### Tests

Tested on Chrome a while ago and on Firefox. I don't have a Mac so I can't test the Cmd key, but it probably works.


addons/paint-skew/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ScratchAddons/manifest-schema/dist/schema.json",
+  "name": "Skewing in costume editor",
```
BroJac5246, line 3: Does this work when you're making a backdrop? If so, this should probably be "Skewing in the paint editor".

CST1229: [The docs](https://scratchaddons.com/docs/reference/addon-manifest/#description) say to use "costume editor" instead of "paint editor".

rgantzos:
When you hold down control and click the costume editor, it opens the context menu (as it should). This gets confusing and can cause small imperfections when shaping objects because you have to select a point before pressing control. Maybe it could prevent the context menu from opening when you're holding control and selecting a point?

CST1229:
> When you hold down control and click the costume editor, it opens the context menu (as it should). This gets confusing and can cause small imperfections when shaping objects because you have to select a point before pressing control. Maybe it could prevent the context menu from opening when you're holding control and selecting a point?

I don't get a context menu when Ctrl+clicking.

rgantzos:
> > When you hold down control and click the costume editor, it opens the context menu (as it should). This gets confusing and can cause small imperfections when shaping objects because you have to select a point before pressing control. Maybe it could prevent the context menu from opening when you're holding control and selecting a point?
> 
> I don't get a context menu when Ctrl+clicking.

I think it only happens on Mac because the command key exists (which should probably be used on Mac instead).

CST1229:
> > > When you hold down control and click the costume editor, it opens the context menu (as it should). This gets confusing and can cause small imperfections when shaping objects because you have to select a point before pressing control. Maybe it could prevent the context menu from opening when you're holding control and selecting a point?
> > 
> > 
> > I don't get a context menu when Ctrl+clicking.
> 
> I think it only happens on Mac because the command key exists (which should probably be used on Mac instead).

I thought I added support for that.

addons/paint-skew/addon.json
mxmou: The name "paint editor" comes from 1.4 and isn't used anywhere in the current version of Scratch. We often use "costume" to mean "costume or backdrop", which isn't completely accurate but I don't think it's wrong. Even Scratch does that:
image

BroJac5246: > [The docs](https://scratchaddons.com/docs/reference/addon-manifest/#description) say to use "costume editor" instead of "paint editor".

Oh, thanks

> The name "paint editor" comes from 1.4 and isn't used anywhere in the current version of Scratch. We often use "costume" to mean "costume or backdrop", which isn't completely accurate but I don't think it's wrong. Even Scratch does that:
> image

That's misleading and somewhat confusing, but thanks for pointing that out.

Endermaster3690:
does it work with text

Samq64:
The addon conflicts with `paint-snap`. When that addon is enabled, this one doesn't do anything.

WorldLanguages:
Not going to release any new editor addons this time.

WorldLanguages:
Let's get this merged soon!

addons/paint-skew/addon.json
```diff
@@ -0,0 +1,27 @@
...
+  "description": "Hold Ctrl while scaling an object on its sides to skew it. Hold Alt while skewing an object to skew it from its center.",
+  "info": [
+    {
+      "text": "On macOS, use the Cmd key instead of the Ctrl key.",
```
WorldLanguages, line 7:
```suggestion
      "text": "On macOS, use the Option key instead of the Alt key and Command key instead of the Control key.",
```

addons/paint-skew/addon.json
```diff
@@ -0,0 +1,27 @@
...
+    }
+  ],
+  "tags": ["editor", "costumeEditor"],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.30.0",
```
WorldLanguages, line 24:
```suggestion
  "versionAdded": "1.35.0",
```

addons/paint-skew/addon.json
```diff
@@ -0,0 +1,27 @@
...
+      "id": "useCmdOnMac"
+    }
+  ],
+  "credits": [
+    {
+      "name": "CST1229",
+      "link": "https://github.com/CST1229/"
+    }
+  ],
+  "tags": ["editor", "costumeEditor"],
```
WorldLanguages, line 17:
```suggestion
  "tags": ["editor", "costumeEditor", "featured"],
```

addons/paint-skew/addon.json
CST1229: I'd say those 2 should be swapped:
```suggestion
      "text": "On macOS, use the Command key instead of the Control key and the Option key instead of the Alt key.",
```

addons/paint-skew/addon.json
WorldLanguages: ugh why can't this be set as resolved

addons/paint-snap/updateScaleTool.js
```diff
@@ -1,6 +1,6 @@
 import createScalePoints from "./genScalePoints.js";
-import createSnapPoints from "./genSnapPoints.js";
```
WorldLanguages, line 2: So this import was previously unused?

addons/paint-skew/userscript.js
```diff
@@ -0,0 +1,136 @@
...
+      ogMouseDown.call(this, hitResult, boundsPath, selectedItems);
+      this.isSkew = false;
+      this.skewCenter = false;
+      this.lastSkx = 0;
+      this.lastSky = 0;
+      this.skewBounds = this.itemGroup.bounds.clone();
+    };
+
+    addons.paintSkew = function (event, original) {
+      if (!this.active) return;
```
WorldLanguages, line 30: Technically this line is unnecessary

addons/paint-skew/userscript.js
```diff
@@ -0,0 +1,136 @@
+import { addons, initialize } from "../paint-snap/compatibility.js";
+
+export default async function ({ addon }) {
+  const paper = await addon.tab.traps.getPaper();
+
+  const addSkew = function () {
+    if (!paper.tool || !paper.tool.boundingBoxTool) return;
+    const ST = paper.tool.boundingBoxTool._modeMap.SCALE.constructor;
```
WorldLanguages, line 8: could we clarify (maybe in a code comment) that ST means ScaleTool?

addons/paint-skew/userscript.js
```diff
@@ -0,0 +1,136 @@
...
+
+      this.skewCenter = event.modifiers.alt;
+
+      let skx = 0;
+      let sky = 0;
+      this.lastSkx = 0;
+      this.lastSky = 0;
+      if ((event.modifiers.control || event.modifiers.command) && !this.isCorner) {
+        // Skew
+        if (this.isSkew === false) {
```
WorldLanguages, line 79: Could `this.isSkew` ever evaluate to true? I can't find any code that would set it to anything other than `false`.

addons/paint-skew/userscript.js
```diff
@@ -0,0 +1,136 @@
...
+
+export default async function ({ addon }) {
+  const paper = await addon.tab.traps.getPaper();
+
+  const addSkew = function () {
+    if (!paper.tool || !paper.tool.boundingBoxTool) return;
+    const ST = paper.tool.boundingBoxTool._modeMap.SCALE.constructor;
+    if (!ST) {
+      return;
+    }
```
WorldLanguages, line 11: What would this case mean?
If `paper.tool.boundingBoxTool._modeMap.SCALE` exists, it should have a (truthy) constructor. If it didn't exist, then this line wouldn't be reached and line 8 would have thrown already.

addons/paint-skew/userscript.js
CST1229: Might have been some leftover code from an earlier implementation of this.

addons/paint-snap/updateScaleTool.js
```diff
@@ -200,7 +200,7 @@ export const updateScaleTool = (paper, tool) => {
     axisLineY.visible = false;
   };
 
-  scaleTool.constructor.prototype.onMouseDrag = function (event) {
+  addons.paintSnap = function (event) {
```
WorldLanguages, line 203: What does currently happen if this function is called while `paint-snap` is dynamically disabled? (doesn't look like this PR is changing the behavior, though)

addons/paint-snap/updateScaleTool.js
CST1229: yes

addons/paint-snap/updateScaleTool.js
CST1229: I don't really have an idea, but it doesn't seem to break anything.

addons/paint-skew/userscript.js
```diff
@@ -0,0 +1,131 @@
+import { addons, initialize } from "../paint-snap/compatibility.js";
+
+export default async function ({ addon }) {
+  const paper = await addon.tab.traps.getPaper();
+
+  const addSkew = function () {
+    if (!paper.tool || !paper.tool.boundingBoxTool) return;
+    // ScaleTool
+    const ST = paper.tool.boundingBoxTool._modeMap.SCALE?.constructor;
```
WorldLanguages, line 9:
```suggestion
    const ST = paper.tool.boundingBoxTool._modeMap?.SCALE?.constructor;
```

WorldLanguages (approved):

addons/paint-skew/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Skewing in costume editor",
+  "description": "Hold Ctrl while scaling an object on its sides to skew it. Hold Alt while skewing an object to skew it from its center.",
```
WorldLanguages, line 3: Maybe we could clarify that you need to hold BOTH Ctrl and Alt to skew from the center. Or is it clear enough, what do you think?
cc @DNin01 @BroJac5246

WorldLanguages:
By the way, I have no idea how I'll translate "skew" into Spanish. Maybe some languages need an info notice explaining what it means.

addons/paint-skew/addon.json
DNin01: There are a few ways we could go with this:
- This piece of information could be excluded because you can already hold Alt to resize from the center. If the user knows that, they could theoretically apply that knowledge to the skewing feature.
- Yeah, why not? Doesn't hurt to add a few words to further clarify.
- It does say "while skewing an object", which was explained in the previous sentence. I would find that clear enough within a few seconds.

Honestly, anything works. I can't really decide.

WorldLanguages: Yeah tbh I read the description wrong twice :P 
The first time I tried out this addon I tried skewing from a corner lol

WorldLanguages: Probably more worth it to consider how "skewing" could be translated into other languages. I have no idea how I'll translate it into Spanish tbh.

DNin01:
> By the way, I have no idea how I'll translate "skew" into Spanish. Maybe some languages need an info notice explaining what it means.

Slant?

WorldLanguages:
Thanks! I added this note to Transifex:

image

Clickable link: https://user-images.githubusercontent.com/17484114/266865412-cd1ca51e-87c7-4eb5-8ceb-fd99843c8ff5.gif

Issue #5282, opened 2022-10-28
 Volume slider resets when entering editor


Samq64:
### Description

`vol-sider` resets to the default volume when when opening or leaving the editor.

### Steps to reproduce

1. Enable `vol-slider`
2. Open any project
3. Adjust the volume
4. Click see inside

### Demonstration(s) and additional context

From https://discord.com/channels/751206349614088204/751207339545198682/1035544753615421629

### Environment

Browser: Firefox 106
Scratch Addons version: 1.29.0 stable
Operating system: Ubuntu 22.10

PR #5339, opened 2022-11-09
Fix more vol-slider bugs

Samq64:
Resolves #5282

### Changes

Fixes a bug in the volume slider addon that causes the project volume to reset when entering or leaving the editor. Also fixes a VSCode warning and uses the default volume when dynamically disabling `mute-project` instead of 1.

### Reason for changes

The volume shouldn't reset when opening or leaving the editor.

### Tests

Tested on Chromium 106.

PR #5405, opened 2022-12-05
Refactor and fix more bugs in mute-project and vol-slider

GarboMuffin:
This fixes for the same bugs as https://github.com/ScratchAddons/ScratchAddons/pull/5339 and thus supersedes it:

 - Fixes https://github.com/ScratchAddons/ScratchAddons/issues/5282
 - Uses default volume when mute-project is disabled instead of always 100%
 - Added user-select: none to mute-project icon
 - Adds non-prefixed appearance: none

Although the way these bugs are fixed is not always the same.

And also includes these:

 - When unmuting, it now uses the volume before the project was muted instead of the default volume setting
 - The vol-slider muted indicator will still be visible in small stage mode
 - Various changes to make the addons work in Scratch mods without further patches


addons/vol-slider/module.js
```diff
@@ -1,40 +1,43 @@
...
+
+export const onVolumeChanged = (callback) => {
+  callbacks.push(callback);
 };
 
 export const setup = (_vm) => {
   if (vm) return;
   vm = _vm;
-  vmVol = vm.runtime.audioEngine.inputNode.gain;
+  gainNode = vm.runtime.audioEngine.inputNode.gain;
```
Samq64, line 42: > Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'inputNode')

I'm not sure why this is happening now, but not before.

GarboMuffin: was my bug introduced by this PR, should be fixed now

WorldLanguages:
In the future, we could add a volume indicator SVG that has 2 waves instead of restricting it to 1 and 3, because the icons are being moved to the extension anyway

Samq64 (approved):
Everything works as expected and all the bug fixes mentioned work. You could also add yourself to the addon credits if you want.

PR #5410, opened 2022-12-05
New addon: remove default values in flyout

### LINKED ISSUE #4473: Clear default block input values
Some blocks in the block palette have a standard input value (like "move (>10<) steps". I'm suggesting an addon that replaces these default input values with a blank value.

TheColaber:
Resolves #4473

Suggested on discord :) https://discord.com/channels/806602307750985799/1043986768652927087

### Tests

dynamic enable and disable don't work and idk how to make it work

### DIFF
--- a/addons/remove-default-flyout/userscript.js
+++ b/addons/remove-default-flyout/userscript.js
@@ -2,6 +2,7 @@ export default async function ({ addon, msg, global, console }) {
   const ScratchBlocks = await addon.tab.traps.getBlockly();
   const oldSetValue = ScratchBlocks.Field.prototype.setValue;
   ScratchBlocks.Field.prototype.setValue = function (newValue) {
+    // console.trace();
     if (!addon.self.disabled && this instanceof ScratchBlocks.FieldTextInput && this.sourceBlock_?.isInFlyout) {
       newValue = "";
     }
@@ -15,6 +16,15 @@ export default async function ({ addon, msg, global, console }) {
   addon.self.addEventListener("reenabled", updateToolbox);
 
   function updateToolbox() {
-    // todo, idk how to do this :(
+    const workspace = addon.tab.traps.getWorkspace();
+    if (!workspace) return;
+    const toolbox = Blockly.getMainWorkspace().getToolbox();
+    if (!toolbox) return;
+    const flyout = toolbox.flyout_;
+    if (!flyout) return;
+    console.log("showing and hiding");
+    flyout.hide();
+    console.log(toolbox);
+    // toolbox.showAll_();
   }
 }


DNin01:
I think the name should be "Remove default values in **blocks**" or something like that, because that's closer to what it's doing.

BroJac5246:
> Dynamic enabling and disabling don't seem too important, it's not a big deal if this gets merged without them in my opinion.

Dynamic disable and enable don't make sense, since the user could change the values, so I agree that this shouldn't have them.

mxmou:
Bug:
gif demonstrating a bug

WorldLanguages:
I think this is the second time Colaber submits an addon that doesn't work as expected when you save the project, we need to be careful :P

mxmou:
> I think this is the second time Colaber submits an addon that doesn't work as expected when you save the project, we need to be careful :P

I did a bit more testing and apparently the bug happens when loading a project. All inputs become empty.
image

Here's what that code should look like:
image

WorldLanguages:
@TheColaber Is this a draft?

addons/remove-default-flyout/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Remove default values in flyout",
+  "description": "todo",
```
DNin01, line 3: My proposed name and description:
```suggestion
  "name": "Remove default block values",
  "description": "Removes the default values of text and number inputs in blocks, leaving them blank.",
```

DNin01 (changes requested):
I think the name could be improved.
1. Most people won't know what the "flyout" is.
2. We should use the word "blocks" somewhere.

TheColaber:
> @TheColaber Is this a draft?

@WorldLanguages 
It's ready to be reviewed, I just couldn't figure out dynamic enable and disable. It could still be merged technically.

Joeclinton1:
> > @TheColaber Is this a draft?
> 
> @WorldLanguages It's ready to be reviewed, I just couldn't figure out dynamic enable and disable. It could still be merged technically.

I think you could've figured this out instead of marking for review. Also conflicts need to be fixed.

WorldLanguages:
Dynamic enable/disable is not high priority, and how exactly it should work on this addon is a bit weird.

addons/remove-default-flyout/addon.json
```diff
@@ -0,0 +1,20 @@
...
+      "name": "TheColaber",
+      "link": "https://scratch.mit.edu/users/TheColaber"
+    }
+  ],
+  "tags": ["editor"],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"],
+      "runAtComplete": false
```
Samq64, line 15:
```suggestion
      "matches": ["projects"]
```

addons/remove-default-flyout/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Remove default block values",
+  "description": "Removes the default values of text and number inputs in blocks, leaving them blank.",
+  "credits": [
+    {
+      "name": "TheColaber",
+      "link": "https://scratch.mit.edu/users/TheColaber"
+    }
+  ],
+  "tags": ["editor"],
```
Samq64, line 10:
```suggestion
  "tags": ["editor", "codeEditor"],
```

addons/remove-default-flyout/userscript.js
```diff
@@ -0,0 +1,30 @@
+export default async function ({ addon, msg, global, console }) {
```
Samq64, line 1:
```suggestion
export default async function ({ addon, console }) {
```

Samq64 (commented):
Like WL said making it dynamic isn't a priority, so that code should be removed since it doesn't work.

WorldLanguages:
+1 from feedback

PR #5413, opened 2022-12-06
Make `60fps` disabled by default and `pause` enabled by default

### LINKED ISSUE #5404: `60fps` shouldn’t be recommended
### Description

Right now, the 60-FPS addon is one of the first addons you see, since it’s actually enabled by default. But I don’t feel like that’s the best place for it. Let me explain…

The 60-FPS addon is one of those addons that only work as intended in very specific circumstances (in this case, projects that support 60-FPS). If it isn’t used on the right project, it can actually break things, which is not what people might expect if they had never known about 60-FPS before - addons usually work as expected and predictably everywhere they come into play.

That’s probably okay once or twice for experimental-level addons, except it doesn’t sound like something we would do with an addon that is enabled by default and recommended - that could cause confusion and make it seem like an unnecessary feature.

Question: would you add an addon to resize the project player? Most projects don’t play well with it, so probably not, and if not, why do we have a 60-FPS mode addon that is enabled by default?

For that reason, I think the 60-FPS addon should be lower down on the list of addons, like maybe on the Featured section instead of Recommended, and/or not enabled by default. Might make it harder to discover or learn about and I don’t dislike the feature, but I wouldn’t expect this feature, as it is currently, to be that high on the list.

### Possible alternatives
Remove that addon and have people go to TurboWarp instead. 💣 (😂 Probably won’t happen)

Could also create an “Experimental” label for it.

DNin01:
Resolves #5404

### Changes

- The 60FPS project player mode addon is now disabled by default and was bumped down from the Recommended section to the Featured section.
- The Pause button addon is now enabled by default.

### Reason for changes

Described in #5404

### Tests

Not tested, but I think it will work fine.

### DIFF
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -36,7 +36,7 @@
       "default": 60
     }
   ],
-  "tags": ["editor", "projectPlayer", "recommended"],
+  "tags": ["editor", "projectPlayer", "featured"],
   "versionAdded": "1.1.0",
-  "enabledByDefault": true
+  "enabledByDefault": false
 }
--- a/addons/pause/addon.json
+++ b/addons/pause/addon.json
@@ -25,5 +25,5 @@
     }
   ],
   "tags": ["editor", "projectPlayer", "recommended"],
-  "enabledByDefault": false
+  "enabledByDefault": true
 }


RedGuy12 (approved):

Samq64 (approved):
Tested

FARTMAN9000000Joemama42069:
my school blocked apps like this so i cant download, anything else i can do?

BroJac5246:
> my school blocked apps like this so i cant download, anything else i can do?
> 

This is not the appropriate place to ask; I'd start a discussion. However, the answer is: not without exploits, which I'd advise against.

CST1229:
> my school blocked apps like this so i cant download, anything else i can do?

If they didn't block [TurboWarp](https://turbowarp.org/) (a mirror of it can also be found [here](https://mirror.turbowarp.xyz)), you can use it to get SA's addons as a separate website.

PR #5421, opened 2022-12-07
comments-linebreaks: add notice "linebreaks only visible to you"

WorldLanguages:
As suggested by @mxmou on https://github.com/ScratchAddons/ScratchAddons/pull/5352#issuecomment-1328186882

### DIFF
--- a/addons/comments-linebreaks/addon.json
+++ b/addons/comments-linebreaks/addon.json
@@ -9,7 +9,7 @@
   ],
   "info": [
     {
-      "text": "The line breaks are only visible to you. Line breaks in your own comments will not appear as new lines to everyone.",
+      "text": "Line breaks are only visible to users with this addon enabled. They will appear as spaces for others.",
       "id": "ownLineBreaks"
     }
   ],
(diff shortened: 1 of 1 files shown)


addons/comments-linebreaks/addon.json
```diff
@@ -7,6 +7,12 @@
       "link": "https://scratch.mit.edu/users/TheColaber/"
     }
   ],
+  "info": [
+    {
+      "text": "The line breaks are only visible to you. Line breaks in your own comments will not appear as new lines to everyone.",
```
fractioneater, line 12: I think this could be worded better. It seems to make more sense with the first "the" left out in my opinion.

BroJac5246: I agree.

Also, the current wording incorrectly implies that others with the addon enabled won't see the long

fractioneater: What about something like "Line breaks are only visible to users with this addon enabled. They will appear as spaces for others."

mxmou:
I think this is important enough to be a `notice`, like the one in the folders addon:
image

mxmou (approved):

Samq64 (approved):

PR #5423, opened 2022-12-07
Update `live-featured-project`

ErrorGamer2000:
### Changes

- Shared project changes
  - No longer tries to fall back to alternative player if the project is unshared.
  - No longer has `forkphorus` as an option in favor of TurboWarp
- Added a "Use Scratch username" option when using TurboWarp player.

### Reason for changes

The "Use Scratch username" option was prompted by this feedback:

> only one problem that I know of: username blocks do not work on the featured projects. plz, fix this.

### Tests

Username is correctly set in TurboWarp player.
Migration code changes forkphorus to TurboWarp.

### DIFF
--- a/addons/live-featured-project/addon.json
+++ b/addons/live-featured-project/addon.json
@@ -12,6 +12,19 @@
       "name": "Hans5958",
       "note": "remake",
       "id": "remake"
+    },
+    {
+      "name": "ErrorGamer2000",
+      "link": "https://github.com/ErrorGamer2000",
+      "note": "update",
+      "id": "update"
+    }
+  ],
+  "info": [
+    {
+      "id": "privacy",
+      "type": "warning",
+      "text": "Enabling the \"Use Scratch username\" when using the TurboWarp player will give TurboWarp access to your Scratch username."
     }
   ],
   "userscripts": [
@@ -35,21 +48,17 @@
     },
     {
       "id": "alternativePlayer",
-      "name": "Alternative player",
+      "name": "Project player",
       "type": "select",
       "default": "none",
       "potentialValues": [
         {
           "id": "none",
-          "name": "None"
+          "name": "Scratch"
         },
         {
           "id": "turbowarp",
           "name": "TurboWarp"
-        },
-        {
-          "id": "forkphorus",
-          "name": "forkphorus"
         }
       ]
     },
@@ -63,17 +72,17 @@
       }
     },
     {
-      "id": "forceAlternative",
-      "name": "Force using alternative player",
+      "id": "enableTWAddons",
+      "name": "Enable TurboWarp addons",
       "type": "boolean",
-      "default": false,
+      "default": true,
       "if": {
-        "settings": { "alternativePlayer": ["turbowarp", "forkphorus"] }
+        "settings": { "alternativePlayer": ["turbowarp"] }
       }
     },
     {
-      "id": "enableTWAddons",
-      "name": "Enable TurboWarp addons",
+      "id": "shareUsername",
+      "name": "Use Scratch username",
       "type": "boolean",
       "default": true,
       "if": {
--- a/addons/live-featured-project/addon.json
+++ b/addons/live-featured-project/addon.json
@@ -64,11 +64,11 @@
     },
     {
       "id": "autoPlay",
-      "name": "Auto-play projects (for alternative players)",
+      "name": "Auto-play projects",
       "type": "boolean",
       "default": false,
       "if": {
-        "settings": { "alternativePlayer": ["turbowarp", "forkphorus"] }
+        "settings": { "alternativePlayer": ["turbowarp"] }
       }
     },
     {
--- a/addons/live-featured-project/script.js
+++ b/addons/live-featured-project/script.js
@@ -1,10 +1,12 @@
 export default async function ({ addon, msg }) {
   const showMenu = addon.settings.get("showMenu");
-  const forceAlternative = addon.settings.get("forceAlternative");
-  const alternativePlayer = addon.settings.get("alternativePlayer");
+  const player = addon.settings.get("alternativePlayer");
   const autoPlay = addon.settings.get("autoPlay");
   const enableTWAddons = addon.settings.get("enableTWAddons");
+  const shareUsername = addon.settings.get("shareUsername");
+
   const enabledAddons = await addon.self.getEnabledAddons("editor");
+  const username = await addon.auth.fetchUsername();
 
   const stageElement = document.querySelector(".stage");
   const projectId = window.Scratch.INIT_DATA.PROFILE.featuredProject.id;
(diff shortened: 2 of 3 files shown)


WorldLanguages:
I think I forgot that there's 2 addons that use turbowarp embeds, so we should make sure there's consistency between the 2 regarding how the embed works and what URL params we pass to them

addons/live-featured-project/addon.json
```diff
@@ -12,6 +12,19 @@
       "name": "Hans5958",
       "note": "remake",
       "id": "remake"
+    },
+    {
+      "name": "ErrorGamer2000",
+      "link": "https://github.com/ErrorGamer2000",
+      "note": "update",
```
Samq64, line 19: Could this be more specific or removed?

addons/live-featured-project/addon.json
```diff
@@ -12,6 +12,19 @@
...
+      "id": "update"
+    }
+  ],
+  "info": [
+    {
+      "id": "privacy",
+      "type": "warning",
```
Samq64, line 26: I don't think this needs to be a warning.

addons/live-featured-project/addon.json
```diff
@@ -35,45 +48,41 @@
...
           "id": "none",
-          "name": "None"
+          "name": "Scratch"
         },
         {
           "id": "turbowarp",
           "name": "TurboWarp"
-        },
-        {
-          "id": "forkphorus",
```
Samq64, line 51: I agree with Hans, this shouldn't be removed.

addons/live-featured-project/script.js
```diff
@@ -75,32 +77,20 @@ export default async function ({ addon, msg }) {
...
-        if (alternativePlayer === "turbowarp") loadTurboWarp();
-        else if (alternativePlayer === "forkphorus") loadForkphorus();
-        else stageElement.removeChild(wrapperElement);
-      }
-    });
-  }
+  if (player === "turbowarp") loadTurboWarp();
+  else loadScratch();
+  iframeElement.addEventListener("load", () => {
+    if (iframeElement.contentDocument.querySelector(".not-available-outer") !== null) {
```
Samq64, line 91: Can you put this in a try, catch or find a way that doesn't throw an error when the project is shared?

Samq64 (changes requested):

addons/live-featured-project/addon.json
```diff
@@ -55,25 +68,25 @@
...
-      "id": "forceAlternative",
-      "name": "Force using alternative player",
+      "id": "enableTWAddons",
+      "name": "Enable TurboWarp addons",
       "type": "boolean",
-      "default": false,
+      "default": true,
       "if": {
-        "settings": { "alternativePlayer": ["turbowarp", "forkphorus"] }
+        "settings": { "alternativePlayer": ["turbowarp"] }
```
Samq64, line 84:
```suggestion
        "settings": { "alternativePlayer": "turbowarp" }
```

addons/live-featured-project/addon.json
```diff
@@ -55,25 +68,25 @@
...
       "default": false,
       "if": {
         "settings": { "alternativePlayer": ["turbowarp", "forkphorus"] }
       }
     },
     {
-      "id": "forceAlternative",
-      "name": "Force using alternative player",
+      "id": "enableTWAddons",
+      "name": "Enable TurboWarp addons",
```
Samq64, line 80:
```suggestion
      "name": "Enable addons",
```

background/get-addon-settings.js
```diff
@@ -74,7 +74,9 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
         addonsEnabled["middle-click-popup"] = false;
       }
     }
-
+    if (addonsEnabled["live-featured-project"]?.alternativePlayer === "forkphorus") {
+      addonsEnabled["live-featured-project"].alternativePlayer = "turbowarp";
+    }
```
Samq64, line 79: Since Forkphorus was re-added the migration code can be removed.

addons/live-featured-project/addon.json
```diff
@@ -12,6 +12,19 @@
...
+      "link": "https://github.com/ErrorGamer2000",
+      "note": "update",
+      "id": "update"
+    }
+  ],
+  "info": [
+    {
+      "id": "privacy",
+      "type": "warning",
+      "text": "Enabling the \"Use Scratch username\" when using the TurboWarp player will give TurboWarp access to your Scratch username."
```
DNin01, line 27: The setting for this is already pretty self-explanatory, so does this warning even need to be here?

WorldLanguages: I think we should stay consistent with the TurboWarp player addon. Doesn't make sense to discuss the same things twice

addons/live-featured-project/script.js
```diff
@@ -90,17 +93,13 @@ export default async function ({ addon, msg }) {
...
-        else if (alternativePlayer === "forkphorus") loadForkphorus();
-        else stageElement.removeChild(wrapperElement);
-      }
-    });
-  }
+  if (player === "turbowarp") loadTurboWarp();
+  else if (player === "forkphorus") loadForkphorus();
+  else loadScratch();
+  iframeElement.addEventListener("load", () => {
+    if (iframeElement.contentDocument.querySelector(".not-available-outer") !== null) {
```
apple502j, line 100: I doubt this works with alternative viewers.

apple502j (changes requested):

addons/live-featured-project/script.js
Samq64: Screenshot_20230528_200117

Samq64: I guess I'll make a call to Scratch's API then.

addons/live-featured-project/script.js
```diff
@@ -1,14 +1,19 @@
...
   const enableTWAddons = addon.settings.get("enableTWAddons");
+  const shareUsername = addon.settings.get("shareUsername");
+
   const enabledAddons = await addon.self.getEnabledAddons("editor");
+  const username = await addon.auth.fetchUsername();
 
   const stageElement = document.querySelector(".stage");
   const projectId = window.Scratch.INIT_DATA.PROFILE.featuredProject.id;
 
+  const fetchedProject = await fetch(`https://api.scratch.mit.edu/projects/${projectId}`);
```
Samq64, line 14: I can add proper error handling.

Samq64 (approved):

Samq64:
@ErrorGamer2000 Could you review my changes?

addons/live-featured-project/script.js
```diff
@@ -1,14 +1,20 @@
...
   const autoPlay = addon.settings.get("autoPlay");
   const enableTWAddons = addon.settings.get("enableTWAddons");
-  const enabledAddons = await addon.self.getEnabledAddons("editor");
+  const shareUsername = addon.settings.get("shareUsername");
 
   const stageElement = document.querySelector(".stage");
   const projectId = window.Scratch.INIT_DATA.PROFILE.featuredProject.id;
 
+  // Check if project is unshared before doing anything
+  const featuredProject = await fetch(`https://api.scratch.mit.edu/projects/${projectId}`);
```
Hans5958, line 12: Could you use the `HEAD` method? It should you save some bytes (with negligble speed difference), since you only check for the status.

```suggestion
  const featuredProject = await fetch(`https://api.scratch.mit.edu/projects/${projectId}`, { 'method': 'HEAD' });
```

Hans5958: PS. Make sure to test it! I haven't run the code in whole!

Samq64: Just did on one shared and one unshared project and it seems to work.

ErrorGamer2000:
Yeah, looks good

WorldLanguages:
Looks good!
Just one question though: how did auto-play work all this time due to browser restrictions on soun auto-play? Also, why haven't we supported auto-play in the vanilla player? There's no URL param for it, but nothing prevents us from doing `iframe.contentWindow.document.querySelector("greenflag").click()`, right?

addons/live-featured-project/addon.json
```diff
@@ -55,33 +68,36 @@
...
       "if": {
-        "settings": { "alternativePlayer": ["turbowarp", "forkphorus"] }
+        "settings": { "alternativePlayer": "turbowarp" }
       }
     },
     {
-      "id": "enableTWAddons",
-      "name": "Enable TurboWarp addons",
+      "id": "shareUsername",
+      "name": "Use Scratch username",
```
WorldLanguages, line 89: Interestingly, this setting does not exist in the `TurboWarp button` addon. I wonder if we should add it.

Samq64: I thought it did, but you're right.

Samq64:
> Just one question though: how did auto-play work all this time due to browser restrictions on soun auto-play?

I think the project just starts with no sound by default unless the user specifically disabled auto-play altogether. I'll have to double-check that though.

addons/live-featured-project/addon.json
Samq64: Actually it always uses the Scratch username in replace player mode.

WorldLanguages: Yeah, I wasn't clear enough. I was wondering if we should be providing a way to turn that off. Completely off-topic to this PR, though

PR #5429, opened 2022-12-08
Fix redux devtools conflict

ErrorGamer2000:
### Changes

Shuffled the `inject-redux` script around a little so that it does not break when redux devtools is enabled. If the `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` variable is already set, it replaces the function and instead of using it's own ducktyped compose function uses the old `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` and only edits middleware. If `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` is set *after* sa updates it, same thing happens with the new value.

### Reason for changes

It's annoying not being able to use Redux DevTools to develop addons.

### Tests

No conflicts on my end, but i need some other people to test too.

### DIFF
--- a/content-scripts/load-redux.js
+++ b/content-scripts/load-redux.js
@@ -39,7 +39,8 @@ fix this warning."
     }
   }
 
-  window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ = function (...args) {
+  let newerCompose;
+  function compose(...args) {
     const scratchAddonsRedux = window.__scratchAddonsRedux;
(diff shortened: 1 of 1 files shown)


apple502j (approved):

WorldLanguages (approved):
Redux appears to work fine

ErrorGamer2000:
Yayyy

WorldLanguages:
Update to the FAQ here: https://github.com/ScratchAddons/website-v2/pull/284 (shouldn't be merged yet)

PR #5430, opened 2022-12-08
Better vm trapping

### LINKED ISSUE #4751: Fallback to Redux using internal React components
It's annoying when you can't use redux devtools at the same time with SA.

If there is no redux, fallback to use `document.querySelector("#app")._reactRootContainer._internalRoot.current.child.stateNode.store` on scratch-www pages. And it's real redux, so probably good

Remove Redux Devtools, but why leave problems when it's easy to fix

cc @apple502j you're the god of redux handling and stuff so

ErrorGamer2000:
Resolves #4751

### Changes

Replaced the function prototype hack with redux.
Also made `Trap.js` use a global cache for all addons, no need to be running the code to find the values again for each addon.

### Reason for changes

Prototype tapping is bad


apple502j (approved):

WorldLanguages:
Since we are removing the `__scratchAddonsTraps` global, could we add a different way to access the VM, blockly and Paper through the console? That's usually useful.
I think `window.scratchAddons.vm`, `window.scratchAddons.blockly` and `window.scratchAddons.paper` would be OK.

ErrorGamer2000:
@WorldLanguages I'll get on that. I'll probably use `window.scratchAddons.traps` though.

mxmou:
`scratchAddons.eventTargets.tab[0]` can be used to access the `addon.tab` API from the console. I think that's good enough.

ErrorGamer2000:
Huh, good point. I think WL's suggestion is probably a better option for devs

WorldLanguages:
> `scratchAddons.eventTargets.tab[0]` can be used to access the `addon.tab` API from the console. I think that's good enough.

That's an interesting protip :P

TheColaber (approved):

PR #5446, opened 2022-12-14
Apply dark mode to TurboWarp iframe fullscreen background

### LINKED ISSUE #4430: Full screen in light mode for turbowarp player.
### Reproduction steps

1. Turn on editor dark mode and turbowarp replace player button
2. Go to a project
3. Go into full screen
4. The screen is dark
5. Press the button to replace the scratch player with the turbowarp player
6. Go into full screen
7. The screen is light

### Expected behavior

The turbowarp player should be dark on the outside

### Actual behavior

The turbowarp player was white on the outside

### Screenshots

Scratch player in full screen:
image

Turbowarp player in full screen:
image

### Environment

- Browser: Firefox 98.0.1
- Scratch Addons version: 1.25.0 
- Operating system:  Windows 10

### Additional context

This is a pretty minor problem.

WorldLanguages:
Resolves #4430

### Changes

Code is self explanatory.
We might want to discuss whether we should try to be consistent with the vanilla Scratch fullscreen, which remains unchanged even if `dark-www` is enabled.
Note that users that had system-level dark mode enabled already had dark mode here. This only applies for users that don't have that.

### Tests

Tested in Chromium

### DIFF
--- a/addons/turbowarp-player/userscript.js
+++ b/addons/turbowarp-player/userscript.js
@@ -52,6 +52,13 @@ export default async function ({ addon, console, msg }) {
           const enabledAddons = await addon.self.getEnabledAddons("editor");
           usp.set("addons", enabledAddons.join(","));
         }
+        if ((await addon.self.getEnabledAddons("theme")).includes("dark-www")) {
+          // TurboWarp player fullscreen uses dark mode if dark-www is enabled.
+          // This might be consistent or not with the vanilla Scratch fullscreen,
+          // depending on the moment you're reading this.
+          usp.set("fullscreen-background", "#000000");
+          // Color #000000 is the default TurboWarp embed bg color if system dark mode detected
+        }
         const iframeUrl = `https://turbowarp.org/${projectId}/embed?${usp}${search}`;
         twIframe.src = "";
         scratchStage.parentElement.prepend(twIframeContainer);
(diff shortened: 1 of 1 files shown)


mxmou (changes requested):
Checking if `dark-www` is enabled doesn't make sense because the same addon can also be used to customize the light mode. I think it's better to use the same color as the vanilla full screen - `--editorDarkMode-fullscreen` if it exists, white if it doesn't.

WorldLanguages:
> Checking if `dark-www` is enabled doesn't make sense because the same addon can also be used to customize the light mode

Good catch - I did consider this at first, then I forgot.

WorldLanguages:
> white if it doesn't

@mxmou If it doesn't exist, should we use "white" or let TurboWarp decide?
Right now on master, we always let TurboWarp decide, so if the OS is set to dark mode, TurboWarp will use dark mode. Do you think we should set it to "white" to make sure it's always consistent with the vanilla player fullscreen?

mxmou:
> > white if it doesn't
> 
> @mxmou If it doesn't exist, should we use "white" or let TurboWarp decide? Right now on master, we always let TurboWarp decide, so if the OS is set to dark mode, TurboWarp will use dark mode. Do you think we should set it to "white" to make sure it's always consistent with the vanilla player fullscreen?

I think always using white is better. Nothing else on the Scratch website changes when the system is in dark mode so the TurboWarp embed probably shouldn't either.

mxmou (approved):

PR #5460, opened 2022-12-19
Add `addon.settings.change` api

### LINKED ISSUE #1950: addon.settings.set

ErrorGamer2000:
Resolves #1950
Should be merged along with ScratchAddons/manifest-schema#60

### Changes

Added the `addon.settings.change` api. Usage is:
```js
addon.settings.change({
  settingId: value
})
```
The api will also validate the change before it is made to prevent errors, and only settings that have `settable` set to true in the manifest can be changed with this method.


WorldLanguages:
> only settings that have `settable` set to true

Maybe the name could be more descriptive... `settableFromUserscripts`?

ErrorGamer2000:
> > only settings that have `settable` set to true
> 
> Maybe the name could be more descriptive... `settableFromUserscripts`?

I feel like that's too verbose... `scriptsCanSet`?

WorldLanguages:
@ErrorGamer2000 It's only ever going to affect userscripts, so I think it would be nice to explicitly have the word "userscript".

Maybe just `userscriptsCanSet`?

ErrorGamer2000:
> @ErrorGamer2000 It's only ever going to affect userscripts, so I think it would be nice to explicitly have the word "userscript".
> 
> Maybe just `userscriptsCanSet`?

sure Change 2morrow

Samq64:
Is there a way the API could be simplified to something like this?
```js
addon.settings.change("mySetting", "myValue");
```

Also, maybe `change` could be changed to `set` so it can't be confused with the change event?

WorldLanguages:
Speaking of the change event, we should disable the `addon.settings.change` method while a settings change event is being dispatched. That way, we could avoid a majority of potential cases where the addon author accidentally creates an infinite loop.

mxmou:
> Is there a way the API could be simplified to something like this?
> 
> ```js
> addon.settings.change("mySetting", "myValue");
> ```

The current form is already very simple and it allows multiple settings to be changed at once.

ErrorGamer2000:
> > Is there a way the API could be simplified to something like this?
> > ```js
> > addon.settings.change("mySetting", "myValue");
> > ```
> 
> The current form is already very simple and it allows multiple settings to be changed at once.

Lol You scared me, I thought you were asking for it to change to this lol

ErrorGamer2000:
> Speaking of the change event, we should disable the `addon.settings.change` method while a settings change event is being dispatched. That way, we could avoid a majority of potential cases where the addon author accidentally creates an infinite loop.

I don't think that's an issue. The api does not trigger the event, IIRC

PR #5464, opened 2022-12-20
New Addon: Fullscreen Editor

ErrorGamer2000:
### Changes

Added the Editor fullscreen addon, which adds a fullscreen button to the editor menu bar, and makes some slight style modifications when fullscreen is enabled, as well as optionally shrinking the stage and expanding it on fullscreen exit.

Also changed all addons that add elements to the project bar to use redux listeneres rather than click listeners.

### Tests

All changed addons hide and show as expected, and new addon does not appear to cause any conflicts.

https://user-images.githubusercontent.com/70162741/208743619-e926c3f5-4a9c-4e21-ad41-0449144d5667.mp4


WorldLanguages:
Please add screenshots/video to the PR's first post :)

ErrorGamer2000:
> Please add screenshots/video to the PR's first post :)

My apologies, updated

WorldLanguages:
Will we need a shared space where the "full screen" icon will be inserted? Technically, "messages in editor" already uses it.

ErrorGamer2000:
> Will we need a shared space where the "full screen" icon will be inserted? Technically, "messages in editor" already uses it.

Already added, way ahead of you there @WorldLanguages.

fractioneater:
In the video, it doesn't look like the stage is filling all of the available space when "shrink stage" is enabled.

Also, shrinking the stage doesn't do anything when "hide stage" is enabled, so you might want to make that clear.

ErrorGamer2000:
> In the video, it doesn't look like the stage is filling all of the available space when "shrink stage" is enabled.

That is a result of the sprite search bar, something that needs to be looked at.

> Also, shrinking the stage doesn't do anything when "hide stage" is enabled, so you might want to make that clear.

IDK

addon-api/content-script/Tab.js
```diff
@@ -555,6 +555,11 @@ export default class Tab extends Listenable {
         },
         until: () => [],
       },
+      beforeEditorProfile: {
+        element: () => q("[class^='menu-bar_account-info-group']"),
```
Samq64, line 559: This element only exists when signed in, so the editor fullscreen addon won't work when signed out.

addons/fullscreen-editor/addon.json
```diff
@@ -0,0 +1,42 @@
...
+  ],
+  "userstyles": [
+    {
+      "matches": ["projects"],
+      "url": "userstyle.css"
+    }
+  ],
+  "settings": [
+    {
+      "id": "shrinkStage",
```
Samq64, line 21: Could this setting be hidden when the hide stage setting is enabled?

addons/fullscreen-editor/userscript.js
```diff
@@ -0,0 +1,61 @@
...
+  fullscreen.className = addon.tab.scratchClass("menu-bar_menu-bar-item", "menu-bar_hoverable", {
+    others: "sa-fullscreeneditor",
+  });
+
+  let isFullscreen = false;
+
+  const enterFullscreen = () => {
+    document.body.requestFullscreen();
+    if (addon.settings.get("shrinkStage")) {
+      if (!addon.settings.get("hide")) {
```
Samq64, line 20: It's possible for both settings to be enabled at the same time, even if the one to hide this stage is hidden from the settings page.

addons/fullscreen-editor/userstyle.css
```diff
@@ -0,0 +1,46 @@
...
+
+.sa-fullscreeneditor::before {
+  /* Icon is in a pseudo-element so dark mode can apply a filter to it
+  without affecting the message count */
+  content: "";
+  display: block;
+  position: absolute;
+  width: 1.7rem;
+  height: 100%;
+  background: url(https://scratch.mit.edu/static/assets/422b48c2206d5240a6717c92496ba21a.svg) no-repeat center center;
```
Samq64, line 15: Ideally both icons would be stored in the addon folder.

addons/mouse-pos/stage-size.js
```diff
@@ -0,0 +1,69 @@
+export const stageEventTarget = new EventTarget();
```
Samq64, line 1: Is there a reason this module is in `mouse-pos`?

addons/fullscreen-editor/addon.json
ErrorGamer2000: Not with the way that the settings page is currently set up, I'll work on fixing that.

Samq64: `"if": {"settings": {"hide": false }}`?

addons/mouse-pos/stage-size.js
WorldLanguages: I think it's the first addon that added its own UI to the top of the stage.

WorldLanguages:
Let's move this to #6422

PR #5465, opened 2022-12-20
Adjust badge colors for better text contrast

### LINKED ISSUE #5462: Edit badge colors in settings page to have better text contrast
### Why this would be helpful

Currently, I can feel the reading troubles some people might have when looking at some of the badges (especially _For editor_) in the full screen settings page.
Current badge colors

### My suggestion

My big request is to change the _For editor_ label to be brighter with black text:
The badge colors with updated "For editor" badge
```
.badge.darkgreen {
  background: #0cd1c7;
  color: #000;
  border-color: #12a39e;
}
```
<sub>(I also sneaked in completely black text to replace the very dark gray text)

If that's not enough, we can get even higher contrast by making all the text black to make them all WCAG-compliant.
The badge colors with all badges updated
```
.badge {
  color: #000;
}

.badge.blue {
  background: #56a9ff;
  border-color: #3c88d7;
}
.badge.green {
  background: #78fd2b;
  border-color: #53b31c;
}
.badge.yellow {
  background: #fddd2b;
  border-color: #fdab12;
}
.badge.red {
  background: #ff9031;
  border-color: #cb5e00;
}
.badge.darkred {
  background: #ff7272;
  border-color: #d53b3b;
}
.badge.darkgreen {
  background: #0cd1c7;
  border-color: #12a39e;
}
.badge.purple {
  background: #ff67de;
  border-color: #e137bd;
}
```

And by also adding this:
```
.badge:hover {
    background: #FFF;
    border-color: #DDD;
}
```
...we can make hovering over the badges basically switch them to a black text on white version, helping to make the text easy to read.

### Possible alternatives

- Make the labels darker with white text.
- Add a high contrast mode and make these optional.

### Additional context

Checked with https://contrastchecker.com/

DNin01:
Resolves #5462, closes #5474

### Changes

Before:
Current badge colors

After:
High contrast badge colors

In words, I made the badges brighter with black text.

### Reason for changes

Some badges, such as the "For editor" badge have hard-to-read text due to how similar the text color is to the badge color.

### Tests

Tested.

### DIFF
--- a/webpages/styles/components/badges.css
+++ b/webpages/styles/components/badges.css
@@ -8,40 +8,39 @@
   margin-inline-start: 10px;
   border-bottom: 2px solid #111;
   white-space: nowrap;
+  color: #000;
 }
 .badge .tooltiptext {
   white-space: normal;
 }
 
 .badge.blue {
-  background: #175ef8;
-  border-color: #0e44b8;
+  background: #56a9ff;
+  border-color: #3c88d7;
 }
 .badge.green {
   background: #78fd2b;
-  color: #333;
   border-color: #53b31c;
 }
 .badge.yellow {
   background: #fddd2b;
-  color: #333;
   border-color: #fdab12;
 }
 .badge.red {
-  background: #fd662b;
-  border-color: #d84a12;
+  background: #ff9031;
+  border-color: #cb5e00;
 }
 .badge.darkred {
-  background: #fd2b2b;
-  border-color: #d81212;
+  background: #ff7272;
+  border-color: #d53b3b;
 }
 .badge.darkgreen {
-  background: #00b1a8;
-  border-color: #047773;
+  background: #0cd1c7;
+  border-color: #12a39e;
 }
 .badge.purple {
-  background: rgb(177, 23, 248);
-  border-color: #b20eb8;
+  background: #ff67de;
+  border-color: #e137bd;
 }
 .filter-option.sel.blue {
   background: #175ef8;


BroJac5246:
Two things:

1. I find it harder to read the "Dangerous" tag now. Could the background be made lighter?
2. I also find it difficult to tell the "For website" and "Forums" tags apart by color. A darker background color for the "Forums" tag, like `#2afc46`, would help with this.

Secret-chest:
Do they all need unique colours?

DNin01:
> 2. I also find it difficult to tell the "For website" and "Forums" tags apart by color. A darker background color for the "Forums" tag, like `#2afc46`, would help with this.

The text is always there to help with differentiation. They don't necessarily have to be distinguishable by color, though it probably makes it look nicer. I could adjust the color a bit.

DNin01:
In #5474, I changed only the "For editor" badge color. We'll see which one people like better.

WorldLanguages:
Note that the "for editor" and "for website" tags may be considered unnecessary. I think they only exist for historical reasons.

BroJac5246:
> Note that the "for editor" and "for website" tags may be considered unnecessary. I think they only exist for historical reasons.

Are they going to be removed in the future? Should an issue be created regarding that? They could be helpful in search results, but don't appear to be displayed right now.

WorldLanguages:
As of now, those 2 tags are only used for addons categorized as "themes", and as BroJac said, they aren't shown while searching. They are only visible if the user is filtering by "themes" only.

Secret-chest:
If for editor and for webiste get removed, forums should too.

fractioneater:
I think that this one looks better than #5474, because everything looks more consistent. But what if people don't like it?

I suggest we put an experimental setting into the Scratch Addons options page that changes the badges to either all black text, all white text, or the default, and see which one users like more.

Samq64:
In the PR description it looks good, but I'm not so sure about when it's used on the settings page.

Screenshot_20221223_105916
image

fractioneater:
> In the PR description it looks good, but I'm not so sure about when it's used on the settings page.

What would it look like with all dark text for dark mode, and all light text for light mode?

DNin01:
> What would it look like with all dark text for dark mode, and all light text for light mode?

Here's a comparison:

### Light background
Dark text
Light text

### Dark background
Dark text
Light text

fractioneater:
I think it looks pretty good with the text the same color as the background.

Secret-chest:
Let's discuss what colours should they all have. I don't think they need to be unique.
Dangerous, Beta - red
For website, Forums, For editor - grey
Recommended - orange
New! - purple

fractioneater:
> Recommended - orange (SA theme colour)

Does this mean it should be blue in beta or pre-release versions?

Secret-chest:
> > Recommended - orange (SA theme colour)
> 
> Does this mean it should be blue in beta or pre-release versions?

That's up to be discussed.

BroJac5246:
> > Recommended - orange (SA theme colour)
> 
> Does this mean it should be blue in beta or pre-release versions?

I think that's too much work- it's really about what the users see. Also, if we do that then it might look similar to other tags in pre-release versions.

RedGuy12:
> I think it looks pretty good with the text the same color as the background (either light or dark, that changes with the theme).

I agree with this - that'd be these colors:
image
image

> Does this mean it should be blue in beta or pre-release versions?

If we do use these, no. Only the theme color should change to blue in prereleases. Not UI elements that happen to be orange.  See #4983.


#4983 also suggests not having unique colors for each badge,

PR #5468, opened 2022-12-21
find-bar: Fix some errors searching for broadcasts

GarboMuffin:
These no longer cause issues:

image

 - Broadcasts with complex inputs will be treated as if they are named ~~"(complex)"~~ "(expression)" instead of throwing errors. The reasoning for this behavior is: generating a meaningful text description for complex blocks is not trivial, very few projects use this anyways, but skipping them entirely seems like a bad idea
 - Broadcasts named after Object.prototype builtins will work reliably now.
 - Closes https://github.com/TurboWarp/scratch-gui/issues/669

### DIFF
--- a/addons-l10n/en/find-bar.json
+++ b/addons-l10n/en/find-bar.json
@@ -1,4 +1,5 @@
 {
   "find-bar/find": "Find",
-  "find-bar/find-placeholder": "Find (Ctrl+F)"
+  "find-bar/find-placeholder": "Find (Ctrl+F)",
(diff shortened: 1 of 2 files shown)


mxmou:
> Broadcasts with complex inputs will be treated as if they are named "(complex)" instead of throwing errors.

Something like "(expression)" might be a better name.

GarboMuffin:
Yes, (complex) was not a good name for this.

Samq64 (approved):

WorldLanguages:
Nice job! Appears to work fine

Issue #3315, opened 2021-08-17
Allow setting dynamic enable/disable behavior per userscript/userstyle


mxmou:
### Is your request related to a problem? Please describe.

`dynamicEnable`, `dynamicDisable`, `injectAsStyleElt` (is that for faster loading? This one might also need a better name) and `updateUserstylesOnSettingChange` can't have different values for each userscript and userstyle. This is sometimes a problem: for addons that add new UI, it would be helpful to be able to not remove the CSS when it's disabled to prevent flashes when enabling and disabling. It also means that it isn't possible to make an addon that runs on multiple different pages only support dynamic enable/disable on some of them.

### Describe the solution you'd like

Make them properties of userscript/userstyle objects, not the addon itself.

apple502j:
`injectAsStyleElt` is used by themes that require dynamic modification of style content, AFAIK.

Could be a problem for userscripts: `Addon` object is per-addon and not per-userscript (this is by design) and so is `addon.self`, which means state management is a mess.

RedGuy12:
injectAsStyleElt is also used for addons that need higher specificity then if they were in a file.

WorldLanguages:
> injectAsStyleElt is also used for addons that need higher specificity then if they were in a file.

No. Specificity stays unchanged.
`injectAsStyleElt` is only for faster loading. Some addons don't need faster loading and still use it, tho.
But realistically, it's built for website dark mode.

WorldLanguages:
To avoid the flash of content, I think we should first make sure the style loads, then send the reenabled event to the userscript.

> It also means that it isn't possible to make an addon that runs on multiple different pages only support dynamic enable/disable on some of them.

Any examples of addons that could make use of that?

WorldLanguages:
> To avoid the flash of content, I think we should first make sure the style loads, then send the reenabled event to the userscript.

This gets a bit more tricky if `addon.tab.displayNoneWhileDisabled` is used instead of JavaScript, but it still can be done.

mxmou:
> > It also means that it isn't possible to make an addon that runs on multiple different pages only support dynamic enable/disable on some of them.
> 
> Any examples of addons that could make use of that?

ocular.

WorldLanguages:
Waiting for response by myself

WorldLanguages:
I think the flashing problem deserves its own "type: bug" issue here

WorldLanguages:
[...]

Issue #173, opened 2020-09-02
We will reject addons using innerHTML or triple curly from now on


apple502j:
Please do not use innerHTML. Use appendChild or textContent or innerText. Existing addons will be changed to use these methods. This includes Vue `{{{`.

YOYITsM3M8:
Kkkkkk I always use appendChild bc innerHTML doesn’t work for me sometimes. innerHTML is easier tho

JeffaloBob:
![image](https://user-images.githubusercontent.com/40470736/92022246-4265dc00-ed5b-11ea-8baf-27cfcb114dad.png)

Explosion-Scratch:
This is gonna sound stupid, but… What is innerhtml?

JeffaloBob:
> This is gonna sound stupid, but… What is innerhtml?

in javascript there are many ways to interact with the html page. one of them is directly modifying the HTML which is bad because it opens up for attacks like XSS and stuff.

Explosion-Scratch:
Oh, I see. Thanks! :)

WorldLanguages:
The issue isn't innerHTML itself really (even tho it's dangerous if you don't know what you're doing), but the review process in extension stores. They don't like innerHTML.

Hans5958:
I don't think you can say that all `innerHTML` implementation is bad. Sure that "`innerHTML` is bad because it parses it directly so they can XSS", but I don't think that's the case here.

But, sure, using `textContent` is feasable.

tb148:
Maybe add an "announcement" label for this type of issues?

Hans5958:
Pins are enough, perhaps?

tb148:
> 
> 
> Pins are enough, perhaps?

We only have limited numbers of pins. So accessing older ones is hard without an unique label.

apple502j:
Vue.js `{{{` is now on the denylist. If you really need to use it, please ask me. The only allowed usage for now is for comments (because they are sanitized server side).

tb148:
What is Vue `{{{`?

apple502j:
Works same as `innerHTML`.

YOYITsM3M8:
what about innerText

apple502j:
innerText is safe.

YOYITsM3M8:
K

NoobTracker:
Question: What about `insertAdjacentHTML()`?

apple502j:
[...]

PR #5471, opened 2022-12-21
Add Copy Reporter addon

TheKodeToad:
Doesn't resolve any issues because I kind of got a bit carried away with implementing it. Oops... sorry about that!

I think the description needs improvement. The copy icon is from Unicons, but I don't know where I should put the license since I know these icons are already used. There's also code copied from scratch-blocks.

### Changes

Added the copy reporter addon. This allows you to copy the value of reporters via either the popup balloon thing (I need to find a better way to describe this) or the monitors on the stage.

### Reason for changes

Manually copying values from balloons that keep disappearing is annoying, especially for long values.

### Tests
It adapts to dark theme and seems to work with other addons.
The monitor copier works with normal, large and slider views. It grabs the value from the DOM so lmk if there's a better way.

[Screencast from 2022-12-21 16-03-09.webm](https://user-images.githubusercontent.com/57493648/208949264-5fb90f64-75db-487f-9e16-757334b3e312.webm)


addons/copy-reporter/addon.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "Copy reporters",
```
Samq64, line 2:
```suggestion
  "name": "Copy reporter text",
```
Can someone come up with abetter name than that?

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,84 @@
...
+      copyButton.setAttribute("width", "14px");
+      copyButton.setAttribute("height", "14px");
+      copyButton.setAttribute("viewBox", "0 0 24 24");
+
+      copyButton.style.cursor = "pointer";
+      copyButton.style.float = "right";
+      copyButton.style.display = "block";
+      copyButton.style.userSelect = "none";
+
+      copyButton.innerHTML = `<path fill="currentColor" d="M21 8.94a1.31 1.31 0 0 0-.06-.27v-.09a1.07 1.07 0 0 0-.19-.28l-6-6a1.07 1.07 0 0 0-.28-.19a.32.32 0 0 0-.09 0a.88.88 0 0 0-.33-.11H10a3 3 0 0 0-3 3v1H6a3 3 0 0 0-3 3v10a3 3 0 0 0 3 3h8a3 3 0 0 0 3-3v-1h1a3 3 0 0 0 3-3V8.94Zm-6-3.53L17.59 8H16a1 1 0 0 1-1-1ZM15 19a1 1 0 0 1-1 1H6a1 1 0 0 1-1-1V9a1 1 0 0 1 1-1h1v7a3 3 0 0 0 3 3h5Zm4-4a1 1 0 0 1-1 1h-8a1 1 0 0 1-1-1V5a1 1 0 0 1 1-1h3v3a3 3 0 0 0 3 3h3Z"/>`;
```
Samq64, line 71: Please don't use `innerHTML` see #173. Use a base64 encoded image or a separate SVG file instead.

Samq64 (changes requested):
It would be nice if the addon added `user-select: text;` to the reporter bubbles and variable monitors so the user can can freely copy whatever portion of the text they want.

addons/copy-reporter/userscript.js
WorldLanguages: Theoretically using innerHTML with a hardcoded string is OK, but a separate SVG file is way better.

TheKodeToad: > Please [don't use innerHTML](https://github.com/ScratchAddons/ScratchAddons/issues/173). Use a base64 encoded image or a separate SVG file instead.

Unfortunately I have to use innerHTML, although it could be in a separate file.

addons/copy-reporter/addon.json
TheKodeToad: I think values is best for consistency. Let me know if you don't like this though.

TheKodeToad:
> Could the addon add `user-select: text;` to the reporter bubbles and variable monitors so the user can can freely copy whatever portion of the text they want?

It's already on the bubbles, but it doesn't seem to work well on monitors :(

addons/copy-reporter/userscript.js
Samq64: Can't it just be set here and then changed from the dark mode addon, like other addons do?

mxmou: > If I don't, the icon colour can't be changed to match the light or theme.

It can be changed using a CSS filter.

TheKodeToad: I didn't think of this - thanks!

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    ScratchBlocks.DropDownDiv.hideWithoutAnimation();
+    ScratchBlocks.DropDownDiv.clearContent();
+
+    let contentDiv = ScratchBlocks.DropDownDiv.getContentDiv();
+
+    let valueReportBox = document.createElement("div");
+    valueReportBox.setAttribute("class", "valueReportBox");
+    valueReportBox.innerHTML = ScratchBlocks.scratchBlocksUtils.encodeEntities(value);
```
Samq64, line 46:
```suggestion
    valueReportBox.innerText = ScratchBlocks.scratchBlocksUtils.encodeEntities(value);
```

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+    valueReportBox.style.userSelect = "text";
```
Samq64, line 47: Since the element has a class, could it be styled with CSS?

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+    // use to get focus and event priority
+    valueReportBox.setAttribute("tabindex", "0");
+    // if the user pressed Ctrl+C, prevent propagation to Blockly
+    valueReportBox.onkeydown = (event) => {
+      if ((event.altKey || event.ctrlKey || event.metaKey) && event.keyCode == 67) {
```
Samq64, line 52: Is there a reason for the Alt and Windows key checks?
```suggestion
      if (event.ctrlKey && event.key === "c") {
```

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+        event.stopPropagation();
+      }
+    };
+
+    if (value.length !== 0) {
+      const copyButton = document.createElement("img");
+      copyButton.setAttribute("role", "button");
+      copyButton.setAttribute("tabindex", "0");
+      copyButton.setAttribute("alt", msg("copy-to-clipboard"));
+      // cannot be done in CSS because of a bug
```
Samq64, line 62: Could the SVG itself be set to that size instead then?

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+      copyButton.setAttribute("width", "14px");
+      copyButton.setAttribute("height", "14px");
+      copyButton.setAttribute("src", addon.self.dir + "/copy.svg");
```
Samq64, line 65:
```suggestion
      copyButton.setAttribute("src", "copy.svg");
```

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+
+      copyButton.classList.add("copy-reporter-icon");
+
+      copyButton.onclick = () => navigator.clipboard.writeText(value);
```
Samq64, line 69:
```suggestion
      copyButton.addEventListener("click", () => {
        navigator.clipboard.writeText(value);
      });
```

addons-l10n/en/copy-reporter.json
```diff
@@ -0,0 +1,4 @@
+{
+  "copy-reporter/copy-to-clipboard": "Copy to Clipboard",
```
Samq64, line 2: If all this is used for is an icon alt, I think just the word copy is enough.

Samq64 (changes requested):

addons/copy-reporter/userscript.js
TheKodeToad: This is a freshly-created button. Is this extra code really necessary?

addons/copy-reporter/userscript.js
TheKodeToad: This is based on the Blockly source.

Samq64: `scratch-blocks`? That's kind of a mess and I think [keyCode is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode).

addons/copy-reporter/userscript.js
Samq64: I think `onClick` is fine here, but on existing elements we prefer `addEventListener`.

addons/copy-reporter/userscript.js
mxmou: `event.keyCode == 67` can be replaced with `event.code == "KeyC"`.

It would be easier and more reliable to simply stop propagation of all `keydown` events though.

addons/copy-reporter/userscript.js
mxmou: I don't think this will work. `src` is resolved relative to the Scratch page.

Samq64: That makes sense

addons/copy-reporter/userscript.js
TheKodeToad: It's designed to only prevent the events which scratch-blocks reads as copying a block.

TheKodeToad: I will replace it with event.key === "c" though as that's the same behaviour but not deprecated. I didn't realise.

mxmou: > I will replace it with event.key === "c" though as that's the same behaviour but not deprecated. I didn't realise.

It's not the same. On some keyboard layouts, the C key is in a different position. Its `key` is still `"c"`, but the `keyCode` is different.

TheKodeToad: Does "KeyC" work like before?

mxmou: > Does "KeyC" work like before?

Yes, unless there are weird edge cases where it behaves differently.

TheKodeToad: On my browser e.key seems to be lowercase C though which is weird.

mxmou: `event.key` is the character ("c"). `event.code` is the name of the key ("KeyC").

TheKodeToad: I got them confused. I can fix this (probably)!

addons/copy-reporter/userscript.js
TheKodeToad: That would invole changing that actual path.

Samq64: No, I mean opening the SVG in an editor and scaling jt to 14x14. It shouldn't effect tge quality because it's a vector.

mxmou: It's usually a good idea to set width and height explicitly, using either CSS or `width` and `height` attributes.

What's the bug that makes it impossible to do this using CSS here?

Samq64: I tried setting the width and height with CSS just now and didn't have any issues. It was probably a caching issue.

TheKodeToad: I'm using Firefox so it's possible something to do with that.

TheKodeToad: Maybe it was a Firefox problem?

mxmou: Were the styles just not getting applied?

TheKodeToad: The attribute was populated :shrug:.

mxmou: I'm not sure what could cause a bug like that. The following worked for me on both Edge and Firefox:
```js
copyButton.style.width = "14px";
copyButton.style.height = "14px";
```

Adding these to the userstyle also worked:
```css
width: 14px;
height: 14px;
```

Samq64: Try setting the styles again and shift-click the reload page button to ignore the browser cache.

Samq64 (approved):

addons/copy-reporter/userscript.js
TheKodeToad: I don't think the cache applies to attributes though. I'll still try.

TheKodeToad: Hmm... it does seem to work, but it's not centred any more. I think something else caused that though.

TheKodeToad:
Idk what happened because the copy button was centred before

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,77 @@
...
+      let element = ctx.target.lastChild.lastChild;
+
+      if (ctx.type === "monitor_slider") {
+        element = element.children[element.children.length - 2].lastChild;
+      } else {
+        element = element.lastChild;
+        if (ctx.type === "monitor_default") {
+          element = element.lastChild;
+        }
+      }
```
mxmou, line 14:
```suggestion
      const className = ctx.type === "monitor_large" ? "monitor_large-value" : "monitor_value";
      const element = ctx.target.querySelector(`[class*='${className}_']`);
```

addons/copy-reporter/userscript.js
```diff
@@ -0,0 +1,77 @@
...
+    // use to get focus and event priority
+    valueReportBox.setAttribute("tabindex", "0");
+    // if the user pressed Ctrl+C, prevent propagation to Blockly
+    valueReportBox.onkeydown = (event) => {
+      if ((event.altKey || event.ctrlKey || event.metaKey) && event.code === "KeyC") {
+        event.stopPropagation();
+      }
+    };
+
+    if (value.length !== 0) {
```
mxmou, line 56:
```suggestion
    if (value.length !== 0 && !addon.self.disabled) {
```
You can add this here and `"dynamicDisable": true` to `addon.json` to allow the addon to be disabled without reloading the page.

TheKodeToad: I can't find a way to remove the context menu entry.

addons/copy-reporter/userscript.js
TheKodeToad: What's wrong with `.className`?

addons/copy-reporter/userscript.js
mxmou: I think that's done automatically by the API.

TheKodeToad: The other context menu plugin didn't seem to but maybe dynamicDisable isn't set right.

mxmou: "Move costume to top or bottom" supports dynamic disable and it doesn't have any extra code for it.

DNin01 (commented):
Might want to add Dynamic Disable support, as most addons support both Dynamic Enable and Disable. You can read the [documentation](https://scratchaddons.com/docs/reference/addon-api/addon.self/#using-addonselfdisabled-and-events-dynamicdisable) for more information. For this addon, though, at least being able to dynamically enable is more important.

TheKodeToad:
> Might want to add Dynamic Disable support, as most addons support both Dynamic Enable and Disable. You can read the [documentation](https://scratchaddons.com/docs/reference/addon-api/addon.self/#using-addonselfdisabled-and-events-dynamicdisable) for more information. For this addon, though, at least being able to dynamically enable is more important.

mxmou also had the same point but I never got around to testing! Seems this is possible after all...

DNin01 (approved):
Seems to work fine!

Dynamic enable and disable are working well too, other than the copy button getting really big if you disable it while a reporter balloon is being displayed, but that's OK.

addons/copy-reporter/addon.json
```diff
@@ -0,0 +1,20 @@
...
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["projects"]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true
```
DNin01, line 19: Oh, and to top it off, we need a `versionAdded` attribute here. This will be...
```suggestion
  "dynamicDisable": true,
  "versionAdded": "1.31.0"
```

WorldLanguages:
> the copy button getting really big if you disable it while a reporter balloon is being displayed

Isn't this easy to fix?

mxmou:
> > the copy button getting really big if you disable it while a reporter balloon is being displayed
> 
> Isn't this easy to fix?

`displayNoneWhileDisabled` on the button should fix it. An even better fix would be to keep the userstyle loaded while the addon is disabled, but that isn't currently possible (see #3315).

WorldLanguages:
> see https://github.com/ScratchAddons/ScratchAddons/issues/3315

Thanks for the reminder!

WorldLanguages:
If interested, anyone reading open a PR to clarify the description field of the addon. I think some users might not be able to find the "copy value" feature because you need to right click.

MaterArc:
> If interested, anyone reading open a PR to clarify the description field of the addon. I think some users might not be able to find the "copy value" feature because you need to right click.

I can do this!

TheKodeToad:
:P i'm just glad my feature has made it! it does need improvement still, yes

Issue #5443, opened 2022-12-14
sprite-properties addon: creating new project breaks "i" icon


WorldLanguages:
### Description

> There's a bug (setting "COLLAPSE PANEL BY DEFAULT" enabled):
Go to scratch.mit.edu/create and wait for the project to load. The "i" icon will appear when the panel is expanded, instead of when it's collapsed, which is the opposite of what normally happens.

### Steps to reproduce

See above
Apparently, you need to be logged-in to Scratch to reproduce.

### Environment

Chrome

WorldLanguages:
We should try to fix this before release, otherwise users will say they can't access the sprite properties panel!

PR #5472, opened 2022-12-21
Move the icon for closing sprite properties

mxmou:
See https://github.com/ScratchAddons/ScratchAddons/pull/5158#issuecomment-1345303499

### Changes

The "collapsing sprite properties" addon adds an arrow to close the properties panel when it's expanded. It looks like this:
image

This PR changes it to look like this instead:
image

A large area around it is clickable.

### Reason for changes

To make it more visible.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/sprite-properties/userscript.js
+++ b/addons/sprite-properties/userscript.js
@@ -81,8 +81,7 @@ export default async function ({ addon, global, console, msg }) {
   }
 
   async function injectCloseButton() {
-    let container = propertiesPanel.querySelector("[class*='sprite-info_row_']:nth-child(2)");
-    injectButton(container, PROPS_CLOSE_BTN_CLASS, "/collapse.svg", msg("close-properties-panel-tooltip"));
+    injectButton(propertiesPanel, PROPS_CLOSE_BTN_CLASS, "/collapse.svg", msg("close-properties-panel-tooltip"));
   }
 
   async function injectButton(container, className, iconPath, tooltip) {
--- a/addons/sprite-properties/userstyle.css
+++ b/addons/sprite-properties/userstyle.css
@@ -17,12 +17,22 @@
 }
 
 .sa-show-sprite-properties [class^="sprite-info_sprite-info_"] {
-  height: 6.3125rem;
+  /* The height needs to be known for the animation to work.
+     6.5rem + 5px is the exact height of sprite info when
+     this addon is enabled:
+       0.75rem (top padding)
+     + 2rem + 2px (.icon-wrappers in the first row)
+     + 0.5rem (margin between rows)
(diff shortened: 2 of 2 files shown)


WorldLanguages:
I like the new position. The old one was very easy to miss.

The only potential problem is that the sprite properties panel needs to be taller than normal. I wonder if that could cause problems in the future, or possibly existing CSS problems which we are missing.

WorldLanguages (approved):
Tested successfully in Chromium

WorldLanguages:
We should also try to fix #5443 before release.

lisa-wolfgang:
Now that I think about it again, a dropdown arrow by itself isn't the most clearly visible solution. We could take inspiration from Scratch 2.0: the button could be colored and include text explaining what it does. We could also take up the full vertical space when the properties panel is expanded if needed.

BroJac5246:
> The only potential problem is that the sprite properties panel needs to be taller than normal. I wonder if that could cause problems in the future

I was also concerned this might not be great for small screens, but I tested it and it's fine in that manner, if anyone was wondering.

BroJac5246:
> Now that I think about it again, a dropdown arrow by itself isn't the most clearly visible solution. We could take inspiration from Scratch 2.0: the button could be colored and include text explaining what it does. We could also take up the full vertical space when the properties panel is expanded if needed.

I had noticed that as well. Could the info button stay on the sprite tile at all times and be able to be used to collapse the sprite properties panel as an optional alternative?

WorldLanguages:
> Now that I think about it again, a dropdown arrow by itself isn't the most clearly visible solution. We could take inspiration from Scratch 2.0: the button could be colored and include text explaining what it does. We could also take up the full vertical space when the properties panel is expanded if needed.

Could you share a Scratch 2.0 screenshot/video if possible? I don't remember how this used to work back then.

mxmou:
> > Now that I think about it again, a dropdown arrow by itself isn't the most clearly visible solution. We could take inspiration from Scratch 2.0: the button could be colored and include text explaining what it does. We could also take up the full vertical space when the properties panel is expanded if needed.
> 
> Could you share a Scratch 2.0 screenshot/video if possible? I don't remember how this used to work back then.

Scratch 2 had a back arrow in the top left corner:
image

The info panel covered the sprite list entirely.

WorldLanguages:
Interesting.

> We could also take up the full vertical space when the properties panel is expanded if needed.

I'm not sure if that would be a great default, but it could be a useful setting for some. Would be part of the "Scratch 2.0 superpreset" I guess :P

Samq64 (approved):

WorldLanguages:
If anyone's interested on the 2.0 model where the sprite info covers the whole area, please open an issue or PR!

PR #5473, opened 2022-12-21
sprite-properties: Fix some bugs

GarboMuffin:
- Prevent show and hide both being simultaneously active. Should fix https://github.com/ScratchAddons/ScratchAddons/issues/5443. This at least prevents the addon breaking after going into and leaving fullscreen.
 - Fix double click and auto hiding after leaving and going back into the editor
 - Fix errors when the editing target is a clone
 - Should not conflict with https://github.com/ScratchAddons/ScratchAddons/pull/5472


Samq64:
```
userscript.js:37 Uncaught TypeError: e.target.matches is not a function
    at document.addEventListener.capture (userscript.js:37:20)
```

GarboMuffin:
Should be fixed now. Was caused by Firefox and Chrome treating `document.addEventListner("mousemove", () => {})` differently (Chrome emits event for `document` itself, Firefox doesn't)

apple502j (approved):

Samq64 (approved):

addons/sprite-properties/userscript.js
```diff
@@ -17,20 +16,36 @@ export default async function ({ addon, global, console, msg }) {
...
+      togglePropertiesPanel();
+    }
   });
+
   // Close properties panel when mouse leaves the entire sprite panel
-  spriteContainer.addEventListener("mouseleave", () => autoHidePanel());
+  document.body.addEventListener(
+    "mouseleave",
+    (e) => {
+      if (e.target.matches('[class*="sprite-selector_sprite-selector_2KgCX"]')) {
```
WorldLanguages, line 37: Was this a mistake? Why use wildcard if we're going to specify the hash (2KgCX)?

GarboMuffin: yeah

PR #5479, opened 2022-12-23
Add editor-random-direction-block addon

Tacodiva:
Superseeds #5153

The author of that PR seems to have abandoned it and was a bit on the wrong track (in my opinion) so I though I'd whip up this solution quickly, because the idea for the addon is pretty good.

This is mostly copypasta from the extra-key-support addon, so should be pretty bug-free.


Tacodude7729:
I wrote 'should be pretty bug-free' than went to the toilet and instantly realized there was a major bug lmao whoop.

addons/editor-random-direction-block/userscript.js
```diff
@@ -0,0 +1,57 @@
...
+      const allBlocks = [...workspace.getAllBlocks(), ...flyout.getWorkspace().getAllBlocks()];
+      for (const block of allBlocks) {
+        if (block.type !== "motion_pointtowards_menu") {
+          continue;
+        }
+        const input = block.inputList[0];
+        if (!input) {
+          continue;
+        }
+        console.log(input.fieldRow);
```
GarboMuffin, line 39:
```suggestion
```

Tacodude7729: Oops, thanks :)

BroJac5246:
The current name implies that it's a new block. I'm thinking something closer to `Myself option in "Point towards ()" block` (that's just an idea, someone will probably come up with something better).

WorldLanguages:
I think the consensus is to use the "beta" tag for addons like this. For example, editor-extra-keys has the beta tag.

Samq64:
Clicking the block with the random direction option selected doesn't row any errors, but it doesn't seem to do anything either.

BroJac5246:
> I think the consensus is to use the "beta" tag for addons like this. For example, editor-extra-keys has the beta tag.

Why? The feature works universally, so is that really necessary?

WorldLanguages:
> Clicking the block with the random direction option selected doesn't throw any errors, but it doesn't seem to do anything either.

But does the block work properly when running scripts normally?

WorldLanguages:
> Why? The feature works universally, so is that really necessary?

@BroJac5246 I guess you could achieve the same result with just the backpack, as long as someone edits the project JSON for you.

Samq64:
> > Clicking the block with the random direction option selected doesn't throw any errors, but it doesn't seem to do anything either.
> 
> But does the block work properly when running scripts normally?

Weird, it's working now... I'll see if I can reproduce it.

WorldLanguages:
PR author is busy with another PR on the same milestone. So postponing to the next one.

Samq64 (approved):
> I can't see how the error reported by samq64 could happen and because there's no repro I suspect some kind of user-error.

I can't repro myself any more, so you're probably right. Everything looks good. The only change I can think of is adding dynamic enable/disable to existing block labels, but I don't think it's worth it if it means reloading the whole workspace.

WorldLanguages:
> The only change I can think of is adding dynamic enable/disable to existing block labels

We should check if related addon editor-extra-keys supports that

WorldLanguages:
> I think the consensus is to use the "beta" tag for addons like this. For example, editor-extra-keys has the beta tag.

Bump

BroJac5246:
> > I think the consensus is to use the "beta" tag for addons like this. For example, editor-extra-keys has the beta tag.
> 
> Bump

I don't think we should because it's not actually in beta. If we really want then we could create a tag that shows it adds features that will impact (not in a negative sense) non-SA users.

WorldLanguages:
@BroJac5246 It wouldn't be worth it to create a dedicated tag for these cases. Keep in mind that the beta addon group also includes addons tagged as "dangerous" - it's not just addons that might have bugs.

BroJac5246:
> @BroJac5246 It wouldn't be worth it to create a dedicated tag for these cases. Keep in mind that the beta addon group also includes addons tagged as "dangerous" - it's not just addons that might have bugs.

I agree that it might not be worth (and probably isn't worth) a dedicated tag. However, I don't think "beta" should be all-encompassing for other miscellaneous categories as well.  To a user, it implies that the features are actually in beta, so it's also a grammatical issue. While a dedicated tag might not be required, I don't think that, from a language perspective, "beta" is being used correctly. (Source on language interpretations: I'm a native American English speaker)

Tacodiva:
I was under the impression the beta tag for the extra keys addon was due to the inclusion of the symbols, and it was unclear if they would be supported on all operating systems. What exactly are you qlwanting to convey with these tags? I don't see anything inherently 'risky' about addons like this.

WorldLanguages:
> I was under the impression the beta tag for the extra keys addon was due to the inclusion of the symbols, and it was unclear if they would be supported on all operating systems.

Good point, I didn't remember the details very well.

BroJac5246:
> I was under the impression the beta tag for the extra keys addon was due to the inclusion of the symbols, and it was unclear if they would be supported on all operating systems. What exactly are you wanting to convey with these tags? I don't see anything inherently 'risky' about addons like this.

Looking at the old PR, this appears to be the case. @WorldLanguages added it in [this commit](https://github.com/ScratchAddons/ScratchAddons/pull/3299/commits/4cd191112369b28b5e6a07b274cec87b4acac974) (with no prior discussion for us to gain information from), at the same time giving [this comment](https://github.com/ScratchAddons/ScratchAddons/pull/3299#issuecomment-917447568) about how he couldn't test some keys. However, he then proceeded to remove support for most keys except for those that are universal. Is it possible the beta tag was just never removed, though? I think this shows that we shouldn't go off of the extra keys addon here, however, because the tag was added because of compatibility issues that are not present here.

addons/editor-random-direction-block/addon.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "name": "Tacodude7729",
+      "link": "https://scratch.mit.edu/users/Tacodude7729/"
+    },
+    {
+      "name": "pufferfish101007",
+      "link": "https://scratch.mit.edu/users/pufferfish101007/"
+    },
+    {
+      "name": "winwinlmfao"
```
WorldLanguages, line 25: Unfortunate name? :P

addons/editor-random-direction-block/addon.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Point towards random direction block",
+  "description": "Adds a \"random position\" option to the \"point towards ()\" block that makes the sprite point in a random direction. This will work for everyone using the project.",
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.31.0",
```
WorldLanguages, line 10:
```suggestion
  "versionAdded": "1.35.0",
```

BroJac5246:
@Tacodiva any progress?

Tacodiva:
@BroJac5246 Pretty much good to go, just waiting on a reply from 'winwinlmfao' to see they want to be credited.

BroJac5246:
> @BroJac5246 Pretty much good to go, just waiting on a reply from 'winwinlmfao' to see how they want to be credited.

@emir4169 how do you want to be credited?


...
If they don't respond then we should probably just pick one (I'm thinking emir4169) and add a link.

WorldLanguages:
We'll wait 48 hours, if no response then we'll credit them as their current github username.

pufferfish101007:
they are also `winwinlmfao` on discord - pinging them there (maybe in https://discord.com/channels/806602307750985799/1025716292822511676) might get a response

Samq64:
https://github.com/ScratchAddons/ScratchAddons/pull/5153#issuecomment-1726945104

addons/editor-random-direction-block/addon.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "name": "Tacodiva",
+      "link": "https://github.com/Tacodiva/"
+    },
+    {
+      "name": "pufferfish101007",
+      "link": "https://scratch.mit.edu/users/pufferfish101007/"
+    },
+    {
+      "name": "winwinlmfao"
```
WorldLanguages, line 25:
```suggestion
      "name": "emir4169"
```

WorldLanguages (approved):

addons/editor-random-direction-block/addon.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Point towards random direction block",
+  "description": "Adds a \"random position\" option to the \"point towards ()\" block that makes the sprite point in a random direction. This will work for everyone using the project.",
```
WorldLanguages, line 3:
```suggestion
  "description": "Adds a \"random direction\" option to the \"point towards ()\" block that makes the sprite point in a random direction. This will work for everyone using the project.",
```

DNin01 (commented):
I don't think I understand the point of this addon if you can just insert a `pick random (1) to (360)` block into `point in direction ()`? The only thing is this saves a few seconds, but is the time save and the fact that we use an otherwise inaccessible block worth it? I don't find myself pointing sprites in random directions that much per project, but I'm not sure how much another user might possibly do that sort of thing - maybe it is useful to someone.

PR #5482, opened 2022-12-23
Add info note to disable-paste-offset

WorldLanguages:
From feedback:

> just figured out there's a SA feature allowing to duplicate an object at the same place on costume editor but we could already do this by holding option/alt and click on the item

### Changes

Add an info note to the addon to make users aware of this. The phrasing can be changed and probably should.

### DIFF
--- a/addons/disable-paste-offset/addon.json
+++ b/addons/disable-paste-offset/addon.json
@@ -1,6 +1,12 @@
 {
   "name": "Do not shift pasted items",
   "description": "Paste copied items at their original position instead of shifted slightly in the costume editor.",
+  "info": [
+    {
+      "text": "This behavior can also be achieved without needing this addon, by Alt+Clicking the item.",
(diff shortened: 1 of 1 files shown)


addons/disable-paste-offset/addon.json
```diff
@@ -1,6 +1,12 @@
 {
   "name": "Do not shift pasted items",
   "description": "Paste copied items at their original position instead of shifted slightly in the costume editor.",
+  "info": [
+    {
+      "text": "This behavior can also be achieved without needing this addon, by Alt+Clicking the item.",
```
Samq64, line 6:
```suggestion
      "text": "This behavior can also be achieved without this addon by Alt+Clicking the item.",
```

Samq64: Current one is fine too.

TheColaber (approved):

Issue #5486, opened 2022-12-25
Volume Slider disables Clone Counter


TJ20201:
### Description

Using both the Volume Slider addon and the Clone Counter addon results in unwanted behaviour of the Volume Slider overriding the Clone Counter instead of being drawn alongside it. 

### Steps to reproduce

1. Enable "Volume Slider" and "Clone Counter" addons. (Pastebin to my settings file: [Click Here](https://pastebin.com/xQ7aRdcg))
2. Go to any project's view. Any at all, as VS and CC apply to every project.
3. Check if the Volume Slider and Clone Counter both appear, if not, you have reproduced the bug.

### Demonstration(s) and additional context

Due to my snipping tool shortcut not working, a Pastebin link has been supplied in the reproduction steps with my exported SA Settings file.

### Environment

Browser: Chrome 108.0.5359.125 (chrome://version/)
Scratch Addons version: v1.29.0
Operating system: Windows 10 Home (10.0.19045 x64)

TJ20201:
I have managed to get my snipping tool keybind to work, and this is the appearance of the Stage of my own project.
![image](https://user-images.githubusercontent.com/86313252/209474191-a0647de8-8c20-494a-b097-3fc28f61f06a.png)

DNin01:
Are you in the editor or on the project page? (The clone counter only appears inside the editor.)

TJ20201:
Ah, I was on the project. However, I just tried on the Editor view on the latest source (with an extra addon from my Fork/custom-brand-image) and it appears as this.
![image](https://user-images.githubusercontent.com/86313252/209479928-f500e945-3571-4fb5-bc6b-6495d58973db.png)

WorldLanguages:
Are you aware about the fact the clone counter isn't shown unless the project has at least 1 clone at the moment?

TJ20201:
> Are you aware about the fact the clone counter isn't shown unless the project has at least 1 clone at the moment?

Yes, it turns out I clicked on the wrong project (my Slideshow API one) instead of my Ping Pong one (the one I originally opened this issue with the attachments from). Maybe a boolean setting could be added to the addon that allows it to show on the non-Editor view? (Just a suggestion.)

DNin01:
> > Are you aware about the fact the clone counter isn't shown unless the project has at least 1 clone at the moment?
> 
[...]

PR #5488, opened 2022-12-25
`clones`: Setting to show on project page

DNin01:
Inspired by https://github.com/ScratchAddons/ScratchAddons/issues/5486#issuecomment-1364735331

### Changes

Added an option to change whether the clone counter also shows on the project page. The previous behavior was only in the editor. Now, this can be changed.

### Reason for changes

Viewers and creators might want to see it outside the editor in some cases.

### Tests

Tested (including the latest changes) on Edge with SA v1.30.0-pre.

### DIFF
--- a/addons/clones/addon.json
+++ b/addons/clones/addon.json
@@ -32,7 +32,7 @@
   "settings": [
     {
       "id": "projectpage",
-      "name": "Also show on project page",
+      "name": "Show on project page",
       "default": false,
       "type": "boolean"
     },
--- a/addons/clones/addon.json
+++ b/addons/clones/addon.json
@@ -8,6 +8,10 @@
     {
       "name": "OregSam",
       "link": "https://github.com/OregSamSas"
+    },
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
     }
   ],
   "dynamicEnable": true,
@@ -40,9 +44,8 @@
     }
   ],
   "latestUpdate": {
-    "version": "1.28.0",
-    "newSettings": ["showicononly"],
-    "isMajor": false
+    "version": "1.30.0",
+    "temporaryNotice": "You can now choose whether the clone counter shows up only in the editor or also on the project page."
   },
   "tags": ["editor", "projectPlayer", "recommended"],
   "enabledByDefault": false
--- a/addons/clones/addon.json
+++ b/addons/clones/addon.json
@@ -44,7 +44,7 @@
     }
   ],
   "latestUpdate": {
-    "version": "1.30.0",
+    "version": "1.31.0",
     "newSettings": ["projectpage"]
   },
   "tags": ["editor", "projectPlayer", "recommended"],
--- a/addons/clones/addon.json
+++ b/addons/clones/addon.json
@@ -45,7 +45,7 @@
   ],
   "latestUpdate": {
     "version": "1.30.0",
-    "newSettings": "projectpage"
+    "newSettings": ["projectpage"]
   },
   "tags": ["editor", "projectPlayer", "recommended"],
   "enabledByDefault": false
--- a/addons/clones/addon.json
+++ b/addons/clones/addon.json
@@ -45,7 +45,7 @@
   ],
   "latestUpdate": {
     "version": "1.30.0",
-    "temporaryNotice": "You can now choose whether the clone counter shows up only in the editor or also on the project page."
+    "newSettings": "projectpage"
   },
   "tags": ["editor", "projectPlayer", "recommended"],
   "enabledByDefault": false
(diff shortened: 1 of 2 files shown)


addons/clones/addon.json
```diff
@@ -26,6 +30,12 @@
     }
   ],
   "settings": [
+    {
+      "id": "editoronly",
+      "name": "Only show in editor",
```
Samq64, line 35: Would a show on project page setting be better?

addons/clones/userscript.js
```diff
@@ -76,7 +78,7 @@ export default async function ({ addon, console, msg }) {
       reduxEvents: ["scratch-gui/mode/SET_PLAYER", "fontsLoaded/SET_FONTS_LOADED", "scratch-gui/locales/SELECT_LOCALE"],
     });
 
-    if (addon.tab.editorMode === "editor") {
+    if (addon.tab.editorMode === "editor" || showOnlyInEditor === false) {
```
Samq64, line 81:
```suggestion
    if (addon.tab.editorMode === "editor" || !showOnlyInEditor) {
```

addons/clones/addon.json
DNin01: Maybe, since "only show in editor" is closer to implying that having the clone counter only appear in the editor is not the default option.

WorldLanguages: I agree. The opposite sounds more intuitive. It feels weird when enabling something removes functionality.

DNin01:
Wait, something's not quite right...

DNin01:
Okay, I fixed the issue.

(Also, I'm not quite sure why you have to enter and leave the editor for the counter to be hidden or shown after you switch the setting... maybe that's not a big deal.)

addons/clones/addon.json
```diff
@@ -31,9 +31,9 @@
   ],
   "settings": [
     {
-      "id": "editoronly",
-      "name": "Only show in editor",
-      "default": true,
+      "id": "projectpage",
+      "name": "Also show on project page",
```
Samq64, line 35: I don't think the word also is necessary here.
```suggestion
      "name": "Show on project page",
```

addons/clones/userscript.js
```diff
@@ -1,7 +1,7 @@
 export default async function ({ addon, console, msg }) {
   const vm = addon.tab.traps.vm;
 
-  let showOnlyInEditor = addon.settings.get("editoronly");
+  let alsoShowOnProjectPage = addon.settings.get("editoronly");
```
Samq64: I think you forgot to change the setting id here.
```suggestion
  let alsoShowOnProjectPage = addon.settings.get("projectpage");
```

Samq64: Never mind, that was already fixed.

addons/clones/addon.json
```diff
@@ -34,9 +44,8 @@
     }
   ],
   "latestUpdate": {
-    "version": "1.28.0",
-    "newSettings": ["showicononly"],
-    "isMajor": false
+    "version": "1.30.0",
+    "temporaryNotice": "You can now choose whether the clone counter shows up only in the editor or also on the project page."
```
Samq64, line 48: I don't think the notice is really necessary sense the new tag appears net to the setting. It it is kept, maybe it could be shortened a little?

addons/clones/userscript.js
```diff
@@ -1,6 +1,7 @@
 export default async function ({ addon, console, msg }) {
   const vm = addon.tab.traps.vm;
 
+  let alsoShowOnProjectPage = addon.settings.get("projectpage");
```
Samq64, line 4:
```suggestion
  let alsoShowOnProjectPage = addon.settings.get("projectpage");
```
"also" could probably be removed here too

addons/clones/addon.json
DNin01: I also forgot to add the "New!" label next to the new setting. I'll do that instead.

Samq64 (approved):

mxmou (changes requested):
This addon supports dynamic enable/disable so I think it's important to allow settings to be changed dynamically too. You can change the line
```js
if (v === 0) countContainerContainer.style.display = "none";
```
to
```js
if (v === 0 || (addon.tab.editorMode !== "editor" && !showOnProjectPage))
  countContainerContainer.style.display = "none";
```
and
```js
if (addon.tab.editorMode === "editor" || showOnProjectPage) {
  addon.tab.appendToSharedSpace({ space: "afterStopButton", element: countContainerContainer, order: 2 });
}
```
to
```js
addon.tab.appendToSharedSpace({ space: "afterStopButton", element: countContainerContainer, order: 2 });
doCloneChecks(true);
```

mxmou (approved):

PR #5069, opened 2022-09-01
Costume editor snapping addon

ErrorGamer2000:
Resolves #3639

### Changes

- Addon API Changes
  - Added new shared space to `addon.tab.appendToSharedSpace` to allow addons to add controls in the paint editor.
- Addon Changes
  - Made minor visual changes to the `onion-skinning` addon.
  - Changed the `onion-skinning` settings page positioning. No longer absolute, always centered over button.
  - Changed `onion-skinning` to use the new shared space.
- New Addon
  - Added the new `paint-snap` addon:
    - Allows users to enable/disable snapping altogether, including the native snap to center.
    - Allows users to customize the snapping threshold (or the minimum distance that the selection can move to snap).
    - Allows users to enable/disable snapping to:
      - Edges of the art board / page
      - Corners of the art board / page
      - Center of the art board / page, including the axes
      - Edges of other objects on the page (Groups are considered to be a single object)
      - Centers of other objects on the page
      - Corners of other objects on the page
    - Allows users to enable/disable snapping from:
      - Center of the selection
      - Corners of the selection
      - Centers of the edges of the selection (Really just acts like the full edge, kinda)
    - Works in both vector and bitmap modes.

### Reason for changes

Paint snapping can be extremely useful in other svg editors, so why not add it to scratch.

### Tests

Can't upload video, but tested all options and snapping, as well as changes to the `onion-skinning` UI as editing was being done as well as when it was finished.

PR #5496, opened 2022-12-28
Fix some onion-skinning and paint-snap visual issues

GarboMuffin:
This is a partial follow-up to https://github.com/ScratchAddons/ScratchAddons/pull/5069, primarily regarding changes to onion-skinning

 - Fix onion skinning setting labels wrapping unexpectedly
 - Restore previous input background color change when an onion skinning settings input was focused so the user can easily tell what input they're editing, reverting https://github.com/ErrorGamer2000/ScratchAddons/commit/7afb158dcfd9e2d3f36e75415059b31f61ed4a68. I feel that the browser's native text caret that is invisible half the time is not sufficient for indicating focus. If that commit was intended to fix a different "visual bug", I can't tell what it was.
 - Use a smaller motion when opening the onion skinning settings menu (similar to the animation pre-paint-snap). I feel that the animation was too large and distracting.

The last 2 also apply to the paint-snap addon.

@ErrorGamer2000

### DIFF
--- a/addons/onion-skinning/style.css
+++ b/addons/onion-skinning/style.css
@@ -91,6 +91,6 @@
   stroke: #ddd;
 }
 
-.sa-onion-label {
+.sa-onion-settings-label {
   white-space: nowrap;
 }
--- a/addons/onion-skinning/style.css
+++ b/addons/onion-skinning/style.css
@@ -1,7 +1,7 @@
 .sa-onion-button {
   position: relative;
 }
-.sa-onion-button:active {
+.sa-onion-button:focus-within {
   background-color: hsla(215, 100%, 65%, 0.2);
 }
 .sa-onion-button[data-enabled="true"] .sa-onion-image {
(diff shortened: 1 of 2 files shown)


ErrorGamer2000:
>  - Restore previous input background color change when an onion skinning settings input was focused so the user can easily tell what input they're editing, reverting https://github.com/ErrorGamer2000/ScratchAddons/commit/7afb158dcfd9e2d3f36e75415059b31f61ed4a68. I feel that the browser's native text caret that is invisible half the time is not sufficient for indicating focus. If that commit was intended to fix a different "visual bug", I can't tell what it was.

I changed that because using focus within will keep the background color if you click inside of the input and drag out, resulting in a highlighted input that is not actually focused.

GarboMuffin:
I am having trouble reproducing the bug you describe. Can you send a video? I'm sure that the bug can be fixed without negatively impacting accessibility.

Samq64 (approved):

WorldLanguages:
Pinging @ErrorGamer2000 for a follow-up

ErrorGamer2000:
I'll check it out tomorrow.

ErrorGamer2000:
@WorldLanguages Seems to be fine

PR #5499, opened 2022-12-28
fullscreen: fix incorrect position of answer input

mxmou:
### Changes

Fixes a bug in the enhanced full screen addon that caused the answer input to be positioned incorrectly (slightly too low and not centered).

### Tests

Before:
image

After:
image


WorldLanguages (approved):
Didn't test.

Samq64 (approved):

PR #5508, opened 2022-12-29
clones: correct count when dynamically enabled

### LINKED ISSUE #4635: Clone counter addon can show negative amount of clones
> It says I have -3 clones. It worked after I reloaded.

https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/clones/userscript.js#L47

mxmou:
Resolves #4635

### Changes

Calculates the correct clone count when `clones` is dynamically enabled.

### Reason for changes

To prevent the counter from showing a negative number.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/clones/userscript.js
+++ b/addons/clones/userscript.js
@@ -70,6 +70,16 @@ export default async function ({ addon, console, msg }) {
     return ret;
   };
 
+  if (addon.self.enabledLate) {
+    // Clone count might be inaccurate if the user deleted sprites
+    // before enabling the addon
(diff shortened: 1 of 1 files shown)


WorldLanguages:
I'm not sure if we should try to correct the count. We might need to end up fixing the edge case of the edge case of the edge case. I think we could merge this, but we should avoid making additional changes like this.

WorldLanguages (approved):
Tested in Chrome

This can be merged now

PR #5511, opened 2022-12-29
Scratch font

valmontechno:
Add the "Scratch" font to the text tool of the costume editor

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -133,6 +133,7 @@
   "rename-broadcasts",
   "middle-click-popup",
   "number-pad",
+  "scratch-font",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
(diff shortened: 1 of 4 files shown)


WorldLanguages:
Has it been removed recently? Where's the context?

CST1229:
> Has it been removed recently? Where's the context?

The font selection item has not been there since 3.0's launch. I don't think the font itself was removed yet.

BroJac5246:
> > Has it been removed recently? Where's the context?
> 
> The font selection item has not been there since 3.0's launch. I don't think the font itself was removed yet.

That's my understanding as well.

WorldLanguages:
Ohh, I see. Would this addon continue to work if Scratch removes the font (not from the list, but from Scratch)

This idea is probably rejected because of licensing issues, right?

CST1229:
> Ohh, I see. Would this addon continue to work if Scratch removes the font (not from the list, but from Scratch)

Probably not.

BroJac5246:
> This idea is probably rejected because of licensing issues, right?

I heard somewhere that that's why it was removed, but I'm not against this. There are still simple workarounds to access the font, so it's not like we're doing anything that's otherwise impossible.

PR #5512, opened 2022-12-29
middle-click-popup: fix rename-broadcasts compatibility

GarboMuffin:
Fixes this bug from feedback https://discord.com/channels/751206349614088204/758464349768515604/1058040403468750898

> if u search b r b when adding a block with middle click it will suggest you broadcast ( rename broadcast ) which isn't supposed to happen as its a custom option to rename a broadcast and not a broadcast you have used and if u click on it, it creates the broadcast "rename broadcast"
> very minor bug but its a bug sooo :P

### DIFF
--- a/addons/middle-click-popup/userscript.js
+++ b/addons/middle-click-popup/userscript.js
@@ -426,7 +426,9 @@ export default async function ({ addon, msg, console }) {
             code === "createLocalVariable" ||
             code === "createGlobalList" ||
             code === "createLocalList" ||
-            code === "createBroadcast"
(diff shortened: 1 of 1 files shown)


WorldLanguages:
I guess we have to remember about this filtering every time a new addon wants to add their own items to dropdowns, right?

Tacodude7729:
For the dropdown options with a function attached it's probably safer to use a whitelist instead of a blacklist. Don't worry about it now I'll change it in my improvement PR when I get the chance

WorldLanguages (approved):

PR #5514, opened 2022-12-29
TTS Captions

GrahamSH-LLK:
### Changes

Adds text to speech captions
image

### Reason for changes

It's widely requested and is a good "accessibility" patch on our shoulder
### Tests

Yes

### DIFF
--- a/addons/tts-captions/addon.json
+++ b/addons/tts-captions/addon.json
@@ -1,0 +1,27 @@
+{
+  "name": "TTS Captions",
+  "description": "Replaces links to the embedded site with real YouTube links.",
+  "credits": [
+    {
+      "name": "GrahamSH"
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["projects"]
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"],
+      "runAtComplete": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": false,
+  "versionAdded": "1.0.0",
+  "tags": ["community", "forums"],
+  "enabledByDefault": false
+}
--- a/addons/tts-captions/addon.json
+++ b/addons/tts-captions/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "TTS Captions",
-  "description": "Replaces links to the embedded site with real YouTube links.",
+  "name": "Text to Speech Captions",
+  "description": "Adds live captions for the text to speech voice.",
   "credits": [
     {
       "name": "GrahamSH"
@@ -19,9 +19,9 @@
       "runAtComplete": false
     }
   ],
-  "dynamicEnable": true,
+  "dynamicEnable": false,
   "dynamicDisable": false,
-  "versionAdded": "1.0.0",
-  "tags": ["community", "forums"],
+  "versionAdded": "1.30.0",
+  "tags": ["projectPage"],
   "enabledByDefault": false
 }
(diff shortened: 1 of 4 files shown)


retronbv:
you seem to have left your sanity check in there

fractioneater:
Shouldn't this have the community tag?

mxmou:
> Shouldn't this have the community tag?

Indeed - that's why schema validation is failing.

Secret-chest:
Could the [Sprite] tag be bold or something?

mxmou:
> Could the [Sprite] tag be bold or something?

Is it even necessary? Text spoken by different sprites doesn't sound different in vanilla Scratch.

Secret-chest:
> > Could the [Sprite] tag be bold or something?
> 
> Is it even necessary? Text spoken by different sprites doesn't sound different in vanilla Scratch.

It would be to distinguish it. Maybe make all sprite tags the same size so text gets aligned nicely.

BroJac5246:
> > Could the [Sprite] tag be bold or something?
> 
> Is it even necessary? Text spoken by different sprites doesn't sound different in vanilla Scratch.

Maybe showing it should be a setting.

CST1229:
> Is it even necessary? Text spoken by different sprites doesn't sound different in vanilla Scratch.

What about e.g different characters with different voices?

Hans5958:
Weak opinion: The style of the captions needs to be done better. Something like [Google Meets' captions would work](https://workspaceupdates.googleblog.com/2019/04/live-captions-in-meet.html) comes in my mind, but anything works really.

addons/tts-captions/addon.json
```diff
@@ -0,0 +1,27 @@
...
+    {
+      "url": "userscript.js",
+      "matches": ["projects"],
+      "runAtComplete": false
+    }
+  ],
+  "dynamicEnable": false,
+  "dynamicDisable": false,
+  "versionAdded": "1.30.0",
+  "tags": ["projectPage"],
```
mybearworld, line 25: This should have the community tag, as mentioned in https://github.com/ScratchAddons/ScratchAddons/pull/5514#issuecomment-1367882657.

```suggestion
  "tags": ["community", "projectPage"],
```

addons/tts-captions/userscript.js
```diff
@@ -0,0 +1,74 @@
+import createLogView from "../debugger/log-view.js";
+export default async function ({ addon, console }) {
+  console.log("sanity check");
+  const vm = addon.tab.traps.vm;
+  const logView = new createLogView();
+  console.log(logView);
+  logView.placeholderElement.textContent = "no logs";
```
mybearworld, line 7: This should probably use `msg`.

addons/tts-captions/style.css
```diff
@@ -0,0 +1,55 @@
...
+  font-style: italic;
+}
+
+.sa-debugger-log-end {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 1px;
+  height: 1px;
+}
```
mybearworld, line 31: This is never used.

```suggestion
```

addons/tts-captions/userscript.js
```diff
@@ -0,0 +1,74 @@
+import createLogView from "../debugger/log-view.js";
```
mybearworld, line 1: Does this really need to use debugger? I think this is overcomplicating it.

mybearworld (changes requested):
My first review on GitHub, so I've probably done everything wrong. Also, the addon has just stopped working for me locally, so I can't really check if changes break anything.

GrahamSH-LLK:
This doesn't work for dynamic values like joins and variables

WorldLanguages:
I'm going to create an issue for this suggestion
Here it is: #6016

PR #5530, opened 2023-01-02
Lint 20230103

apple502j:

PR #5533, opened 2023-01-03
mute-project: Fix mute icon

GarboMuffin:
#5530 broke the icon:

image

Tested in Firefox on Linux


apple502j:
Will be reverting the PR entirely, so closing. Also this is potentially a SVGO bug.

PR #5424, opened 2022-12-07
Fix hide-flyout and sprite-folders conflicts

ErrorGamer2000:
### Changes

Slightly tweaked the `z-index` change on the stage wrapper to prevent sprite context menus from being cut off with the folders addon enabled.

PR #5534, opened 2023-01-03
Unbreak hide-flyout and hide-stage compatibility

GarboMuffin:
I broke this in https://github.com/ScratchAddons/ScratchAddons/pull/5424. The stage would still use some space on the right:

image

Tested fix in RTL and LTR in Firefox

### DIFF
--- a/addons/hide-stage/style.css
+++ b/addons/hide-stage/style.css
@@ -16,10 +16,13 @@
   border-radius: 0;
 }
 
-.sa-stage-hidden [class*="gui_stage-and-target-wrapper_"],
+/* [class*="gui_flex-wrapper_"] is for specificity over hide-flyout */
+.sa-stage-hidden [class*="gui_flex-wrapper_"] [class*="gui_stage-and-target-wrapper_"],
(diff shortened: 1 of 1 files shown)


addons/hide-stage/style.css
```diff
@@ -16,10 +16,13 @@
...
 }
 
-.sa-stage-hidden [class*="gui_stage-and-target-wrapper_"],
+/* [class*="gui_flex-wrapper_"] is for specificity over hide-flyout */
+.sa-stage-hidden [class*="gui_flex-wrapper_"] [class*="gui_stage-and-target-wrapper_"],
 .sa-stage-hidden [class*="stage-wrapper_stage-wrapper_"]:not([class*="stage-wrapper_full-screen_"]),
 .sa-stage-hidden [class*="gui_target-wrapper_"] {
   padding: 0;
+  /* override hide-flyout */
+  padding-inline: 0;
```
mxmou, line 25: This line doesn't appear to have any effect - the specificity change is enough to fix the bug.

GarboMuffin: Removed, thanks

mxmou (approved):

PR #5540, opened 2023-01-04
Speed up open full screen settings popup glide-in

DNin01:
### Subject

The notification that pops up when you try to enable the Scratch Notifier addon from the extension popup. It needs you to open the full screen settings page because I guess the extension popup cannot request permissions.
Popup reading, "to enable the 'Scratch Notifier' addon, go to the full screen settings page."

### Changes

Made the drop-down duration of this popup 0.35s and the dismiss duration 0.6s. Basically, I just sped up the animation. It may seem like a lot, but it was pretty slow to begin with.

### Reason for changes

With the previous, slower animation speed, it would take a second for it to come onscreen. This change makes it snappier and more responsive, which can help make it easier to notice.

### Tests

Tested on Microsoft Edge.

### DIFF
--- a/webpages/settings/components/addon-body.js
+++ b/webpages/settings/components/addon-body.js
@@ -96,7 +96,7 @@ export default async function ({ template }) {
           if (result === false) {
             if (isIframe) {
               this.$root.addonToEnable = this.addon;
-              document.querySelector(".popup").style.animation = "dropDown 1.6s 1";
+              document.querySelector(".popup").style.animation = "dropDown 0.5s 1";
               this.$root.showPopupModal = true;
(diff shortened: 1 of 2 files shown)


apple502j (dismissed):

mxmou:
I think it's still too slow.

Samq64 (dismissed):

Samq64:
It could be slightly faster, but I think it's fin like this too.

DNin01:
I have increased the speed further.

mxmou (approved):

Samq64 (approved):

PR #5562, opened 2023-01-07
gamepad: Remove hints cache

GarboMuffin:
This fixes an issue from my Scratch comments:

>  Problem with Turbowarp. I'm using the desktop client, so I don't know if that's the issue. Basically, when I try to change the gamepad controls using the gamepad support addon, it resets the controls when I unplug the controller, even after saving the project. Am I doing something wrong? 

1. Plug in a controller
2. Customize settings
3. Save them to the project
4. Unplug controller
5. Plug controller in again
6. It should have the settings that were customized, not the default ones

Controllers are not plugged and unplugged often enough to justify have a fragile cache here

### DIFF
--- a/addons/gamepad/gamepadlib.js
+++ b/addons/gamepad/gamepadlib.js
@@ -153,6 +153,7 @@ class GamepadData {
   }
 
   resetMappings() {
+    this.hints = this.gamepadLib.getHints();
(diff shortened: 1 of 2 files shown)


apple502j (approved):

Samq64 (approved):
Tested on Firefox with a Bluetooth Xbox controller.

Issue #5561, opened 2023-01-07
Sprite properties disapear when changing language


RedGuy12:
### Description

The properties and the I icon both disappear so i cant get them back

### Steps to reproduce

1. Open editor
2. Change language
3. Observe bug

### Demonstration(s) and additional context

![image](https://user-images.githubusercontent.com/61329810/211131120-96b0ed7d-320e-42b4-a9fb-3766508a914f.png)

### Environment

Browser: Edge
Scratch Addons version: 1.30.0
Operating system: Windows

Issue #5557, opened 2023-01-06
Sprite Panel Collapse Speed is always instant


CubesterYT:
### Description

When having the Sprite Panel Collapse addon on, when collapsing and un collapsing it, the speed is the exact same, for all settings, which is instant:

https://user-images.githubusercontent.com/78769806/211115550-2a233486-e359-4a42-992a-3ad14a99eeb7.mp4

### Steps to reproduce

1. Enable the addon
2. Use the addon/change all speed settings
3. Nothing else

### Environment

Browser: Chrome 108.0.5359.125
Scratch Addons version: 1.30.0
Operating system: Windows 10

GarboMuffin:
it's a compatibility issue with drag-drop -- they both want to transition the same element

CubesterYT:
Oh yeah, I have drag and drop on:

![image](https://user-images.githubusercontent.com/78769806/211116379-ff89fd55-bba4-4e59-86b4-95e063e538e8.png)

PR #5563, opened 2023-01-07
sprite-properties: Fix some bugs

GarboMuffin:
- Fixes https://github.com/ScratchAddons/ScratchAddons/issues/5561
 - Fixes https://github.com/ScratchAddons/ScratchAddons/issues/5572 / https://discord.com/channels/751206349614088204/758464349768515604/1061135466251943946 - properties panel now works in ["wide" languages](https://github.com/LLK/scratch-gui/blob/e15b2dfa3a2e58e80fae8d1586c7f56aa0cc0ede/src/lib/locale-utils.js#L6-L18) such as Japanese and Italian
 - Fixes collapse icon being draggable
 - Fixes RTL support
 - Fixes properties panel after entering editor momentarily being hidden when auto-collapse is disabled
 - Whether the panel is collapsed or not is now preserved through leaving the editor and going back in rather than resetting to the default from settings
 - Fixes info button disappearing after renaming sprite
 - Fixes performance issue caused by running document.querySelector every time targets update (oftentimes every frame)
 - Fixes compatibility with folders addon - if the current target was in a closed folder and the folder was opened, the info button would not be added

drag-drop related changes and fix to #5557 moved to https://github.com/ScratchAddons/ScratchAddons/pull/5570

### DIFF
--- a/addons/drag-drop/dragged-over.css
+++ b/addons/drag-drop/dragged-over.css
@@ -1,11 +1,13 @@
 /* Based on the `stage-selector_raised` class */
-[class*="stage-selector_header_"],
-[class*="sprite-selector_sprite-selector"],
-[class*="monitor_list-body"] {
+div[class*="stage-selector_header_"],
+div[class*="sprite-selector_sprite-selector"],
+div[class*="sprite-info_sprite-info"],
+div[class*="monitor_list-body"] {
   transition: background-color 0.25s ease;
 }
 .sa-dragged-over,
-.sa-dragged-over [class*="stage-selector_header_"],
-.sa-dragged-over [class*="monitor_list-body"] {
+.sa-dragged-over div[class*="stage-selector_header_"],
+.sa-dragged-over div[class*="sprite-info_sprite-info"],
+.sa-dragged-over div[class*="monitor_list-body"] {
   background-color: hsla(215, 100%, 77%, 1);
 }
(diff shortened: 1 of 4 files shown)


Samq64:
Oops, I didn't mean to do that...

Samq64:
The sprite properties state always resets to collapsed when changing the language which causes this when collapse by default is disabled:
[Screencast from 2023-01-08 12-43-59.webm](https://user-images.githubusercontent.com/81489795/211211823-d3a3d87c-c84c-4c82-9b93-1580545bc7ca.webm)

GarboMuffin:
The scope of this PR has grown significantly larger than I had initially expected but I believe it is basically done now, maybe with some minor code style or comment clean up

apple502j (approved):

Samq64:
I noticed that `Automatically collapse when mouse leaves sprite panel` can be enabled even if  `Collapse panel by default` isn't. Should that setting just be hidden?

WorldLanguages:
@Samq64 I think that combination of settings is valid like any other.

Samq64 (approved):

Samq64:
I meant adding an if to the collapse by default setting so this can't happen:
image

It doesn't really matter though.

mxmou:
> I meant adding an if to the collapse by default setting so this can't happen: image
> 
> It doesn't really matter though.

I think it makes sense to try to prevent this configuration because it behaves weirdly (the properties are initially visible but moving the mouse to the sprite pane and back hides them.).

PR #5566, opened 2023-01-07
variable-manager: Require a click to display variables that are way too big

GarboMuffin:
When variables exceed around 1,000,000 characters in length, they cause excessive performance issues and sometimes cause the whole tab to crash. (there was an issue in TW/scratch-gui about this, but I can't find it anymore)

Instead of that, this PR shows a warning:

image

The thresholds are set high enough that most people will never see this message.

An earlier version of this PR with similar thresholds has been in TurboWarp for a very long time with the message "Too big to safely display. If this limit is too low, use the feedback button at the top of the screen." No one has submitted feedback about this.

### DIFF
--- a/addons-l10n/en/variable-manager.json
+++ b/addons-l10n/en/variable-manager.json
@@ -2,5 +2,6 @@
   "variable-manager/for-this-sprite": "Variables for this sprite",
   "variable-manager/for-all-sprites": "Variables for all sprites",
   "variable-manager/variables": "Variables",
-  "variable-manager/search": "Search"
+  "variable-manager/search": "Search",
+  "variable-manager/too-big": "Too big to display."
 }
--- a/addons-l10n/en/variable-manager.json
+++ b/addons-l10n/en/variable-manager.json
@@ -3,5 +3,5 @@
   "variable-manager/for-all-sprites": "Variables for all sprites",
   "variable-manager/variables": "Variables",
   "variable-manager/search": "Search",
-  "variable-manager/too-big": "Too big to display."
+  "variable-manager/too-big": "Click to display very large value."
 }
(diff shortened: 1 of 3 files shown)


addons-l10n/en/variable-manager.json
```diff
@@ -2,5 +2,6 @@
   "variable-manager/for-this-sprite": "Variables for this sprite",
   "variable-manager/for-all-sprites": "Variables for all sprites",
   "variable-manager/variables": "Variables",
-  "variable-manager/search": "Search"
+  "variable-manager/search": "Search",
+  "variable-manager/too-big": "Click to display very large value."
```
jeffalo, line 6: Could this message state the max length? eg. `Click to display very large value. (1000000+ characters)` (with a separate message for lists) or would that be unnecessary information?

jeffalo (approved):

addons-l10n/en/variable-manager.json
WorldLanguages: I don't think it's really necessary to tell users what the exact limit is.

addons/variable-manager/userscript.js
```diff
@@ -85,18 +85,32 @@ export default async function ({ addon, console, msg }) {
...
       if (!this.visible && !force) return;
+
       let newValue;
+      let maxSafeLength;
       if (this.scratchVariable.type === "list") {
         newValue = this.scratchVariable.value.join("\n");
+        maxSafeLength = 5000000;
       } else {
         newValue = this.scratchVariable.value;
+        maxSafeLength = 1000000;
```
Samq64, line 102: If the value can be viewed by clicking on it, I think the limit could probbably be one tenth this size. The current one is fine though.

Samq64 (approved):

Issue #5145, opened 2022-09-29
Improve Keybind/Hotkey Handling


Tacodude7729:
As I see it, currently, the way SA handles keybindings is not ideal. 
- There is no way to rebind keys on anything
- There is no way to keybind a lot of the vanilla behaviors without an addon ( and addons like 'duplicate scripts with alt key' don't support changing the key either)
- The handling of keybinds must be implemented by every addon individually  

The only real progress that has been made on this was the addition of the currently unused 'key' setting type in #793.

========================

_Not sure if an issue is the best place to put this wall of text, but here we go._

I've put a bit of though into it and what follows is the solution I think would work best to mitigate all of these problems. It involves some substantial changes and contains a large amount of what I consider my personal opinion on how things should be done, so I want feedback before I try and actually do any of it. I also cannot do all of this by myself (I have no clue about vue.js :sweat_smile: ) so I would need some help.

The idea of keybinds as a setting is good but the current implementation has issues. Right now, when an addons uses `settings.get()` on a 'key' setting, you get a string like  `Ctrl + Shift + F`, for an addon to use this it has to be parsed then code has to be written to detect that keybind. Instead, for key settings, `settings.get()` should return an object that contains helpful things for implementing the keybind, such as an isPressed property, a way to get the keys in the bind without needing to parse a string and events that can be subscribed to for when the keybind is pressed and when it is released. I think this would have to be implemented through an additional script which gets injected to every page and listens for the keybinds of the addons that are enabled on that page. The 'key' settings also needs more options for constraining what types of keys can be bound to it.

As well as these core changes, a new addon (maybe enabled by default) would need to be added for the keybind manager. This addon adds a new popup and tab which allows you to search keybinds from all addons by name or by the actual keybinding and change them from there (I'm thinking something like the keyboard shortcuts menu in vscode). It would also provide keybindings for vanilla scratch features that don't have one, so addons like 'duplicate scripts with alt key' or #705 could be put into this one. Keybinds should be mostly or all empty by default so loads of keybinds could be added without worrying about users having every key on their keyboard doing something they didn't set it to. 
This addon could be split in two, one for the manager and another for providing keybinds for vanilla features but I'm not sure why you'd want one without the other.

[...]

Issue #4946, opened 2022-08-03
Add a Hotkey Manager


BroJac5246:
### The Problem
Okay, there's been a ton of discussion about this, but no one has actually opened this issue, so I'm gonna do it right here :)

Basically, Scratch Addons now has a ton of addons that use hotkeys/keyboard shortcuts/etc. This is great! You can perform tons of neat little actions. However, this comes with two major problems:

1. There are now way too many hotkeys to keep track of- and when you throw in some of Scratch's lesser known ones, it can just be too confusing, for both the users and the developers
2.  The wide range of devices & browsers can cause compatibility issues. See the list of related issues below.

In #4662 (also linked below), WorldLanguages and I came to the conclusion that a hotkey section in the settings would be the best solution. However, the specifics need to be discussed, which is part of why I'm opening this.

Here's what we need to consider:

1. Where would the hotkeys settings show up? In both the full screen settings _and_ the popup, or just the full screen settings? Because of how important this will be, I think we should go with the former, but that's just my opinion.
2. How will the new approach to hotkeys be? This is a HUGE question. We already know there will be a dedicated manager, but how will this work? How will hotkeys be managed? Will addons specify in their `addon.json` that they would like to request a hotkey(s)?
3. Will the hotkey manager allow users to change the hotkeys included in vanilla scratch? [This comment by WL](https://github.com/ScratchAddons/ScratchAddons/issues/4662#issuecomment-1166376638) demonstrates how this could be useful. (Read my comment after that one as well.) This could be complicated, though, with we need to consider.
4. Will different hotkeys be suggested, or even allowed, based on OS, to help users fix compatibility issues before they even arise?
5. Will additional hotkeys be introduced when this is implemented, such as those recommended in #4725 
6. What will happen to addons that use hotkeys? Will these be hidden from users? What if they need to have additional settings?
7. How will migration work? This was somewhat mentioned in Points 2 and 6, but what will happen to existing addons? Will they need to be modified for the new approach? Will we have to add a `hotkey` option to the addon manifest (`addon.json`)? How will users' settings be ported?
8. Not a big problem, but we also might need to remember to implement credits into the hotkey manager, depending on the approach we take.

### Describe alternatives you've considered
_Not_ implement a hotkey manager and let users suffer?

### Issues that Demonstrate the Need for this
Issues that demonstrate compatibility issues:
 #4466 and #4662

Issues that suggest keyboard shortcuts:
[...]

PR #5574, opened 2023-01-08
Hold ctrl to enable or disable snapping

### LINKED ISSUE #5564: Offer a different set of snapping guidelines while holding ctrl
### Why this would be helpful

Some people are overwhelmed by the massive amount of snapping settings, and it can be hard to drag an object to where you want it with all of the snapping getting in the way.

### My suggestion

If there are two different settings for snapping, one while ctrl is held, and one where it isn't held, people will be able to potentially work faster once they get used to it. The cmd key would probably be used on mac.

Example: If I set all corner snapping to the ctrl key, the corners wouldn't get in the way unless I was intentionally holding down the ctrl key.

### Possible alternatives

This would go well with the settings in #5555.

### Additional context

_No response_

fractioneater:
Resolves #5564

This is my first PR, so if I did anything incorrect, please tell me.

### Changes

You can now hold ctrl (or meta) to temporarily enable or disable snapping, depending on if snapping is toggled.

### Reason for changes

Snapping is helpful, but not always, so now it can be temporarily disabled without having to move your mouse to the toggle button.

### Tests

Tested on ChromeOS 108
Dynamic enable and disable work.


addons/paint-snap/addon.json
```diff
@@ -30,6 +30,39 @@
...
       "allowTransparency": true
     },
+    {
+      "type": "positive_integer",
+      "id": "threshold",
+      "name": "Snapping distance",
+      "default": 10,
+      "min": 4,
+      "max": 100
+    },
```
fractioneater, line 40: This is to make it more consistent with the popup inside the costume editor.

addons/paint-snap/addon.json
```diff
@@ -30,6 +36,39 @@
...
+    {
+      "type": "boolean",
+      "id": "ctrl",
+      "name": "Ctrl key modifier",
```
Samq64, line 50: Could this toggle be removed and just be a third option in the setting below?

mxmou:
I don't think this feature needs any settings. If snapping is disabled, holding Ctrl should enable it, if it's enabled, holding Ctrl should disable it.

mxmou:
> > I don't think this feature needs any settings. If snapping is disabled, holding Ctrl should enable it, if it's enabled, holding Ctrl should disable it.
> 
> I think it should at least have a toggle, because the ctrl button will conflict with #5400, and some people might want to turn it off.

I think #5400 could be implemented in a way that doesn't require a modifier key, for example by adding a separate skew tool.

fractioneater:
Oops, I forgot to take out the other one.

I have to go right now, though.

Samq64 (dismissed):

ErrorGamer2000:
I have somewhat mixed feelings about this PR. While I can see how it would be useful to some users, it is not exactly a necessary feature. You could simply take the half of a second it takes to press the button before moving the object, and reenable it after. This would also, as has already been mentioned, cause a conflict with skewing, as well as limit future addons that may change the costume editor. I also tried to remain a faithful as possible to the common snapping aspects of other big svg editors, such as inkscape or affinity designer. Snapping, in my opinion, should not be affected by a keybind that could limit the functionality of other aspects, or a keybind that could also change anything else on the workspace.

fractioneater:
> You could simply take the half of a second it takes to press the button before moving the object, and reenable it after.

The main reason why I created this PR is because I was trying to drag something to a point that I needed snapping to get to, and then I needed snapping disabled to move it some more in a different direction. I couldn't move my mouse to the toggle button, because I needed precise control over the object, so I thought a key to toggle it would be useful.

ErrorGamer2000:
> I needed snapping disabled to move it some more in a different direction

I understand, but there is also the fact that you could easily drop the object at the snapped position, turn off the snapping, and then move it again, maybe with a lower mouse DPI. You can also zoom in in the paint editor to get finer movement control

WorldLanguages:
@fractioneater What do you think?

fractioneater:
> @fractioneater What do you think?

I think that the ctrl key can be used in a lot of ways, but only once, so it should only be used when it is really important that you can let go of it and hold it down again while dragging something.

#5400 and #5596 could potentially conflict with this. The costume editor has a lot of actions being added to it, but skewing could be a separate tool (as @mxmou suggested), and middle clicking is not used anywhere else. If this PR is merged, the rest of the addons that affect dragging in the costume editor will have to use a different modifier key (there aren't any more) or have a separate tool for that action.

ErrorGamer2000:
This is the reason I think it would be better to not merge this pr. I'm sorry @fractioneater but I think is a better idea to leave the modifier key for a different addon that may actually need it.

fractioneater:
> I think it is a better idea to leave the modifier key for a different addon that may actually need it.

Yes, I agree with this. It's probably not a good idea to use up all of the modifier keys on features that probably won't even get noticed by users.

WorldLanguages:
I'd say we put this on hold until we have a proper hotkey managment system (#4946, #5145), because until we have something like that, all hotkeys we add will be "default hotkeys" that will be difficult to change later.
In the future, we could offer this feature with no default key modifier and let the user choose what hotkey to use, and it will be the user's job to find a modifier key that isn't already taken.

PR #5582, opened 2023-01-09
block-switching: Add more switching

GarboMuffin:
- It is now possible for switches to create inputs that don't already exist
   - merged say <-> think and say for seconds <-> think for seconds into one group
   - added delete item of list <-> delete all of list switch
   - added go to front/back layer <-> go forward/backward n layers switch
   - added touching color <-> color touching color switch
   - inputs with values such as "1" are created when needed
   - going the other way, inputs that just contain a text field such as "banana" will be deleted while more complicated inputs will be split off
 - Converted copy+pasted code from scratch-blocks to getBlockly() (this addon predates getBlockly())


Samq64 (approved):
I don't know much about Blockly, but I tried every switch group and there were no console errors and the blocks worked as expected.

WorldLanguages:
Merging, as this has been cherry-picked on Turbowarp for 2 weeks with no bug reports.

PR #5583, opened 2023-01-09
`progress-bar`: Round the progress bar and set default height to 8px

BroJac5246:
Resolves no open issues.

### Changes
The progress bar in the `progress-bar` addon is now rounded. It previously had disgusting square edges that stuck out, so this helps it look better.

I also set the default `Progress bar height (px)` value to ~~10~~ 8 (previously 5), since 5 is very small and hides away.

### Reason for changes
The square edges were aesthetically unpleasing, and the small size made it appear as if the addon was trying to hide as opposed to being useful, so I improved it.

### Tests
Working on Chrome 109 (beta, 32-bit) on a Chromebook. I would appreciate it if someone could test this on Firefox.

### DIFF
--- a/addons/progress-bar/addon.json
+++ b/addons/progress-bar/addon.json
@@ -32,7 +32,7 @@
       "type": "integer",
       "min": 3,
       "max": 25,
-      "default": 10
+      "default": 8
     }
   ],
   "dynamicEnable": true,
--- a/addons/progress-bar/addon.json
+++ b/addons/progress-bar/addon.json
@@ -32,7 +32,7 @@
       "type": "integer",
       "min": 3,
       "max": 25,
-      "default": 5
+      "default": 10
     }
   ],
   "dynamicEnable": true,
--- a/addons/progress-bar/userstyle.css
+++ b/addons/progress-bar/userstyle.css
@@ -3,11 +3,13 @@
   height: calc(1px * var(--progressBar-height));
   box-sizing: border-box;
   border: 1px solid white;
+  border-radius: calc(1px * var(--progressBar-height));
 }
 .u-progress-bar-inner {
   width: 0;
   height: 100%;
   background-color: white;
+  border-radius: calc(1px * var(--progressBar-height));
 }
 
 .u-progress-bar-top {
--- a/addons/progress-bar/userstyle.css
+++ b/addons/progress-bar/userstyle.css
@@ -23,6 +23,7 @@
 }
 .u-progress-bar-top .u-progress-bar-inner {
   transition: width 0.2s;
+  border-radius: 0px;
 }
 
 .u-progress-bar-caption {


addons/progress-bar/addon.json
```diff
@@ -32,7 +32,7 @@
       "type": "integer",
       "min": 3,
       "max": 25,
-      "default": 5
+      "default": 10
```
Samq64, line 35:
```suggestion
      "default": 8
```
I think 10 pixels is a little too big in the editor menubar to be used as the default.

addons/progress-bar/userstyle.css
```diff
@@ -3,11 +3,13 @@
   height: calc(1px * var(--progressBar-height));
   box-sizing: border-box;
   border: 1px solid white;
+  border-radius: calc(1px * var(--progressBar-height));
```
Samq64, line 6:
```suggestion
  border-radius: 50%;
```

addons/progress-bar/userstyle.css
```diff
@@ -3,11 +3,13 @@
...
 }
 .u-progress-bar-inner {
   width: 0;
   height: 100%;
   background-color: white;
+  border-radius: calc(1px * var(--progressBar-height));
```
Samq64, line 12:
```suggestion
  border-radius: 50%;
```

Samq64 (approved):

DNin01 (commented):
It looks really nice!

By the way, it also gets rounded when placed at the top of the screen (if you have the setting enabled). Not sure if that's intentional.

BroJac5246:
> It looks really nice!
> 
> By the way, it also gets rounded when placed at the top of the screen (if you have the setting enabled). Not sure if that's intentional.

It's intentional, but do you think it shouldn't do that?

DNin01:
> > It looks really nice!
> > By the way, it also gets rounded when placed at the top of the screen (if you have the setting enabled). Not sure if that's intentional.
> 
> It's intentional, but do you think it shouldn't do that?

When the corners are square on the top-of-the-page progress bar, it just looks more attached to the top of the page (or something like that), and that's the way it is on most websites.

BroJac5246:
> > > It looks really nice!
> > > By the way, it also gets rounded when placed at the top of the screen (if you have the setting enabled). Not sure if that's intentional.
> > 
> > It's intentional, but do you think it shouldn't do that?
> 
> When the corners are square on the top-of-the-page progress bar, it just looks more attached to the top of the page (or something like that), and that's the way it is on most websites.

So flat on the top, round on the bottom?

DNin01:
> So flat on the top, round on the bottom?

I'd probably be fine with that too.

BroJac5246:
> > So flat on the top, round on the bottom?
> 
> I'd probably be fine with that too.

Which would you prefer?

DNin01:
> > > So flat on the top, round on the bottom?
> > 
> > 
> > I'd probably be fine with that too.
> 
> Which would you prefer?

I'd prefer if the top-of-the-page progress bar had all corners square.

mxmou:
This already looks much better than the current design, but I think it would be even nicer if it had a 25% transparent white background and no border, like this:
image

Secret-chest:
If the edges are rounded and the show at the top of the page option is enabled, it wouldn't sit flush with the window.

CST1229:
We could also use something like one of the designers' [mockups](https://github.com/LLK/scratch-gui/issues/827#issuecomment-368538369), which looks like this (colored progress bar, dark background):

mxmou:
@CST1229 That looks good too, but I think white can adapt to dark mode settings more easily.

BroJac5246:
@CST1229 I prefer the look of mxmou's design, plus they make a good point about dark mode. If you do want a colored progress bar then you can always create an issue, since I can see some users wanting that.

BroJac5246:
I've been busy recently, but I'll try to update this when I can.

DNin01 (approved):
I like it!

BroJac5246:
Should I have added myself to the credits? Do we have any specific guidelines on if one should do so?

WorldLanguages:
@BroJac5246 It's not mandatory, but I think you could have added yourself if you want.
If you want to be credited you could suggest mxmou to add you on #5628

WorldLanguages:
This PR is being continued here: https://github.com/ScratchAddons/ScratchAddons/pull/5628

PR #5586, opened 2023-01-09
zebra-striping: compare colors instead of categories

### LINKED ISSUE #5194: Alternating nested block colors (zebra striping): Fails when two categories are the same color
### Description

"Alternating nested block colors" works by checking if a block and the parent it's directly nested in are of the same category (Looks, Motion, Control, Operators, etc). This fails to consider that the user might override two categories to have the same color.

### Steps to reproduce

1. Ensure two block categories are the same color, ex. with the "Customizable block colors" addon
2. Ensure "Alternating nested block colors" is enabled and has an effect normally
3. Nest a block of one category inside a block of another category with the same color
4. Note that the color of the nested block isn't affected by zebra striping

Variable and list categories overridden to be the same color, with zebra striping lightly darkening nested blocks (see "Self -> Heal" in the middle):

Several list blocks nested in variable blocks or vice versa. Only the variable block directly nested in another variable block is affected by zebra striping.

### Environment

Browser: Firefox 105.0.3
Scratch Addons version: 1.28.0
Operating system: macOS 11.7 Big Sur

mxmou:
Resolves #5194

### Changes

Enables zebra striping when two blocks have the same color, even if the category is different. This only works if the two colors are exactly the same.

### Reason for changes

It's possible to choose the same color for two categories.

### Tests

Tested on Edge and Firefox, using Appel as an example project. In the first screenshot, variables and lists have the same color. In the second screenshot, operators and variables have the same color.
image
image

### DIFF
--- a/addons/zebra-striping/userscript.js
+++ b/addons/zebra-striping/userscript.js
@@ -19,7 +19,7 @@ export default async function ({ addon, msg, console }) {
         if (parent) {
           if (block.isShadow()) {
             isStriped = !!stripeState.get(parent);
-          } else if (parent.getCategory() === block.getCategory()) {
+          } else if (parent.getColour() === block.getColour()) {
             isStriped = !stripeState.get(parent);
           }
         }


WorldLanguages:
Does Blockly spell color as "colour"? Interesting :P

mxmou:
> Does Blockly spell color as "colour"? Interesting :P

It does, and I always forget that and write "color" and get errors.

Samq64 (approved):

GarboMuffin (approved):

CST1229:
Fun fact: The addon actually used to do this until I changed it (because Snap<i>!</i>, which I used as a reference, compares by categories, not colors).

mxmou:
One thing I didn't mention in the PR description is that this causes all nested blocks to be striped if the "black" preset for `editor-theme3` is used. I think that's better than the previous behavior.

WorldLanguages:
@mxmou Could you share a screenshot of the old behavior vs the new one? (when using the "black" preset)

mxmou:
Previous behavior:
image

After this PR:
image

PR #5588, opened 2023-01-09
editor-searchable-dropdowns: Preventing creating existing variables and trim names

GarboMuffin:
Previously the options would always be shown, but using them do the wrong thing. For example, when creating a global variable whose name conflicts with a local variable, it would seem like a variable was created, but in reality one was not.

Additionally, variable names will now always be trimmed. This matches Scratch's behavior.


apple502j (approved):

Samq64 (approved):

WorldLanguages:
> variable names will now always be trimmed

Do all addons with the ability of creating variables currently trim it too?

PR #5594, opened 2023-01-10
Fix "presets" label in the Addons popup

### LINKED ISSUE #5549: Fix position and layout of preset buttons
### Description

In the setting menu of the `dark-www` (website dark mode and customizable colors) addon, in the "presets" section, the "presets" heading is touching the first preset button, and there is no linebreak between the "presets" heading and the first preset.

In fullscreen settings, the presets are side-by-side, which is fine, but I would prefer them to be on top of each other and turn them into radio buttons (as we may add more and more presets).

### Steps to reproduce

1. Open the Scratch Addons extension popup.
2. Find the `dark-www` extension (I did this by typing dark-www into the addon search bar).
3. Click on the down caret icon to expand the addon and show its setting menu.
4. Find the `presets` section by scrolling down. 
5. Notice that the presets are positioned oddly, the **presets** label/heading is touching the first preset, and thus the presets do not align. See additional context.

Screenshot 2023-01-05 at 1 54 26 PM

^^^ There is no space between "presets" heading and the preset buttons.

I think this is because the presets are supposed to be side-by-side, but the extension popup is not wide enough, so the presets move down because there is not enough space. The first preset remains close to the "presets" label (which now acts like a heading) and even touches it. This results in the addon buttons being oddly positioned and the lack of spacing between the "presets" label and the actual preset buttons.

We should make the presets more responsive and go on top of each other with enough spacing.

### Environment

Browser: Chrome version 108.0.5359.124 (Official Build) (arm64)
Operating system: MacOS Ventura 13.0 on 14" 2021 MacBook Pro M1 pro

mxmou:
Resolves #5549

### Changes

Moves the "presets" label in the popup above the list of presets.

### Reason for changes

It was previously next to the first preset, with no spacing around it, which didn't look very good.

### Tests

image
image

### DIFF
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -251,11 +251,16 @@ body.iframe .addon-setting.boolean-setting .setting-label {
 }
 body.iframe .presets-column > .setting-label {
   margin-inline-start: 0;
+  margin-top: 10px;
   margin-bottom: 0;
+  width: 100%;
 }
 body.iframe .presets-column > .addon-setting {
   margin: 5px 0;
   margin-inline-end: 10px;
+}
+body.iframe .presets-column > .addon-setting:last-child {
+  margin-inline-end: 0;
 }
 
 body.iframe .addons-container {


apple502j (approved):

penguinmoose:
Thanks, mxmou!

PR #5602, opened 2023-01-11
Add message indicator color option for website dark mode and customizable colors

actuallyronan:
### Changes

Adds a color setting to the "Website Dark Mode and Customizable Colors" add-on that lets you select the color for the message indicator.

### Reason for changes

Many people (including myself) would prefer to be able to change the color of the message indicator, as it is a key part of the Scratch website.

### Tests

Tested on Chrome 108 and it seems to work.


addons/dark-www/experimental_scratchr2.css
```diff
@@ -671,3 +671,7 @@ input.link.black {
 .iframeshim {
   color-scheme: light;
 }
+
+.message-count, .notificationsCount {
+  background-color: var(--darkWww-message-indicator-color) !important; 
+}
```
mxmou, line 677:
```suggestion
#topnav ul.account-nav .messages .notificationsCount {
  background-color: var(--darkWww-message-indicator-color); 
}
```

addons/dark-www/experimental_scratchwww.css
```diff
@@ -862,3 +862,7 @@ p {
   /* Change the solid color background to a transparent one */
   background-color: rgba(0, 165, 0, 0.07);
 }
+
+.message-count, .notificationsCount {
+  background-color: var(--darkWww-message-indicator-color) !important; 
+}
```
mxmou, line 868:
```suggestion
#navigation .messages .message-count.show {
  background-color: var(--darkWww-message-indicator-color); 
}
```

mxmou (changes requested):
You don't need to define a `customCssVariable` for this - one will be created automatically based on the setting name (`--darkWww-messageIndicatorColor` in this case).

Samq64:
Could you please enable actions? (https://github.com/actuallyronan/ScratchAddons/actions)

DNin01:
> Could you please prettier on the actions page? (https://github.com/actuallyronan/ScratchAddons/actions)

I think it's not that Prettier isn't working, it's that it's not turning tab characters into two spaces.

@actuallyronan could you change all the tab characters to two spaces?

actuallyronan:
> > Could you please prettier on the actions page? (https://github.com/actuallyronan/ScratchAddons/actions)
> 
> I think it's not that Prettier isn't working, it's that it's not turning tab characters into two spaces. The CGs already recommend enabling workflows, so they probably already enabled it.
> 
> @actuallyronan could you change all the tab characters to two spaces?

Done

addons/dark-www/addon.json
```diff
@@ -1,2107 +1,2183 @@
...
+      "allowTransparency":true,
+      "default":"#606060"
     },
     {
-      "name": "Border color",
-      "id": "border",
-      "type": "color",
-      "allowTransparency": true,
-      "default": "#606060"
+      "name":"Message indicator color",
```
DNin01, line 2081: The messages "indicator" is also known as the messages badge.

DNin01 (commented):
> Done

Thanks!

Also, sorry to bug you _again_, but another formatting thing in the `addon.json` is that there is usually a space after the `:` right after each key (example: `"key": "value"`), so maybe change that as well, to keep the formatting consistent with other files.

mxmou:
@actuallyronan Formatting is now fixed.

DNin01 (changes requested):
Four things:
1. You ~~need to~~ might want to add the default message badge color to the other presets (look under `"presets": [`).
2. This doesn't apply to the messages page itself.[^1]
3. If you set the badge color to a bright color, the number on it should turn black.
4. This is generally known as the messages badge.

[^1]: See image: 
An orange messages badge on the Messages page

mxmou:
> You need to add the default message badge color to the other presets (look under `"presets": [`).

This isn't actually necessary, but it's probably a good idea - I think it makes sense to reset the color when selecting a preset.

Samq64:
Are there any other orange/yellow elements on the site that could be customized with this setting?

DNin01:
> Are there any other orange/yellow elements on the site that could be customized with this setting?

The Share button.

DNin01:
> > You need to add the default message badge color to the other presets (look under `"presets": [`).
> 
> This isn't actually necessary, but it's probably a good idea - I think it makes sense to reset the color when selecting a preset.

The reset menu for that color will look like this, though.

The preset color choices menu with transparent backgrounds from the presets

Clicking on one of them, however, still resets the color to the default, not a transparent background as shown.

mxmou:
> > > You need to add the default message badge color to the other presets (look under `"presets": [`).
> > 
> > 
> > This isn't actually necessary, but it's probably a good idea - I think it makes sense to reset the color when selecting a preset.
> 
> The reset menu for that color will look like this, though.
> 
> The preset color choices menu with transparent backgrounds from the presets
> 
> Clicking on one of them, however, still resets the color to the default, not a transparent background as shown.

That's a bug with the settings page.

WorldLanguages:
> That's a bug with the settings page.

Could you create a GitHub issue about it? (I think it doesn't exist yet)

mxmou:
@WorldLanguages I'll just open a PR to fix it - it's a very simple change.

mxmou:
If a light color is chosen, the message count should become black. You can use `customCssVariables` to implement this.

WorldLanguages:
@mxmou We should probably move this to the v1.32 milestone?

Secret-chest:
What about the Explore orange?

mxmou:
> What about the Explore orange?

That should probably be in a separate addon for customizing banner colors, similar to #4187.

WorldLanguages:
What's the status of this PR?

mxmou:
> What's the status of this PR?

This feature is still missing:

> If a light color is chosen, the message count should become black. You can use `customCssVariables` to implement this.

DNin01:
Scratch changed the message counter background in the update. Do we need to worry about making it customizable anymore? If not, I'll close this.

DNin01:
Well, there is the fact that some people might not like it to be dark red.

This PR isn't being actively maintained right now, so maybe we can start a new one if anyone is interested.

WorldLanguages:
> Well, there is the fact that some people might not like it to be dark red.

There were one of two people sending feedback about this. Maybe it could be part of this addon.

mxmou:
> > Well, there is the fact that some people might not like it to be dark red.
> 
> There were one of two people sending feedback about this. Maybe it could be part of this addon.

The option this PR currently adds changes both counters - the one in the navigation bar and the one on the Messages page. It should probably be split into two.

DNin01:
> > > Well, there is the fact that some people might not like it to be dark red.
> 
> > 
> 
> > There were one of two people sending feedback about this. Maybe it could be part of this addon.
> 
> 
> 
> The option this PR currently adds changes both counters - the one in the navigation bar and the one on the Messages page. It should probably be split into two.

That would work, but it feels like too much. I wonder how we could implement this while only adding one color to the settings.

WorldLanguages:
> I don't think there's a (good) way to make both colors customizable with just one setting.

I agree. It would be nice if we could provide a `Scratch default colors (blue)` preset (or maybe another name) for this addon.
@mxmou It's up to you if you want to work on this PR or create a new one

DNin01:
I don't feel like customizing the badge above the list of messages counting the number of unread messages is as important. If we were to go with a one-setting method, I'd allow customizing the badge on the navigation bar and not the one on the messages page.

WorldLanguages:
From feedback:

> you forgot the option to change the message notifier from a putrid orange red to any color

mxmou:
> I don't feel like customizing the badge above the list of messages counting the number of unread messages is as important. If we were to go with a one-setting method, I'd allow customizing the badge on the navigation bar and not the one on the messages page.

People want to change the one on the messages page back to orange though.

mxmou (approved):

WorldLanguages:
Looks good!
Does this take into account that the scratchr2 message button orange is a slightly different shade? (Not saying it should, but I've checked whether they have the same exact hex code). Looks like it does. Nice!

WorldLanguages:
I guess this will need a settings migration for users who already selected these presets (specifically, the ones with `(blue)` in the name). I'd say we create an issue after merging this and deal with it later.

WorldLanguages:
(We should also add a latestUpdate so that users notice this update)

DNin01:
Alright, this seems like the best we can do, and that's fine.

WorldLanguages (approved):

PR #5609, opened 2023-01-12
Scroll project titles

actuallyronan:
Suggested in the SA Community Discord server: https://discord.com/channels/806602307750985799/1062485661959082055

### Changes

This allows you to scroll through the title of projects that you don't own, that way you can view the entire title, and not have it get cut off. I also made sure that the title resets to its original position when the add-on is disabled.

### Reason for changes

You should be able to read the full title of all projects, not just ones that you own. There are a lot of longer titles or titles that get cut off due to little space in the UI.

### Tests

Tested on Chrome 109

Screenshot from the Discord server where it was originally suggested


Samq64 (changes requested):
Could you add some tags to the addon? Without the community or editor tag the addon can't be enabled.

addons/scroll-project-titles/style.css
```diff
@@ -0,0 +1,4 @@
+.project-title.no-edit {
+  overflow-x: scroll !important;
+  text-overflow: clip !important;
```
Samq64, line 3: I don't think `!important` necessary on thse two lines.

addons/scroll-project-titles/addon.json
```diff
@@ -1,28 +1,22 @@
...
   ],
-  "userscripts": [
-    {
-      "url": "script.js",
-      "matches": ["projects"],
-      "runAtComplete": false
-    }
-  ],
-  "versionAdded": "1.31.0",
-  "tags": ["community", "projectPage"],
```
DNin01: It looks like you accidentally removed the tags in this change.

actuallyronan: Thanks, I was wondering what happened there.

addons/scroll-project-titles/style.css
actuallyronan: Thanks, fixed it.

actuallyronan:
@Samq64 added the tags 👍

addons/scroll-project-titles/style.css
```diff
@@ -0,0 +1,4 @@
+.project-title.no-edit {
```
Samq64, line 1: Scratch's selector is more specific, so yours will need to be the same to override the styles.
```suggestion
.preview .project-title.no-edit {
```

addons/scroll-project-titles/script.js
```diff
@@ -0,0 +1,7 @@
+export default async function ({ addon, console, msg }) {
+  addon.self.addEventListener("disabled", function() {
+    if (document.querySelector(".project-title.no-edit")) {
```
Samq64, line 3: This check is probably unnecessary, especially since it only runs when dynamically disabling the addon.

Samq64 (changes requested):
I think you forgot to add your userscript to `addon.json`.

addons/scroll-project-titles/script.js
actuallyronan: It can still prevent errors, and it also makes sure that it doesn't affect titles that you can edit. It's important to reset the scroll, that way you don't get all the project titles stuck halfway through scrolling without any way of resetting it.

Secret-chest:
What about a marquee effect?

mxmou:
> What about a marquee effect?

That would be annoying.

BroJac5246:
> What about a marquee effect?

_Maybe_ on hover? But at that point you might as well manually scroll...

Joeclinton1:
> Should this be part of full areas instead?

Agree

Samq64:
#7720 @actuallyronan I can add you to the credits if you want.

DNin01:
@actuallyronan, are you @rgantzos?

WorldLanguages:
Replaced by PR #7720

PR #5610, opened 2023-01-12
Fix GitHub action stale-issues

WorldLanguages:
Resolves error logs from https://github.com/ScratchAddons/ScratchAddons/actions/runs/3897982638

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -20,10 +20,10 @@ jobs:
         with:
           stale-issue-message: 'Ping! There has been no activity for 7 days.'
           stale-pr-message: 'Ping! There has been no activity for 7 days.'
-          only-label: 'status: awaiting answer/followup'
+          only-labels: 'status: awaiting answer/followup'
           days-before-stale: 7
           days-before-close: -1
-          stale-issue-label: ''
-          stale-pr-label: ''
+          stale-issue-label: 'Stale'
+          stale-pr-label: 'Stale'
           operations-per-run: 30
           repo-token: ${{ steps.generate-token.outputs.token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-issues.yml
```diff
@@ -20,10 +20,10 @@ jobs:
...
           stale-issue-message: 'Ping! There has been no activity for 7 days.'
           stale-pr-message: 'Ping! There has been no activity for 7 days.'
-          only-label: 'status: awaiting answer/followup'
+          only-labels: 'status: awaiting answer/followup'
           days-before-stale: 7
           days-before-close: -1
-          stale-issue-label: ''
-          stale-pr-label: ''
+          stale-issue-label: 'Stale'
+          stale-pr-label: 'Stale'
```
Hans5958, line 27: I just made a proper label in which you should use.
```suggestion
          stale-issue-label: 'status: stale'
          stale-pr-label: 'status: stale'
```

Hans5958 (changes requested):

.github/workflows/stale-issues.yml
WorldLanguages: I'm not sure what the behavior of the action is when the label doesn't exist.
I would prefer if no tag was added.

Hans5958: Based on my knowledge, if there is no such label, it would just create the label beforehand, by force. I could delete the `status: stale` label if you want to test this hypothesis, but I would still like to keep it as a fallback, if what I think is right.

WorldLanguages: Sure, let's keep it

Hans5958 (approved):
LGTM. Thanks for the edit.

apple502j (approved):

PR #5613, opened 2023-01-13
`user-id`: Option to show only on hover

### LINKED ISSUE #5329: `user-id`: Option to only show user IDs when hovering over username

What if someone wants to have the option to quickly see anyone's user ID but doesn't want the `#12345678` thing to always appear next to every username on profile pages?

### What this change does

The addon would have a new option to show user IDs only when hovering over the username on a profile page. We could call it "only show when hovering over username".

### Possible alternatives

We could also use a tooltip or put it in a less distracting place.

_No response_

DNin01:
Resolves #5329

### Changes

Added an option to the `user-id` addon that makes the user ID only appear when you're hovering over the username.

### Reason for changes

Having the user ID next to the username is useful whenever you want to use it, but what if you don't always like it to show up?

### Tests

Tested on Edge 108 with SA v1.30.0-pre and it works.

### DIFF
--- a/addons/user-id/addon.json
+++ b/addons/user-id/addon.json
@@ -5,6 +5,10 @@
     {
       "name": "yes-i-am-lanky",
       "link": "https://scratch.mit.edu/users/yes-i-am-lanky/"
+    },
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
     }
   ],
   "userscripts": [
@@ -17,11 +21,32 @@
     {
       "url": "userstyle.css",
       "matches": ["profiles"]
+    },
+    {
+      "url": "hover.css",
+      "matches": ["profiles"],
+      "if": {
+        "settings": { "onlyonhover": true }
+      }
+    }
+  ],
+  "settings": [
+    {
+      "id": "onlyonhover",
+      "name": "Only show when hovering over username",
+      "default": false,
+      "type": "boolean"
     }
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
+  "injectAsStyleElt": true,
+  "updateUserstylesOnSettingsChange": true,
   "tags": ["community", "profiles"],
   "versionAdded": "1.19.0",
+  "latestUpdate": {
+    "version": "1.31.0",
+    "newSettings": ["onlyonhover"]
+  },
   "enabledByDefault": false
 }
--- a/addons/user-id/addon.json
+++ b/addons/user-id/addon.json
@@ -40,7 +40,6 @@
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "injectAsStyleElt": true,
   "updateUserstylesOnSettingsChange": true,
   "tags": ["community", "profiles"],
   "versionAdded": "1.19.0",
--- a/addons/user-id/addon.json
+++ b/addons/user-id/addon.json
@@ -44,7 +44,7 @@
   "tags": ["community", "profiles"],
   "versionAdded": "1.19.0",
   "latestUpdate": {
-    "version": "1.31.0",
+    "version": "1.32.0",
     "newSettings": ["onlyonhover"]
   },
   "enabledByDefault": false
--- a/addons/user-id/addon.json
+++ b/addons/user-id/addon.json
@@ -26,13 +26,13 @@
       "url": "hover.css",
       "matches": ["profiles"],
       "if": {
-        "settings": { "onlyonhover": true }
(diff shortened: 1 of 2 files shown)


Hans5958:
Pardon regarding my knowledge of this addon, but why setting the opacity when you can just `display: none`?

DNin01:
> Pardon if I missed something regarding this addon, but why setting the opacity when you can just `display: none`?

I chose to set the opacity over hiding it completely for some reason the first time I made this feature. It was probably so that it could detect hovering, but now one of the parent elements does that. I'll see later if I can change this.

WorldLanguages:
Using opacity is probably unnecessary, unless you want to make an animation

DNin01:
Specifically, anytime I add `display: none` or `display: block`, it gets canceled out by `display: var(--userId-_displayNoneWhileDisabledValue);`

WorldLanguages:
@DNin01 If you want to force an element that you passed to displayNoneWhileDisabled() to hide, you must use `!important`. See example: https://scratchaddons.com/docs/reference/addon-api/addon.tab/#using-addontabdisplaynonewhiledisabled-dynamicdisable

DNin01:
I'm still not exactly sure what I need to do make it optionally show on hover _and_ go away properly when you disable the addon. It does one or the other. Unless there's some way to make it show but hide when NOT hovering over it, not hide but show when hovering over it.

WorldLanguages:
@DNin01 That's confusing. Maybe you could make a table or diagram with all the possible states? Isn't the logic simply "hide except if hovered and addon enabled"?

DNin01:
Yeah, I was a little confusing. Does the following make more sense?
```
if addon enabled {
  if hovering over element {
    show
  } else {
    hide
  }
} else {
  hide
}
```

Or...
```
// We show it only if the addon is on
if addon enabled {
  show
} else {
  hide
}

// then, whether it is being shown or not, we check if the user is hovering over it, and if they aren't, we hide it.
if NOT hovering over element {
  hide
}
```

mxmou:
> I'm still not exactly sure what I need to do make it optionally show on hover _and_ go away properly when you disable the addon. It does one or the other. Unless there's some way to make it show but hide when NOT hovering over it, not hide but show when hovering over it.

Would the following work?
```css
.header-text > h2:not(:hover) > .sa-user-id {
  display: none !important;
}
```

DNin01:
> ```
> h2:not(:hover)
> ```

Aha!

Samq64 (approved):
I thought it would be a tooltip, but this works too.

DNin01:
> I thought it would be a tooltip, but this works too.

It makes it possible to copy the user ID.

addons/user-id/addon.json
```diff
@@ -17,11 +21,32 @@
...
+    {
+      "id": "onlyonhover",
+      "name": "Only show when hovering over username",
+      "default": false,
+      "type": "boolean"
     }
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
+  "injectAsStyleElt": true,
```
WorldLanguages, line 43: injectAsStyleElt isn't really necessary

WorldLanguages:
Tested and implementation looks okay

@mxmou Any comments? Do we have any better way to only show the user ID on hover?
I've always had mixed opinions about the reasoning for this addon to exist, as well as the way we decided to style it. The bold numbers look out of place.
About this specific PR, maybe it would feel snappier if we added a short CSS transition when hovering?

mxmou:
> Do we have any better way to only show the user ID on hover?

I don't think so.

> I've always had mixed opinions about the reasoning for this addon to exist, as well as the way we decided to style it. The bold numbers look out of place.

I don't know what the reason for this addon to exist is - it doesn't seem very useful to me. The only thing about the styling I would change is add a bit more spacing between the username and the ID.

> About this specific PR, maybe it would feel snappier if we added a short CSS transition when hovering?

I think it's more important to be consistent with Scratch. The actions that appear when hovering a comment don't have a transition.

TheColaber:
> I'm still not exactly sure what I need to do make it optionally show on hover _and_ go away properly when you disable the addon. It does one or the other. Unless there's some way to make it show but hide when NOT hovering over it, not hide but show when hovering over it.

This would probably be fixed with that PR I made that reworked it. https://github.com/ScratchAddons/ScratchAddons/pull/4924 GarboMuffin also linked to a lot of commits that would not have been a problem in the first place if that pr was merged. I recommend taking a look.

WorldLanguages (approved):

PR #5603, opened 2023-01-11
Update README

WorldLanguages:
Changes:
- Changed header from `Welcome to Scratch Addons' repository` to `Scratch Addons browser extension`
- Add Edge version and user count badges. Remove some unnecessary badges that repeat information that is already on the page, such as the star count. Chrome badges are now red, and Edge badges are blue.
- Use new mission statement for the "about" section. Is now consistent with the org README.
- Change `Google Chrome/Microsoft Edge` to `Google Chrome / Microsoft Edge` (added spaces to both sides of the slash)
- Fixed broken link to contributing guidelines

TODOs:
- [ ] Provide a link to the org README somewhere on this repo README, which has more information
- [ ] Link to the docs instead of explaining what addons are inside the README. Some basic explanation can stay, but surely not the current one, which hasn't been updated for a while
- [ ] Link to https://scratchaddons.com/docs/getting-started/installing/ in the "install" section. This page has additional information, such as browser support.
- [ ] Provide a link to the security policy
- [ ] Provide a link to the "how to join Transifex" page regarding translation
- [ ] Explain what each folder's purpose is
- [ ] Mention VSCode is the recommended editor
- [ ] The "from source" section has a list of alternatives that looks like a list of steps, that's confusing
- [ ] Rewrite the "contributing" section
- [ ] Spellcheck whole README and attempt to use formal English

PR #5617, opened 2023-01-14
Fix broken links

RedGuy12:
Regression from #5519


WorldLanguages:
Note: the README broken link fix is also part of #5603

apple502j (approved):

PR #5619, opened 2023-01-15
IsOnline 2: The Search For More Money

GrahamSH-LLK:
### Changes

adds a super duper cool isonline addon
### Reason for changes

because who _doesn't_ like isonline
### Tests

duh
### Technical details

I have a small node server running in a docker container on a beefy oracle VM (free tier)
It is attached to a redis database
On every post to `/ping` with a body, it sets a key for each user to `true`, and sets it to expire in 5 minutes
`/status/${username}` responds with `{"online": true}` or false if no such record exists
It's very (very) fast due to the nature of redis


addons/isonline/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "IsOnline",
+  "description": "Adds a button to the bottom-right of the project page to set the thumbnail of your project to any image (including gifs).",
```
fractioneater, line 3: I don't think this description is correct.

Also, does this have an effect on the project page? If not, you should remove the `projectPage` tag.

Samq64: It's just the description and tags from the thumbnail setter addon. This should probbably be a draft.

GrahamSH-LLK: yeah that was me copying an existing addon.json because i was lazy

GrahamSH-LLK:
This is waiting on:
- Figuring out compatibility with our privacy policy
- Making a better online indicator
- Showing it on pages other than profile pages (or maybe this can wait?)
- Having somebody other than me look over the setup and code for the server, and most likely giving somebody (WL, maybe garbo) access to the server
- Showing in the UI clearly that this addon regularly pings a server

PR #5624, opened 2023-01-15
Fix condition for showing reset dropdown

mxmou:
### Changes

Fixes the condition for showing a reset dropdown next to a setting.

### Reason for changes

The dropdown should be hidden if the setting isn't changed by any preset.

### Tests

Tested by copying the `addon.json` change from #5602.

### DIFF
--- a/webpages/settings/components/addon-setting.js
+++ b/webpages/settings/components/addon-setting.js
@@ -46,10 +46,12 @@ export default async function ({ template }) {
         return (
           !this.tableChild &&
           this.addon.presets &&
-          this.addon.presets.some((preset) =>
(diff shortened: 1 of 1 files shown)


Samq64:
What if the value is changed by some presets, but not others?

DNin01:
> What if the value is changed by some presets, but not others?

Would that ever happen anyway?

Samq64:
> > What if the value is changed by some presets, but not others?
> 
> Would that ever happen anyway?

In most cases probbably not, so it's not really a big deal.

apple502j (approved):

Samq64 (approved):

WorldLanguages:
> What if the value is changed by some presets, but not others?

Ideally that shouldn't be the case.

PR #5634, opened 2023-01-18
Rename "Mute" to "Do not disturb"

### LINKED ISSUE #4955: Call the "Mute" option something else
This:
Right clicking the extension icon shows an option called Mute

"Mute" could mean:
- Turn off the sound
- Stop notifications
- Turn off the addons

It's slightly vague. Not too bad, but some percentage of users would get confused.

WorldLanguages said:
> It hides the message count, and disables notifications from the "Scratch Notifier" feature

Based on that, a more fitting, self-explanatory name might be:
- Do Not Disturb, or
- Disable Notifications

DNin01:
Resolves #4955

### Changes

When right clicking the Scratch Addons extension icon, there was a Mute option, which silenced and hid notifications from the Scratch Notifier addon and hid the messages badge. I renamed this setting to "Do not disturb" and added a note to notification-related addons to tell the user that enabling this feature will hide the message count.

### Reason for changes

It described the feature better.

### Tests

Changes tested.

### DIFF
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -7,7 +7,7 @@
   "info": [
     {
       "type": "info",
-      "text": "You can use Do Not Disturb (right-click extension icon \u2192 Do not disturb) to hide the message count.",
+      "text": "You can use Do Not Disturb (right click extension icon \u2192 Do not disturb) to temporarily hide the message count.",
       "id": "dndtohide"
     }
   ],
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -4,6 +4,13 @@
   "tags": ["popup", "recommended"],
   "enabledByDefault": true,
   "versionAdded": "1.0.0",
+  "info": [
+    {
+      "type": "notice",
+      "text": "Turning on Do Not Disturb (right-click extension icon \u2192 Do not disturb) will hide the message count.",
+      "id": "dndtohide"
+    }
+  ],
   "settings": [
     {
       "name": "Badge color",
--- a/addons/msg-count-badge/addon.json
+++ b/addons/msg-count-badge/addon.json
@@ -6,8 +6,8 @@
   "versionAdded": "1.0.0",
   "info": [
     {
-      "type": "notice",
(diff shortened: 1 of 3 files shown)


WorldLanguages:
I think I'm indifferent about this

addons/msg-count-badge/addon.json
```diff
@@ -4,6 +4,13 @@
   "tags": ["popup", "recommended"],
   "enabledByDefault": true,
   "versionAdded": "1.0.0",
+  "info": [
+    {
+      "type": "notice",
+      "text": "Turning on Do Not Disturb (right-click extension icon \u2192 Do not disturb) will hide the message count.",
```
WorldLanguages, line 10: Maybe this notice should also be added to scratch-notifier

WorldLanguages: So users won't get confused over why they're not receiving notifications

DNin01: I think that's already implied by Do not disturb, but we could add a notice anyway to inform the user that they can hide the notifications temporarily.

BroJac5246: > I think that's already implied by Do not disturb, but we could add a notice anyway to inform the user that they can hide the notifications temporarily.

I agree, as I wasn't even aware "Mute" existed before we decided to rename it.

WorldLanguages:
@DNin01 Are my changes okay? I changed `right-click` because it's inconsistent with other addon descriptions. Other changes I made were opinionated.

DNin01:
> @DNin01 Are my changes okay? I changed `right-click` because it's inconsistent with other addon descriptions. Other changes I made were opinionated.

Looks good to me 👍

addons/msg-count-badge/addon.json
```diff
@@ -4,6 +4,13 @@
   "tags": ["popup", "recommended"],
   "enabledByDefault": true,
   "versionAdded": "1.0.0",
+  "info": [
+    {
+      "type": "info",
```
Samq64, line 9: I think the type defaults to info, so this is unnecessary.

Samq64 (approved):

Hans5958 (commented):
You have my weak approval. DND is a usual term, but let's see how others think.

WorldLanguages:
No one is explicitly against this change, so I'm merging it.

Issue #2057, opened 2021-04-02
mediarecorder: add setting to record monitors and cursor (using window.CropTarget)


Scratchfangs:
Recording won't record lists and variables. A hunch of mine is because they are outside the canvas but I am wondering if there is a fix.

WorldLanguages:
Probably will never happen, unless I'm missing something - there's no API to record an area of the page, only to record the canvas.
One "workaround" could be using the screen sharing feature, but that would also show the cursor, and the result wouldn't be a 4:3 video.

mxmou:
@WorldLanguages Would it be possible to create another, invisible canvas and render the content of the canvas and the monitors onto it every frame?

apple502j:
Too laggy IMO.

mxmou:
> Too laggy IMO.

Could it be optional?

WorldLanguages:
@mxmou Not sure how you could accurately render HTML/CSS based monitors into a canvas.

mxmou:
> @mxmou Not sure how you could accurately render HTML/CSS based monitors into a canvas.

I think it's possible to embed them inside a SVG and render it.

TecFuture:
What about rendering variables? 
There is this option called "variables" as a folder which you can view variables data. 
We can render the variables data (along with how it like like slider) into the video.

One question
Can you get the data if variable is slider,large readout, normal variable (Such as API, or a quick CSS/HTML check)?

WorldLanguages:
> there's no API to record an area of the page, only to record the canvas

This is outdated now!! Good news

https://developer.chrome.com/blog/new-in-chrome-104/#region-capture

apple502j:
Rejected for now, due to compatibility concerns.

mxmou:
> Rejected for now, due to compatibility concerns.

Couldn't the canvas hack be implemented as a temporary solution?

WorldLanguages:
See https://region-capture-demo.glitch.me/ for a demo.
Still not supported on Firefox.

PR #5638, opened 2023-01-18
Minor text updates to recorder & remove find bar label

iqnite:
### Changes
Find bar: Removes the useless "Find" label next to the bar.
Project video recorder: Changed button description to "Record" and added a warning that says that variables are not supported.

### Reason for changes
"Start recording" doesn't start a recording but opens a popup. "Record" better matches the button's purpose and needs less space in the menu bar.
The "Find" label next to the editor find bar was pretty useless because it is already in the bar itself, so it only took up space and looked weird.

### Tests
All changes were tested and work correctly, as they're only changes to the text.


iqnite:
Done   Von: Samq64Gesendet: Mittwoch, 18. Januar 2023 14:37An: ScratchAddons/ScratchAddonsCc: IQBit; AuthorBetreff: Re: [ScratchAddons/ScratchAddons] Minor text updates to find bar and recorder (PR ***@***.*** commented on this pull request.In addons-l10n/en/find-bar.json:> @@ -1,5 +1,5 @@ {-  "find-bar/find": "Find",+  "find-bar/find": "",Could you remove this line and and also remove it from the JS file?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

addons-l10n/en/find-bar.json
```diff
@@ -1,5 +1,5 @@
 {
-  "find-bar/find": "Find",
+  "find-bar/find": "",
```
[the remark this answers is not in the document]
iqnite: Done

Samq64 (changes requested):
The label itself still exists and falls back to some placeholder text since the message doesn't exist. Could you remove the actual label by deleting https://github.com/iqnite/ScratchAddons/blob/54128ca6d81c983e881d148e028352ed7a8c9b21/addons/find-bar/userscript.js#L41 and the two lines below it?

Samq64 (changes requested):
It turns out [this line](https://github.com/iqnite/ScratchAddons/blob/d77bc021e8c2b7b8be007a9b3b88b5c417ac98ae/addons/find-bar/userscript.js#L215) still uses the element so that part of it can be removed along with `this.findLabel = null;` and the label styles. The find bar should also to be shifted to the right a little.

iqnite:
> It turns out [this line](https://github.com/iqnite/ScratchAddons/blob/d77bc021e8c2b7b8be007a9b3b88b5c417ac98ae/addons/find-bar/userscript.js#L215) still uses the element so that part of it can be removed along with `this.findLabel = null;` and the label styles. The find bar should also to be shifted to the right a little. Everything else looks good to me.

Fixed. Anything else?

mxmou:
I think what @Samq64 meant is that `this.findLabel = null` and other references to `findLabel` should be removed, not added.

iqnite:
> I think what @Samq64 meant is that `this.findLabel = null` and other references to `findLabel` should be removed, not added.

Done in 68da1fc1e7a6f4162bccebb677e65c086ddb7903

Samq64 (commented):
All the styles for [`.sa-find-bar > label`](https://github.com/iqnite/ScratchAddons/blob/7ab8d89047160fff76bead3efbaf260cad6c7d13/addons/find-bar/userstyle.css#L16) and [`[dir="rtl"] .sa-find-bar > label`](https://github.com/iqnite/ScratchAddons/blob/7ab8d89047160fff76bead3efbaf260cad6c7d13/addons/find-bar/userstyle.css#L28) are also unused now, sonthey can be removed.
As I mentioned before, the find bar looks like it needs to be shifted a little to the right, so try adding `margin-left: 1em;` to [`.sa-find-bar`](https://github.com/iqnite/ScratchAddons/blob/7ab8d89047160fff76bead3efbaf260cad6c7d13/addons/find-bar/userstyle.css#L3) and see if it looks better.

Samq64 (approved):

DNin01 (approved):

addons-l10n/en/mediarecorder.json
```diff
@@ -10,7 +10,7 @@
   "mediarecorder/record-audio-description": "This does not include Text-to-Speech.",
   "mediarecorder/record-after-flag": "Do not start recording until the Green Flag is clicked",
   "mediarecorder/record-until-stop": "Stop recording after the project has stopped",
-  "mediarecorder/record-description": "Record the stage as a WebM file. You can save it to your computer after the recording is finished.",
+  "mediarecorder/record-description": "Record the stage as a WebM file. You can save it to your computer after the recording is finished.\nNote: stage monitors and lists will not be visible.",
```
WorldLanguages, line 13: This new sentence is confusing - aren't lists also monitors?

iqnite: Many people don't know it.

Samq64:
```suggestion
  "mediarecorder/record-description": "Record the stage as a WebM file. You can save it to your computer after the recording is finished.\nNote: Variable and list monitors will not be visible.",
```

WorldLanguages:
Griffpatch accepted the find bar change.

WorldLanguages:
This can be merged now (I didn't test the find bar myself)

By the way, it would now be possible to record monitors in Chromium. See https://github.com/ScratchAddons/ScratchAddons/issues/2057#issuecomment-1249835176

PR #5653, opened 2023-01-20
Replace "What I've been doing" with the user's statistics.

Dr-Lego:
### Replace "What I've been doing" with the user's statistics

- Data from ScratchDB and the Scratch Website
- Displays exact follower count (fetched from "scratch.mit.edu/users/[username]/followers")
- You can still switch beetween the Statistics and WIBD.
- Lets you instantly see the follower count, rank, etc. when you visit someone's profile page
- Adds also a link to the user's ScratchStats page
- Looks nice in dark and normal mode

### How it looks like:
image

image

image


Dr-Lego:
It says I should add "VersionAdded" to the addon manifest, but I'm not supposed to do that, I guess?

addons/show-user-stats/addon.json
```diff
@@ -0,0 +1,18 @@
...
+        "link": "https://scratch.mit.edu/users/Dr_Lego/"
+      }
+    ],
+    "userscripts": [
+      {
+        "url": "userscript.js",
+        "matches": ["https://scratch.mit.edu/users/*"]
+      }
+    ],
+    "tags": ["community"],
```
DNin01, line 16:
```suggestion
    "tags": ["community"],
    "dynamicEnable": true,
    "dynamicDisable": true,
```

DNin01 (changes requested):
> It says I should add "VersionAdded" to the addon manifest, but I'm not supposed to do that, I guess?

Add `"versionAdded": "1.31.0"` - that's the upcoming version. You can do this later if you want.

Also, if possible, make sure the addon supports [dynamic enable/disable](https://scratchaddons.com/docs/reference/addon-api/addon.self/#using-addonselfdisabled-and-events-dynamicdisable), meaning that the addon's effects disappear when disabled and appear when enabled, all without refreshing the page.

mxmou:
There's already an addon that displays statistics on profiles. It should probably be combined with this one.

BroJac5246:
Can there be a setting to show WIBD by default?

DNin01:
In addition to adding dynamicEnable and dynamicDisable to the addon manifest, I believe you might need to make some changes to the userscript to make it automatically revert the layout when you disable the addon and automatically apply when you enable it, without even refreshing the page, unless it already does that.

WorldLanguages:
> There's already an addon that displays statistics on profiles. It should probably be combined with this one.

Yeah, I'd prefer to see this as a setting for the `Profile statistics` addon.

WorldLanguages:
Does this provide any additional functionality not currently part of "profile statistics"?
Is the only difference that this isn't a new row, but instead uses the "What I've been doing" section? I just created an issue for that: #5834

WorldLanguages:
Closing PR because it creates a new addon with the same goal as an existing one.

Issue #5834 was created because some users might want a setting for the different positioning that this PR suggested.

PR #5684, opened 2023-02-02
Link changelog link directly to version

Hans5958:
### Changes

Do adjustments to the changelog link so it goes directly to the version, instead stuck on the top.

### Reason for changes

So the link would stick to the current extension version that is used, even if it is not latest.

### Tests

Negligble to be tested, but I haven't done it.

### DIFF
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -56,7 +56,7 @@ const vue = new Vue({
       const uiLanguage = chrome.i18n.getUILanguage();
       const localeSlash = uiLanguage.startsWith("en") ? "" : `${uiLanguage.split("-")[0]}/`;
       const utm = `utm_source=extension&utm_medium=popup&utm_campaign=v${chrome.runtime.getManifest().version}`;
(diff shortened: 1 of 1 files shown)


webpages/popup/index.js
```diff
@@ -56,7 +56,7 @@ const vue = new Vue({
       const uiLanguage = chrome.i18n.getUILanguage();
       const localeSlash = uiLanguage.startsWith("en") ? "" : `${uiLanguage.split("-")[0]}/`;
       const utm = `utm_source=extension&utm_medium=popup&utm_campaign=v${chrome.runtime.getManifest().version}`;
-      return `https://scratchaddons.com/${localeSlash}changelog/?${utm}`;
+      return `https://scratchaddons.com/${localeSlash}changelog/#v${chrome.runtime.getManifest().version}?${utm}`;
```
WorldLanguages, line 59: I think the query params go before the hash

WorldLanguages (approved):
Tested successfully

PR #5696, opened 2023-02-05
Use different fix for sprite pane glitch

DNin01:
Fixes a bug from `scratch-gui`: https://github.com/LLK/scratch-gui/issues/5471.

### Changes

The existing Disable sprite wobble addon fixed the above issue by preventing the sprites from wobbling when dragging items over them, since that would usually cause a scrollbar to appear that interferes with the wrapping and makes the sprite cards in the sprite pane jump/glitch around.

Now, my "rebuild" of this addon just adds `overflow-y: hidden` to some part of the sprite pane (as [mxmou said](https://github.com/LLK/scratch-gui/issues/5471#issuecomment-1345238522)) that fixes this without removing any animations.

There's still a setting to disable the wobble as before.

### Reason for changes

It's a more practical solution.

### Tests

- [x] Test on Chromium-based browser
- [x] Test on Firefox


addons/disable-sprite-wobble/addon.json
```diff
@@ -1,14 +1,10 @@
...
     {
-      "name": "Xan",
-      "link": "https://scratch.mit.edu/users/-Xanimation-/"
-    },
-    {
-      "name": "CST1229",
-      "link": "https://github.com/cst1229/"
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
     }
```
DNin01, line 8: I removed the previous contributors from this addon's credits because I completely redid the code (as if this was a new addon). Is that okay?

BTW, I also used [mxmou's code](https://github.com/LLK/scratch-gui/issues/5471#issuecomment-1345238522).

BroJac5246: I think I'm leaning towards leaving them, since they still had an effect on the addon.

pufferfish101007:
Was the point of the original addon not that the animations were annoying and not just that it fixed the scrollbar bug?
IMO there should maybe be an option (enabled by default) to disable the wobble, that can then be disabled to fix the bug whilst keeping the wobble.

WorldLanguages:
> Was the point of the original addon not that the animations were annoying and not just that it fixed the scrollbar bug?

Good point. The original addon intends to disable the animation AND fix the bug. I'm not sure how we should offer a new way that fixes  the bug without disabling the animation. Also, if the bug is ever fixed, the addon will continue to exist because it would still remove the animation.

If the fix is this easy, couldn't we convince the Scratch Team to update it soon, so that all users benefit?

BroJac5246:
> If the fix is this easy, couldn't we convince the Scratch Team to update it soon, so that all users benefit?

Good point, someone should just create a PR and fix it for everyone

CST1229:
I'd say it should be a different addon, as some users might want to disable sprite wobble for reasons other than fixing that bug (and this PR essentially removes a feature).

WorldLanguages:
Let's see if the Scratch Team is willing to make the issue help-wanted.

WorldLanguages:
> Let's see if the Scratch Team is willing to make the issue help-wanted.

No luck there. Setting a reminder 7 days from now:

Secret-chest:
Well, I hate the wooble, and the way all sprites preceding the one I'm dragging to get highlighted

Samq64:
If this does get merged, I think there should be a toggle to also disable the wobble.

iqnite:
> Also tested on Scratch Desktop 3.27.0.
How?

WorldLanguages:
> Let's see if the Scratch Team is willing to make the issue help-wanted.

No luck after 6 months, let's release this.
I think this should be a new addon though. I don't think it makes sense to remove the option to remove the wobble, nor it makes sense to merge a bug fix and a feature into the same addon

Secret-chest:
Regarding the wooble, it feels very weird that all sprites before the one being dragged to get highlighted as well.

BroJac5246:
> > Let's see if the Scratch Team is willing to make the issue help-wanted.
> 
> No luck after 6 months, let's release this.
> I think this should be a new addon though. I don't think it makes sense to remove the option to remove the wobble, nor it makes sense to merge a bug fix and a feature into the same addon

What would happen if a PR were to be sent despite the issue it resolves not being marked as help-wanted? Would they reject the PR or would they say, "Whatever, this works and it's less work for us 🤷"?

mxmou:
> Regarding the wooble, it feels very weird that all sprites before the one being dragged to get highlighted as well.

All sprites get highlighted in vanilla Scratch. If only some do, that might be a SA bug.

mxmou:
> > Let's see if the Scratch Team is willing to make the issue help-wanted.
> 
> No luck after 6 months, let's release this. I think this should be a new addon though. I don't think it makes sense to remove the option to remove the wobble, nor it makes sense to merge a bug fix and a feature into the same addon

If we keep the option to remove the wobble, it should only remove that, not the grow animation.

WorldLanguages:
> No luck after 6 months, let's release this.

No luck after ~9 months. @DNin01 Let's get this ready for release?

As mxmou said, let's keep the setting to remove the wobble. And I'd say we rename the addon and keep the "remove wobble" setting disabled by default, except for previous users of the addon.

DNin01:
Sure, but what should we do about the removed credits and whether or not we should keep the option to disable the wobble?

DNin01:
I guess all the credits should stay because we're keeping the option to remove the wobble.

addons/disable-sprite-wobble/addon.json
DNin01: I guess all the credits should stay because we're keeping the option to remove the wobble.

WorldLanguages:
Yeah, keep the credits

DNin01:
All done!

addons/disable-sprite-wobble/addon.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "Disable sprite wobble in sprite pane",
-  "description": "Allow scripts, costumes, and sounds to be moved to different sprites without the sprites moving around within the sprite pane.",
+  "name": "Fix sprite pane glitching",
+  "description": "Allow scripts, costumes, and sounds to be dragged and dropped onto different sprites without the sprites jumping around within the sprite pane.",
```
WorldLanguages, line 3: Maybe the description should start with `Fixes a bug in the Scratch editor where...` or `Fixes a Scratch bug that...` just like other bug fix addons

addons/disable-sprite-wobble/addon.json
```diff
@@ -9,17 +9,43 @@
...
+        "settings": {
+          "disable-wobble": true
+        }
+      }
+    }
+  ],
+  "settings": [
+    {
+      "id": "disable-wobble",
+      "name": "Disable sprite wobble",
```
WorldLanguages, line 36: I wonder how we could name this setting so that it doesn't feel "mandatory" to enable it. Right now, I believe some users might believe the addon does nothing if this isn't enabled.

addons/disable-sprite-wobble/addon.json
WorldLanguages: Also, is it "drag and drop into" or "drag and drop onto"?

Samq64: into I think

addons/disable-sprite-wobble/addon.json
WorldLanguages: We can add a `Fix glitch` setting that defaults to true (a bit stupid, but it would work).
Instead we could also rename this setting to something like `Also disable sprite wobble` (?)
Or maybe this is just me and users will understand the setting is an addition, and that the bug is fixed in default settings.

DNin01: Or maybe Disable sprite wobble should be removed entirely. Probably no one would have thought to add it had it not caused the sprite pane column fluctuation bug.

Samq64: Unless it was part of a reduce animations addon but I think think Scratch really has any others to begin with.

WorldLanguages: > Or maybe Disable sprite wobble should be removed entirely. Probably no one would have thought to add it had the sprite pane column fluctuation bug not existed.

After receiving feedback about the removal of "fix footer", I unfortunately disagree :p

WorldLanguages: Maybe a select setting:

Mode: `Fix glitch` / `Fix glitch and disable sprite wobble`

WorldLanguages:
Note: we won't transition settings.

image

WorldLanguages (approved):

iqnite:
> Note: we won't transition settings.
> 
> image

The Mode setting should be on the second option by default, so the current behavior of the add on won’t be changed for users with the add on enabled before the update.

WorldLanguages:
> The Mode setting should be on the second option by default, so the current behavior of the add on won’t be changed for users with the add on enabled before the update.

I think it is okay in this case, as previously there was no way to only fix the bug without remove the wobbling.

But yeah, we could try to migrate the setting if we want to.

addons/disable-sprite-wobble/addon.json
```diff
@@ -9,17 +9,53 @@
...
+      "name": "Mode",
+      "type": "select",
+      "potentialValues": [
+        {
+          "id": "none",
+          "name": "Fix bug"
+        },
+        {
+          "id": "no-wobble",
+          "name": "Fix bug and disable sprite wobble"
```
DNin01, line 45: I would like if it was clearer why this setting was here. With the update notification when this is initially released, this will make sense, but after that, it wouldn't fit here.

<details><summary>One idea...</summary>
<p>

```suggestion
          "name": "Fix bug and disable sprite wobble (original)"
```

</p>
</details>

We could also think about eventually moving either this or the bug fix to a new addon or something.

WorldLanguages: Maybe add an info.

WorldLanguages: @DNin01Adding "original" to the option is also okay. Commit your suggestion if you like it.

DNin01: > @DNin01 Adding "original" to the option is also okay. Commit your suggestion if you like it.

For now, with the update notice to go along, I think the current option name is okay. Not sure what to do afterward but we'll figure something out.

PR #5699, opened 2023-02-05
Dynamic enable/disable for `one-click-studio-invites`

DNin01:
A step towards #1673

### Changes

Made the "Accept Invite" button show when you enable the addon and hide when you disable it, all without a refresh.

### Reason for changes

We're trying to add dynamic enable/disable support to all addons at some point.

### Tests

Tested on Edge with SA v1.30.0-pre.


WorldLanguages (approved):
Haven't tested

Samq64 (approved):

PR #5705, opened 2023-02-08
Edit descriptions/names of 5 addons

DNin01:
### Changes

Check `Files changed` for the changes.

### Reason for changes

These changes are meant to make a few of the addons easier to find, understand, and use.

### Tests

Should work.

### DIFF
--- a/addons/hide-new-variables/addon.json
+++ b/addons/hide-new-variables/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Hide new variables",
-  "description": "Do not automatically create monitors for newly created variables or lists.",
+  "description": "Do not automatically show new variables or lists on the project player.",
   "credits": [
     {
       "name": "GarboMuffin"
--- a/addons/hide-new-variables/addon.json
+++ b/addons/hide-new-variables/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Hide new variables",
-  "description": "Do not automatically show new variables or lists on the project player.",
+  "description": "Do not automatically show new variables or lists on the stage.",
   "credits": [
     {
       "name": "GarboMuffin"
--- a/addons/hide-new-variables/addon.json
+++ b/addons/hide-new-variables/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Hide new variables",
-  "description": "Do not automatically show new variables or lists on the stage.",
+  "description": "Do not automatically show monitors for newly created variables or lists.",
   "credits": [
     {
       "name": "GarboMuffin"
--- a/addons/pause/addon.json
+++ b/addons/pause/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Pause button",
+  "name": "Project pause button",
   "description": "Adds a pause button next to the green flag.",
   "credits": [
     {
--- a/addons/pause/addon.json
+++ b/addons/pause/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "Project pause button",
-  "description": "Adds a pause button next to the green flag.",
+  "name": "Pause button",
+  "description": "Adds a button to pause the project next to the green flag.",
   "credits": [
     {
       "name": "Jeffalo"
--- a/addons/pick-colors-from-stage/addon.json
+++ b/addons/pick-colors-from-stage/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "Select stage colors in the costume editor",
-  "description": "Allows the costume editor's eye dropper to also pick colors from the stage.",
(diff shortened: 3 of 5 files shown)


addons/hide-new-variables/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Hide new variables",
-  "description": "Do not automatically create monitors for newly created variables or lists.",
+  "description": "Do not automatically show new variables or lists on the project player.",
```
mxmou, line 3:
```suggestion
  "description": "Do not automatically show new variables or lists on the stage.",
```

addons/pause/addon.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Pause button",
+  "name": "Project pause button",
```
mxmou, line 2: I think short names are better for simple addons like this. You can add more detail to the description if you think it's necessary, for example "Adds a button next to the green flag that pauses the project."

addons/pick-colors-from-stage/addon.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "Select stage colors in the costume editor",
-  "description": "Allows the costume editor's eye dropper to also pick colors from the stage.",
+  "name": "Pick colors on stage with eyedropper",
+  "description": "Allows the costume editor's eyedropper tool to also pick colors from the stage.",
```
mxmou, line 3: "Tool" usually means a mode like select or brush.

addons/pause/addon.json
DNin01: Other related addons such as "muted project player mode" and "project volume slider" have "project" in their names.

WorldLanguages: It's obvious that "pause button" refers to projects. What else in the Scratch website could be paused?

mxmou (approved):

Samq64 (approved):

addons/hide-new-variables/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Hide new variables",
-  "description": "Do not automatically create monitors for newly created variables or lists.",
+  "description": "Do not automatically show new variables or lists on the stage.",
```
WorldLanguages, line 3: So are we going to avoid using the word "monitor"?

Samq64: I don't think we need to, but I think this description is fine.

DNin01: Here are a few variations that use the word “monitors”:
```suggestion
  "description": "Do not automatically show variable and list monitors on the stage.",
```
```suggestion
  "description": "Do not automatically show monitors for new variables or lists.",
```
```suggestion
  "description": "Do not automatically show monitors for newly created variables or lists.",
```

Or one other option:
```suggestion
  "description": "Hide newly created variables and lists by default.",
```

BroJac5246: I like the third one best, as it's clear and uses 'monitor.'

WorldLanguages: I didn't mean to say we should be using the word "monitor". Honestly, I think we should not

WorldLanguages:
Is there enough consensus here?

WorldLanguages:
Regarding the use of the word "monitor", a new addon was added after the creation of this PR, which uses it for the description (copy-reporter).
Are we actively trying to avoid using that word? If so, we should also edit that description on this PR (git pull master first)

DNin01:
> Regarding the use of the word "monitor", a new addon was added after the creation of this PR, which uses it for the description (copy-reporter). Are we actively trying to avoid using that word? If so, we should also edit that description on this PR (git pull master first)

I think there are cases, including `copy-reporter`'s description where using the word "monitor" makes more sense than not.

mxmou:
The word "monitor" was used in the "show variable" block's help text in 2.0.
image

I don't think it's used anywhere in Scratch 3.0's UI, but a lot of Scratchers use it and might want to be able to search for it on the settings page. There are currently two addons that mention monitors in the description: "hide new variables" and "copy reporter values".

WorldLanguages:
> There are currently two addons that mention monitors in the description: "hide new variables" and "copy reporter values".

Note that "hide new variables" would stop mentioning monitors if we merged this PR in the current state (unless I'm missing something). So searching for "monitor" would only match 1 addon.

DNin01:
> So searching for "monitor" would only match 1 addon.

That's something to think about.

WorldLanguages:
I think we could add (keep) the word "monitor" in hide-new-variable's description before merging this

WorldLanguages (approved):

PR #5714, opened 2023-02-10
Prevent visited links from turning blue on messages page

### LINKED ISSUE #5709: The style of username links is inconsistent on the Messages page with `dark-www` enabled
### Description

**Expected behavior:** All usernames on the Messages page should only turn blue when you're hovering over them, like vanilla Scratch.

**Actual behavior:** With `dark-www` enabled, usernames for user profiles you have visited will be highlighted blue always, and others will only be white and highlighted blue while you're hovering over them.

### Steps to reproduce

1. Enable the Website dark mode and customizable colors addon. You can use any theme.
2. Open your messages page.
3. Check various usernames in your messages. Notice how some usernames are colored blue when they shouldn't be.

Screenshot of Messages page in dark mode

### Environment

Browser: Edge 109
Scratch Addons version: 1.30.2 and 1.30.0-pre
Operating system: Windows 10

fractioneater:
Resolves #5709

### Changes

Links that are visited no longer turn blue on the messages page when dark-www is enabled.

### Reason for changes

What's happening currently is inconsistent with the behavior of vanilla Scratch.

### Tests

Tested on a chromium browser.

### DIFF
--- a/addons/dark-www/experimental_scratchwww.css
+++ b/addons/dark-www/experimental_scratchwww.css
@@ -221,6 +221,7 @@ body:not(.sa-body-editor) .button.white,
 .tabs li.active:hover,
 .grid .thumbnail .thumbnail-info .thumbnail-title .thumbnail-creator a,
 .unsupported-browser .faq-link-text,
+a.social-messages-profile-link,
 a.social-messages-profile-link:link,
 .studio-selector-button-text-unselected,
 .studio-project-tile .studio-project-username,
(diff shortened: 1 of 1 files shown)


DNin01 (approved):

WorldLanguages:
Looks like this does fix the issue, but there might be a better way to do so. Let's wait for mxmou review.

addons/dark-www/experimental_scratchwww.css
```diff
@@ -221,6 +221,7 @@ body:not(.sa-body-editor) .button.white,
 .tabs li.active:hover,
 .grid .thumbnail .thumbnail-info .thumbnail-title .thumbnail-creator a,
 .unsupported-browser .faq-link-text,
+a.social-messages-profile-link,
 a.social-messages-profile-link:link,
```
mxmou, line 225: Add `a.social-messages-profile-link:visited` here and remove the `:not()`s under "Link color" - they make selectors more specific, which will probably cause problems.

mxmou (changes requested):

mxmou (approved):

PR #5716, opened 2023-02-10
Fix user dropdown text overflowing

### LINKED ISSUE #5715: scratchr2 addon account menu is narrower than vanilla
### Description

From feedback:

> using scratch 2.0 -> 3.0, the 2.0 pages have an issue where "Configuración de la cuenta" in spanish language clips out of the box which does not happen on 3.0 pages

Technically it also "clips out of the box" in vanilla Scratch too, but it's barely visible.

### Steps to reproduce

Make sure to edit the HTML if your Scratch username is long.

Vanilla Scratch:
image

Scratch 2.0→3.0 addon:
image

Probably happens in some other languages too

### Environment

Any

DNin01:
### Changes

Adds `white-space: normal;` to the `#topnav ul.account-nav ul.user-nav li a` style rule in `addons/scratchr2/scratchr2.css`

### Reason for changes

To resolve #5715

After:
Text word-wrapping in the user dropdown

### Tests

Tested on Edge 111.

### DIFF
--- a/addons/discuss-button/compact-nav.css
+++ b/addons/discuss-button/compact-nav.css
@@ -9,8 +9,9 @@
   display: block;
   font-size: 0.75rem;
   font-weight: normal;
-  line-height: 0px;
-  padding-bottom: 15px;
+  line-height: normal;
+  padding-bottom: 8px;
+  white-space: normal;
 }
 
 .sa-compact-profile-icon {
--- a/addons/scratchr2/scratchr2.css
+++ b/addons/scratchr2/scratchr2.css
@@ -540,6 +540,7 @@ select:focus {
 #topnav ul.account-nav ul.user-nav li a {
   color: var(--darkWww-navbar-text, white);
   font-size: 13px;
+  white-space: normal;
 }
 #topnav ul.account-nav ul.user-nav li:hover {
   background: var(--darkWww-navbar-border, rgba(0, 0, 0, 0.1));


mxmou:
> #### Issues
> A long username with the two lines of text on top on each other
> (When the "compact user dropdown" setting in the `discuss-button` addon is enabled)
> But probably not a huge concern.

That happens because the `line-height` of the username is set to `0px` in `compact-nav.css`. Changing it to `normal` would fix this, but the spacing might need to be changed.

penguinmoose:
I think the dropdown should be made wider to fit the text, instead of the text going into a new line.

mxmou:
> I think the dropdown should be made wider to fit the text, instead of the text going into a new line.

scratch-www doesn't do that though, and the goal of this addon is to make everything as similar to scratch-www as possible.

WorldLanguages:
> That happens because the `line-height` of the username is set to `0px` in `compact-nav.css`. Changing it to `normal` would fix this, but the spacing might need to be changed.

@DNin01 Wish to look into this?

DNin01:
> > That happens because the `line-height` of the username is set to `0px` in `compact-nav.css`. Changing it to `normal` would fix this, but the spacing might need to be changed.
> 
> @DNin01 Wish to look into this?

I just made a change and it should be pretty good now.

mxmou (approved):

WorldLanguages (approved):

PR #5731, opened 2023-02-16
New addon: Remember collapsed forum categories

### LINKED ISSUE #5710: Remember Collapsed Forum Categories
### Why this would be helpful

Collapsing categories is pretty useless. It doesn't automatically collapse previously closed categories(in other words, remembering the categories you've collapsed in the past).

### How this addon works

Scratch already saves a cookie when you collapse a category. When it's been collapsed the cookie's value is `collapsed`, otherwise it's just empty. The name of the cookie uses the category's number proceeded by `category_body_`.

`.box-content` needs an inline `display` of `none`, and `#category_head_`(followed up by the category's number) needs a class of `collapsed`.'

_I have actually made a working version of this, though it's very basic [Steve0Greatness/ScratchAddons:rememeber-collapsed-categories](https://github.com/Steve0Greatness/ScratchAddons/tree/rememeber-collapsed-categories)_

### Possible alternatives

_No response_

### Additional context

_No response_

Steve0Greatness:
Resolves #5710

### Changes

This addon fixes the ability for the forums to remember the categories that have been closed from session to session.

### Reason for changes

At some point, this seems to have been intended, however it just doesn't work for some reason.

### Tests

It works, and it probably doesn't seem major issues.


addons/remember-collapsed-categories/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Remember Closed Categories",
+  "description": "Fixes the forums not remembering what forum categories you've closed.",
```
DNin01, line 3: May I suggest a new name and description?
```suggestion
  "name": "Remember collapsed forum categories",
  "description": "Categories you collapsed on the forum home page (the list of forums) will stay collapsed.",
```

Why my description change? To most people, this will seem like a new feature rather than a fix.

addons/remember-collapsed-categories/userscript.js
```diff
@@ -0,0 +1,54 @@
...
+  function shouldCollapseCategory(categoryNumber) {
+    let cookieName = `category_body_${categoryNumber}`;
+    let storedInCookie = document.cookie.includes(`${cookieName}=collapsed`);
+    return storedInCookie;
+  }
+
+  function changeCategoryCookieExistenceLength(categoryNumber) {
+    let cookieName = `category_body_${categoryNumber}`;
+    console.log(`Extending ${cookieName} for an extra 30 days`);
+    document.cookie = `${cookieName}=collapsed;max-age=${30 * 24 * 60 * 60}`;
```
WorldLanguages, line 47: Does this log the user out?

WorldLanguages: Nevermind, it wouldn't because auth cookies are HTTP only.

However, this would clear any other cookies. Please write or "borrow" `function setCookie()` and use it

addons/remember-collapsed-categories/addon.json
WorldLanguages: I don't think we should try really hard to make it clear this is technically a bug fix. Most users aren't aware of this broken feature.

addons/remember-collapsed-categories/userscript.js
WorldLanguages: Unfortunately Firefox does not support the [Cookie Store API](https://developer.mozilla.org/en-US/docs/Web/API/Cookie_Store_API) yet. So we can't use that.

mxmou: I thought setting `document.cookie` only changes one cookie.

WorldLanguages: @mxmou Huh, I think you are right. Weird old JS API. I always assumed the `setCookie()` boilerplate was used because the `document.cookie` setter wasn't magical.

Steve0Greatness: `document.cookie` only effects 1 cookie, it's a an odd behavior.

addons/remember-collapsed-categories/userscript.js
```diff
@@ -0,0 +1,54 @@
+export default async function ({ console }) {
+  const categories = document.querySelectorAll('[id^="category_body_"]');
```
WorldLanguages, line 2: Using document.querySelectorAll instead of waitForElement could break because you're running this with `"runAtComplete" false`.
Either run this at complete, or use a while(true) pattern with waitForElement like other addons.

Steve0Greatness: Done... I think.

addons/remember-collapsed-categories/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Remember collapsed forum categories",
+  "description": "Categories you collapsed on the forum home page (the list of forums) will stay collapsed when you leave and come back.",
```
WorldLanguages, line 3: The description can probably be improved, but I don't have any concrete suggestions. Low priority.

Some notes:
1. It should be clear that the collapsing feature is already available in vanilla Scratch, but that it doesn't persist tab sessions
2. Is /discuss/ called "the forum home page"? Do other addons name it differently?
3. "when you leave and come back" can probably be made clearer and less informal.

addons/remember-collapsed-categories/userscript.js
```diff
@@ -0,0 +1,57 @@
+export default async function ({ console, addon }) {
+  while (true) {
+    const categoryBody = await addon.tab.waitForElement('[id^="category_body_"]', {
+      markAsSeen: true,
+    });
+    console.log(`Found category ${categoryBody.id}`);
+
+    let categoryCollapse = categoryBody.querySelector(".toggle");
+    categoryCollapse.addEventListener("click", (event) => {
+      event.preventDefault();
```
WorldLanguages, line 10: What's the purpose of `preventDefault()` here?

addons/remember-collapsed-categories/addon.json
```diff
@@ -0,0 +1,21 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/discuss/"],
+      "runAtComplete": false
+    }
+  ],
+  "tags": ["community", "forums"],
+  "dynamicDisable": true,
+  "dynamicEnable": true,
```
WorldLanguages, line 19: I don't think this addon supports dynamicDisable. But because of the nature of this addon, it's hard to tell what supporting dynamicDisable would mean 🤔

addons/remember-collapsed-categories/userscript.js
```diff
@@ -0,0 +1,56 @@
...
+      continue;
+    }
+    extendCategoryCookieExistenceLength(categoryNumber);
+
+    const categoryContent = categoryBody.querySelector(".box-content");
+    categoryContent.style.display = "none";
+
+    const categoryHead = categoryBody.querySelector(`.box-head h4`);
+    categoryHead.classList.add("collapsed");
+    categoryHead.classList.add("sa-collapsed");
```
WorldLanguages, line 33: Does Scratch style category heads that have the "collapsed" class any different?

Does the "sa-collapsed" class have any specific purpose? I was expecting a userstyle targeting this class name, but there is none,

addons/remember-collapsed-categories/userscript.js
```diff
@@ -0,0 +1,56 @@
...
+
+    const categoryName = categoryHead.innerText.split("\n")[1]; // This is exclusively for debug purposes
+    console.log(`Category ${categoryName}(#${categoryNumber}) should be hidden`);
```
WorldLanguages, line 36: Reminder: debugging code should be removed before merging

addons/remember-collapsed-categories/userscript.js
```diff
@@ -0,0 +1,56 @@
...
+        console.log(`${categoryHead.id.split("_")[2]} is being opened`);
+        return;
+      } else {
+        extendCategoryCookieExistenceLength(categoryNumber);
+      }
+    });
+
+    const categoryNumber = categoryBody.id.split("_").pop();
+    if (!shouldCollapseCategory(categoryNumber)) {
+      removeCategoryCookie(categoryNumber);
```
WorldLanguages, line 23: Will this ever delete an existing cookie? I can't think of any situations where this function call would do anything.

addons/remember-collapsed-categories/userscript.js
mxmou: Probably to prevent scrolling to the top when clicking the buttons (which are actually `<a>` elements with `href="#"`).

addons/remember-collapsed-categories/userscript.js
Steve0Greatness: I think there was a reliability issue with `collapsed`, if I'm remember correctly from initially doing that.

addons/remember-collapsed-categories/userscript.js
Steve0Greatness: Normally Scratch sets the cookie to just an empty value, it doesn't delete it outright. Although it set the cookie's `max-length` to the session.

But I'm not sure about your question, are you asking if it could potentially delete the `scratchsession` cookie? If that is your question: the answer is no, it takes in a number as the input, this is used to delete the cookie that uses the naming scheme of `category_body_[number]`, in other words, the category cookies.

WorldLanguages: > are you asking if it could potentially delete the `scratchsession` cookie?

I was not referring to that.

addons/remember-collapsed-categories/addon.json
Samq64:
```suggestion
  "description": "Fixes a bug in Scratch where categories on the forum front page don't stay collapsed when reloading the page.",
```
It might not be specific enough though.

DNin01: I don't think we need to indicate that it's a bug. Most Scratchers will think it simply wasn't a feature.

WorldLanguages: @DNin01 I agree. I'm not sure why I said "it should be clear that the collapsing feature is already available in vanilla Scratch, but that it doesn't persist tab sessions".

WorldLanguages: I think that I was trying to say that Scratch already offers the ability to collapse forum categories, and we shouldn't imply that we are the ones adding that. In other words, that the collapsing UI/UX already exists, but that it doesn't persist sessions by default.
But clearly, the user doesn't care that technically this fixes a bug.

addons/remember-collapsed-categories/addon.json
```diff
@@ -0,0 +1,20 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/discuss/"],
+      "runAtComplete": false
+    }
+  ],
+  "tags": ["community", "forums"],
+  "dynamicEnable": true,
+  "versionAdded": "1.31.0"
```
WorldLanguages, line 19:
```suggestion
  "versionAdded": "1.32.0"
```

WorldLanguages (approved):

WorldLanguages:
@DNin01 Do you plan to re-review this? Please remove your pending review if you don't plan to

WorldLanguages:
Merging this PR as admin. This new addon is relatively simple and the risk of breaking other addons is very low.

PR #5734, opened 2023-02-17
New addon: Follow topics after posting

### LINKED ISSUE #5626: Auto-follow forum topics
### Why this would be helpful

Some people always want to be notified when they participate in a forum topic.

### How this addon works

This addon would automatically subscribe you to a topic when you post on it.

### Possible alternatives

_No response_

### Additional context

_No response_

Chiroyce1:
Resolves #5626 

### Changes

Adds a new addon called `auto-follow-topics`

### Tests

Tested manually twice

### Considerations
- Improve naming (addon ID and description)
- This addon was bodged together at 10pm in like 10 minutes, please do suggest changes in the code. I have made it to a state that works without too much hassle

i think i messed up the schema but will fix soon


addons/auto-follow-topics/userscript.js
```diff
@@ -0,0 +1,13 @@
+export default async function ({ addon, console, msg }) {
+  document.querySelector("[name=AddPostForm]").addEventListener("click", (event) => {
+    localStorage.setItem(location.href.split("/")[5], true);
+  });
+
+  if (localStorage.getItem(location.href.split("/")[5])) {
+    console.log("Following topic");
```
BroJac5246, line 7:
```suggestion
```

addons/auto-follow-topics/userscript.js
```diff
@@ -0,0 +1,13 @@
...
+    const followBtn = document.querySelectorAll(".follow-button")[1];
+    followBtn.focus();
+    followBtn.click();
+    localStorage.clear();
```
DNin01, line 11: Would this clear everything in local storage and possibly affect Scratch, or...?

WorldLanguages: Yeah, please avoid clearing all local storage. Use `localStorage.removeItem("key")` instead.

WorldLanguages: The root cause of the problem is that the local storage key is being used to store information, not just as an identifier. That's a pattern to avoid.
@Chiroyce I'd suggest having a single local storage key, with a constant name, that stores an array of IDs in JSON format.

WorldLanguages:
Any reason not to use sessionStorage instead?

addons/auto-follow-topics/userscript.js
Chiroyce1: My bad, did not consider that. I *somehow* thought that every addon had unique storage (Note to self: Always read docs)

> The root cause of the problem is that the local storage key is being used to store information, not just as an identifier. That's a pattern to avoid.

Got that, will make a fixed key like `auto-follow-topics`


> @Chiroyce I'd suggest having a single local storage key, with a constant name, that stores an array of IDs in JSON format.

A small problem with that I can think of, if a user unfollows a topic and revisits that page, it would follow it again. So I think its best not to make it persistent, but instead just check if a user has posted in a topic just before the page reloaded

Chiroyce1:
@WorldLanguages Nope, updated code to use session storage, and made sure everything still works. After someone clicks "Post", the ID is appended to an array in sessionStorage, and after the reload, it follows the topic and removes that specific topicID from the array, instead of entirely clearing the value or all of sessionStorage.

image

~~A slight issue that I've found now is that~~ Ah thankfully the ctrl+enter addon also clicks the submit button which triggers this addon, so yay

CST1229:
> Also I think the manifest check is failing, I probably messed up since I copied the schema from better-emojis and just changed it.

I think the tags must contain at least one of `["community", "editor", "popup"]`.

DNin01:
> > Also I think the manifest check is failing, I probably messed up since I copied the schema from better-emojis and just changed it.
> 
> 
> 
> I think the tags must contain at least one of `["community", "editor", "popup"]`.

Oh, okay. "Community" would probably fit here best.

DNin01 (commented):
I tested it in Edge and it's working fine for me. Looks like it also doesn't unfollow topics you were already following, which is good.

All we need to do now is fix the addon manifest (that check is failing).

addons/auto-follow-topics/addon.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Auto Follow Topics",
+  "description": "Automatically follow forum topics you post on",
```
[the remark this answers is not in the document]
BroJac5246: I think it should be "Automatically follow forum topics".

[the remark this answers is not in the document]
Chiroyce1: Well yeah, the entire addon manifest requires review because I just scraped one together to be good enough for testing.

"Auto follow topics" sounds like the best to me, the forums tag is very clearly visible next to the addon name, so I don't think adding "forums" to the addon name is a good idea.

Chiroyce1:
@DNin01 thanks for testing, is it worth checking if the user ran into the 60 second rule? It technically should be as easy as checking if that element exists, probably.

BroJac5246:
> @DNin01 thanks for testing, is it worth checking if the user ran into the 60 second rule? It technically should be as easy as checking if that element exists, probably.

I was originally against it, but now I think that we might as well.

DNin01:
> @DNin01 thanks for testing, is it worth checking if the user ran into the 60 second rule? It technically should be as easy as checking if that element exists, probably.
> 
> 

It's not really that important. Those users would probably end up posting on the topic 99% of the time anyway.

Chiroyce1:
@DNin01 it was rather easy to implement and it handles all 3 errors I've tested (60 sec rule, blank post, unsupported image host) - and it should work for any other ones as well. It just checks for an element with an `errorlist` class, if it isn't there, then follow the topic.

DNin01 (approved):

Samq64 (approved):

addons/auto-follow-topics/userscript.js
```diff
@@ -0,0 +1,22 @@
+export default async function ({ addon, console, msg }) {
+  const getTopics = () => JSON.parse(sessionStorage.getItem("auto-follow-topics") || "[]");
+  const setTopics = (topics) => sessionStorage.setItem("auto-follow-topics", JSON.stringify(topics));
+  const postError = () => Boolean(document.querySelector(".errorlist"));
+  const topicID = location.href.split("/")[5];
+
+  document.querySelector("[name=AddPostForm]").addEventListener("click", (event) => {
```
WorldLanguages, line 7:
```suggestion
  document.querySelector("[name=AddPostForm]")?.addEventListener("click", (event) => {
```
Form doesn't exist if user is logged out

addons/auto-follow-topics/addon.json
```diff
@@ -0,0 +1,21 @@
...
+  "credits": [
+    {
+      "name": "Chiroyce",
+      "link": "https://github.com/Chiroyce1"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/discuss/topic/*"]
```
WorldLanguages, line 14:
```suggestion
      "matches": ["topics"]
```

WorldLanguages (approved):

addons/auto-follow-topics/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Auto follow topics",
```
WorldLanguages, line 2: Just wondering if we can think of a better name. Ideally it should imply that the "auto follow" happens after posting.

WorldLanguages: I'm thinking of removing the word "auto" and keeping it simple: `Follow topics after posting`

BroJac5246: > I'm thinking of removing the word "auto" and keeping it simple: `Follow topics after posting`

How about `Automatically follow topics after posting`? It feels more automated that way.

DNin01: > > I'm thinking of removing the word "auto" and keeping it simple: `Follow topics after posting`
> 
> How about `Automatically follow topics after posting`? It feels more automated that way.

But that's a mouthful.

WorldLanguages: The word "automatically" can stay as the first word of the description.

PR #5738, opened 2023-02-18
Ctrl+Click to open open in new TurboWarp tab

iqnite:
### Changes
If the "Replace player" mode is selected in the `turbowarp-player` addon, the user can <kbd>Ctrl</kbd>+Click the button to open the project in a new TurboWarp tab, leaving the current Scratch tab open. Clicking the button will replace the Scratch player. If the "Open new tab" mode is selected, <kbd>Ctrl</kbd>+Click will have no effect.

### Reason for changes
Sometimes you want to open the current project in a new tab, and in actual browsers <kbd>Ctrl</kbd>+Clicking a link opens it to a new tab. Similar behavior in the TurboWarp button would make the GUI even more intuitive.

### Tests
It works fine with the "Replace player" mode, however, I've only roughly tested it with the other mode. Although it should work also there, somebody should try to test it again.


fractioneater:
Can you make it do the same thing when you middle-click the button? That's many people's preferred shortcut.

BroJac5246:
> Can you make it do the same thing when you middle-click the button? That's many people's preferred shortcut.

Yeah, I prefer three-finger-click/middle-click to open links in new tabs, since they're so much easier.

DNin01:
> Can you make it do the same thing when you middle-click the button? That's many people's preferred shortcut.

Middle click also generally opens links in a new tab.

DNin01 (approved):
Tested. I noticed that Ctrl+Clicking opens a new tab without switching to that tab, whereas using the Open new tab mode and clicking without holding Ctrl will switch to that tab. I'm guessing this is a browser thing (it seems like it is), so unless you know of a way to fix that, it's not a big deal.

CST1229:
> Tested. I noticed that Ctrl+Clicking opens a new tab without switching to that tab, whereas using the Open new tab mode and clicking without holding Ctrl will switch to that tab. I'm guessing this is a browser thing (it seems like it is), so unless you know of a way to fix that, it's not a big deal.

At least in Firefox, ctrl+clicking links doesn't switch to that tab (so that behavior is actually somewhat more consistent).

DNin01:
> > Tested. I noticed that Ctrl+Clicking opens a new tab without switching to that tab, whereas using the Open new tab mode and clicking without holding Ctrl will switch to that tab. I'm guessing this is a browser thing (it seems like it is), so unless you know of a way to fix that, it's not a big deal.
> 
> At least in Firefox, ctrl+clicking links doesn't switch to that tab (so that behavior is actually somewhat more consistent with the browser).

It happens in Edge, too.

BroJac5246:
> Tested. I noticed that Ctrl+Clicking opens a new tab without switching to that tab, whereas using the Open new tab mode and clicking without holding Ctrl will switch to that tab. I'm guessing this is a browser thing (it seems like it is), so unless you know of a way to fix that, it's not a big deal.

In Chrome, Ctrl-clicking opens the tab in the background.

DNin01:
> > Tested. I noticed that Ctrl+Clicking opens a new tab without switching to that tab, whereas using the Open new tab mode and clicking without holding Ctrl will switch to that tab. I'm guessing this is a browser thing (it seems like it is), so unless you know of a way to fix that, it's not a big deal.
> 
> In Chrome, Ctrl-clicking opens the tab in the background.

That's basically the same thing. So, it's a thing across browsers.

fractioneater:
> > Tested. I noticed that Ctrl+Clicking opens a new tab without switching to that tab, whereas using the Open new tab mode and clicking without holding Ctrl will switch to that tab. I'm guessing this is a browser thing (it seems like it is), so unless you know of a way to fix that, it's not a big deal.
> 
> At least in Firefox, ctrl+clicking links doesn't switch to that tab (so that behavior is actually somewhat more consistent with the browser).

Firefox has a setting for that, so it's only consistent with the browser for users that have that toggle disabled.

CST1229:
We could instead make the button an `<a>` with a click event handler that prevents default; that way it works just like a link unless you click on it.

addons/turbowarp-player/addon.json
```diff
@@ -9,8 +9,13 @@
     },
     {
       "type": "notice",
-      "text": "TurboWarp has separate cloud data servers. \"Replace player\" mode will send your Scratch username to TurboWarp.org for use in cloud data.",
+      "text": "\"Replace player\" mode will send your Scratch username to TurboWarp.org for use in cloud data. TurboWarp has separate cloud data servers, so the cloud variables will be different than the ones on the Scratch player.",
```
DNin01, line 12:
```suggestion
      "text": "The \"Replace player\" mode will send your Scratch username to TurboWarp.org for use in cloud data. TurboWarp has separate cloud data servers, so the cloud variables will be different than the ones on the Scratch player.",
```

DNin01 (approved):

addons/editor-unshare-button/button.css
```diff
@@ -1,4 +1,4 @@
 .sa-unshare-button {
-  background: hsla(30, 100%, 55%, 1);
+  background: #ff4d4d;
```
Samq64, line 2: Please don't make unrelated changes in this pull request.

addons/turbowarp-player/userscript.js
```diff
@@ -41,7 +41,13 @@ export default async function ({ addon, console, msg }) {
...
       search = `#?token=${projectToken}`;
     }
-    if (action === "player") {
+    if ((action === "link") || (e.ctrlKey) || (e.metaKey)) {
+      window.open(
+        `https://turbowarp.org/${window.location.pathname.split("/")[2]}${search}`,
+        "_blank",
+        "noopener,noreferrer"
+      );
+    } else if (action === "player") {
```
Samq64, line 50: I'm pretty sure this if check isn't necessary.

addons/turbowarp-player/userscript.js
```diff
@@ -99,4 +99,4 @@ export default async function ({ addon, console, msg }) {
 
     scratchStage = document.querySelector(".guiPlayer");
   }
-}
+}
```
Samq64, line 102: Adding back the new line should fix the Prettier checks.

iqnite: Thanks!

addons/editor-unshare-button/button.css
DNin01, line 2: Changing the color of something from another addon should be discussed and committed separately.

DNin01 (changes requested):

addons/editor-unshare-button/button.css
iqnite: Sorry! The reason why I made them on this branch is that the Minor text and design changes also were on the master branch. I'll undo it.

DNin01 (approved):

WorldLanguages (approved):

PR #5739, opened 2023-02-19
Darken TurboWarp button to improve contrast

DNin01:
A step towards #5616

### Changes

Slightly darkened the color of the TurboWarp button and changed the color of the Scratch button to match highlight color (purple by default).

New TurboWarp and Scratch player button colors

### Reason for changes

To make the icons inside easier to see.

### Tests

Tested on Edge 116.


Samq64:
I'm not so sure I like these changes. The current colours are brand colours and don't think they should be changed. A possible solution for the Scratch button is to make the background Scratch blue. I can't think of any alternatives for the TurboWarp button though.

Endermaster3690:
are they customizable?

BroJac5246:
> are they customizable?

No

DNin01:
> are they customizable?

In my opinion, this isn't big enough a detail to deserve a setting switch.

WorldLanguages:
Is it important to ensure there's enough contrast between the background color and the icons? There's no text here

WorldLanguages:
> Is it important to ensure there's enough contrast between the background color and the icons? There's no text here

Hmm, what do you think?

Secret-chest:
> Hmm, what do you think?

It's good as it was, the icons have outlines and are a S and a T respectively.

DNin01:
> @DNin01 🤔

According to APCA, [the outlines on the icon may not be thick enough](https://www.myndex.com/APCA/?BG=ffffff&TXT=ff4d4d). But since the white outer outline of the icon contrasts okay with the button background, it would be somewhat recognizable even if you can't see the inside parts of it (think of that like if the whole T was white), and with a hover title, I don't think the icon matters _too_ much.

The red button as a whole is accessible on the background according to APCA, but _almost_ passes WCAG 2.0 at about 2.9:1.

I'm interested in the idea of making the background of the Scratch button purple or even thinking about a different design such as a rounded-rectangle-shaped toggle. I do think [the Scratch button needs to be changed](https://www.myndex.com/APCA/?BG=ffffff&TXT=f8aa36).

So yeah, I'll probably make some further adjustments. I would love if I could keep the brand colors as close as possible.

Secret-chest:
Remember that when APCA says OK, it means it passes AAA. Passing only AA requires a score 10 points lower.

WorldLanguages:
> I do think [the Scratch button needs to be changed](https://www.myndex.com/APCA/?BG=ffffff&TXT=f8aa36).

Yeah, we should probably increase the contrast in both

DNin01:
I updated the design. The TurboWarp button is a little lighter and the Scratch one now has a purple background and adapts to your `dark-www` settings.

WorldLanguages:
> and the Scratch one now has a purple background and adapts to your dark-www settings.

Looks like the Scratch button now uses the same exact background as the "See Inside" button, so the contrast will vary depending on the "highlight color" setting

The TurboWarp button bg color looks great!

Samq64 (approved):

WorldLanguages:
> Looks like the Scratch button now uses the same exact background as the "See Inside" button, so the contrast will vary depending on the "highlight color" setting

Is this superior to providing a fixed color? The default highlight color probably has more contrast than the old unchangeable background color, but the contrast will vary if the user changes the setting.

DNin01:
> Is this superior to providing a fixed color? The default highlight color probably has more contrast than the old unchangeable background color, but the contrast will vary if the user changes the setting.

Contrast in user-set settings is not really a problem - if the user can't see something in a color scheme they made, they're not going to keep using it.

WorldLanguages:
> Ensuring colors chosen by the user have enough contrast doesn't really matter - if the user can't see something in a color scheme they made, they're not going to keep using it.

In theory, if you have enough image buttons, there might be a case where the user has no color choice that would work for all at the same :P
But it's just this one, so it's probably fine

WorldLanguages (approved):

PR #5742, opened 2023-02-19
New addon-blink animation

DNin01:
### Changes

- The `addon-blink` animation now has a drop-shadow to make the addon glow orange
- The animation's timing has been modified 

### Reason for changes

I feel like the animation is more noticeable with these changes.

### Steps to reproduce

To see the animation for yourself:
1. Search for "Scratch Notifier" in the popup.
2. Disable the addon if it's already enabled, then enable the addon, then on the following prompt, choose "Open full screen settings".
3. Observe the animation. Refresh the page or use developer tools to replay it.

### Tests

Tested on Edge 109.

### DIFF
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -79,17 +79,12 @@ h1 {
 }
 
 @keyframes addonBlink {
-  10% {
+  0% {
     border: 1px solid var(--content-border);
-  }
-  25% {
-    border: 1px solid var(--orange);
   }
   50% {
-    border: 1px solid var(--content-border);
-  }
-  75% {
     border: 1px solid var(--orange);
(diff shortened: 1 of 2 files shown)


webpages/settings/style.css
```diff
@@ -79,17 +79,12 @@ h1 {
...
   }
-  25% {
-    border: 1px solid var(--orange);
-  }
   50% {
-    border: 1px solid var(--content-border);
-  }
-  75% {
     border: 1px solid var(--orange);
+    filter: drop-shadow(0px 0px 7px var(--orange));
```
Samq64, line 87: I think 4px would be a little better.

Samq64 (approved):

PR #5767, opened 2023-03-02
Add v1.31 update notification

WorldLanguages:
YT thumbnail and video still pending

image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -242,14 +242,14 @@
   "extensionHasUpdated": {
     "message": "Scratch Addons has updated to v$1"
   },
-  "extensionUpdateInfo1_v1_30": {
-    "message": "New addon: \"costume editor snapping\", which lets you snap objects in the costume editor to bounding boxes and vector nodes. Go to $1 to enable it."
+  "extensionUpdateInfo1_v1_31": {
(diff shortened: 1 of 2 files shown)


Samq64 (approved):

BroJac5246:
Should the project progress bar revamp be mentioned somewhere? Some people might have been scared away from it before v1.30.1. It's okay if we don't (this is the v1.31.0 notification, after all), but this might be our best opportunity.

Also, if someone creates a blog post, should that be linked?

Samq64:
> Should the project progress bar revamp be mentioned somewhere? Some people might have been scared away from it before v1.30.1. It's okay if we don't (this is the v1.31.0 notification, after all), but this might be our best opportunity.

Oh, right I forgot about that. It should probably get the updated tag even though it's been available to users for a little while.

WorldLanguages:
> Should the project progress bar revamp be mentioned somewhere?

We could add a latestUpdate field. But since it's actually from 1.30.x then it would be weird. If you think it's worth it, open a PR soon and we can discuss it.

WorldLanguages:
> Also, if someone creates a blog post, should that be linked?

Good idea

michaeleldar:
Is there gonna be a video for this release, because I didn't see one.

CST1229:
> Is there gonna be a video for this release, because I didn't see one.

there isn't one - #5785

PR #5768, opened 2023-03-03
Add `latestUpdate` to `progress-bar`

Samq64:
### Changes

Adds the update tag to the progress bar addon with a temporary notice about the changes from 1.30.1.

### Reason for changes

A lot of users might have missed it since there was no update tag on the settings page.

### Tests

Tested on Chromium 110.

### DIFF
--- a/addons/progress-bar/addon.json
+++ b/addons/progress-bar/addon.json
@@ -12,7 +12,7 @@
   ],
   "latestUpdate": {
     "version": "1.31.0",
-    "temporaryNotice": "The progress bar has had a redisign since Scratch Addons 1.30.1."
+    "temporaryNotice": "The progress bar has had a redesign since v1.30.1."
   },
   "userscripts": [
     {
--- a/addons/progress-bar/addon.json
+++ b/addons/progress-bar/addon.json
@@ -10,6 +10,10 @@
       "link": "https://scratch.mit.edu/users/jbthepig/"
     }
   ],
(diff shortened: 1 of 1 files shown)


addons/progress-bar/addon.json
```diff
@@ -10,6 +10,10 @@
       "link": "https://scratch.mit.edu/users/jbthepig/"
     }
   ],
+  "latestUpdate": {
+    "version": "1.31.0",
+    "temporaryNotice": "The progress bar has had a redisign since Scratch Addons 1.30.1."
```
WorldLanguages, line 15: Typo: "redisign"

WorldLanguages: And add "v" before the version

TheColaber: I don't think we normally even include the version in the notice tho. Maybe just exclude that part?

WorldLanguages: @TheColaber Well, it's not bad to clarify that the change isn't technically new.

DNin01: > And add "v" before the version

Clarification: Keep `"version": "1.31.0"` the same but add the letter "v" in the `temporaryNotice` description.

Samq64: Should I remove "Scratch Addons" when I add the V?

WorldLanguages (approved):

PR #5770, opened 2023-03-03
forum-id: auto-add post number small & in brackets

### LINKED ISSUE #5766: forum-id: autoadded IDs no longer are small and in parentheses
### Description

(Unsure if this is a regression or an intentional thing.)

Probably before #5240 was merged, "Quote post number"'s auto-added IDs looked like this, in [small] tags and (parentheses):

In master, they now look like this, exactly the same as when manually pressing the Quote post number button:

### Steps to reproduce

1. Enable Quote post number and its "Automatically add post link to quotes" setting
2. Quote a post on the forums
3. Notice the text is normal-size and not in parentheses

(see the images in the description)

### Environment

Browser: Firefox 111.0b7
Scratch Addons version: Pre-regression: v1.30.2, post-regression: 9558611e16e0b3e30fde7cdbf2964aa7979585e7
Operating system: Windows 10

pufferfish101007:
Resolves #5766 

### Changes

Fixes a regression caused by #5240 that made the link added by `forum-id` with the 'auto add post link to quotes' setting enabled large and without brackets.

### Reason for changes

Keeps the behaviour consistent with the current release.

### Tests

Tested on chromium 107 - the standalone 'quote post number' button works as it always has and auto-adding the link doea format it as before #5240.

### DIFF
--- a/addons/better-quoter/module.js
+++ b/addons/better-quoter/module.js
@@ -197,7 +197,11 @@ function setup() {
     const username = post.find(".username").text();
     const idText =
       !forumIdAddon?.self?.disabled && forumIdAddon?.settings?.get?.("auto_add")
-        ? `[small](${getIDLink(id.substring(1), post["0"].querySelector(".box-head > .conr").textContent, false)})[/small]`
+        ? `[small](${getIDLink(
+            id.substring(1),
+            post["0"].querySelector(".box-head > .conr").textContent,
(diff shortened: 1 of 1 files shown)


CST1229 (approved):

WorldLanguages (approved):
Untested. Change looks right

PR #5773, opened 2023-03-04
"Variable manager" addon: trigger resize event after switching to another tab

WorldLanguages:
Resolves bug report from feedback:

> When I'm in the variable tab and I click on a link to a block in the deguger a poster bug happens.

[scrnli_04_03_2023_12-33-48.webm](https://user-images.githubusercontent.com/17484114/222927479-0826c6fa-df7c-4393-858e-7328d6ddd1c4.webm)

### Changes

`variable-manager` now dispatches a "resize" event if it detects that the user (or an addon) navigated away from the varmanager tab into a different tab.

~~Dispatch a "resize" event on the window object after the redux dispatch that changes the tab.~~

~~**It's possible that we're also missing this in other parts where we dispatch ACTIVATE_TAB**. I still have to check this.
I'm also wondering if maybe dispatching ACTIVATE_TAB is the right way to change the selected tab. Does vanilla Scratch also have this bug?~~

### Tests

Fixes the bug in Chromium

### DIFF
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -295,7 +295,6 @@ export default async function ({ addon, console, msg }) {
         type: "scratch-gui/navigation/ACTIVATE_TAB",
         activeTabIndex: 0,
       });
-      window.dispatchEvent(new Event("resize"));
     }
   };
 
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -295,6 +295,7 @@ export default async function ({ addon, console, msg }) {
         type: "scratch-gui/navigation/ACTIVATE_TAB",
         activeTabIndex: 0,
       });
+      window.dispatchEvent(new Event("resize"));
     }
   };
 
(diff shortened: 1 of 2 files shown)


mxmou:
> I'm also wondering if maybe dispatching ACTIVATE_TAB is the right way to change the selected tab. Does vanilla Scratch also have this bug?

Scratch does dispatch the resize event when switching to the code tab. I think the problem is that the code tab is shown and resized before the variable manager is hidden.

WorldLanguages:
@mxmou Could you clarify?

Samq64 (approved):

WorldLanguages:
@mxmou Do you have any objections against merging this PR as-is? Do you think we should try to hide the variable manager completely before attempting to switch to the code tab? (If both answers are no, you don't have to reply)

mxmou (approved):
A fix in `variable-manager` would probably be better, but this works too.

WorldLanguages:
Ohh, you mean that `variable-manager` should listen for the Redux event and resize the window, instead of every addon resizing the window after changing the currently active tab? Haven't considered that. That's probably a better solution, since it will be compatible with future code.

mxmou:
> Ohh, you mean that `variable-manager` should listen for the Redux event and resize the window, instead of every addon resizing the window after changing the currently active tab? Haven't considered that. That's probably a better solution, since it will be compatible with future code.

Hiding the variable manager tab before Scratch triggers the resize event would be even better, but that's probably more complex.

Issue #5762, opened 2023-03-01
Editor crashes when dragging sound to sprite folder


WorldLanguages:
### Description

Editor crash. Requires "sprite folders" addon to be enabled.

### Steps to reproduce

Easy to reproduce. See gif.

![brave_2l8fG0LI2y](https://user-images.githubusercontent.com/17484114/222276818-30781778-07b3-472a-84ea-89ee318fe27b.gif)

### Demonstration(s) and additional context

This is not a recent regression. It's very likely that this bug has happened for months.

### Environment

Report is Chromium. I reproduced in Chromium. Probably happens in Firefox too.

WorldLanguages:
cc @GarboMuffin addon crashing editor

PR #5784, opened 2023-03-05
Temporal bug fix for #5762 ("folders" editor crash)

WorldLanguages:
Temporal bug fix for #5762, hoping to avoid to make a new release that can crash the editor. But the bug has probably been here for months.

Appears to avoid crashes. Sharing sounds between sprites appears to work normally after this fix.

### DIFF
--- a/addons/folders/userscript.js
+++ b/addons/folders/userscript.js
@@ -1218,6 +1218,19 @@ export default async function ({ addon, console, msg }) {
         newIndex
       );
     };
+
+    // Temporal bug fix for #5762
+    const originalShareSoundToTarget = vm.shareSoundToTarget;
+    vm.shareSoundToTarget = function (...args) {
+      const target = this.runtime.getTargetById(args[1]);
+      if (!target) {
+        // Avoid reading property from null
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Pinging for easy access: @GarboMuffin

addons/folders/userscript.js
```diff
@@ -1218,6 +1218,19 @@ export default async function ({ addon, console, msg }) {
...
       );
     };
+
+    // Temporal bug fix for #5762
+    const originalShareSoundToTarget = vm.shareSoundToTarget;
+    vm.shareSoundToTarget = function (...args) {
+      const target = this.runtime.getTargetById(args[1]);
+      if (!target) {
+        // Avoid reading property from null
+        return Promise.reject();
```
GarboMuffin, line 1228: Since it's going to end up logged as a scary red warning in the JS console, let's be a bit more specific than `Uncaught (in promise) undefined`
```suggestion
        return Promise.reject(new Error("Dropping sound into folder is not supported"));
```

apple502j (approved):

PR #5788, opened 2023-03-06
New "customizable navigation bar" (discuss-button) setting: stick to top of screen/page

callumjt:
new setting for customizable-navigation-bar setting:
no navbar stick
makes the navbar not stick to the top

not really expecting this to be accepted. just made it for fun and decided to do a pull request


callumjt:
sorry about there being no labels, not really sure how to add them

DNin01 (changes requested):
Great job! However, it doesn't work on scratchr2-based pages (such as profile pages, my stuff, or the forums - the ones with the Scratch 2.0 design).

callumjt:
> Great job! However, it doesn't work on scratchr2-based pages (such as profile pages, my stuff, or the forums - the ones with the Scratch 2.0 design).

oh ill try to fix that

DNin01 (approved):
Thanks to the change I made, it should work on all pages now.

callumjt:
just updated it. for some reason the my stuff page had a different class for the navbar so i added that class

DNin01:
> just updated it. for some reason the my stuff page had a different class for the navbar so i added that class

Actually, I don't think that's necessary - adding `.innerwrap` only shifts the navigation bar to the left on `scratchr2` pages. It still works fine without `.innerwrap`, and all `scratchr2` pages seem to have that element within `#topnav`.

<sub>Microsoft Edge 110, Windows 10</sub>

callumjt:
> > just updated it. for some reason the my stuff page had a different class for the navbar so i added that class
> 
> 
> 
> Actually, I don't think that's necessary - adding `.innerwrap` only shifts the navigation bar to the left on `scratchr2` pages. It still works fine without `.innerwrap`, and all `scratchr2` pages seem to have that element within `#topnav`.
> 
> 
> 
> <sub>Microsoft Edge 110, Windows 10</sub>

When using the 2.0 -> 3.0 addon it uses that class for the navbar

mxmou:
> > > just updated it. for some reason the my stuff page had a different class for the navbar so i added that class
> > 
> > 
> > Actually, I don't think that's necessary - adding `.innerwrap` only shifts the navigation bar to the left on `scratchr2` pages. It still works fine without `.innerwrap`, and all `scratchr2` pages seem to have that element within `#topnav`.
> > Microsoft Edge 110, Windows 10
> 
> When using the 2.0 -> 3.0 addon it uses that class for the navbar

2.0 → 3.0 doesn't change any IDs or classes.

callumjt:
> > > > just updated it. for some reason the my stuff page had a different class for the navbar so i added that class
> 
> > > 
> 
> > > 
> 
> > > Actually, I don't think that's necessary - adding `.innerwrap` only shifts the navigation bar to the left on `scratchr2` pages. It still works fine without `.innerwrap`, and all `scratchr2` pages seem to have that element within `#topnav`.
> 
> > > Microsoft Edge 110, Windows 10
> 
> > 
> 
> > When using the 2.0 -> 3.0 addon it uses that class for the navbar
> 
> 
> 
> 2.0 → 3.0 doesn't change any IDs or classes.

Huh am i just dumb and read the wrong thing. Well either way it didn't work on the pages that 2.0 -> 3.0 changes. I might have just had an old update not sure

callumjt:
ye sorry i had an older version so it didnt work so i removed the .innerwrap

callumjt:
Reviewing sucks

iqnite:
Prettier check fails because you removed the line at the end of the userscript. In order to make it succeed, do a linebreak after the `}` at the end of every file.

callumjt:
> Prettier check fails because you removed the line at the end of the userscript. In order to make it succeed, do a linebreak after the `}` at the end of every file.

still not working

DNin01:
> > Prettier check fails because you removed the line at the end of the userscript. In order to make it succeed, do a linebreak after the `}` at the end of every file.
> 
> still not working

You already had a newline at the end, otherwise GitHub it would display a ⛔ icon. Try navigating to the Actions tab on your fork and enabling workflows. After making another commit, see if the code is formatted automatically, which will tell you what Prettier thought was "unpretty".

Prettier failing now is the least of our concerns. I believe it automatically runs again after the pull request is merged and should fix any problems.

WorldLanguages:
Regarding the new feature idea, it seems pretty reasonable. Theoretically a general purpose extension could be used to fix sticky elements, but we already have a relevant addon where we can add this setting.

addons/discuss-button/addon.json
```diff
@@ -86,14 +95,20 @@
       "id": "compact-nav",
       "type": "boolean",
       "default": false
+    },
+    {
+      "name": "Stick to top of screen",
+      "id": "stick",
+      "type": "boolean",
+      "default": true
```
WorldLanguages, line 103: The default should probably be `false` here, unless I'm misunderstanding what the setting does

callumjt: its true by default cause if its false it enables the setting

WorldLanguages: Oh, I see.
Generally, enabling an addon/setting replaces vanilla Scratch behavior with a custom one. Here, it does the opposite. I'd say it's a bit confusing.

callumjt: ye it was originally disabled by default but dnin wanted it like that so i just went with it

mxmou: > Generally, enabling an addon/setting replaces vanilla Scratch behavior with a custom one. Here, it does the opposite. I'd say it's a bit confusing.

Negative switches ("Don't stick to top of screen") can also be confusing.

DNin01: > ye it was originally disabled by default but dnin wanted it like that so i just went with it

Yeah, I switched it around because of the fact that you turn something on to disable something. But then it sounds like something that should be off by default, but that state is different than vanilla Scratch.

So how should we handle that?

DNin01: Should this be a two-choice single-select setting with the following options?

Stick to: `top of screen` `top of page`

callumjt: > Should this be a two-choice single-select setting with the following options?
> 
> Stick to: `top of screen` `top of page`

ye i think this is the best way to do it

WorldLanguages: I don't think there's a good way to phrase "don't stick to the top" without using negatives. In that case, I guess a two-choice select setting would make sense

callumjt:
just updated it to where it will give a setting (stick to) with 2 options: "top of page" and "top of screen"

addons/discuss-button/addon.json
DNin01: I'll work on adding that.

callumjt:
oh i cant request 2 people

DNin01:
> oh i cant request 2 people 

I know, that's kinda strange. At least I'm an organization member and I can request multiple users so I just did that.

addons/discuss-button/stick.css
```diff
@@ -0,0 +1,3 @@
+#navigation {
```
Samq64, line 1: Could yiu re-add `topnav`?
```suggestion
#navigation, #topnav {

```

addons/discuss-button/stick.css
```diff
@@ -0,0 +1,3 @@
+#navigation {
+    position: absolute !important;
```
Samq64, line 2:
```suggestion
    position: absolute;

```

addons/discuss-button/stick.css
callumjt: i didnt even noticed it was removed

Samq64 (approved):

callumjt:
Does wl need to merge this?

WorldLanguages:
Yes, I'll need to Prettier lint first

PR #5791, opened 2023-03-07
"Adding blocks by name improvement" improvement

Tacodiva:
It finally dawned on me how to make it so not typing the first part of a block (like just the word `clones`) could return all the blocks that string in them, and not turn the whole codebase into a cluster*frick*. Griffpatch suggested this and I think this is a important change which (should) mean users have no reason to go back to the old version.

### DIFF
--- a/addons/middle-click-popup/WorkspaceQuerier.js
+++ b/addons/middle-click-popup/WorkspaceQuerier.js
@@ -471,14 +471,11 @@ class TokenTypeStringEnum extends TokenType {
           yieldedToken = true;
         }
       } else {
-        if (query.lowercase.startsWith(valueInfo.lower, idx) && TokenTypeStringLiteral.TERMINATORS.indexOf(query.lowercase[idx + valueInfo.lower.length]) !== -1) {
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Pending improvement reminder:

image

Tacodiva:
Some more bugs for later today me to look into:

String going in input somehow:
image

Direction block spawning in the wrong colour:
image

DNin01:
> Direction block spawning in the wrong colour:
> image

Looks like the `zebra-striping` addon?

Tacodiva:
> > Direction block spawning in the wrong colour:
> > image
> 
> Looks like the `zebra-striping` addon?

Hahaha, yeah it is too. Well... that's one less thing I have to do :sweat_smile:

WorldLanguages:
This PR should ideally resolve #5808

Samq64 (approved):
I didn't review most of the code, but everything works great in both browsers and there aren't any console errors when using hex colors.

WorldLanguages (approved):

PR #5797, opened 2023-03-12
Volume slider updates

DNin01:
### Changes

- The volume slider now collapses with an animation when you're not hovering over its area. You can toggle this behavior with the "Always show slider" setting.
- The cursor turns into a pointer when hovering over the icon or the slider.
- The handle has been darkened to improve contrast.

### Tests

Tested on Edge 110 in combination with `mouse-pos` and `clones`.


addons/vol-slider/userstyle.css
```diff
@@ -42,13 +47,19 @@
...
   border: none;
   border-radius: 50%;
-  background-color: #cf63cf;
+  background-color: #c234c2;
+}
+
+.sa-vol-slider:not(:hover) > .sa-vol-slider-input {
+  opacity: 0;
+  width: 0;
+  margin-left: -3px;
```
DNin01, line 64: Will this be slightly annoying for touch screen users? Is that a problem?

Samq64: What about making it a setting?

DNin01: > What about making it a setting?

Could always do that. I was thinking about it.

Samq64 (approved):
I like the animation! I haven't tested on Firefox yet.

WorldLanguages:
Have you considered this idea:
If the "always show slider" setting is disabled, and the user is hovering the volume icon (hence forcing the slider to show), hide the clone count and the mouse position indicator.

This has a few benefits:
1. The clone count and mouse position indicators won't move (see what Samq said above)
2. If there's not enough room in the stage for some reason, it won't cause any weird overflows
3. The user won't be paying attention to the other addons while adjusting the volume anyway
4. Potentially, we could allow the volume slider to work while on "small stage mode" if we do this

DNin01:
@WorldLanguages I tried to do that with CSS, but couldn't figure out how.

WorldLanguages:
@DNin01 You're right, that sounds tricky to do in CSS. Maybe [this will do it](https://css-tricks.com/almanac/selectors/g/general-sibling/)? Or we could go with a JS solution instead.

DNin01:
> @DNin01 You're right, that sounds tricky to do in CSS. Maybe [this will do it](https://css-tricks.com/almanac/selectors/g/general-sibling/)? Or we could go with a JS solution instead.

It did work!

DNin01:
@WorldLanguages I'm done - I think it looks okay.

addons/vol-slider/addon.json
```diff
@@ -30,10 +41,21 @@
       "min": 0,
       "max": 100,
       "default": 100
+    },
+    {
+      "name": "Always show slider",
+      "id": "always",
+      "type": "boolean",
+      "default": false
```
WorldLanguages, line 49: Are we going to change the default?

WorldLanguages:
I think we should also update `mute-project` so that clicking the muted icon unmutes the project.

WorldLanguages:
Beautiful:

brave_O6bjoYENx2

addons/vol-slider/addon.json
DNin01: You mean have "always show slider" turned off even for existing users of the addon? In that case, yes, that sounds good. It's an upgrade.

addons/vol-slider/addon.json
```diff
@@ -30,10 +41,21 @@
...
     }
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
+  "updateUserstylesOnSettingsChange": true,
   "versionAdded": "1.29.0",
+  "latestUpdate": {
+    "version": "1.32.0",
+    "newSettings": ["always"]
```
WorldLanguages, line 58:
```suggestion
    "newSettings": ["always"],
    "isMajor": true
```

WorldLanguages (approved):

WorldLanguages:
> I think we should also update `mute-project` so that clicking the muted icon unmutes the project.

This was out of scope for this PR but we could add it at some point in the future.

WorldLanguages:
I think we could also consider what the standard margin between the elements in the stage header should be.

It currently looks wrong, specially if mouse-pos is showing all 6 digits at the same time. The volume icon and mouse position are too close to each other:

image

BroJac5246:
> I think we could also consider what the standard margin between the elements in the stage header should be.
> 
> It currently looks wrong, specially if mouse-pos is showing all 6 digits at the same time. The volume icon and mouse position are too close to each other:
> 
> image
> 

The distance between the flag and stop buttons in vanilla Scratch would be the answer that makes the most sense because it's consistent.

Issue #5556, opened 2023-01-06
Promisify `addon.tab.traps.vm` - addons should use `addon.tab.traps.getVm()`


WorldLanguages:
### Why this would be helpful

It would be consistent with getBlockly() and getPaper().
It would avoid race conditions.

### My suggestion

Remove the `addon.tab.traps.vm` getter.
Replace it with an `addon.tab.traps.getVm()` function that would return a promise. It would resolve to the VM if found, and reject if it is definitely not available. If the VM is not available just yet, the promise is expected to resolve as soon as it's available, instead of rejecting.
Userscripts would be expected to `await` it, so addon code wouldn't change substantially.

### Possible alternatives

Theoretically, an alternative would be to wait until the VM is available before running userscripts, but this would only make sense for project pages. If the VM failed to load, all addons, even those that aren't using the VM, would be affected. So I don't really like this alternative. (Source: https://github.com/ScratchAddons/ScratchAddons/issues/5517#issuecomment-1367689741)

### Additional context

See #5517 for context

WorldLanguages:
I believe #5962 is a better solution.

PR #5800, opened 2023-03-14
New addon 'Faster project loading'

Tacodiva:
Jazza requested a fix for this in the discord so I found the bug and patched it.


WorldLanguages:
I would suggest adding this as a new addon, named "faster project loading" (or similar) with a description that explains more in detail. The addon would be recommended and probably be enabled by default.

DNin01 (dismissed):
> I would suggest adding this as a new addon, named "faster project loading" (or similar) with a description that explains more in detail. The addon would be recommended and probably be enabled by default.

I think so too.

BroJac5246:
Link to the original issue if anyone wants it: https://github.com/LLK/scratch-gui/issues/8805

If we can do anything about this, then we definitely should, although we should wait to see if this will be fixed in Scratch first. That's the ideal situation since it benefits all Scratch users.

DNin01:
> we should wait to see if this will be fixed in Scratch first. That's the ideal situation since it benefits all Scratch users.

I think that's a good plan too. This issue doesn't seem to affect loading starting from the project page, which is what users will do most of the time, so I don't think we need to prioritize this. I think we can wait a bit to see if this will get fixed in Scratch first.

Tacodiva:
I agree, I think we should wait a little while to see if the ST patches it quickly before we merge this. It's a pretty bad issue that I'm sure is not kind to their backpack server so hopefully that's enough incentive for them to fix it.

WorldLanguages:
Let's wait 2 or 3 weeks to see if the Scratch Team fixes the bug?

Jazza-231:
I'm not sure if I'm allowed to talk here, but why do we need to wait to see if the scratch team fixes it? If they don't, we just wasted that time waiting for them, and if they do, then we can either leave this now-redundant fix (if it doesn't start causing its own problems) or revert these changes? By fixing it before the scratch team, at least it provides a fix for some people. Anyway, the people this is most likely to affect (advanced users) are also the same people who are more likely to use SA.

Tacodiva:
True, but we're not going to release for a while so it really doesn't matter if it's merged. If we merge it now and the ST fixes it before the next release we'd have to revert it and my effort creating a proper addon and everything would have been a waste of time. If 1.31.1 is gonna release and it's still not fixed then we'll merge.

WorldLanguages:
> If they don't, we just wasted that time waiting for them

In the opposite case, we waste some time creating and reviewing this new addon, translators also waste time, etc.
It's not standard practice for us to fix vanilla Scratch bugs that were *just reported some days ago*. It only makes sense to provide a bug fix if the Scratch Team is not capable of fixing the issue relatively quickly for all users.

WorldLanguages:
Setting a reminder 1 week from now:

BroJac5246:
> Ping! There has been no activity for 7 days.

Hmm... we have some discussion to do. I think we would all like to see a fix in Scratch's side, but they haven't responded. I'm not familiar with their policy on third-party PRs, but if we created one on their repository, would it be accepted? If not, then I support this addon, otherwise I believe we should wait further.

Samq64:
> I'm not familiar with their policy on third-party PRs, but if we created one on their repository, would it be accepted?

Probably not because the issue has no labels yet.

> We will only accept Pull Requests for bugs that have an issue filed that has a priority label

Tacodiva:
I would submit a PR if I knew the best way to fix this (even if they don't accept it, it could at least guide them to where the problem is), but as far as I can tell there's no obvious one-liner that will fix this problem, it's a bit more nuanced then that.

BroJac5246:
> I would submit a PR if I knew the best way to fix this (even if they don't accept it, it could at least guide them to where the problem is), but as far as I can tell there's no obvious one-liner that will fix this problem, it's a bit more nuanced then that.

Could the solution from this PR be ported?

Tacodiva:
> Could the solution from this PR be ported?

No. The solution in this PR doesn't even really fix the problem all it does is intercept the bad requests, make the loading bar ignore them and make them return an error instantly rather than having to actually send the request to backpack server and wait. A proper solution would make it so these requests don't happen in the first place and would definitely not overwrite `window.fetch` like this addon does.

WorldLanguages:
> Could the solution from this PR be ported?

Theoretically, yes - but userscripts are "hacky" by nature. The actual Scratch editor code is supposed to be clean.
I'm not sure if we're familiar enough with their codebase in order to tell how to properly resolve the issue without trapping window.fetch... If someone is, I'd motivate them to submit a PR to the LLK repo even if the "help wanted" label is missing.

WorldLanguages:
We aren't going to remove the "awaiting answer/followup" label, so we'll get reminders here every 7 days of inactivity 👍

BroJac5246:
Aaaaaaaaahhh!!!

This should really be fixed on Scratch's end, but they're not doing anything! I still think we should just go ahead and create a PR in their repo under the (sound) logic that this is too severe to delay upon.

Simultaneously, we should prepare this PR, just in case nothing happens on their end even after a PR is sent.

Tacodiva:
> I still think we should just go ahead and create a PR in their repo

@BroJac5246 Go ahead lmao. I don't know how to fix this and I can't be bothered working it out :P

BroJac5246:
> > I still think we should just go ahead and create a PR in their repo
> 
> @BroJac5246 Go ahead lmao. I don't know how to fix this and I can't be bothered working it out :P

I'm not familiar with what is causing the issue, nor how to fix it

DNin01:
> ...they're not doing anything!

Like any development team, the Scratch Team has a long task list, can't do anything instantly, needs to do other things first to avoid disruption and needs to check over each change to make for the best experience, so we must be patient. Even we take a while to add highly requested features sometimes (example: opacity slider).

WorldLanguages:
> Simultaneously, we should prepare this PR

Agreed. This should be a recommended, enabled by default, new addon.

Tacodiva:
Ok I've moved the fix into it's own, recommended and enabled by default addon. Tested it and everything seems to work as expected.

WorldLanguages:
@Tacodiva Could you test whether the bug is still fixed after my commit? If it is, then we can give further insights to the Scratch Team on the root cause of the issue (a race condition)

Tacodiva:
You are right that it is a race condition and the theory of this fix is sound. However, there are multiple reasons that fix doesn't work:
1. The userscript has runAtComplete set to false so `addon.tab.traps.vm` is undefined when this is run, leading to an error.
2. Even if the runAtComplete is true, when the VM loads the stores have already been added so it's too late to intercept the `addStore` method.
3. Even if we rearranged the stores after the VM has loaded, chances are many requests to the backpack server have already been sent out, so it's not really an ideal solution.

WorldLanguages:
1. That's a bug and we're tracking it at #5556 
2. Did you mean "even if runAtComplete is **false**?" In my machine I was able to intercept all addStore calls.
3. Of course, we have to rearrange the stores immediately. We either find a way to do it immediately, or we use a different solution.

@Tacodiva I find it funny that you ran into a race condition for a fix for a race condition.

WorldLanguages:
> I find it funny that you ran into a race condition for a fix for a race condition.

Tracking that on #5962 
Let's keep your userscript solution then @Tacodiva

Tacodiva:
Yeah, I tried to make a solution like yours work but couldn't because of the race condition

addons/faster-project-loading/userscript.js
```diff
@@ -0,0 +1,30 @@
...
+    if (method === "GET" && url.startsWith(BACKPACK_URL)) {
+      /*
+              We don't want to block actual requests for backpack assets.
+              A backpack request URL looks like:
+                https://backpack.scratch.mit.edu/Tacodiva?limit=20&offset=0
+              We can check for the '?' to differentiate them from wrong requests which look like:
+                https://backpack.scratch.mit.edu/de9e1ed35f087dcb297f1339b16deaf6.svg
+            */
+
+      if (url.indexOf("?") == -1) throw new Error("Request blocked by Scratch Addons faster project loading. ");
```
WorldLanguages, line 16: Is a `!addon.self.disabled` check required here?

WorldLanguages:
In general the PR looks good to merge.
Might have missed something because we had a lot of off-topic discussion as well as a failed reimplementation by me.

addons/faster-project-loading/userscript.js
WorldLanguages: Asking because I thought there should be one here too. There's a check in the postMessage trap below.

addons/faster-project-loading/addon.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Faster project loading",
+  "description": "Makes project loading faster when opening directly to the editor.",
```
WorldLanguages, line 3: Maybe the description could be improved.

DNin01: How about this?
```suggestion
  "description": "Fixes a Scratch bug that slows down project loading if the editor is opened from a link.",
```

WorldLanguages:
Tacodiva is researching some race conditions even with the current implementation of the addon userscript.
It's hard because since this feature is togglable and checking whether it's enabled is async, realistically all runAtComplete:false addons might run late in some cases.

Tacodiva:
The race condition wl is referring to above is described in more detail in issue #5975. There is currently no plans to address this due to the combined factors of it being very rare and very annoying to fix.

WorldLanguages (approved):

PR #5811, opened 2023-03-17
Add random query to userstyle link href

### LINKED ISSUE #5787: (master) Middle-click popup doesn't appear properly (possibly only in Firefox?)
### Description

The Insert blocks by name addon's popup seems to have no styling; it appears as an unstyled list of blocks below the page.
Might have been a regression from after the middle-click popup PR was merged.

### Steps to reproduce

1. Enable the Insert blocks by name addon
2. Go to the editor and trigger the middle-click popup
3. Scroll down on the page (not in the workspace)

https://user-images.githubusercontent.com/68464103/223028532-1148ccc0-a028-463e-baa1-22127083b682.mp4

### Environment

Browser: Firefox 111.0b8
Scratch Addons version: 1.31.0-prerelease 298db20becad110a7da9af1b3f24d2780c564c86
List of enabled addons: [g5kwfut2n6rhk2xtvcjfxyrz724lwx](https://scratchaddons.github.io/list-of-enabled-addons/?version=master&number=g5kwfut2n6rhk2xtvcjfxyrz724lwx)
Settings file: [scratch-addons-settings.zip](https://github.com/ScratchAddons/ScratchAddons/files/10894782/scratch-addons-settings.zip)
Operating system: Windows 10

Tacodiva:
Possibly resolves #5787

### DIFF
--- a/content-scripts/cs.js
+++ b/content-scripts/cs.js
@@ -220,7 +220,7 @@ function addStyle(addon) {
       link.setAttribute("data-addon-id", addon.addonId);
       link.setAttribute("data-addon-index", styleIndex);
       link.classList.add("scratch-addons-style");
-      link.href = userstyle.href;
+      link.href = userstyle.href + "?" + Math.floor(Math.random() * 1e16);
       appendByIndex(link, styleIndex);
     }
   }


apple502j (approved):
Maybe timestamps? though this looks okay

WorldLanguages (changes requested):
Unfortunately some other parts of the cs.js file don't expect the href property to be changed this way. Those will need to be changed from `.href ` to `.getAttribute("data-style-href")`.
We'd also have to double check for side effects in other files. There's 1 or 2 addons that directly interact with their <link> elements, hopefully not through their href property.

WorldLanguages:
I just committed some progress, but it remains untested. And I'm not sure if I missed some other file (or line within cs.js) that needs to be changed too.

Samq64 (changes requested):
> And I'm not sure if I missed some other file (or line within cs.js) that needs to be changed too.

Things are still broken in a lot of places.
image

WorldLanguages:
This line of code won't be a problem because the userscript manually adds the link element to the DOM (not the extension), but we need to be careful if any other addons are expecting URLs to end in `.css`, since that wouldn't be the case anymore because we would be adding URL params:

https://github.com/ScratchAddons/ScratchAddons/blob/61fd2a948df8cf4cc97a2265491e21c64e068c35/addons/dark-www/experimental_forums_preview.js#L20

Tacodiva:
Just looked into this again quickly. All the issues where just being caused by a mistake w_l made which I have fixed. 
I also did a dirty search in the repo for js files containing `.css` and I'm pretty sure "experimental_forums_preview" is the only problematic file, which I have fixed.

I've done some testing and it seems to me like this is good to go.

Tacodiva:
Ping! There has been no activity for 28 days.

((I think this should be a part of the 1.35 milestone given it has potential (and has actually) caused some major and hard to diagnose issues))

PR #5812, opened 2023-03-17
Hex color picker fixes

### LINKED ISSUE #5296: Changing hex color picker doesn't update the actual color
#4847 regression

> It´s very difficult, to get the main scratch color-picker to take the color from the hexcode of the scratch-addon-color-picker-extension.
> Enter does not seem to work well for that.

_Originally posted by @Drachenbauer in https://github.com/ScratchAddons/ScratchAddons/issues/311#issuecomment-1296238433_

CST1229:
Resolves #5296

### Changes

To the `color-picker` (Hex color picker) addon:
- Fixes the picker in the costume editor not actually changing the sliders
- Fixes the costume editor hex picker not updating in the following cases:
  - Clicking the Swap button
  - Moving the sliders
  - Clicking the transparent button 
  - Using the eyedropper
  - Switching from a gradient to a solid color while the second gradient color is selected
- Slightly refactored the code editor hex picker; it uses more Blockly now
- Fixed the code editor hex picker not updating at all after the picker has been opened

### Reason for changes

Fixing bugs.

### Tests

Seems to work on both Chrome and Firefox.


addons/color-picker/paint-editor.js
```diff
@@ -36,7 +36,7 @@ export default async ({ addon, console, msg }) => {
         const previousTool = addon.tab.redux.state.scratchPaint.color.eyeDropper.previousTool;
         if (previousTool) previousTool.activate();
         addon.tab.redux.state.scratchPaint.color.eyeDropper.callback(
-          tinycolor(hex).setAlpha(scratchAddons.opacitySliderAlpha).toRgbString()
+          tinycolor(hex).setAlpha(scratchAddons.opacitySliderAlpha).toHexString()
```
mxmou, line 39: `toHexString` ignores the alpha channel, so this will always return the same result as just `tinycolor(hex).toHexString()`.

WorldLanguages (approved):

PR #5816, opened 2023-03-18
share-through-mystuff: simplify CSS

mxmou:
### Changes

* Moves 2.0 → 3.0 compatibility code to `addons/scratchr2/mystuff.css`.
* Replaces a workaround for a Scratch bug with a much simpler one that doesn't need CSS.
* The addon currently tries to align the Delete and Share buttons on unshared projects with the Unshare button on shared projects. This is implemented in a way that doesn't work in languages other than English. I removed it in this PR because I don't think visual changes belong in this addon.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/share-through-mystuff/userscript.js
+++ b/addons/share-through-mystuff/userscript.js
@@ -1,5 +1,5 @@
 export default async function ({ addon, console, msg }) {
-  Scratch.MyStuff.ProjectThumbnailCollectionView.prototype.shared = function(project) {
+  Scratch.MyStuff.ProjectThumbnailCollectionView.prototype.shared = function (project) {
     // Scratch's implementation incorrectly calls fadeOut on the entire list
     if (this.model.options.collectionType === "notshared") {
       project.$el.fadeOut();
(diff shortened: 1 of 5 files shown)


WorldLanguages:
Tagging addon author @Steve0Greatness in case they are interested on taking a look (optional)

addons/share-through-mystuff/userscript.js
```diff
@@ -1,4 +1,11 @@
 export default async function ({ addon, console, msg }) {
+  Scratch.MyStuff.ProjectThumbnailCollectionView.prototype.shared = function (project) {
+    // Scratch's implementation incorrectly calls fadeOut on the entire list
+    if (this.model.options.collectionType === "notshared") {
+      project.$el.fadeOut();
+    }
+  };
+
```
WorldLanguages, line 8: Good catch here. I don't think we tested sharing projects with the "unsharing projects" tab selected.

WorldLanguages (approved):

Steve0Greatness (approved):

addons/share-through-mystuff/userscript.js
mxmou: The addon already includes a fix for this, but it's more complex than the one I added in this PR.

Samq64 (approved):

Issue #5764, opened 2023-03-01
New userscript docs available on the website


WorldLanguages:
There's 3 new pages available on the docs:

- About userscripts: https://scratchaddons.com/docs/develop/userscripts/
- Userscript Best Practices: https://scratchaddons.com/docs/develop/userscripts/best-practices/
- Userscript Debugging Tips: https://scratchaddons.com/docs/develop/userscripts/debugging-userscripts/

This are userscript docs, not addon.* API docs (they are separate). Any feedback is welcome. We hope to document the most common edge cases, implicit best practices, and basically anything that's good to now for both new and old contributors.
The userstyle docs will be remade to follow this pattern as well soon.

WorldLanguages:
Self-feedback: 

Would be nice to add a best-practice: avoid `x.parentElement.parentElement` in favor of `x.closest(selector)`.

Samq64:
I like how the the examples try to use code snippets from real addons, but wouldn't an addon that only has a userscript be a better option than using `scratch-messaging` for the example manifest?

WorldLanguages:
@Samq64 I agree - I think Scratch Messaging is mentioned there just for historical reasons.

WorldLanguages:
@Samq64 See https://github.com/ScratchAddons/website-v2/pull/310

PR #5826, opened 2023-03-19
Account info

StioStudio:
(the following might have typos)

Resolves that you don't have any easy way to get information about you.

### Changes
*Added the account-info addon. And added some words to the english version of the addon.
*The addon adds a page in the account settings
*The info page link is https://scratch.mit.edu/accounts/settings/#info

### Reason for changes
So that the user can easily see their information
The reason I have it as enabledByDefault is because it is a new page. And if someone links you to it, you will not see it

### Tests
I have tested it with oprea GX. It uses chromium. I cloned the repositery and added it to my extensions. Then I added my changes by using Visual Studio Code.
It have been tested on firefox too by callumjt


addons/account-info/addon.json
```diff
@@ -0,0 +1,30 @@
...
+        "matches": ["https://scratch.mit.edu/accounts/settings/"]
+      },
+      {
+        "url": "info.js",
+        "matches": ["https://scratch.mit.edu/accounts/password_change/"]
+      },
+      {
+        "url": "info.js",
+        "matches": ["https://scratch.mit.edu/accounts/email_change/"]
+      }
```
DNin01, line 24: Since these are arrays, these can be combined into one.
```suggestion
      {
        "url": "info.js",
        "matches": ["https://scratch.mit.edu/accounts/settings/", "https://scratch.mit.edu/accounts/password_change/", "https://scratch.mit.edu/accounts/email_change/"]
      }
```

Samq64: That could be simplified even more.
```suggestion
      {
        "url": "info.js",
        "matches": ["https://scratch.mit.edu/accounts/*"]
      }
```

WorldLanguages:
I think we should discuss the addon idea before doing code review @DNin01 @Samq64

WorldLanguages:
@StioStudio For any upcoming PRs please create an issue first, specially new addon PRs.
Personally I'm not convinced about the usefulness of an "account info" addon. Isn't this need already covered by websites like ScratchStats.com?

StioStudio:
> 

I think it is an good addon. You will be able to see a lot of info that you else can't see (not even on ScratchStats.com) and I thought to add more.

StioStudio:
And you will not need to go to other websites

StioStudio:
And because it it #info will it not break for non SA users

WorldLanguages:
> You will be able to see a lot of info that you else can't see (not even on ScratchStats.com) and I thought to add more.

Which pieces of information do you think would be the most useful? Specially those that aren't available in websites (such as ScratchStats.com) already.

> And you will not need to go to other websites

I don't think this page would be visited very often by users, so unlike addons like "profile statistics", I'm not sure if this would save users much time, unless there's useful information not available in public websites.

> And because it it #info will it not break for non SA users

Non SA users wouldn't see a 404 page, but they still wouldn't see what they're expecting. Also, we can discuss the implementation details after discussing the usefulness first.

StioStudio:
As you can see here: https://github.com/StioStudio/ScratchAddons/blob/master/addons/account-info/info.js#L195
You can see a lot of false and true things about your account. That can be useful

WorldLanguages:
@StioStudio None of those seems useful - and even if we find something that is, building an entire new addon to show a single value does not make sense.

1. If the user is banned, the page won't even load
2. Projects and studios don't let you type comments if you're muted. There's also a "message from Scratch Team" (AKA alert) that notifies you when you are muted or unmuted.
3. No admins use Scratch Addons, which makes sense for security reasons
4. All others can be seen in the user's own profile.

StioStudio:
Awwww ok. I mostly did this addon to learn, but I acctuly thought it could be cool. I guess I can just close then.

WorldLanguages:
No problem. Wish you the best in the process of learning JavaScript and browser extension development.
If you have any new addon/feature ideas, [create a new issue](https://github.com/ScratchAddons/ScratchAddons/issues/new/choose).

WorldLanguages:
Also, if you have any feedback regarding the addon docs (I assume you've read parts of it) feel free to comment on #5764

StioStudio:
I will use the addon my self were that extenstion has only that addon

StioStudio:
@WorldLanguages hey do you know why scratch addons chraches if I have the custom and the original on on the same time? And maybe how to fix it.

WorldLanguages:
@StioStudio At the moment we don't support running multiple instances at once.

StioStudio:
> @StioStudio At the moment we don't support running multiple instances at once.

Awww ok. Any idea how I could fix that?
Can I add my own to SA but still update without deleting my addon? (I am not to used with git)

Samq64:
> Awww ok. Any idea how I could fix that? Can I add my own to SA but still update without deleting my addon? (I am not to used with git)

You could try GitHub desktop.

PR #5829, opened 2023-03-19
Fix Comlink not defined in module.js (race condition)

### LINKED ISSUE #5753: Comlink not defined in module.js (race condition)
### Description

It's unusual, but theoretically possible, that module.js runs before libraries/comlink.js does.
No userscripts run when that happens.

We should make sure comlink.js has fully executed before module.js <script> gets added to the DOM at all, or we can make modifications to module.js to account for this case.

image

### Steps to reproduce

Be unlucky (like the  Discord user that reported this), or artificially add a setTimeout to the code (found in cs.js) that adds comlink.js <script> tag to the DOM.

_No response_

### Environment

Reproduced in Chrome by Discord feedback (race condition)

WorldLanguages:
Resolves #5753

### Changes

Converts comlink.js to a module dependency. It will now be impossible for our own code in module.js to run if comlink.js hasn't executed fully yet.

### Reason for changes

We need to ensure module.js code has access to `window.Comlink` when executing. The previous approach to load comlink.js did not ensure that in 100% of cases.
If you've experienced that userstyles run but userscripts don't, and it gets fixed after a reload, this was probably the root cause.

### Tests

Needs some more testing, since we're changing the execution environment of comlink.js to strict mode.

### DIFF
--- a/content-scripts/cs.js
+++ b/content-scripts/cs.js
@@ -123,10 +123,6 @@ const cs = {
 };
 Comlink.expose(cs, Comlink.windowEndpoint(comlinkIframe1.contentWindow, comlinkIframe2.contentWindow));
 
-const pageComlinkScript = document.createElement("script");
-pageComlinkScript.src = chrome.runtime.getURL("libraries/thirdparty/cs/comlink.js");
-document.documentElement.appendChild(pageComlinkScript);
-
 const moduleScript = document.createElement("script");
 moduleScript.type = "module";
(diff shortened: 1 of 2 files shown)


WorldLanguages:
> since we're changing the execution environment of comlink.js to strict mode.

As far as I can see, it's already strict mode through `'use strict'`

Samq64 (approved):

Issue #5431, opened 2022-12-08
Extend possible enabled addon states, not just true/false


WorldLanguages:
### Why this would be helpful/solve a problem

As of now, addons can only be enabled or disabled. When the user installs for the first time, or when an update adds new addons that didn't previously exist, we check at the value of `enabledByDefault` and that's all. After the extension is restarted, we have no idea if the user disabled an addon, or if the user never tried it out. There's little information we store on `chrome.storage`.

### What this change does

Instead of false/true, we could use the following possibilities:
1. `default`: the user didn't explicitly opt-in or opt-out into the feature. The value of `enabledByDefault` on the manifest would be respected. We might want to add a new possible property on addon manifests named `enabledOnInstall` to recover the old behavior of `enabledByDefault` in case we need it (for example, in case we want to enable an addon by default for new users, but not for existing ones).
2. `disabled`: the addon is explicitly disabled.
3. `enabled`: the addon is explicitly enabled.
4. `recommendedEnabled`: the user enabled a feature (which doesn't exist yet) such as "automatically enable all recommended addons", and a recommended addon wasn't previously enabled manually. This would allow us to show these addons in a different way in the settings page, so the user can know the addon has been automatically enabled, and not manually enabled.

The concept of `recommendedEnabled` could be expanded for other concepts, such as "enabled because of a superpreset", or it could be extended for other addon tags.

Additionally, no matter the enabled state stored in `chrome.storage`, if we add the ability to disable addons remotely for all users (#5248), the addon should be logically disabled. There are also other reasons we might want to allow addons to be logically enabled/disabled during the span of the session (in all tabs or a specific one) such as addon tours (#5407)

It may also be useful to store the date where the user last changed the enabled value of an addon. We could also store a log of historical changes (instead of just the most recent one per each addon), but I'll leave that for another issue. Knowing whether a user recently tested out an addon can be valuable information at the time of considering what addons to suggest to the user.

### Possible alternatives

There's a lot of possible discussion here on different ways we could do this.
Any alternative is probably better than the current true/false model.

### Demonstration(s) and additional context

This concept could also be extended to addon settings, potentially.

DNin01:
I think it would be cool to see this help us with "personalized" addon suggestions and recently used lists. But also, IMO, users should be aware of the fact that the addons they're using are being tracked, even if Scratch Addons doesn't send any data to databases or servers for analytics or other things.

BroJac5246:
[...]

PR #5831, opened 2023-03-19
Add AprilFools2023 addon: "featured dangos"

### LINKED ISSUE #5807: 🔴 April Fools: Discussing our own easter egg
2021: Dango rain (#1931)
2022: Scratch Avocados (#4436)
2023: (?) Hopefully dangos again? We love dangos!

\
Some suggestions found on the Discord server from various authors:
- Merging #3991 
- Hide little dancing pixel Scratch cats in various places on the settings page, popup, Scratch website, etc. And locally track how many the user has found and clicked on
- "Featured Dangos" row in front page. We could feature multiple dango related projects that already exist, from various Scratchers.
- Replace the pause button icon with the dango icon. I don't think it would be fine to replace some native Scratch elements, but we can do it with our own.

WorldLanguages:
Resolves #5807

### Changes

New addon that adds a "Featured Dangos" row to the front page. It includes 5 hand-picked projects.
It will be enabled by default, but it won't be visible in the settings page unless it's April Fools.

image

TODOs:

- [x] Addon name / description
- [x] Hide addon from settings page unless current timestamp matches April Fools
- [x] Early return in userscript if current timestamp doesn't match April Fools

### Reason for changes

This idea appeared to receive positive feedback in the April Fools discussion issue.

### Tests

Works correctly in Chromium
The layout doesn't work very well in mobile portrait view.

### DIFF
--- a/addons/featured-dangos/addon.json
+++ b/addons/featured-dangos/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Featured Dangos (April Fools 2023)",
-  "description": "(description TODO)",
+  "description": "Adds some dango-related projects to the front page.",
   "userscripts": [
     {
       "url": "userscript.js",
--- a/addons/featured-dangos/addon.json
+++ b/addons/featured-dangos/addon.json
@@ -14,5 +14,6 @@
     }
   ],
   "tags": ["community"],
+  "versionAdded": "1.31.1",
   "enabledByDefault": true
 }
--- a/addons/featured-dangos/addon.json
+++ b/addons/featured-dangos/addon.json
@@ -13,7 +13,7 @@
       "matches": ["https://scratch.mit.edu/"]
     }
   ],
-  "tags": ["community"],
+  "tags": ["community", "easterEgg", "featured"],
   "versionAdded": "1.31.1",
   "enabledByDefault": true
 }
--- a/addons/featured-dangos/addon.json
+++ b/addons/featured-dangos/addon.json
@@ -1,0 +1,18 @@
+{
+  "name": "Featured Dangos (April Fools 2023)",
+  "description": "(description TODO)",
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/"]
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/"]
+    }
+  ],
+  "tags": ["community"],
+  "enabledByDefault": true
+}
--- a/addons-l10n/en/featured-dangos.json
+++ b/addons-l10n/en/featured-dangos.json
@@ -1,0 +1,4 @@
+{
+  "featured-dangos/row-title": "Featured Dangos",
+  "featured-dangos/note": "Added by Scratch Addons browser extension"
+}
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -136,6 +136,7 @@
   "number-pad",
   "copy-reporter",
   "share-through-mystuff",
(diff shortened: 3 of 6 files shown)


WorldLanguages:
I remember it was possible to make the front page crash when modifying its DOM (probably a React bug), so we want to make sure that's not the case.
I have tested logging in, clicking around in every possible button, scrolling every possible container, and there were no errors. But I might have missed something.

WorldLanguages:
Feedback about the 5 projects I chose is also welcome.

WorldLanguages:
Other ideas:

- Instead of hardcoding the projects, fetch from a studio
- Choose 5 random projects on each load, instead of selecting only five

addons/featured-dangos/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Featured Dangos (April Fools 2023)",
+  "description": "(description TODO)",
```
DNin01, line 3:
```suggestion
  "description": "Adds some dango-related projects to the front page.",
```

addons/featured-dangos/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Featured Dangos (April Fools 2023)",
```
DNin01, line 2: I'm pretty sure the plural for dango is the same as the singular.

DNin01 (commented):
Tested and looks okay to me.

mxmou:
> I remember it was possible to make the front page crash when modifying its DOM (probably a React bug), so we want to make sure that's not the case.

Removing an element controlled by React causes a crash - this is the case everywhere, not just on the front page. Adding elements shouldn't cause any issues.

MaterArc:
Are the projects also going to include the users that made the project. I think we should also include that to give them the proper credit.

addons/featured-dangos/addon.json
WorldLanguages: It's hard to know because it's a foreign word. I've seen the plural "dangos" in many places. There's thousands of results from Reddit posts: https://www.google.com/search?q=%22dangos%22+site:www.reddit.com

DNin01: It's like "emoji"/"emojis".

Either is fine with me, then.

WorldLanguages:
I'm sorry, some commits will have a wrong date because I changed my system's datetime to test them.

WorldLanguages:
This looks ready to merge! Any comments?

Samq64:
The addon doesn't work, probably because of https://scratch.mit.edu/discuss/topic/667012/

DNin01:
> The addon doesn't work, possibly a Scratch bug since half the rows are missing even with SA disabled?

It performs a date check and says visiting scratch.mit.edu/#dangos will bypass the date check.

DNin01 (approved):

Samq64:
> It performs a date check and says visiting scratch.mit.edu/#dangos will bypass the date check.

I see. But if the user manually enabled it, I think it should work normally.

Samq64 (approved):

DNin01:
> > It performs a date check and says visiting scratch.mit.edu/#dangos will bypass the date check.
> 
> I see. But if the user manually enabled it, I think it should work without adding `#dangos` any time of year.

I think we might change this behavior in the future.

WorldLanguages:
> I see. But if the user manually enabled it, I think it should work without adding `#dangos` any time of year.

This addon is enabled by default in order to avoid more complex code that enables the addon in a specific date. So currently there's no way to know whether the user manually enabled it. Related: #5431

WorldLanguages:
See #5891

PR #5838, opened 2023-03-22
`pause`: Automatically pause project when the page is not focused

### LINKED ISSUE #5836: Auto-pause on focus leave
### Why this would be helpful

Many users would like the project to be paused when they enter another window, e. g. to don't lose progress in runner games.

### My suggestion

Add an option to the `pause` addon to pause the project when the tab is not focused, possibly with dynamic change so that users can quickly toggle it without having to reload.

### Possible alternatives

_No response_

### Additional context

_No response_

iqnite:
Resolves #5836 

### Changes
Add an option to the pause addon to pause the project when the tab is not focused.

### Reason for changes
Many users would like the project to be paused when they enter another window, e. g. to not lose progress in runner games.

### Tests
~~It only pauses once when you change the addon's settings. Can anyone look at the code?~~

- [x] Basic addon functionality
- [x] Auto-pause (fixed thanks to @Samq64)
- [x] Auto-pause in the editor
- [x] Dynamic enable support
- [x] Dynamic disable support


addons/pause/userscript.js
```diff
@@ -15,6 +15,12 @@ export default async function ({ addon, console, msg }) {
   setSrc();
   onPauseChanged(setSrc);
 
+  addon.settings.addEventListener("change", () => {
+    console.log("Settings changed!");
+    if (addon.settings.get("auto-pause") === true) document.addEventListener("focusout", setPaused(false));
```
Samq64, line 20: The setting change event is only triggered when  settings change after the addon is already loaded. Also, you probably meant to run `setPaused(true)` here, false will un-pause the project.

iqnite: Oh I misunderstood the method. Thanks!

WorldLanguages:
Should the project automatically unpause when the user comes back to the tab?

addons/pause/addon.json
```diff
@@ -24,6 +24,14 @@
       "matches": ["projects", "projectEmbeds"]
     }
   ],
+  "settings": [
+    {
+      "type": "boolean",
+      "id": "auto-pause",
+      "name": "Pause when page loses focus",
```
DNin01, line 31: This setting named like this may be a little unclear to the average user, but I can't think of anything better right now.

BroJac5246: > This setting named like this may be a little unclear to the average user, but I can't think of anything better right now.

"Pause project when browser tab changes [from Scratch]?"

DNin01: > > This setting named like this may be a little unclear to the average user, but I can't think of anything better right now.
> 
> "Pause project when browser tab changes [from Scratch]"?

What if the user focuses a different _window_?

DNin01: We could potentially go with "pause when project page is minimized".

BroJac5246: > > > This setting named like this may be a little unclear to the average user, but I can't think of anything better right now.
> > 
> > "Pause project when browser tab changes [from Scratch]"?
> 
> What if the user focuses a different _window_?

Since we're avoiding the word 'focus,' I'm not sure there's much we can do about that. I think this gets the point across well enough, but I don't particularly care.

WorldLanguages: "Pause when leaving tab"?

DNin01: > "Pause when leaving tab"?

Concise, but then users might think that means closing the tab or clicking a link to a different page (which if that were what it did, wouldn't be very helpful).

Samq64:
> Guys, this addon has bigger problems than the wrong name. It doesn't work in first place.

Try adding
```js
if (addon.settings.get("auto-pause")) document.addEventListener("focusout", setPaused(true));
```
above the settings change event listener.

addons/pause/addon.json
WorldLanguages: @iqnite Both things can be dealt with in parallel.

WorldLanguages:
@Samq64 You meant `() => setPaused(true)` instead of just `setPaused(true)`

Samq64:
> @Samq64 You meant `() => setPaused(true)` instead of just `setPaused(true)`

Yeah, thanks for catching that.

iqnite:
> @Samq64 You meant `() => setPaused(true)` instead of just `setPaused(true)`

What's the difference?

BroJac5246:
> > @Samq64 You meant `() => setPaused(true)` instead of just `setPaused(true)`
> 
> What's the difference?

One should work; the other doesn't.

addons/pause/userscript.js
```diff
@@ -15,6 +15,14 @@ export default async function ({ addon, console, msg }) {
   setSrc();
   onPauseChanged(setSrc);
 
+  if (addon.settings.get("auto-pause")) document.addEventListener("focusout", () => setPaused(true));
```
Samq64, line 18: I got this working by using `window` and `blur` instead of `document` and `focusout`.
```suggestion
  if (addon.settings.get("auto-pause")) window.addEventListener("blur", () => setPaused(true));
```

addons/pause/userscript.js
```diff
@@ -15,6 +15,14 @@ export default async function ({ addon, console, msg }) {
...
+  addon.settings.addEventListener("change", () => {
+    console.log("Settings changed!");
+    if (addon.settings.get("auto-pause") === true) document.addEventListener("focusout", () => setPaused(true));
+    else if (addon.settings.get("auto-pause") === false)
+      document.removeEventListener("focusout", () => setPaused(false));
```
Samq64, line 23: I don't think putting an else on the next line like that works. You'll probably have to add the braces to the condition or use a [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator).

WorldLanguages: > I don't think putting the else on the next line without braces works. 

I don't see why it wouldn't work.

addons/pause/addon.json
iqnite: What about "Auto-pause when switching windows"?

BroJac5246: > What about "Auto-pause when switching windows"?

But then there's switching tabs.

iqnite: > But then there's switching tabs.

It's technically like switching tabs, and I know browsers that don't even support tabs. Also, many sites talk about "windows" meaning tabs and even Griffpatch says "Close all browser windows and retry".

BroJac5246: > > But then there's switching tabs.
> 
> It's technically like switching tabs, and I know browsers that don't even support tabs. Also, many sites talk about "windows" meaning tabs and even Griffpatch says "Close all browser windows and retry".

Yes, but all of the browsers we officially support do have tabs, and Griffpatch is referring to windows, not tabs.

Chiroyce1: >@iqnite: and even Griffpatch says "Close all browser windows and retry".

Well thats because closing all windows indirectly closes all tabs

Samq64 (commented):
Should the project automatically resume when the setting is dynamically disabled?

iqnite:
> Should the project automatically resume when the setting is dynamically disabled?

No. That would only add extra complication. And if the user manually pauses and then disables the setting, the project would resume, even if it shouldn't.

addons/pause/addon.json
WorldLanguages: Oh boy, I don't think we found a clear way to name this setting 🤦‍♂️

BroJac5246: > Oh boy, I don't think we found a clear way to name this setting 🤦‍♂️ 

Since the most accurate is "Automatically pause when tab loses focus," I prefer that one.

iqnite: Look at the actual one please.

Samq64 (approved):

iqnite:
@DNin01 your review adds a new perspective I didn't think about. Maybe we should add an option to prevent pause for projects with cloud variables, maybe even mitigations for the addon itself (not just this option)?

DNin01:
> maybe even mitigations for the addon itself (not just this option)?

Indeed, we could consider entirely disabling the pause function on these types of projects.

iqnite:
> Indeed, we could consider entirely disabling the pause function on these types of projects.

I wouldn't say disabling though. People with the knowledge could still bypass the restriction (via devtools). Maybe we should add a "Allow pausing on projects with cloud variables" setting and mark it as Dangerous?

DNin01:
Whaddya say we just don't auto-pause if cloud data activity (send/receive) occurred on the cloud WebSocket connection in the last 15 seconds? That should mitigate most of the issues potentially introduced by pausing those projects.

iqnite:
> don't auto-pause if cloud data activity (send/receive) occurred on the cloud WebSocket connection in the last 15 seconds?

Sounds good, maybe also add a notice in the settings.

DNin01:
> Whaddya say we just don't auto-pause if cloud data activity (send/receive) occurred on the cloud WebSocket connection in the last 15 seconds? That should mitigate most of the issues potentially introduced by pausing those projects.

Done in #8555.

PR #5844, opened 2023-03-24
Improve profile icon from "search profile" addon

Hexagonl:
### Changes
Improved the Search Profile addon by cleaning up the icon used.

- Fixed stroke width to be 1px instead of 1.333px, which was inconsistent with this icon's vanilla counterparts (Projects and Studios).
- Used strokes instead of fill to be more consistent with this icon's counterparts.
- Changed HEX color from #626262 to #6B6B6B to be more consistent with this icon's counterparts.
- Fixed bits poking out at the bottom (by using rounded stroke caps instead).
- Condensed SVG code from 1319 bytes to 301 bytes.
- Added my name to the list of contributors.

### Reason for changes
- Line thickness of icon was inconsistent with its counterparts.
- Color of icon was slightly inconsistent with its counterparts.

### Tests
Works correctly in Firefox 111.0 and 111.0. 

### Before and After
before
after


WorldLanguages:
Thanks for the detailed PR body!

WorldLanguages (approved):
Change looks good to me.

It appears that the previous version of icon was chosen by me in the commit link below, without much thought, so I appreciate you're suggesting a better alternative: https://github.com/Geotale/ScratchAddons/commit/45d1f44ee75a599eabc2d52d772a41d55f1a43fd

mxmou (approved):

Issue #5281, opened 2022-10-28
Official way to test SA with local scratch-www instance


WorldLanguages:
### Why this would be helpful/solve a problem

It would be easier to test upcoming changes to the Scratch website and editor.

### What this change does

We may have to do some changes to the code, or possibly not, as long as we have documentation that lists step-by-step how to test SA with a local scratch-www instance.
Of course, you could add 127.0.0.1 to /etc/hosts but we can probably do better.

### Possible alternatives

Documenting the way @mxmou has been doing it until now, without any code changes. I'm not sure if he has ever publicly stated it (if so, please send a link, if not please say no)

### Demonstration(s) and additional context

#5248 went offtopic and I moved the conversation here.
Relevant comment: https://github.com/ScratchAddons/ScratchAddons/issues/5248#issuecomment-1290808909

mxmou:
> Documenting the way @mxmou has been doing it until now, without any code changes. I'm not sure if he has ever publicly stated it (if so, please send a link, if not please tell us how you've been doing it)

I had to make changes to the code because it expects to be running on `scratch.mit.edu` in a lot of places. I also used [this extension](https://github.com/einaregilsson/Redirector) to redirect `localhost:8333/session` to a data URI.

WorldLanguages:
> I had to make changes to the code because it expects to be running on `scratch.mit.edu` in a lot of places.

Do you have a list of those code changes? Or is it outdated?
If it would be better than starting all over again, please send a list of changes or submit a PR.
I think it's good to make sure addons will continue to work even if the URL is not exactly "scratch.mit.edu", for example, it would mean SA would work in a proxied version of Scratch.

mxmou:
> Do you have a list of those code changes?

* In the extension manifest, I added `http://localhost/*` as a content script match and a host permission.
* `background/get-userscripts.js` listens to two `webRequest` events (line [240](https://github.com/ScratchAddons/ScratchAddons/blob/master/background/get-userscripts.js#L240) and [273](https://github.com/ScratchAddons/ScratchAddons/blob/master/background/get-userscripts.js#L273)) with the filter `urls: ["https://scratch.mit.edu/*"]`.
* The function [`userscriptMatches()`](https://github.com/ScratchAddons/ScratchAddons/blob/master/background/get-userscripts.js#L396-L431) in the same file checks if the origin is `scratch.mit.edu`. That probably isn't necessary - the old implementation (before #2305) ignored the origin.
[...]

PR #5848, opened 2023-03-25
Handle Scratch totally-normal mode (April Fools 2023)

### LINKED ISSUE #5806: 🔴 April Fools: Compatibility with upcoming Scratch editor update
**See PR in LLK repo: [feat: bring back cat blocks](https://github.com/LLK/scratch-gui/pull/x/../8811/)**

It looks like it will be pretty much identical to last year's editor April Fools. So our fix should be similar to our PR #4404 from last year, excluding the "world stage mode" code (that feature didn't reach users last year either).

**Fixes available at PR #5848**

WorldLanguages:
Resolves #5806 
Resolves part of #5281 since it's closely related

*Many code changes from this PR are inspired by or copied from last year's April Fools bug fixes PR: #4404*

### Changes

- In order to debug, addons can run on https://llk.github.io/scratch-gui/develop/. It will be removed before merging.
- Enables April Fools mode when going to https://llk.github.io/scratch-gui/develop/ - note that the "mode" button is not available without the changes to fix-console.js made in this branch
- Changes various parts of the code that expect to be running only in scratch.mit.edu. Related to #5281 
- Adds `totally-normal` class to the \<body> element if appropriate, just like last year

Pending changes and TODOs:

- [x] Make "cat blocks" addon do nothing if editor is running in totally-normal mode
- [ ] Check whether other editor addons don't work well or crash the editor when changing to cat mode
- [ ] Remove debug code before merging

### Reason for changes

See #5806 

### Tests

Not tested extensively yet.
Priority number one is that we disable any addons that would crash the editor if the user uses the totally-normal features from the Scratch editor.


WorldLanguages:
Apparently the glitch that I thought was from hide-flyout is actually a vanilla Scratch glitch:

1. Go to https://llk.github.io/scratch-gui/develop/ with no addons enabled (but you need the extension enabled to get the "mode" menu)
2. Go to the "costumes" or "sounds" tab (it's indifferent)
3. Enable caturday mode for the first time
4. Come back to the "code" tab

image

BroJac5246:
When will this be merged? April Fools' is _very_ close, and we also have to go through the webstore approval processes, right?

Samq64:
With all addons enabled, there aren't any editor crashes when enabling cat mode, but some addon styles break, but that could just be a develop branch issue.

addon-api/content-script/Tab.js
```diff
@@ -823,4 +824,9 @@ export default class Tab extends Listenable {
   prompt(title, message, defaultValue, opts) {
     return modal.prompt(this, title, message, defaultValue, opts);
   }
+  isScratchAprilFools23() {
+    if (!this.redux.state) return true; // better safe than sorry
+    if (typeof this.redux.state?.scratchGui?.timeTravel === "object") return true;
+    return false;
+  }
```
WorldLanguages, line 831: Here, it would have been more suitable to check for the session flag:
```js
addon.tab.redux.state.session.session.flags.everything_is_totally_normal
=== true
```
The code that shipped to users unintentionally disabled addons before April Fools because the timeTravel object is available, but not accessible through the UI

WorldLanguages: This mistake costed a release

PR #5850, opened 2023-03-26
New Feature: Old Shaped Categories

WildmonkGitHub:
Resolves # This will bring the nostalgic of Scratch 2.0, because there are so many options in SA to make Scratch 3 look like Scratch 2. I've added one more feature that changes the shape of categories from round to a thin square.
Screenshot 2023-03-26 135303

### Changes

This feature is pre-release.

### Reason for changes

I found this idea when I saw some tutorials that how to make Scratch 3 look like Scratch 2 using SA, so I notice that there is one thing missing, and I made that - "Old Shaped Categories".

### Tests

I have tested this, there is no issue.

### DIFF
--- a/addons/shape-categories/addon.json
+++ b/addons/shape-categories/addon.json
@@ -1,7 +1,7 @@
 {
     "name": "Old Shaped Categories",
     "description": "Removes the round shape of categories and change into thin square, just like Scratch 2.0",
-    "tags": ["editor", "theme", "category"],
+    "tags": ["editor", "theme"],
     "credits": [
       {
         "name": "Wildmonk",
--- a/addons/shape-categories/addon.json
+++ b/addons/shape-categories/addon.json
@@ -1,5 +1,5 @@
 {
-    "name": "Old Shaped Categories",
+    "name": "Old shaped categories",
     "description": "Removes the round shape of categories and change into thin square, just like Scratch 2.0",
     "tags": ["editor", "theme"],
     "credits": [
@@ -11,7 +11,7 @@
     "enabledByDefault": false,
     "dynamicDisable": true,
     "dynamicEnable": true,
-    "versionAdded": "1.31.1",
+    "versionAdded": "1.31.0",
     "userstyles": [
       {
         "url": "userstyle.css",
--- a/addons/shape-categories/addon.json
+++ b/addons/shape-categories/addon.json
@@ -1,0 +1,21 @@
+{
+    "name": "Old Shaped Categories",
+    "description": "Removes the round shape of categories and change into thin square, just like Scratch 2.0",
+    "tags": ["editor", "theme", "category"],
+    "credits": [
+      {
+        "name": "Wildmonk",
+        "link": "https://scratch.mit.edu/users/Wildmonk-2"
+      }
+    ],
+    "enabledByDefault": false,
+    "dynamicDisable": true,
+    "dynamicEnable": true,
+    "versionAdded": "1.31.1",
+    "userstyles": [
+      {
+        "url": "userstyle.css",
+        "matches": ["projects"]
+      }
+    ]
+}
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -136,6 +136,7 @@
   "number-pad",
   "copy-reporter",
   "share-through-mystuff",
+  "shape-categories",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/shape-categories/userstyle.css
+++ b/addons/shape-categories/userstyle.css
@@ -1,0 +1,6 @@
+.scratchCategoryItemBubble {
+    border-radius:0cm;
+    width:0.25cm;
+    height:0.68cm;
+    border: none;
+}


addons/shape-categories/addon.json
```diff
@@ -0,0 +1,21 @@
...
+    "credits": [
+      {
+        "name": "Wildmonk",
+        "link": "https://scratch.mit.edu/users/Wildmonk-2"
+      }
+    ],
+    "enabledByDefault": false,
+    "dynamicDisable": true,
+    "dynamicEnable": true,
+    "versionAdded": "1.31.1",
```
DNin01, line 14: We usually add addons in major versions (e.g., 1.32.0) only.

addons/shape-categories/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "Old Shaped Categories",
```
DNin01, line 2: Minor detail, all of the other addons (with a few exceptions) don't start every word with an uppercase letter. They actually look like this:
```suggestion
    "name": "Old shaped categories",
```

DNin01 (commented):
This seems to work best in combination with the two-column category menu (`addons/columns`), so maybe this would work better as a setting for that addon instead of a new addon, unless you can find a way to make it compatible with that addon _and_ the normal design.

Be sure to test the addon with related addons enabled _and_ all other addons such as two-column category menu disabled.

WorldLanguages:
> This seems to work best in combination with the two-column category menu (`addons/columns`), so maybe this would work better as a setting for that addon instead of a new addon

I initially agree with this

Samq64:
What about the category icons addon?

WorldLanguages:
> What about the category icons addon?

We could make it so that this new addon/setting overrides that addon (overriding the other way wouldn't make much sense).

WorldLanguages:
We could also consider renaming `block palette category icons` and providing a setting on it that lets the user choose between the icons and this "old shaped thin square".

addons/shape-categories/addon.json
WildmonkGitHub: I have changed it.

addons/shape-categories/addon.json
WildmonkGitHub: Done with changes

WildmonkGitHub:
> This seems to work best in combination with the two-column category menu (`addons/columns`), so maybe this would work better as a setting for that addon instead of a new addon, unless you can find a way to make it compatible with that addon _and_ the normal design.
> 
> Be sure to test the addon with related addons enabled _and_ all other addons such as two-column category menu disabled.

I have tested it, but it looks like it would not look good when you turn on (`block palette category icons`) while using this feature.

DNin01:
> it looks like it would not look good when you turn on (`block palette category icons`) while using this feature.

That should definitely be considered.

Samq64:
> We could also consider renaming `block palette category icons` and providing a setting on it that lets the user choose between the icons and this "old shaped thin square".

I think that's he best solution.

WildmonkGitHub:
> That should definitely be considered. 

So, this feature will be not accepted unless I just fix the block palette category icon bug while using this feature.

DNin01 (changes requested):
> > That should definitely be considered.
> 
> So, this feature will be not accepted unless I just fix the block palette category icon bug while using this feature.

Probably. It seems like a good thing to support. We don't really want one addon to conflict with another.

We could probably just move the icon _next_ to the category's color instead of inside it, only when this addon is enabled, of course.

Also, all by itself, it looks like this:

Colored bars replace the circles indicating category colors

I would expect the bars to be placed further to the left. This is why it's good to consider how an addon looks and behaves without the need for other addons. If we don't want to do that, then we could also consider moving this feature to another addon that it works well with (in this case, Two-column category menu), as I mentioned as an option earlier.

So, those are the things we need to do before we wrap up.

Secret-chest:
> > > That should definitely be considered.
> > 
> > So, this feature will be not accepted unless I just fix the block palette category icon bug while using this feature.
> 
> Probably. It seems like a good thing to support. We don't really want one addon to conflict with another.
> 
> We could probably just move the icon _next_ to the category's color instead of inside it, only when this addon is enabled, of course.
> 
> Also, all by itself, it looks like this:
> 
> Colored bars replace the circles indicating category colors
> 
> I would expect the bars to be placed further to the left. This is why it's good to consider how an addon looks and behaves without the need for other addons. If we don't want to do that, then we could also consider moving this feature to another addon that it works well with (in this case, Two-column category menu), as I mentioned as an option earlier.
> 
> So, those are the things we need to do before we wrap up.

How about making it more compact (colours always on the same line?)

WorldLanguages:
@WildmonkGitHub Why was this closed?

WorldLanguages:
🤔

WildmonkGitHub:
> 🤔

I don't think that it is a nice idea. I better work on other thing that has no problems.

WorldLanguages:
I think the feature idea is nice. I guess you mean the implementation had problems.

WildmonkGitHub:
> I think the feature idea is nice. I guess you mean the implementation had problems.

Better I work on next, this UI is not looking good in my opinion.

PR #5852, opened 2023-03-27
Find bar: make some strings translatable

### LINKED ISSUE #5822: find-bar: "event", "var", "list" should be translatable strings
### Why this would be helpful

`find-bar` adds the following prefixes to variable and broadcast names:
* "event" for broadcasts
* "var" and "list" for local variables and lists
* "VAR" and "LIST" for global variables and lists

These currently aren't translated.

### My suggestion

Make them translatable.

### Possible alternatives

_No response_

### Additional context

_No response_

iqnite:
Resolves #5822 

### Changes

`find-bar`: Make `event`, `var`, and `list` accessible through the l10n file to make them translatable. That should also improve consistency, as they're not hard-coded in the source.

### Reason for changes

For most languages, the current labels are OK, but for some, they're unclear (e.g. "Ereignis" is the German word for "event").

### Tests

*Tested on Vivaldi 7.5.2921.*

**Problems:**

- [x] Clicking on them doesn't search for all instances, it only jumps to the declaration in the block panel instead.
- [x] The items in the find bar dropdown don't have the right color.
- [x] It still forces a strict word order, will try adding {name} to l10n file.


addons/find-bar/userscript.js
```diff
@@ -305,17 +305,25 @@ export default async function ({ addon, msg, console }) {
 
       let vars = map.getVariablesOfType("");
       for (const row of vars) {
-        addBlock(row.isLocal ? "var" : "VAR", (row.isLocal ? "var " : "VAR ") + row.name, row);
+        addBlock(
+          row.isLocal ? msg("var") + " " : msg("VAR") + " ",
+          (row.isLocal ? msg("var") + " " : msg("VAR") + " ") + row.name,
+          row
+        );
```
mxmou, line 312: `addBlock`'s first parameter is a class name, which is used to style the item and doesn't affect the displayed text. Only the second one should be translated.

addons/find-bar/userscript.js
```diff
@@ -305,17 +305,25 @@ export default async function ({ addon, msg, console }) {
...
       }
 
       let lists = map.getVariablesOfType("list");
       for (const row of lists) {
-        addBlock(row.isLocal ? "list" : "LIST", (row.isLocal ? "list " : "LIST ") + row.name, row);
+        addBlock(
+          row.isLocal ? msg("list") + " " : msg("LIST") + " ",
+          (row.isLocal ? msg("list") + " " : msg("LIST") + " ") + row.name,
+          row
+        );
```
mxmou, line 321: Same as above.

CST1229:
Also, I'd say the string names should be `var-local` and `var-global` instead of `var` and `VAR`, for clarity (and possibly also add the variable name as a string input, like `variable {name}`, so translators can put the text after the name if they need to).

addons/find-bar/userscript.js
iqnite: Oh, thanks!

addons-l10n/en/find-bar.json
```diff
@@ -1,4 +1,9 @@
 {
   "find-bar/find-placeholder": "Find (Ctrl+F)",
-  "find-bar/complex-broadcast": "(expression)"
+  "find-bar/complex-broadcast": "(expression)",
+  "find-bar/var": "variable",
+  "find-bar/VAR": "VARIABLE",
```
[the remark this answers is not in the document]
mxmou: In the current version of SA, the prefix for variables is just "var". I don't think it should be changed to "variable".

Samq64 (approved):

addons-l10n/en/find-bar.json
```diff
@@ -1,4 +1,9 @@
 {
   "find-bar/find-placeholder": "Find (Ctrl+F)",
-  "find-bar/complex-broadcast": "(expression)"
+  "find-bar/complex-broadcast": "(expression)",
+  "find-bar/var-local": "variable {name}",
+  "find-bar/var-global": "VARIABLE {name}",
+  "find-bar/list-local": "list {name}",
+  "find-bar/list-global": "LIST {name}",
```
WorldLanguages, line 7: Self TODO: add translation notes here for languages that don't distinguish lowercase and uppercase. They could explicitly translate the strings as "local variable" or "global variable" so that the distinction remains.

WorldLanguages: Please tag me if I haven't looked into this after a few days

DNin01: @WorldLanguages Here's your reminder.

addons-l10n/en/find-bar.json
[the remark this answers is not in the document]
iqnite: It would make it clearer and easier to translate.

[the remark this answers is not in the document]
WorldLanguages: @iqnite Could we keep the strings intact for the scope of this PR? See the other conversation below - I'm going to add translator notes. Other languages may choose to use the full words if they want to. But I don't see why the English words need to change.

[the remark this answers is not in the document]
iqnite: OK, OK, I'm changing them back.

WorldLanguages:
@iqnite Does everything look good?

iqnite:
> @iqnite Does everything look good?

I think yes. I use it daily and until yet I didn't find any problems. After all, it was just a text change.

addons/find-bar/userscript.js
```diff
@@ -285,7 +285,7 @@ export default async function ({ addon, msg, console }) {
         if (root.type === "event_whenbroadcastreceived") {
           const fieldRow = root.inputList[0].fieldRow;
           let eventName = fieldRow.find((input) => input.name === "BROADCAST_OPTION").getText();
-          addBlock("receive", "event " + eventName, root).eventName = eventName;
+          addBlock("receive", msg("event") + " " + eventName, root).eventName = eventName;
```
WorldLanguages, line 288: I think you forgot to pass `{name: eventName}` here.

iqnite: Thanks!

WorldLanguages (approved):

PR #5869, opened 2023-04-02
New Feature: `Favourite addons`

### LINKED ISSUE #1250: Ability to make selected addons always show fist in the popup.

There are some addons (2.0 → 3.0, dark mode) that I want to be able to easily enable or disable, which can be harder if many addons are running on the current page so the list gets long.

Ability to "pin" addons to the top of the addons tab.

Platformer444:
Resolves #1250 

### Changes
This enhancement allows certain addons to be favourited and be always show on the top.

### Reason for changes
Favouriting Addons allows easy enabling and disabling of the addon.

### Tests
Tested on Chrome 111 and it works fine.


images/icons/star-filled.svg
fractioneater, line 1: That right "arm" is a bit off.
image

WorldLanguages:
This is going in a good direction, but it will need many more changes and questions to reach users.
Firstly, we would store the "favorited" addons in chrome.storage, not localStorage. Should it be chrome.storage.sync or chrome.storage.local? I'm not sure. Would import/export also store the list of favorited addons?

WorldLanguages:
Also, we use American English when naming variables, so multiple instances of "favourite" will need to get the U removed.

Platformer444:
> This is going in a good direction, but it will need many more changes and questions to reach users. Firstly, we would store the "favorited" addons in chrome.storage, not localStorage. Should it be chrome.storage.sync or chrome.storage.local? I'm not sure. Would import/export also store the list of favorited addons?

I think `chrome.storage.sync` would be a better option as addon settings are also saved in it.

Platformer444:
> > This is going in a good direction, but it will need many more changes and questions to reach users. Firstly, we would store the "favorited" addons in chrome.storage, not localStorage. Should it be chrome.storage.sync or chrome.storage.local? I'm not sure. Would import/export also store the list of favorited addons?
> 
> I think `chrome.storage.sync` would be a better option as addon settings are also saved in it.

@WorldLanguages

TheColaber:
this needs discussion

PR #5870, opened 2023-04-02
change canvas color button in costume editor

callumjt:
ive added a button in the costume editor to change the canvas color.

this should be added as its easier to make a sprite if its light or dark

tested on firefox

sorry if there are bugs im still very new to js

preview:
[https://cdn.discordapp.com/attachments/1089456285910573066/1092077085213859940/2023-04-02_14-18-56.mp4](url)

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -136,7 +136,7 @@
   "number-pad",
   "copy-reporter",
   "share-through-mystuff",
-  "featured-dangos",
+  "scene-color",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/scene-color/addon.json
+++ b/addons/scene-color/addon.json
@@ -1,0 +1,50 @@
+{
+    "name": "Scene color",
+    "description": "Adds a button to change the canvas color in the editor.",
+    "tags": ["projects"],
+    "credits": [
+      {
+        "name": "callumjt",
+        "link": "https://scratch.mit.edu/users/callumjt/"
+      }
+    ],
+    "userstyles": [
+        {
+            "url": "userstyle.css",
+            "matches": ["projects"]
+        }
+    ],
+    "userscripts": [
+        {
+            "url": "userscript.js",
+            "matches": ["projects"]
+        }
+    ],
+    "settings": [
+        {
+            "id": "default",
+            "name": "default color",
+            "type": "select",
+            "default": "black",
+            "potentialValues": [
+                {
+                    "id": "black",
+                    "name": "Black"
+                },
+                {
+                    "id": "white",
+                    "name": "White"
+                }
+            ]
+        }
+    ],
+    "customCssVariables": [
+        {
+            "name": "color",
+            "value": "default"
+        }
+    ],
+    "enabledByDefault": false,
+    "dynamicEnable": true,
+    "dynamicDisable": true
+}
--- a/addons/scene-color/addon.json
+++ b/addons/scene-color/addon.json
@@ -46,5 +46,6 @@
     ],
     "enabledByDefault": false,
     "dynamicEnable": true,
-    "dynamicDisable": true
+    "dynamicDisable": true,
+    "versionAdded": "1.32.0"
 }
--- a/addons/scene-color/addon.json
+++ b/addons/scene-color/addon.json
@@ -20,24 +20,6 @@
             "matches": ["projects"]
         }
     ],
-    "settings": [
-        {
-            "id": "default",
-            "name": "default color",
-            "type": "select",
-            "default": "black",
-            "potentialValues": [
-                {
-                    "id": "black",
-                    "name": "Black"
-                },
-                {
-                    "id": "white",
-                    "name": "White"
-                }
-            ]
-        }
-    ],
     "customCssVariables": [
         {
             "name": "color",
--- a/addons/scene-color/icon.svg
+++ b/addons/scene-color/icon.svg
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- https://material.io/resources/icons/?search=brightness_3&icon=brightness_3&style=baseline -->
+<svg width="24px" height="24px" fill="white" version="1.1" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+ <path d="m9.5919 3.6911c-0.87243 0-1.7033 0.13294-2.4927 0.38221 3.3734 1.0552 5.8162 4.2043 5.8162 7.9267 0 3.7224-2.4428 6.8714-5.8162 7.9267 0.78934 0.24927 1.6202 0.38221 2.4927 0.38221 4.5865 0 8.3089-3.7224 8.3089-8.3089 0-4.5865-3.7224-8.3089-8.3089-8.3089z" fill="#575e75" stroke-width=".83089"/>
+</svg>
--- a/addons/scene-color/userscript.js
+++ b/addons/scene-color/userscript.js
@@ -1,0 +1,58 @@
+export default async function ({ addon, console }) {
+
+    while (true) {
+        await addon.tab.waitForElement('[class*="paint-editor_zoom-controls"]', {
+        markAsSeen: true,
+    });
+
+    var color = "#111111";
+    var color2 = "#1a1a1a";
+    var color3 = "#2a2a2a";
+    var color4 = "#333333";
+    const paper = await addon.tab.traps.getPaper();
+    Update()
+
+    const div = document.createElement("div")
+    div.classList.add("button-group_button-group_2_h4y")
+
+    div.addEventListener("click", function() {
+        if (color == "#111111") {
+            color = "white";
+            color2 = "white";
+            color3 = "white";
+            color4 = "white";
+        } else {
+            color = "#111111";
+            color2 = "#1a1a1a";
+            color3 = "#2a2a2a";
+            color4 = "#333333";
+        }
+        Update()
+    })
+
+    const span = document.createElement("span")
+    span.classList.add("button_button_u6SE2", "paint-editor_button-group-button_1I1tm")
+    span.role = "button"
+
+    const img = document.createElement("img")
+    img.classList.add("paint-editor_button-group-button-icon_10kVn")
+    img.src = addon.self.dir + "/icon.svg"
+    img.draggable = "false"
+
+    div.appendChild(span)
+    span.appendChild(img)
+    document.querySelector('[class*="paint-editor_zoom-controls"]').appendChild(div)
+
+    function Update() {
+    for (let layer of paper.project.layers) {
+        if (layer.data.isBackgroundGuideLayer) {
+          layer.vectorBackground._children[0].fillColor = color3;
+          layer.vectorBackground._children[1].fillColor = color4;
+          layer.vectorBackground._children[1]._children[0].fillColor = color;
+          layer.vectorBackground._children[1]._children[1].fillColor = color2;
+          layer.bitmapBackground._children[0].fillColor = color;
+          layer.bitmapBackground._children[1].fillColor = color2;
+        }
+    }
+}
+}}


DNin01:
If I'm correct, this is already covered by:
https://github.com/ScratchAddons/ScratchAddons/blob/5f39d79a88330bf72be8cf9070d9df596d92983b/addons/editor-dark-mode/addon.json#L970
But it might be possible to make these two features compatible with each other.

DNin01 (changes requested):
If you intend for this to be a new addon, don't forget to add the new addon's ID to the `addons.json` file.

Or, as I was suggesting, we could also move this feature to `editor-dark-mode` so they work together.

callumjt:
> If you intend for this to be a new addon, don't forget to add the new addon's ID to the `addons.json` file.
> 
> Or, as I was suggesting, we could also move this feature to `editor-dark-mode` so they work together.

oops i forgot to upload addons.json

callumjt:
ummm why did it request a review from wl?

DNin01:
> ummm why did it request a review from wl?

They're a ["code owner"](https://github.com/ScratchAddons/ScratchAddons/blob/5f39d79a88330bf72be8cf9070d9df596d92983b/.github/CODEOWNERS#L2). It's a GitHub feature that allows reviewers to automatically be requested for certain changes.

callumjt:
> 



> > ummm why did it request a review from wl?
> 
> They're a ["code owner"](https://github.com/ScratchAddons/ScratchAddons/blob/5f39d79a88330bf72be8cf9070d9df596d92983b/.github/CODEOWNERS#L2). It's a GitHub feature that allows reviewers to automatically be requested for certain changes.

ah ok

BroJac5246:
> Or, as I was suggesting, we could also move this feature to `editor-dark-mode` so they work together.

Personally, I think this would be more useful as a separate addon. This is something I would use, but I would not think to look in the editor dark mode addon.

WorldLanguages:
Could we first discuss in a newly created issue about why this feature would be useful? @callumjt Have you seen anyone suggesting this? I can only remember requests of adding a way to dynamically disable the canvas dark mode.

callumjt:
> 



> Could we first discuss in a newly created issue about why this feature would be useful? @callumjt Have you seen anyone suggesting this? I can only remember requests of adding a way to dynamically disable the canvas dark mode.

someone in the discord requested this and i thought it was a good idea so i made it.

WorldLanguages:
Actually, I think I got confused and I was referring here to what you just made:

> I can only remember requests of adding a way to dynamically disable the canvas dark mode.

WorldLanguages:
I think we could possibly remove "canvas dark mode" from editor-dark-mode into a separate addon, what do you think @mxmou?

mxmou:
> I think we could possibly remove "canvas dark mode" from editor-dark-mode into a separate addon, what do you think @mxmou?

The toggle inside the costume editor should probably be a separate addon, but I think the editor dark mode setting would still be useful as a default value.

DNin01:
I just realized, if it were a separate addon, then users could use it even if they prefer light themes for the rest of the editor.

Secret-chest:
You should add settings to customise the colours, and generate the checkerboard colours dynamically

WorldLanguages:
> I just realized, if it were a separate addon, then users could use it even if they prefer light themes for the rest of the editor.

You can already achieve this through the "vanilla Scratch colors" preset.

mxmou:
> > I just realized, if it were a separate addon, then users could use it even if they prefer light themes for the rest of the editor.
> 
> You can already achieve this through the "vanilla Scratch colors" preset.

You can't currently have a dark canvas background in a light theme.

callumjt:
hmm

callumjt:
> hmm

shoot wrong thing

callumjt:
so do you think that this would be accepted? :)

callumjt:
:(

callumjt:
Ye I don't think this is gonna be accepted :(

BroJac5246:
> Ye I don't think this is gonna be accepted :(

It hasn't been closed, so I wouldn't be surprised if it will be accepted.

@WorldLanguages any thoughts?

DNin01:
> Ye I don't think this is gonna be accepted :(

It takes a while for some things to get reviewed - we always have quite a lot to do.

callumjt:
> > Ye I don't think this is gonna be accepted :(
> 
> It takes a while for some things to get reviewed - we always have quite a lot to do.

sorry i am quite impatient

DNin01 (changes requested):
Unfortunately, there are a few things that need to be addressed:
- Dynamic enable/disable don't work. It takes more than specifying `dynamicEnable`/`Disable` in the manifest - sometimes you need to modify code too. Check [this documentation page](https://scratchaddons.com/docs/reference/addon-api/addon.self/#using-addonselfdisabled-and-events-dynamicdisable) for more info.
- This addon could take `Editor dark mode and customizable colors (editor-dark-mode)` into account. For instance, `scene-color` could automatically set the canvas color to light when `editor-dark-mode`'s "Change the costume editor background and selection borders" setting is disabled, and dark when it is enabled.
- The checkerboard pattern should be visible and it should look the same as in `Editor dark mode and customizable colors` and vanilla Scratch.

callumjt:
> Unfortunately, there are a few things that need to be addressed:
> 
> - Dynamic enable/disable don't work. It takes more than specifying `dynamicEnable`/`Disable` in the manifest - sometimes you need to modify code too. Check [this documentation page](https://scratchaddons.com/docs/reference/addon-api/addon.self/#using-addonselfdisabled-and-events-dynamicdisable) for more info.
> 
> - This addon could take `Editor dark mode and customizable colors (editor-dark-mode)` into account. For instance, `scene-color` could automatically set the canvas color to light when `editor-dark-mode`'s "Change the costume editor background and selection borders" setting is disabled, and dark when it is enabled.
> 
> - The checkerboard pattern should be visible and it should look the same as in `Editor dark mode and customizable colors` and vanilla Scratch.

I should be able to do all of those tomorrow but I'm not sure what you mean on 3? Do you
Mean on light mode?

DNin01:
> > Unfortunately, there are a few things that need to be addressed:
> > 
> > * Dynamic enable/disable don't work. It takes more than specifying `dynamicEnable`/`Disable` in the manifest - sometimes you need to modify code too. Check [this documentation page](https://scratchaddons.com/docs/reference/addon-api/addon.self/#using-addonselfdisabled-and-events-dynamicdisable) for more info.
> > * This addon could take `Editor dark mode and customizable colors (editor-dark-mode)` into account. For instance, `scene-color` could automatically set the canvas color to light when `editor-dark-mode`'s "Change the costume editor background and selection borders" setting is disabled, and dark when it is enabled.
> > * The checkerboard pattern should be visible and it should look the same as in `Editor dark mode and customizable colors` and vanilla Scratch.
> 
> I should be able to do all of those tomorrow but I'm not sure what you mean on 3? Do you Mean on light mode?

Here's a comparison:

### Editor dark mode and customizable colors
Light
Dark

### Scene color
Light
Dark
(Also, this picture doesn't show it, but in dark mode, the opaque area on the outside does not change, but `editor-dark-mode` does change it.)

Ideally, these should match. Also note that the background is customizable on `editor-dark-mode`.

WorldLanguages:
> @WorldLanguages any thoughts?

I'm a bit confused over how exactly this new addon would play out with the editor-dark-mode toggle that also changes the canvas color. I think we'd need to discuss that before the PR author continues to make changes to the implementation.
Reviewing this would be a lot simpler if we had discussed this on an issue before the PR submission or if there wasn't an existing toggle that switched the canvas color.

Some more concerns:
1. We'd need to avoid code duplication
2. Does the current implementation support 4 hardcoded color values? Why where those values chosen?

callumjt:
> > @WorldLanguages any thoughts?
> 
> 
> 
> I'm a bit confused over how exactly this new addon would play out with the editor-dark-mode toggle that also changes the canvas color. I think we'd need to discuss that before the PR author continues to make changes to the implementation.
> 
> Reviewing this would be a lot simpler if we had discussed this on an issue before the PR submission or if there wasn't an existing toggle that switched the canvas color.
> 
> 
> 
> Some more concerns:
> 
> 1. We'd need to avoid code duplication
> 
> 2. Does the current implementation support 4 hardcoded color values? Why where those values chosen?

What if on editor dark mode, we see if this addon is enabled and if so we don't change the canvas but we still do the colour variables and this addon gets those colour variable and sets them?

Samq64:
I think the button should be added to `editor-dark-mode` or the canvas setting should be removed from it. Either was I think it could be a colour input instead of a toggle.

callumjt:
ok so let me get this right. in editor dark mode there will now be a switch that has the options light and dark and that will be the default value then this addon will change it? will the light and dark colours be the same for both addons?

WorldLanguages:
>  in editor dark mode there will now be a switch that has the options light and dark and that will be the default value then this addon will change it?

That's one of the best ways IMO, but let's see if someone can think of a better way

> will the light and dark colours be the same for both addons

Don't see why they woudl need to be different.

callumjt:
> > in editor dark mode there will now be a switch that has the options light and dark and that will be the default value then this addon will change it?
> 
> That's one of the best ways IMO, but let's see if someone can think of a better way
> 
> > will the light and dark colours be the same for both addons
> 
> Don't see why they would need to be different.

ok ill start working on that. should i change all the editor dark mode presets too?

mxmou:
> How does editor-dark-mode **currently** determine the background color of the canvas?

If the setting to "change the costume editor background and selection borders" is enabled, it uses the accent color setting. It also generates a few variants of that color. If the setting is disabled, it doesn't make any changes to the costume editor.

> I also wonder what would happen if this new addon is dynamically disabled. Or possibly this addon shouldn't support dynamicDisable at all, to avoid thinking about this edge cases. But it could support dynamicEnable normally.

I imagine it would revert to the vanilla colors, or to the default set in editor dark mode if that is enabled.

> And if editor-dark-mode is dynamically enabled after this new addon runs, then it shouldn't have any effect in the canvas, right? Any other edge cases we need to consider? What happens if the user changes the editor-dark-mode default from light to dark while the editor is open?

Editor dark mode wouldn't affect the state of the light/dark toggle but it might change the exact colors.

Changing the default would have no effect if `scene-color` is enabled. If it's disabled, I would expect the editor dark mode setting to be updated dynamically.

WorldLanguages:
> Editor dark mode wouldn't affect the state of the light/dark toggle but it might change the exact colors.

Hmm, I think I see what you mean, but it's a bit hard to wrap my head around the idea.
So editor-dark-mode may affect both the default when the editor is loading, as well as the exact colors that this other new addon lets the user choose. Am I getting this right?

Honestly, I don't care exactly how we solve this, as long as it's somehow coherent, user choices are migrated when we release a new version, and dynamicDisable/Enable works as expected (or are intentionally unsupported). We need to support all migration cases (for example, I don't think we should enable this new addon by default to people that had the vanilla colors preset), as well as all combinations of dynamicDisable/Enable in all possible orders (both addons can be dynamically enabled in 2 possible orders, or only one of the addons can be dynamically enabled, and if this new addon also supports dynamicDisable then multiply this by two).

Secret-chest:
I can't withstand this ugly black. Please add settings to set exact colours, like we have now.

callumjt:
Shoot forgot about this, I might be able to make it on Thursday

WorldLanguages:
@mxmou Could you write a small pseudo-specification on how exactly this new addon should interact with editor-dark-mode?

mxmou:
> @mxmou Could you write a small pseudo-specification on how exactly this new addon should interact with editor-dark-mode?

Depending on which addons are enabled and how they're configured, a light or dark color scheme may be used:
* If editor dark mode is enabled and `scene-color` is disabled, the active color scheme should be controlled by a new select setting in editor dark mode. The existing setting to change the "costume editor background and selection border" should be removed.
* If `scene-color` is enabled and editor dark mode is disabled, the color scheme is set by the user using the switch added by `scene-color`. The switch is initially set to light mode.
* If both addons are enabled, the color scheme is controlled by the switch. The initial value of the switch depends on the editor dark mode setting.

A color scheme defines the following colors:
* artboard: the 480x360 area in the middle of the canvas. Scratch's default color: `#ffffff`
* workspace: the darker outer area. Default color: `#ecf1f9`
* checkerboard pattern. Default color: `#d9e3f2`
* background: the background of the `<canvas>` element. It's only visible when zoomed very far out. Default color: `#d9e3f2`

Editor dark mode currently also changes a few other colors. Those shouldn't be affected by the new addon.

How exactly each color scheme looks depends on editor dark mode settings:
* If editor dark mode is disabled, the light color scheme is the same as Scratch's defaults. The dark color scheme could use `#575e75`, Scratch's dark gray color, as a base color.
* If editor dark mode is enabled, both color schemes should be generated based on its settings.

Enabling or disabling editor dark mode while `scene-color` is enabled shouldn't change the state of the switch. Disabling `scene-color` while editor dark mode is enabled should reset the color scheme to the default set in editor dark mode, but it should probably remember the state of the switch and use it again when the addon is reenabled.

callumjt:
> > > Unfortunately, there are a few things that need to be addressed:
> > > 
> > > * Dynamic enable/disable don't work. It takes more than specifying `dynamicEnable`/`Disable` in the manifest - sometimes you need to modify code too. Check [this documentation page](https://scratchaddons.com/docs/reference/addon-api/addon.self/#using-addonselfdisabled-and-events-dynamicdisable) for more info.
> > > * This addon could take `Editor dark mode and customizable colors (editor-dark-mode)` into account. For instance, `scene-color` could automatically set the canvas color to light when `editor-dark-mode`'s "Change the costume editor background and selection borders" setting is disabled, and dark when it is enabled.
> > > * The checkerboard pattern should be visible and it should look the same as in `Editor dark mode and customizable colors` and vanilla Scratch.
> > 
> > 
> > I should be able to do all of those tomorrow but I'm not sure what you mean on 3? Do you Mean on light mode?
> 
> Here's a comparison:
> ### Editor dark mode and customizable colors
> 
> Light Dark
> ### Scene color
> 
> Light Dark (Also, this picture doesn't show it, but in dark mode, the opaque area on the outside does not change, but `editor-dark-mode` does change it.)
> 
> Ideally, these should match. Also note that the background is customizable on `editor-dark-mode`.

how do i change the opaque area? i thought the this code would do the job but it doesnt seem to work 
image

Secret-chest:
I still don't understand, can't the colours be customisable?

callumjt:
> I still don't understand, can't the colours be customisable?

yeee umm i have no clue on how to do that

callumjt:
I have no motivation to make this so imma just close it

PR #5871, opened 2023-04-02
AutoBumper Addons List

emma144:
Hello! I've come across this extension recently and I thought it would've been cool if this was added. It is an extension that adds a button at the bottom of the page on all Scratch topics that, once clicked, will send a bump message. The bump message is, of-course, customizable in the settings. Please let me know if I have missed anything! I also tested this locally and it works.

### DIFF
--- a/addons/auto-bumper/addon.json
+++ b/addons/auto-bumper/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Auto-Bumper",
+  "name": "One-click bump",
   "description": "Bump a topic automatically just by clicking one button.",
   "tags": ["community", "forums"],
   "settings": [
--- a/addons/auto-bumper/addon.json
+++ b/addons/auto-bumper/addon.json
@@ -1,0 +1,20 @@
+{
+  "name": "Auto-Bumper",
+  "description": "Bump a topic automatically just by clicking one button.",
+  "tags": ["community", "forums"],
+  "settings": [
+    {
+      "id": "bumptext",
+      "name": "Bump Text",
+      "type": "string",
+      "default": "Bump!"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["discuss", "https://scratch.mit.edu/"]
+    }
+  ],
+  "enabledByDefault": false
+}
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -137,6 +137,7 @@
   "copy-reporter",
   "share-through-mystuff",
   "featured-dangos",
+  "auto-bumper",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/auto-bumper/userscript.js
+++ b/addons/auto-bumper/userscript.js
@@ -17,6 +17,7 @@ export default async function ({ addon, console }) {
   
   // listen to the bump button click
   button.addEventListener("click", async () => {
+    if (!prompt('Are you sure you would like to bump this topic?')) return;
     textarea.value = addon.settings.get('bumptext');
     postButton.click();
   });
--- a/addons/auto-bumper/userscript.js
+++ b/addons/auto-bumper/userscript.js
@@ -2,9 +2,15 @@ export default async function ({ addon, console }) {
   
   // create a button and append it
   const button = document.createElement('button');
-  button.textContent = 'Bump';
   button.style.float = 'right';
+  button.classList.add('grey');
+  button.classList.add('button');
   document.querySelector('.form-submit').append(button);
+  
+  // add the button content
+  const span = document.createElement('span');
+  span.textContent = 'Bump';
+  button.append(span);
   
   let textarea = document.querySelector('.markItUpEditor');
   let postButton = document.querySelector('.form-submit').firstElementChild;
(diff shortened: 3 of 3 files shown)


addons/auto-bumper/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Auto-Bumper",
```
DNin01, line 2: With this name, it sounds like topics could be bumped without user interaction. A more fitting name might be:
```suggestion
  "name": "One-click bump",
```

BroJac5246:
Suggestion: the button should only show if there hasn't been a post in a certain amount of time. This will prevent bumping when it shouldn't be done. The period could be a setting, or it could be hard-coded to a set period, such as 24 hours.

emma144:
> Suggestion: the button should only show if there hasn't been a post in a certain amount of time. This will prevent bumping when it shouldn't be done. The period could be a setting, or it could be hard-coded to a set period, such as 24 hours.

I am not sure how I could do that, and I believe it would also be a bit complicated

WorldLanguages:
@emma144 Congrats on making your own addon!

WorldLanguages:
Personally I'm doubtful on whether this belongs on Scratch Addons. We could be indirectly encouraging spam and accidental bump posts.

emma144:
Hmm, well, we could find a way to add a cooldown and a confirmation box?

BroJac5246:
> Hmm, well, we could find a way to add a cooldown and a confirmation box?

The cooldown should be the 60/120- second rule; the confirmation box would probably be advisable to help prevent missclicks.

emma144:
> > Hmm, well, we could find a way to add a cooldown and a confirmation box?
> 
> The cooldown should be the 60/120- second rule; the confirmation box would probably be advisable to help prevent missclicks.

The cooldown is currently the 120/60 second post rule already

BroJac5246:
> > > Hmm, well, we could find a way to add a cooldown and a confirmation box?
> > 
> > The cooldown should be the 60/120- second rule; the confirmation box would probably be advisable to help prevent missclicks.
> 
> The cooldown is currently the 120/60 second post rule already

I'm personally cool with it staying at that

WorldLanguages:
> Personally I'm doubtful on whether this belongs on Scratch Addons. We could be indirectly encouraging spam and accidental bump posts.

Note that by "spam" here, I was not only referring to bumping often. Simply bumping 1 forum post unnecessarily would also count as spam.
If we release this addon to the public we will be indirectly encouraging users to bump their posts because it will be a bit easier. It's also possible that some users might learn about the concept of bumping because of our addon explanation.

WorldLanguages:
I'm not a hardcore forums user, but I believe there are some forum categories where bumping makes no sense and is disallowed, right? Would we need to consider those cases?
I don't like the fact publishing this addon could possibly mean we need to educate its users about forum etiquette regarding bumping.

BroJac5246:
> > Personally I'm doubtful on whether this belongs on Scratch Addons. We could be indirectly encouraging spam and accidental bump posts.
> 
> Note that by "spam" here, I was not only referring to bumping often. Simply bumping 1 forum post unnecessarily would also count as spam.

[This topic](https://scratch.mit.edu/discuss/topic/395720/) suggests that you shouldn't bump in the following situations:
```
1. Bumping more than twice a day. This is considered spam.
2. Bumping really old topics which are no longer useful. (Like old questions in which the owner is no longer looking for answer)
3. Stickies (These are already at the top so don't need to be bumped). 
```

The first can be fixed by only showing the button on topics that haven't received a post in 24 hours; the third can be fixed by not showing the button on stickies. The second would be hard to determine, but we could also show a warning if the topic hasn't received a post in, say, 6 months.

If we also include a note in the settings menu then this should be okay.

WorldLanguages:
I don't like where this is going. We need to implement some of these rules into the code, and educate the user about others

emma144:
Actually, @WorldLanguages, it seems complicated and from a second thought is not very practical

WorldLanguages:
If anyone wants to re-discuss the addon idea, please create a "new addon" issue.

GrahamSH-LLK:
> > > Personally I'm doubtful on whether this belongs on Scratch Addons. We could be indirectly encouraging spam and accidental bump posts.
> > 
> > 
> > Note that by "spam" here, I was not only referring to bumping often. Simply bumping 1 forum post unnecessarily would also count as spam.
> 
> [This topic](https://scratch.mit.edu/discuss/topic/395720/) suggests that you shouldn't bump in the following situations:
> 
> ```
> 1. Bumping more than twice a day. This is considered spam.
> 2. Bumping really old topics which are no longer useful. (Like old questions in which the owner is no longer looking for answer)
> 3. Stickies (These are already at the top so don't need to be bumped). 
> ```
> 
> The first can be fixed by only showing the button on topics that haven't received a post in 24 hours; the third can be fixed by not showing the button on stickies. The second would be hard to determine, but we could also show a warning if the topic hasn't received a post in, say, 6 months.
> 
> If we also include a note in the settings menu then this should be okay.

Just a comment, @BroJac5246, you aren't a maintainer of this project by any sense of the term. You keep giving people a sense that _you_ approve and _you're_ the person who has to approve. The problem is that that's not true. Just a thought.

BroJac5246:
> Just a comment, @BroJac5246, you aren't a maintainer of this project by any sense of the term. You keep giving people a sense that _you_ approve and _you're_ the person who has to approve. The problem is that that's not true. Just a thought.

Oops, thank you for pointing that out.

Issue #4746, opened 2022-07-02
Warn users before posting "scratch addons" on forums


WorldLanguages:
It's happening [quite a lot](https://ocular.jeffalo.net/search?q=%22scratch%20addons%22&sort=newest), and apparently the Scratch forums don't have a list of filtered words like comments, so the ST can't easily block it like they did with comments.

We already warn the user when about to comment "scratch addons", so we should reuse the code and do the same for posting forum topics.

WorldLanguages:
We already add a notice when posting to bugs&glitches asking users to please disable SA and check if the bug happens on vanilla Scratch. Unfortunately, some users mention "this happens with Scratch Addons disabled" on their post after reading that, breaking extension policy accidentally.

mxmou:
I think this would be helpful - the ST usually sends an alert with a link to the browser extension policy but sometimes nobody reports the post.

> apparently the Scratch forums don't have a list of filtered words like comments, so the ST can't easily block it like they did with comments

They do. There isn't an easy way to add new words to the list because it's hardcoded.

PR #5872, opened 2023-04-02
Change message when user posts "Scratch Addons"

DNin01:
### Changes

Modified the message that appears when you try to post a comment containing "Scratch Addons".

### Reason for changes

To encourage internal motivation. While bans and mutes do stop users from continuing harmful behavior, punishments can also make them avoid situations _because they don't want to be punished again_, when the actual reason is, in this case, to keep kids with little to no experience with staying safe from dangerous/malicious software, safe. This new message takes the focus away from the "punishment" of being muted and moves it towards keeping other kids safe by not talking about being muted.

### Tests

Not tested, but it should work.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -344,7 +344,7 @@
     "message": "No results found in \"$1\". Perhaps try different keywords?"
   },
   "captureCommentError": {
-    "message": "Message added by Scratch Addons: it appears that your comment doesn't follow $1, which asks Scratchers not to advertise or name browser extensions. This comment wasn't posted because it will probably result in your account getting muted from commenting for 5 minutes."
+    "message": "Message added by Scratch Addons: it appears that your comment doesn't follow $1, which asks Scratchers not to advertise or name browser extensions for safety reasons. This comment wasn't posted because mentioning the name of browser extensions violates the aforementioned policy."
   },
   "captureCommentPolicy": {
     "message": "Scratch Community's Browser Extension/Userscript Policy"
@@ -353,7 +353,7 @@
     "message": "Post anyway"
   },
   "captureCommentConfirm": {
-    "message": "Are you sure you want to post this comment? Scratch will probably mute your account from commenting anywhere on Scratch for at least 5 minutes."
+    "message": "Are you sure you want to post this comment?"
   },
   "extensionStoreDescription1": {
     "message": "Choose from 100+ options:",


_locales/en/messages.json
```diff
@@ -344,16 +344,16 @@
     "message": "No results found in \"$1\". Perhaps try different keywords?"
   },
   "captureCommentError": {
-    "message": "Message added by Scratch Addons: it appears that your comment doesn't follow $1, which asks Scratchers not to advertise or name browser extensions. This comment wasn't posted because it will probably result in your account getting muted from commenting for 5 minutes."
+    "message": "Message added by Scratch Addons: it appears that your comment doesn't follow $1, which asks Scratchers not to advertise or name browser extensions for safety reasons. This comment wasn't posted because mentioning the name of browser extensions violates the aforementioned policy."
```
Samq64, line 347: I prefer the old message except without mentioning how long the user will be muted.
```suggestion
    "message": "Message added by Scratch Addons: it appears that your comment doesn't follow $1, which asks Scratchers not to advertise or name browser extensions. This comment wasn't posted because it will probably result in your account being temporarily muted from commenting."
```

WorldLanguages:
The motivation of the original message is to give more context to the user about "why does scratch addons get me muted", where Scratch is not helpful at all. So, linking to the extension policy was the most important thing to show the user.
However, would it make sense for us to wait until the user gets muted in order to tell them? Not really, so that's why we stop the comment from reaching the internet, even though the user clicked the post button.
But at this point, we need to be careful because we are preventing the user from posting a comment. No other feature on the extension does that. So, I assume that telling the user that "we blocked you from doing what you wanted, but that's better than being muted" seemed reasonable.

Not opposing this change, just trying to follow my mental justification of the original message.

WorldLanguages:
Clarification: of course the user can click "post anyway" and confirm that they're aware that they'll probably get muted.

WorldLanguages:
We plan to add this same blocking mechanism to the forums (#4746) so having a single message that explains why it's wrong to mention SA, outside from muting, might be a good idea. The forums have a different bad word mechanism that doesn't include "scratch addons" as far as I know

DNin01:
> So, I assume that telling the user that "we blocked you from doing what you wanted, but that's better than being muted" seemed reasonable.

That makes sense. While reading your comment, I was starting to wonder if knowing that we prevented the user from being muted would mean the fact that SA stopped the comment from being posted would make more sense.

WorldLanguages:
@DNin01 I think you should rephrase what you just said :P

DNin01:
Basically, maybe the user would be more likely to understand why we prevented them from posting the comment if we tell them that they would've been muted.

WorldLanguages:
> Basically, maybe a user would be more likely to understand why we prevented them from posting the comment if we told them that they would've been muted. Is that true?

It's probably true, and I'm pretty sure that's the motivation for the original message.

_locales/en/messages.json
DNin01, line 347: So, then how about...?
```suggestion
    "message": "Message added by Scratch Addons: it appears that your comment doesn't follow $1, which asks Scratchers not to advertise or name browser extensions for safety reasons. This comment wasn't posted because it would likely result in your account being temporarily muted."
```

Samq64: Could this be interpreted as our extension being unsafe?

WorldLanguages: @Samq64 Possibly it could.

DNin01: The other parts of that message might be interpreted like that too.

WorldLanguages:
My suggestion below. Tries to be a compromise between both of suggestions above.
It is also slightly inspired by actual sentences that Scratch uses when swearing or posting links to "unmoderated chat".

Message when posting a comment:
> Message added by Scratch Addons: your comment mentions a browser extension. The {Browser Extension Policy} urges Scratchers not to advertise them, as they have not been checked for safety by the Scratch Team. Please follow the aforementioned policy by avoiding any references to browser extensions when posting on Scratch. {Post anyway}

Confirmation message after clicking "post anyway":
> Posting this comment may lead to automatic muting of your account, temporarily preventing you from commenting. Persistent muting can result in your account being blocked from Scratch. Are you sure you want to post this comment?

WorldLanguages:
> This new message tries to take the focus away from the "punishment" of being muted and moves it towards keeping other kids safe by not talking about being muted.

I believe my suggestion does take the focus away from the punishment.
Compared to the old version, it also clarifies that *any reference* to a browser extension is against the policy.

DNin01:
Looks good. I'd suggest a slight modification to one of the sentences:
```diff
- The {Browser Extension Policy} urges Scratchers not to advertise them, as they have not been checked for safety by the Scratch Team.
+ The {Browser Extension Policy} urges Scratchers not to advertise them, as the Scratch Team have not checked them for safety and can't possibly check every single one.
```

WorldLanguages:
Screenshots:

image
image

WorldLanguages:
The captureCommentError string can be reused for the forums (#4746)
The confirmation string will need to be different because there's no such thing as muted from the forums.

Samq64 (approved):

Issue #3930, opened 2021-12-15
Make `if` work for notices and warings


Samq64:
### Is your request related to a problem? Please describe.

The are some notices and warnings that are useless unless a specific setting enabled and it clutters up the settings page.

### Describe the solution you'd like

Make the `if` used for addon settings also work for notices and warnings, so they can be hidden when they are not needed.

### Describe alternatives you've considered

I haven't thought or any alternatives, but I guess a few notices could be a removed entirely.

### Additional context

I tried using `if` on the `fullscreen` notice, but it had no effect.

Related: #3015

Samq64:
While working on this I noticed there isn't an easy way to prevent the layout from shifting. Is that problematic?

Also while there are a few places where this makes sense, most setting-specific notices would be better off as setting descriptions (#5320).

mxmou:
It would also be useful if addon names and descriptions could be different depending on the platform (to change Ctrl to Cmd on macOS, for example).

PR #5878, opened 2023-04-04
Add `makeymakey` to `load-extensions`

### LINKED ISSUE #5876: Add Makey Makey to automatically add extensions
### Why this would be helpful

I use makey makey alot in my games and tbh its not super important I feel it would be nice

### My suggestion

Very simple add a slider for makey makey

### Possible alternatives

_No response_

### Additional context

_No response_

Samq64:
Resolves #5876

### Changes

Adds the Makey Makey extension to the automatically add extensions addon.

### Reason for changes

Someone want the extension added and it's simple to add.

### Tests

Tested on Firefox.

### DIFF
--- a/addons/load-extensions/addon.json
+++ b/addons/load-extensions/addon.json
@@ -31,10 +31,20 @@
       "id": "translate",
       "type": "boolean",
       "default": false
+    },
+    {
+      "name": "Makey Makey",
+      "id": "makeymakey",
+      "type": "boolean",
+      "default": false
     }
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
   "versionAdded": "1.10.0",
+  "latestUpdate": {
+    "version": "1.32.0",
+    "newSettings": ["makeymakey"]
+  },
   "tags": ["editor", "codeEditor", "featured"]
 }
--- a/addons/load-extensions/userscript.js
+++ b/addons/load-extensions/userscript.js
@@ -3,7 +3,7 @@ export default async function ({ addon, console }) {
   const loadExtensions = () => {
     if (addon.self.disabled) return;
     // IDs are taken from https://github.com/LLK/scratch-vm/blob/ffa78b91b8645b6a8c80f698a3637bb73abf2931/src/extension-support/extension-manager.js#L11
-    const EXTENSIONS = ["music", "pen", "text2speech", "translate"];
+    const EXTENSIONS = ["music", "pen", "text2speech", "translate", "makeymakey"];
     for (let ext of EXTENSIONS) {
       // Check if setting enabled and it's not already loaded
       if (addon.settings.get(ext) && !vm.extensionManager.isExtensionLoaded(ext)) {


BroJac5246:
While we're at it, we might as well just add all of the extensions. There's not a real reason not to.

DNin01:
How often is the Makey Makey extension needed? To me, it doesn't seem like this would save much time for most users, but maybe I'm wrong.

BroJac5246:
> How often is the Makey Makey extension needed? To me, it doesn't seem like this would save much time for most users, but maybe I'm wrong.
> 
> (Okay, that's not exactly what I meant.)

It's a valid point, but we might as well. Some users benefit, and those who don't aren't harmed. All they have to do is keep the setting off.

mxmou:
> While we're at it, we might as well just add all of the extensions. There's not a real reason not to.

There used to be an option for video sensing, but it was removed in #1799 because adding the video sensing extension automatically enables the camera.

There probably isn't a reason to not add hardware extensions though.

WorldLanguages:
> There probably isn't a reason to not add hardware extensions though.

As far as I know, we didn't intentionally exclude them. But auto-loading hardware extensions makes less sense than typical extensions like pen.

Samq64:
I'll add the hardware extensions then. Should I also add video sensing?

BroJac5246:
> I'll add the hardware extensions then. Should I also add video sensing?

If you can find a good solution to the problem mentioned by WorldLanguages then yes, otherwise just pass on it.

Samq64:
I added all of the extensions and added a warning about the webcam turning on when loading the video extension.

addons/load-extensions/addon.json
```diff
@@ -6,6 +6,11 @@
       "type": "warning",
       "text": "When the video sensing option is enabled, the webcam will turn on every time the editor is opened.",
       "id": "webcam"
+    },
+    {
+      "type": "notice",
+      "text": "If an extension monitor, for example, the tempo is shown on the stage, that extension will stay loaded in that project even after hiding all monitors and disabling the addon.",
```
WorldLanguages, line 12: Very sad

DNin01: I didn't know that.

WorldLanguages: The combination of these two warnings is very scary.

Samq64: > The combination of these two warnings is very scary.

That's what I was worried about.

Samq64: #3930 would help here.

addons/load-extensions/addon.json
```diff
@@ -1,6 +1,18 @@
 {
   "name": "Automatically add extensions",
-  "description": "Auto-shows music, pen, and other extensions in the block category menu in the editor.",
+  "description": "Automatically adds the selected block extensions such as music or pen to the block category menu in the editor.",
+  "info": [
+    {
+      "type": "warning",
```
BroJac5246, line 6: Does this need to be a warning? What effects could it have?

Samq64: I could probably reduce it to a notice. I set it to a warning because users probably won't want the camera automatically tuning on so often.

BroJac5246: > I could probably reduce it to a notice. I set it to a warning because users probably won't want the camera automatically tuning on so often.

Changing it would help with World_Languages' point, and with your reasoning it would be a better notice.

Samq64:
I didn't realise the video extensions also turn on the webcam by default on the project page too, so they can't reasonably be included unless there's a way to override that behavour.

DNin01:
> I didn't realise the video extensions also turn on the webcam by default on the project page too, so they can't reasonably be included unless there's a way to override that behavour.

That can probably be overriden—some projects use these blocks but have the camera default to turned off. [Case in point](https://scratch.mit.edu/projects/370757928/)

Samq64:
Yes, the video state is normally remembered but the video sensing extension is [hardcoded](https://github.com/scratchfoundation/scratch-editor/blob/4d8530e21d54ac0ef40be7ccb011367c5ca1421c/packages/scratch-vm/src/extensions/scratch3_video_sensing/index.js#L410-L414) to turn it on the first time it loads.

mxmou:
> Yes, the video state is normally remembered but the video sensing extension is [hardcoded](https://github.com/scratchfoundation/scratch-editor/blob/4d8530e21d54ac0ef40be7ccb011367c5ca1421c/packages/scratch-vm/src/extensions/scratch3_video_sensing/index.js#L410-L414) to turn it on the first time it loads.

[The video state is `ON` by default, even if the extension isn't loaded.](https://github.com/scratchfoundation/scratch-editor/blob/4d8530e21d54ac0ef40be7ccb011367c5ca1421c/packages/scratch-vm/src/sprites/rendered-target.js#L142-L155) In my Scratch mod, which loads it by default, I [changed the project loading code](https://github.com/mxmou/scratch-vm/commit/afb86caf28a0cda044ada37e912f1263c6499b3f) to set video to `OFF` when loading a project that doesn't contain video sensing blocks.

DNin01:
Video sensing is indeed initialized to an enabled state in new projects. Theoretically we could just override that—but only if the addon's setting is enabled—to prevent video sensing from automatically starting.

The result of this would be that the projects saved by any users of this addon setting would have video sensing disabled within the project data. (Is that ideal? I mean, anyone who tries to add the Video Sensing extension to one of these projects would have to turn on the camera manually, which is not how it normally works.)

PR #5889, opened 2023-04-06
Update PR Template

Samq64:
Resolves #4956

### Changes

Adds an additional context section, Chrome and Firefox check boxes and a comment telling the users they should make an issue first to the pull request template.

### Reason for changes

To expand the pull request template without cluttering it too much.

### Tests

- [ ] Chrome
- [ ] Firefox

~~Tested by pasting the new template as this pull request's description.~~

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,4 @@
-<!-- Which issue(s) does this pull request fix or resolve? If there aren't any, please create one first unless this is a bug fix or minor string update -->
+<!-- Which issue(s) does this pull request fix or resolve? If there aren't any, please submit one first unless this is a minor string update or bug fix. -->
 
 Resolves #
 
@@ -16,3 +16,7 @@ Resolves #
 - [ ] Firefox
 
 <!-- Please add any known issues or other test notes here. -->
+
+### Additional context
+
+<!-- Any other information like screenshots, videos, or context about this pull request -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,10 +1,10 @@
-<!-- Which issue(s) does this pull request fix or resolve? If there aren't any, please submit one first unless this is a minor string update or bug fix. -->
+<!-- Which issue(s) does this pull request fix or resolve? If there aren't any, please submit one first unless this is a hotfix or minor string update. -->
 
 Resolves #
 
 ### Changes
 
-<!-- Please describe the changes you've made. Add any screenshots or videos here. -->
+<!-- Please describe the changes you've made. Add any screenshots or videos here if applicable. -->
 
 ### Reason for changes
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,7 +12,7 @@ Resolves #
 
 ### Tests
 
-- [ ] Chrome
+- [ ] Chrome/Chromium-based browser
 - [ ] Firefox
 
 <!-- Please add any known issues or other test notes here. -->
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -12,4 +12,12 @@ Resolves #
 
 ### Tests
 
-<!-- If applicable. Have you tested this pull request? If so, how? -->
+- [ ] Chrome
```
DNin01, line 15: Might want to expand this to say:
```suggestion
- [ ] Chrome/Chromium-based browser
```

.github/pull_request_template.md
```diff
@@ -12,4 +12,12 @@ Resolves #
 
 ### Tests
```
DNin01, line 13: It might be helpful to note that it is only required to test one browser, since reviewers can help. Unless that rule has changed.

DNin01 (commented):
Reviewed the whole thing. Just a few things to say, but otherwise it looks OK! I’m sure this will help contributors.

.github/pull_request_template.md
Samq64: Should Edge also be an option since we also publish to their store, or do you think Chromium based browser covers that?

DNin01 (approved):

.github/pull_request_template.md
BroJac5246: > Should Edge also be an option since we also publish to their store, or do you think Chromium based browser covers that?

Stick with Chromium.

WorldLanguages:
I'm not convinced about most changes.

- The clarification on the first line is very good.
- Adding checkboxes to all PRs will add unnecessary clutter, even to the PR list on the repo (not just to every individual PR) when a big majority of pull requests work without a problem in other browsers. I think we should face browser compatibility problems and cross-browser testing differently, not through the PR template.
- The "additional context" section seems overkill. Instead, I'd instead suggest adding a comment on the "changes" section that encourages the PR author to include screenshots/videos if applicable.

Samq64:
> Adding checkboxes to all PRs will add unnecessary clutter

At least I'm not PRing LLK's test coverage template.

image

.github/pull_request_template.md
```diff
@@ -1,15 +1,15 @@
...
+<!-- Please describe the changes you've made. Add any screenshots or videos here if applicable. -->
 
 ### Reason for changes
 
 <!-- Why should these changes be made? -->
 
 ### Tests
 
-<!-- If applicable. Have you tested this pull request? If so, how? -->
+<!-- Please test your chnages in a Chromium based browser and add any known issues or other testing notes here. Bigger changes should also be tested on Firefox. -->
```
WorldLanguages, line 15:
```suggestion
<!-- Please test your changes in a Chromium-based browser and add any known issues or other testing notes here. Bigger changes should also be tested on Firefox. -->
```

.github/pull_request_template.md
```diff
@@ -1,15 +1,15 @@
...
+<!-- Please describe the changes you've made. Add any screenshots or videos here if applicable. -->
 
 ### Reason for changes
 
 <!-- Why should these changes be made? -->
 
 ### Tests
 
-<!-- If applicable. Have you tested this pull request? If so, how? -->
+<!-- Please test your changes in a Chromium-based browser and add any known issues or other testing notes here. Bigger changes should also be tested on Firefox. -->
```
WorldLanguages, line 15: Maybe it would be better if we worded this to include PR authors that use Firefox as their main browser. The requirement is to test it in 1 browser, not to test it in a Chromium-based one 🤔

DNin01 (approved):
Still looks good to me!

I'm fine if we save browser test checkboxes for later - we haven't really discussed it or come to a conclusion about it.

WorldLanguages (approved):

PR #5890, opened 2023-04-06
Fixed direction cursor bug in sprite properties

### LINKED ISSUE #5898: sprite-properties: "collapse when mouse leaves sprite panel" not working properly
### Description

> the panel closes when you go on pop-up elements

### Steps to reproduce

See video.
Make sure setting `AUTOMATICALLY COLLAPSE WHEN MOUSE LEAVES SPRITE PANEL` is enabled.

https://user-images.githubusercontent.com/17484114/230694271-784ad36a-6d3a-48a2-a409-8fce24a056ae.mp4

None

### Environment

Bug has been around since this addon was released AFAIK

valmontechno:
Resolves #5898

when the addon the "automatically collapse when mouse leaves sprite panel" option of the "Collapsing sprite properties" addon is activated, the properties of the sprites disappear if you hover over the direction cursor.


addons/sprite-properties/userscript.js
```diff
@@ -36,11 +36,22 @@ export default async function ({ addon, console, msg }) {
     }
   }
   // Close properties panel when mouse leaves the entire sprite panel
+  let popoverAbove;
   document.body.addEventListener(
-    "mouseleave",
+    "mousemove",
```
Samq64, line 41: The event will fire every constantly even when on the project page. Is there a better solution?

addons/sprite-properties/userscript.js
```diff
@@ -36,11 +36,22 @@ export default async function ({ addon, console, msg }) {
...
     (e) => {
-      if (e.target.matches('[class*="sprite-selector_sprite-selector_2KgCX"]')) {
+      let popoverAbove = document.querySelector('[class*="Popover-above"]');
```
GarboMuffin, line 43: `document.querySelector` is slow on large projects. perhaps you can use eg. `e.target.closest(..)` instead

addons/sprite-properties/userscript.js
```diff
@@ -36,11 +36,22 @@ export default async function ({ addon, console, msg }) {
     }
   }
   // Close properties panel when mouse leaves the entire sprite panel
+  let popoverAbove;
   document.body.addEventListener(
-    "mouseleave",
+    "mousemove",
     (e) => {
-      if (e.target.matches('[class*="sprite-selector_sprite-selector_2KgCX"]')) {
```
WorldLanguages, line 42: Isn't a possible fix just adding `&& !directionPopoverCurrentlyOpen()` here?
The sprite properties wouldn't collapse until the user moves the cursor once again to the properties, and then leaves it again. But it wouldn't affect performance.

WorldLanguages: A possible implementation of `directionPopoverCurrentlyOpen` could be checking if `e.target.closest('[class*="Popover-above"]')` exists.

DNin01:
I'd suggest you try [branching](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-and-deleting-branches-within-your-repository) so you don't have to close old pull requests to make new ones.

WorldLanguages:
I opened #6090 as an alternative fix to this bug.

addons/sprite-properties/userscript.js
WorldLanguages: See #6090

Issue #5309, opened 2022-11-02
Block colors update support checklist


DNin01:
This issue is a to-do list for updating the Customizable block colors addon to take Scratch's upcoming block color switcher into account when it releases.

Discussion: #5192

- [x] Tweak the `editor-theme3` addon to make blocks look closer to the new colors (#5257)
  - Adjust the borders, text, etc.
- [x] Add a preset to use the new colors now (#5308)
- [x] Update the Debugger addon blocks' default colors to be consistent (#6152)
- [x] Add a preset for high contrast to `editor-theme3`'s [default setting values](https://github.com/ScratchAddons/ScratchAddons/blob/48106c8599fca407ded70aeb073085cd7b6b655a/addons/editor-theme3/addon.json#L196)

WorldLanguages:
Should we change the colors of debugger's blocks depending on whether the user is on default colors or accessible colors?

DNin01:
> Should we change the colors of debugger's blocks depending on whether the user is on default colors or accessible colors?

I think we definitely should.

WorldLanguages:
> Should we change the colors of debugger's blocks depending on whether the user is on default colors or accessible colors?

I think this is the only pending TODO in this checklist.

WorldLanguages:
> I think this is the only pending TODO in this checklist.

@DNin01 Can you confirm? If that's the only pending TODO I suggest creating a new issue for it with a priority-4 label.

DNin01:
> > I think this is the only pending TODO in this checklist.
> 
> 
> 
> @DNin01 Can you confirm? If that's the only pending TODO I suggest creating a new issue for it with a priority-4 label.

Created issue: #6123

WorldLanguages:
@DNin01 Can this be closed?

DNin01:
> @DNin01 Can this be closed?

Yeah, we're pretty much done here.

PR #5896, opened 2023-04-07
Use high-contrast colors in Black preset

DNin01:
A step towards #5309

### Changes

Replaced the block color values in the Black preset with the ones from the high contrast preset. The borders and text will look brighter.

### Reason for changes

To make it easier to see block borders and text.

### Tests

Tested on Edge 111.


WorldLanguages:
I don't think I currently remember why these need to be changed, but here's my question, do we need to migrate previous users of the preset into the new preset values? Or should this only affect new instances of users choosing the preset after this change is released?

DNin01:
> I don't think I currently remember why these need to be changed, but here's my question, do we need to migrate previous users of the preset into the new preset values? Or should this only affect new instances of users choosing the preset after this change is released?

We could do either. And I'm mainly changing these because I thought they would be hard to see and I thought Scratch's new block colors would be a good color palette to make them more accessible.

Secret-chest:
Shouldn't this be a separate preset?

DNin01:
> Shouldn't this be a separate preset?

It's not different enough to deserve a separate preset. Think of this as an upgrade to the existing one.

If you want the default colors with this preset, you can select "Scratch's default colors" and then choose "Colored on black background" as the text color.

mxmou (approved):
> > I don't think I currently remember why these need to be changed, but here's my question, do we need to migrate previous users of the preset into the new preset values? Or should this only affect new instances of users choosing the preset after this change is released?
>
> We could do either.

I think it's better to update the settings automatically for existing users of the preset.

WorldLanguages:
Adding "pending" label as this lacks migration code and it would be inconvenient to do that in a different PR.

WorldLanguages:
This is another PR which wasn't merged because it lacks migration. Working on it.

WorldLanguages:
Migration code I just committed is untested. I will test it

Please confirm if the migration logic is correct. If all settings (except `comment-color`, `input-color`, `text` and `forums`) exactly match the old Black preset, then the new whole Black preset is applied.
As I'm writing this, I notice this means that we might be changing the user's `comment-color`, `input-color` and/or `text` settings in the process. Should we check if those exactly match the old Black preset? Or should we ignore them, but leave them intact after the migration? Or, third alternative, do you think the current implementation should stay?

mxmou:
> As I'm writing this, I notice this means that we might be changing the user's `comment-color`, `input-color` and/or `text` settings in the process. Should we check if those exactly match the old Black preset?

Yes, because that's the only way to differentiate the Black preset from the default colors.

WorldLanguages:
@mxmou Good catch. I'll update the migration logic.

WorldLanguages:
Now, the only setting which is ignored is `CHANGE THE COLORS OF BLOCKS IN THE FORUMS`. It does not determine in any way if the migration happens, and its value always stays unchanged after the migration.

background/get-addon-settings.js
```diff
@@ -194,6 +194,28 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
             },
             manifest.presets.find((p) => p.id === "dark")
           );
+          updatePresetIfMatching(
+            settings,
+            3,
+            {
+              "motion-color": "#4C97FF",
```
WorldLanguages, line 201: After some testing, migration might not go as expected if, for example, this setting is set to `#4c97ffff` instead of the default `#4C97FF`. These 2 represent the same color.

WorldLanguages: Simply "looking" (clicking the setting without changing any sliders or clicking anything) into a color setting that allows transparency changes its value to lowercase and adds `ff` at the end to represent 100% alpha.

If we don't change this behavior, only users that applied the preset and never explored any color setting will be migrated.

WorldLanguages: It's likely that we've missed this edge case in previous migrations, but it's too late to do anything about those now.

WorldLanguages:
I will add migration code as part of another PR.

background/get-addon-settings.js
WorldLanguages: See #5992

PR #5901, opened 2023-04-08
editor-theme3: add black versions of icons

mxmou:
### Changes

Changes to the black text color option in `editor-theme3`:
* Lighter background color for hovered dropdown items
image
* Block icons from LLK/scratch-gui#8843
image

### Reason for changes

This will make the high contrast preset more similar to Scratch's high contrast theme.

### Tests

Tested on Edge and Firefox.


WorldLanguages:
Everything looks good in Chromium. Wait for @DNin01's review

DNin01 (commented):
These new black icons would probably also be good to use when the text color is set to "colored on white background".

DNin01 (approved):

PR #5546, opened 2023-01-05
Rounded thumbnails addon

DNin01:
Resolves #5328

### Changes

Added a new addon that makes thumbnails and profile pictures in various areas around the website have rounded corners.

### Reason for changes

Just to allow that customization, you know.

### Tests

Tested on Edge together with a few other addons.

PR #5903, opened 2023-04-08
Rounded corners UI addon

Valer100:
### Changes
I added a new addon which makes the Scratch website look much better by rounding the corners of some objects.

### Reason for changes
I think the Scratch website would look better, thanks to this addon

### Tests
I tested the addon on Google Chrome version 111.0.5563.148

### DIFF
--- a/addons-l10n/ro/rounded-ui.json
+++ b/addons-l10n/ro/rounded-ui.json
@@ -1,0 +1,1 @@
+{"rounded-ui/@name":"Interfață cu colțuri rotunjite","rounded-ui/@description":"Face interfața site-ului Scratch mult mai plăcută, rotunjind colțurile unor obiecte."}
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -138,6 +138,7 @@
   "share-through-mystuff",
   "auto-follow-topics",
   "featured-dangos",
+  "rounded-ui",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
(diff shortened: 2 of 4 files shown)


addons-l10n/ro/rounded-ui.json
```diff
@@ -0,0 +1 @@
+{"rounded-ui/@name":"Interfață cu colțuri rotunjite","rounded-ui/@description":"Face interfața site-ului Scratch mult mai plăcută, rotunjind colțurile unor obiecte."}
```
CST1229, line 1: You shouldn't add localized strings by yourself; those should be done through Transifex, after this PR is merged.

Valer100: Now I deleted the Romanian translation of my addon.

Secret-chest:
What about a setting to choose how much rounding to use?

Samq64:
Related: #5546

WorldLanguages:
@Valer100 We should discuss the addon idea first, before discussing the implementation. Could you create an issue on this repository with the "new addon" template where you summarize what criteria is used to choose which elements get rounded corners? And you could also mention what you think should happen to elements that already have rounded corners.

Also, I think some of your CSS selectors are targeting elements created by other browser extensions, and no addon does that already, so I'm not sure if we'll accept that since it would require more maintenance work for us in the future if other browser extensions change their CSS.

WorldLanguages:
Related PR: #5546

PR #5907, opened 2023-04-09
Update `number-pad` description

### LINKED ISSUE #5906: "Always show number pad" description is confusing
### Description

If you look at the description for the "always show number pad" addon. It says 

> A number pad will show when editing number inputs from certain blocks, such as "set x to".

But, since settings a variable in scratch can also be a string, the number pad doesn't actually show on this block. A more a appropriate example could be

> A number pad will show when editing number inputs from certain blocks such as "change x by ()"

Just a little thing that I noticed while browsing scratch addons.

### Steps to reproduce

1. Open scratch addons settings
2. Find the "always show number pad" addon
3. Look the description

_No response_

### Environment

Browser: Firefox
Scratch Addons version: 1.31.2
Operating system: Windows 10

DNin01:
Resolves #5906

### Changes

Updated the description and info message in the `number-pad` addon.

### Reason for changes

To make it less confusing.

### Tests

Tested.

### DIFF
--- a/addons/number-pad/addon.json
+++ b/addons/number-pad/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Always show number pad",
-  "description": "Show Scratch's number pad input when editing number fields on all devices, instead of only touchscreen devices.",
+  "description": "Show Scratch's virtual number pad when editing number fields in blocks on all devices, instead of only touchscreen devices.",
   "credits": [
     {
       "name": "GarboMuffin"
@@ -8,7 +8,7 @@
   ],
   "info": [
     {
-      "text": "A number pad will show when editing number inputs from certain blocks, such as \"set x to\".",
+      "text": "A number pad will show when editing number inputs from blocks such as \"move () steps\".",
       "id": "explanation"
     }
   ],


addons/number-pad/addon.json
```diff
@@ -1,14 +1,14 @@
...
+  "description": "Show Scratch's virtual number pad when editing number fields in blocks on all devices, instead of only touchscreen devices.",
   "credits": [
     {
       "name": "GarboMuffin"
     }
   ],
   "info": [
     {
-      "text": "A number pad will show when editing number inputs from certain blocks, such as \"set x to\".",
+      "text": "A number pad will show when editing number inputs from blocks such as \"move () steps\".",
```
Samq64, line 11: Is this note necessary?

BroJac5246: > Is this note necessary?

See #5906.

WorldLanguages (approved):

addons/number-pad/addon.json
DNin01: What do you want us to do with it? Do you think it should be part of the description? Reworded? Removed entirely?

Samq64 (approved):

Issue #4919, opened 2022-07-28
Workspace context menu submenus


Samq64:
### Is your request related to a problem? Please describe.

The workspace context menu can get very long, especially with devttols and custom-block-shape enabled.

### Describe the solution you'd like

An addon that adds submenus to to the context menu, for example, `Swap with... > my variable` or `Screenshot > Save script as PNG`

### Additional context

Related: #4914

PR #5917, opened 2023-04-11
Large block text setting

### LINKED ISSUE #2645: Bolder / larger symbols in math operator blocks

### LINKED ISSUE #4913: Customizable block text style doesn't redraw blocks
### Description

When enabling or disabling customizable block text style or its "bold text" setting, the size of the blocks doesn't change. This results in weird spacing until the page is reloaded.

### Reproduction steps

1. Disable customizable block text style
2. Open the editor. Spacing is correct:
  image
4. Enable customizable block text style
5. Enable the "bold text" setting. The space around the text is too small:
  image
6. Reload. Spacing is correct:
  image
8. Disable the "bold text" setting. The space around the text is too large:
  image

### Expected behavior

A reload would not be required to adjust the width of blocks.

DNin01:
### Changes

Adds a percentage setting to the `custom-block-text` addon to increase the size of text on blocks up to 200%. You can also lower the size down to 75% in case for some reason you need to.

Image

### Reason for changes

To make the blocks easier to read.

### Tests

Tested in Edge 114 in combination with `middle-click-popup`.

Resolves #4913

- [x] [TODOs](https://github.com/ScratchAddons/ScratchAddons/pull/5917#issuecomment-1616180035)


mxmou:
I don't think this should close #2645. I agree with [this comment](https://github.com/ScratchAddons/ScratchAddons/issues/2645#issuecomment-893058277):

> Some people don't have issues with the general font of the editor, only the arithmetic blocks since they have such small symbols.

BroJac5246:
> I don't think this should close #2645. I agree with [this comment](https://github.com/ScratchAddons/ScratchAddons/issues/2645#issuecomment-893058277):
> 
> > Some people don't have issues with the general font of the editor, only the arithmetic blocks since they have such small symbols.

Could there be separate settings for text and operators/symbols?

Secret-chest:
Why no settings?

WorldLanguages:
I also wonder why you've discarded the idea of making the font size customizable.

DNin01:
> Why no settings? (font size)
> 
> They're really easy to implement.

> I also wonder why you've discarded the idea of making the font size customizable.

Alright, so if I make the size customizable, should I use an integer input or a couple single-select size options like "normal", "large" and "larger"?

WorldLanguages:
@DNin01 We can combine both concepts. I'd prefer having a select with options default/large/larger/custom. If user chose "custom" they could manually enter an integer value.

This integer value doesn't necessarily need to match one-to-one to CSS units. Ideally we should allow users to change the font one step at a time as defined by browser behavior. For example, I don't think most browsers distinguish 14.2pt from 14.3pt. Maybe browsers actually round it to the nearest integer and I've written all of this for no reason? Do we need to read the CSS spec?

DNin01:
I have added the option to choose between 5 different larger text sizes.

mxmou:
I think it's better to just make the font size fully customizable (with reasonable minimum and maximum values). A checkbox that enables a number or text setting is sometimes useful, but in this case it just adds unnecessary complexity.

Secret-chest:
> I think it's better to just make the font size fully customizable (with reasonable minimum and maximum values). A checkbox that enables a number or text setting is sometimes useful, but in this case it just adds unnecessary complexity.

Completely agree. It should be a multiplier, maybe 75% to 250%.

Secret-chest:
Since it's a percentage, you can also tell how large it's going to be.

Secret-chest:
> Maybe browsers actually round it to the nearest integer and I've written all of this for no reason? Do we need to read the CSS spec?

It could depend on the scaling / zoom. And even on a normal 1920x1080px screen with 100% zoom, `13px` and `13.5px` look different.

WorldLanguages:
> I think it's better to just make the font size fully customizable (with reasonable minimum and maximum values). A checkbox that enables a number or text setting is sometimes useful, but in this case it just adds unnecessary complexity.

My concern was that the settings page currently doesn't have a per-setting reset button if the addon has no presets. Unless we use percentages, we wouldn't really expect the user to know or remember what the default text size is.
This might cause some confusion:
1. The user might think the addon defaults to smaller or larger text. How would they know if vanilla Scratch's font size is 3, 13, or whatever? We know that this addon doesn't change anything by default, but will users know?
2.  The user might want to return to the default text size (after playing with the setting) and won't remember what the vanilla Scratch value was. So their only option is to reset the whole addon to default settings.

DNin01:
> > I think it's better to just make the font size fully customizable (with reasonable minimum and maximum values). A checkbox that enables a number or text setting is sometimes useful, but in this case it just adds unnecessary complexity.
> 
> Completely agree. It should be a multiplier, maybe 75% to 250%.
> This would also remove the need for a switch, since 100% would be unchanged.

Interesting 🤔

DNin01:
> 1. The user might think the addon defaults to smaller or larger text. How would they know if vanilla Scratch's font size is 3, 13, or whatever? We know that this addon doesn't change anything by default, but will users know?

Turning off the "large text" setting would revert the text size back to the default.

Secret-chest:
Still, why have 5 levels? A percentage would be simpler to understand.

DNin01:
Yeah, let's try percentages.

DNin01:
It might look nicer if the dropdown arrows changed size too.

Secret-chest:
You can have the minimum and default be 100, then remove the toggle.

mxmou:
> You can have the minimum and default be 100, then remove the toggle.

The minimum should probably be less than 100. I don't think there's a reason to not let people make the font smaller.

WorldLanguages:
> Turning off the "large text" setting would revert the text size back to the default.

Note that if the default font size setting is 100, the "large text" toggle would become unnecessary. I only proposed a toggle so that users understand that the default font size of the addon is exactly equal to vanilla Scratch's font size, but that would become clear if the number 100 meant "same as vanilla Scratch".

DNin01:
Removed the toggle. For no change, text size can be set to 100% but it can go up to 150%.

Secret-chest:
Great. You can make the limits 50 to 200%.

DNin01:
> Great. You can make the limits 50 to 200%.

I'm still unsure why people would want to _decrease_ the text size below the already small default size, unless they typically zoom in. I could also increase the maximum but the text might not fit on the blocks.

I'll keep thinking about it. I suppose it wouldn't hurt to have the option, right?

BroJac5246:
> > Great. You can make the limits 50 to 200%.
> 
> I'm still unsure why people would want to _decrease_ the text size below the already small default size, unless they typically zoom in. I could also increase the maximum but the text might not fit on the blocks unless people use a different addon to make them even bigger.
> 
> I'll keep thinking about it. I suppose it wouldn't hurt to have the option, right?

I'm sure there are reasons (some people just find things too large), and it doesn't hurt

addons/custom-block-text/addon.json
```diff
@@ -36,11 +44,23 @@
...
+      "max": 150,
+      "default": 100
     }
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
   "updateUserstylesOnSettingsChange": true,
   "versionAdded": "1.24.0",
+  "latestUpdate": {
+    "version": "1.32.0",
```
WorldLanguages, line 62:
```suggestion
    "version": "1.33.0",
```

WorldLanguages:
Unless decreasing the font size leads to visual bugs, I think we should offer the option.

WorldLanguages:
Is it just me, or the size of the blocks is determined only on first editor load? Dynamically changing the value of the setting results in this, and I can't find a way to fix it without a reload:

image

In the screenshot I opened the editor, then changed the setting from 100 to 150 by typing in the 3 digits.

DNin01:
> Is it just me, or the size of the blocks is determined only on first editor load? Dynamically changing the value of the setting results in this, and I can't find a way to fix it without a reload:
> 
> In the screenshot I opened the editor, then changed the setting from 100 to 150 by typing in the 3 digits.

Looks like this bug: #4913

WorldLanguages:
If we don't find a way to fix it, we can tell users that a reload is required for the settings to apply.
The bug wasn't that noticeable with the bold text.

DNin01:
@mxmou Any chance you know how to fix this bug?

WorldLanguages:
> If we don't find a way to fix it, we can tell users that a reload is required for the settings to apply.

This might not be good enough - it looks like there's some code (probably from Blockly) which at some specific point of time calculates the size of the blocks based on the font size. If our CSS, for some reason, gets injected after that moment, the bug will be visible to the user even though they haven't changed the settings dynamically.

It would be nice to know which Scratch code is responsible for this. We might try to find a way to call it multiple times, or at least being able to detect if it has ran already, to avoid race condition bugs.

WorldLanguages:
Thanks @towerofnix for the help! The bug is now fixed.

DNin01:
The block widths should be updated when the addon is dynamically disabled or when "bold text" is toggled. Also, after a second dynamic enable, it seems the text size isn't re-applied until a setting is changed.

WorldLanguages:
> when "bold text" is toggled

Yeah, handling that separately is causing its own set of issues because even though the font size did adjust properly, it looks wrong because it adjusted before the bold text had a chance to apply.

Will fix these tomorrow

DNin01:
Yes! It works great!

But I just realized that the blocks inside the Insert blocks by name addon's popup still require a page refresh to update.

WorldLanguages:
> But I just realized that the blocks inside the Insert blocks by name addon's popup still require a page refresh to update.

Forgot about that addon!

We should also test changing the font size while the "make a block" modal is open.

WorldLanguages:
Also, looks like we're currently styling the "make a X" buttons as if they were blocks:

image

DNin01:
> Also, looks like we're currently styling the "make a X" buttons as if they were blocks:
> 
> image

Interesting. Maybe we could turn that into a feature?

Tacodiva:
Just pushed fixes the issues with the middle click popup :blush:

addons/custom-block-text/userscript.js
```diff
@@ -0,0 +1,89 @@
...
+  textShadowCss.disabled = true;
+  document.head.appendChild(textShadowCss);
+
+  const updateBlockly = () => {
+    blocklyInstance.Field.cacheWidths_ = {}; // Clear text width cache
+    updateAllBlocks(vm);
+  };
+
+  const setFontSize = (wantedSize) => {
+    if (wantedSize !== 100) document.documentElement.style.setProperty("--customBlockText-sizeSetting", wantedSize);
```
Tacodiva, line 47: Nitpick, but could you not use the "customCssVariables" key in `addon.json` to do this automatically?

WorldLanguages: Yeah, but everything here is intentionally handled only by the userscript to make timing 0% dependent on any code outside this addon.
This addon doesn't use userstyles either for this same reason.

WorldLanguages:
TODOs:

- [ ] This affects the "make a block" modal. Should it? Is it easy to avoid it? Also, test changing the font size while this the modal is being shown, and right after closing the modal.
- [ ] "Make a Variable", "Make a List" and "Make a Block" ideally shouldn't be styled, is it easy to exclude them?
- [ ] This addon could benefit from #4919 as it interacts with Blockly workspaces.

CST1229:
> Question: is there really a difference between this setting to increase the font size, and zooming into the workspace? There really isn't right? That's a bit confusing.

Increasing the font size doesn't increase outline and padding sizes.

DNin01:
> Question: is there really a difference between this setting to increase the font size, and zooming into the workspace? There really isn't right? That's a bit confusing.

You can't change the zoom level of the blocks palette, so this is the only way to make that text larger, other than changing the entire page's zoom level.

DNin01:
Interesting issue: when pulling up the Make a Block popup, the blocks on the blocks palette suddenly stop adapting to text style changes, but everything else adapts.

To reproduce:
1. Click "Make a Block"
2. Change the text size

The block in the Make a Block menu itself also doesn't adapt.

WorldLanguages:
> Interesting issue: when pulling up the Make a Block popup, the blocks on the blocks palette suddenly stop adapting to text style changes, but everything else continues to adapt.

Yeah, #6227 should fix that

DNin01:
@WorldLanguages Apparently this branch also [got on the ScratchAddons repo](https://github.com/ScratchAddons/ScratchAddons/activity?ref=large-block-text-setting)? Noticed while I was trying to merge `ScratchAddons:master` into the branch from the web UI. Can we safety delete it?

WorldLanguages:
I'm not sure, we can delete it after merging this

DNin01:
> I'm not sure, we can delete it after merging this

It looks to me like we won't need it. We can restore it if deleting it causes issues.

DNin01:
Problem solved!

Samq64 (commented):
Could the style elements be moved to one or more CSS files?

WorldLanguages:
> Could the style elements be moved to one or more CSS files?

@Samq64 I'm not sure if that can be done without sacrificing reliability, as loading CSS files from disk may take a few milliseconds, leading to race conditions.

WorldLanguages:
> > Interesting issue: when pulling up the Make a Block popup, the blocks on the blocks palette suddenly stop adapting to text style changes, but everything else continues to adapt.
> 
> Yeah, #6227 should fix that

The PR was postponed to v1.35.0 milestone, so this PR should also be postponed

WorldLanguages:
> Interesting issue: when pulling up the Make a Block popup, the blocks on the blocks palette suddenly stop adapting to text style changes, but everything else continues to adapt.
> 
> To reproduce:
> 
> 1. Click "Make a Block"
> 2. Change the text size
> 
> The block in the Make a Block menu itself also doesn't adapt.

Fixed!

addons/custom-block-text/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Customizable block text style",
-  "description": "Changes the thickness of the text on blocks and optionally adds a text shadow.",
+  "description": "Allows you to make a few adjustments to block text to make it easier to read.",
```
WorldLanguages, line 3: Maybe add the keywords thickness, bold, and/or size to the description?

addons/custom-block-text/userscript.js
```diff
@@ -0,0 +1,93 @@
...
+  const blocklyInstance = await addon.tab.traps.getBlockly();
+
+  // Handling the CSS from here instead of a userstyle is much more stable, as
+  // there's no code outside of this addon dynamically toggling the styles.
+  // This way, we can clearly control the execution order of style operations.
+  // For example, we always want to call updateAllBlocks() after the styles
+  // were updated according to the user's settings, not before.
+  const fontSizeCss = document.createElement("style");
+  fontSizeCss.textContent = `
+    .blocklyText,
```
WorldLanguages, line 17:
```suggestion
    .blocklyText:not(.blocklyFlyoutButton > .blocklyText),
```

This change would make the CSS not affect the font size of the "make a variable", "make a block", etc. buttons.

image

WorldLanguages:
Should this addon and `custom-block-shape` affect the "Make new block" block preview?
Currently, they do. It doesn't cause any issues unless you change the settings while the "new block" modal is open. In that case, you need to close and reopen it for it to readjust. No other known issues or incompatibilities with other addons.

WorldLanguages:
We could also attempt to make the block category labels bigger/smaller depending on the block size.
Currently there's no way to change their size.
If we decide to change these, the "Make a Variable" and other buttons would be affected too.

This is how it looks right now: (font size 175%)
image

mxmou:
> Should this addon and `custom-block-shape` affect the "Make new block" block preview?

I think so.

WorldLanguages:
> We could also attempt to make the block category labels bigger/smaller depending on the block size.
> Currently there's no way to change their size.
> If we decide to change these, the "Make a Variable" and other buttons would be affected too.

I think we should also increase the font size of those, and merge this PR.

addons/custom-block-text/userscript.js
WorldLanguages: I think we decided against this.

WorldLanguages (approved):

mxmou (approved):

WorldLanguages:
From feedback: custom font size for editor comments. Seems like the missing piece of the puzzle. It's the only UI element in the Code tab that we don't provide a custom font size setting for.

PR #5931, opened 2023-04-15
Allow changing editor popup backdrop color

### LINKED ISSUE #5930: Change popup background colors with `editor-dark-mode`
### Why this would be helpful

Currently, the colors of backgrounds of popups (for renaming/creating variables, etc.) are determined by the Highlight color setting, but what if you wanted to customize this color separately from the rest of the UI?

Also, what if you wanted to customize the transparency?

### My suggestion

Add a color setting for popup backgrounds in `editor-dark-mode`.

It'd also be cool if we could add an optional blur effect to blur out everything under the popup. This could be done using `backdrop-filter: blur(x);`.

### Possible alternatives

_No response_

### Additional context

_No response_

DNin01:
Resolves #5930

### Changes

Added a setting to `editor-dark-mode` for changing the color of the popup backdrop (let me know if you can think of a better name). This is the blue cover behind popups in the editor like the popup that appears when you are creating or renaming a variable, list or broadcast message. It also supports adjusting transparency!

All of the presets except for "Scratch's default colors" have also been updated to make use of this feature.

Screenshot:

A popup with a transparent gray background behind it

### Reason for changes

So that people could make it different than the highlight color.

### Tests

Tested on Edge 112.


WorldLanguages:
Reminder if this gets merged/approved: this would require preset migration

addons/editor-dark-mode/addon.json
```diff
@@ -1139,6 +1152,7 @@
         "primary": "#5498c7",
         "highlightText": "#21211f",
         "menuBar": "#c0c3c6",
+        "popup": "#5498c7e6",
```
mxmou, line 1155:
```suggestion
        "popup": "#00000099",
```
Transparent black would look better in this preset too.

addons/editor-dark-mode/experimental_editor.css
```diff
@@ -400,7 +400,7 @@ img[src="/static/assets/7bd7487b704797cb5ab3cb441486ea70.svg"],
   background-color: var(--editorDarkMode-primary-transparent35);
 }
 [class*="modal_modal-overlay_"] {
-  background-color: var(--editorDarkMode-primary-overlay);
+  background-color: var(--editorDarkMode-popup);
```
mxmou, line 403: `primary-overlay` isn't used anywhere else, so you can remove its definition from `addon.json`.

mxmou (approved):

Samq64:
I think blue is too bright for the darker preset.

Secret-chest:
I think the dark ones should just have dark grey.

WorldLanguages:
Note: I won't review the color changes, but I'll implement the preset migrations.

I encourage we discuss what Samq and Secret-chest said before merging.

DNin01:
@WorldLanguages Alright, I just went through and changed all the dark themes that were still using the blue color.

addons/editor-dark-mode/addon.json
```diff
@@ -1014,6 +1011,7 @@
         "primary": "#4d97ff",
         "highlightText": "#4d97ff",
         "menuBar": "#2f4066",
+        "popup": "#333a",
```
mxmou, line 1014:
```suggestion
        "popup": "#000a",
```
I think transparent black looks better than dark gray - especially in this preset because it mostly uses shades of blue.

mxmou (changes requested):

addons/editor-dark-mode/addon.json
```diff
@@ -1064,6 +1063,7 @@
         "primary": "#4d97ff",
         "highlightText": "#4d97ff",
         "menuBar": "#4d97ff",
+        "popup": "#333a",
```
mxmou, line 1066: Can you change it to `#000a` here as well? Everything else looks good.

mxmou (approved):

WorldLanguages:
I will implement the following migration:

- If settings match the "Dark Editor" preset, set the "popup" setting to `#47566be6`
- If settings match the "TurboWarp Dark" preset, set the "popup" setting to `#333a`
- If settings match the "Scratch 2.0" preset (or alternatively the "Scratch 1.x preset"), set the "popup" setting to `#00000099`
- If none of the conditions above match (which includes the 3.Darker, 3.Dark and Experimental Dark presets) then no migration will be done and the setting will be set to its default, `#000a`, which matches the new defaults in these 3 presets.

I need some help at defining what "if settings match X preset" should mean. @mxmou @DNin01 Would be great if you could define which set of settings should be matched to check if the user was previously using a preset. I don't think we want to be strict and match against ALL of the preset defaults, right?

WorldLanguages:
See #3904 for reference. That appears to be the last time that the editor-dark-mode presets have gone through a change that required a preset migration.

mxmou:
> I will implement the following migration:
> 
> * If settings match the "Dark Editor" preset, set the "popup" setting to `#47566be6`
> * If settings match the "TurboWarp Dark" preset, set the "popup" setting to `#333a`
> * If settings match the "Scratch 2.0" preset (or alternatively the "Scratch 1.x preset"), set the "popup" setting to `#00000099`
> * If none of the conditions above match (which includes the 3.Darker, 3.Dark and Experimental Dark presets) then no migration will be done and the setting will be set to its default, `#000a`, which matches the new defaults in these 3 presets.

This will set the color to the default `#000a` if the settings don't match any preset. I think it's better to use a transparent version of the highlight color setting to make the modal backdrop look the same as before this PR.

mxmou:
> See #3904 for reference. That appears to be the last time that the editor-dark-mode presets have gone through a change that required a preset migration. In that PR, the following settings had to exactly match the preset for the migration to happen: `page`, `primary`, `highlightText`, `menuBar`, `activeTab`, `tab`, `selector`, `selector2`, `selectorSelection`, `accent`, `input`, `workspace`, `categoryMenu`, `palette`, `fullscreen`, `stageHeader` and `border`. At that time, the excluded settings were: `dots`, `textShadow` (now part of a different addon), `affectPaper`, `affectStage`, `darkComments` (now part of a different addon) and `darkScrollbars` (now removed).

I think it still makes sense to check these settings, except `fullscreen` and `stageHeader` - full screen mode isn't really a part of the editor.

WorldLanguages:
> This will set the color to the default `#000a` if the settings don't match any preset. I think it's better to use a transparent version of the highlight color setting to make the modal backdrop look the same as before this PR.

I see. How exactly is the "transparent version of the highlight color" calculated as of v1.31?

mxmou:
> > This will set the color to the default `#000a` if the settings don't match any preset. I think it's better to use a transparent version of the highlight color setting to make the modal backdrop look the same as before this PR.
> 
> I see. How exactly is the "transparent version of the highlight color" calculated as of v1.31?

It's the highlight color with 90% opacity. The migration code could just append `e6` to the setting value.

Secret-chest:
The dark presets should use a black backdrop.

mxmou:
> The dark presets should use a black backdrop.

Most of them will use transparent black or dark gray. Dark Editor will still use dark blue.

WorldLanguages:
When trying to write robust migration code I'm getting into the same problem as described here: https://github.com/ScratchAddons/ScratchAddons/pull/5896#discussion_r1181981293

I'd say we merge both PRs and I'll later submit a single new PR that adds robust migration for both cases

addons/editor-dark-mode/addon.json
```diff
@@ -1188,6 +1192,10 @@
     "type": "palette",
     "colors": ["primary", "menuBar", "page", "accent", "workspace", "selector"]
   },
+  "latestUpdate": {
+    "version": "1.32.0",
+    "newSettings": ["popup"]
```
WorldLanguages, line 1197:
```suggestion
    "newSettings": ["popup"],
    "temporaryNotice": "You can now customize the backdrop color of popups, such as the create new variable prompt."
```

My wording is probably bad, but I think it would be worth to explain what the new setting does. In the future this will be explained with setting descriptions.

DNin01: Could we go with this?
```suggestion
    "newSettings": ["popup"],
    "temporaryNotice": "You can now customize the color of popup backdrops (the solid-color background behind prompts for actions like naming a new variable) separately."
```

WorldLanguages (approved):

Issue #3854, opened 2021-11-27
Highlight project creator username


mxmou:
### Is your request related to a problem? Please describe.

Scratch doesn't make a distinction between comments posted by the creator of a project and comments by other users.

### Describe the solution you'd like

An addon that adds an indicator to the creator's username - either a background color:
![image](https://user-images.githubusercontent.com/51849865/143686482-53316263-5b5a-459d-9d2f-5430993e675f.png)
or a label:
![image](https://user-images.githubusercontent.com/51849865/143686727-f97f9acf-ba20-475a-80eb-f8fedfe03ea6.png)

Something similar could be done on profiles, studios (with labels like "Host", "Manager", and "Curator"), and in the forums (for the original poster).

WorldLanguages:
Screenshot from #5932:

![image](https://github.com/ScratchAddons/ScratchAddons/assets/17484114/99813281-3af9-4af5-9c00-317f9c1e0758)

PR #5932, opened 2023-04-15
[new addon] OP badge for forum topics

mdwalters:
### Changes
Adds an OP badge to the user that created a forum topic.
### Reason for changes
I feel like there is a need for an OP badge so that people can see who made the topic, without looking at the first post on that topic.
### Tests
It has been tested on Chrome 112.


addons/op-badge/userscript.js
```diff
@@ -0,0 +1,17 @@
+export default async function({ addon, console }) {
+    const posts = document.querySelectorAll(".blockpost");
+
+    const op = await fetch(`https://scratchdb.lefty.one/v3/forum/topic/posts/${window.location.href.split("/")[5]}/0?o=oldest`).then(res => res.json());
```
DNin01, line 4: What is this fetch used for? Since forum topic pages display poster usernames, there's probably a way to do this without having to fetch from ScratchDB, right?

I haven't tried this myself yet, by the way.

Samq64: That wouldn't be hard to do on the first page, but what about other pages.

Samq64:
Related: #3854

addons/op-badge/userscript.js
```diff
@@ -0,0 +1,17 @@
...
+
+    for (const i of posts) {
```
Samq64, line 6: Could [addon.tab.waitForElement](https://scratchaddons.com/docs/reference/addon-api/addon.tab/#using-addontabwaitforelement) be used instead?

addons/op-badge/userscript.js
```diff
@@ -0,0 +1,17 @@
...
+        if (i.querySelector(".username").innerText == op[0].username) {
+            let br = document.createElement("br");
+            let op_badge = document.createElement("span");
+
+            op_badge.innerText = "OP";
```
Samq64, line 11: I think "Original poster" should be used instead of "OP". Also, all addon string should be [localized](https://scratchaddons.com/docs/localization/localizing-addons/).

DNin01: We could also use the word "Author", like GitHub does.

addons/op-badge/userscript.js
WorldLanguages: @Samq64 waitForElements is not needed in forum posts (unless you set runAtComplete to false)

addons/op-badge/userscript.js
WorldLanguages: Also see https://scratchaddons.com/docs/develop/userscripts/#localizing-userscripts

addons/op-badge/userscript.js
WorldLanguages: I think it would be worth it to avoid this fetch call if it's the 1st page. But it's not strictly necessary.

WorldLanguages: Is it necessary to use a third party service however? (ScratchDB)

addons/op-badge/userscript.js
```diff
@@ -0,0 +1,17 @@
+export default async function({ addon, console }) {
+    const posts = document.querySelectorAll(".blockpost");
+
+    const op = await fetch(`https://scratchdb.lefty.one/v3/forum/topic/posts/${window.location.href.split("/")[5]}/0?o=oldest`).then(res => res.json());
+
+    for (const i of posts) {
+        if (i.querySelector(".username").innerText == op[0].username) {
```
WorldLanguages, line 7: This won't work for Scratch Team members since their username will end with a `star*`

addons/op-badge/userscript.js
```diff
@@ -0,0 +1,17 @@
...
+            let br = document.createElement("br");
+            let op_badge = document.createElement("span");
```
WorldLanguages, line 9: You can probably achieve the same result (br + span) by using a div. We usually don't use <br> tags in addons but it's not enforced

addons/op-badge/userscript.js
rgantzos: There's not really an API as far as I'm aware, but you could just fetch the HTML like this:
```
let id = "499628"
let data = await (await fetch(`https://scratch.mit.edu/discuss/topic/${id}/`)).text()
let html = document.createElement("html")
html.innerHTML = data.replace("<!DOCTYPE html>", "")
let author = html.querySelector(".black.username").textContent)
```

WorldLanguages: You should probably use DOMParser instead of creating an "html" element like that

addons/op-badge/userscript.js
mdwalters: It displays the scratch team usernames normally on ScratchDB...

WorldLanguages: > It displays the scratch team usernames normally on ScratchDB...

Sure, but `i.querySelector(".username").innerText` will include the asterisk, so both strings won't be equal.
This is just an edge-case that I think was worth mentioning, since I wasn't sure someone else would bring it up.

WorldLanguages: You're right, unlike profile pages, an asterisk isn't added in the forums. My bad.

addons/op-badge/userscript.js
WorldLanguages: Setting this as resolved as waitForElement is discouraged in this context.

addons/op-badge/userscript.js
```diff
@@ -0,0 +1,16 @@
+export default async function({ addon, console, msg }) {
+    const posts = document.querySelectorAll(".blockpost");
+
+    const html = new DOMParser().parseFromString(await fetch(`https://scratch.mit.edu/discuss/topic/${window.location.href.split("/")[5]}/`).then(res => res.text()), "text/html");
```
WorldLanguages, line 4: If this is already running on the first page, we can set `html` to `document` and avoid the DOMParser call and the fetch.

WorldLanguages: Self TODO: it would also be better if the fetch call was unauthenticated to let Scratch hit the cache. Haven't checked whether current code sends auth cookies or not.

Chiroyce1:
I personally think it should have a badge like design instead of just plaintext
image
Of course this is just a mockup - better design ideas would be great
(also why is there a text-shadow by default on it? djangobb styling?)

mdwalters:
> I personally think it should have a badge like design instead of just plaintext image Of course this is just a mockup - better design ideas would be great (also why is there a text-shadow by default on it? djangobb styling?)

yeah, I think I could do that...

addons/op-badge/userscript.js
```diff
@@ -1,12 +1,12 @@
 export default async function({ addon, console, msg }) {
     const posts = document.querySelectorAll(".blockpost");
 
-    if (window.location.href.split("?").join("").split("&")[0] == "page=1" || window.location.href.split("?").join("").split("&").length == 0) {
+    if (window.location.href.split("?")[1] == "page=1" || window.location.href.split("?").length == 1) {
```
WorldLanguages, line 4: Try using `(new URLSearchParams(location.search).get("page")) || 1`

WorldLanguages:
It's probably a good idea to make this PR fully resolve #3854 (by @mxmou) by also adding this functionality to project comments. If the PR author isn't interested, one of the maintainers can work on it.

Secret-chest:
> It's probably a good idea to make this PR fully resolve #3854 (by @mxmou) by also adding this functionality to project comments. If the PR author isn't interested, one of the maintainers can work on it.

Profile comments should have an Owner badge, and studio comments should have Curator, Manager and Host ones.

Chiroyce1:
> Profile comments should have an Owner badge, and studio comments should have Curator, Manager and Host ones.

I understand for projects and studio owner/manager - but we'd have to load all curators at first to put a badge next to every curator comment, some studios have a TON of curators.

WorldLanguages:
@BroJac5246 Good point, we shouldn't forget that some studios may have more than 40 managers. But those are probably all inactive, since you can't promote anyone until you go below the limit.

BroJac5246:
> @BroJac5246 Good point, we shouldn't forget that some studios may have more than 40 managers. But those are probably all inactive, since you can't promote anyone until you go below the limit.

It's still worth keeping in mind, though.

mdwalters:
> Supporting a "curator" badge would be tricky because of what Chiroyce1 just said.

how about we just do a label for the comment thread creator?

BroJac5246:
> > Supporting a "curator" badge would be tricky because of what Chiroyce1 just said.
> 
> how about we just do a label for the comment thread creator?

Some people will probably disagree because that's not what we'd do in project comments, but I agree because studio and project comments have a different dynamic.

WorldLanguages:
> how about we just do a label for the comment thread creator?

Hmm, would that really be useful? Studios have a 25 reply limit per thread anyway.

Chiroyce1:
@mdwalters Did you forget to stage `forums.js` and `projects.js` before committing/pushing?

WorldLanguages:
image
image

Secret-chest:
Please set the border radius in pixels so ot doesn't look like this.

addons/op-badge/addon.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "Show OP badge in forums and projects",
+  "description": "Adds a OP badge to the user that created a forum topic or project.",
```
WorldLanguages, line 3: Name and description need to be changed. Ideally at least one of them should mention "original poster" somewhere. Possibly "project author" too.
Should also mention the word "comment(s)".

addons-l10n/en/op-badge.json
```diff
@@ -0,0 +1,3 @@
+{
+  "op-badge/op": "Original poster"
```
WorldLanguages, line 2: Should this be changed to `Project author` (or similar) in project pages?

mdwalters:
I think it should

addons-l10n/en/op-badge.json
mdwalters: i think so too

WorldLanguages:
image

Secret-chest:
I'd set the border radius to a very high value so it's fully rounded

Chiroyce1:
I agree with @Secret-chest, maybe also increase the horizontal padding a bit?

addons/op-badge/style.css
```diff
@@ -0,0 +1,17 @@
+.sa-original-poster {
+  display: inline-block;
+  background-color: var(--darkWww-button, #4d97ff);
+  color: var(--darkWww-button-text, white);
+  border-radius: 8px;
```
WorldLanguages, line 5:
```suggestion
  border-radius: 16px;
```
?

addons/op-badge/style.css
```diff
@@ -0,0 +1,17 @@
...
+  text-shadow: none;
+  padding: 4px;
```
WorldLanguages, line 7:
```suggestion
  padding: 6px;
```
?

addons/op-badge/addon.json
WorldLanguages: @DNin01 Help?

WorldLanguages:
image

addons/op-badge/addon.json
DNin01, line 3: @WorldLanguages Sure thing! How about this?
```suggestion
  "name": "Highlight forum topic/project creator in conversations",
  "description": "Adds a badge next to the original poster in forum topics and next to the project creator in project comments.",
```

addons-l10n/en/op-badge.json
```diff
@@ -0,0 +1,4 @@
+{
+  "op-badge/op": "Original poster",
+  "op-badge/project-author": "Project author"
```
DNin01, line 3:
```suggestion
  "op-badge/project-author": "Project creator"
```

addons/op-badge/addon.json
WorldLanguages: This one's hard...

Samq64: Should we mention forum topics or projects in the name? I think mentioning both makes it too long.

WorldLanguages: I would say we prioritize project comments and "project creator" and maybe in the future we can consider whether it makes sense to split this addon into two. It might make sense if #6009 and #5937 both go live 🤔

DNin01:
```suggestion
  "name": "Highlight project creator in comments",
  "description": "Adds a badge next to the project creator in project comments sections. Also adds a badge next to the original poster in forum topics.",
```

Samq64 (approved):

WorldLanguages:
Any last minute reviews...?

addons-l10n/en/op-badge.json
```diff
@@ -0,0 +1,4 @@
+{
+  "op-badge/op": "Original poster",
+  "op-badge/project-author": "Project creator"
```
mxmou, line 3:
```suggestion
  "op-badge/op": "Original Poster",
  "op-badge/project-author": "Project Creator"
```
Title case is consistent with similar labels in Scratch ("Scratch Team", "Forum Moderator").

addons/op-badge/style.css
```diff
@@ -0,0 +1,19 @@
+.sa-original-poster {
+  display: inline-block;
+  background-color: var(--darkWww-button, #4d97ff);
```
mxmou, line 3: Could this be a different color in the forums when 2.0 → 3.0 is disabled - maybe `var(--darkWww-button-scratchr2PostHeader)`?

The font size should probably be set explicitly instead of relying on the default for `<small>`.

addons-l10n/en/op-badge.json
DNin01: Is this capitalization necessary in project comments?

mxmou: Most strings on the project page also use title case: "Add to Studio", "Copy Link", "Load More".

WorldLanguages: It's not 100% consistent though: "See inside", "View all"

BroJac5246: > It's not 100% consistent though: "See inside", "View all"

True, but mxmou's examples are more closely related to this.

addons/op-badge/style.css
WorldLanguages: > Could this be a different color in the forums when 2.0 → 3.0 is disabled - maybe `var(--darkWww-button-scratchr2PostHeader)`?

@mxmou Could you clarify? Did you mean "when dark mode is disabled" instead of "when 2.0 → 3.0 is disabled"?

mxmou: This shade of blue (`#4d97ff`) is a part of the Scratch 3.0 color scheme. It isn't used anywhere on scratchr2 pages unless 2.0 → 3.0 is enabled.

WorldLanguages: So what color/CSS variable should we use in the forums?

mxmou: Probably `--darkWww-button-scratchr2PostHeader`.

WorldLanguages: And what would be the fallback if website dark mode is disabled? 🤔

mxmou: `#28a5da`

WorldLanguages (approved):

PR #5939, opened 2023-04-17
Keep "featured dangos" addon

### LINKED ISSUE #5891: Remove (or keep) featured-dangos addon
### Why this would be helpful

April Fools has ended already. We need to decide what to do about the featured-dangos addon (#5831)

### My suggestion

Removing the featured-dangos addon completely.

### Possible alternatives

Keeping it as an easter egg which the user can toggle. If we do this, I suggest changing the addon id to `featured-dangos-v2` so that we can easily set `enabledByDefault: false` to the "new" addon.

### Additional context

See #5831 and #5807

Samq64:
Closes #5891
Closes #5934

### Changes

Applies the changes in https://github.com/ScratchAddons/ScratchAddons/issues/5891#issuecomment-1510030004 so the featured dangos addon won't get removed.

### Reason for changes

Some people want the featured dangos addon to stay as an Easter egg.

### Tests

Tested on Chromium and Firefox. I didn't test the migration code or cases where the projects in the studio are changed or removed.


CST1229:
I think the ID should be renamed to featured-dangos-v2 (or something similar), so the addon won't stay enabled even when the migration code gets removed.

WorldLanguages:
As I said in the issue, please let's avoid writing migration code. Just rename the addon ID like CST just said, and don't set enabledByDefault to true.

addons/featured-dangos-v2/userscript.js
```diff
@@ -0,0 +1,47 @@
+export default async function ({ addon, console, msg }) {
+  // Fetch projects from a studio
+  const fetchStudio = async (studio) => {
+    const response = await fetch(`https://api.scratch.mit.edu/studios/${studio}/projects/`);
+    return response.json();
+  };
+
+  const projects = await fetchStudio("33036797");
```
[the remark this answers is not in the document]
WorldLanguages: Would be ideal if the studio was created by an org member who voted in favor of keeping the addon

Secret-chest:
What about turning this into an addon to add any studio to the front page?

WorldLanguages:
@Secret-chest Idea is creative, but it's probably a better idea to discuss it in a dedicated issue

addons/featured-dangos-v2/userscript.js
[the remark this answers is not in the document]
WorldLanguages: Are you interested? @Hans5958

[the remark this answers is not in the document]
Hans5958: I can host it, but not sure if I would maintain the contents.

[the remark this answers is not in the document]
WorldLanguages: @Hans5958 Then should we keep the addon?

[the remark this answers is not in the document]
Hans5958: It's still a yes to me, but I'm just can't gurantee my availability regarding the studio.

Hans5958:
Has the issue been made? I wanted to see that.

Samq64:
> Has the issue been made? I wanted to see that.

#5990

WorldLanguages:
The "featured dangos" addon has been technically inaccessible by the average user for almost 2 months. Still interested in merging this?

Samq64:
> The "featured dangos" addon has been technically inaccessible by the average user for almost 2 months. Still interested in merging this?

Only if this PR doesn't need too much extra work.

addons/featured-dangos-v2/addon.json
```diff
@@ -0,0 +1,18 @@
...
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/"]
+    }
+  ],
+  "tags": ["community", "easterEgg"],
+  "versionAdded": "1.31.1"
```
WorldLanguages, line 17:
```suggestion
  "versionAdded": "1.33.0"
```

addons/featured-dangos-v2/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Featured dangos",
+  "description": "An Apil Fools joke from 2023 that adds a row to the front page for dango-related projects.",
```
Hans5958, line 3:
```suggestion
  "description": "An April Fools joke from 2023 that adds a row to the front page for dango-related projects.",
```

BroJac5246: Should we also specify that this was originally added by Scratch Addons?

BroJac5246:
The addon ID in `addons.json` needs to be updated. Also, how did you test this since it immediately crashes Scratch Addons?

Samq64:
> The addon ID in `addons.json` needs to be updated. Also, how did you test this since it immediately crashes Scratch Addons?

I tested it before switching addon ids. It's fixed now.

addons/featured-dangos-v2/addon.json
```diff
@@ -0,0 +1,18 @@
...
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/"]
+    }
+  ],
+  "tags": ["community", "easterEgg"],
+  "versionAdded": "1.33.0"
```
WorldLanguages, line 17: Apparently easter egg addons might appear under "other new addons" which is probably unintended.

addons/featured-dangos-v2/userscript.js
```diff
@@ -0,0 +1,47 @@
...
+
+    // Use correct project thumbnail
+    project.querySelector("img").src = `//uploads.scratch.mit.edu/projects/thumbnails/${id}.png`;
+
+    // Use correct project title
+    project.querySelector("a[title]").textContent = title;
+    project.querySelector("a[title]").title = title;
+
+    // Use correct project author
+    project.querySelector(".thumbnail-creator").firstElementChild.innerText = username;
```
WorldLanguages, line 41: `TypeError: Cannot read properties of null (reading 'firstElementChild')`

See comment above:
> At this point, the Featured Dangos row might be 5 "Project" placeholders linking to /projects/1, or possibly the actual featured projects. We will be handling both cases.

This is not handling the second case properly, crashing the addon.

addons/featured-dangos-v2/userscript.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Clone Featured Projects row and its children
+  const featuredDangosRow = featuredProjectsRow.cloneNode(true);
+  featuredDangosRow.classList.add("sa-featured-dangos-row");
+
+  // At this point, the Featured Dangos row might be 5 "Project" placeholders linking to
+  // /projects/1, or possibly the actual featured projects. We will be handling both cases.
+
+  // Change row title
+  featuredDangosRow.querySelector(".box-header > h4").textContent = msg("row-title");
```
WorldLanguages, line 24: The addon ID also needs to be changed in the file name of `addons-l10n/en/featured-dangos.json`, or this won't properly get the message.

addons/featured-dangos-v2/userscript.js
WorldLanguages: You can emulate this case by using Chrome Devtools' "network request blocking" feature to block `api.scratch.mit.edu/proxy/featured`

addons/featured-dangos-v2/addon.json
WorldLanguages: It's not wrong though, I simply didn't expect this 🤔 

image

addons/featured-dangos-v2/userscript.js
Samq64: Is it just to prevent the addon from crashing when Scratch's API isn't working like it should?

addons/featured-dangos-v2/addon.json
Samq64: Should the version added be reverted to prevent this?

addons/featured-dangos-v2/userscript.js
WorldLanguages: No, it's a race condition that can depend on browser caching, Scratch server speed or the user's internet speed.
runAtComplete:true has very few guarantees in scratch-www pages.

WorldLanguages: I recommended the "network request blocking" feature because there's no easy way to slow down a specific request.

addons/featured-dangos-v2/addon.json
WorldLanguages: Yeah, I guess that's easier than changing the settings page code.

addons-l10n/en/featured-dangos-v2.json
```diff
@@ -0,0 +1,4 @@
+{
+  "featured-dangos-v2/row-title": "Featured Dangos",
+  "featured-dangos-v2/note": "Added by Scratch Addons browser extension"
```
WorldLanguages, line 3: This string can be removed as it's now unused by the userscript.

addons/featured-dangos-v2/userscript.js
Samq64: Ok, I'll need to use a different solution because for now I surround the username code with an if.

WorldLanguages: That's good enough IMO, the original version of the addon didn't include the authors at all.

WorldLanguages (approved):

addons/featured-dangos-v2/userscript.js
WorldLanguages: This probably wouldn't happen in practice because the studio fetch is done first. My bad. But technically it was a bug.

PR #5940, opened 2023-04-17
Split "addonSettings" storage item into 3 different items

### LINKED ISSUE #5849: Split "addonSettings" storage item into 3 different items to avoid reaching 8192 byte limit
### Why this would be helpful

As we add more settings each release, we'll soon reach 8192 bytes for the `addonSettings` storage item. This item is stored using the `chrome.storage.sync` API, which has this limit. See QUOTA_BYTES_PER_ITEM on https://developer.chrome.com/docs/extensions/reference/storage/. Also in MDN: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/sync#storage_quotas_for_sync_data

The maximum number of bytes that an extension can use as a whole is 102400 bytes, which is 12 times greater than the per-item limit.

### My suggestion

Initially suggested here: #5190

My suggestion is to use three different storage items: `addonSettings1`, `addonSettings2` and `addonSettings3`. This would basically triple the available storage for addon settings, and could stay as a semi-permanent solution for a reasonable amount of time. After reading these 3 separate items from disk, they will be merged in memory. When writing to disk, they will need to be split using some specific criteria before setting the storage items.
We could separate those 3 by addon ID, such as those that start with a-e, then f-o and lastly p-z. But we could also use any other criteria that would make sense.
The `chrome.storage` APIs let you easily get and set multiple items at the same time, so the amount of code changes needed to implement this won't be very big.
This solution would be transparent for content script and addon code. Only background page, popup and settings page code will need modifications. Settings JSON files will continue to be valid and their structure won't change.

### Possible alternatives

DNin01 has suggested dedicated storage items for addons that have table settings, such as `discuss-button` and `cloud-games`. This is also related to issue #5666

### Additional context

Related issues:
Issue #4721 and PR #4750 (we did reach the per-key limit already, but only if the user had used the settings import feature, which didn't have any basic compression considering that most addons don't have any settings)
#5190
#5666

WorldLanguages:
Resolves #5849

### Changes

For new users, it creates three different storage items for addon settings: addonSettings1, addonSettings2 and addonSettings3.
For existing users, it splits the addonSettings storage item in three parts (using the first character of the addon ID) and reloads the extension.
The structure of settings JSON files (which can be exported/imported) remains untouched.

### Reason for changes

See #5849. There's a very low size per-key limit.

### Tests

Still a draft, so not tested yet.
Will need deep testing in both Chromium and Firefox.

### DIFF
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -36,7 +36,28 @@ const updatePresetIfMatching = (settings, version, oldPreset = null, preset = nu
   }
 };
 
-chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {}, addonsEnabled = {} }) => {
+async function transitionToNewStorageKeys(addonSettings) {
+  // This will not delete the addonSettings storage item
+  chrome.storage.sync.set(
+    {
+      ...minifySettings(addonSettings, null),
+    },
+    () => {
+      // Reload extension after 500ms
+      setTimeout(() => chrome.runtime.reload(), 500);
+    }
+  );
+}
+
+const ADDON_SETTINGS_KEYS = ["addonSettings", "addonSettings1", "addonSettings2", "addonSettings3"];
+chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems) => {
+  if (storageItems.addonSettings && !storageItems.addonSettings3) {
+    return transitionToNewStorageKeys(storageItems.addonSettings);
+  }
+  const addonsEnabled = storageItems.addonsEnabled || {};
+  const addonSettings = storageItems["addonSettings3"]
+    ? { ...storageItems.addonSettings1, ...storageItems.addonSettings2, ...storageItems.addonSettings3 }
+    : {};
   const func = () => {
     let madeAnyChanges = false;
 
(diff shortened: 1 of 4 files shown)


background/get-addon-settings.js
```diff
@@ -36,7 +36,28 @@ const updatePresetIfMatching = (settings, version, oldPreset = null, preset = nu
   }
 };
 
-chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {}, addonsEnabled = {} }) => {
+async function transitionToNewStorageKeys(addonSettings) {
+  // This will not delete the addonSettings storage item
+  chrome.storage.sync.set(
+    {
+      ...minifySettings(addonSettings, null),
+    },
```
WorldLanguages, line 44: Self TODO: it might be a better idea to rename addonSettings to oldAddonSettings.

Contributors might often jump between PRs that are still in old versions. I still need to have some thought on the best way to solve that. Keeping addonSettings intact does not help, as you end up with 2 parallel storages for settings depending on the branch you're on.

WorldLanguages:
Needs testing in both browsers:

- [ ] Launch the extension for the first time (by clearing the storage)
- [ ] Transitioning the old settings format into the new one
- [ ] Importing settings
- [ ] Exporting settings
- [ ] Changing addon settings

background/get-addon-settings.js
WorldLanguages: Moving this discussion to the thread instead of this comment.

WorldLanguages:
Also worth thinking about the best way to deal with this:

> Contributors might often jump between PRs that are still in old versions. I still need to have some thought on the best way to solve that. Keeping addonSettings intact does not help, as you end up with 2 parallel storages for settings depending on the branch you're on.

The easiest fix is to git pull upstream on every PR branch, most of them shouldn't have conflicts.
We could also try to keep `addonSettings` synced with the new 3 storage items (and viceversa), but that could get complex quickly and we would need to dynamically toggle this behavior or remove it for the actual release. We could theoretically keep this behavior for the actual release but it wouldn't benefit our users in any way in exchange for more complex transition code.

WorldLanguages:
If something goes wrong with this update, keep in mind the `addonSettingsOld` storage item will be a copy of the user's addon settings at the time of the migration.

PR #5942, opened 2023-04-18
Clarify support for Cmd key in turbowarp-player info

iqnite:
Small string change for https://github.com/ScratchAddons/ScratchAddons/pull/5738

### Changes

Clarified support for Cmd+Click on button for macOS users.

### Reason for changes

Mac users might think the Ctrl+Click for new tab feature won't work for them.

### DIFF
--- a/addons/turbowarp-player/addon.json
+++ b/addons/turbowarp-player/addon.json
@@ -14,7 +14,7 @@
     },
     {
       "type": "info",
-      "text": "If the \"Replace player\" mode is selected, you can Ctrl+Click (on macOS Cmd+Click) the button to open the project in a new TurboWarp tab.",
+      "text": "If the \"Replace player\" mode is selected, you can Ctrl+Click (Cmd+Click on macOS) the button to open the project in a new TurboWarp tab.",
       "id": "ctrlToOpen"
     }
   ],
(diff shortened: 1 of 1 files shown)


addons/turbowarp-player/addon.json
```diff
@@ -14,7 +14,7 @@
     },
     {
       "type": "info",
-      "text": "If the \"Replace player\" mode is selected, you can Ctrl+Click the button to open the project in a new TurboWarp tab.",
+      "text": "If the \"Replace player\" mode is selected, you can Ctrl+Click (on macOS Cmd+Click) the button to open the project in a new TurboWarp tab.",
```
WorldLanguages, line 17:
```suggestion
      "text": "If the \"Replace player\" mode is selected, you can Ctrl+Click (Cmd+Click on macOS) the button to open the project in a new TurboWarp tab.",
```

I think the above is more readable, but I can't tell for sure because I'm not a native English speaker.

DNin01: Speaking grammatically, this might work better as "on MacOS, Cmd+Click" or what WorldLanguages suggested.

WorldLanguages (approved):

addons/turbowarp-player/addon.json
BroJac5246: Both are correct, but as a native English speaker I would use W_L's solution if I were writing this.

DNin01 (approved):

Issue #5719, opened 2023-02-11
Groups of settings


DNin01:
### Why this would be helpful

There are a few addons with a large number of settings, most of which being basically the same feature but for a different type of whatever, like which kinds of blocks show related blocks in their right click menus or which types of activity you get notified of. It can make the settings a little harder to navigate to have them all stand out and look just as important and separate as every other setting in that addon.

### My suggestion

Mainly for addons like `scratch-notifier` or `block-switching`, it'd help to have similar settings grouped together. This would allow us to make what could be looked at as one big setting with a list of options.

It's slightly hard to explain, especially in the way I see it, so here's an image:
![A mock-up of a settings page for a "push notifications" feature. Below the other on/off switches, a group of settings called "receive notifications for..." contains checkboxes for specific types of activity.](https://user-images.githubusercontent.com/106490990/218284097-510a25f7-1efb-4e61-92e2-705dfb8ae0a3.png)

### Additional context

This won't improve the look of the settings for most addons, but it's worth considering, especially if we plan to redesign the settings page. Given our easy-to-use format, this design improvement is worth it.

WorldLanguages:
This would be interesting for some addons that have too many settings

Samq64:
This could be done with `<fieldset>`.

mybearworld:
It does.
![A demonstration of it working with fieldset in the pop up](https://github.com/ScratchAddons/ScratchAddons/assets/130385691/dff7c4fb-aedd-4555-8ead-7e03fba84e81)

DNin01:
I think option 1 would allow for more flexibility (you can put the groups exactly where you want and the names could easily be localized). I don't know which one would be the easiest to parse or code.

CST1229:
> I don't know which one would be the easiest to parse or code.

Probably 1.

WorldLanguages:
Option 2 requires the least line changes for existing manifests and doesn't add too much complexity to the manifest schema.
But for translation of the addon groups to actually work, we must assign an ID for each setting group and, optionally, provide a name that will be read by the user and that will be sent for translation.

So the final result would be:

```jsonc
// Option 2 - modified:
"settingGroups": [
 {
    "id": "notifyOn",
    "name": "Notify on" // Can be `null`. See #6068
 }
],
"settings": [
  {
    "name": "Loves",
    "id": "loves",
    "type": "boolean",
    "default": true,
    "group": "notifyOn"
  },
  {
    "name": "Favorites",
[...]

PR #5960, opened 2023-04-23
Add "Button for removing items from profile pages" addon

### LINKED ISSUE #5958: On the favorite projects, following, etc. pages, add an option to remove the item from that page
### Why this would be helpful

Sometimes, you want to clean up your shared or favorites projects, for example. You also don't want to open every single project in a new tab.

### How this addon works

This addon adds an X button to every item of almost every page of rows on your profile page that does the following:
| Page | What the X does |
|--------|--------|
| Shared Projects | Unshare the project |
| Favorite Projects | Unfavorite the project |
| Studios I'm Following | Unfollow the studio |
| Studios I Curate | Leave the studio |
| Following | Unfollow the user | 

### Possible alternatives

* A way to open the user/project without leaving the page. Getting this right would be more difficult than just adding this undo option, though.
* This could be merged with `scratchstats` for a "Better Profiles" addon, but these addons are pretty unrelated.

### Additional context

_No response_

mybearworld:
Resolves #5958

### Changes

Added "Button for removing items from profile pages" addon (internally `remove-from-profile-pages`) that adds buttons that remove certain projects/user/studios from the page.

### Reason for changes

Sometimes, you want to clean up your shared or favorites projects, for example. You also don't want to open every single project in a new tab.

### Tests

Tested on Edge 114.0.1793.0 (Official build) dev (64-bit).


mybearworld:
I'm not sure what I did wrong for the Prettier check to fail...

rgantzos:
@mybearworld You don't need to update the translation files; it's already done by the system that translates it with Transifex

If prettier fails, it means that the code is formatted incorrectly. I'm assuming there will be changes to this code before it's merged (and it looks pretty good right now), so it's not something you need to worry about too much right now, but you might want to just put it into a [prettier formatter](https://prettier.io/playground/).

mybearworld:
> @mybearworld You don't need to update the translation files; it's already done by the system that translates it with Transifex

How else would it know what text to use?

> If prettier fails, it means that the code is formatted incorrectly. I'm assuming there will be changes to this code before it's merged (and it looks pretty good right now), so it's not something you need to worry about too much right now, but you might want to just put it into a [prettier formatter](https://prettier.io/playground/).

I have formatted the code with Prettier on VSCode already (with the .prettierrc.json in the repository).

rgantzos:
> > @mybearworld You don't need to update the translation files; it's already done by the system that translates it with Transifex
> 
> How else would it know what text to use?
> 
Sorry my bad, I misread it.

mybearworld:
The error `fatal: You are not currently on a branch.` doesn't really sound like a problem with the formatting to me - is it just a weird error message?

WorldLanguages:
Don't worry about Prettier

mybearworld:
> Don't worry about Prettier

Okay

mybearworld:
I'll implement dynamic setting changes as well soon.

mybearworld:
Is this too many settings?

Nine settings.

Secret-chest:
Dedicated pages only would be very useful.

BroJac5246:
> Is this too many settings?
> 
> Nine settings.

It's a lot, but they're all useful settings, so it's _probably_ okay.

DNin01:
For user-experience reasons, I don't know if every option has to stay, but I still have to review.

DNin01:
I sure wish we could [https://github.com/ScratchAddons/ScratchAddons/issues/5719](better organize these settings).

addons/remove-from-profile-rows/addon.json
```diff
@@ -0,0 +1,107 @@
...
+      "matches": ["profiles", "https://scratch.mit.edu/users/*/*"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Only show on the dedicated pages",
+      "id": "dedicated_pages",
+      "type": "boolean",
+      "default": false
+    },
```
DNin01, line 25: If this means only show the button to remove items on full lists and not profile pages, then I don't think this option is necessary. Instead, I think the buttons should always only appear on dedicated pages. Profile pages are more likely used for browsing and not editing.

addons/remove-from-profile-rows/addon.json
```diff
@@ -0,0 +1,107 @@
...
+    {
+      "name": "Only show the button on hover",
```
DNin01, line 27: Could probably be renamed:
```suggestion
      "name": "Only show remove button while hovering over items",
```

addons/remove-from-profile-rows/addon.json
```diff
@@ -0,0 +1,107 @@
...
+    },
+    {
+      "name": "Only show confirmation on destructive actions (Unsharing, leaving studios)",
+      "id": "only_destructive",
+      "type": "boolean",
+      "default": true,
+      "if": {
+        "settings": {"show_confirmation": true}
+      }
+    },
```
DNin01, line 46: I don't know about you, but I feel these two settings could be improved. The second setting here _removes_ functionality when _enabled_, and it's also slightly unclear what this should do if disabled - it seems to not imply by itself that when disabled, a confirmation will appear for all actions.

My idea to fix this problem would be to replace these on/off switches with a multi-choice setting named "confirmation before removing" with three choices, "off", "before unsharing projects and leaving studios" and "all actions", or something else.

DNin01 (commented):
I've yet to test the addon, but here are some suggestions for its settings.

addons/remove-from-profile-rows/addon.json
mybearworld: That does sound a lot better, and it also cuts down on the amount of settings required.

mybearworld: I did that, but now it looks like this:
image
And that isn't really good.

Secret-chest: The three-way setting could be first.

mybearworld: image
That's better, but I don't think it's very ideal.

DNin01 (commented):
We still need to work on Dynamic Disable for this addon.
Xs placed over images

Luckily, 2.0 → 3.0 doesn't break anything. Nice!

What is slightly concerning is accidentally clicking the X, especially on profile pictures which are already small. Maybe it could be placed _on_ the corner of the thumbnails, like in the Scratch-2.0 studio design, further out of the way. It might also be cool if it looked like the delete button on the sprites list from the editor.

BroJac5246:
> We still need to work on Dynamic Disable for this addon. Xs placed over images
> 
> Luckily, 2.0 → 3.0 doesn't break anything. Nice!
> 
> What is slightly concerning is accidentally clicking the X, especially on profile pictures which are already small. Maybe it could be placed _on_ the corner of the thumbnails, like in the Scratch-2.0 studio design, further out of the way. It might also be cool if it looked like the delete button on the sprites list from the editor.

On the thumbnails' corners would probably be the better placement because it's a lot more efficient.

mybearworld:
> Maybe it could be placed _on_ the corner of the thumbnails, like in the Scratch-2.0 studio design, further out of the way.

> On the thumbnails' corners would probably be the better placement because it's a lot more efficient.

It already is on the corner of the thumbnails - your screenshot seems to be an issue with CSS not getting applied for some reason.
image


> It might also be cool if it looked like the delete button on the sprites list from the editor.

That would only work for when `scratchr2` is enabled, otherwise it would feel out of place.

Secret-chest:
What about a setting to only show it on hover?

Secret-chest:
The button should be round and have a proper icon, not an `X`.

mybearworld:
Dynamic disable is fixed now - I accidentally used `() => disable` instead of just `disable`.

> What about a setting to only show it on hover?

Already exists.

> The button should be round and have a proper icon, not an `X`.

I'm using a simple <button> element, so it has the same style as all other buttons, I thought that would make the most sense.

mybearworld:
Confirmation messages could probably use some more detail (e.g. "Unfavorite Sample Project?" instead of "Remove this item from this row?")

addons/remove-from-profile-rows/addon.json
mybearworld: I think when confirmations are disabled, it's expected that you won't get any confirmations.

BroJac5246:
> > The button should be round and have a proper icon, not an `X`.
> 
> I'm using a simple `<button>` element, so it has the same style as all other buttons, I thought that would make the most sense.

For consistency, it might make sense to use what is used in studios, including based on if the studio is 2.0- or 3.0-styled.

mybearworld:
> > > The button should be round and have a proper icon, not an `X`.
> > 
> > 
> > I'm using a simple `<button>` element, so it has the same style as all other buttons, I thought that would make the most sense.
> 
> For consistency, it might make sense to use what is used in studios, including based on if the studio is 2.0- or 3.0-styled.

That would make sense.

BroJac5246:
> > > > The button should be round and have a proper icon, not an `X`.
> > > 
> > > 
> > > I'm using a simple `<button>` element, so it has the same style as all other buttons, I thought that would make the most sense.
> > 
> > For consistency, it might make sense to use what is used in studios, including based on if the studio is 2.0- or 3.0-styled.
> 
> There's not really a remove button on studios.

I mean the buttons to remove a project from a studio while within a studio, sorry.

mybearworld:
> > > > > The button should be round and have a proper icon, not an `X`.
> > > > 
> > > > 
> > > > I'm using a simple `<button>` element, so it has the same style as all other buttons, I thought that would make the most sense.
> > > 
> > > 
> > > For consistency, it might make sense to use what is used in studios, including based on if the studio is 2.0- or 3.0-styled.
> > 
> > 
> > There's not really a remove button on studios.
> 
> I mean the buttons to remove a project from a studio while within a studio, sorry.

Again, there isn't really a button for that - just three dots with an option to remove there. This seems like too much for an addon like this.

BroJac5246:
> > > > > > The button should be round and have a proper icon, not an `X`.
> > > > > 
> > > > > 
> > > > > I'm using a simple `<button>` element, so it has the same style as all other buttons, I thought that would make the most sense.
> > > > 
> > > > 
> > > > For consistency, it might make sense to use what is used in studios, including based on if the studio is 2.0- or 3.0-styled.
> > > 
> > > 
> > > There's not really a remove button on studios.
> > 
> > 
> > I mean the buttons to remove a project from a studio while within a studio, sorry.
> 
> Again, there isn't really a button for that - just three dots with an option to remove there. This seems like too much for an addon like this.

I just realized I haven't actually removed a project from a studio since they got changed to scratch-www 🤦
I was thinking more along the lines of the 2.0 studios, where there was just a circular "x" button in the top right (I believe). If the 3.0-styled pages addon is enabled then we can change the button to be blue to better fit the 3.0 theme.

mybearworld:
> I was thinking more along the lines of the 2.0 studios, where there was just a circular "x" button in the top right (I believe). If the 3.0-styled pages addon is enabled then we can change the button to be blue to better fit the 3.0 theme.

The changing button to blue part is automatically done with the `scratchr2` addon already. It being circular and on the top right would work, but I don't think it's too necessary.

mybearworld:
Error messages are now more detailed:
scratch.mit.edu says: Are you sure you want to leave "Untitled Studio"?
scratch.mit.edu says: Failed leaving "Untitled Studio". This could be because the studio doesn't exist anymore, you own the studio, or you're signed out.

DNin01:
> > Maybe it could be placed _on_ the corner of the thumbnails, like in the Scratch-2.0 studio design, further out of the way.
> 
> 
> 
> It already is on the corner of the thumbnails - your screenshot only happened because CSS isn't getting applied due to the problems with dynamic disable (that should be an easy fix) (it was).

By "_on_ the corner" I meant placed on top of the border, half on top of the image and half outside the border. Right now it's inside the image.

mybearworld:
> > > Maybe it could be placed _on_ the corner of the thumbnails, like in the Scratch-2.0 studio design, further out of the way.
> > 
> > 
> > It already is on the corner of the thumbnails - your screenshot only happened because CSS isn't getting applied due to the problems with dynamic disable (that should be an easy fix) (it was).
> 
> By "_on_ the corner" I meant placed on top of the border, half on top of the image and half outside the border. Right now it's inside the image.

Like this, but somehow with the border not disappearing?
image

BroJac5246:
> > > > Maybe it could be placed _on_ the corner of the thumbnails, like in the Scratch-2.0 studio design, further out of the way.
> > > 
> > > 
> > > It already is on the corner of the thumbnails - your screenshot only happened because CSS isn't getting applied due to the problems with dynamic disable (that should be an easy fix) (it was).
> > 
> > By "_on_ the corner" I meant placed on top of the border, half on top of the image and half outside the border. Right now it's inside the image.
> 
> Like this, but somehow with the border not disappearing?
> image
> 

Much better!

mybearworld:
> > > > > Maybe it could be placed _on_ the corner of the thumbnails, like in the Scratch-2.0 studio design, further out of the way.
> > > > 
> > > > 
> > > > It already is on the corner of the thumbnails - your screenshot only happened because CSS isn't getting applied due to the problems with dynamic disable (that should be an easy fix) (it was).
> > > 
> > > 
> > > By "_on_ the corner" I meant placed on top of the border, half on top of the image and half outside the border. Right now it's inside the image.
> > 
> > 
> > Like this, but somehow with the border not disappearing?
> 
> Much better!

I think so too! But I can't figure out how to have the border not disappear, because it goes beyond the `item`.

mybearworld:
The only way I can think of is having a margin at the top of the items - but this looks weird.
image

DNin01:
> I think so too! But I can't figure out how to have the border not disappear, because it goes beyond the `item`.

This doesn't seem to happen on the dedicated pages, which is one advantage of only supporting dedicated pages, as I suggested earlier.

mybearworld:
> > I think so too! But I can't figure out how to have the border not disappear, because it goes beyond the `item`.
> 
> This doesn't seem to happen on the dedicated pages, which is one advantage of only supporting dedicated pages, as I suggested earlier.

That probably does make sense - it'd have to have a different name, though.

mybearworld:
Updated settings page
This is much better.

mybearworld:
I didn't know renaming the branch closed the issue...

addons/remove-from-profile-rows/addon.json
mybearworld: Done - https://github.com/ScratchAddons/ScratchAddons/pull/5960#issuecomment-1528868196

addons/remove-from-profile-rows/addon.json
mybearworld: Done with https://github.com/ScratchAddons/ScratchAddons/pull/5960/commits/8fc107c0f36b3b74dff9e769cb11ae1d6d900f6c.

PR #5783, opened 2023-03-05
Temporal bugfix for #5697 (message-filter spams requests)

WorldLanguages:
Temporal bug fix for #5697 

Couldn't fully check if this fixes the bug, but it should. @DNin01 Maybe you could confirm?

PR #5983, opened 2023-05-01
message-filters: fix load more spam

### LINKED ISSUE #5697: `message-filters` spams requests
### Description

When `message-filters` is enabled, requests to fetch messages are quickly and repeatedly being sent. They stop after a second, but there are a lot of repetitive requests, which is unnecessary. The addon should only send the same request once and wait until it's done before sending more.

### Steps to reproduce

1. Enable `message-filters`
2. Have your browser's developer tools open and recording network activity
3. Open the messages page
4. Check the network activity. You'll see a lot of duplicate requests.

The network tab in DevTools with a bunch of "messages?limit=40&offset=40 or 80" requests

### Environment

Browser: Edge 109
Scratch Addons version: v1.30.0-pre and v1.30.1
Operating system: Windows 10

mxmou:
Resolves #5697
Closes #5783

### Changes

Fixes incorrect logic in the `message-filters` userscript. The previous code used an infinite loop with `waitForElement`, which caused the button to be clicked for every message until there were at least 40. This PR uses Redux to detect messages being loaded instead.

### Tests

Tested on Edge and Firefox. Verified that every request is sent only once using the network log.

### DIFF
--- a/addons/message-filters/userscript.js
+++ b/addons/message-filters/userscript.js
@@ -121,22 +121,23 @@ export default async function ({ addon, console, msg }) {
     localStorage.setItem("scratchAddonsMessageFiltersSettings", JSON.stringify(active));
     // Log how many messages are showing.
     console.log(`${count} messages showing.`);
-  }
-  // Add the checkboxes element.
-  document.querySelector(".messages-social-title").appendChild(container);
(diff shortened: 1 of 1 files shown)


DNin01 (approved):
Tested on Edge 112.

WorldLanguages:
Thanks!

PR #5989, opened 2023-05-02
Fix popup layout bug in Chrome

### LINKED ISSUE #2593: Weird popup styling issues in Chrome, just sometimes

mxmou:
Resolves #2593

### Changes

Adds a `width` to some `img` elements.

### Reason for changes

Chrome sometimes doesn't update the layout properly when the image loads, which causes weird bugs.

### Tests

I wasn't able to reproduce the issue with this change. It happens quite often on v1.31.2, so I believe that means it's fixed.

### DIFF
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -32,6 +32,7 @@ body {
   padding: 0 20px;
 }
 #logo {
+  width: 30px;
   height: 30px;
   margin-inline-end: 20px;
   vertical-align: middle;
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Good catch! Chrome not knowing the aspect ratio of images until they load would explain all layout issues.

WorldLanguages (approved):
Tested in Chromium

PR #6010, opened 2023-05-05
Gradient bar/nav bar.

fin-github:
Resolves #
N/A

### Changes

At the top it changes it from boring-blue to a blue-to-purple gradient.

### Reason for changes

Because, at the top it is boring. With this, it gives it a new neat touch.

### Tests

Enable it, look at the nav bar.

### ~~WIP~~
~~I am currently making old.css work. main.css works fine.~~

image


DNin01 (changes requested):
Be aware that this will conflict with Editor dark mode and customizable colors, I'm pretty sure.

WorldLanguages:
Note that I've talked with the PR author on Discord and they're creating addons mostly for learning purposes (web technologies and git) as well as personal use.

Secret-chest:
This shouldn't be an addon, having no settings is bad.

WorldLanguages:
I'm closing this for now.

fin-github:
> This shouldn't be an addon, having no settings is bad.

i am still working on the settings part

fin-github:
> Be aware that this will conflict with Editor dark mode and customizable colors, I'm pretty sure. It also has a setting that allows changing the navigation bar color, so we would have to find a way to consider that or maybe not have this proposed addon at all.

I have known that beforehand, 2 reasons why I proceeded:



1. This is a gradient not a solid color
2. This is just light. If you dont want to use editor dark mode then you can use this

BroJac5246:
If something like this is added then I think it should be of either `discuss-button` or website dark theme/customizable colors.

Secret-chest:
> If something like this is added then I think it should be of either `discuss-button` or website dark theme/customizable colors.

Probably `dark-www` since `discuss-button` is getting too many unrelated features.

DNin01:
I would be interested to see what this turns into.

Samq64:
Let's move the discussion to #6022.

PR #6013, opened 2023-05-06
Add "Cooldown display for posting in the forums" addon

### LINKED ISSUE #5984: Show how many seconds left on 60 second rule
### Why this would be helpful

I want to know how many seconds before I can post again

### How this addon works

When the error is displayed, it also shows how many seconds left
maybe a live timer too? (possibly)

### Possible alternatives

putting a timer on your phone after posting is not a good idea

### Additional context

_No response_

mybearworld:
Resolves #5984 

### Changes

Added a "Cooldown display for posting in the forums" (internally `forum-post-countdown`) addon which keeps track of when you can post again.

The addon in the settings.
The countdown in the post button.

### Reason for changes

It makes it easier to know when exactly you can post again. It also makes time feel a lot faster if you can see it pass in real time.

### Tests

Tested in Microsoft Edge 114.0.1807.6 (Official build) dev (64-bit).


WorldLanguages:
I would suggest to simply forget New Scratchers exist. Most users of this addon will be full Scratchers. Users shouldn't expect a "post" button with no countdown to mean with 100% certainty that the second rule has expired.

addons/forum-post-countdown/addon.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "Countdown for posting in the forums",
+  "description": "When posting in the forums, show a countdown on every page showing when you can post again.",
```
DNin01, line 3: Here's my way of wording the name and description which you can apply and/or modify if you want:
```suggestion
  "name": "Display cooldown for posting in the forums",
  "description": "After submitting a forum post, display a countdown showing when you can post again.",
```

BroJac5246:
> I would suggest to simply forget New Scratchers exist. Most users of this addon will be full Scratchers. Users shouldn't expect a "post" button with no countdown to mean with 100% certainty that the "seconds rule" has expired.

I would disagree. Users will (rightly) interpret the countdown as "time until you can post," not "time until you can potentially post." Also, do we actually know what percent of our users are New Scratchers? It wouldn't surprise me if we do have New Scratchers who use the addon; some people just never become Scratchers.

mybearworld:
> > I would suggest to simply forget New Scratchers exist. Most users of this addon will be full Scratchers. Users shouldn't expect a "post" button with no countdown to mean with 100% certainty that the "seconds rule" has expired.
> 
> I would disagree. Users will (rightly) interpret the countdown as "time until you can post," not "time until you can potentially post." Also, do we actually know what percent of our users are New Scratchers? It wouldn't surprise me if we do have New Scratchers who use the addon; some people just never become Scratchers.

And SA users who've switched accounts as well.

addons/forum-post-countdown/addon.json
mybearworld: > Use nouns instead of verbs (for example, `Customizable new sprite position` instead of `Change new sprite position`) unless doing so would result in the addon name being overly complicated. (https://scratchaddons.com/docs/reference/addon-manifest/)

Because of that, I think cooldown and display should be flipped (so that display is the noun):

```suggestion
  "name": "Cooldown display for posting in the forums",
  "description": "After submitting a forum post, display a countdown showing when you can post again.",
```

Other than that, this seems like a better way to phrase it (I am not good with making names, as you can tell)

Secret-chest:
Can't you fetch the Scratcher status from the API and make it automatic?

WorldLanguages:
> Can't you fetch the Scratcher status from the API and make it automatic?

Yeah, I'd say we fetch `/session` on each page load (within this addon's userscript) to check whether the account is New Scratcher or not.

mybearworld:
> > Can't you fetch the Scratcher status from the API and make it automatic?
> 
> Yeah, I'd say we fetch `/session` on each page load (within this addon's userscript) to check whether the account is New Scratcher or not.

That might have edge cases (Do we know what happens if you post, then become a Scratcher? Do you still have to wait 120s until you can post or is it reduced to 60s?) But those are rare enough that I think we can ignore them,

WorldLanguages:
There's infinite edge cases, like logging in into a different account in a different browser tab.

mybearworld:
> There's infinite edge cases, like logging in into a different account in a different browser tab.

Logging out just deletes the entry in localStorage and thereby removes the countdown, because I didn't want to deal with making that work :P

RedGuy12:
> Users shouldn't expect a "post" button with no countdown to mean with 100% certainty that the "seconds rule" has expired.

regardless of if we consider new scratchers, this should be noted in settings

mybearworld:
> > Users shouldn't expect a "post" button with no countdown to mean with 100% certainty that the "seconds rule" has expired.
> 
> regardless of if we consider new scratchers, this should be noted in settings

The addon now automatically adjusts based on the status, I don't think that needs to be mentioned.

RedGuy12:
as you said yourself, there's infinite edge cases - other tabs/browsers/devices, switching accounts, etc, etc. we don't want lots of bug reports for stuff we won't fix, so we should have a note saying it doesn't have 100% guaranteed accuracy but is good enough for the majority of use cases

mybearworld:
> as you said yourself, there's infinite edge cases - other tabs/browsers/devices, switching accounts, etc, etc. we don't want lots of bug reports for stuff we won't fix, so we should have a note saying it doesn't have 100% guaranteed accuracy but is good enough for the majority of use cases

That makes sense. I've added a note.

Samq64:
> as you said yourself, there's infinite edge cases - other tabs/browsers/devices, switching accounts, etc, etc. we don't want lots of bug reports for stuff we won't fix, so we should have a note saying it doesn't have 100% guaranteed accuracy but is good enough for the majority of use cases

The same can be said with pretty much all addons (and software in general)

RedGuy12:
i feel like users of this addon are more likely to encounter those edge cases, but in the end i don't really care that much tbh

mybearworld:
> i feel like users of this addon are more likely to encounter those edge cases, but in the end i don't really care that much tbh

Now that I think about it, it's really only when changing devices - and I think someone changing devices will probably take at least minute to do so.

addons/forum-post-countdown/addon.json
```diff
@@ -0,0 +1,30 @@
...
+  "versionAdded": "1.33.0",
+  "userscripts": [
+    {
+      "url": "post-create.js",
+      "matches": ["newPostScreens"]
+    },
+    {
+      "url": "userscript.js",
+      "matches": ["https://scratch.mit.edu/*"]
+    }
```
WorldLanguages, line 14: Why does this need to run on every page?

mybearworld: So the cooldown is shown on every page.

WorldLanguages: Is that really necessary?

WorldLanguages: I was thinking just post buttons in the forums?

mybearworld: That probably does make sense.

WorldLanguages:
Does the cooldown affect changing the signature or editing posts? Just asking so we can test against those cases.

mybearworld:
> Does the cooldown affect changing the signature or editing posts? Just asking so we can test against those cases. (And surely this shouldn't affect the "report a post" screen)

It does not affect any of these.

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+
+  setInterval(async () => {
+    const now = Date.now();
+    if (now > Number(countdown) + secondCount * 1000 || !(await addon.auth.fetchIsLoggedIn())) {
+      localStorage.removeItem("sa-forum-post-countdown");
+      elt.remove();
+      return;
+    }
+    elt.textContent = msg("seconds-left", { seconds: secondCount - Math.floor((now - countdown) / 1000) });
+  });
```
WorldLanguages, line 25: Doing this once a second is probably okay? I believe setInterval with no second argument runs multiple times per second.

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+
+  addon.settings.addEventListener("change", () => {
+    secondCount = addon.settings.get("120-second") ? 120 : 60;
+  });
```
WorldLanguages, line 29:
```suggestion
```

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+
+  addon.self.addEventListener("disabled", () => {
+    elt.style.display = "none";
+  });
+  addon.self.addEventListener("reenabled", () => {
+    elt.style.display = "";
+  });
+}
```
WorldLanguages, line 37:
```suggestion
  addon.tab.displayNoneWhileDisabled(elt);
```

addons/forum-post-countdown/post-create.js
```diff
@@ -0,0 +1,10 @@
+export default async function () {
+  const submitButtons = document.querySelectorAll("#djangobbwrap .form-submit [type=submit]");
+  const submitButton = submitButtons[submitButtons.length - 1];
```
WorldLanguages, line 3: Can a single page have more than one submit button? I'm curious because I think other addons don't consider this

WorldLanguages:
We could use `sessionStorage` instead of `localStorage` here.

addons/forum-post-countdown/post-create.js
mybearworld: No, they can't - this is from when this addon duplicated the post button, and I wanted to make sure nothing weird could happen.

```suggestion
  const submitButton = document.querySelector("#djangobbwrap .form-submit [type=submit]");
```

addons/forum-post-countdown/userscript.js
mybearworld: Then it could be delayed by about a second - might not be too big of a deal, though.

WorldLanguages: @mybearworld At least do 250 or 500 ms, the default is like 3ms which is excessive

addons/forum-post-countdown/addon.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "Cooldown display for posting in the forums",
+  "description": "After submitting a forum post, display a countdown showing when you can post again.",
+  "tags": ["community", "forums"],
+  "versionAdded": "1.33.0",
+  "userscripts": [
+    {
+      "url": "post-create.js",
```
Samq64, line 8: Since this runs on the same pages as the main script could it be merged with it?

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,28 @@
...
+    ? 120
+    : 60;
+
+  const elt = document.createElement("div");
+  elt.id = "sa-forum-post-countdown";
+  document.body.appendChild(elt);
+
+  setInterval(async () => {
+    const now = Date.now();
+    if (now > Number(countdown) + secondCount * 1000 || !(await addon.auth.fetchIsLoggedIn())) {
```
Samq64, line 19: Could the login check only be run once?

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,28 @@
...
+      localStorage.removeItem("sa-forum-post-countdown");
+      elt.remove();
+      return;
+    }
+    elt.textContent = msg("seconds-left", { seconds: secondCount - Math.floor((now - countdown) / 1000) });
+  });
+
+
+  addon.tab.displayNoneWhileDisabled(elt);
```
Samq64, line 28: You forgot the closing brace at the end of the file.

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,28 @@
...
+  document.body.appendChild(elt);
+
+  setInterval(async () => {
+    const now = Date.now();
+    if (now > Number(countdown) + secondCount * 1000 || !(await addon.auth.fetchIsLoggedIn())) {
+      localStorage.removeItem("sa-forum-post-countdown");
+      elt.remove();
+      return;
+    }
+    elt.textContent = msg("seconds-left", { seconds: secondCount - Math.floor((now - countdown) / 1000) });
```
Samq64, line 24: This would probably be a little easier to understand if it was a variable.

mybearworld:
image
what

WorldLanguages:
> We could use `sessionStorage` instead of `localStorage` here.

Is it okay if I change this?

mybearworld:
> > We could use `sessionStorage` instead of `localStorage` here.
> 
> Is it okay if I change this?

Go ahead,

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,40 @@
...
+    await fetch("/session", { headers: { "x-requested-with": "XMLHttpRequest" } }).then((resp) => resp.json())
+  ).permissions.new_scratcher
+    ? 120
+    : 60;
+
+  const elt = document.createElement("div");
+  elt.id = "sa-forum-post-countdown";
+  document.body.appendChild(elt);
+
+  const seconds = secondCount - Math.floor((now - countdown) / 100);
```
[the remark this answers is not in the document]
mybearworld: I am really not thinking here... I'm keeping it the way it is, probably doesn't need to be ava riable.

Samq64:
Where is countdown displayed?

mybearworld:
> Where is countdown displayed?

On every page with a post button on the bottom right on the page.

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,38 @@
+export default async function ({ addon, msg }) {
+  const countdown = localStorage.getItem("sa-forum-post-countdown");
+  if (!countdown) {
+    return;
```
Samq64, line 4: I was wrong about my review because now this script will never run since the storage item is set later.

mybearworld: Should now be fixed (among with a bunch of other things)

WorldLanguages:
> We could use `sessionStorage` instead of `localStorage` here.

I was wrong:

> Whenever a document is loaded in a particular tab in the browser, a unique page session gets created and assigned to that particular tab. That page session is valid only for that particular tab.

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+    ? 120
+    : 60;
+
+  const elt = document.createElement("div");
+  elt.id = "sa-forum-post-countdown";
+  document.body.appendChild(elt);
+
+  setInterval(async () => {
+    const now = Date.now();
+    if (now > Number(countdown) + secondCount * 1000 || !loggedIn) {
```
WorldLanguages, line 28: `loggedIn` isn't going to change with time, so it doesn't make sense to use its value on the interval.

addons/forum-post-countdown/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+    if (!localStorage.getItem("sa-forum-post-countdown")) {
+      localStorage.setItem("sa-forum-post-countdown", Date.now());
+    }
+  });
+
+  const countdown = localStorage.getItem("sa-forum-post-countdown");
+  if (!countdown) {
+    return;
+  }
+
```
WorldLanguages, line 14: Ideally, at this point of the code we would check if the countdown is already overdue (for example, it might have been 24 hours), delete the storage item and return in that case.

WorldLanguages:
@mxmou You might have better ideas for the positioning and CSS of the countdown (but we can deal with this after submitting strings for translation)

The current design is sticky and is positioned in the bottom right of topics and the "new post" screens:
image

addons/forum-post-countdown/userscript.js
WorldLanguages: I think that would avoid the countdown from showing for 1 second then disappearing. Haven't actually tried to test or emulate this case, though

Samq64:
I expected it to to be vertically aligned with the post button.

mybearworld:
Maybe inside the post button, like [Post (in 23s)]?

addons/forum-post-countdown/userscript.js
mybearworld: Good catch

mybearworld:
image

mybearworld:
I also used the opportunity to disable the post button.

addons/forum-post-countdown/userscript.js
mybearworld: With the countdown now being in the post button, is this actually that important? Worst case, there's a bit too much padding on the submit button for a second.

WorldLanguages (approved):
I'm not sure if we should disable the post button, but it looks okay

mybearworld:
> I'm not sure if we should disable the post button, but it looks okay

I thought it makes sense because you can't post anyways - maybe it could just have lower opacity but still allow you to use it as usual?

WorldLanguages:
> I thought it makes sense because you can't post anyways - maybe it could just have lower opacity but still allow you to use it as usual?

Yeah, I think that would be better.

mybearworld:
> Yeah, I think that would be better.

Alright, done.

BroJac5246:
> I'm not sure if we should disable the post button, but it looks okay

Why? Is there any reason not to disable it? There's no reason you would want to use it anyway.

mybearworld:
> > I'm not sure if we should disable the post button, but it looks okay
> 
> Why? Is there any reason not to disable it? There's no reason you would want to use it anyway.

Probably just not blocking functionality that Scratch allows, or making sure it's still usable if the addon somehow breaks.

PR #6019, opened 2023-05-07
Add "Separate "Copy Link" and "Copy Embed" buttons on project page" addon

### LINKED ISSUE #5518: Seperate link copy and embed copy
### Why this would be helpful

I hate it when I have to click the Copy Link button and then be given a popup where I have to select a link that copies the Scratch project URL/HTML.

### How this addon works

This addon makes the "Copy Link" button copy the link straight away, and adds a "Copy Embed" button that copies an HTML embed code like the one in the vanilla Scratch dialog.

### Possible alternatives

_No response_

### Additional context

_No response_

mybearworld:
Resolves #5518

### Changes

Added "Separate "Copy Link" and "Copy Embed" buttons on project page" (internally `copy-link-embed`) addon, which separates the copy link and copy embed functionality.

The addon in the settings pageThe addon in use.

### Reason for changes

In order to streamline the process of copying links/embeds which also works fine on small screens.

### Tests

Tested on Microsoft Edge 114.0.1807.6 (Official build) dev (64-bit).


fractioneater:
I would recommend following the naming scheme of most other addons, where the title is what changes when it is activated. My suggestion is "Separate buttons for link and embed copy."

mybearworld:
> I would recommend following the naming scheme of most other addons, where the title is what changes when it is activated. My suggestion is "Separate buttons for link and embed copy."

That is a lot better.

addons/copy-link-embed/addon.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "Separate buttons for link and embed copy.",
```
CST1229, line 2: I suggest
```suggestion
  "name": "Separate \"Copy Link\" and \"Copy Embed\" buttons on project page",
```

Samq64:
I think there should be a setting to remove the copy embed button since most people probably won't use that.

mybearworld:
> I think there should be a setting to remove the copy embed button since most people probably won't use that.

There already is, and you can also hide the Copy Link button.

PR #6020, opened 2023-05-07
Add "Arrow key incrementation in editor" addon

### LINKED ISSUE #4928: Change numbers in block inputs with arrow keys
Sometimes I want to increment/decrement a number in a block input by small numbers. To speed things up, I'd kind of like to see an addon that lets you use `↑`/`↓` while they're selected to increment/decrement them. This exists in Chrome's DevTools:

Increment by 1: `↑`
Increment by 10: `Shift`+`↑`
Increment by 100: `Ctrl`+`↑`
Increment by 0.1: `Alt`+`↑`

Just an idea.

mybearworld:
Resolves #4928

### Changes

Added addon called "Arrow key incrementation in editor" (internally `editor-number-arrow-keys`), that allows incrementing/decrementing values through arrow keys.

(this screenshot is outdated)
The addon in the settings page.

### Reason for changes

Make it more easy to quickly increment or decrement a value.

### Tests

Tested on Microsoft Edge 114.0.1807.6 (Official build) dev (64-bit)


mxmou:
> Draft because of JavaScript floating point math - trying to increase ".2" by .1 will give you 0.30000000000000004.

You can multiply the number by 10, add 1, then divide by 10 to get an accurate result.

mybearworld:
> > Draft because of JavaScript floating point math - trying to increase ".2" by .1 will give you 0.30000000000000004.
> 
> You can multiply the number by 10, add 1, then divide by 10 to get an accurate result.

Thanks, that worked!

WorldLanguages:
> You can multiply the number by 10, add 1, then divide by 10 to get an accurate result.

Won't that break if the number has two decimal places?

mybearworld:
> > You can multiply the number by 10, add 1, then divide by 10 to get an accurate result.
> 
> Won't that break if the number has two decimal places?

It does, but this'd work: `(Number(elt.value) * (10**amountOfDecimalPlaces) + 1) / (10**amountOfDecimalPlaces)`

mybearworld:
For now, this'll do.

addons/editor-number-arrow-keys/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Arrow key shortcuts for number input in editor",
+  "description": "Use the up and down arrow keys to increase or decrease numbers. Holding Shift increments it by 10, holding Ctrl increments it by 100 and holding Alt increments it by 1000.",
```
DNin01, line 3: My suggested name and description:
```suggestion
  "name": "Arrow key incrementation in editor",
  "description": "Use the up and down arrow keys to increase or decrease number values in block inputs. Holding Shift changes by 10, holding Ctrl changes by 100 and holding Alt changes by 1,000.",
```

DNin01 (approved):
Tested on Edge 112.

mybearworld:
It could probably be customizable now when floats are not used.

DNin01:
> Huh? Why does the `positive_integer` type require a `row` to be present?

Maybe try using `integer` and setting minimum and maximum values instead? Perhaps > 1 and < 1,000,000?

(I just did some searching and it seems like `positive_integer` is only still an option for backwards compatibility. Now, we can use `integer` and just set the minimum value to 0 to achieve the same functionality.)

mybearworld:
It still wants a `row`...

apple502j:
That error shows up if the setting definition violates the schema file in any way. (It is set to fall back to table when everything fails.)

Integers should not be stringified.

mybearworld:
Shouldn't it just fail the check if it can't figure out what the `type` is, though?

apple502j:
Ajv looks for the pair of all keys and values, and does not prioritize one over the others. An integer with a string default value and an "animal" type are equally invalid; it does not check the type then the conformance of other properties in that order, but instead checks everything at once.

DNin01:
> Integers should not be stringified.

Oh, no wonder.

The minimums should probably be 1, though - when set to 0, the keyboard shortcut won't do anything.

WorldLanguages:
Do we really need to provide four different increase levels? That feels like a lot (at least as a default)

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -0,0 +1,31 @@
+export default async function ({ addon }) {
+  setInterval(() => {
+    [...document.querySelectorAll(".blocklyHtmlInput:not(.sa-editor-number-arrow-keys)")].forEach((elt) => {
+      elt.addEventListener("keydown", (e) => {
```
WorldLanguages, line 4: No need to use `setInterval`, which won't be nice for performance, and no other addons use it this way.

Simply register (once!) a single `keydown` event on `document.body` and check if `e.target.classList.has("blocklyHtmlInput")`.
This also removes the need for a `sa-editor-number-arrow-keys` class.

WorldLanguages:
Let's also not forget [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) exists, to avoid the 0.3000004 bug

WorldLanguages:
@mybearworld Can you allow contributors to push into your branch?

mybearworld:
> @mybearworld Can you allow contributors to push into your branch? (in all of your PRs)

Done!

WorldLanguages:
> Do we really need to provide four different increase levels? That feels like a lot (at least as a default)

Bump!
Also, do these need to be number inputs? Wouldn't a select with a few options be better? We could also offer a `0.1` option which might be quite handy for wait blocks.

mybearworld:
> Wouldn't a select with a few options be better?

That's a good idea.

> e could also offer a `0.1` option which might be quite handy for wait blocks.

That was the original plan but failed because of the 0.1 + 0.2 issue - I'll try BigInt.

mxmou:
> Let's also not forget [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) exists, to avoid the 0.3000004 bug

How would BigInt fix that? Isn't it an integer type?

WorldLanguages:
> That was the original plan but failed because of the 0.1 + 0.2 issue - I'll try BigInt.

BigInt only supports integers, so we still need to multiply to 10^n and `n` will be the supported amount of "digits(?)". BigInt will let us support up to 10 digits without issue I believe (not sure about normal numbers). If the block already contains a number with more than a certain amount of digits then we should not try to do math operations on it. The idea is to support most use cases but not the extreme ones, so that we don't have to deal with the bugs.

WorldLanguages:
> How would BigInt fix that? Isn't it an integer type?

Didn't mean to imply BigInt would save us from converting decimal numbers into an integer. But we get higher precision when doing operation with the possibly big integers

mybearworld:
It now uses a select with the options `None`, `1`, `10`, `100`, `1,000`.

DNin01:
> It now uses a select with the options `None`, `1`, `10`, `100`, `1,000`. Do these work?

It’s reminiscent of #6352. I think only being able to choose powers of 10 works okay, and it’s definitely less cumbersome than typing out the numbers yourself. Though, some people may want more control over this.

WorldLanguages:
> Do we really need to provide four different increase levels? That feels like a lot (at least as a default)

Bump! I think it's a very bad idea to provide four different increase levels. Maybe we can consider it for future versions of the addon.

mybearworld:
> Hm, what would other defaults be?

WorldLanguages:
IMO we should provide the following settings:

- Normal keypress: `0.01`, `0.1`, `1` (default), `10`, `100`
- Some other key: same options but either `0.1` or `10` as default (we should discuss which one is more useful)

Samq64:
Maybe the amount a value is changed by can depend on how large the number is? The problem is trying to make that work in a way users would expect.

mybearworld:
> Maybe the amount a value is changed by can depend on how large the number is? The problem is trying to make that work in a way users would expect.

I don't think there would be a way to do that a majority of users would agree on.

Secret-chest:
Maybe it could be done based on the block?

mybearworld:
> Maybe it could be done based on the block?

Is there a block where it only makes sense to put in numbers divisible by a certain number (that's not one)?

Secret-chest:
> > Maybe it could be done based on the block?
> 
> Is there a block where it only makes sense to put in numbers divisible by a certain number (that's not one)?

No, but for example in the movement blocks it should be 1, whereas in the wait block it should be 0.1.

Samq64:
What if it's 1 for whole numbers and 0.1 for decimals?

WorldLanguages:
> Maybe it could be done based on the block?

I was about to post about providing a different logic for the "wait for X seconds" block, but is it really worth it?

mybearworld:
> What if it's 1 for whole numbers and 0.1 for decimals?

Maybe it could adjust based on the level of precision in the block? e.g. 1 -> 2, 1.1 -> 1.2, 1.11 -> 1.12, etc
Still, it would have to be actually implemented well

WorldLanguages:
I have asked griffpatch for his opinion, shouldn't take too long...

Samq64:
> > What if it's 1 for whole numbers and 0.1 for decimals?
> 
> Maybe it could adjust based on the level of precision in the block? e.g. 1 -> 2, 1.1 -> 1.2, 1.11 -> 1.12, etc Still, it would have to be actually implemented well

That's what I was thinking, but what if the user presses the up arrow to go from 9 to 10? Would it jump to 20 after that?

BroJac5246:
> > > What if it's 1 for whole numbers and 0.1 for decimals?
> > 
> > Maybe it could adjust based on the level of precision in the block? e.g. 1 -> 2, 1.1 -> 1.2, 1.11 -> 1.12, etc Still, it would have to be actually implemented well
> 
> That's another idea I had, but what if the user presses the up arrow to go from 9 to 10? Would it jump to 20 after that?

I'm thinking up/down should be increments of one and shift+up/down should be increments of powers of 10, as suggested above. The two could also potentially be swapped, but if you're making large increments then you should just type in the number, so increments of one make the most sense for fine-tuning.

mybearworld:
> That's another idea I had, but what if the user presses the up arrow to go from 9 to 10? Would it jump to 20 after that?
> 
> Edit: Are you saying it would only do that for decimals and not whole numbers?

I don't think it would make much sense to do this for whole numbers - but the same problem can apply with decimals (e.g. 1.7 -> 1.8 -> 1.9 -> 2 -> 3)

WorldLanguages:
> I have asked griffpatch for his opinion, shouldn't take too long...

He said 0.1 would be useful, not much more

mybearworld:
Ok, so it generally seems like that in some way, decimals should definitely exist.

WorldLanguages:
Been trying to play with decimals. It's quite hard without manually shifting the decimal point by messing with the number as a string of characters.

WorldLanguages:
brave_S7qvRzhJfP

WorldLanguages:
Defaults are now the same as Chrome DevTools: (see original suggestion: https://github.com/ScratchAddons/ScratchAddons/issues/4928#issue-1323303406)

image

WorldLanguages:
We could offer a `0.01` setting too, but using `100` would probably be rare, so I'm not sure if we should provide a `100` option

DNin01:
> image

This is quite visually pleasing.

WorldLanguages:
@mybearworld Could you test my changes?
Please also read the TODOs I left in the code

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+    return Number(numStr) * (numberIsNegative ? -1 : 1);
+  };
+
+  document.body.addEventListener("keydown", (e) => {
+    if (!e.target.classList.contains("blocklyHtmlInput")) return;
+    if (!["ArrowUp", "ArrowDown"].includes(e.code)) return;
+    if (Number(e.target.value).toString().replace(/^0*/, "") !== e.target.value.replace(/^0*/, "")) return;
+    // TODO: why remove leading zeros to a result of Number().toString ?
```
mybearworld, line 35: Because `e.target.value` might be `.1`, where `Number(e.target.value).toString()` would be "0.1", and therefore different from `e.target.value` even though they should be counted as the same.

```suggestion
```

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -0,0 +1,66 @@
...
+  document.body.addEventListener("keydown", (e) => {
+    if (!e.target.classList.contains("blocklyHtmlInput")) return;
+    if (!["ArrowUp", "ArrowDown"].includes(e.code)) return;
+    if (
+      e.target.value &&
+      Number(e.target.value)
+        .toString()
+        .replace(/^0*|0*$/, "") !== e.target.value.replace(/^0*|0*$/, "")
+    )
+      return;
```
WorldLanguages, line 40: This will not return if `e.target.value` is null, so it will fail on line 42 still

mybearworld: 1. When would `e.target.value` be null?
2. Doesn't the simple `e.target.value` check take care of that?

mybearworld: Oh, wait, yeah, true, I didn't think this through

```suggestion
    if (
      !e.target.value ||
      Number(e.target.value)
        .toString()
        .replace(/^0*|0*$/, "") !== e.target.value.replace(/^0*|0*$/, "")
    )
      return;
```

WorldLanguages: > 1. When would `e.target.value` be null?

Well, I meant falsey. It might also be an empty string

mybearworld: An empty string was what I was originally trying to prevent, but I messed up the logic.

WorldLanguages (approved):
Everything looks okay

WorldLanguages:
@DNin01 Could you test everything again on Edge?

DNin01 (commented):
Tested in Edge 115, no issues found.

Still would like to see dynamic enable/disable.

DNin01 (approved):

WorldLanguages:
> Still would like to see dynamic enable/disable.

Whoops, forgot. It was quite easy. Thanks @mybearworld!

DNin01:
> That was the original plan but failed because of the 0.1 + 0.2 issue - I'll try BigInt.

Today, I've been trying to work on a BigInt solution as a better alternative to what we came up with already. Once I finish it (if I do), I'll open another PR to replace the current ["crappy decimal support"](https://github.com/ScratchAddons/ScratchAddons/pull/6020/commits/9867844ff320ad45f83d3dd078117877c64f1743).

PR #4157, opened 2022-02-01
More consistent icons in `material-forum-editor-buttons`

Mr-Anonymous75:
## Changes

I made all the icons more constistent, and also removed unnecessary borders from all of them.

dark-www on
dark-www off

## Motivation

The icons just seemed kinda inconsistent.

## Tests

I've tested on Chrome 97 and Firefox Dev Edition 97 on Windows 11 21H2, with `dark-www` both on and off.

---

<sup>I usually spend hours cleaning up the SVG code, but I was tired enough by the time I got there - you won't believe how tough link and scratchblocks were.</sup>

PR #6047, opened 2023-05-14
Completely revamp `material-forum-editor-buttons`

just-a-hriday:
So I was going through my old issues and PRs, and I found #4157 - a PR trying to make `material-forum-editor-buttons`'s icons more consistent.

I decided to retry doing that, and I ended up completely revamping `material-forum-editor-buttons`.
<sub>Although in the end, I don't think I did anything at all for consistency</sub>

## Changes

- The addon now uses Google's variable material symbol font instead of adding an SVG for each button.
- The buttons are aligned with the separators now.
- The addon has a `userscript.js` file.

## Tests

Tested on Edge 113 (chromium) and Firefox 114.

Compatible with `scratchr2`, `dark-www`, `forum-toolbar`, and `image-uploader`.

## Comparison

scratch mit edu_discuss_topic_684116_ <sup>Old addon. </sup>

scratch mit edu_discuss_topic_684116_ <sup>Vanilla scratch.</sup>

scratch mit edu_discuss_topic_684116_ <sup>New addon, outlined.</sup>

scratch mit edu_discuss_topic_684116_ <sup>New addon, filled.</sup>

<sub>Sorry for the switched italic and underline buttons in the screenshots, it's fixed now</sub>

---

## Issues <sub><sup><sub>There are lots</sub></sup></sub>

- [x] The emoji icons get deleted from the dropdown menu.
- [x] The scratchblocks options have question marks inside it.
- [x] Incompatible with `forum-toolbar` and `image-uploader`.
- [x] The background images (containing the bitmap expand arrows) in the dropdown buttons aren't getting removed while `dark-www` is on.
- [x] The buttons take a lot of time to appear.
- [x] Some text (the name of the glyph) is visible for about a second before the buttons appear.
- [x] `scratchblocks` (forums 3.0 scratchblocks addon) removes the expand buttons from the scratchblocks dropdown menu for some reason.
- [ ] I couldn't find suitable icons for **scratch wiki, <strike>wikipedia</strike>, and <strike>dictionary.com</strike>** from `forum-toolbar`.
**Please suggest icons.**
- [ ] No icon for scratchblocks button - google doesn't have one. I've submitted an issue in their github repo, though. If they want to they'll add it in a month or two.<br><sup>hey that rhymed</sup>


just-a-hriday:
I feel like the list item and os version icons don't properly convey the purpose of the button.

Does anyone else know a better icon from [Google's website](https://fonts.google.com/icons)?

just-a-hriday:
I tried adding this at the start of `userscript.js` to fix the incompatibilities.

```
await addon.tab.waitForElement(".sa-forum-toolbar-code");
await addon.tab.waitForElement(".markItUpButton17");
```

The addon works now, but only when both `forum-toolbar` and `image-uploader` are enabled, _and_ `forum-toolbar` has code enabled.

Otherwise, it just gets stuck at the await statements because either one of the elements doesn't appear.

Secret-chest:
I don't think you should use a font, and you should still use classic material icons.

Secret-chest:
Even better than Material Symbols, there is <http://icon-sets.iconify.design/mdi/>.

just-a-hriday:
> and you should still use classic material icons.

Uhh, they _are_ material icons. What do you mean?

> I don't think you should use a font

Why not? It's much easier to maintain once it's implemented, it's much easier to add more buttons, and it's easy to update in case google changes any icons.

Secret-chest:
They aren't material _icons_, they are material _symbols_, at least update the description.

Secret-chest:
> Why not? It's much easier to maintain once it's implemented, it's much easier to add more buttons, and it's easy to update in case google changes any icons.

The font is going to have all the symbols when you only need around ten of them and it's going to take a while to load

just-a-hriday:
> The font is going to have all the symbols when you only need around ten of them

Well, yeah, I guess.

> and it's going to take a while to load

It does, yeah.

**BUT** my point still stands

mybearworld:
Instead of `computer`, `devices` might be a better fit for the browser/os button.

Secret-chest:
> **BUT** my point still stands

How?

just-a-hriday:
> This should use the svgs. Iconfonts are lowerres and you can't pick and choose which you want.

Lower res? These are vectors. There is no resolution. And these are more consistent.

<sub><sub>is it fine if i just ignore the rest of the comment for now-</sub</sub>

just-a-hriday:
> Instead of `computer`, `devices` might be a better fit for the browser/os button.

Huh, good idea. I'll do that with the next push.

mybearworld:
> And these are more consistent.

You can download Material Symbols as well.

GrahamSH-LLK:
> > This should use the svgs. Iconfonts are lowerres and you can't pick and choose which you want.
> 
> Lower res? These are vectors. There is no resolution. And these are more consistent. is it fine if i just ignore the rest of the comment for now-

You should be able to get SVGs for each of these. Instead, we're shipping a 2.45mb file that includes tons of icons we don't need.

just-a-hriday:
> > > This should use the svgs. Iconfonts are lowerres and you can't pick and choose which you want.
> > 
> > Lower res? These are vectors. There is no resolution. And these are more consistent. is it fine if i just ignore the rest of the comment for now-
> 
> You should be able to get SVGs for each of these. Instead, we're shipping a 2.45mb file that includes tons of icons we don't need. Keep in mind that SA is only ~70mb.

Oh wait, I didn't even think about the file size-

I'll do something about that problem.

addons/material-forum-editor-buttons/userstyle.css
```diff
@@ -0,0 +1,131 @@
+@import url("https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200");
```
WorldLanguages, line 1: I'm not sure if Chrome Web Store policy allows this

WorldLanguages: Any "choose a custom font" implementation would do something like this, however. It's kind of a gray area.

just-a-hriday: Oh. Any better solutions?

Earlier, I'd added the woff2 file directly, but @GrahamSH-LLK suggested I change it.
And well, he's right - the font file is huge!

WorldLanguages: I guess the best solution would be an SVG file for each icon, included in the source of the extension.

mybearworld:
I'd suggest `library_books` for the dictionary. Wikipedia, the Scratch Wiki and Scratchblocks are not services that are "official", so there aren't really icons - maybe `public`, `pets` and `code_blocks`?

Secret-chest:
> I'd suggest `library_books` for the dictionary. Wikipedia, the Scratch Wiki and Scratchblocks are not services that are "official", so there aren't really icons - maybe `public`, `pets` and `code_blocks`?

They could have some custom ones.

just-a-hriday:
> > I'd suggest `library_books` for the dictionary. Wikipedia, the Scratch Wiki and Scratchblocks are not services that are "official", so there aren't really icons - maybe `public`, `pets` and `code_blocks`?
> 
> They could have some custom ones.

`library_books` is good for dictionary.com, and `public` is alright for Wikipedia.

But what would `pets` be used for? It's just a paw.

And `code_blocks` is the same as the `code` icon, it's just surrounded by a square. Also, scratch's blocks are based on google's Blockly library, and I've posted an issue in google's material icon repo suggesting that they add an icon for a Blockly block. They might make one in a few months.

addons/material-forum-editor-buttons/userstyle.css
just-a-hriday: > Any "choose a custom font" implementation would do something like this, however. It's kind of a gray area.

Shouldn't google be fine with it? It's their own icon font from their own api.

BroJac5246:
> I'd suggest `library_books` for the dictionary. Wikipedia, the Scratch Wiki and Scratchblocks are not services that are "official", so there aren't really icons - maybe `public`, `pets` and `code_blocks`?

I think `language` would be better for Wikipedia. But why `pets` for the Scratch Wiki?

mybearworld:
> > I'd suggest `library_books` for the dictionary. Wikipedia, the Scratch Wiki and Scratchblocks are not services that are "official", so there aren't really icons - maybe `public`, `pets` and `code_blocks`?
> 
> I think `language` would be better for Wikipedia. But why `pets` for the Scratch Wiki?

It was the one that made most sense to me (as it's like a cat). It doesn't make much sense, though.

BroJac5246:
> > > I'd suggest `library_books` for the dictionary. Wikipedia, the Scratch Wiki and Scratchblocks are not services that are "official", so there aren't really icons - maybe `public`, `pets` and `code_blocks`?
> > 
> > I think `language` would be better for Wikipedia. But why `pets` for the Scratch Wiki?
> 
> It was the one that made most sense to me (as it's like a cat). It doesn't make much sense, though.

I don't think we should use it because most people aren't going to make the connection.

Secret-chest:
Also, Material Symbols are sometimes much uglier than Material Icons, or at least lost the personality for the sake of "simplification".

Left is Material Symbols, right is Material Icons.

imageimage

imageimage

just-a-hriday:
> Also, Material Symbols are sometimes much uglier than Material Icons, or at least lost the personality for the sake of "simplification".
> 
> Left is Material Symbols, right is Material Icons.
> 
> imageimage
> 
> imageimage

I prefer the material symbols in both cases, assuming they're the ones on the left.

Secret-chest:
I forgot the home icon. The worst one.

just-a-hriday:
> I forgot the home icon. The worst one.

I admit that the home symbol is a bit oversimplified, but the cottage symbol is a reasonable replacement.

Issue #5190, opened 2022-10-12
Scaling SA to 150+ addons: performance and memory usage


WorldLanguages:
We're getting close to the 150+ addon count.
This issue intends to list the technical side of handling this amount of addons. The user experience (UX) side of having so many addons belongs in a different issue.

### Memory usage (background page)
It's important to consider the memory usage of our background page. It's true that Scratch itself is pretty memory intensive, but SA is running all the time, even if the user isn't currently using Scratch!
Since v1.0.0, all manifests are stored inside a variable in the background page.
This includes all this data from all addons, no matter if they're enabled or not:
  - Addon ID
  - The entire addon.json file as an Object
  - If the browser language is not set to English, and it's a language we support (such as Spanish), the English name and description are stored in memory too (see https://github.com/ScratchAddons/ScratchAddons/blob/master/background/load-addon-manifests.js#L58)

This `scratchAddons.manifests` array gets larger with every release. I'm not sure how to properly check the actual memory usage of this "big" object, but according to Chrome Devtools, it appears to use about 0.25MB.

We could also check whether `scratchAddons.l10n` stores any unnecessary messages in memory in the background page.

### chrome.storage.sync per-key size limit

Related: #4721

As we add more settings each release, we'll soon reach 8192 bytes for the `addonSettings` key.
See QUOTA_BYTES_PER_ITEM on https://developer.chrome.com/docs/extensions/reference/storage/
It was previously easy to reach this limit by exporting/importing settings, but we fixed it. However, the problem will return at some point, if we keep adding new settings.

### Settings page performance

This is less important than the other things I mentioned above, but it belongs on this list.
We're using Vue v1 and I've implemented some weird performance hacks to try to avoid performance bottlenecks.
However, of course the settings page has more DOM nodes each release, as well as more JavaScript (vdom?) objects handled by Vue.
However, the solution here is more straight forward, we could just migrate to Vue 3, and we'll get better performance with less hacks. We would also avoid shipping the CSP version which is slower.

### Anything else we need to consider?

Please tell me whether I'm missing something here.

iqnite:
Performance isn't the only problem: just think about the SSD storage it will need when it will have 250 addons with all the GUI icons that belong to them. Actually, it's not relevant, but like in the boiled frogs principle, Scratch Addons will grow and grow until it will be too big.

iqnite:
[...]

PR #6049, opened 2023-05-14
Vue 3 for webpages

### LINKED ISSUE #2531: Settings page takes way too long to load

GrahamSH-LLK:
Factor in  #5190
Resolves #2531 (in my testing)

### Changes

This is a large change.
This change rewrites the settings page, license page, and popup in Vue 3, compiled with vite. 
### Reason for changes

Loads nearly instantly
We aren't using _vue 1_
Single js/css file
### Tests

Tested on chrome and firefox


DNin01:
It would be nice to see some screenshots.

Samq64:
> It would be nice to see some screenshots.

As far as I know this PR just ports the settings page to Vue 3 and doesn't change anything visually.

DNin01:
> This PR just ports the settings page to Vue 3 and doesn't change the design. I haven't tested it though and there might be some small differences.

Okay!

webpagesish/src/assets/logo.svg
WorldLanguages, line 1: lol

WorldLanguages:
Note that if this gets merged, quite a lot of code can be removed because they were Vue v1 optimizations.

WorldLanguages:
> If this updates the design of the settings page, it would be nice to see some screenshots.

This PR should ideally keep everything as-is unless it's code that wouldn't work on Vue v3. Unnecessary code should also stay. We can remove it later.

webpagesish/.gitignore
```diff
@@ -0,0 +1,28 @@
...
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
+
+node_modules
+.DS_Store
+dist
```
WorldLanguages, line 12: We wouldn't gitignore `dist`, right?

webpagesish/data/addon-groups.js
```diff
@@ -0,0 +1,99 @@
+export default [
+  // Only popup:
+  {
+    id: "recentlyUsed",
+    name: () => {
+      chrome.i18n.getMessage("recentlyUsed");
+    },
```
TheColaber, line 7: probably a mistake, there is no return here

DNin01 (commented):
Does this new settings page require a build step in order for a developer to use it? If so, we should update any installation instructions such as in the top-level README. We could also add this info to CONTRIBUTING.md.

webpagesish/data/addon-groups.js
GrahamSH-LLK: that is in fact a mistake

webpagesish/.gitignore
GrahamSH-LLK: True

GrahamSH-LLK:
> Does this new settings page require a build step in order for a developer to use it? If so, we should update any installation instructions such as in the top-level README. We could also add this info to CONTRIBUTING.md.

Yes, it does (but only if they're editing it)

GrahamSH-LLK:
TODO: 
- [ ] Fix tables- the inputs aren't showing up bc they aren't defined. @TheColaber I think implemented this originally?
- [ ] Fix sidebar- its too low for some reason
- [ ] Implement unsupported page - just add the js file??

webpages/settings/README.md
```diff
@@ -0,0 +1,29 @@
+# webpagesish
```
DNin01, line 1:
```suggestion
# webpages
```

webpagesish/.gitignore
Hans5958: > We wouldn't gitignore `dist`, right?

I was wondering the reason on this. A self-reloading build script would save space in this case. Is this a compromise to avoid build steps?

GrahamSH-LLK: Yep, this way the settings page works for people without installing node, `npm install`ing and running a build script. The bundle is pretty small anyways

Hans5958: That's too bad. I wish we can have a full-fledged build script. At least this is one step of it.

.prettierignore
```diff
@@ -8,3 +8,5 @@ _locales/*/messages.json
 !_locales/en/messages.json
 
 .github/*
+
+webpages/settings/dist/*
```
RedGuy12, line 12: eslint ignore too

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
```
RedGuy12, line 2: why another package.json?

webpagesish/.gitignore
WorldLanguages: > At least this is one step of it.

Yeah, this would still be a step forwards. I don't think "saving space" is really a concern here.

WorldLanguages: Also, there's nothing directly wrong with checking-in outputs. The Chrome DevTools team [even checks in node_modules](https://www.jackfranklin.co.uk/blog/check-in-your-node-dependencies/).

GrahamSH-LLK:
Okay this like.. works. Perfectly. The only thing I have left is reimplementing the licenses page (which should be trivial AF)

CST1229:
Maybe building webpagesish could be made into a workflow.

GrahamSH-LLK:
Some testing: 
- Resources load 200ms faster than current.
- JS blocking time is down from 212ms to 67ms (3x faster loads)
- Overall, it loads 345ms faster (on my relatively fast computer, larger differences would be seen on slower computers)

DNin01 (commented):
Still need to address some things in the More Settings popup (Vue 3 is the second one):
Normal
Vue 3 webpage

GrahamSH-LLK:
> Still need to address some things in the More Settings popup (Vue 3 is the second one): Normal Vue 3 webpage

strange, I'll take a look

webpages/README.md
```diff
@@ -0,0 +1,29 @@
+# Settings Page
+
+This directory is built, but is has already been compiled. You may recompile it only if you make changes.
```
mybearworld, line 3: I think the is is not supposed to be there
```suggestion
This directory is built, but has already been compiled. You may recompile it only if you make changes.
```

GrahamSH-LLK:
Superseded by #6272

PR #6050, opened 2023-05-15
Add "Skip post button" addon

### LINKED ISSUE #6046: Skip post button in forums
### Why this would be helpful

Some topics contain very large posts that you don't always need to read, and therefore need to skip by scrolling a lot.

### How this addon works

This addon would:
1. Fix the post header
2. Add a button to the post header that skips to the next post (or the reply section, if it's the last post on a page)

### Possible alternatives

_No response_

### Additional context

_No response_

mybearworld:
Resolves #6046

### Changes

Added "Skip post button" `skip-post` addon that adds a button to skip the post to every forum post.

### Reason for changes

Sometimes there are very long posts you don't always want to read.

### Tests

Tested in Microsoft Edge 115.0.1843.0 (Official build) dev (64-bit) 

### To do

- [ ] **Making the "Fix post header" option actually work.** Just `position: sticky; top: 0;` isn't working, neither is also removing the overflow from its parent's `.box`.


mybearworld:
Shouldn't this have https://github.com/ScratchAddons/ScratchAddons/labels/scope%3A%20addon as well?

mybearworld:
> Also I didn't know GitHub let you put labels directly in comments like that.

It's too bad these types of things aren't documented in one place.

PR #6051, opened 2023-05-15
Fix grammar

1enss:
I fixed some grammar (just a tad)

### Changes
I just fixed some grammar problems.

### Reason for changes

For easy readability
### Tests
Tested in readme.


GrahamSH-LLK:
keep WebExtension, and it prolly doesn't need `that is easy to use`

Samq64 (approved):
Note there's a bigger draft readme PR open: #5603

WorldLanguages (changes requested):
I would move some of these changes to #5603 instead.

PR #6053, opened 2023-05-15
Run addons on `scratchfoundation.github.io` for testing

WorldLanguages:
### Changes

Runs addons on https://scratchfoundation.github.io/scratch-gui/beta/ and https://scratchfoundation.github.io/scratch-gui/develop/

Related: #5281

### Reason for changes

Makes it easier to test the upcoming changes.
This will be reverted before v1.33.0 release.

### Tests

Most things work normally in Chromium and Firefox. This includes dynamic enable, dynamic disable and dynamic setting changes.
There's some problems with `addon.tab.scratchClass` but a reload often fixes it. If, for example, the find bar lacks CSS, a reload will often fix it. It's not a bug that will happen this often in real scratch.mit.edu.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -207,7 +207,7 @@ export default class Tab extends Listenable {
    * @type {?string}
    */
   get editorMode() {
-    if (location.origin === "https://scratchfoundation.github.io") return "editor";
+    if (location.origin === "https://scratchfoundation.github.io" || location.port === "8601") return "editor";
     const pathname = location.pathname.toLowerCase();
     const split = pathname.split("/").filter(Boolean);
     if (!split[0] || split[0] !== "projects") return null;
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -207,6 +207,7 @@ export default class Tab extends Listenable {
    * @type {?string}
    */
   get editorMode() {
+    if (location.origin === "https://scratchfoundation.github.io") return "editor";
     const pathname = location.pathname.toLowerCase();
     const split = pathname.split("/").filter(Boolean);
     if (!split[0] || split[0] !== "projects") return null;
@@ -307,6 +308,12 @@ export default class Tab extends Listenable {
     res = res.slice(0, -1);
     // Sanitize just in case
     res = res.replace(/"/g, "");
+
+    if (res === "" && location.origin === "https://scratchfoundation.github.io" && !window._scratchClassAlert) {
+      window._scratchClassAlert = true;
+      alert("addon.tab.scratchClass might have failed in this page load. Please reload.");
+    }
+
     return res;
   }
 
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -308,12 +308,6 @@ export default class Tab extends Listenable {
     res = res.slice(0, -1);
     // Sanitize just in case
     res = res.replace(/"/g, "");
-
-    if (res === "" && location.origin === "https://scratchfoundation.github.io" && !window._scratchClassAlert) {
-      window._scratchClassAlert = true;
-      alert("addon.tab.scratchClass might have failed in this page load. Please reload.");
-    }
-
     return res;
   }
 
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -223,7 +223,8 @@ function onDataReady() {
 }
 
 function bodyIsEditorClassCheck() {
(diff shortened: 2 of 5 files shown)


WorldLanguages:
@mxmou This can be extended to a specific localhost port too in case that helps too.

Samq64:
> This will be reverted before v1.33.0 release.

Why? is it because of extension permissions?

WorldLanguages:
> Why? is it because of extension permissions?

Correct. See https://developer.chrome.com/docs/extensions/mv3/permission_warnings/#update_permissions

mxmou:
> @mxmou This can be extended to a specific localhost port too in case that helps too.

That would be very useful because building locally is the only way to test scratch-www. The default port is 8333 for scratch-www and 8601 for scratch-gui.

> > Why? is it because of extension permissions?
> 
> Correct. See https://developer.chrome.com/docs/extensions/mv3/permission_warnings/#update_permissions

Some permissions are already removed from the published version of the extension depending on the browser. Could the same be done with this host permission?

WorldLanguages:
> That would be very useful because building locally is the only way to test scratch-www. The default port is 8333 for scratch-www and 8601 for scratch-gui.

Okay, I will add support for those two. Could you remind me what a scratch-gui URL in localhost might look like?

> Some permissions are already removed from the published version of the extension depending on the browser. Could the same be done with this host permission?

Yes, we will consider that later.

mxmou:
> > That would be very useful because building locally is the only way to test scratch-www. The default port is 8333 for scratch-www and 8601 for scratch-gui.
> 
> Okay, I will add support for those two. Could you remind me what a scratch-gui URL in localhost might look like?

Usually just `http://localhost:8601/`. There are also `localhost:8601/blocks-only.html` and `localhost:8601/player.html`, but those aren't very useful for SA development.

DNin01 (changes requested):
As soon as I reloaded Scratch Addons, the site access permissions page appeared and I needed to give Scratch Addons permission to access scratchfoundation.github.io. That could be a potential user-experience issue - users might be surprised or even suspicious.

So, potentially, this could be an optional permission? It's not like the average user will need Scratch Addons to run on scratchfoundation. And besides, one of the best practices is to not use required permissions for things that aren't absolutely necessary for an extension to run. Developers that needed to would simply either allow Scratch Addons to access all sites or all necessary sites plus scratchfoundation.

<sub>Also, the "Scratch Addons has updated to 1.***33***.0" message appeared.</sub>

rgantzos:
> That could be a potential user-experience issue - users might be surprised or even suspicious.

That's why WL said it would be removed before release, that way the new permissions don't reach the normal users.

BroJac5246:
> As soon as I reloaded Scratch Addons, the site access permissions page appeared and I needed to give Scratch Addons permission to access scratchfoundation.github.io. That could be a potential user-experience issue - users might be surprised or even suspicious.

World_Languages wrote:

> This will be reverted before v1.33.0 release.

And responded to this question by Samq64:

> > Why? is it because of extension permissions?
> 
> Correct. See https://developer.chrome.com/docs/extensions/mv3/permission_warnings/#update_permissions

Secret-chest:
(can't quote DNin01)

I'd really like this being an option in "More Settings" so we can also help report bugs.

WorldLanguages:
@mxmou Forgot if there was anything pending to add. Could you check if localhost is properly supported? You still need a different extension to redirect `localhost:8333/session` to a data URI, however. And userstyles may not load as fast, which only really matters for the "website dark mode" addon.

content-scripts/inject/module.js
```diff
@@ -91,6 +91,7 @@ const page = {
   },
   isFetching: false,
   async refetchSession() {
+    if (location.origin === "https://scratchfoundation.github.io") return;
```
mxmou, line 94: This function doesn't work when running scratch-www locally because the request is blocked by CORS. It should probably fetch `localhost:8333/session` instead of `scratch.mit.edu/session` when on `localhost:8333`.

Everything else works as expected.

CST1229: It could fetch `/session` instead of having a set domain.

WorldLanguages:
Anything else?

mxmou (approved):

addon-api/content-script/Tab.js
```diff
@@ -207,6 +207,7 @@ export default class Tab extends Listenable {
    * @type {?string}
    */
   get editorMode() {
+    if (location.origin === "https://scratchfoundation.github.io" || location.port === "8601") return "editor";
```
WorldLanguages, line 210: From https://github.com/ScratchAddons/ScratchAddons/pull/6253#issuecomment-1603850164

This should have considered fullscreen mode.

PR #6056, opened 2023-05-15
Added that the cat blocks were from 2020-2023

jabinstech:
I changed the cat-blocks json file to state that the cat blocks were from 2020-2023, not just 2020

i do not know how to make a pull request

### DIFF
--- a/addons/cat-blocks/addon.json
+++ b/addons/cat-blocks/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Cat blocks",
-  "description": "Brings back editor cat hat blocks from April Fools 2020-2023.",
+  "description": "Brings back editor cat hat blocks that were first introduced for April Fools' Day 2020.",
   "credits": [
     {
       "name": "TheColaber",
--- a/addons/cat-blocks/addon.json
+++ b/addons/cat-blocks/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Cat blocks",
-  "description": "Brings back editor cat hat blocks from April Fools 2020.",
+  "description": "Brings back editor cat hat blocks from April Fools 2020-2023.",
   "credits": [
     {
       "name": "TheColaber",


GrahamSH-LLK:
You just made one! Nice job, if this is your first time!

Samq64 (approved):

DNin01 (commented):
Is it possible to say something along the lines of "from April Fools 2020 and after" (not exactly that) or just leave out the year entirely?

BroJac5246:
> Is it possible to say something along the lines of "from April Fools 2020 and after" (not exactly that) or just leave out the year entirely?

I've been thinking about that as well. If we accept this then we will need to update the description every year. Additionally, do we consider Cat blocks and Caturday mode to be the same thing? Previous discussions have shown that we haven't reached a consensus.

jabinstech:
> > Is it possible to say something along the lines of "from April Fools 2020 and after" (not exactly that) or just leave out the year entirely?
> 
> I've been thinking about that as well. If we accept this then we will need to update the description every year. Additionally, do we consider Cat blocks and Caturday mode to be the same thing? Previous discussions have shown that we haven't reached a consensus.

Ah, yeah, that makes sense.
<sup><sub>what do i do now do i close it or something or do i change it and make it say caturday mode...</sub></sup>

addons/cat-blocks/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Cat blocks",
-  "description": "Brings back editor cat hat blocks from April Fools 2020.",
+  "description": "Brings back editor cat hat blocks from April Fools 2020-2023.",
```
BroJac5246, line 3:
```suggestion
  "description": "Brings back editor cat hat blocks that were first introduced for April Fools' Day 2020.",
```
This mitigates the issue I mentioned because it refers only to 2020 but still acknowledges that future versions do exist. While at it I, I also improved the name of April Fools' Day, but you don't have to accept that it you prefer the previous name.

jabinstech:
That probably makes more sense.

BroJac5246 (approved):

DNin01 (approved):
That sounds reasonable.

PR #6060, opened 2023-05-16
Add hiding vanilla buttons to forum-toolbar

### LINKED ISSUE #6059: Add vanilla buttons to forum-toolbar settings
### Why this would be helpful

A lot of people probably don't use some buttons of the forum toolbar. They may want to get rid of them.

### My suggestion

Add options to `forums-toolbar` to remove the vanilla buttons as well.

### Possible alternatives

We could create a separate addon instead, but this would make more sense.

### Additional context

_No response_

mybearworld:
Resolves #6059

### Changes

Added the ability to customize the vanilla buttons to `forum-toolbar`.

### Reason for changes

A lot of people probably don't use some buttons of the forum toolbar. They may want to get rid of them.

### Tests

Tested in Microsoft Edge 115.0.1843.0 (Official build) dev (64-bit) 

### Other

I've made this too quickly, when I should've waited on replies on the issue - but there's no point holding the PR back now.


mybearworld:
Is there any better way to indicate Scratch Addons' settings than adding a wrench? It gets a bit confusing without it. (This is how it would look without the New tag)
The settings

Samq64:
Now that there's a screenshot, I see why World Languages doesn't want it. I think grouping Scratch Addons options together would help a little though.

WorldLanguages:
> Now that there's a screenshot, I see why World Languages doesn't want it.

Are you referring to the fact this adds fifteen new settings to hide buttons added by vanilla Scratch?

mybearworld:
> > Now that there's a screenshot, I see why World Languages doesn't want it.
> 
> Are you referring to the fact this adds fifteen new settings to hide buttons added by vanilla Scratch?

What else could they be referring to? That's the only thing you can see in the screenshot.

mybearworld:
This would greatly benefit from #5719 as well.

just-a-hriday:
Can you add a new screenshot with the grouped options?

PR #6061, opened 2023-05-16
vol-slider: compatibility with RTL and editor-buttons-reverse-order

### LINKED ISSUE #6055: Auto-hiding volume slider incompatible with reverse order of project controls

mxmou:
Resolves #6055

### Changes

* RTL support
* Compatibility with `editor-buttons-reverse-order`. I changed the order of elements as suggested in https://github.com/ScratchAddons/ScratchAddons/issues/6055#issuecomment-1550191655.
* Changed the CSS to use nested elements with `position: absolute` instead of animating the margin. This isn't absolutely necessary but I prefer this solution.

### Reason for changes

To fix bugs.

### Tests

Tested different combinations of settings in both LTR and RTL.


DNin01 (approved):
Nice work! Tested all combinations I could think of.

There was a small animation (see at https://github.com/ScratchAddons/ScratchAddons/pull/5797#issuecomment-1516801676) where the slider also "moves" slightly that was sacrificed in this change, but I don't care, it makes the code simpler.

Secret-chest:
Can we re-add the animation?

WorldLanguages:
Is RTL language + "reverse order of project controls" supposed to look like this when volume is at 100%?
I guess it makes sense but it's confusing for me. Just making sure.

image

WorldLanguages:
Here (LTR with "reverse project controls") I would have expected a different order, but it's ok.
I think `clone counter - mouse position - volume slider` would be the best order (this order is used on RTL languages with "reverse project controls"), but it's probably hard to do.

image

WorldLanguages:
I'm merging this as-is for the v1.32.1 release. Any bugs that might persist will be lower-prio than the current volume slider bugs from v1.32.0

Secret-chest:
Still no animation?

mxmou:
> Is RTL language + "reverse order of project controls" supposed to look like this when volume is at 100%? I guess it makes sense but it's confusing for me. Just making sure.

I believe reversing the direction of the slider in RTL is correct. The layout would make more sense if the slider was on the left side of the icon, but in that case it would cover the green flag and stop sign.

The icon should probably be mirrored in RTL though.

> Here (LTR with "reverse project controls") I would have expected a different order, but it's ok. I think `clone counter - mouse position - volume slider` would be the best order (this order is used on RTL languages without "reverse project controls"), but it's probably hard to do.

I like this suggestion - it's not hard to do and it would prevent the clone counter, which doesn't have a fixed width, from causing the mouse position to move.

> Still no animation?

I added it back after @DNin01 and you commented about it.

WorldLanguages:
> I like this suggestion - it's not hard to do and it would prevent the clone counter, which doesn't have a fixed width, from causing the mouse position to move.

👍 If interested, open a new PR

PR #6063, opened 2023-05-17
Fix unable to insert sprites from backpack

### LINKED ISSUE #6062: Faster project loading prevents inserting sprites from backpack
idk what addon did it bc i tried every one that shouldve affected it but only way to fix it is by turning off addons pls help i cant put stuff from my backpack in projects

DNin01:
Resolves #6062

### Changes

`faster-project-loading` blocks requests to insert sprites from the backpack. To my understanding, these requests always look like this: `https://backpack.scratch.mit.edu/*.zip`. Because they begin with backpack.scratch.mit.edu, this addon blocks them, but that was not the intended behavior. This change should fix that but not affect project load speed.

### Tests

Tested in Edge 113.

### DIFF
--- a/addons/faster-project-loading/userscript.js
+++ b/addons/faster-project-loading/userscript.js
@@ -2,10 +2,12 @@
 
 export default async function ({ addon }) {
   const BACKPACK_URL = "https://backpack.scratch.mit.edu/";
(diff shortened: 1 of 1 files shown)


WorldLanguages (approved):
Actually this is OK

Tacodiva:
Thanks for fixing this :smile:

PR #6070, opened 2023-05-18
Change string `gamepad/store-in-project`

### LINKED ISSUE #5458: `gamepad`: Clarify what storing settings in the project means
### Why this would be helpful

There are a few things the "Store these settings in the project to override the default configuration (Experimental tool for project creators)" checkbox could mean. Maybe it's to have SA remember these settings. Or maybe to save the settings somewhere inside the project (this is the answer).

"Experimental tool for project creators" does give us a hint, but it's still generally vague.

Even then, this switch is useless for anyone but the project creator (or people downloading the project to play it in TurboWarp or remixing it). Also, the creator has to enter the editor and save the project for the configurations to apply to others, I'm assuming.

And for anyone else, they won't be the owner of the project in order to save their edits (unless they make a remix), so they can't save the settings, but the way it's worded makes it sound like it might get saved in SA, if you're not reading carefully. These users may see it **and might misunderstand and think that SA would remember the controls they configured** when it actually doesn't - this actually has to be inserted into the project by its creator.

### My suggestion

We could do one of a few things:

1. Reword it to better describe the function. Perhaps "Store these settings **inside** the project"?
2. Add an extended description like "the creator of a project can use this setting to store the controls above inside the project so that other Scratch Addons users playing it will have the same controls".
3. Only make it visible to the project creator.
4. If you aren't the creator, have SA remember the configuration too.

### Possible alternatives

_No response_

### Additional context

_No response_

DNin01:
Resolves #5458

### Changes

Changed the description of the checkbox for storing the gamepad configuration inside the project.

### Reason for changes

For clarification.

### Tests

Tested.

### DIFF
--- a/addons-l10n/en/gamepad.json
+++ b/addons-l10n/en/gamepad.json
@@ -18,7 +18,7 @@
   "gamepad/keyinput-title": "Click and press a key or click to change button. Escape to cancel. Backspace or delete to clear.",
   "gamepad/browser-support": "This browser and operating system have known bugs that may make this addon difficult to use. Try another browser if you encounter problems.",
   "gamepad/config-header": "This comment contains configuration for gamepad support in third-party tools or websites such as https://turbowarp.org/\nDo not edit by hand",
-  "gamepad/store-in-project": "Store these settings in the project to override the default configuration (Experimental tool for project creators)",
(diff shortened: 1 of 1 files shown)


DNin01:
It might work even better if the checkbox was only visible on your own projects.

GarboMuffin (approved):

PR #6074, opened 2023-05-19
Change default editor highlight color to purple

DNin01:
### Changes

Scratch's editor accent color is updating to `#855cd6`. This change updates some of the presets to use this new color. Presets with a blue accent color are still available.

### Tests

Tested in Edge 113 on scratch.mit.edu and scratchfoundation.github.io.


Samq64:
I don't think we need a 3.dark purple preset.

addons/editor-dark-mode/addon.json
```diff
@@ -1162,6 +1188,32 @@
...
+        "categoryMenu": "#ffffff",
+        "palette": "#f9f9f9cc",
+        "fullscreen": "#ffffff",
+        "stageHeader": "#e8edf1",
+        "border": "#00000026",
+        "dots": true
+      }
+    },
+    {
+      "name": "Scratch's default colors (blue)",
```
Samq64, line 1214: They won't be default after the change.
```suggestion
      "name": "Scratch 3.0 (blue)",
```
Also I think it should be moved before Scratch 2 so the versions are in order, but it doesn't matter.

addons/editor-dark-mode/addon.json
```diff
@@ -1162,6 +1188,32 @@
...
+      "id": "scratch",
+      "description": "The colors originally used by Scratch 3.0",
```
Samq64, line 1216:
```suggestion
      "description": "The colors originally used by Scratch 3.0 until June 2023",
```

DNin01:
> I don't think we need a 3.dark purple preset.

What about new users - those who will have never used Scratch before the editor update? This is the first accent color they would've known about. It would make the most sense to them if presets were based off the purple color. 3.Dark isn't based off Scratch 2.0 colors because they're not the current default.

DNin01:
Adding `status: pending` because we'll need migration code.

addons/editor-dark-mode/addon.json
DNin01: I don't think that piece of information is important, especially in the future.

Samq64:
I think we can remove the 

> > I don't think we need a 3.dark purple preset.
> 
> What about new users - those who will have never used Scratch before the editor update? This is the first accent color they would've known about. It would make the most sense to them if presets were based off the purple color. 3.Dark isn't based off Scratch 2.0 colors because they're not the current default.

I misunderstood. I thought you added a second 3.Dark preset, not 3.Dark**er**. I think all the changes are fine except I don't think we need a second 3.Darker preset. Users can easily choose the blue from the colour dropdown.

WorldLanguages:
I personally believe it's reasonable to add a preset for the old defaults (`3.Darker blue`)

WorldLanguages:
This increases the number of presets to 10 so we may reconsider the order and naming of all the presets, I think.

This is how I would order and name them:
- `3.Darker` (default)
- `3.Darker (2022)`
- `3.Dark`
- `Dark Editor`
- `"Experimental" Dark`
- `TurboWarp Dark`
- `Scratch 1.x`
- `Scratch 2.0`
- `Scratch 3.0 (2022)`
- `Scratch's default colors`

WorldLanguages:
> Adding `status: pending` because we'll need migration code.

What migration logic are we going to use? If "highlight color" is the old blue, change it to the default purple. Is that all?

Samq64:
I think "blue" makes more sense than "2022".

addons/editor-dark-mode/addon.json
```diff
@@ -1162,6 +1188,32 @@
...
+        "palette": "#f9f9f9cc",
+        "fullscreen": "#ffffff",
+        "stageHeader": "#e8edf1",
+        "border": "#00000026",
+        "dots": true
+      }
+    },
+    {
+      "name": "Scratch's default colors (blue)",
+      "id": "scratch",
```
Samq64, line 1215: Two presets can't have the same id.

background/get-addon-settings.js
```diff
@@ -296,6 +296,19 @@ chrome.storage.sync.get(["addonSettings", "addonsEnabled"], ({ addonSettings = {
             settings.popup = newPopupSettingValue;
             madeAnyChanges = madeChangesToAddon = true;
           }
+
+          updatePresetIfMatching(
+            settings,
+            7,
+            {
+              // Old highlight color (blue)
+              primary: "#4d97ff",
```
Samq64, line 305: `highlightText` also needs to be changed.

Samq64: And `menuBar` either only on Scratch's default colours or always if we're also migrating Experimental Dark.

Secret-chest:
Why `Scratch 3.0 (2022)`? `Scratch 3.0 (2019)` or `Scratch 3.0 (original)` makes more sense.

addons/editor-dark-mode/addon.json
```diff
@@ -1059,14 +1085,14 @@
...
-        "menuBar": "#4d97ff",
+        "page": "#2f2641",
+        "primary": "#9966ff",
+        "highlightText": "#ad7eff",
+        "menuBar": "#855cd6",
         "popup": "#000a",
         "activeTab": "#282828",
         "tab": "#202020",
-        "selector": "#252c37",
+        "selector": "#2b2537",
```
mxmou, line 1095: I understand why you changed the `page` and `selector` colors - the dark blue doesn't look very good together with the purple accents. But the equivalent colors in vanilla Scratch will remain light blue, so I think it's better to choose one of the folllowing two options:

* Less saturated blue
  * page: `#2e3238`
  * selector: `#292d32`

* Dark gray
  * page: `#202020`
  * activeTab: `#333333`
  * tab: `#282828`
  * selector: `#333333`
  * selector2: `#333333`

See the comments above for `primary` and `highlightText`.

addons/editor-dark-mode/addon.json
```diff
@@ -1162,6 +1188,32 @@
       "name": "Scratch's default colors",
       "id": "scratch",
       "description": "The colors normally used by Scratch",
+      "values": {
+        "page": "#e5f0ff",
+        "primary": "#855cd6",
```
mxmou, line 1193: The definition of the `primary-variant` CSS variable will need to be changed to correctly generate a darker version of this new color. Specifically, `r`, `g` and `b` in the `multiply` part should all be `0.85`.

mxmou (changes requested):

background/get-addon-settings.js
WorldLanguages: The migration code I added is probably alright but only as a fallback if none of the presets that changed match exactly. We can discuss it after the other part of the PR is approved by mxmou.

WorldLanguages: > `highlightText` also needs to be changed.

Right. That default has changed too.

WorldLanguages:
I assume the setting migration steps might look something like this, please correct me:
(each item is a different migration and they run top to bottom - note that migrations can only see the settings as set by the migration above it)

1. If current settings look like* Experimental Dark, change the following settings to their new preset defaults: `page`, `primary`, `highlightText`, `menuBar`, `selector`. Other settings are left intact.
2. If current settings look like* the "Scratch's default colors" preset (as it was before this PR) the `menuBar` setting will be changed to its new preset default.
3. If setting `primary` matches the old default, set it to the new default.
4. If setting `highlightText` matches the old default, set it to the new default.

\*checks all settings except `fullscreen`, `stageHeader`, `dots`, `affectPaper`, `affectStage`.

Feel free to edit this comment and ping me if you've added more migration steps or changed the existing ones.

Samq64:
That looks good except I'm not sure if all experimental dark users should be migrated since there's no easy way to undo the changes, unlike the other presets.

WorldLanguages:
> That looks good except I'm not sure if all experimental dark users should be migrated since there's no easy way to undo the changes, unlike the other presets.

Well said 🤔

WorldLanguages:
Regarding migration, we have two alternatives:

- Force the update to the new purple as soon as possible. This might be a bit confusing for users, as we'll need to tell them that some color changes might reflect in their browser before Scratch actually rolls them out to the public.
- Apply the migration only after Scratch rolls out the update. This can be done in many ways (including pings to a server, but we won't do that), each way has it's trade-off, usually user experience vs. complexity of the code and probability of it not working as expected.

It might be better if I send a separate pull request for the migration code to both dark-www and editor-dark-mode, but I was wondering what you think.

WorldLanguages:
@mxmou Can you review this or are you busy at the moment? Please keep us informed as there's only a few days left.

DNin01:
Can we auto-detect when Scratch rolls this out by looking at webpage elements and apply the changes after the fact?

apple502j:
No, settings are modified by the background who knows nothing about the website rendering.

I think the best way is to hold this PR pending the actual changes, then release our update **after** Scratch does.

WorldLanguages:
> I think the best way is to hold this PR pending the actual changes, then release our update **after** Scratch does.

Yeah, that's also an option. That would mean Scratch Addons users will get the new colors late, not early.

DNin01:
> we could just delay the Web Store releases

Upload it to the store early and wait for it to be reviewed so we can release it immediately as needed. That's a great idea!

mxmou:
> I assume the setting migration steps might look something like this, please correct me: (each item is a different migration and they run top to bottom - note that migrations can only see the settings as set by the migration above it)
> 
> 1. If current settings look like* Experimental Dark, change the following settings to their new preset defaults: `page`, `primary`, `highlightText`, `menuBar`, `selector`. Other settings are left intact.
> 2. If current settings look like* the "Scratch's default colors" preset (as it was before this PR) the `menuBar` setting will be changed to its new preset default.
> 3. If setting `primary` matches the old default, set it to the new default.
> 4. If setting `highlightText` matches the old default, set it to the new default.
> 
> *checks all settings except `fullscreen`, `stageHeader`, `dots`, `affectPaper`, `affectStage`.

Some presets (3.Dark, Dark Editor) still use the old values for `primary` and `highlightText`, so updating them unconditionally won't work. I think it's best to keep the user's previous settings unless they match one of the presets.

WorldLanguages:
> Some presets (3.Dark, Dark Editor) still use the old values for `primary` and `highlightText`, so updating them unconditionally won't work. I think it's best to keep the user's previous settings unless they match one of the presets.

I see. I'll update my comment

WorldLanguages:
I think I got confused back then. I thought it would be a good idea to update the `menuBar` ("menu bar background") setting unconditionally to purple, as long as it matched the old blue color (#4D97FF), even if some other settings don't look like "Scratch's default colors".

mxmou:
> I think I got confused back then. I think it would be a good idea to update the `menuBar` ("menu bar background") setting unconditionally to purple, as long as it matched the old blue color (#4D97FF), even if some other settings don't look like "Scratch's default colors". Does this make sense?

Yes, that would work too.

WorldLanguages:
Feel free to directly edit my comment if you wish. https://github.com/ScratchAddons/ScratchAddons/pull/6074#issuecomment-1556031309
I'll update it as soon as I have some time with the idea I just said.

WorldLanguages:
So I think this would be the settings migration then(?)

1. If current settings look like* Experimental Dark, change the following settings to their new preset defaults: `page`, `primary`, `highlightText`, `menuBar`, `selector`. Other settings are left intact.
2. If current settings look like* 3.Darker, change the following settings to their new preset defaults: `primary`, `highlightText`. Other settings are left intact.
3. If current settings look like* Scratch 3.0 default colors, change the following settings to their new preset defaults: `primary`, `highlightText`. Other settings are left intact. Setting `menuBar` is changed on the next step.
4. If the current value of `menuBar` is the old blue default, the `menuBar` setting will be changed to its new preset default. This may or may not complement migration 3.

\*checks all settings except `fullscreen`, `stageHeader`, `dots`, `affectPaper`, `affectStage`.

Samq64:
I think `primary` and `highlightText` should also be updated unconditionally. That means experimental dark would be the only present that would need special migration.

WorldLanguages:
@Samq64 See comment https://github.com/ScratchAddons/ScratchAddons/pull/6074#issuecomment-1565648852

WorldLanguages:
Ideally we should merge this and release v1.33.2 a few days before July 28th.

WorldLanguages:
I implemented the settings migration as described on this comment: https://github.com/ScratchAddons/ScratchAddons/pull/6074#issuecomment-1596295264

WorldLanguages:
As a reminder, these are the current names (and order) of the presets on this PR branch:

image

WorldLanguages:
Adding a latestUpdate with a temporaryNotice to this addon (editor-dark-mode) should be discussed at #6257.

Samq64:
Like I said in https://github.com/ScratchAddons/ScratchAddons/pull/6074#discussion_r1198425200 I don't think the word defeat should be used in the 3.0 blue preset name because it's not the default any more.

WorldLanguages:
> Like I said in [#6074 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/6074#discussion_r1198425200) I don't think the word default should be used in the 3.0 blue preset name because it's not the default any more.

Good point, but let's postpone this discussion for v1.34.0 unless your comment gets a huge amount of upvotes.

Secret-chest:
What about `Scratch 3.0 (original)`?

addons/editor-dark-mode/addon.json
```diff
@@ -992,6 +992,32 @@
...
+        "categoryMenu": "#202020",
+        "palette": "#202020cc",
+        "fullscreen": "#000000",
+        "stageHeader": "#000000",
+        "border": "#ffffff0d",
+        "dots": false
+      }
+    },
+    {
+      "name": "3.Darker Blue",
```
mxmou, line 1018:
```suggestion
      "name": "3.Darker (blue)",
```
(for consistency with "Scratch's default colors (blue)")

mxmou (approved):

background/get-addon-settings.js
```diff
@@ -53,10 +53,21 @@ const updatePresetIfMatching = (settings, version, oldPreset = null, presetOrFn
...
+    console.groupCollapsed("updatePresetIfMatching log");
     for (const key of Object.keys(oldPreset)) {
-      if (!areSettingsEqual(settings[key], oldPreset[key])) return console.log(settings, oldPreset, key);
+      if (!areSettingsEqual(settings[key], oldPreset[key])) {
+        console.log(`Did not match old preset (version ${version})`, {
+          settingsSnapshot: JSON.parse(JSON.stringify(settings)),
+          checkedAgainst: oldPreset,
+          keyFailed: key,
+        });
+        console.groupEnd();
```
mxmou, line 64: This is probably a mistake - there's another call to `groupEnd()` a few lines below this.

background/get-addon-settings.js
```diff
@@ -53,10 +53,21 @@ const updatePresetIfMatching = (settings, version, oldPreset = null, presetOrFn
...
+        return;
+      }
       if (typeof presetOrFn === "object") map[key] = presetOrFn.values[key];
     }
+    console.groupEnd();
+    console.warn(`Matched old preset (version ${version}) with settings:`, JSON.parse(JSON.stringify(settings)));
```
mxmou, line 70: Why `warn`?

background/get-addon-settings.js
```diff
@@ -322,6 +333,110 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
...
+              palette: "#333333cc",
+              border: "#444444",
+            },
+            () => {
+              console.log("Migrated Experimental Dark preset.");
+              madeAnyChanges = madeChangesToAddon = true;
+              settings.page = "#2e3238";
+              settings.primary = "#855cd6";
+              settings.highlightText = "#ccb3ff";
+              settings.menuBar = "#855cd6";
```
mxmou, line 364: Is this line necessary? The menu bar color is updated unconditionally.

background/get-addon-settings.js
WorldLanguages: Note that line 65 is `return;` so the lines below you are mentioning aren't actually being executed.

background/get-addon-settings.js
WorldLanguages: Yeah it's probably unnecessary but I think I got the color wrong in migration #10. Please see code comment below I just created.

background/get-addon-settings.js
```diff
@@ -322,6 +333,110 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
...
+            settings,
+            10,
+            {
+              // Old vanilla "menu bar background" (blue)
+              menuBar: "#4d97ff",
+            },
+            () => {
+              console.log("Migrated 'menu bar background' setting from old blue to new purple.");
+              madeAnyChanges = madeChangesToAddon = true;
+              settings.menuBar = "#9966ff"; // New vanilla "menu bar background" (purple)
```
WorldLanguages, line 437: Did I go crazy here? Shouldn't this be `#855cd6`?

background/get-addon-settings.js
WorldLanguages: I'll change it to `log`

background/get-addon-settings.js
mxmou: It should definitely be `#855cd6`. `#9966ff` is the color of Looks blocks.

WorldLanguages: No idea how I got those 2 colors confused

background/get-addon-settings.js
WorldLanguages: Yeah, this line is unnecessary then.

background/get-addon-settings.js
mxmou: I still don't think I understand why a group needs to be made. It doesn't look like there is ever more than one message in one of these groups.
Screenshot of console

WorldLanguages: I'll just reverse the console log changes before merging

background/get-addon-settings.js
mxmou: Probably because this PR initially changed the default for `primary` to `#9966ff`.

WorldLanguages: Fixed by https://github.com/ScratchAddons/ScratchAddons/pull/6074/commits/487ed9c392bff7908c1837a5a24f4856bd5abe4e

background/get-addon-settings.js
WorldLanguages: Fixed by https://github.com/ScratchAddons/ScratchAddons/pull/6074/commits/487ed9c392bff7908c1837a5a24f4856bd5abe4e

WorldLanguages:
TODOs for myself:

- [ ] Carefully check migration code to make sure all colors match either the setting values from the old presets or the new presets.
- [ ] Revert console logging changes to `updatePresetIfMatching` function (behavior of the function always stayed unchanged, I only added more logs)

WorldLanguages:
@mxmou Make sure to double check all color changes on `/editor-dark-mode/addon.json` (new defaults, preset changes, presets that should identically match old blue presets) one more time as we've seen it's easy to miss some mistakes. I'll double check the migration code

WorldLanguages (approved):

PR #6083, opened 2023-05-20
Update code related to stage size buttons

### LINKED ISSUE #6075: The stage size buttons were updated and break some addons (in `scratch-gui/beta`)
### Description

Since the Scratch color changes, which you can see on https://scratchfoundation.github.io/scratch-gui/beta/, some addons that have anything to do with small stage mode don't work properly.

### Shared space overflow

The shared space containing the volume slider plus the mouse position and clone count displays don't disappear when entering small stage mode.

The shared space overflows

### Hide stage and sprite pane doesn't work properly

The button to hide the stage doesn't display correctly.

A button placed incorrectly

It was actually different the second time I checked it out for whatever reason:

A button placed almost correctly

### Steps to reproduce

1. Run https://github.com/WorldLanguages/ScratchAddons/tree/test-on-scratchfoundation on the URL above.
2. Enable the relevant addon(s).

And no, this does not happen on `develop`.

### Environment

Browser: Edge 113
Scratch Addons version: v1.33.0-pre; https://github.com/ScratchAddons/ScratchAddons/pull/6053/commits/816be015c7fea281c47c32fd4957820c2287927e
Operating system: Windows 10

mxmou:
Resolves #6075

### Changes

* Fixes `hide-stage`, which didn't work at all in the beta version of scratch-gui.
* Changes the way `clones`, `mouse-pos`, `vol-slider`, `debugger`, `gamepad` and `sprite-properties` detect when the stage size is changed. The toggle buttons no longer have class names that would make them easy to identify. This could be solved using more complex selectors, but I decided to use Redux instead.

### Reason for changes

To make these addons work correctly when the update is released.

### Tests

Tested locally and on scratch.mit.edu.


addons/hide-stage/userscript.js
```diff
@@ -20,38 +25,56 @@ export default async function ({ addon, console, msg }) {
...
 
   addon.self.addEventListener("disabled", () => unhideStage());
 
   while (true) {
-    const stageControls = await addon.tab.waitForElement("[class*='stage-header_stage-size-toggle-group_']", {
+    let stageControls = await addon.tab.waitForElement("[class*='stage-header_stage-size-toggle-group_']", {
       markAsSeen: true,
       reduxCondition: (state) => !state.scratchGui.mode.isPlayerOnly,
     });
+    // Temporary code to support both current and future Scratch version
```
WorldLanguages, line 49: I assume we'll need to do this in many more places, it might be better to create a dedicated property or `addon.tab` or a module that can be imported.
Possibly a class name can be added to the body tag as well. What do you think?

WorldLanguages: Let me clarify: I was wondering if a boolean property/method would be useful to conditionally run code depending on whether Scratch is running in the old version or the new version.

WorldLanguages:
Just confirmed with GarboMuffin that there's no concrete reason why we previously used onclick events instead of Redux events. Probably just because Redux was a bit unstable at the time if you had some other extensions installed (this is not the case anymore).

WorldLanguages (approved):

Samq64 (approved):

PR #6088, opened 2023-05-21
Reference the blog from update notifications

### LINKED ISSUE #6079: Reference the blog from update notifications
### Why this would be helpful

Since the last two major updates didn't have a video to go with them, I think we should take the opportunity to link to our blog to provide a more exciting and complete list of changes.

### My suggestion

Add a link to the blog or most recent blog post at the time to the [major update notification](https://user-images.githubusercontent.com/17484114/236056243-36b797a2-15e5-4ccd-8c5c-7f92d75ac0cc.png).

### Possible alternatives

We could also link to the blog from the changelog.

### Additional context

At the time of posting this, [about 186 people have visited the blog in the past 14 days](https://analytics.scratchaddons.com/scratchaddons.com?page=%2Fblog%2F&period=custom&from=2023-05-05&to=2023-05-18), so this feature could bring more people to it.

### A new blog team member?

Speaking of the blog, I would also be available to write blog posts for major updates that include images.

<details><summary>Sample</summary>
<p>

We've been hard at work bringing in new customization options and a few features to speed things up, so let's take a look at what's new this time!

### Jump into editing faster

One of our newest addons, Faster Project Loading, reduces the time it takes to load into the editor directly (like by clicking "See inside" from My Stuff or taking a link to the editor). This actually happens because of a Scratch bug that adds unnecessary time to project loading only if you start in the editor, but we were able to fix that in this new, enabled-by-default addon!

### Up to the top!

Sometimes you want to quickly bring a sprite in front of all other sprites. We made an addon for that, too. Just hold Shift and click a sprite in the sprite pane to bring it to the front layer without dragging out any blocks.

### Never miss a post

Whether you're used to getting notifications or you want to keep up with your conversations, Auto Follow Topics can help. With this addon, any discussion forum topic you post in you will automatically follow. This means you automatically get messages for new posts just by joining the conversation.

### Save some time sifting through topics

While we're still looking at the forums, we have another addon called Necropost Highlighter. A necropost is when someone posts something that derails an inactive or resolved discussion and brings it to the top of the list (because the topics are sorted by latest activity). This addon flags old topics to indicate potential necroposts.

This isn't 100% accurate but you still might find it useful from time to time, like if you want to prioritize active conversations when you don't have a lot of free time.

### Don't use certain categories of forums?

We have one more forum enhancement - Remember Collapsed Forum Categories. If you didn't know, the forum listing page actually has collapsible groups, so you can hide certain categories of forums, but they would always expand again whenever you left that page and came back. But now, we made this addon which simply keeps them collapsed.

### A more compact volume control

The volume slider now defaults to only show an icon of a speaker, expanding into a slider by hovering over it, which looks cleaner and more fun, plus it saves some space up there.

You can choose between this behavior and the original, always expanded slider in the addon's settings.

### Popups can now have their own backdrop color

Previously, if you used Editor Dark Mode and Customizable Colors to make a custom theme, the backdrop behind popups could only be the same as the highlight color. Now, that's no longer the case! You can set it to whatever you want now, which means you can have black or white backdrops instead of blue or colored ones.

### Set aside the navigation bar

Don't like it when the navigation bar is anchored to the top of the screen _all the time_? Well, within the Customizable Navigation Bar addon, we've added an option to make it stick to the top of the page instead. With that setting, if you scroll down, the navigation bar won't follow you.

Enjoy the wider range of available customizations and we'll see you again for the next update!

</p>
</details>

iqnite:
Resolves #6079

### Changes

Added link to the Scratch Addons blog home page to the SA update notification.

### Reason for changes

It would bring more people to the blog, and it's easier to read than the changelog.

### Tests

- [x] Chrome
- [ ] Firefox


WorldLanguages:
See comment https://github.com/ScratchAddons/ScratchAddons/issues/6079#issuecomment-1556025050

WorldLanguages:
#6138 added the necessary strings. This is outdated

PR #6089, opened 2023-05-21
Fix unshared projects not always getting their iframes removed in `live-forum-preview`

### LINKED ISSUE #6086: live-featured-project doesn't always hide the project on 404
### Description

You can set unshared projects as featured projects, so live-featured-project will not work. In that case, it should hide the embed - it sometimes does that, but often not.

### Steps to reproduce

1. Go to any profile with an unshared featured project (for example, https://scratch.mit.edu/users/mybearworld/)
2. Reload the page until you see Giga with a magnifying glass

A demonstration

### Environment

Browser: Microsoft Edge 115.0.1851.0 (Official build) dev (64-bit) 
Scratch Addons version: 1.32.1
Operating system: Windows 11

mybearworld:
Resolves #6086

### Changes

Change the code to remove unshared projects so that it waits until the page has actually loaded before checking if a project is unshared. This also gets rid of the fallback to alternative editors when an unshared project is present, as that no longer works, and therefore also the `forceAlternative` setting.

### Reason for changes

To fix a bug.

### Tests
Tested in Microsoft Edge 115.0.1851.0 (Official build) dev (64-bit).

### DIFF
--- a/addons/live-featured-project/addon.json
+++ b/addons/live-featured-project/addon.json
@@ -63,15 +63,6 @@
       }
     },
     {
-      "id": "forceAlternative",
-      "name": "Force using alternative player",
-      "type": "boolean",
-      "default": false,
-      "if": {
-        "settings": { "alternativePlayer": ["turbowarp", "forkphorus"] }
-      }
-    },
-    {
       "id": "enableTWAddons",
       "name": "Enable TurboWarp addons",
       "type": "boolean",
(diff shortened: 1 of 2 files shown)


mybearworld:
Does this warrant migration code? The behavior of the settings does change.

Samq64:
Does this warrant migration code? The behavior of the settings does change.

> I don't think so.

Samq64:
Note there's another `live-featured-project` PR on the 1.33 milestone: #5423

Hans5958:
It is a duplicate of #5423. Unless there's something new on this PR, this can be closed for now.

PR #6097, opened 2023-05-23
New remix tree link placement option

### LINKED ISSUE #3676: Move the remix tree button

The remix tree button shouldn't belong there.

Untitled

Move the remix tree button somewhere else, maybe next to see inside?

DNin01:
Resolves #3676

### Changes

Moved the location of the remix tree button. It is now located next to the remixes side pane as a link.

The remixes list with a "Remix tree" link

It is also possible to change this behavior back to what it was before.

### Reason for changes

The previous placement of the remix tree button uses up more space than I think is necessary, and how often does an average user visit the remix tree, anyway?

### Tests

Tested in Edge 115 to the best of my ability, including after my refactor. Dynamically enabling, disabling and changing settings and switching between the project page and the editor worked great!


addons/remix-tree-button/addon.json
```diff
@@ -21,7 +25,30 @@
...
+      "default": "new"
+    }
+  ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "versionAdded": "1.2.0"
+  "versionAdded": "1.2.0",
+  "latestUpdate": {
+    "version": "1.33.0",
+    "temporaryNotice": "The link to the remix tree has moved. If you'd like to go back to the old behavior, you can change the setting below.",
```
WorldLanguages, line 51: Not sure if it's a good idea to force this... I bet many users won't find the button

WorldLanguages: Unless we do something like this: https://github.com/ScratchAddons/ScratchAddons/pull/3603#issuecomment-961820896

WorldLanguages: We could also try changing the default but only for users who have this addon disabled at the moment of the update (and for new users too, of course).

DNin01: > Not sure if it's a good idea to force this... I bet many users won't find the button
>
> We could also try changing the default but only for users who have this addon disabled at the moment of the update (and for new users too, of course).

I was definitely thinking about that.

Secret-chest:
The old remixes link, though.

DNin01:
> The old remixes link, though.

Which one are you taking about?

DNin01:
Hang on, I forgot about l10n.

I'm working on another feature for this addon so give me a minute...

Secret-chest:
Wasn't there a "remixes" link in the remixes column?

mxmou:
> > The old remixes link, though.
> 
> Which one are you taking about?

Probably the "View all" link which this change removes.

DNin01:
> Wasn't there a "remixes" link in the remixes column?

The one that goes to `/remixes`? If the user enables this addon, then they most likely aren't asking for that page.

addons/remix-tree-button/addon.json
DNin01: Turns out, I accidentally set the default to "new" instead of "replace", so the default hasn't changed, but I'm currently reconsidering what the default should be anyway.

DNin01: The default is now "new".

Secret-chest:
> The one that goes to /remixes? If the user enables this addon, then they most likely aren't asking for that page.

They're both useful, in different ways.

addons/remix-tree-button/addon.json
```diff
@@ -21,7 +25,42 @@
...
+        },
+        {
+          "id": "replace",
+          "name": "Replace \"Remixes \u2192 View all\" link"
+        }
+      ],
+      "default": "new"
+    },
+    {
+      "name": "Open remix tree by clicking remixes icon next to @heart.svg@star.svg",
```
Samq64, line 46: I think this should be one of the options above instead of its own toggle

addons/remix-tree-button/main.js
```diff
@@ -1,35 +1,174 @@
...
+      }
+    }
+  }
+
+  function undoLoadRemixButton() {
+    // Removes the button.
+    if (document.querySelector("#scratchAddonsRemixTreeBtn")) {
+      document.querySelector("#scratchAddonsRemixTreeBtn").style.display = "none";
+    }
+  }
```
Samq64, line 69: Could [addon.tab.displayNoneWhileDisabled](https://scratchaddons.com/docs/reference/addon-api/addon.tab/#using-addontabdisplaynonewhiledisabled-dynamicdisable) be used here? It would also mean you wouldn't need to change the display type when the addon is dynamically re-enabled.

addons/remix-tree-button/main.js
```diff
@@ -1,35 +1,174 @@
...
+      }
+    }
+  }
+
+  function goToRemixTree(e) {
+    // Follow the link to the remix tree after clicking the button.
+    if (addon.settings.get("open-with-icon") && !addon.self.disabled) {
+      if (addon.settings.get("new-tab") || e.ctrlKey || e.metaKey) {
+        window.open(
+          `https://scratch.mit.edu/projects/${window.location.href.split("projects")[1].split("/")[1]}/remixtree`,
```
Samq64, line 52: Just adding `remixtree` would be enough assuming there's a trailing slash in the URL, but if you want to be sure it would probably be better to store the project ID in a variable.

addons/remix-tree-button/main.js
```diff
@@ -1,35 +1,174 @@
...
+    if (addon.tab.editorMode === "projectpage") {
+      document.querySelector(".remix-list a").removeAttribute("target");
+      document.querySelector(".remix-list a").removeAttribute("rel");
+    }
+  }
+
+  // Initialize:
+  initialSetup();
+
+  function initialSetup() {
```
Samq64, line 120: I don't think this needs to be a function.

addons/remix-tree-button/addon.json
DNin01: This would be a good opportunity to allow multiple of these to be turned on at the same time. Maybe I should split everything into toggles so you don't have to lose one button to add another.

addons/remix-tree-button/main.js
DNin01: I'll look into it. It might be more complicated when dealing with settings that can be dynamically changed.

Samq64: Oh, right, I guess that's the problem with having multiple placements for the same addon.

DNin01:
Just redid everything and fixed a bug or two. It's so much easier to look at now.

I also removed the setting to click the 🌀 icon to go to the remix tree. It seems like you'd expect that to remix the project or something? Maybe I'm wrong. I might add it back, but I think we should prioritize reviewing what I've done already.

WorldLanguages:
Well, it's been a while... It's nice to see you've cleaned up the code with time though

addons/remix-tree-button/addon.json
```diff
@@ -21,7 +25,36 @@
...
+    }
+  ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "versionAdded": "1.2.0"
+  "versionAdded": "1.2.0",
+  "latestUpdate": {
+    "version": "1.33.0",
+    "temporaryNotice": "The remix tree link can now be moved so it doesn't get in the way.",
+    "newSettings": ["type", "open-with-icon"]
```
WorldLanguages, line 58: Wrong setting listed

addons/remix-tree-button/main.js
```diff
@@ -1,35 +1,121 @@
...
+      .waitForElement(".remix-list", {
+        reduxCondition: (state) => state.scratchGui.mode.isPlayerOnly,
+      })
+      .then(() => {
+        if (enabled) {
+          const link = document.querySelector(".remix-list a");
+          const linkSpan = link.firstChild;
+
+          if (!originalTextContent) {
+            originalTextContent = linkSpan.textContent; // Store what the button's text content was before the first time we change it (because different languages).
```
WorldLanguages, line 95: If you want, you can remove the global variable and just use `addon.tab.scratchMessage("project.viewAllInList")`

addons/remix-tree-button/addon.json
```diff
@@ -21,7 +25,36 @@
...
+      "type": "boolean",
+      "default": false
+    }
+  ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "versionAdded": "1.2.0"
+  "versionAdded": "1.2.0",
+  "latestUpdate": {
+    "version": "1.33.0",
```
WorldLanguages, line 56:
```suggestion
    "version": "1.36.0",
```

DNin01:
> > The one that goes to /remixes? If the user enables this addon, then they most likely aren't asking for that page.
> 
> They're both useful, in different ways.

Maybe having it go to the full remixes list instead of the remix tree when Shift-clicking the link would be okay?

Secret-chest:
> Maybe having it go to the full remixes list instead of the remix tree when Shift-clicking the link would be okay?

Pretty undiscoverable. Why not add both?

DNin01:
> > Maybe having it go to the full remixes list instead of the remix tree when Shift-clicking the link would be okay?
> 
> Pretty undiscoverable. Why not add both?

I could add it to the addon description or in a note or in a `title` property or something.

DNin01:
A Shift alternative wouldn't be touch-screen friendly, however.

BroJac5246:
> > > The one that goes to /remixes? If the user enables this addon, then they most likely aren't asking for that page.
> > 
> > They're both useful, in different ways.
> 
> Maybe having it go to the full remixes list instead of the remix tree when Shift-clicking the link would be okay?

Isn't that the browser shortcut for open page in new window?

DNin01:
> > Maybe having it go to the full remixes list instead of the remix tree when Shift-clicking the link would be okay?
> 
> Isn't that the browser shortcut for open page in new window?

Never mind, then.

WorldLanguages:
Well, the default behavior of the addon is not changed, so it should be fine. We could also think of creative ways to keep both the "View all" and remix tree buttons, maybe by showing a tree icon instead of "remix tree", but I'm not sure if all languages have space for that. In touchscreen it would also be difficult to touch a small tree icon.

Secret-chest:
Just make it like in 2.0 where you can click the remix icon to view a tree.

DNin01:
> Just make it like in 2.0 where you can click the remix icon to view a tree.

And make it possible for it to be enabled and for the two other options to be disabled?

That's actually an interesting idea.

Samq64:
`/remixtree` now redirects to `/remixs`.

Issue #5801, opened 2023-03-15
VSCode extension for SA developers


iqnite:
Make Scratch Addon's development even faster by creating an extension for Visual Studio Code. That extension would have the following features:

- Add syntax highlight and autocomplete support for Scratch Addons keywords (like `addon.tab`)
- Quickly set up addon folders (like what [create-sa-addon](https://github.com/scratchaddons-community/create-sa-addon) does)
- GUI for editing `addon.json` manifest files, which would be like a form with parameters to fill in (name, version added, credits, ...)
- Automatically link an addon's folder in the `addons.json` file when a new manifest file is created
- Show the equivalent Scratch block for JavaScript commands (and vice-versa)?

I'm aware that this very probably will not be accepted, but maybe it will serve as an inspiration to someone.

Samq64:
Related: #3968

WorldLanguages:
Related:
Our project has a lot of subfolders, it would be nice to provide a way to hide all other addon folders except the one that is currently under development.

iqnite:
> Related: Our project has a lot of subfolders, it would be nice to provide a way to hide all other addon folders except the one that is currently under development.

Adding a way to create folders in the "Open editors" panel would be good enough. That would make the extension useful even for non-SA-devs!

PR #6100, opened 2023-05-24
Add few default settings to `.vscode` dir (search excludes and JSON schemas)

WorldLanguages:
Related issue: #5801
Related PR: #3968

### Changes

- Hide non-English locale files from search and quick open
- Exclude third-party libraries from search by default
- Apply manifest schema checking by default to all addon manifest files
- Apply schema to /manifest.json file
- Apply inline schema to addon IDs (addons/addons.json file)

Multiple of these changes are inspired by or directly taken from #3968. Thanks RedGuy!


TheColaber:
> Group multiple files that sit in root (such as gitignore, LICENSE, readme, eslintrc) into a virtual folder (file nesting)

Why are they under the prettier file? Maybe it should be under the readme instead?

WorldLanguages:
> Why are they under the prettier file? Maybe it should be under the readme instead?

It could be under any of the files. But we need to choose one.

apple502j (changes requested):
Vetoing.

BroJac5246:
> Vetoing.

If you're going to do that then please explain **why**, otherwise there's no point to doing such a thing.

Also, you don't even have the right to veto in the first place. The definition of `veto` is:

> The constitutional power of the chief executive of a state or nation to prevent or delay the enactment of legislation passed by the legislature

The only people that that could refer to if we create an analogy are W_L and/or Griffpatch.

WorldLanguages:
Don't worry, we'll deal with this after the Scratch update on June 7th.

Hans5958 (approved):
Early support in response of veto. I'm not sure why we have to block things that improve editor experience (including, but not limited to, automated checks)

@BroJac5246 Apple had a history of "vetoing" without reason. I can say it's been a problem since the inception of the project. (see https://github.com/ScratchAddons/ScratchAddons/discussions/4886#discussioncomment-3232636)

.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["rslfrkndmrky.rsl-vsc-focused-folder"]
```
Hans5958, line 2: Should we suggest Prettier, since that's what we do on the CI? Or should we just leave it on the CI's job?

WorldLanguages:
Most of these changes clearly enhance the editor experience without negatively affecting users of other code editors. But keep in mind this PR changes multiple VSCode settings at once and they aren't all obvious, such as recommending an editor extension by an unknown author. It's very likely apple502j will only object a portion of the changes.

.vscode/extensions.json
WorldLanguages: We can probably add ESLint.
We could also add Prettier so that "format document" works as expected, but I don't think we should enable "lint on save".

apple502j, line 2: This extension was last updated 2.5 years ago. Shouldn't be on the recommended list.

.vscode/settings.json
```diff
@@ -0,0 +1,48 @@
+{
+  "explorer.fileNesting.enabled": true,
+  "explorer.fileNesting.expand": false,
+  "explorer.fileNesting.patterns": {
+    ".prettierrc.json": ".*, *.md, LICENSE"
```
apple502j, line 5: This makes no sense.

.vscode/settings.json
```diff
@@ -0,0 +1,48 @@
...
+    // All languages but English
+    "_locales/[abcdfghijklmnopqrstuvwxyz]*": true,
+    "_locales/[abcdefghijklmnopqrstuvwxyz][abcdefghijklmopqrstuvwxyz]*": true,
+    // All languages but English
+    "addons-l10n/[abcdfghijklmnopqrstuvwxyz]*": true,
+    "addons-l10n/[abcdefghijklmnopqrstuvwxyz][abcdefghijklmopqrstuvwxyz]*": true
+  },
+
+  "search.exclude": {
+    "libraries/**": true
```
apple502j, line 20: This should only exclude thirdparty libs

.vscode/settings.json
```diff
@@ -0,0 +1,48 @@
+{
+  "explorer.fileNesting.enabled": true,
+  "explorer.fileNesting.expand": false,
+  "explorer.fileNesting.patterns": {
+    ".prettierrc.json": ".*, *.md, LICENSE"
+  },
+
+  // Hides directories/files from sidebar.
+  // Excludes them from search by default, as well as from quick open.
+  "files.exclude": {
```
apple502j, line 10: Is it possible to exclude from search but not from sidebar? If not, please drop.

apple502j (changes requested):
Detailed review. If someone can file a PR with only the json schema change I'd approve.

.vscode/extensions.json
Hans5958: On second though, I think it would be more appropriate if we have `package.json` to bind specific versions of ESLint and Prettier.

> *ahem* one step to build step *ahem*

WorldLanguages: Is that bad on its own? VS Code APIs are pretty stable over time and this extension doesn't appear to have any dependencies that need updating.

.vscode/settings.json
WorldLanguages: Why? This groups multiple files that are rarely edited by contributors.

.vscode/settings.json
WorldLanguages: Okay. I'll exclude the license texts as well.

.vscode/settings.json
WorldLanguages: Yes, it is possible, but how often do you use VSCode to access these directories...?

apple502j (approved):

Issue #6009, opened 2023-05-05
More clearly represent the scope of each addon


DNin01:
### Why this would be helpful

There are icons on each addon in the list indicating the scope of the addon (whether it's for the editor, the website, etc.), but it's not very obvious to all users, so if we required users to use that information (for example, to make more sense of addons), some users could have trouble. Making it easier to see the scope of each addon is useful because:

- The addon's name doesn't have to say where it comes into play:
  - **Non-draggable sprites in editor** → Editing / **Non-draggable sprites**
  - **Costume editor snapping** → Editing / Costume Editor / **Object snapping**
  - **60FPS project player mode** → Projects / **60FPS mode**
  - **Live forum post preview** → Website / Forums / **Live preview**

- Users can look for addons for specific areas of Scratch (for example, find addons just for the project player, such as the volume slider). This is already possible through the filters on the left sidebar but could be made even easier with a rework of categories and grouping.

### My suggestion

Instead of grouping addons the way we currently do, which is by having a bunch of expandable sections like Recommended, Featured, Others, etc., we could change it to something like this:
![Groups for each scope](https://user-images.githubusercontent.com/106490990/236575833-a7e24803-55f6-4e60-be7b-d07f96c1d8cb.png)

If that solution has some problems, we could take that idea and combine it with the grouping setup we already have so that basically, within each current group (Recommended, Featured, etc.), there would be sections for each scope category of addons.
!["Recommended" group followed by "Editing: Code" subgroup](https://user-images.githubusercontent.com/106490990/236577345-e5c01ed7-7549-47da-b8ee-c48e40cb5f50.png)

This is actually done already - editor addons are placed above website addons, but there isn't an obvious separation between them - there's only the icons which are pretty easy to miss and only the major categories are separated between each other, so editor addons for all the different parts of the editor are all bunched together when they could be separated so that enhancements for the code editor, costume editor and other parts and aspects of the editor are all divided up into their own sections.

We also need to make it clearer how you find addons for the project player, such as the volume slider. Related issue: #3014

WorldLanguages:
> If that solution has some problems, we could take that idea and combine it with the grouping setup we already have so that basically, within each current group (Recommended, Featured, etc.), there would be sections for each scope category of addons.

This suggestion is probably the easiest to implement without us being forced to go deep into other setting page redesign discussions, which will take more time and effort.

[...]

PR #6104, opened 2023-05-25
Add "related addons" feature to settings page

### LINKED ISSUE #3933: "See also" section in addons

### LINKED ISSUE #5937: UX: "related addons" section (modal) in settings page
### Why this would be helpful

Finding related addons. IE all of the addons related to Block Customization

#### **Related Addons Modal:**
A button could be added which would pop up a modal with an explicitly defined list of addons in the `addons.json`
```json
"relatedAddons": ["addon1", "addon2", "etc"]
```
**Quick mockup:**
image
image

Additionally, if you would like to highlight a specific setting in the linked extension, you could provide an object instead like so
```json
"relatedAddons": [
  {
    "addon": "addon1",
    "settings": ["setting1", "setting2"]
  },
  "addon2",
  "etc"
]
```
image

### Possible alternatives

One alternative that I have considered is the button instead replacing your current search with the list of addons, but this would probably require changes to the fuzzy search engine to support direct enumerations.

### Additional context
See this previous post by from @lisa-wolfgang 
#3933

WorldLanguages:
Resolves #5937
Resolves #3933
Reverts #7991

### Changes (Outdated)

- Addons can specify related addons in their manifests. In the exact same order, these addons are displayed as a pseudo-info box that provides links to the related addon(s). Note that these relationships are not bidirectional - if you want addon A to recommend addon B and viceversa, both manifests need to be changed-
- When clicking a related addon, the "related addons" modal is opened and is automatically scrolled to the addon that the user clicked. The clicked addon is also highlighted using the already existing addon-blink animation.
- Toggling addons and changing settings inside the modal works the same as searching the addons normally, and is kept in sync as expected. If the user clicks a related addon inside this view, the link opens in a new tab and only highlights the addon that the user clicked.
- This functionality is not available in the popup view of the settings page.

brave_fVustW7hKO


Samq64:
Here's a table with some addon relations World Languages and a couple other came up with on Discord. Feel free to edit it.
| addon 1                        |      | addon 2                      |                              |                        |
|--------------------------------|------|------------------------------|------------------------------|------------------------|
| dark-www                       | <--> | editor-dark-mode             |                              |                        |
| editor-dark-mode               | -->  | editor-theme3                |                              |                        |
| editor-stage-left              | <--> | editor-buttons-reverse-order |                              |                        |
| block-duplicate                | <--> | block-cherry-picking         |                              |                        |
| discuss-button                 | <--> | expanding-search-bar         |                              |                        |
| editor-theme3                  | <--> | custom-block-shape           |                              |                        |
| vol-slider                     | <--> | mute-project                 |                              |                        |
| editor-theme3                  | <--> | transparent-orphans          |                              |                        |
| better-emojis                  | <--> | emoji-picker                 |                              |                        |
| curator-link                   | <--> | more-links                   |                              |                        |
| scratchr2                      | <--> | horizontal-mystuff-tabs      |                              |                        |
| zebra-striping                 | <--> | editor-theme3                |                              |                        |
| share-through-mystuff          | <--> | editor-unshare-button        |                              |                        |
| scratch-messaging              | <--> | scratch-notifier             | msg-count-badge              |                        |
| fix-editor-comments            | <--> | editor-comment-previews      |                              |                        |
| scratchr2                      | <--> | old-studio-layout            |                              |                        |
| color-picker                   | <--> | 2d-color-picker              | default-costume-editor-color | pick-colors-from-stage |
| remove-search-bar-autocomplete | <--> | expanding-search-bar         |                              |                        |
| search-sprites                 | <--> | editor-searchable-dropdowns  | middle-click-popup           |                        |
| hide-stage                     | <--> | hide-flyout                  | tutorials-button             | sprite-properties      |
| remove-sprite-confirm          | <--> | hide-delete-button           |                              |                        |
| block-switching                | <--> | middle-click-popup           |                              |                        |
| old-studio-layout              | <--> | items-per-row                |                              |                        |
| old-studio-layout              | <--> | studio-followers             |                              |                        |
| items-per-row                  | <--> | custom-zoom                  |                              |                        |
| studio-tools                   | <--> | studio-followers             |                              |                        |
| expanded-backpack              | <--> | items-per-row                |                              |                        |
| block-count                    | <--> | project-info                 |                              |                        |
| data-category-tweaks-v2        | <--> | columns                      |                              |                        |
| live-featured-project          | <--> | feature-unshared             |                              |                        |
| feature-unshared               | <--> | remix-button                 | longer-wiwo                  |                        |
| confirm-actions                | <--> | remove-sprite-confirm        |                              |                        |
| move-to-top-bottom             | <--> | folders                      |                              |                        |
| youtube-fullscreen             | <--> | true-youtube-links           |                              |                        |
| full-signature                 | <--> | hide-signatures              |                              |                        |
| mediarecorder                  | <--> | blocks2image                 |                              |                        |

WorldLanguages:
There's some conversation in a different issue #6009 about grouping addons by more specific subcategories, such as "color picker" and "messages page".
We should ensure a majority of the related-addon relationships we add aren't just subcategories. For example, I'm not sure if color picker addons should be related to each other. They aren't directly related - they simply affect the same area of the editor.

WorldLanguages:
I would also suggest that we add the related-addon relationships as part of a different PR, after this one is merged.

@mxmou @lisa-wolfgang Any feedback about the implementation of the feature? (not the suggested relationships themselves)

towerofnix:
Yay, exciting!

Your second "resolves" link at the top of this PR may be broken, it links to itself 😅

WorldLanguages:
Reminder: I haven't tested whether clicking outside the "related addons" modal closes the modal or not. It should.

DNin01:
> Reminder: I haven't tested whether clicking outside the "related addons" modal closes the modal or not. It should.

Clicking outside does close the modal.

lisa-wolfgang:
The implementation overall looks good to me.
Opening related addons from inside the related addons modal in a new tab seems a bit awkward, but it works well enough. One alternative would be to have a back button for the related addons window to allow for multiple layers of related addon navigation within the same tab, but I don't know if that would really be worth the effort.

DNin01:
> The implementation overall looks good to me. Opening related addons from inside the related addons modal in a new tab seems a bit awkward, but it works well enough. One alternative would be to have a back button for the related addons window to allow for multiple layers of related addon navigation within the same tab, but I don't know if that would really be worth the effort.

Or, alternatively, just not display related addons from inside the modal.

WorldLanguages:
> Or, alternatively, just not display related addons from inside the modal.

I have considered that, but I thought that was worse.

> One alternative would be to have a back button for the related addons window to allow for multiple layers of related addon navigation within the same tab, but I don't know if that would really be worth the effort.

It's not worth the effort right now, I think.
IMO opening in a new tab is better than providing no way to go back, and that's the best we can do for now.

WorldLanguages:
@BroJac5246 Reason for the dislike emoji reaction?

BroJac5246:
> @BroJac5246 Reason for the dislike emoji reaction?

I just don't really like this concept; I feel like we should find different methods of achieving the benefits of this. I'm somewhat _okay_ with it, but I don't exactly _like_ it.

MaterArc:
I kinda like the idea. I think it's a good way for people to search and find what they want to enable, it's a little bit overwhelming while looking at so many features anyways, so narrowing it down and picking out features that relate to the ones that the user already enabled is a good way to explore other addons that they might like.

Racecar-Roaster:
Like the fact that this can help find similar addons that people can also want in their scratch browser. I think this is helpful for someone who does animation on scratch. There can be similar tools to the colour HEX and I can find something also useful that I did not know I could use at all. And this can make the list of addons more organized and easier to find.

iqnite (changes requested):
This change would resolve many concerns, but there's one thing I don't understand: Wouldn't it be more effective to scroll to the related addons, instead of opening them in a modal? The user could then jump back through the other addon's "Related" link.
For example: dark-www is related to dark-editor. So the user is viewing the dark-www addon and clicks on the dark-editor link in the "Related addons" section of the dark-www addon, which opens dark-editor. If the user wants to go back, they click on dark-www in the "Related addons" section of dark-editor.

WorldLanguages:
> Wouldn't it be more effective to scroll to the related addons, instead of opening them in a modal?

This has a few downsides:
1. For a good user experience we'd need to properly support the back button, which is not trivial
2. It's sometimes useful to see all of the related addons in the same list

Also, "related addon" relationships might not be bi-directional in some cases.

DNin01:
60fps and turbowarp-player could probably be linked.

WorldLanguages:
@mxmou @samq64 I will resolve conflicts and try to get this merged this month

webpages/settings/components/addon-body.html
```diff
@@ -141,17 +149,24 @@
...
   }
 
-  .addon-credits > span:not(:first-child):not(:last-child)::after {
+  .addon-credits > span:not(:first-child):not(:last-child)::after,
+  .related-addons span:not(:first-child):not(:last-child)::after {
     content: ",";
   }
+  .related-addons span:first-child {
+    font-weight: bold;
+  }
```
WorldLanguages, line 169: These CSS changes only apply some .addon-credits styles to .related-addons as well. There are no new rules.

webpages/settings/components/addon-body.html
```diff
@@ -141,17 +149,24 @@
...
 
-  .addon-credits > span {
+  .addon-credits > span,
+  .related-addons span {
     margin-inline-end: 0.5rem;
   }
 
-  .addon-credits > span:not(:first-child):not(:last-child)::after {
+  .addon-credits > span:not(:first-child):not(:last-child)::after,
+  .related-addons span:not(:first-child):not(:last-child)::after {
```
WorldLanguages, line 164: These CSS changes only apply some .addon-credits styles to .related-addons as well. There are no new rules.

webpages/settings/components/addon-body.html
WorldLanguages, line 169: This makes `Related addons:` bold.

webpages/settings/components/addon-body.html
```diff
@@ -286,6 +304,9 @@
     display: flex;
     align-items: center;
   }
+  .related-addons > .addon-message {
+    flex-wrap: wrap;
+  }
```
WorldLanguages, line 309: Wrap is needed if the window is not wide enough to prevent the following unwanted layout:

image

Samq64:
Right now the scrollbar extends into the header. Shouldn't it stop at the top of the scrollable content?

WorldLanguages:
> Right now the scrollbar extends into the header. Shouldn't it stop at the top of the scrollable content?

Thanks, I accidentally fixed that with my latest commit. I unnecessarily reinvented the wheel instead of copying the CSS layout of the existing More Settings modal.

webpages/settings/components/modal.html
```diff
@@ -36,13 +36,17 @@ <h1>{{ title }}</h1>
     margin: auto;
     padding: 20px;
     width: 50rem;
-    max-width: 75%;
+    max-width: 80%;
```
WorldLanguages, line 39: This also makes the More Settings modal a bit wider.

webpages/settings/components/modal.html
```diff
@@ -1,6 +1,6 @@
 <template>
   <div v-cloak v-show="isOpen" class="modal">
-    <div class="modal-content" v-click-outside="clickOutside">
+    <div class="modal-content" v-click-outside="clickOutside" @click="closePickers">
```
WorldLanguages, line 3: This is necessary so that color pickers can be closed properly inside the modal

WorldLanguages:
This is how it looks now:

image

image

image

WorldLanguages:
@mxmou @DNin01 Maybe you could provide some feedback, do you think the screenshots I just sent look reasonable?

I'm thinking most addons might eventually get a "related addons" infobox assigned to them, that should mean we can't make it too distracting, or else it's equivalent to adding an info notice to virtually every addon.

Samq64:
What about putting it at the bottom?
image

WorldLanguages:
image

addons/custom-block-text/addon.json
```diff
@@ -16,6 +16,7 @@
       "link": "https://github.com/DNin01"
     }
   ],
+  "relatedAddons": ["editor-theme3", "zebra-striping", "custom-block-shape"],
```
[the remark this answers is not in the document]
WorldLanguages: These are for testing, I will propose the actual ones in a followup PR 👍

_locales/en/messages.json
```diff
@@ -149,6 +149,9 @@
   "resetToDefault": {
     "message": "Reset to default settings"
   },
+  "relatedAddons": {
+    "message": "Related addons:"
```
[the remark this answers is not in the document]
WorldLanguages: Good point, right now I'm reusing the string for the modal title which is not ideal.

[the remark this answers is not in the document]
Samq64: For the modal title it could be something like `Addons related to "{addon}"`?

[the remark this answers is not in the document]
mxmou: > For the modal title it could be something like `Addons related to "{addon}":`?

I think that's too long, especially if the addon has a long name. I suggest "Related addons:" for the label and "Related Addons" (title case) for the modal title.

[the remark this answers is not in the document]
WorldLanguages: @mxmou Good idea! Fixed

mxmou:
> If the user clicks a related addon inside this view, the link opens in a new tab and only highlights the addon that the user clicked.

This makes following a chain of "related addon" links - which I think is a reasonable thing to do - a very strange experience because the behavior of links alternates between opening a modal and opening a new tab. The simplest way to fix this is to remove the modal entirely and instead scroll to the related addon and highlight it. If the relations between addons are bidirectional, it will still be easy to go back to the previous addon.

Even if this problem was fixed in a different way, showing addons in a modal is quite weird with the current structure of the settings page. I'd prefer something like this:

image

mxmou:
> What about putting it at the bottom? image

Placing it at the bottom is a good idea but I think it should look like this screenshot, without a background. A better way to make it stand out a bit is to use a larger font size (14px) and a brighter text color (`--content-text` instead of `--description-text`).

WorldLanguages:
> Even if this problem was fixed in a different way, showing addons in a modal is quite weird with the current structure of the settings page. I'd prefer something like this:
> 
> image

That looks nice!
I would assume the value of the search bar, as well as any applied sidebar filters, would be ignored when on this view. That's some weirdness that we can ignore if we go for a modal (but by itself doesn't justify a modal being a superior choice)

WorldLanguages:
> This makes following a chain of "related addon" links - which I think is a reasonable thing to do - a very strange experience because the behavior of links alternates between opening a modal and opening a new tab. The simplest way to fix this is to remove the modal entirely and instead scroll to the related addon and highlight it.

There's some added complexity if we want to do that, for example we should think whether we want to clear the user's search and filters in order to scroll to the related addon, in cases where it's not already on the list.

Another approach we can consider, is to keep all of the Related Addons experience inside the same modal. It would be equivalent to stacking multiple modals one above the other, but the ones that get behind the new modal get lost.

Samq64:
If the modal is kept, a back button could be added.

mxmou:
> If the user doesn't see the back arrow for some reason, I wonder if they'll try to go back by clicking the All button on the sidebar. Kinda tricky to decide what that button should do in this situation.

I think clicking any sidebar item should exit the related addons view and switch to that category.

> That looks nice!
> I would assume the value of the search bar, as well as any applied sidebar filters, would be ignored when on this view. That's some weirdness that we can ignore if we go for a modal (but by itself doesn't justify a modal being a superior choice)

There should probably be a visual indication of this, for example making all items in the sidebar appear unselected.

Another thing to consider is that users might want to search for something while on this view. One solution is to exit it automatically when the value of the search bar is changed. Another option is to hide the search bar when viewing related addons - with the current position of the search bar that would be quite weird, but we could do something like this:





This would use more vertical space than the current layout with the search bar in the header, but it has another advantage: putting the search bar next to the name of the selected category might make it clearer that it will only search in that category. It also makes the selected category visible if the sidebar is closed. Unfortunately, it makes the search bar "hidden" in the corner, but it could be highlighted when focused to make it more noticeable (because it gets focus automatically when the page loads):

WorldLanguages:
Anyone want to take over?

Samq64:
This now works looks similar to mxmou's screenshots. There is no navigation tracking yet though so the back button exits the related addons view entirely and doesn't scroll to the originally clicked addon.

Samq64:
https://github.com/ScratchAddons/ScratchAddons/pull/6104/commits/322b460dc79d264fcd8b57fc15875e088af687c3 changed the 2 instances of `.addons-container.placeholder` to `.addons-container .placeholder` in `/webpages/settings/style.css` but GitHub doesn't like the fact that it's a white space only change.

Samq64:
The implementation isn't perfect but it works well apart from a race condition with the header text.

https://github.com/user-attachments/assets/b8de1d37-1118-44f1-ae30-55291caa214a

Samq64:
This was also a bug in the modal version but when clicking a related addon from the themes category all related addons also have the theme tag.

Adding a check in the `addon-tag.js` had a huge performance hit, so it's not fixed yet.

Samq64:
What should be done about performance? Currently navigating away from the related view causes the addon list to be reloaded without the placeholder.

WorldLanguages:
By placeholder do you mean the "loading indicators"? I think it's fine to only show them on first load.

The usefulness of listing "related addons" wins over any performance concerns.

webpages/settings/style.css
```diff
@@ -119,6 +112,41 @@ h1 {
...
+  border-bottom: 1px solid var(--control-border);
+  display: flex;
+  align-items: center;
+  padding: 10px 40px;
+  user-select: none;
+}
+
+.category-header-title {
+  flex-grow: 1;
+  font-size: 16px;
```
mxmou, line 126:
```suggestion
  font-size: 16px;
  font-weight: 500;
```

mxmou (commented):
I don't like the way the list of related addons is styled - see https://github.com/ScratchAddons/ScratchAddons/pull/6104#issuecomment-2579833331.

webpages/settings/components/addon-body.js
```diff
@@ -73,8 +73,9 @@ export default async function ({ template }) {
           this.addon._wasEverEnabled = this.addon._enabled || newState;
           this.addon._enabled = newState;
           // Do not extend when enabling in popup mode, unless addon has warnings
```
mxmou, line 75:
```suggestion
          // Do not extend when enabling in popup mode, unless addon has warnings
          // Do not collapse when disabling in related addons view
```

webpages/styles/components/search.css
```diff
@@ -1,14 +1,14 @@
 .search-box {
-  background: var(--search-toolbar-background);
-  border-bottom: 1px solid var(--control-border);
   display: flex;
   align-items: center;
-  padding-inline-start: 30px;
   user-select: none;
+  border-radius: 4px;
+  border: 1px solid transparent; /* prevent layout shift */
```
mxmou, line 6:
```suggestion
  border: 1px solid transparent; /* prevent layout shift when focused */
```

mxmou (approved):
Now that the name of the selected category is shown in the toolbar, it would make sense to put the button that opens the sidebar in small mode next to it. I can do that in a future PR.

iqnite (approved):
The implementation is good, but would merging this PR really "close" the issues? What about the addons listed in https://github.com/ScratchAddons/ScratchAddons/pull/6104#issuecomment-1563406851?

Also, the addon manifest validation workflow should be updated to not mark the "related addons" feature as incorrect.

Samq64:
https://github.com/ScratchAddons/ScratchAddons/pull/6104#discussion_r1904767014:

> These are for testing, I will propose the actual ones in a followup PR 👍

WorldLanguages:
The search bar on the right edge feels a bit weird to me.
Will new users find it easily?

WorldLanguages:
Other than that, I would like to say the back-button implementation is excellent.

Samq64:
> Will new users find it easily? (I see that a focus color was added, but I'm not sure if that's enough)

I considered using the same focus styles as other inputs but thought it was too distracting in the popup. What do you think?

Samq64:
I've pushed the suggested style changes.

One more minor thing: currently the related header is displayed as `Related to Addon name` keeping the name sentence case. Would one of these be better?
- `Related to addon name` (lowercase)
- `Related to: Addon name`
- `Related to "Addon name"`

mxmou:
> I've pushed the suggested style changes.
> 
> One more minor thing: currently the related header is displayed as `Related to Addon name` keeping the name sentence case. Would one of these be better?
> - `Related to addon name` (lowercase)
> - `Related to: Addon name`
> - `Related to "Addon name"`

I think `Related to: Addon name` is the most translation-friendly.

WorldLanguages:
@mxmou @Samq64 Good to merge?

PR #6105, opened 2023-05-26
Enhance `addon.tab.scratchClass` performance and consistency

### LINKED ISSUE #4106: Performance: change `addon.tab.scratchClass` to avoid using a mutation observer
This does not affect TurboWarp.

Currently, a mutation observer needs to be attached to the whole document forever, even if the page already loaded:
https://github.com/ScratchAddons/ScratchAddons/blob/master/content-scripts/inject/module.js#L316

This is because addons might use `addon.tab.scratchClass` synchronously but might not add the element to the DOM until needed, and scratchClass might return a placeholder class name starting with `scratchAddonsScratchClass/`.

The mutation observer was added as a fix for #1690. That issue states:
> We can fix this by adding a mutation observer. This may hurt performance slightly but we have many mutation observers already, one per each addon running addon.tab.waitForElement...

That's not true anymore since #2268

Possible solutions:
1. Making scratchClass async
2. Documenting how scratchClass should be properly used to avoid this
3. TODO, haven't thought too much about this. This was in my mind for a lot but I never wrote an issue until now.

It would also be cool to have cleaner code that handles the CSS classes differences between scratch-gui and scratch-paint cc @mxmou any suggestions?

WorldLanguages:
Resolves #4106

### Changes

- Userscripts that don't specify runAtComplete (or that explicitly set runAtComplete:true) won't run until `addon.tab.scratchClass()` is guaranteed to work properly.
- A persistent mutation observer that made a somehow expensive (haven't measured) document.querySelector call in the callback has been removed as a consequence of the above.
- Previously, `addon.tab.scratchClass()` failed often (depending on computer and internet speeds) when running scratch-gui locally or visiting https://scratchfoundation.github.io/scratch-gui/beta/ or https://scratchfoundation.github.io/scratch-gui/develop/. Now, it should always work properly. Exceptionally, some users suffered from this bug on scratch.mit.edu too in the release version (not 100% confirmed whether this PR should fix those cases, but surely it doesn't make them worse).
- Userscripts that run before complete will need to call `await addon.tab.scratchClassReady()` before using `addon.tab.scratchClass()`.

The algorithm that matches Scratch classes with the provided string remains unchanged.
TurboWarp only needs to implement `addon.tab.scratchClassReady = () => Promise.resolve()`.

### Reason for changes

Performance and bugs like this (find bar lacks styling, etc.):
image

### Tests

Tested in Chromium, in both scratch.mit.edu and /scratch-gui/beta

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -316,7 +316,7 @@ export default class Tab extends Listenable {
 
   scratchClassReady() {
     // Make sure to return a resolved promise if this is not a project!
-    const isProject = location.pathname.split("/")[1] === "projects";
+    const isProject = location.pathname.split("/")[1] === "projects" && location.pathname.split("/")[3] !== "embed";
     const isScratchGui = location.origin === "https://scratchfoundation.github.io" || location.port === "8601";
     if (!isProject && !isScratchGui) return Promise.resolve();
 
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -285,6 +285,9 @@ export default class Tab extends Listenable {
    * @returns {string} Hashed class names.
    */
   scratchClass(...args) {
+    if (!this._calledScratchClassReady)
+      throw new Error("Wait until addon.tab.scratchClassReady() resolves before using addon.tab.scratchClass");
+
     let res = "";
     args
       .filter((arg) => typeof arg === "string")
@@ -296,7 +299,7 @@ export default class Tab extends Listenable {
                 className.startsWith(classNameToFind + "_") && className.length === classNameToFind.length + 6
             ) || "";
         } else {
-          res += `scratchAddonsScratchClass/${classNameToFind}`;
+          throw new Error("addon.tab.scratchClass call failed. Class names are not ready yet");
         }
         res += " ";
       });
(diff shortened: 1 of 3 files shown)


addon-api/content-script/Tab.js
```diff
@@ -285,6 +285,9 @@ export default class Tab extends Listenable {
    * @returns {string} Hashed class names.
    */
   scratchClass(...args) {
+    if (!this._calledScratchClassReady)
```
apple502j, line 288: I think this would not catch cases where someone forgets to await the method call.

apple502j (approved):

addon-api/content-script/Tab.js
WorldLanguages: This will only catch cases where a runAtComplete:false script forgot to await the method call and is using `addon.tab.scratchClass` before any other scripts from the same addon on the same tab that run on full complete start running.

addon-api/content-script/Tab.js
```diff
@@ -311,6 +314,19 @@ export default class Tab extends Listenable {
     return res;
   }
 
+  scratchClassReady() {
+    // Make sure to return a resolved promise if this is not a project!
+    const isProject = location.pathname.split("/")[1] === "projects" && location.pathname.split("/")[3] !== "embed";
```
WorldLanguages, line 319: Not quite. https://scratch.mit.edu/projects/104/remixes/ isn't a project.

WorldLanguages: https://scratch.mit.edu/projects/104/studios/ too.

WorldLanguages: #6120

addon-api/content-script/Tab.js
```diff
@@ -296,7 +299,7 @@ export default class Tab extends Listenable {
                 className.startsWith(classNameToFind + "_") && className.length === classNameToFind.length + 6
             ) || "";
         } else {
-          res += `scratchAddonsScratchClass/${classNameToFind}`;
+          throw new Error("addon.tab.scratchClass call failed. Class names are not ready yet");
```
WorldLanguages, line 302: Can this actually happen? I think not

PR #5995, opened 2023-05-03
Add some setting descriptions

DNin01:
A step towards #5320

### Changes

Added several extended setting descriptions across a few addons. These will be available once setting descriptions are supported by Scratch Addons. Related discussion is available in the issue linked above.

### Reason for changes

To clarify some settings.

### Tests

Scratch Addons loads fine with these changes (although, of course, they are invisible because we haven't added the feature yet).

PR #6125, opened 2023-05-29
Rename some settings in Debugger addon

DNin01:
### Changes

Made some minor edits to the names of some settings and the update message in the Debugger addon.

### Tests

Tested.

### DIFF
--- a/addons/debugger/addon.json
+++ b/addons/debugger/addon.json
@@ -85,7 +85,7 @@
   "versionAdded": "1.16.0",
   "latestUpdate": {
     "version": "1.33.0",
-    "temporaryNotice": "There's now a low detail mode for graphs to increase performance. Turning on \"high detail graphs\" fills and animates graphs.",
+    "temporaryNotice": "There is now a low-detail mode for graphs to increase performance. Turning on \"high detail graphs\" fills and animates graphs.",
     "newSettings": ["fancy_graphs"]
   },
   "libraries": ["chartjs"]
--- a/addons/debugger/addon.json
+++ b/addons/debugger/addon.json
@@ -51,7 +51,7 @@
       "default": false
     },
     {
-      "name": "Log when max clones exceeded",
+      "name": "Log when clone limit exceeded",
       "id": "log_failed_clone_creation",
       "type": "boolean",
       "default": true
@@ -75,7 +75,7 @@
       "default": false
     },
     {
-      "name": "Fill and animate graphs",
+      "name": "High detail graphs",
       "id": "fancy_graphs",
       "type": "boolean",
       "default": false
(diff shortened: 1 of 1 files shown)


addons/debugger/addon.json
```diff
@@ -75,7 +75,7 @@
       "default": false
     },
     {
-      "name": "Fill and animate graphs",
+      "name": "High detail graphs",
```
Samq64, line 78: I'm not sure if this might be interpreted as increasing the precision of graph values.

DNin01: #5995 would help with this.

DNin01: Would switching this setting around and saying "low detail graphs" work better?

addons/debugger/addon.json
```diff
@@ -85,7 +85,7 @@
   "versionAdded": "1.16.0",
   "latestUpdate": {
     "version": "1.33.0",
-    "temporaryNotice": "There is now an option to disable the filling and animating of graphs.",
+    "temporaryNotice": "There's now a low detail mode for graphs to increase performance. Turning on \"high detail graphs\" fills and animates graphs.",
```
BroJac5246, line 88:
```suggestion
    "temporaryNotice": "There is now a low-detail mode for graphs to increase performance. Turning on \"high detail graphs\" fills and animates graphs.",
```

BroJac5246: Keep in mind that this will need to be revised again if Samq64's advice is accepted.

addons/debugger/addon.json
WorldLanguages: > #5995 would help with this.

Not really.

DNin01: > > #5995 would help with this.
> 
> 
> 
> Not really.

I mean, we could say turning it off improves performance.

WorldLanguages: > I mean, we could say turning it off improves performance.

Sure, but the setting name shouldn't imply that this setting has anything to do with precision anyway

DNin01:
```suggestion
      "name": "Animated graphs (may affect performance)",
```

addons/debugger/addon.json
WorldLanguages: Needs to be changed

addons/debugger/addon.json
```diff
@@ -85,7 +85,7 @@
   "versionAdded": "1.16.0",
   "latestUpdate": {
     "version": "1.33.0",
-    "temporaryNotice": "There is now an option to disable the filling and animating of graphs.",
+    "temporaryNotice": "There is now a low-detail mode for graphs to increase performance. Turning on \"high detail graphs\" fills and animates graphs.",
```
DNin01, line 88: We should also update this part. How's this?
```suggestion
    "temporaryNotice": "Graph animation is now disabled by default to increase performance. You can re-enable it if you want.",
```

WorldLanguages (approved):

Samq64 (approved):

Issue #6115, opened 2023-05-28
[June 28th] Scratch will be updating the website's accent color and other colors


DNin01:
The website is being updated to use the new purple accent color too! You can see the progress [here](https://github.com/scratchfoundation/scratch-www/tree/feature/color-contrast).

Trouble is, there are quite a few more addons that will need to be modified.

As per usual, we'll probably offer a preset to go back to the blue color for users that prefer that.

Samq64:
Will these changes also be released on June 7th?

DNin01:
> Will these changes also be released on June 7th?

I don't know for sure, but I'm guessing they will be.

Samq64:
Do you know if the changes would also effect `scratchr2` pages?

DNin01:
> Do you know if the changes would also effect `scratchr2` pages?

Probably not, but `dark-www` would affect both pages, so we could either not allow changing the accent color on `scratchr2` pages or have `dark-www` change the color and update any addons to adapt to it. If `dark-www` is disabled, those addons should stay the same to be consistent.

mxmou:
> > Do you know if the changes would also effect `scratchr2` pages?
> 
> Probably not, but `dark-www` would affect both pages, so we could either not allow changing the accent color on `scratchr2` pages or have `dark-www` change the color and update any addons to adapt to it. If `dark-www` is disabled, those addons should stay the same to be consistent.

Another option is to add separate settings for 2.0-styled pages.

DNin01:
> > > Do you know if the changes would also effect `scratchr2` pages?
> > 
> > 
> > Probably not, but `dark-www` would affect both pages, so we could either not allow changing the accent color on `scratchr2` pages or have `dark-www` change the color and update any addons to adapt to it. If `dark-www` is disabled, those addons should stay the same to be consistent.
> 
> Another option is to add separate settings for 2.0-styled pages.

That amount of customization would be overwhelming unless we figure something out.

Samq64:
> Another option is to add separate settings for 2.0-styled pages.

If we do that, I think only the navigation bar and link colours need separate settings, and they should be hidden when Scratch 2 to Scratch 3 is enabled.

DNin01:
> > Another option is to add separate settings for 2.0-styled pages.
> 
> If we do that, I think only the navigation bar and link colours need separate settings, and they should be hidden when Scratch 2 to Scratch 3 is enabled.

[...]

PR #6128, opened 2023-05-30
Change default website highlight color to purple

DNin01:
Part of #6115

### Changes

Scratch's editor accent color is updating to `#855cd6`. This updates each preset in `dark-www` to use purple instead of blue. I also kept a blue version of Scratch's default colors in case that's anyone's preference, but all other presets have had their blue changed to purple.

We'll need to do a bit more work to make it look good, but that can be a separate pull request if that's okay.

### Tests

Tested in Edge 113.

### Tasks

- [ ] Add migration code

### DIFF
--- a/addons/dark-www/addon.json
+++ b/addons/dark-www/addon.json
@@ -2353,7 +2353,7 @@
     {
       "name": "Scratch's default colors (blue)",
       "id": "scratch",
-      "description": "The colors normally used by Scratch",
+      "description": "The colors originally used by Scratch",
       "values": {
         "page": "#fcfcfc",
         "navbar": "#4d97ff",
--- a/addons/dark-www/addon.json
+++ b/addons/dark-www/addon.json
@@ -2214,7 +2214,7 @@
       "id": "navbar",
       "type": "color",
       "allowTransparency": true,
-      "default": "#4d97ff"
+      "default": "#855cd6"
     },
     {
       "name": "Content background",
@@ -2244,13 +2244,13 @@
       "name": "Highlight color",
       "id": "button",
       "type": "color",
-      "default": "#4d97ff"
+      "default": "#855cd6"
     },
     {
       "name": "Link color",
       "id": "link",
       "type": "color",
-      "default": "#4d97ff"
+      "default": "#ccb3ff"
     },
     {
       "name": "Footer background",
@@ -2299,13 +2299,13 @@
       "description": "A dark theme by Maximouse based on a userstyle. Not experimental",
       "values": {
         "page": "#202020",
-        "navbar": "#4d97ff",
+        "navbar": "#855cd6",
         "box": "#282828",
         "gray": "#333333",
         "blue": "#252c37",
         "input": "#202020",
-        "button": "#4d97ff",
-        "link": "#4d97ff",
+        "button": "#855cd6",
+        "link": "#ccb3ff",
         "footer": "#333333",
         "border": "#606060",
         "darkBanners": "darker",
@@ -2318,13 +2318,13 @@
       "description": "A dark theme by Ucrash",
       "values": {
         "page": "#242527",
-        "navbar": "#4d97ff",
+        "navbar": "#855cd6",
         "box": "#2f3137",
         "gray": "#424346",
         "blue": "#1b1d1f",
         "input": "#3a3a3a",
-        "button": "#4d97ff",
-        "link": "#4d97ff",
+        "button": "#855cd6",
+        "link": "#ccb3ff",
         "footer": "#17181a",
         "border": "#000000",
         "darkBanners": "desaturated",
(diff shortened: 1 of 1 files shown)


addons/dark-www/addon.json
```diff
@@ -2343,7 +2362,7 @@
         "blue": "#e9f1fc",
         "input": "#fafafa",
         "button": "#4d97ff",
-        "link": "#4d97ff",
+        "link": "#2a77e4",
```
DNin01, line 2365: Also note that I darkened the link color for Scratch's default colors (blue).

WorldLanguages:
Bump

We will also need to migrate settings to new presets here.

WorldLanguages:
bump @mxmou 

We need to decide the preset changes and the setting migration logic.

addons/dark-www/addon.json
mxmou: I think users will prefer an unmodified version of the old colors.

WorldLanguages: Note that this line refers to the new `Scratch's default colors` preset (not the "blue" version).

@DNin01 Did Scratch modify the link color for links in the scratch-www repo?

DNin01: > ~Note that this line refers to the new `Scratch's default colors` preset (not the "blue" version).~ Sorry, the line numbers confused me. This does affect the "blue" version of the default colors.
> 
> @DNin01 What's the motivation for changing this preset instead of directly cloning the old preset as of v1.33.0?

I just wanted to increase the contrast a little because the old color probably wouldn't pass any guidelines.

Thinking about it now, this might not be the only element that is inaccessible with this blue color and the user can always change it if necessary, so maybe we shouldn't default to a darker link color in this preset.
```suggestion
        "link": "#4d97ff",
```

WorldLanguages:
waiting

addons/dark-www/addon.json
```diff
@@ -2297,6 +2297,25 @@
       "name": "\"Experimental\" Dark",
       "id": "experimentalDark",
       "description": "A dark theme by Maximouse based on a userstyle. Not experimental",
+      "values": {
+        "page": "#202020",
+        "navbar": "#855cd6",
+        "box": "#282828",
+        "gray": "#333333",
+        "blue": "#252c37",
```
mxmou, line 2305:
```suggestion
        "blue": "#292d32",
```
I think the website and editor versions of Experimental Dark should stay as similar as possible. This color was changed in the editor version to be less saturated.

addons/dark-www/addon.json
```diff
@@ -2297,6 +2297,25 @@
...
+        "footer": "#333333",
+        "border": "#606060",
+        "darkBanners": "darker",
+        "darkForumCode": true
+      }
+    },
+    {
+      "name": "\"Experimental\" Dark (blue)",
+      "id": "experimentalDark",
+      "description": "A dark theme by Maximouse based on a userstyle. Not experimental",
```
mxmou, line 2318:
```suggestion
      "description": "\"Experimental\" Dark with a blue accent color",
```
(based on the description for the blue version of 3.Darker)

addons/dark-www/addon.json
```diff
@@ -2305,7 +2324,7 @@
         "blue": "#252c37",
         "input": "#202020",
         "button": "#4d97ff",
-        "link": "#4d97ff",
+        "link": "#67a7ff",
```
mxmou, line 2327: I don't think this change is necessary - users who decide to revert to the blue color probably don't care about contrast. But if we do change this color, the same change should be made to the blue version of 3.Darker.

mxmou:
I think `navbar` and `button` can be migrated even if the other settings don't match any preset. `link` can't because its new value isn't the same in all presets.

WorldLanguages:
@mxmou Got it. We can migrate `navbar` and `button` unconditionally, just like migration 10 from editor-dark-mode.

Regarding `link`, should we ignore some settings when checking whether a specific preset is currently active? Probably doesn't make sense to check whether `darkForumCode` is enabled or not.

mxmou:
> Regarding `link`, should we ignore some settings when checking whether a specific preset is currently active? Probably doesn't make sense to check whether `darkForumCode` is enabled or not.

Checking color settings is probably enough - `darkBanners` and `darkForumCode` can be ignored.

WorldLanguages:
@DNin01

DNin01:
It looks like the setting migration worked. I don't remember if I changed any settings (I don't think I did). I don't know how to test migration logic more than once yet.

Anyway, as expected, the navigation bar and highlight colors became purple and the link is still blue for now.

WorldLanguages:
I meant the unresolved code reviews :P

DNin01:
> I meant the unresolved code reviews :P

Oh okay lol

All ready!

WorldLanguages:
I implemented migration steps:
1. If `navbar` matches old default, set it to new default
2. If `button` matches old default, set it to new default
3. If current settings look like* Experimental Dark, set `blue` and `link` to their new preset defaults
4. If current settings look like* Dark WWW, set `link` to its new preset default
5. If current settings look like* Scratch default colors, set `link` to its new preset default

\* Checks settings `page`, `box`, `gray`, `blue`, `input`, `link`, `footer`, `border`. **Does not check `navbar` nor `button`.** Does not check the remaining boolean settings.

DNin01:
The link color setting has now been migrated too!

addons/dark-www/addon.json
```diff
@@ -2297,6 +2297,25 @@
       "name": "\"Experimental\" Dark",
       "id": "experimentalDark",
       "description": "A dark theme by Maximouse based on a userstyle. Not experimental",
+      "values": {
+        "page": "#202020",
+        "navbar": "#855cd6",
+        "box": "#282828",
+        "gray": "#333333",
+        "blue": "#292d32",
```
mxmou, line 2305: The default value of this setting needs to be updated to match the Experimental Dark preset.

mxmou (changes requested):

addons/dark-www/addon.json
WorldLanguages: Good catch, the addon defaults should match the `"Experimental" Dark` preset!

WorldLanguages (approved):

WorldLanguages:
@mxmou We can release v1.33.2 to stores (but not yet to users) after your approval here.

mxmou (approved):

PR #6131, opened 2023-05-31
Shorten URLs for Cloud Games

DNin01:
Resolves comment https://github.com/ScratchAddons/ScratchAddons/issues/5190#issuecomment-1404142737

### Changes

Removed the common start `https://scratch.mit.edu/` from URLs in the Cloud Games addon's preset URLs, so instead of having a URL like `https://scratch.mit.edu/studios/539952`, it will just be `studios/539952`.

From my testing, Cloud Games is already capable of processing shortened URLs acting as relative paths to parts of `scratch.mit.edu`, so no code changes were required.

Note that the project URLs don't need to begin with `projects/` (if they don't specify whether to load a project or a studio, the default will be a project), but I kept them there for simplicity.

### Reason for changes

To save space in the input boxes in the settings page and take up less space in storage.

### Tests

Tested in Edge 113.

### Tasks

- [ ] Ideally, we should have migration code set up to remove it for existing users.

### DIFF
--- a/addons/cloud-games/addon.json
+++ b/addons/cloud-games/addon.json
@@ -29,7 +29,7 @@
       "type": "table",
       "default": [
         {
-          "url": "https://scratch.mit.edu/studios/539952/"
+          "url": "studios/539952"
         }
       ],
       "row": [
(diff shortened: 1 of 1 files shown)


DNin01:
> This is okay as long as users still know they can paste full URLs.

Perhaps if a full URL is inputted, it can be shortened automatically?

WorldLanguages:
That would work great for compression but the user still has no way to know that they can paste a full URL and that we will take care of it.

DNin01:
> That would work great for compression but the user still has no way to know that they can paste a full URL and that we will take care of it.

Yeah... 🤔

Just wondering, how might that be an issue anyway? It only takes a few seconds to trim off the start of the URL out and most users will probably not add that many projects in a short time, so it doesn't seem too important to me for the user to know this.

mybearworld:
`discuss-button` has links cut off by default without any mention that they can also be regular, which doesn't seem to cause any problems.

mybearworld:
That's true, but that's just an implementation detail. To the end user, there is no difference.

PR #6137, opened 2023-06-01
Fix `addon.tab.scratchClass` bugs after refactor

### LINKED ISSUE #6136: Many addons don't work if dynamically enabled
### Description

Dynamically enabling any addon that uses `addon.tab.scratchClass` in a userscript with `"runAtComplete": true` results in the following error:
```
Uncaught (in promise) Error: Wait until addon.tab.scratchClassReady() resolves before using addon.tab.scratchClass
    at Tab.scratchClass (Tab.js:289:13)
```

The following addons are affected:
* better-img-uploads
* color-picker (only in code editor)
* echo-effect
* editor-messages
* find-bar
* gamepad
* hide-stage
* swap-local-global
* variable-manager

### Steps to reproduce

1. Reload the editor
2. Enable one of the addons listed above
3. Try to use the addon

_No response_

### Environment

Browser: Edge 115 dev
Scratch Addons version: f9813518
Operating system: Windows

### LINKED ISSUE #6153: confirm-actions confirmations don't show description and have broken close buttons
### Description

When using the confirmations addon in 1.33.0-prerelease, the confirmation boxes don't have a description and the close button doesn't do anything.

### Steps to reproduce

1. Enable 1.33.0-prerelease
2. Go to a profile page
3. Enable confirm-actions
4. Trigger a confirmation

A broken unfollow dialog
On scratch-www pages, the white part doesn't show up either.

### Environment

Browser: Edge 114
Scratch Addons version: 1.33.0-prerelease, I don't know how to get a hash
Operating system: Windows 11

WorldLanguages:
Resolves #6136
Resolves #6153

### Changes

Fixes multiple very stupid mistakes.

### DIFF
--- a/content-scripts/inject/run-userscript.js
+++ b/content-scripts/inject/run-userscript.js
@@ -9,7 +9,7 @@ export default async function runAddonUserscripts({ addonId, scripts, enabledLat
     const loadUserscript = async () => {
       await scratchAddons.l10n.loadByAddonId(addonId);
       const module = await import(scriptUrl);
-      await addonObj.tab.scratchClassReady();
+      if (runAtComplete) await addonObj.tab.scratchClassReady();
       const msg = (key, placeholders) =>
         scratchAddons.l10n.get(key.startsWith("/") ? key.slice(1) : `${addonId}/${key}`, placeholders);
       msg.locale = scratchAddons.l10n.locale;
--- a/content-scripts/inject/run-userscript.js
+++ b/content-scripts/inject/run-userscript.js
@@ -7,9 +7,11 @@ export default async function runAddonUserscripts({ addonId, scripts, enabledLat
     const { url: scriptPath, runAtComplete } = scriptInfo;
     const scriptUrl = `${new URL(import.meta.url).origin}/addons/${addonId}/${scriptPath}`;
     const loadUserscript = async () => {
-      await scratchAddons.l10n.loadByAddonId(addonId);
-      const module = await import(scriptUrl);
-      if (runAtComplete) await addonObj.tab.scratchClassReady();
+      const [module] = await Promise.all([
+        import(scriptUrl),
+        scratchAddons.l10n.loadByAddonId(addonId),
+        runAtComplete ? addonObj.tab.scratchClassReady() : Promise.resolve(),
+      ]);
       const msg = (key, placeholders) =>
         scratchAddons.l10n.get(key.startsWith("/") ? key.slice(1) : `${addonId}/${key}`, placeholders);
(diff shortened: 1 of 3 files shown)


content-scripts/inject/run-userscript.js
```diff
@@ -9,6 +9,7 @@ export default async function runAddonUserscripts({ addonId, scripts, enabledLat
     const loadUserscript = async () => {
       await scratchAddons.l10n.loadByAddonId(addonId);
       const module = await import(scriptUrl);
+      await addonObj.tab.scratchClassReady();
```
WorldLanguages, line 12: In the future we can probably speed this up a bit with `Promise.all()`
This change does not make it slower though. But the l10n and module import could happen in parallel.

mxmou (approved):

content-scripts/inject/run-userscript.js
mxmou: Is there a reason for not making this improvement now?

WorldLanguages: Well, technically the bug fix for this regression is also a speed up (just noticed) because the order of the await calls has been changed, so it can be done now.

apple502j (approved):

apple502j (changes requested):
Actually, not ready. `runAtComplete: false` should not call the method unless dynamically enabled.

WorldLanguages:
> `runAtComplete: false` should not call the method unless dynamically enabled.

I think it shouldn't call the method, no matter if the addon was dynamically enabled or not.

content-scripts/inject/run-userscript.js
WorldLanguages: Done

WorldLanguages:
I think it's okay now, but I got this wrong too many times

CST1229:
Why not use `await addon.tab.scratchClassReady()` in addons instead of delaying loading all userscripts?

WorldLanguages:
> Why not use `await addon.tab.scratchClassReady()` in addons instead of delaying loading all userscripts?

Only userscripts that don't "run at complete" need to manually call it. We could theoretically extend it to all userscripts, but I thought it made more sense to add this as a precondition for `runAtComplete:true`. Would you say the page is "complete" if the classes aren't available yet?

Samq64:
Ideally this would also fix #6153.

apple502j (approved):

WorldLanguages:
There was one pending case I was about to investigate, `addon.tab.createEditorContextMenu` calls `addon.tab.scratchClass()`, which affects these two addons that don't run at complete: `folders` and `move-to-top-bottom`.

PR #6138, opened 2023-06-01
Add v1.33.0 update notification

WorldLanguages:
image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -251,7 +251,7 @@
   "extensionUpdateInfo2_v1_33": {
     "message": "Multiple addons have been adjusted to work with the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
   },
-  "extensionUpdateInfo3_v1_33": {
+  "readTheBlogPost": {
     "message": "Read the $1 for more information.",
     "description": "The placeholder is \"blog post\""
   },
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -242,14 +242,21 @@
   "extensionHasUpdated": {
     "message": "Scratch Addons has updated to v$1"
   },
-  "extensionUpdateInfo1_v1_32": {
-    "message": "New addon: \"move sprite to front layer\", which lets you Shift+Click a sprite within the sprite pane to move it to the front (top layer) of the stage. Go to $1 to enable it."
+  "extensionUpdateInfo1_v1_33": {
+    "message": "New addon: \"highlight project creator in comments\", which adds a badge next to the project creator in project comments sections. Go to $1 to enable it."
   },
   "scratchAddonsSettings": {
     "message": "Scratch Addons settings"
   },
-  "extensionUpdateInfo2_v1_32": {
-    "message": "Other new addons added this update are \"faster project loading\", \"follow topics after posting\", \"necropost highlighter\", and \"remember collapsed forum categories\"."
+  "extensionUpdateInfo2_v1_33": {
+    "message": "Multiple addons have been adjusted to work with the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
+  },
+  "extensionUpdateInfo3_v1_33": {
+    "message": "Read the $1 for more information.",
+    "description": "The placeholder is \"blog post\""
+  },
+  "blogPost": {
+    "message": "blog post"
   },
   "notAffiliated": {
     "message": "Scratch Addons is not affiliated with Scratch."
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -249,7 +249,7 @@
     "message": "Scratch Addons settings"
   },
   "extensionUpdateInfo2_v1_33": {
-    "message": "Multiple addons have been adjusted to work with the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
+    "message": "Multiple addons have been adjusted in preparation for the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
   },
   "readTheBlogPost": {
     "message": "Read the $1 for more information.",
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -249,7 +249,7 @@
     "message": "Scratch Addons settings"
   },
   "extensionUpdateInfo2_v1_33": {
-    "message": "Multiple addons have been adjusted in preparation for the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
+    "message": "Multiple addons have been adjusted in preparation for the upcoming Scratch changes. This update also brings performance and stability improvements."
   },
   "readTheBlogPost": {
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Note: version v1.33.1 will *probably* have its own dedicated update notification message.

_locales/en/messages.json
```diff
@@ -242,14 +242,21 @@
...
-    "message": "Other new addons added this update are \"faster project loading\", \"follow topics after posting\", \"necropost highlighter\", and \"remember collapsed forum categories\"."
+  "extensionUpdateInfo2_v1_33": {
+    "message": "Multiple addons have been adjusted to work with the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
+  },
+  "readTheBlogPost": {
+    "message": "Read the $1 for more information.",
+    "description": "The placeholder is \"blog post\""
+  },
+  "blogPost": {
+    "message": "blog post"
```
WorldLanguages, line 259: These 2 strings above are unused in the code as of now, but it's a good idea to send them for translation.

_locales/en/messages.json
```diff
@@ -242,14 +242,21 @@
...
+  "extensionUpdateInfo1_v1_33": {
+    "message": "New addon: \"highlight project creator in comments\", which adds a badge next to the project creator in project comments sections. Go to $1 to enable it."
   },
   "scratchAddonsSettings": {
     "message": "Scratch Addons settings"
   },
-  "extensionUpdateInfo2_v1_32": {
-    "message": "Other new addons added this update are \"faster project loading\", \"follow topics after posting\", \"necropost highlighter\", and \"remember collapsed forum categories\"."
+  "extensionUpdateInfo2_v1_33": {
+    "message": "Multiple addons have been adjusted to work with the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
```
DNin01, line 252: > Multiple addons have been adjusted to work with the upcoming Scratch changes.

The changes we're making to accommodate for the Scratch changes are behind-the-scenes, and most users are not aware of them, so do we need to say this?

> increased storage for addon settings,

This sounds almost like we're raising a limit that's part of Scratch Addons, when in reality, it's a limit imposed by Chrome's storage API, and we've never documented or intended such a limit. Plus, most people have probably never made enough customizations to hit the storage limit. This enhancement is kind of technical and if we're trying to put it in a user-friendly summary of changes, I think we should only bring it up if we can explain it in further detail. Maybe we could just include it in the changelog, because patch notes tend to include the technical stuff.

DNin01 (commented):
I have some UX suggestions related to the update message.

_locales/en/messages.json
WorldLanguages: Sure, but what would you highlight instead?

DNin01: > Sure, but what would you highlight instead?

I know... 😕

Samq64: > The changes we're making to accommodate for the Scratch changes are behind-the-scenes, and most users are not aware of them, so do we need to say this?

I think it should stay in the notification because it's the main focus of the update.

_locales/en/messages.json
Samq64: I think just adding it with the other links is fine, but in the notification text it will get more attention.

_locales/en/messages.json
DNin01: > > The changes we're making to accommodate for the Scratch changes are behind-the-scenes, and most users are not aware of them, so do we need to say this?
> 
> I think it should stay in the notification because I think it's the main focus of the update.

I haven't seen any announcements about the color changes on Scratch yet, so we would need to explain what this was all about.

DNin01: Maybe:
```suggestion
    "message": "Multiple addons have been adjusted in preparation for an upcoming Scratch update. This update also brings performance improvements, increased storage for addon settings, and more."
```
Vague, but true.

Saying "work with" might be a bit much, because there are still some problems that need to be worked out for this to be considered total compatibility.

And I realize this message is pretty important if we don't make it to v1.33.1 in time. Then, at least users that read the notification will have a better understanding.

DNin01: My only concern with that is:
> upcoming Scratch update. This update also

"Update" appearing close to another "update", which could imply that they are related, which they are not - one is about Scratch and the other is about Scratch Addons.

DNin01: We might be able to say "the upcoming Scratch changes" after all, since the Scratch Team appears to be setting up a notification before the update.

DNin01 (approved):
I'd still suggest a clarification of "increased storage for addon settings" if possible, but everything else looks good to me.

_locales/en/messages.json
```diff
@@ -242,14 +242,21 @@
...
+  "extensionUpdateInfo1_v1_33": {
+    "message": "New addon: \"highlight project creator in comments\", which adds a badge next to the project creator in project comments sections. Go to $1 to enable it."
   },
   "scratchAddonsSettings": {
     "message": "Scratch Addons settings"
   },
-  "extensionUpdateInfo2_v1_32": {
-    "message": "Other new addons added this update are \"faster project loading\", \"follow topics after posting\", \"necropost highlighter\", and \"remember collapsed forum categories\"."
+  "extensionUpdateInfo2_v1_33": {
+    "message": "Multiple addons have been adjusted in preparation for the upcoming Scratch changes. This update also brings performance improvements, increased storage for addon settings, and more."
```
WorldLanguages, line 252:
```suggestion
    "message": "Multiple addons have been adjusted in preparation for the upcoming Scratch changes. This update also brings performance and stability improvements."
```

DNin01: Nice!

WorldLanguages: I know understand why the update messages in the App Store back in 2013 were so vague. Miss my iPod Touch running iOS 6.

_locales/en/messages.json
DNin01: > We might be able to say "the upcoming Scratch changes" after all, since the Scratch Team appears to be setting up a notification before the update.

Although now I'm not so sure if users will understand if we publish this release before Scratch notifies us of the update.

WorldLanguages: Probably not going to happen before.

DNin01: > I'd still suggest a clarification of "increased storage for addon settings" if possible, but everything else looks good to me.

It looks like [it will become visible on June 21st](https://github.com/scratchfoundation/scratch-www/commit/b7c1bada77c0d17e5362cfbfc3e57e4c3c819279#diff-70634dad4e029ddb3a0f1490f5b44368374234f6e852a6d8960cb9a7cfb1da61R22-R23), which is two weeks from now.

WorldLanguages: Don't know, I've seen too many dates by now.

mxmou: The code for the announcement banner will be released this week, but it won't be visible until the 21st. It links to a forum post, which will probably be made public at the same time.

PR #6142, opened 2023-06-02
Dynamic enable and disable for sounds-duration

DNin01:
A step towards https://github.com/ScratchAddons/ScratchAddons/issues/1673

### Changes

Added dynamic enable and disable support to `sounds-duration`. The addon actually continues running in the background but is invisible while disabled.

### Reason for changes

We're trying to add dynamic enable/disable support to all addons at some point.

### Tests

Tested on Edge 113.

### DIFF
--- a/addons/sounds-duration/addon.json
+++ b/addons/sounds-duration/addon.json
@@ -20,5 +20,7 @@
       "matches": ["projects"]
     }
   ],
+  "dynamicEnable": true,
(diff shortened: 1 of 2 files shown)


Samq64 (approved):
Tested on Firefox. It looks like it stays accurate when dynamically enabling while a sound is playing.

apple502j (approved):

PR #6146, opened 2023-06-04
editor-theme3: compatibility with Scratch high contrast theme

### LINKED ISSUE #5314: Motion block icon outline is not colored
### Description

I look closer that Motion Icon outline block is not colored

### Steps to reproduce

1. Enable Customizable block colors
2. Customize Motion Color Block

Screenshot 2022-10-29 180622

### Environment

Browser: Opera
Operating system: Windows 11

mxmou:
Resolves #5314

### Changes

Fixes bugs:
* Matrix inputs were colored incorrectly - this was broken by #5257
* Octave buttons in "play note" dropdown didn't become black in black text mode

The following bugs only happened in the beta version of scratch-gui:
* Changing the language or switching to the project page and back would reset the colors
* Icons remained black if the addon was enabled and the Scratch theme was set to high contrast
  * I fixed this by including white versions of icons in the addon, which made it very easy to fix #5314
* If the Scratch theme was set to high contrast, the text color was wrong after the addon was dynamically disabled

### Reason for changes

To improve compatibility with the Scratch changes.

### Tests

Tested locally and on scratch.mit.edu.


Samq64 (approved):
Tested on Firefox. I haven't tested very many edge cases, but dynamically enabling and disabling the addon with both of Scratch's modes and using black and white text works as expected.

apple502j (approved):

WorldLanguages:
This was reverted by #6216. This change made the bug described at #6212 easier to experience by users by simply attempting to create a new block. The root cause of the bug is described at #6212.

WorldLanguages:
> The following bugs only happened in the beta version of scratch-gui:

@mxmou We would need to merge these bug fixes back for v1.33.2, right?

mxmou:
> > The following bugs only happened in the beta version of scratch-gui:
> 
> @mxmou We would need to merge these bug fixes back for v1.33.2, right?

Yes, but #6227 will need to be merged first to prevent the crash from happening again.

addons/editor-theme3/theme3.js
```diff
@@ -393,6 +447,20 @@ export default async function ({ addon, console, msg }) {
   addon.self.addEventListener("disabled", updateColors);
   addon.self.addEventListener("reenabled", updateColors);
 
+  // inject() and overrideColours() are called when changing the Scratch theme, language, or editor mode
```
WorldLanguages, line 450: It is also called after clicking the "create new block" button, it seems.

addons/editor-theme3/theme3.js
```diff
@@ -393,6 +447,20 @@ export default async function ({ addon, console, msg }) {
...
+  const oldInject = Blockly.inject;
+  Blockly.inject = function (...args) {
+    const workspace = oldInject.call(this, ...args);
+    updateColors();
```
WorldLanguages, line 454:
```js
if (Blockly.getMainWorkspace()?.getFlyout?.()) updateColors();
```

WorldLanguages: There's some in-depth discussion about the behavior of `getMainWorkspace()` in the issue: #6212

But for this specific crash, doing nothing if `Blockly.inject()` was called because the "create new custom block" modal has been opened doesn't sound wrong. I don't think this code was actually written to try to cover that case anyways.

PR #6151, opened 2023-06-05
data-category-tweaks-v2: high contrast mode fix

mxmou:
### Changes

The "data category tweaks" addon uses Scratch's extension category system to replace the data category with separate categories for variables and lists. That didn't work in high contrast mode because Scratch tries to change the color of all extension categories and the code responsible for that only expects one category per item.

This PR fixes the issue.

### Reason for changes

To make the feature work after the Scratch update.

### Tests


Samq64 (approved):
Tested on Firefox on both versions.

DNin01 (approved):

WorldLanguages:
This could also be merged now if we want.

PR #5071, opened 2022-09-02
Change colors of extension icons based on background and block color

mxmou:
### Changes

Updates editor dark mode and customizable block colors addons to change the colors of extension icons in the category menu. The affected icons are debugger, music, video sensing, and text to speech.

PR #6152, opened 2023-06-05
Change debugger blocks in high contrast mode

### LINKED ISSUE #6123: The Debugger addon’s blocks should have a high contrast variant
Scratch is adding a high contrast block color palette. We have an addon with its own set of blocks, the Debugger blocks, but they don’t have a high contrast variant yet. They could use the color `#34E4D0` when the high contrast color mode is selected. That color is what `editor-theme3`’s High Contrast preset sets them to.

mxmou:
Resolves #6123

### Changes

Adds high contrast versions of the debugger block colors and icon.

### Tests


Samq64 (approved):
Is there a reason why the icons aren't separate files? Either way I tested on Firefox and it looks good.

DNin01 (approved):

WorldLanguages:
This can be merged for v1.33.0 if you want. If not, please wait a few days before merging.

mxmou:
> Is there a reason why the icons aren't separate files?

I don't think there was a reason when that code was originally written, but it's much easier for other addons to make changes to the icon (#5071) if it's a data URI.

PR #6157, opened 2023-06-06
Add `SET_THEME` to `reduxEvents` list in some addons

mxmou:
Fixes two bugs that happened when switching theme:
* `hide-flyout` stopped working.
* The `zebra-striping` bug happened when a reporter inside another block from the same category was being replaced:
image
When doing this after changing the theme, the replacement glow wouldn't appear.

### Tests


WorldLanguages (approved):

WorldLanguages:
Does it hurt to merge for v1.33.1?

PR #6166, opened 2023-06-07
Add dynamic disable to true-youtube-links

mybearworld:
Resolves part of #1673

### Changes

I made `true-youtube-links` dynamically disablable.

### Reason for changes

#1673

### Tests

Tested on Edge


WorldLanguages:
Is it worth it to add dynamicDisable to this addon?

addons/true-youtube-links/userscript.js
```diff
@@ -9,6 +16,10 @@
         },
       }
     );
+    if (addon.self.disabled) {
+      continue;
+    }
```
WorldLanguages, line 21: Technically you shouldn't skip the link here, you should add it to a list and handle it later if the addon gets reenabled. Which makes this code extra ugly.
dynamicDisable support = disabling support + reenabled support

WorldLanguages: So that means this loop will be continuously running as long as the addon is disabled in the session? 🤔

mybearworld: Now that you say that, this should probably stop and restart with disable/enable, I'll do that later

BroJac5246:
> Is it worth it to add dynamicDisable to this addon?

I don't think so. It would be confusing for links to change if the page hasn't reloaded.

mxmou:
> > Is it worth it to add dynamicDisable to this addon?
> 
> I don't think so. It would be confusing for links to change if the page hasn't reloaded.

The addon already supports dynamic enable.

addons/true-youtube-links/userscript.js
mybearworld: Would this be better?

WorldLanguages: It's still not ideal, as you could still have two active while loops at the same time.

mybearworld: That shouldn't have too much impact, unless you'd disable/enable the addon a lot of times.

WorldLanguages: It can happen by reenabling just once

mybearworld: Yes, but that shouldn't be too bad. There's not really a way to prevent it.

WorldLanguages: Probably shouldn't be too bad, but there's always a way.

TheColaber:
the way this was done is very buggy. the waitforelement should only be ran once

WorldLanguages:
Superseded by #6476

PR #6168, opened 2023-06-07
Properly support declarativeNetRequest in Firefox

### LINKED ISSUE #6167: Firefox 113+ partially supports declarativeNetRequest
### Description

Users on Firefox 113+ can't post comments from Scratch Messaging nor mark their messages as read. This is because URLs that end with `?sareferer` have lost their special treatment (background/handle-fetch.js)
**Edit: this didn't affect users that installed SA from the Firefox Addons store as we don't ship the declarativeNetRequestWithHostAccess permission as part of our [Firefox manifest](https://github.com/ScratchAddons/ScratchAddons/blob/master/.github/gen-manifest.mjs#L2).**

First column is Chrome, second column is Edge, third column is Firefox:
image

(source: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/declarativeNetRequest)

### Steps to reproduce

Inspect the extension background page on Firefox.

image

`domains` is deprecated, but the alternative `initiatorDomains` is only supported since Chrome 101+

### Environment

Firefox 113+
Scratch Addons v1.32.1 or master branch

WorldLanguages:
Resolves #6167

### Changes

- Use `initiatorDomains` instead of `domains` in Firefox. We can also do this in Chrome if we decide to support only Chrome 101 and newer.
- Use `new URL(chrome.runtime.getURL("")).host` instead of `chrome.runtime.id` for Firefox support. Firefox considers `griffpatch@griffpatch.co.uk` to be the extension ID, and all we care about here is the hostname when looking at the URL of our background page. 

### Tests

Everything appears to work fine in both Chrome and Firefox after this change: posting comments from Scratch Messaging, deleting comments, etc.
Be careful when testing: browsers may use old rulesets if there was an error while adding new ones. Run `chrome.declarativeNetRequest.updateDynamicRules({removeRuleIds: [1]})` to be sure.

### DIFF
--- a/background/handle-fetch.js
+++ b/background/handle-fetch.js
@@ -1,5 +1,15 @@
 if (chrome.declarativeNetRequest) {
   // DNR available, use this instead
+
+  const baseUrl = new URL(chrome.runtime.getURL(""));
+  const initiatorDomains = {
+    // `domains` is deprecated and wasn't implemented by Firefox, but it's the only
+    // option for Chrome 96-100 support.
+    [baseUrl.protocol === "moz-extension:" ? "initiatorDomains" : "domains"]: [baseUrl.host],
+    // `baseUrl.host` is equivalent to `chrome.runtime.id` in Chrome. In Firefox, it's
+    // the extension id that is used in URLs (not the one specified in manifest.json)
+  };
+
   chrome.declarativeNetRequest.updateDynamicRules({
     removeRuleIds: [1],
     addRules: [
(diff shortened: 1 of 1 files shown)


apple502j (approved):

Samq64 (approved):
Scratch messaging works as expected on both Firefox and Chrome.

WorldLanguages:
Note that DNR won't be used in Firefox releases until we update gen-manifest.mjs, so this only affects local development for now.

PR #6182, opened 2023-06-10
Fix `addon.tab.scratchClass` regression breaking folders addon

WorldLanguages:
### Changes

Resolves https://github.com/ScratchAddons/ScratchAddons/pull/6137#issuecomment-1584936150
Userscripts that call `addon.tab.createEditorContextMenu()` and that run before complete need to wait until scratchClass is ready.
It was reported that `move-to-top-bottom` was broken, but it was a byproduct of the "sprite folders" addon breaking.

### Tests

Hopefully adding this `await` doesn't break anything (it's now the first one in the file). We can use `.then()` instead in that case.

### DIFF
--- a/addons/folders/userscript.js
+++ b/addons/folders/userscript.js
@@ -708,6 +708,7 @@ export default async function ({ addon, console, msg }) {
     });
   };
 
+  await addon.tab.scratchClassReady();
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Apparently, there's no concrete reason for why the folders addon doesn't support dynamicEnable.

PR #6200, opened 2023-06-12
New Addon: Username warning remover

### LINKED ISSUE #6199: Addon: Username block warning remover
### Why this would be helpful

Many scratch projects use this block and e.g. for me it is not important if I go in a project with username blocks

### How this addon works

This addon removes the warning in scratch projects

### Possible alternatives

_No response_

### Additional context

[YouTube Video](https://www.youtube.com/watch?v=iuoGk8BtGRk)

Onako2:
Resolves #6199

### Changes

Remove the warning when going in projects that contain the username block. 
[Video on YouTube](https://www.youtube.com/watch?v=iuoGk8BtGRk)

### Reason for changes

Many projects have this block. I usually ignore these warnings however it may be for some users a bit disruptive

### Tests

Works perfectly fine! [Video](https://www.youtube.com/watch?v=iuoGk8BtGRk)
In that video Chrome is used. In Firefox it also works!


WorldLanguages:
Everything looking great @Onako2, we'll just discuss the addon idea a bit (in the issue) and decide how the addon will be named.

CST1229:
This addon also (likely unintentionally) hides the video sensing alert:
image
(https://scratch.mit.edu/projects/764400055/)

addons/element-remover/addon.json
```diff
@@ -0,0 +1,53 @@
...
+  "versionAdded": "1.33.0",
+  "settings": [
+    {
+      "name": "Remove username Sensing block warning",
+      "id": "username",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "name": "remove publish project banner on project page",
```
DNin01, line 47: We might want the terms to be consistent with Scratch.
```suggestion
      "name": "Remove share project banner on project page",
```

addons/element-remover/addon.json
```diff
@@ -0,0 +1,53 @@
...
+  ],
+  "credits": [
+    {
+      "name": "Onako2",
+      "link": "https://scratch.mit.edu/users/Onako2"
+    }
+  ],
+  "latestUpdate": {
+    "version": "1.0.0"
+  },
```
DNin01, line 37: I don't think it's necessary to insert latestUpdate in the first iteration of the addon.

addons/element-remover/addon.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "Element remover",
+  "description": "Remove elements inside scratch",
```
DNin01, line 3: This is a pretty general addon name. It sounds like it would be used for something with many options. We may want to rename it to something more specific to the features we're offering, or to keep it general-ish, something like "remove distractions".

Onako2: > This is a pretty general addon name. It sounds like it would be used for something with many options. We may want to rename it to something more specific to the features we're offering, or to keep it general-ish, something like "remove distractions".

Good idea. I just did it as a temporary name. I'll rename it after school 😀

Onako2:
> This addon also (likely unintentionally) hides the video sensing+cloud variables alert: image (https://scratch.mit.edu/projects/764400055/)

Thanks for reporting. Seems like I'll need to do it with javascript somehow 😟

mybearworld:
> > As this is a userstyle, this could probably pretty easily be disabled and enabled dynamically.
> 
> I know. But warning about camera and cloud vars is useful because some users may be confused when the project doesn't work

No, that's not what I meant - I'm talking about this: https://scratchaddons.com/docs/reference/addon-manifest/#dynamicdisable

Onako2:
> > > As this is a userstyle, this could probably pretty easily be disabled and enabled dynamically.
> > 
> > 
> > I know. But warning about camera and cloud vars is useful because some users may be confused when the project doesn't work
> 
> No, that's not what I meant - I'm talking about this: https://scratchaddons.com/docs/reference/addon-manifest/#dynamicdisable

I don't understand what dynamic disable is. Could you explain it in simple terms?

Samq64:
> I don't understand what dynamic disable is. Could you explain it in simple terms?

It makes it so your addon can be turned off without reloading the page.

Issue #6212, opened 2023-06-16
Blockly#getMainWorkspace is unreliable (Customizable block colors causes crash when making custom block)


callumjt:
### Description

when you click on "create block" the editor crashes and you have to reload
![image](https://github.com/ScratchAddons/ScratchAddons/assets/116295240/f165f566-2ccc-403b-b0ab-6e18a4ae7a6c)

### Steps to reproduce

1.turn on Customizable block colors
2.try to create a custom block

### Environment

Browser: firefox 114.0.1
Scratch Addons version: 1.33
Operating system: windows 11

DNin01:
Can reproduce.

medlans:
Can replicate.
Browser: Chrome 114.0
Scratch Addons version: 1.33 (it didn't happen before this version)
Operating system: MacOS X Macintosh 11.3
![flash](https://github.com/ScratchAddons/ScratchAddons/assets/124630024/dca58d8e-b7c8-434d-b922-4731650590f1)
I turned that addon off and was able to create one:
![off](https://github.com/ScratchAddons/ScratchAddons/assets/124630024/c7a8a544-80ff-467d-8591-0611bcc79f6e)

Samq64:
This looks like an `editor-theme3` crash.
```
gui Unhandled Error with action='Blocks': TypeError: Cannot read properties of null (reading 'getWorkspace')    at updateColors (chrome-extension://ahjndmkmpohijgkaglnfmmiiflfpkcll/addons/editor-theme3/theme3.js:439:36) ...
```

apple502j:
This affects most, if not all, Blockly addons.

apple502j:
This basically breaks all addons that use `Blockly#getMainWorkspace` and `Blockly#mainWorkspace`. After a definition prompt is opened, it returns the prompt's detached workspace, instead of the actual workspace. A workspace refresh, triggered by going to and back from project page, resets it to actual one.

This bug has existed since the very early versions - but it only became apparent with this crash.

The solution is to query the workspace not from `Blockly`, but from the React state where we grab the `ScratchBlocks` from. However, since multiple instances may exist during an addon's life, we cannot store this in a const, instead querying on every occasion. While I believe the performance cost is low, not a easy patch when this affects dozens of addons.

@towerofnix @GarboMuffin  Is my analysis correct?

GarboMuffin:
Yes, that matches some prior observations I've had

[...]

PR #6216, opened 2023-06-17
Revert "editor-theme3: compatibility with Scratch high contrast theme"

apple502j:
Reverts ScratchAddons/ScratchAddons#6146
See #6212 

The `updateColors` method is broken. It should not be called within overwritten method. Since this needs to be fixed ASAP I am just reverting this.


mxmou (approved):

WorldLanguages:
Merging, but I'll quickly double-check that this actually fixes the bug that users are experiencing.

WorldLanguages:
> The `updateColors` method is broken. It should not be called within overwritten method. Since this needs to be fixed ASAP I am just reverting this.

Is it good enough to avoid calling `updateColors()` if we know it's going to fail because Blockly.inject() is being called as a side-effect of the user wanting to create a new block? (and not because the language changed) https://github.com/ScratchAddons/ScratchAddons/pull/6146#pullrequestreview-1489472913

Issue #3906, opened 2021-12-10
[ Addon Bug : Auto Hide Block Palette ] Blocks get deleted when dragged onto palette area.


teahwiz:
### Description

When blocks are dragged over the collapsed palette area in "Category hover mode", the blocks will be deleted. Opening and closing the palette fixes this issue temporarily, Mousing over the category icons re-initiates this issue.

### Repoduction steps

1. Mouse over the category icons from the coding area towards the stage.
2. Drag blocks over the collapsed palette area.

### Expected behaviour

Blocks should not be deleted

### Actual behaviour

Blocks are deleted

### Environment

- Browser: Turbowarp
- Scratch Addons version: 55ac5a5
- Operating system: Void Linux

### Additional context

A addition to fixing this issue could be to allow blocks to be deleted when dropped over the palette category icons, while the palette only expands if not dragging a block.

mxmou:
More detailed reproduction steps from #6224:

> 1. Have the Auto-hiding block palette addon enabled with the category hover or category click setting on
> 2. Have a block in the canvas off to the right to drag into the palette area
> 3. Make a new custom block (UPDATE: You don't even need to make the custom block, if you press cancel instead of ok and continue as normal it will still cause the bug to occur), and if you chose category hover the palette would have collapsed already, if you chose category click, click the “my blocks” category to hide the block palette
> 4. Try to drag a block (I recommend holding alt to make a duplicate block to be able to test repeatedly easily) into the area where the block palette should be but isn't. It should delete the block, even though you are dragging it over canvas space.
> 5. If you chose category hover, hover over the category icons then back over the block canvas, or if you chose category click, click on the “my blocks” category twice, and you will find that blocks no longer delete when you drag them into the area of the canvas where the block palette would normally be (in case it's not clear, this is what should happen always) UPDATE: You need to move a block anywhere on the canvas before opening and closing the palette, just doing it will not work

WorldLanguages:
+1 from Discord server

WorldLanguages:
Possibly related to #6212? As you have to make a custom block.
See also https://github.com/ScratchAddons/ScratchAddons/pull/6227#issuecomment-1599952073

WorldLanguages:
This should be resolved by #6613

Issue #6224, opened 2023-06-18
Blocks delete weirdly with auto hiding block palette


Jazza-231:
### Description

What's happening: After making a custom block and collapsing the block palette, dragging blocks into the space where the palette normally is deletes them, even though the palette is collapsed.
What should happen: Blocks only delete when dragged over the icons while the palette is collapsed.

See below for more details, and the original bug report: [SA discord server](https://discord.com/channels/806602307750985799/1116706384180936714)

### Steps to reproduce

1. Have the Auto-hiding block palette addon enabled with the category hover or category click setting on
2. Have a block in the canvas off to the right to drag into the palette area
3. Make a new custom block (UPDATE: You don't even need to make the custom block, if you press cancel instead of ok and continue as normal it will still cause the bug to occur), and if you chose category hover the palette would have collapsed already, if you chose category click, click the “my blocks” category to hide the block palette
4. Try to drag a block (I recommend holding alt to make a duplicate block to be able to test repeatedly easily) into the area where the block palette should be but isn't. It should delete the block, even though you are dragging it over canvas space.
5. If you chose category hover, hover over the category icons then back over the block canvas, or if you chose category click, click on the “my blocks” category twice, and you will find that blocks no longer delete when you drag them into the area of the canvas where the block palette would normally be (in case it's not clear, this is what should happen always) UPDATE: You need to move a block anywhere on the canvas before opening and closing the palette, just doing it will not work

### Demonstration(s) and additional context

A video demonstrating a few repetitions: [video](https://cdn.discordapp.com/attachments/1116706384180936714/1116706693183701062/Auto_hiding_block_palette_bug.mp4)

### Environment

Originally tested on turbowarp desktop 1.7.1
Browser: Brave Version 1.52.126, Chromium: 114.0.5735.133
Scratch Addons version: Still happens on 1.33.0 and whatever version TW uses right now
Operating system: Windows 11 Pro

mxmou:
Duplicate of #3906

WorldLanguages:
This should be fixed by #6613

PR #6227, opened 2023-06-18
Add an API to query Blockly workspace

### LINKED ISSUE #6212: Blockly#getMainWorkspace is unreliable (Customizable block colors causes crash when making custom block)

apple502j:
Resolves #6212 

TODO:

- [x] actually make the addons use this API
- [ ] ensure they are not stored addon-scope, and instead queried per use
- [ ] ensure they are not queried outside editor
- [ ] test

### DIFF
--- a/addon-api/content-script/Trap.js
+++ b/addon-api/content-script/Trap.js
@@ -31,15 +31,31 @@ export default class Trap extends Listenable {
     return "__reactInternalInstance$";
   }
 
-  /**
-   * Gets Blockly instance actually used by Scratch.
-   * This is different from window.Blockly.
-   * @async
-   * @throws when on non-project page.
-   * @returns {Promise<object>}
-   */
-  async getBlockly() {
-    if (this._cache.Blockly) return this._cache.Blockly;
+  _getBlocksComponent(wrapper) {
+    if (!this._react_internal_key) {
+      this._react_internal_key = Object.keys(wrapper).find((key) => key.startsWith(this.REACT_INTERNAL_PREFIX));
+    }
+    const internal = wrapper[this._react_internal_key];
+    let childable = internal;
+    /* eslint-disable no-empty */
+    while (((childable = childable.child), !childable || !childable.stateNode || !childable.stateNode.ScratchBlocks)) {}
+    /* eslint-enable no-empty */
+    return childable;
+  }
+
+  _getBlocksWrapperComponentSync() {
+    const editorMode = this._getEditorMode();
+    if (!editorMode || editorMode === "embed")
+      throw new Error(`Cannot access Blockly on ${editorMode} page (${location.pathname})`);
+    const BLOCKS_CLASS = '[class^="gui_blocks-wrapper"]';
+    let elem = document.querySelector(BLOCKS_CLASS);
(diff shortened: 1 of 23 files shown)


WorldLanguages:
> TODO: many addons call this method in settings, dynamic enable/disable callback, etc. This is illegal; workspace operation must be done only when the workspace is visible.
> 
> How should we fix all those criminals?

Maybe we should add `if(workspace)` every time? This is already the case on data-category-tweaks-v2 for example: https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/data-category-tweaks-v2/userscript.js#L192-L223

WorldLanguages:
We probably should not change the behavior of any addons except editor-theme3 for the v1.33.2 release. This means that if we do end up adding a new API for this, only editor-theme3 should be changed to use it.

As far as I've experienced, changing the language and changing the theme color (which is only possible if editor-theme3 is disabled) have the same side-effects, so we may get more bug reports than normal - but most bugs would have existed for a while. Changing the theme color will just be more frequent than changing the language inside the editor, but will probably not introduce any new bugs (after some reduxEvents updates and other PRs that were released for v1.33.0)

Related: https://github.com/ScratchAddons/ScratchAddons/pull/6216#issuecomment-1599939884

WorldLanguages:
As far as I've seen, the `window.Blockly.getWorkspace()` method:
- May possibly return `null` already. Most addons already consider this case.
- Returns a different workspace after changing the language or changing the theme color.
- Returns a different workspace each time the user clicks "see inside".
- Returns the wrong workspace after the user clicks "make a block" in the block pallete and until one of the two conditions above happen. 

So if we create our own dedicated method to get the workspace, we would probably make these decisions:
- It can still return `null`, of course.
- It always returns the intended, most recent workspace. This is already the case with `window.Blockly.getWorkspace()` with one exception: our method should never return the "make new block" modal workspace. It should always return the most recent workspace that is not connected to the "new block creation" modal.
- We could expose an event that addons can listen to, `"workspaceChanged"`. It would not trigger if the "main" workspace has changed as a result of opening the "new block" modal. Addon `editor-theme3` already attempted to do this by trapping `Blockly.inject`, which returns workspace objects and internally calls `Blockly.createMainWorkspace_`. I suggested a way to exclude the unintended "new block workspace" [here](https://github.com/ScratchAddons/ScratchAddons/pull/6146#discussion_r1236169094).

WorldLanguages:
Doesn't feel necessary to implement this by v1.33.x, because #6247 was merged.

WorldLanguages:
Using this new API on hide-flyout (auto-hide block palette) would probably fix #3906

WorldLanguages:
apple502j requested to postpone this to next milestone

WorldLanguages:
This blocks: #3906 (see also #6224) and #5917

WorldLanguages:
PR replaced by #6613

PR #6229, opened 2023-06-18
Add option to disable-stage-drag-select

mr-mph:
### Changes

Added a new setting in the _Non-draggable sprites in editor_ addon that allows the user to enable dragging while the project is stopped.

### Reason for changes

Gives users more options for customization.

### Tests

Tested in Chrome 114.0.5735.106

### DIFF
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -8,6 +8,10 @@
     },
     {
       "name": "GarboMuffin"
+    },
+    {
+      "name": "Mr_MPH",
+      "link": "https://scratch.mit.edu/users/Mr_MPH/"
     }
   ],
   "userscripts": [
--- a/addons/disable-stage-drag-select/addon.json
+++ b/addons/disable-stage-drag-select/addon.json
@@ -16,6 +16,14 @@
       "matches": ["projects"]
     }
   ],
+  "settings": [
+    {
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Note: this lets you drag sprites while the project is paused, too. That's what should happen I think.

WorldLanguages (approved):

addons/disable-stage-drag-select/addon.json
```diff
@@ -16,9 +20,21 @@
       "matches": ["projects"]
     }
   ],
+  "settings": [
+    {
+      "name": "Allow dragging while project is stopped",
+      "id": "drag_while_stopped",
+      "type": "boolean",
+      "default": false
```
Samq64, line 28: I think this should be true by default, but this is fine too.

WorldLanguages:
cc @Joeclinton1 @GarboMuffin

addons/disable-stage-drag-select/addon.json
WorldLanguages: The ideal default setting vs. backwards compatibility dilemma...

WorldLanguages: I think keeping the default as off is fine since you can hold Shift to drag, no matter if the project is running or not!

Samq64 (approved):

PR #6240, opened 2023-06-19
Increase category.sel and switch contrast

DNin01:
A step towards #5616

### Changes

Made some changes to the style of the selected category (on the left side pane). The text is much easier to read because it's black/white instead of orange.

Changed the design of the setting switches, specifically, the one set to off:
Two pictures each with a filled-in switch set to on and an outlined switch set to off, one picture is light mode and one is dark mode
This is great because the background of the switch set to off in dark mode was almost impossible to make out.

### Reason for changes

To make things easier to see.

### Tests

Tested in Edge 114. RTL direction considered and tested as needed.


WorldLanguages:
Here's a screenshot (dark mode):
image

WorldLanguages:
Looks good to me, not sure if someone needs to be reviewed in detail here.

Samq64 (approved):

webpages/settings/components/category-selector.html
```diff
@@ -40,7 +40,6 @@
   }
 
   .category.sel {
-    color: var(--orange);
     font-weight: var(--brand-orange-min-font-weight);
```
mxmou, line 43:
```suggestion
    font-weight: 600;
```
This CSS variable specifies the font weight of orange text. It shouldn't be used here.

webpages/styles/colors-light.css
```diff
@@ -11,8 +10,8 @@
   --button-hover-background: #d4d3d3;
   --input-background: #fff;
   --search-input-background: #f0f0f0;
-  --switch-background: #888;
-  --switch-inner-background: #fff;
+  --switch-background: #555;
```
mxmou, line 13:
```suggestion
  --switch-border: #555;
```

webpages/styles/colors.css
```diff
@@ -1,6 +1,6 @@
 :root {
   --brand-orange: #ff7b26;
-  --brand-orange-min-font-weight: normal;
+  --brand-orange-min-font-weight: 600;
```
mxmou, line 3: This changes the links at the bottom of the Messaging popup. Was that intended?

webpages/styles/colors.css
```diff
@@ -47,7 +47,7 @@
   --modal-overlay: rgba(0, 0, 0, 0.65);
   --hover-highlight: rgba(255, 255, 255, 0.05);
   --hover-highlight-reduced: rgba(255, 255, 255, 0.02);
-  --hover-darken: rgba(0, 0, 0, 0.1);
+  --hover-darken: rgba(0, 0, 0, 0.2);
```
mxmou, line 50: I don't think this change is necessary. The color doesn't have much contrast with the background either way, and it doesn't need to - the orange border and the font weight change are enough to indicate selection.

mxmou (changes requested):

mxmou:
I think this design makes switches stand out too much. Something like this would look better and still have enough contrast (3:1 is the WCAG requirement for user interface components):
image
image
* The border color is `#808080` in both light and dark mode.
* The border width is 1px like most other elements (cards, buttons, inputs).
* The background-color is `--button-background`.

webpages/styles/colors-light.css
```diff
@@ -1,5 +1,4 @@
 :root {
-  --brand-orange-min-font-weight: 600;
```
mxmou, line 2: This can now be readded.

mxmou (approved):

webpages/styles/colors-light.css
DNin01: Okay, but why is `--brand-orange-min-font-weight` 600 in light and `normal` in dark? For me, it makes certain text bold only when I'm in light mode. That's why I removed this line initially.

mxmou: The variable is used for these links:
Mark messages as read | Open messages page

They become orange on hover, which makes them hard to read in light mode. The bold font makes it slightly better. It would be even better to use a different color (probably `--blue-text`), but that should be a separate PR.

DNin01:
Ready to merge.

Samq64 (approved):

DNin01:
@WorldLanguages We can probably merge by now, right?

WorldLanguages:
> @WorldLanguages We can probably merge by now, right?

Yep, I may merge it in a few days depending on how other settings page PRs progress.

DNin01:
It's been a while...

Samq64:
> It's been a while...

I added the pending tag because I assumed #6272 was going to be merged soon and I didn't want to cause more conflicts but it's still not merged and the settings page can't stay frozen for ever.

DNin01:
We ready yet, @WorldLanguages? Why is this being held back?

WorldLanguages:
Yeah let's merge this

PR #6242, opened 2023-06-20
better-img-uploads (HD image uploads) bug fixes

### LINKED ISSUE #6238: Minor tweaks to better-img-uploads
### Why this would be helpful

1. One of the changes fixes uploading an image with the exact same ratio as the scratch stage importing it 1 pixel smaller than the stage (e.g. uploading a 480x360 image would import it as 479x359, which annoyed me for ages)
2. Another change allows .sprite3 files to be uploaded with the hd image upload button. This isn't a big issue in itself, but consequently, this means that if you use the drag-and-drop addon with the hd image upload option enabled, you can now drag sprite3 files into the sprite pane (which was also something that annoyed me for ages)

### My suggestion

See above, but specifically all this does is add some special logic for sprite3 files and change 1 number.

### Possible alternatives

These are such small changes I don't think there's any other way to do it (unless I committed coding sins with the way I implemented them, oops)

### Additional context

A before (in turbowarp) and after (on scratch) for uploading a 960x720 image and a sprite file.
https://github.com/ScratchAddons/ScratchAddons/assets/81956724/02043519-adb9-4635-a546-6b77d2351409

Also, I know these are tiny tiny changes and arent really necessary so it makes sense if this is gonna get rejected for being too small.

Jazza-231:
Resolves  #6238 

### Changes

Added support for importing .sprite3 files (they are accepted but passed through with no changes), and changed 1 number in the logic for image sizing from 479 to 480 to stop it resizing certain things incorrectly.

### Reason for changes

Cuz they were annoying me and mxmou and WL said they were good 👍

### Tests

Tested in Brave Version 1.52.126, Chromium version 114.0.5735.133

### DIFF
--- a/addons/better-img-uploads/userscript.js
+++ b/addons/better-img-uploads/userscript.js
@@ -110,7 +110,7 @@ export default async function ({ addon, console, msg }) {
     let processed = new Array();
 
     for (let file of files) {
-      if (file.type.endsWith("/svg") || file.name.endsWith(".sprite3")) {
+      if (file.type.endsWith("/svg+xml") || file.name.endsWith(".sprite3")) {
         // The file is either an SVG or a sprite3 file, we should not change it...
         processed.push(file);
         continue;
--- a/addons/better-img-uploads/userscript.js
+++ b/addons/better-img-uploads/userscript.js
@@ -25,7 +25,7 @@ export default async function ({ addon, console, msg }) {
     });
     button.append(img);
     const input = Object.assign(document.createElement("input"), {
-      accept: ".svg, .png, .bmp, .jpg, .jpeg, .sprite3",
+      accept: ".svg, .png, .bmp, .jpg, .jpeg, .sprite2, .sprite3",
       className: `${addon.tab.scratchClass(
         "action-menu_file-input" /* TODO: when adding dynamicDisable, ensure compat with drag-drop */
       )} sa-better-img-uploads-input`,
@@ -110,8 +110,15 @@ export default async function ({ addon, console, msg }) {
     let processed = new Array();
 
     for (let file of files) {
-      if (file.type.endsWith("/svg+xml") || file.name.endsWith(".sprite3")) {
-        // The file is either an SVG or a sprite3 file, we should not change it...
+      if (
+        !(
+          file.name.endsWith(".png") ||
+          file.name.endsWith(".bmp") ||
+          file.name.endsWith(".jpg") ||
+          file.name.endsWith(".jpeg")
+        )
+      ) {
+        // The file is not in the whitelist, so we should ignore it, and let scratch deal with it..
         processed.push(file);
         continue;
       }
--- a/addons/better-img-uploads/userscript.js
+++ b/addons/better-img-uploads/userscript.js
@@ -25,7 +25,7 @@ export default async function ({ addon, console, msg }) {
     });
     button.append(img);
     const input = Object.assign(document.createElement("input"), {
-      accept: ".svg, .png, .bmp, .jpg, .jpeg",
+      accept: ".svg, .png, .bmp, .jpg, .jpeg, .sprite3",
       className: `${addon.tab.scratchClass(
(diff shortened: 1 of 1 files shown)


addons/better-img-uploads/userscript.js
```diff
@@ -110,8 +110,8 @@ export default async function ({ addon, console, msg }) {
     let processed = new Array();
 
     for (let file of files) {
-      if (file.type.includes("svg")) {
-        //The file is already a svg, we should not change it...
+      if (file.type.endsWith("/svg") || file.name.endsWith(".sprite3")) {
```
Samq64, line 113:
```suggestion
      if (file.type.endsWith("/svg+xml") || file.name.endsWith(".sprite3")) {
```

WorldLanguages: Yeah, let's be careful with the MIME types here.

addons/better-img-uploads/userscript.js
```diff
@@ -150,7 +150,7 @@ export default async function ({ addon, console, msg }) {
       } //Otherwise just leave the image the same size
 
       function getResizedWidthHeight(oldWidth, oldHeight) {
-        const STAGE_WIDTH = 479;
+        const STAGE_WIDTH = 480;
```
WorldLanguages, line 153: This was initially changed to 479 by commit https://github.com/ErrorGamer2000/ScratchAddons/commit/6da9ddc95ba8d999014dfe781b7bdbbd43dc480f#diff-854cc3c818d6760de96084d1f214859d4919296db14211aa9204b9c5b218cd4cR114

It was probably unintentional, it's based on vanilla Scratch code which uses 480 of course: https://github.com/scratchfoundation/scratch-svg-renderer/blob/a4ad7080761a9dba8c5ffadad703cfd07f62841a/src/bitmap-adapter.js#L70

Jazza-231:
Realive because i rly want something of mine to be merged lmao

Samq64 (approved):

Jazza-231:
image
Thoughts on adding a temporary notice? (even though the changes weren't to the drap and drop addon)

WorldLanguages:
@Jazza-231 I don't think a notice is necessary.

addons/better-img-uploads/userscript.js
```diff
@@ -25,7 +25,7 @@ export default async function ({ addon, console, msg }) {
     });
     button.append(img);
     const input = Object.assign(document.createElement("input"), {
-      accept: ".svg, .png, .bmp, .jpg, .jpeg",
+      accept: ".svg, .png, .bmp, .jpg, .jpeg, .sprite3",
```
WorldLanguages, line 28: https://github.com/scratchfoundation/scratch-gui/blob/a5bbc4c488aa530fa89cb24a0eecc07cd40f5d57/src/components/sprite-selector/sprite-selector.jsx#L123

WorldLanguages: We should add `.sprite2` here

WorldLanguages (approved):

addons/better-img-uploads/userscript.js
```diff
@@ -25,7 +25,7 @@ export default async function ({ addon, console, msg }) {
     });
     button.append(img);
     const input = Object.assign(document.createElement("input"), {
-      accept: ".svg, .png, .bmp, .jpg, .jpeg",
+      accept: ".svg, .png, .bmp, .jpg, .jpeg, .sprite2, .sprite3",
```
WorldLanguages, line 28: Note that this didn't fix a bug per-se, but it tells the OS file picker to accept sprite2 and sprite3 file uploads.

WorldLanguages:
Feedback: apparently, prior to this change, the HD uploads addon (better-img-uploads) had the side-effect of supporting webp image uploads.

Related: #6666

BroJac5246:
> Feedback: apparently, prior to this change, the HD uploads addon (better-img-uploads) had the side-effect of supporting webp image uploads.
> 
> Related: #6666

This reminds me of https://github.com/ScratchAddons/ScratchAddons/issues/6515#issuecomment-1681451448 as well.

PR #6252, opened 2023-06-22
Use setAttribute to set ARIA attributes

mxmou:
### Changes

Changes code that sets `role` and `aria-*` attributes to use `setAttribute()` because `.role` and `.aria*` don't work on Firefox.

### Reason for changes

Some things didn't work correctly on Firefox:
* The tab added by `search-profile` was read as "button" instead of "tab, not selected" by screen readers.
* In the beta version of scratch-gui, the hide stage button didn't appear selected when the stage was hidden.

### Tests

Tested on Edge and Firefox. The `hide-stage` change was tested by running scratch-www locally.


WorldLanguages:
> because `.role` and `.aria*` don't work on Firefox.

Huh. Seems like an obvious webcompat issue.

WorldLanguages (approved):

apple502j (approved):

PR #6253, opened 2023-06-22
Fix debugger and gamepad buttons in full screen

mxmou:
### Changes

Fixes a bug that happens in the beta version of scratch-gui. When debugger and gamepad are enabled, full screen mode looks like this:
image

This happens because React reuses the `stage-header_stage-size-row` when entering full screen mode and changes its class name instead of removing it from the DOM and inserting a new element.

### Reason for changes

To make addons work correctly when the update is released.

### Tests

Tested in both browsers, locally and on scratch.mit.edu.

### DIFF
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -625,18 +625,22 @@ export default async function ({ addon, console, msg }) {
   };
 
   while (true) {
-    await addon.tab.waitForElement('[class*="stage-header_stage-size-row"]', {
-      markAsSeen: true,
-      reduxEvents: [
-        "scratch-gui/mode/SET_PLAYER",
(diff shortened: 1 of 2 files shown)


WorldLanguages:
I'm seeing this on https://scratchfoundation.github.io/scratch-gui/develop/ sometimes:

image

mxmou:
> I'm seeing this on https://scratchfoundation.github.io/scratch-gui/develop/ sometimes:
> 
> image

That probably happens because `editorMode` is always `"editor"` on `scratchfoundation.github.io`, even in full screen mode. I tested this PR locally and it worked correctly.

WorldLanguages:
Good catch

WorldLanguages (approved):

PR #6254, opened 2023-06-22
Enhance "username search" addon

valmontechno:
### Changes

Fixes the bug that if we search for a user that does not exist like griffpatch2, the button redirects to page 404.
And added a new setting in the Username search addon that allows the user to show icon of searched user.

### Reason for changes

These modifications  enhance the user experience by eliminating redirection errors and providing better visual recognition of searched users. This enables users to navigate more efficiently.

scrnli_22_06_2023 10-44-07

### Tests

Google Chrome : 114.0.5735.134
Microsoft Edge : 114.0.1823.51

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3,7 +3,7 @@
   "name": "__MSG_extensionName__",
   "description": "__MSG_extensionDescription__",
   "version": "1.33.2",
-  "version_name": "1.33.2-prerelease",
+  "version_name": "1.33.2",
   "default_locale": "en",
   "background": {
     "page": "background/background.html"


addons/search-profile/userscript.js
```diff
@@ -5,14 +5,13 @@ export default async function ({ addon, console, msg }) {
     img = tab.appendChild(document.createElement("img")),
     span = tab.appendChild(document.createElement("span")),
     user = document.querySelector('[name="q"]').value.trim(),
-    valid = /^[\w-]{3,20}$/g.test(user);
+    valid = (/^[\w-]{3,20}$/g.test(user) && (await fetch(`https://scratch.mit.edu/users/${user}/`)).status == 200);
```
mxmou, line 8: Because of this `fetch`, it takes some time for the tab to appear. I think it would be better to show it immediately, then fetch this URL and disable the button element if it fails.

addons/search-profile/style.css
```diff
@@ -1,10 +1,27 @@
-.tabs button:disabled {
+.user-tab {
+  display: grid !important;
+  align-items: end;
+  justify-items: center;
+}
```
mxmou, line 5: You can add `display: block` to the `span` instead.

addons/search-profile/style.css
```diff
@@ -1,10 +1,27 @@
...
+
+.user-tab:not(:disabled) img {
+  border-radius: 3px;
+}
```
mxmou, line 9: Dark mode currently breaks the icon.
image
You can fix that by adding `filter: none`:
```suggestion
.user-tab:not(:disabled) img {
  border-radius: 3px;
  filter: none !important;
}
```

mxmou (changes requested):

WorldLanguages:
> Fixes the bug that if we search for a user that does not exist like griffpatch2, the button redirects to page 404.

Note that this is not a "bug": ([source](https://github.com/ScratchAddons/ScratchAddons/pull/770#pullrequestreview-535659416))

image

WorldLanguages:
Why not just rely on the api.scratch.mit.edu endpoint (and only that one fetch)? There would still be some false positives, such as searching `kaj`, but it's still better than the previous behavior.

I like the idea of embedding the profile picture.

addons/search-profile/addon.json
```diff
@@ -14,10 +14,23 @@
...
+  ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "credits": [{ "name": "Geotale" }, { "name": "pufferfish101007" }, { "name": "Hexagonal" }],
+  "credits": [{ "name": "Geotale" }, { "name": "pufferfish101007" }, { "name": "Hexagonal" }, { "name": "valmontechno" }],
   "tags": ["community", "recommended", "profiles"],
   "versionAdded": "1.1.0",
+  "latestUpdate": {
+    "version": "1.33.2",
+    "temporaryNotice": "Now displays the icon and name of the searched user.",
```
RedGuy12, line 32: temporaryNotice is unnecessary.

BroJac5246:
> Why not just rely on the api.scratch.mit.edu endpoint (and only that one fetch)? There would still be some false positives, such as searching `kaj`, but it's still better than the previous behavior.
> 
> I like the idea of embedding the profile picture.

I'm not familiar with the API, but what will happen if you request the profile picture of a deleted or banned user?

Samq64:
> I'm not familiar with the API, but what will happen if you request the profile picture of a deleted or banned user?

It either returns the real one or the default.

BroJac5246:
> > I'm not familiar with the API, but what will happen if you request the profile picture of a deleted or banned user?
> 
> It either returns the real one or the default.

How is "either" determined?

CST1229:
> How is "either" determined?

I think if the user is banned (or possibly deleted), it returns the real one (which can be the default one if they never set a PFP), and if it never existed it returns the default one

valmontechno:
> Why not just rely on the api.scratch.mit.edu endpoint (and only that one fetch)? There would still be some false positives, such as searching `kaj`, but it's still better than the previous behavior.

Maybe we should first do the simple verification of the character string, then fetch the scratch 2 profile page and deactivate the button if necessary.

WorldLanguages:
> Maybe we should first do the simple verification of the character string, then fetch the scratch 2 profile page and deactivate the button if necessary.

Is it really necessary to fetch the scratch2 page?

valmontechno:
> Is it really necessary to fetch the scratch2 page?

I think we could use the [ScratchDB API](https://scratchdb.lefty.one/v3/user/info/kaj). It returns `"id":null` , `"error":"UserNotFoundError"` or `"error":"UserNotValidError"` if the user does not exist.

WorldLanguages:
Why can't we just use the normal Scratch API and forget about the edge cases such as "kaj"? It will work fine for 99% of searches.

PR #6259, opened 2023-06-24
Clarify 60 FPS breaking projects

DNin01:
### Changes

Clarifies why most projects will not behave as intended if you raise the frame rate.

### Reason for changes

These types of addons aren't usually seen in Scratch Addons, but I guess people really like having 60 FPS without having to switch over to TurboWarp. If we want to have addons like this that can break things, it should at least be exceptionally clear to the user that they can't be used everywhere. This change makes it clearer that this isn't an everyday addon.


addons/60fps/addon.json
```diff
@@ -4,7 +4,7 @@
   "info": [
     {
       "type": "notice",
```
DNin01, line 6: I'd even understand if this was a warning (though the warnings are pretty distracting).

addons/60fps/addon.json
```diff
@@ -4,7 +4,7 @@
...
-      "text": "Projects will run twice as fast at 60FPS, because they typically run at 30FPS.",
+      "text": "Many projects won't behave properly when running at 60 FPS, because increasing the frame rate also increases the speed at which scripts run.",
```
Samq64, line 7:
```suggestion
      "text": "Many projects won't behave properly when running at 60 FPS, because increasing the frame rate also increases the speed at which most scripts run.",
```

Samq64 (approved):

Samq64:
I also just noticed there's a space missing in the setting name. I'm not sure if it's intentional, if it isn't I guess it could be fixed here since this PR only changes this file.

DNin01:
> I also just noticed there's a space missing in the setting name. I'm not sure if it's intentional, if it isn't I guess it could be fixed here since this PR only changes this file.

It looks better to me that way, but it's inconsistent with the name.

DNin01:
What was that all about?

CST1229:
> It just modified 135 files for no reason?

image

Samq64:
> > I also just noticed there's a space missing in the setting name. I'm not sure if it's intentional, if it isn't I guess it could be fixed here since this PR only changes this file.
> 
> It looks better to me with a space, but it's inconsistent with the name.

I meant `Alt+GreenFlag FPS`.

DNin01:
@WorldLanguages Are we ready to merge?

PR #6271, opened 2023-06-25
Rename hide-signatures

DNin01:
I'm not totally sure why it had to be "Hide signatures _on_ forums".

### DIFF
--- a/addons/hide-signatures/addon.json
+++ b/addons/hide-signatures/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Hide signatures on forums",
+  "name": "Hide forum signatures",
   "description": "Hides signatures in forum posts and adds a \"Show signature\" button.",
   "credits": [
     {


WorldLanguages:
Looks like the addon author initially named it `Hide signatures in forums` (pretty much the same): https://github.com/FunctionalMetatable/ScratchAddons/commit/1b376084e2a14a5ee4126fb80f20573f00ca9fc4#diff-71564b07538b4c97f01941d973083a8cd693c5e8881bc3f3480e9a1a00ca42b5R2

mxmou:
"Hide forum signatures" still isn't a noun. "Hidden forum signatures" or "Forum signature hiding" would be, but in this case it's more important that the name is easy to understand.

BroJac5246:
> "Hide forum signatures" still isn't a noun. "Hidden forum signatures" or "Forum signature hiding" would be, but in this case it's more important that the name is easy to understand.

I agree, it's more important to be concise.

BroJac5246 (approved):

WorldLanguages:
@mxmou To clarify, you aren't opposed to this change, right? In that case, I would go ahead with merging this.

WorldLanguages:
If #6009 is implemented, we could rename this to simply `Hide signatures`.

mxmou:
> @mxmou To clarify, you aren't opposed to this change, right? In that case, I would go ahead with merging this.

I like the change, I just didn't understand the original explanation.

WorldLanguages (approved):

WorldLanguages:
How on earth did this break Prettier?

PR #6272, opened 2023-06-25
Vue3 for webpages (new)

### LINKED ISSUE #2531: Settings page takes way too long to load

GrahamSH-LLK:
Factor in  #5190
Resolves #2531 (in my testing)

### Changes

This is a large change.
This change rewrites the settings page, license page, and popup in Vue 3, compiled with vite. 
### Reason for changes

Loads nearly instantly
We aren't using _vue 1_
Single js/css file
### Tests

Tested on chrome and firefox


Samq64:
Previous PR: #6049

Samq64:
A couple differences things I've noticed:
- There's a space missing between addon credits and notes, and in more settings
- The reset/preset dropdown next to addon settings is wider on the Vue 3 version.

I also had missing settings inputs once, but I wasn't able to repro that bug again.

BackThePortal:
> There's a space missing before addon credit notes, and in the export/import description in more settings

Can't really see what you're saying, everything in credit notes and the more settings popup looks exactly the same (in Chrome), except for this space before the commas, but imo it looks better this way.
Vue 3:
image
Orginal:
image

> The reset/preset dropdown next to addon settings is wider on the Vue 3 version.
Yup, just saw that. Again, some change in style rendering order. Gonna fix it now.

Samq64 (changes requested):
Actually it looks like 1.33.3 settings can't be imported into the Vue 3 version.

WorldLanguages:
Before merging, we want to make sure:
- Everything works: toggling addons, search, changing settings, filtering, changing theme, importing/exporting settings
- The layout looks the same (vs. the Vue 1 settings page) in all screen sizes, both in light mode and dark mode, in all languages, and also in RTL

After merging this (and after 1 release), I'll probably work on some cleanup of the existing code (which remains mostly unchanged) to remove some now unnecessary Vue v1 optimizations and ugly old code.

Tagging some people that work or plan to work on the settings page so I make sure they're aware of this PR: @mxmou, @lisa-wolfgang, @TheColaber

GrahamSH-LLK:
> 

ah shit thanks

DNin01:
The Prettier config on your branch needs an update. You'll need to revert the "Format code" commit and update the branch. Temporarily disable workflows if it's getting in the way.

webpages/src/lib/serialize.js
```diff
@@ -2,6 +2,7 @@ const promisify =
   (callbackFn) =>
   (...args) =>
     new Promise((resolve) => callbackFn(...args, resolve));
+const browserLevelPermissions = ["notifications"];
```
RedGuy12, line 5: What about `"clipboardWrite"`?

Couldn't you just import this into App.vue?

WorldLanguages: Yeah, this also needs the `if (typeof browser !== "undefined")` check for clipboardWrite.

Samq64:
The addon settings still won't import and I noticed the category switching animation is gone.

webpages/dist/rtl-list.js
```diff
@@ -1,5 +1,6 @@
 const rtlList = [
   // List of language codes (xx or xx-XX)
+  "en-US",
```
WorldLanguages, line 3: Remove this. It was just for testing.

webpages/src/lib/rtl-list.js
```diff
@@ -1,5 +1,8 @@
+/* Languages for which RTL layout is enabled.
+   Only included languages should be added to this list. */
 const rtlList = [
   // List of language codes (xx or xx-XX)
+  "en-US",
```
WorldLanguages, line 5: Remove this. It was just for testing.

libraries/license-info.json
```diff
@@ -4,13 +4,16 @@
   "scratchblocks": "MIT",
   "scratch-blocks": "Apache-2.0",
   "icu-message-formatter": "Apache-2.0",
-  "vue": "MIT",
+  "vue1": "MIT",
```
Samq64, line 7: Is Vue 1 still used anywhere?

Samq64 (commented):
I've noticed some pages like permissions have a `.vue` file but also static html files in both the main webpages folder and dist.

GrahamSH-LLK:
> I've noticed some pages like permissions have a `.vue` file but also static html files in both the main webpages folder and dist.

that's intentional??

libraries/license-info.json
GrahamSH-LLK: Yes, the addon popups.

GrahamSH-LLK:
> The colour picker dark mode still needs some work. Hovering over the buttons on the right turns the background black. image

Fixed!

webpages/src/App.vue
```diff
@@ -0,0 +1,814 @@
...
+  <div class="popup" v-cloak v-show="showPopupModal">
+    <div class="label">{{ msg("settingsPagePermission", addonToEnable ? addonToEnable.name : "") }}</div>
+    <div>
+      <button class="large-button" @click="openFullSettings()">{{ msg("openFullSettings") }}</button>
+      <button class="large-button" @click="hidePopup()">{{ msg("skipOpenFullSettings") }}</button>
+    </div>
+  </div>
+</template>
+
+<script>
```
WorldLanguages, line 199: Can't this be a separate JS file?

webpages/src/App.vue
```diff
@@ -0,0 +1,814 @@
...
+          const [addonMajor, addonMinor, __] = manifest.versionAdded.split(".");
+          const excluded_1_33_0 = manifest.versionAdded === "1.33.0";
+          if (!excluded_1_33_0 && extMajor === addonMajor && extMinor === addonMinor) {
+            manifest.tags.push("new");
+            manifest._groups.push(
+              manifest.tags.includes("recommended") || manifest.tags.includes("featured") ? "featuredNew" : "new"
+            );
+          } else if (excluded_1_33_0) {
+            // Addon: op-badge
+            // TODO: remove this for v1.34.0 release.
```
WorldLanguages, line 357: This code was updated in a recent commit.

GrahamSH-LLK:
> When I tried to install Node.js, my graphics driver got corrupted.

oh yeah i forgot i installed the graphic-driver-corrupting package. npm really has everything!

Samq64 (commented):
Prettier should work correctly again assuming you're branch isn't too outdated. I'll test again to make sure all the other conflicts were fixed correctly.

WorldLanguages:
I made a small change to the `webpages/settings/index.html` file on PR #6572

WorldLanguages:
I think it's unlikely we'll merge this exact PR, as it's hard to track the changes and it has some issues.

- I believe the order of stylesheets was not kept, so that lead to layout bugs that were fixed by changing the CSS, instead of trying to keep the order intact
- There is no need to update the color picker, we can upgrade the settings page to Vue 3 but continue to use the current one, until we find one that suits best and we find ideal.

Now that we have the experience of looking at a working implementation of the settings page built on top of Vue 3, it will be quite easier to rebuild it again with some better changes tracking, with consideration on which lines really need to change and which don't, and less chance of things breaking.

We can probably do the upgrade in two separate steps:
1. First merge a PR that keeps Vue 1, but modularizes everything that this PR had to modularize. We should attempt to create a similar file structure to the final Vue 3 one, but without actually upgrading to Vue 3.
2. Then we can actually upgrade to Vue 3. All resulting CSS (after compilation) should stay mostly identical after the upgrade, and the color picker will still work as usual, which leaves fewer things to check.

After those 2 are done, we should look into other goals, such as 1. decoupling the settings page from the `chrome.*` APIs, which means TurboWarp could theoretically just use our settings page, and 2. considering settings page redesigns we've discussed all over the repository.

GrahamSH-LLK:
This PR didn't really have to modularize anything? It's basically a 1:1 port. I don't understand the point of the first PR. The only CSS differences on this PR have already been changed. On the color picker side, source code is no longer available for my library, just built files, which I'm fairly sure is against Mozilla standards.

webpages/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "webpagesish",
```
Samq64, line 2: Could the package be renamed?

mxmou:
> The only CSS differences on this PR have already been changed.

The `!important`s in `main.css` are still there.

WorldLanguages:
> This PR didn't really have to modularize anything?

What are the new files on src/lib then? There are also other changes that can be merged directly into the Vue 1 project without requiring a Vue 3 upgrade.



> > The only CSS differences on this PR have already been changed.
> 
> The `!important`s in `main.css` are still there.

Would removing those be enough? I was wondering if it made sense to start again. Personally, I would find it easier to start again than trying to modify this PR even more to get it to perfectly match, but you may choose any of the two options as long as I don't work on it :P

> Also all of these PRs get squashed as far as I know so change history shouldn't be a big deal.

Yeah, the number of commits is not a problem, but we want to have the least amount of file diffs. It is also a bad idea to move .js file contents into .vue files, as we continue to modify them and Git has no way to know which lines need updating.

mxmou:
> > The `!important`s in `main.css` are still there.
> 
> Would removing those be enough? I was wondering if it made sense to start again. Personally, I would find it easier to start again than trying to modify this PR even more to get it to perfectly match, but you may choose any of the two options as long as I don't work on it :P

Removing them without changing anything else would break the CSS, but there are two very simple workarounds:
* changing the selector from `.clear-button` to `.large-button.clear-button`
* (not tested) removing `import "../assets/main.css"` from `src/main.js` and importing `main.css` from `settings.css` instead, after the other styles.

Samq64:
If we're not going to be merging this into master soon could we merge into a new branch on the main repo so I can stop pushing to Graham's branch? I know I don't need to resolve the conflicts every time there's a change but I also don't want to have to deal with multiple PRs worth of conflicts later on.

Samq64:
Just give be a few minutes I'm really sorry.

Waakul:
i found a bug, i was just switching tabs and switched to games tab, nothing was there

Waakul:
Also the styles take time to get injected, it doesnt get injected until it's fully loaded

Waakul:
Literally everything in webpages/dist/assets is unformatted and single-line

Waakul:
https://github.com/GrahamSH-LLK/ScratchAddons/pull/7

Samq64:
> i found a bug, i was just switching tabs and switched to games tab, nothing was there

It does it just takes time to load since it makes a lot of network requests.

Waakul:
> > i found a bug, i was just switching tabs and switched to games tab, nothing was there
> 
> It does it just takes time to load since it makes a lot of network requests.

I mean complete empty, literally nothing

Samq64:
> I mean complete empty, literally nothing

Are there any errors in the console?

Waakul:
> > I mean complete empty, literally nothing
> 
> Are there any errors in the console?

it only appeared once, dunno how to get it back

webpages/assets/main.css
```diff
@@ -47,7 +50,7 @@ h1 {
 
 .main {
   display: flex;
-  height: calc(100% - 60px);
+  height: calc(100vh - 60px);
```
Samq64, line 53: This causes the license page background to change colour when scrolling, Do you remember why it was changed?

GrahamSH-LLK: hmm, i can't reproduce.

GrahamSH-LLK:
i thought i already closed this. use as reference impl if needed, but we should probably fully rewrite

PR #6276, opened 2023-06-26
Store setting migrations that were applied

WorldLanguages:
### Changes

Store the list of setting migrations that were applied. Might be useful in the future, in case we want to distinguish users that have some settings because of the migration vs. they already had those settings before then. It's also possible that we'll never use this. It won't take a lot of space to store a few integer numbers.

### Tests

To be tested

### DIFF
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -58,6 +58,8 @@ const updatePresetIfMatching = (settings, version, oldPreset = null, presetOrFn 
       if (typeof presetOrFn === "object") map[key] = presetOrFn.values[key];
     }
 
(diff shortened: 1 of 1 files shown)


background/get-addon-settings.js
```diff
@@ -58,6 +58,8 @@ const updatePresetIfMatching = (settings, version, oldPreset = null, presetOrFn
       if (typeof presetOrFn === "object") map[key] = presetOrFn.values[key];
     }
 
+    if (settings._appliedVersions) settings._appliedVersions.push(version);
```
apple502j, line 61:
```suggestion
    if (Array.isArray(settings._appliedVersions)) settings._appliedVersions.push(version);
```

PR #6277, opened 2023-06-26
editor-dark-mode: use blue color for some elements

mxmou:
### Changes

Makes selected text, hovered block category names, and checkboxes in the block palette blue even if a different highlight color is chosen.

A blue loading screen would look very bright in a dark theme, so the page background is used in dark themes instead.

### Reason for changes

In Scratch, these things are still blue and not purple.

### Tests

Tested by running scratch-www locally.

### DIFF
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -736,6 +736,18 @@
       }
     },
     {
+      "name": "categoryMenu-hoverText",
+      "value": {
+        "type": "textColor",
+        "black": "#3373cc",
+        "white": "#80b5ff",
+        "source": {
+          "type": "settingValue",
+          "settingId": "categoryMenu"
+        }
+      }
+    },
+    {
       "name": "palette-text",
       "value": {
         "type": "textColor",
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -123,17 +123,6 @@
       }
     },
     {
-      "name": "primary-transparent25",
-      "value": {
-        "type": "multiply",
-        "source": {
-          "type": "settingValue",
-          "settingId": "primary"
-        },
-        "a": 0.25
-      }
-    },
-    {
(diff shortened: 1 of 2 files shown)


Secret-chest:
NOOOO! Please add a setting to change the blue colour, as well as cyan.

apple502j (approved):

DNin01 (commented):
This isn't customizable?

I think it would be a missed opportunity not to allow this to be changed by the user (especially for custom light themes such as recreations of TurboWarp's colors), if those are the kinds of customization flexibilities we want to offer.

WorldLanguages:
> This isn't customizable?

We could make it customizable in the future. Or is it strictly needed now?

DNin01:
> We could make it customizable in the future.

Yeah, that would be fine.

mxmou:
I think a setting for the loading screen background would be useful, but it can be added later. The other blue elements probably aren't something many people would want to customize.

DNin01:
Wait a minute, could the popup backdrop color be reused for this? (It could be converted to solid instead of opaque.) If we figured out how to do that, then we wouldn't have to add an "advanced" (potentially confusing) setting to change the blue color.

DNin01:
I understand the difficulty behind making this customizable now.

mxmou:
> Wait a minute, could the popup backdrop color be reused for this? (It could be converted to solid instead of opaque.) If we figured out how to do that, then we wouldn't have to add an "advanced" (potentially confusing) setting to change the blue color.

It could be, but that would make the loading screen completely black in some presets (including 2.0 and 1.4).

WorldLanguages (approved):

Secret-chest:
I want to change it

WorldLanguages:
Feedback:

> So, when I clicked on a sprite in the sprite pane, I noticed that the box on the stage is still purple, even with my settings turning the project editor into 2.0. Maybe there could be a way to make it the color before the update, or the grey and blue that showed up in Scratch 2.0?

PR #6285, opened 2023-06-27
New addon: pagination-top

### LINKED ISSUE #4620: Move the page numbers to the top
I may not want to always have to scroll down to the bottom of my [projects list page](https://scratch.mit.edu/users/D-ScratchNinja/projects/) to see older stuff. That's why I suggested this new addon, "Pages at the top".

What I'm saying is, it would be cool if the page numbers that are usually seen at the bottom to be moved up to the top for easy reaching (like they are on the Discussion Forums). I mean, sure, the bottom is just an _End_ key press away, but for some, it's still convenient to have them up high.

### Mockup
(using DevTools/Inspect Element - check out the top right)

s-paginationattop

I've previously shared this idea on Scratch in [this project](https://scratch.mit.edu/projects/615769983/) and [this forum topic](https://scratch.mit.edu/discuss/topic/571280/).

As for alternatives, there could be a bar with page numbers that floats on some corner on the screen, too (attached to the screen, not the page, like the fix footer feature in the Infinite scrolling addon).

DNin01:
Resolves #4620

### Changes

Adds an addon that adds a copy of the pagination to the top of lists of shared and favorited projects as well as lists of studios someone follows or curates.

### Reason for changes

So you can navigate to different pages faster.

### Tests

Tested in Edge 114 in combination with `scratchr2`.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -143,6 +143,7 @@
   "op-badge",
   "featured-dangos-v2",
   "remember-collapsed-categories",
+  "pagination-top",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/pagination-top/addon.json
+++ b/addons/pagination-top/addon.json
@@ -1,0 +1,27 @@
+{
+  "name": "Quick-access pagination",
+  "description": "Moves the page links for lists of projects and studios to the top.",
+  "tags": ["community", "profiles"],
+  "credits": [
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": [
+        "https://scratch.mit.edu/users/*/projects/",
+        "https://scratch.mit.edu/users/*/favorites/",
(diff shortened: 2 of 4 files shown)


WorldLanguages:
Looks good, just wondering what our approach should be with these tiny addons that are not easy to explain unless we show a screenshot, and that most users wouldn't care if it was enabled by default

GloomyFM:
> Looks good, just wondering what our approach should be with these tiny addons that are not easy to explain unless we show a screenshot, and that most users wouldn't care if it was enabled by default

you could make a new tag like the editor preview thing but with screenshots. It could be in an array.

DNin01:
I was thinking about the idea of adding another feature or two to this to make it a little more significant. You could use keyboard shortcuts to navigate through pages.

WorldLanguages:
Well, theoretically we could also support infinite scrolling on these pages

DNin01:
> Looks good, just wondering what our approach should be with these tiny addons that are not easy to explain unless we show a screenshot, and that most users wouldn't care if it was enabled by default

I agree with you.

This addon might be more interesting if it was a more general pagination-enhancements addon. What if this addon also offered keyboard shortcuts for flipping through a page at a time, the ability to jump to a specific page with <kbd>Ctrl</kbd>+<kbd>G</kbd>, and the option to truncate the page numbers?

PR #6288, opened 2023-06-28
"Username search" addon: show profile picture of searched user

valmontechno:
responds to this pull request #6254

### Changes

Added a new setting in the Username search addon that allows the user to show icon of searched user.

### Reason for changes

This modification enhance the user experience by providing better visual recognition of searched users. This enables users to navigate more efficiently.

### Tests

Google Chrome	114.0.5735.198
Microsoft Edge	114.0.1823.58

### DIFF
--- a/addons-l10n/en/search-profile.json
+++ b/addons-l10n/en/search-profile.json
@@ -1,4 +1,4 @@
 {
   "search-profile/profile": "Profile",
-  "search-profile/invalid-username": "\"{username}\" is not a valid username or the account doesn't exist"
+  "search-profile/invalid-username": "\"{username}\" is not a valid username or the account does not exist"
 }
--- a/addons-l10n/en/search-profile.json
+++ b/addons-l10n/en/search-profile.json
@@ -1,4 +1,4 @@
 {
   "search-profile/profile": "Profile",
-  "search-profile/invalid-username": "\"{username}\" is not a valid username"
+  "search-profile/invalid-username": "\"{username}\" is not a valid username or the account doesn't exist"
 }
--- a/addons/search-profile/style.css
+++ b/addons/search-profile/style.css
@@ -1,4 +1,4 @@
-.tabs button .tab-icon.profile {
+.tabs button .tab-icon.sa-search-profile-icon {
   width: 24px;
   height: 24px;
   border-radius: 3px;
--- a/addons/search-profile/style.css
+++ b/addons/search-profile/style.css
@@ -4,15 +4,11 @@
   border-radius: 3px;
 }
 
-.tabs button.sa-search-profile-btn {
-  /* Vanilla style is width:8.75em */
-  width: auto;
-  min-width: 8.75em;
-}
-
 .sa-search-profile-btn span {
-  display: flex;
-  justify-content: center;
+  display: block;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  text-align: center;
 }
 
 .tabs button:disabled {
--- a/addons/search-profile/style.css
+++ b/addons/search-profile/style.css
@@ -4,6 +4,12 @@
   border-radius: 3px;
 }
 
+.tabs button.sa-search-profile-btn {
+  /* Vanilla style is width:8.75em */
+  width: auto;
+  min-width: 8.75em;
+}
+
 .sa-search-profile-btn span {
   display: flex;
   justify-content: center;
--- a/addons/search-profile/style.css
+++ b/addons/search-profile/style.css
@@ -4,6 +4,11 @@
   border-radius: 3px;
 }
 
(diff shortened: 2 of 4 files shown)


WorldLanguages:
See #6254 for previous discussion

kokofixcomputers:
It would be great I would like this. I will review this for you.

kokofixcomputers:
i could not get this working so i cannot review this for you

WorldLanguages:
Hmmm

image

addons/search-profile/style.css
```diff
@@ -1,3 +1,20 @@
...
+.tabs button.sa-search-profile-btn {
+  /* Vanilla style is width:8.75em */
+  width: auto;
+  min-width: 8.75em;
+}
+
+.sa-search-profile-btn span {
+  display: flex;
+  justify-content: center;
+}
```
mxmou, line 16: This causes a layout shift when the username is shown. I think it would be better to truncate the text.

```suggestion
.sa-search-profile-btn span {
  display: block;
  overflow: hidden;
  text-overflow: ellipsis;
  text-align: center;
}
```

addons/search-profile/userscript.js
```diff
@@ -7,19 +7,36 @@ export default async function ({ addon, console, msg }) {
...
   tab.tabIndex = -1; // unselected tabs should only be focusable using arrow keys
   img.src = addon.self.dir + "/user.svg";
-  img.className = "tab-icon";
+  img.className = "tab-icon sa-search-profile-icon";
   span.innerText = msg("profile");
   addon.tab.displayNoneWhileDisabled(tab);
 
+  const setInvalidUsername = () => {
+    tab.disabled = true;
+    tab.title = msg("invalid-username", { username: user });
```
mxmou, line 21: The error message should now probably say that the user doesn't exist, not that the username is invalid.

addons/search-profile/userscript.js
```diff
@@ -7,19 +7,36 @@ export default async function ({ addon, console, msg }) {
...
   if (valid) {
     tab.addEventListener("click", () => {
       location = `/users/${user}/`;
     });
+    fetch(`https://api.scratch.mit.edu/users/${user}`)
+      .then((res) => res.json())
+      .then((data) => {
+        if (data.code == "NotFound") {
+          setInvalidUsername();
+        } else {
```
mxmou, line 34: There could be an error code other than `NotFound`.
```suggestion
        if (data.code == "NotFound") {
          setInvalidUsername();
        } else if (!data.code) {
```

mxmou (changes requested):

addons/search-profile/userscript.js
WorldLanguages: We could have 2 strings or change the current one to contemplate both cases.

`\"{username}\" is an invalid username or the account doesn't exist`

mxmou: I don't think the difference between an invalid username and a valid one that doesn't exist is very important to the user.

WorldLanguages: Okay, I'll change the existing string

WorldLanguages (approved):

addons-l10n/en/search-profile.json
```diff
@@ -1,4 +1,4 @@
 {
   "search-profile/profile": "Profile",
-  "search-profile/invalid-username": "\"{username}\" is not a valid username"
+  "search-profile/invalid-username": "\"{username}\" is not a valid username or the account doesn't exist"
```
mxmou, line 3:
```suggestion
  "search-profile/invalid-username": "\"{username}\" is not a valid username or the account does not exist"
```
Using "is not" and "doesn't" in the same sentence sounds weird.

addons/search-profile/userscript.js
```diff
@@ -7,19 +7,36 @@ export default async function ({ addon, console, msg }) {
...
     tab.addEventListener("click", () => {
       location = `/users/${user}/`;
     });
+    fetch(`https://api.scratch.mit.edu/users/${user}`)
+      .then((res) => res.json())
+      .then((data) => {
+        if (data.code == "NotFound") {
+          setInvalidUsername();
+        } else if (!data.code) {
+          span.innerText = "@" + data.username;
```
mxmou, line 35: The `@` makes `@Maximouse` just long enough to be cut off 😿
Could the horizontal padding be decreased so that longer usernames can be displayed? I think `0.5rem` is enough.

mxmou (approved):

addons/search-profile/userscript.js
WorldLanguages: > The `@` makes `@Maximouse` just long enough to be cut off 😿

And I wouldn't consider Maximouse to be a long username!

Issue #6294, opened 2023-06-28
Hide editor menu bar icons


WorldLanguages:
### Why this would be helpful

> Scratch added icons to the menu bar at the top, which kinda makes it an eyesore. Maybe make an addon to remove those and get rid of the extra spacing, so it looks like it did before? Side note, using an element blocker from an adblocker doesn't work; the icons come back when you click on something else. 

> they take up a lot of space

> Please add a addon to remove the icons added to the nav bar on project editor in settings/file/edit - it's spacey

### My suggestion

Could be a new addon
Potentially could be merged with the "hide tutorials" addon

medlans:
Yeah, I would like this too. Maybe name it hide header stuff or something, and you could choose to hide tutorials and/or icons?

WorldLanguages:
> It would be cool to have an option to only show either the text or the icons (I like the icons)

callumjt:
i could make this if no ones else is making it

Samq64:
Sure, but it's a feature so it probably won't be added until 1.34.

BroJac5246:
> Potentially could be merged with the \"hide tutorials\" addon

I think it should be this. I was thinking earlier today that the hide tutorials addon is a little strange, but a "Customizable Editor Header" or similar addon would make sense.

medlans:
> > Potentially could be merged with the "hide tutorials" addon
> 
> I think it should be this. I was thinking earlier today that the hide tutorials addon is a little strange, but a "Customizable Editor Header" or similar addon would make sense.

I learned the header tools were removed in 3.0 today, could this be there?

DNin01:
Note that Scratch removes the text labels if the screen is not wide enough, leaving only the icons. If we aren't careful, we could hide the icons and they buttons would become basically invisible.

Secret-chest:
**Menus:** select Icons only / Text only / Both
**Tutorials:** select Show / Icon only / Text only / None
**Only show user icon:** bool
**Padding size (%):** int (percentage)

Secret-chest:
> I learned the header tools were removed in 3.0 today, could this be there?
> 
> @medlans What are "header tools"?
I think it means the scissors, help, grow, shrink, duplicate buttons on the Scratch 2.0 and 1.x header

WorldLanguages:
Note that addon compact-editor (PR #4018) added this functionality, and it cannot be toggled. It keeps only the text in the settings/file/edit menus.
[...]

PR #5761, opened 2023-02-28
New addon: Compact menu bar user dropdown

iqnite:
New version at #6297
Resolves #4648
Related to #4018

### Changes

image

With this new addon, users can hide the username from the editor menu bar, leaving only the profile picture. The name is added to the dropdown menu, like in the website implementation.

### Reason for changes

It saves space, which is suitable for small screens, and there's the same feature for the website.

### Tests

_Tested on Vivaldi 5.7.2921._
~~The username hides appropriately from the menu bar, but it is shown in the account menu only one time, when clicking outside and re-opening it the username disappears. Any ideas on how I could fix this so that it's always visible in the menu (like in the website implementation)?~~

~~**Update:** They're now always shown (thanks @CST1229 ), but they are visible on every menu item (so there are 4 displays). Fix?~~

**Update:** All issues mentioned above were addressed thanks to @CST1229.

PR #6297, opened 2023-06-28
[Updated] New addon: Compact menu bar user dropdown

### LINKED ISSUE #4648: Adjustments to some menu bar items to save space
For those with a small screen, the menu bar can get crowded, especially with addons that add to it. I have some ideas for how we could save space by providing options and updates that shorten or remove some text like so:

Scratch's editor menu bar but modified

### What changed in this image:
- (From an addon) **"Start Recording" changed to "Record"**
- **"Tutorials" text hidden, leaving just the icon**
- **"See Project Page" changed to "Project Page"**
- **Username hidden**

Other ideas to save space:
- (From an addon) **"blocks" in "42 blocks" replaced with an icon**

When I shrunk the window down, nothing was forced to compress or hide to fit thanks to the shorter menu bar.

These are just ideas, and not all of these have to be implemented, or there may be another change we can make, but one or more of these changes in the form of a new addon for Scratch's menu bar items and updates to existing addons for anything added by them would be nice for those with a smaller screen.

Kind of related to #2110

iqnite:
Based on #5761 
Resolves #4648
Related to #4018

### Changes

With this new addon, users can hide the following:
- The menu labels "Settings", "File", and "Edit", leaving only the pictogram.
- The username, leaving only the profile picture. (The name is added back inside the menu, like in the website implementation.)
- The label on the "Tutorials" button.

### Reason for changes

It saves space, which is suitable for small screens, and there's the same feature for the website. Also, Scratch just released an update that makes the menu bar overflow using Scratch Addons.

### Tests

_Tested on Vivaldi 5.7.3035._

The main functionality seems to work without issues, dynamic reload is supported but messes up the menu when it changes.

### Additional info

Thanks to @CST1229 for debugging the original PR.

If you wonder why I made a new PR, it's because the original PR could still be merged directly if there are any problems with this new one.

### Demonstration of the changes

image


iqnite (commented):
Some questions:
- Should we remove the `tutorials-button` addon? Its main purpose is implemented here now.
- OK if there's no separate option for which labels to hide? I prefer putting code in userstyles, so dynamic reload is automatic and performance is less impacted.

Samq64:
> OK if there's no separate option for which labels to hide?

That should be fine.

People have also requested hiding the icons. Could this be updated so users can choose between the two?

iqnite:
> > OK if there's no separate option for which labels to hide?
> 
> That should be fine.
> 
> People have also requested hiding the icons. Could this be updated so users can choose between the two?

That would blow up my userstyle concept. Is there any way to toggle userstyles directly from the addon settings?

Samq64:
> That would blow up my userstyle concept. Is there any way to toggle userstyles directly from the addon settings?

Using an addonn manifest [if](https://scratchaddons.com/docs/reference/addon-manifest/#if) and setting `updateUserstylesOnSettingsChange` to true should work.

iqnite:
> Adding an [if](https://scratchaddons.com/docs/reference/addon-manifest/#if) to your userstyle matches and setting `updateUserstylesOnSettingsChange` to true should work.

Thanks, I'll look at it later.

iqnite:
### Update

I'm planning to merge together `tutorials-button`, `editor-messages`, and this into a new `custom-menu-bar` addon.

iqnite:
I've heard something about "Settings migration", how do I import settings from the old addons?

WorldLanguages:
> Merged `tutorials-button`, `editor-messages`, `compact-menu-bar`, and this into a new `custom-menu-bar` addon.

Should `editor-messages` be merged? I'm not sure.
Also, `compact-menu-bar` does not exist yet. You should clarify that

Samq64:
> Should `editor-messages` be merged? I'm not sure.

I don't think so because it's a new feature and everything else just modifies existing things.

iqnite:
I think it should be merged, so it's easier to customize the elements on the menu bar from one place.

addons/custom-menu-bar/addon.json
```diff
@@ -0,0 +1,138 @@
...
+      "potentialValues": [
+        {
+          "name": "Show icon and text",
+          "id": "both"
+        },
+        {
+          "name": "Show icon only",
+          "id": "icon"
+        }
+      ],
```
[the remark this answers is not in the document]
iqnite: > Yeah, I see. You could add the setting, though. `Hide icon and text (unimplemented)`

I'll wait until it gets fixed.

iqnite:
> I don't really think editor-message really fits. It's the kind of addon that we'd like to showcase with a screenshot of the messages icon, and that the user may want to toggle with a single click. In contrast, most addons whose names start with "customizable" do not change anything unless you change the default settings of the addon. Note that we can always use the "related addons" feature to link the user to similar addons (PR not merged).

It's already merged, and the `discuss-button` addon had an option to remove the "Ideas" button, which unrelated.

iqnite:
Can someone help me with the Messages button option? It does a weird thing:
1. When I enable it, I have to reload the editor in order to apply the changes
2. Once reloaded, I can disable it without having to reload
3. If I didn't reload the editor in step 2, I can also reenable it and the button appears instantly

Does anyone know where the problem is?

WorldLanguages:
> Can someone help me with the Messages button option? It does a weird thing:

We could help you, but I find it unlikely the messages button will be merged with this addon. Maybe we could leave this discussion for later, at least?

iqnite:
> I find it unlikely the messages button will be merged with this addon.

I repeat that it's already merged, and it should be. This addon allows you to fully customize the menu bar, which means that you can add or remove buttons. Although the Messages button isn't part of vanilla Scratch, users should be able to manage all the items from one place, and it would feel more intuitive.

BroJac5246:
> Although the Messages button isn't part of vanilla Scratch

Hence the issue with it being merged into this. I agree with WorldLanguages that it should remain separate. Yes, it's part of the navbar, but it serves a different purpose than the rest of the addons here.

iqnite:
Democracy sees it differently: https://discord.com/channels/806602307750985799/1123995427688169605/1125861745290903612

DNin01:
Note: Prettier, the code formatter we use, updated and changed a default setting, so it modified a bunch of files. You can turn off Actions temporarily and revert any Format code commits from Git if this gets in your way.

iqnite:
What do you think: Should the Tutorials button option be on hide by default or should we add an extra migration code for this update?
The first one would be way more easier to implement, but normally such addons don't do anything until the users change the settings.

addons/tutorials-button/menu-icons.css
```diff
@@ -0,0 +1,7 @@
+[class*="settings-menu_dropdown-label"] {
+  display: none;
+}
+
+[class*="menu-bar_collapsible-label"] {
+  display: none;
+}
```
Samq64, line 7:
```suggestion
[class*="settings-menu_dropdown-label"],
[class*="menu-bar_collapsible-label"] {
  display: none;
}
```

addons/tutorials-button/addon.json
```diff
@@ -1,21 +1,139 @@
...
+        }
+      }
+    }
+  ],
   "dynamicDisable": true,
-  "enabledByDefault": false
+  "dynamicEnable": true,
+  "updateUserstylesOnSettingsChange": true,
+  "enabledByDefault": false,
+  "versionAdded": "1.22.0",
```
Samq64, line 132: Should we consider this a new addon or not?

Samq64 (commented):
Could you remove the changes from `editor-messages` and `addons.json`?

Samq64:
> Resolves #6294

I know showing just the icons also saves space, but that issue is specifically about hiding the icons and only showing the labels. Could you unlink the issue or add the option?

addons/tutorials-button/addon.json
iqnite: That update from "Show Discuss button" to "Custom navbar" wasn't considered as new addon. They both have the same addon ID, so...

iqnite:
> Could you unlink the issue or add the option?

I've unlinked it now, thanks for reminding me. I want to add that option, but I'm not able to locate the icons with CSS as they don't belong to any class and don't have any ID. Do you have an idea how we could do that, possibly without using a userscript?

addons/tutorials-button/addon.json
WorldLanguages: I'd encourage you to change the addon ID, as we'll need to add settings migration code nevertheless

iqnite: So... basically revert commit [451347e](https://github.com/ScratchAddons/ScratchAddons/pull/6297/commits/451347e8d754fb439574566f42ebf6a80196f891)?

WorldLanguages: Yep

iqnite: Done!

WorldLanguages:
> I know showing just the icons also saves space, but that issue is specifically about hiding the icons and only showing the labels. Could you unlink the issue or add the option?

@Samq64 Note that the "compact editor" addon fixed the issue you're referencing.
More discussion on how we should handle customizations of the editor menu bar belongs here: #6432

addons/addons.json
```diff
@@ -143,6 +141,8 @@
   "op-badge",
   "featured-dangos-v2",
   "remember-collapsed-categories",
+  "custom-menu-bar",
+  "editor-messages",
```
Samq64, line 145: Could `editor-messages` be put back where it was under `cat-blocks`?

addons/editor-messages/userscript.js
```diff
@@ -1,7 +1,6 @@
-export default async function ({ addon, console, msg }) {
+export default async function ({ addon, console }) {
   const messages = document.createElement("a");
   messages.href = "/messages/";
-  messages.title = msg("messages");
```
Samq64, line 4: Could this be re-added? It was the tooltip that appears when hovering over the icon.

addons/editor-messages/userstyle.css
```diff
@@ -6,7 +6,7 @@
 
 .sa-editormessages::before {
   /* Icon is in a pseudo-element so dark mode can apply a filter to it
-  without affecting the message count */
+    without affecting the message count */
```
Samq64, line 9:
```suggestion
  without affecting the message count */
```

addons-l10n/en/editor-messages.json
Samq64, line 1: Could this be re-added too?

Samq64 (commented):
This looks pretty good apart form some unnecessary diffs. Could a notice be added mentioning that the hide tutorials button addon was merged with this?

addons/custom-menu-bar/addon.json
```diff
@@ -0,0 +1,133 @@
+{
+  "name": "Custom menu bar",
```
Samq64, line 2: Most addons use customizable
```suggestion
  "name": "Customizable menu bar",
```

addons/custom-menu-bar/addon.json
```diff
@@ -0,0 +1,133 @@
...
+  "description": "Hide specific editor menu bar items.",
```
Samq64, line 3:
```suggestion
  "description": "Allows you to hide specific editor menu bar items or remove their labels.",
```

addons/custom-menu-bar/addon.json
```diff
@@ -0,0 +1,133 @@
...
+        },
+        {
+          "name": "Only show icons",
+          "id": "icons"
+        }
+      ],
+      "default": "both"
+    },
+    {
+      "name": "Tutorials button",
```
Samq64, line 43: I think this setting should be a show/hide boolean and follow the the other setting when shown.

Samq64:
When the customizable navigation bar addon was added we kept the existing addon ID. I think the same should be done here.

WorldLanguages:
I'll postpone this for the next milestone as I believe #6432 might require more discussion

WorldLanguages:
> I've noticed that the menus don't have icons or labels when the Compact Editor addon and the Hide Menu Bar Labels option are enabled together. That part of @lisa-wolfgang 's addon should be moved to my addon as an additional selection.

I proposed to discuss that in this issue: https://github.com/ScratchAddons/ScratchAddons/issues/6432#issuecomment-1656920393, but received no responses.

I said:
> Personally, I don't think we should remove the editor menu bar changes from compact-editor. But that leaves the "customizable editor menu bar" addon in a quite awkward position.

iqnite:
I think I found a solution for the Hide Icons thing, but it will mean that:
1. Menus and Tutorials button will be controlled together
2. You can't entirely hide the Tutorials button anymore
3. The rest will still work normally
I think that that wouldn't be a problem at all for new users, and it wouldn't be a drastic change for previous ones. The "Compact editor" button won't hide icons anymore, but the setting would just be moved. What do you think? Should I commit it?

Secret-chest:
> You can't entirely hide the Tutorials button anymore

Tutorials are worthless.

iqnite:
> Tutorials are worthless.

You have the choice: A simple and consistent code that can restore the old behavior or 3 inconsistent features based on old addons.

DNin01:
> > You can't entirely hide the Tutorials button anymore
> 
> Tutorials are worthless.

Minimizing it is still better than nothing.

WorldLanguages:
> 2\. You can't entirely hide the Tutorials button anymore

What does that mean and why?

Secret-chest:
I still want icons+text or text only for the menus, but I want to hide the tutorials or at least not have the text.

iqnite:
The reason why you can't modify the Tutorials button separately is that I'm applying the userstyle to every button in the `file-area`, which contains menus and the Tutorials button. They will all be treated the same in the new behavior.
If someone has a better solution to make this addon compatible with the Compact Editor addon, please tell me!

addons/custom-menu-bar/addon.json
iqnite: > I think this setting should be a show/hide boolean and follow the the other setting when shown.

Wdym?

Secret-chest:
> If someone has a better solution to make this addon compatible with the Compact Editor addon, please tell me!

Use the solution in `editor-compact`.

iqnite:
> > If someone has a better solution to make this addon compatible with the Compact Editor addon, please tell me!
> 
> Use the solution in `editor-compact`.

That's what I did.

iqnite:
Wait, I have an idea. ;)

iqnite:
Now it's compatible at 100% and it has an option to hide menu bar icons.

addons/custom-menu-bar/menu-labels.css
```diff
@@ -0,0 +1,3 @@
+[class*="menu-bar_file-group"] > [class*="menu-bar_hoverable"] > img {
+  display: none;
```
Samq64, line 2: image

Could the margin or padding be reduced? When hovered they also look very slightly off-centred but I don't think that's a big deal.

addons/editor-compact/addon.json
```diff
@@ -12,6 +12,12 @@
       "link": "https://scratch.mit.edu/users/Maximouse/"
     }
   ],
+  "info": [
+    {
+      "id": "hide-icons-update",
+      "text": "To hide menu bar icons, use the \"Customizable menu bar\" addon."
```
Samq64, line 18: After #6104 the addons could be linked instead.

iqnite: It's to tell users that the behavior of Compact editor was moved to Customizable menu bar.

addons/custom-menu-bar/menu-labels.css
iqnite: I can try...

iqnite: Is it better now?

Samq64:
With labels only mode the labels still disappear when the window is narrow. What should be done about about that?

iqnite:
> With labels only mode the labels still disappear when the window is narrow. What should be done about about that?

I've noticed the issue, and I don't know how to handle that without userscripts. I think this isn't that important, as narrow windows aren't the most common usage for Scratch and people with narrow screens would use the other option.

DNin01:
> > With labels only mode the labels still disappear when the window is narrow. What should be done about about that?
> 
> 
> 
> I've noticed the issue, and I don't know how to handle that without userscripts. I think this isn't that important, as narrow windows aren't the most common usage for Scratch and people with narrow screens would use the other option.

`@media (max-width: 1024px) { ... }` should work. It's a CSS feature - any style rules inside the brackets will only take effect if a condition (usually user preferences or device/browser/window information) is met (in this case, if the window is at least 1,024 pixels wide).

iqnite:
Thanks!

Samq64 (approved):

addons/custom-menu-bar/menu-labels.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+[class*="menu-bar_hoverable"] {
+  justify-content: center;
+}
+
+[class*="menu-bar_hoverable"] > span {
+  margin-right: 0;
+  padding-right: 0;
+}
```
DNin01, line 12: Looks like there's the tiniest bit of padding/margin to the left of the labels - you can see while hovering over them.
Labels with leftover space to the left

This can probably be fixed by changing both margin right and margin left or simply removing it from all sides with `margin: 0` and `padding: 0`.

(In that case, you might also want to move that section into `@media (max-width: 1024px) {}` so it doesn't apply to the icons.)

WorldLanguages:
Regarding comment https://github.com/ScratchAddons/ScratchAddons/issues/6432#issuecomment-1656920393, what will we be doing? Will we remove the setting from "compact editor" and migrate it here?

iqnite:
> Regarding comment [#6432 (comment)](https://github.com/ScratchAddons/ScratchAddons/issues/6432#issuecomment-1656920393), what will we be doing? Will we remove the setting from "compact editor" and migrate it here?

Yes, I've already made the necessary changes. What we still need are the migration codes for editor-compact and tutorials-button.

addons/custom-menu-bar/addon.json
```diff
@@ -0,0 +1,127 @@
...
+      "link": "https://scratch.mit.edu/users/lisa_wolfgang/",
+      "note": "hide tutorials button",
+      "id": "tutorials-button"
+    }
+  ],
+  "info": [
+    {
+      "id": "tutorials-button-update",
+      "text": "The \"Hide Tutorials button\" addon has been moved here as a setting."
+    }
```
DNin01, line 26: Just in case of confusion:
```suggestion
    },
    {
      "id": "small-screens",
      "text": "The text labels may be automatically replaced by icons on smaller windows."
    }
```

DNin01 (approved):
Looks OK to me! Sorry for the wait.

iqnite:
> Sorry for the wait.

It's fine! I just think about all those people with small screens like me that have been waiting for this addon...

Samq64:
We'll need setting migrations for users with the tutorials button or compact editor enabled.

iqnite:
> We'll need setting migrations for users with the tutorials button or compact editor enabled.

Yeah I know, btw I have no idea how that stuff works

DNin01:
> We'll need setting migrations for users with the tutorials button or compact editor enabled.

I'll add a task to the PR description.

DNin01:
I added migration logic. To test it...

1. Switch to this branch and pull origin
2. Set the state of the `custom-menu-bar` addon to `undefined` (the easiest way to do this is by adding `addonsEnabled["custom-menu-bar"] = undefined;` somewhere around line 141 of the `get-addon-settings.js` file, then reload Scratch Addons). `custom-menu-bar` should now be disabled.
3. Switch to the `master` branch
4. Enable the `tutorials-button` addon
5. Switch to this branch
6. Reload Scratch Addons
7. The `custom-menu-bar` addon should now be enabled

Samq64:
The migration works when updating, but not when importing. Is it possible the settings are being minified before the extension reloads even on per-release versions? Also there should probably be a second migration for compact editor users but I can do that.

DNin01:
> The migration works when updating, but not when importing. Is it possible the settings are being minified before the extension reloads even on pre-release versions? Also there should probably be a second migration for compact editor users but I can push that.

Yeah, the state of removed addons is probably not stored in or pulled from settings exports, but that's okay, those users will probably be able to find the customization option again by searching.

DNin01:
Having `editor-compact` enabled now sets `menu-labels` to "labels".

Samq64 (approved):

iqnite:
🎉

background/get-addon-settings.js
```diff
@@ -139,6 +139,24 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
       }
     }
 
+    if (addonsEnabled["custom-menu-bar"] === undefined) {
+      // Transition v1.35 to v1.36
+      if (addonsEnabled["tutorials-button"] === true) {
+        // Hide Tutorials button is now a setting in Customizable menu bar. Enable it for existing addon users.
+        madeAnyChanges = true;
+        addonsEnabled["custom-menu-bar"] = true;
+        addonSettings["custom-menu-bar"] = { ["hide-tutorials-button"]: true };
```
WorldLanguages, line 148: No idea why there's brackets [] here, but in this case it's equivalent as not having them.

WorldLanguages: Double-checked the migration code, looks alright

addons/custom-menu-bar/username.js
```diff
@@ -0,0 +1,14 @@
+export default async function ({ addon }) {
+  if (await addon.auth.fetchIsLoggedIn()) {
+    const username = await addon.auth.fetchUsername();
+    addon.tab.redux.initialize(); // Start listening to Redux events
+    addon.tab.redux.addEventListener("statechanged", async (e) => {
+      if (!addon.settings.get("compact-username")) return; // Skip if the option is disabled
+      if (e.detail.action.type !== "scratch-gui/menus/OPEN_MENU" || e.detail.action.menu !== "accountMenu") return; // Skip if another Redux event occurs
+      const profileSpans = await addon.tab.waitForElement(".menu_menu-item_3EwYA.menu_hoverable_3u9dt"); // Do this if the menu is opened
```
WorldLanguages, line 8: Bad idea to target a hashed class name.

WorldLanguages: This is also happening inside the editor, so it isn't a good idea to use waitForElement lightly, without any optimizations!

WorldLanguages: cc @DNin01 @Samq64

addons/editor-compact/addon.json
WorldLanguages, line 18: Technically, the convention is to start the addon name with lowercase (for reference, see all the info's that reference "developer tools")

addons/custom-menu-bar/username.js
```diff
@@ -0,0 +1,14 @@
+export default async function ({ addon }) {
+  if (await addon.auth.fetchIsLoggedIn()) {
+    const username = await addon.auth.fetchUsername();
+    addon.tab.redux.initialize(); // Start listening to Redux events
+    addon.tab.redux.addEventListener("statechanged", async (e) => {
+      if (!addon.settings.get("compact-username")) return; // Skip if the option is disabled
+      if (e.detail.action.type !== "scratch-gui/menus/OPEN_MENU" || e.detail.action.menu !== "accountMenu") return; // Skip if another Redux event occurs
```
WorldLanguages, line 7: So there's a total of 3 possible Redux events? I don't get what this is supposed to mean.

addons/custom-menu-bar/username.js
Samq64: I'm working on the userscript now.

addons/custom-menu-bar/username.js
Samq64: The first is the event that's triggered when any menu is open and the second is which menu was just opened. I think not returning early here makes the code easier to understand.

background/get-addon-settings.js
DNin01: JavaScript doesn't really like hyphens in symbols.
```js
JSON.stringify({ a-b: 1 }); // > Uncaught SyntaxError: Unexpected token '-'
```
To work around this, brackets are needed.

WorldLanguages: @DNin01 Sure, but the quotes are fixing that, not the brackets :P

addons/custom-menu-bar/username.js
WorldLanguages: Huh, looks like I misread. These check for inequality, not equality. Sorry

background/get-addon-settings.js
DNin01: > @DNin01 Sure, but the quotes are fixing that, not the brackets :P

Oh.

addons/custom-menu-bar/username.js
iqnite: Working on a fix.

Samq64: @iqnite You'll probably want to take a quick look at #7187 before doing anything.

PR #6308, opened 2023-06-29
Fix stage header addons not showing in full screen

### LINKED ISSUE #6275: Clone counter not showing in fullscreen editor mode
### Description

If you go in the editor, create a clone, make the stage smaller, and click fullscreen, you will not see the clone counter. It should show in fullscreen even if the stage is smaller.

### Steps to reproduce

1. Produce at least a clone in a project.
2. Go in the editor.
3. Enable the "Clone counter" extension.
4. Press the "Switch to small stage" icon which will make the stage small and the code area big.
5. Press the fullscreen button.

A video about the bug:
https://github.com/ScratchAddons/ScratchAddons/assets/78285518/50ab1f57-3a3b-4a70-b887-d3099eac9693

### Environment

Browser: Firefox ESR 102.12.0esr (64-bit)
Scratch Addons version: v1.33.1
Operating system: Kali Linux

mxmou:
Resolves #6275

### Changes

Fixes a bug in `clones`, `mouse-pos` and `vol-slider`: switching to full screen or the project page while small stage mode was selected didn't show the addon's UI. `gamepad` doesn't seem to be affected.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/clones/style.css
+++ b/addons/clones/style.css
@@ -37,6 +37,6 @@
   content: attr(data-str);
 }
 
-.sa-clones-small .clone-container-container {
+.sa-small-stage .clone-container-container {
   display: none !important;
 }
--- a/addons/clones/userscript.js
+++ b/addons/clones/userscript.js
@@ -1,27 +1,12 @@
+import addSmallStageClass from "../../libraries/common/cs/small-stage.js";
+
 export default async function ({ addon, console, msg }) {
   const vm = addon.tab.traps.vm;
 
   let showOnProjectPage = addon.settings.get("projectpage");
   let showIconOnly = addon.settings.get("showicononly");
 
-  const updateStageSize = () => {
-    if (!addon.tab.redux.state) return;
-    const isSmallStage = addon.tab.redux.state.scratchGui.stageSize.stageSize === "small";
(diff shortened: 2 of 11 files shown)


WorldLanguages:
> `gamepad` doesn't seem to be affected.

Huh, why is that?

WorldLanguages:
Ideally, shouldn't we avoid duplicated code and merge all of this logic into a single JS module?
It also does not make sense to have 3 separate HTML class names that mean "the stage is in small mode"

mxmou:
> > `gamepad` doesn't seem to be affected.
> 
> Huh, why is that?

https://github.com/ScratchAddons/ScratchAddons/blob/4151aa8/addons/gamepad/style.css#L50
In full screen mode, `data-editor-mode` will be set to `fullscreen`.

> Ideally, shouldn't we avoid duplicated code and merge all of this logic into a single JS module? It also does not make sense to have 3 separate HTML class names that mean "the stage is in small mode"

I'll move it into a module.

apple502j (approved):

Samq64 (approved):

WorldLanguages:
We can merge this for v1.33.3 and let's hope we're not missing any bugs!

hueychen27:
Thank you for fixing this.

Issue #6296, opened 2023-06-28
Update default customizable quotes on forums colour


Samq64:
### Why this would be helpful

`scratchr2` (mostly) switched over to using the new purple, but the default customizable quotes on forums colour is still the old Scratch 2 blue.

### My suggestion

Change the default colour to the new purple.

mxmou:
> `scratchr2` (mostly) switched over to using the new purple, but the default customizable quotes on forums colour is still the old Scratch 2 blue.

The default is actually the same as the post header color, which is still blue. I think we should wait for the ST to update it.

WorldLanguages:
Related feedback:

> Also, the links on https://scratch.mit.edu/help/videos/ still showed purple.

Samq64:
I didn't even know that page existed.

medlans:
> Related feedback:
> 
> > Also, the links on https://scratch.mit.edu/help/videos/ still showed purple.
> 
> It's a Scratch page that is rarely used.

By the way, it also happens when I hover over usernames on profile pages, and I think it only happens on scratchr2. Also, when I applied all my settings, I could see purple borders around unfollow discussion.

> I didn't even know that page existed.

It still does, and I have it linked as “Help” on my navbar because the actual help page doesn’t exist anymore and just redirects to ideas. I remember it used to be linked on there.

WorldLanguages:
> it also happens when I hover over usernames on profile pages

Yeah, on the "default colors (blue)" preset, hovering over a comment author still shows purple:

![image](https://github.com/ScratchAddons/ScratchAddons/assets/17484114/e8738204-481d-4cb5-8907-d089b39412c7)

mxmou:
> Related feedback:
> 
> > Also, the links on https://scratch.mit.edu/help/videos/ still showed purple.
> 
> It's a Scratch page that is rarely used.

I don't think it makes sense to try to support that page. There's even an open PR in scratch-www that makes it redirect to Ideas: scratchfoundation/scratch-www#7307

medlans:
> > Related feedback:
> > > Also, the links on https://scratch.mit.edu/help/videos/ still showed purple.
> > 
> > 
> > It's a Scratch page that is rarely used.
> 
> I don't think it makes sense to try to support that page. There's even an open PR in scratch-www that makes it redirect to Ideas: [scratchfoundation/scratch-www#7307](https://github.com/scratchfoundation/scratch-www/pull/7307)

[...]

PR #6268, opened 2023-06-25
dark-www: compatibility with Scratch color changes

mxmou:
### Changes

Makes `dark-www` work better with Scratch's color contrast update. I didn't make changes to other addons in this PR unless they were referencing CSS variables I renamed or removed.

We don't know what changes, if any, will be made to scratchr2. I decided to use the same colors on scratch-www and scratchr2 (see screenshots below). If the ST does something different, we can change it in a future update.

I used the updated Experimental Dark preset proposed in #6128 in the screenshots.

Explore page
Ideas page
Profile page

### Tests

Tested by running scratch-www locally.

PR #6315, opened 2023-06-29
dark-www: old scratchr2 blue and other changes

### LINKED ISSUE #6314: The scratchr2 navbar border color changed
Before the navbar changed to purple, there were borders between each button on scratchr2 pages, but now they are the same color as the navbar background, making them invisible.

Scratch navbar before-after

So we can actually just remove this CSS variable from `dark-www`:
https://github.com/ScratchAddons/ScratchAddons/blob/a60c89103303a7bfe114199115f5a270f079fe06/addons/dark-www/addon.json#L233-L245

and change the uses of it:
https://github.com/ScratchAddons/ScratchAddons/blob/a60c89103303a7bfe114199115f5a270f079fe06/addons/dark-www/experimental_scratchr2.css#L68-L72

mxmou:
Resolves #6314

### Changes

* Fixes the color of hovered usernames in profile comments (https://github.com/ScratchAddons/ScratchAddons/issues/6296#issuecomment-1612304865)
* Removes separators between menu bar items (#6314)
* Changes the blue color on scratchr2 pages if a blue preset is selected (https://github.com/ScratchAddons/ScratchAddons/pull/6268#issuecomment-1613487815)
* `customCssVariables` changes:
  * New `default` property for `"type": "map"` value providers. It's a value provider that's used as a fallback if none of the options match.
  * The `options` of a `"type": "map"` value provider can now also be value providers.

### Reason for changes

To make the "Scratch's default colors" preset more similar to Scratch's new design and the blue presets to the old one.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/dark-www/experimental_scratchr2.css
+++ b/addons/dark-www/experimental_scratchr2.css
@@ -448,6 +448,7 @@ input.link:hover,
 input.link:active,
 input.link.black:hover,
 input.link.black:active,
+#comments .comment .info .name a:hover,
 #comments .more-replies .pulldown,
 #email-resend-box a {
   color: var(--darkWww-link);
--- a/addons/dark-www/experimental_scratchr2.css
+++ b/addons/dark-www/experimental_scratchr2.css
@@ -24,7 +24,7 @@ label,
 #topnav ul.account-nav .logged-in-user .dropdown-menu,
 #topnav li.logout.divider input,
 #topnav ul.account-nav .sign-in .popover {
-  background-color: var(--darkWww-navbar);
+  background-color: var(--darkWww-navbar-scratchr2);
 }
 #topnav ul.site-nav li,
 #topnav ul.site-nav li.last,
@@ -33,11 +33,11 @@ label,
 #topnav ul.account-nav.logged-in > li,
 #topnav ul.account-nav ul.user-nav li.logout.divider,
 #topnav ul.account-nav .sign-in .popover {
-  border-color: var(--darkWww-navbar);
+  border-color: var(--darkWww-navbar-scratchr2);
 }
 #topnav ul.account-nav .sign-in .popover .arrow,
 #topnav ul.account-nav .sign-in .popover .arrow::after {
-  border-bottom-color: var(--darkWww-navbar);
+  border-bottom-color: var(--darkWww-navbar-scratchr2);
 }
 #topnav ul.site-nav li a,
 #topnav ul.account-nav > li > span,
@@ -382,9 +382,9 @@ ul.pagination span /* mobile forums */ {
   color: var(--darkWww-button-scratchr2ButtonText);
 }
 .button {
-  background-color: var(--darkWww-button);
+  background-color: var(--darkWww-button-scratchr2);
   background-image: none; /* Fixes Scratch bug: https://scratch.mit.edu/discuss/topic/694218/ */
-  border-color: var(--darkWww-button);
+  border-color: var(--darkWww-button-scratchr2);
   color: var(--darkWww-button-scratchr2ButtonText);
 }
 .button a,
@@ -396,7 +396,7 @@ ul.pagination span /* mobile forums */ {
   color: var(--darkWww-button-scratchr2ButtonText);
 }
 .button:hover {
-  background-color: var(--darkWww-button-variant);
+  background-color: var(--darkWww-button-scratchr2Hover);
   background-image: none; /* Fixes Scratch bug: https://scratch.mit.edu/discuss/topic/694218/ */
 }
 input:focus,
@@ -414,14 +414,14 @@ textarea:focus,
   box-shadow: none;
 }
 .blockpost div.box {
-  background-color: var(--darkWww-button);
-  border-top-color: var(--darkWww-button);
+  background-color: var(--darkWww-button-scratchr2PostHeader);
+  border-top-color: var(--darkWww-button-scratchr2PostHeader);
 }
 .my-ocular-reaction-button {
-  box-shadow: 0 0 2px var(--darkWww-button);
+  box-shadow: 0 0 2px var(--darkWww-button-scratchr2PostHeader);
 }
 .my-ocular-reaction-button.selected {
-  background-color: var(--darkWww-button);
+  background-color: var(--darkWww-button-scratchr2PostHeader);
   color: var(--darkWww-button-scratchr2PostHeaderText);
 }
 .blockpost .box-head {
@@ -450,7 +450,7 @@ input.link.black:active,
 #comments .comment .info .name a:hover,
 #comments .more-replies .pulldown,
 #email-resend-box a {
-  color: var(--darkWww-link);
+  color: var(--darkWww-link-scratchr2);
 }
 .inew,
 .isticky:not(.iclosed) {
--- a/addons/dark-www/experimental_scratchr2.css
+++ b/addons/dark-www/experimental_scratchr2.css
@@ -22,9 +22,12 @@ label,
 /* Navigation bar background */
 #topnav .innerwrap,
 #topnav ul.account-nav .logged-in-user .dropdown-menu,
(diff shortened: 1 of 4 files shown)


medlans:
So would this fix the issue where the 3.0 colors are applied to the 2.0 pages?
Anyway, I was thinking of being able to have the 2.0 colors on both 3.0 and 2.0 pages, or alternatively, 2.0 <-> 3.0 (I mean we have a bunch of things you can add to the editor to make it look more like 2.0) for a while (before the pull requests were made), maybe with presets or something??

addons/dark-www/experimental_scratchr2.css
```diff
@@ -384,9 +382,9 @@ ul.pagination span /* mobile forums */ {
   color: var(--darkWww-button-scratchr2ButtonText);
 }
 .button {
-  background-image: linear-gradient(var(--darkWww-button), var(--darkWww-button-variant));
-  border-color: rgba(0, 0, 0, 0.2);
-  border-bottom-color: rgba(0, 0, 0, 0.3);
+  background-color: var(--darkWww-button-scratchr2);
+  background-image: none; /* Fixes Scratch bug: https://scratch.mit.edu/discuss/topic/694218/ */
```
WorldLanguages, line 386: Wow, hasn't this been fixed yet?

WorldLanguages (approved):

WorldLanguages:
> So would this fix the issue where the 3.0 colors are applied to the 2.0 pages?

@medlans Yeah, this fixes quite a lot of colors, including the 2.0 navbar and some colors in the forums.
There's probably some colors that still don't match 100% the old blues, such as the "post comment" button on profiles. But I'm not really sure.

**Before this PR: (v1.32.2):** ("Default colors (blue)" preset)
image

**After this PR:** (Also "Default colors (blue)" preset)
image

WorldLanguages:
> Anyway, I was thinking of being able to have the 2.0 colors on both 3.0 and 2.0 pages, or alternatively, 2.0 <-> 3.0 (I mean we have a bunch of things you can add to the editor to make it look more like 2.0) for a while (before the pull requests were made), maybe with presets or something??

@medlans I guess you can already achieve this. But I assume you mean that you'd like to have a "2.0 colors" preset that can be applied with just two clicks? If that's the case, I encourage you to open a [new issue](https://github.com/ScratchAddons/ScratchAddons/issues/new?assignees=&labels=status%3A+needs+triage&projects=&template=blank.md&title=) and specify which colors you think should be assigned to each color setting on this preset.

addons/dark-www/experimental_scratchr2.css
CST1229: Doesn't seem to be.
image

medlans:
> > So would this fix the issue where the 3.0 colors are applied to the 2.0 pages?
> 
> @medlans Yeah, this fixes quite a lot of colors, including the 2.0 navbar and some colors in the forums. There's probably some colors that still don't match 100% the old blues, such as the "post comment" button on profiles. But I'm not really sure.
> 
> **Before this PR: (v1.32.2):** ("Default colors (blue)" preset) image
> 
> **After this PR:** (Also "Default colors (blue)" preset) image

Yeah, normally there's separators between stuff in the navbar, and a gradient for the follow discussion button, which you can see through here:
https://web.archive.org/web/20141015105751/http://scratch.mit.edu/discuss/topic/10588

> > Anyway, I was thinking of being able to have the 2.0 colors on both 3.0 and 2.0 pages, or alternatively, 2.0 <-> 3.0 (I mean we have a bunch of things you can add to the editor to make it look more like 2.0) for a while (before the pull requests were made), maybe with presets or something??
> 
> @medlans I guess you can already achieve this. But I assume you mean that you'd like to have a "2.0 colors" preset that can be applied with just two clicks? If that's the case, I encourage you to open a [new issue](https://github.com/ScratchAddons/ScratchAddons/issues/new?assignees=&labels=status%3A+needs+triage&projects=&template=blank.md&title=) and specify which colors you think should be assigned to each color setting on this preset.

Yeah that's what I meant LOL.

addons/dark-www/experimental_scratchr2.css
```diff
@@ -24,10 +24,20 @@ label,
...
-#topnav ul.account-nav .sign-in .popover .arrow:after {
-  border-bottom-color: var(--darkWww-navbar);
+#topnav ul.site-nav li,
+#topnav ul.site-nav li.last,
+#topnav ul.account-nav > li,
+#topnav ul.account-nav > li.join-scratch,
+#topnav ul.account-nav.logged-in > li,
+#topnav ul.account-nav ul.user-nav li.logout.divider,
+#topnav ul.account-nav .sign-in .popover {
+  border-color: var(--darkWww-navbar-scratchr2);
```
DNin01, line 36: A minor thing - when the navigation bar is set to a transparent color, the borders become visible.

A transparent purple navigation bar with visible borders between each link

DNin01 (approved):
I think this is good!

Secret-chest:
does it affect `scratchr2`?

mxmou:
> does it affect `scratchr2`?

It doesn't change anything when 2.0 → 3.0 is enabled.

addons/dark-www/experimental_scratchr2.css
mxmou: This bug was never visible with `dark-www` enabled. I had to add `background-image: none` to fix it in this PR because I removed the gradient from purple buttons.

mxmou:
> There's probably some colors that still don't match 100% the old blues, such as the "post comment" button on profiles. But I'm not really sure.

The only place where the 3.0 blue is used on scratchr2 pages after this PR is the border of focused inputs.

> Yeah, normally there's separators between stuff in the navbar, and a gradient for the follow discussion button, which you can see through here: https://web.archive.org/web/20141015105751/http://scratch.mit.edu/discuss/topic/10588

I decided to only change the colors in this PR because making other visual changes isn't really the purpose of this addon.

addons/dark-www/experimental_scratchr2.css
mxmou: Fixed.

apple502j:
Hmm, why is the schema check successful? Did someone edit it?

mxmou:
> Hmm, why is the schema check successful? Did someone edit it?

Apparently adding a new property doesn't cause it to fail.

apple502j (changes requested):
Blocking merge pending manifest update by me.

apple502j:
@mxmou 
> The options of a "type": "map" value provider can now also be value providers.

Apparently the schema already allows this:

image

mxmou:
> @mxmou
> 
> > The options of a "type": "map" value provider can now also be value providers.
> 
> Apparently the schema already allows this:
> 
> (image)

That's weird. It didn't actually work before this PR.

apple502j (changes requested):

content-scripts/cs.js
```diff
@@ -279,7 +279,7 @@ function setCssVariables(addonSettings, addonsWithUserstyles) {
         // this is not even a color lol
         return getColor(addonId, obj.source) ? getColor(addonId, obj.true) : getColor(addonId, obj.false);
       case "map":
-        return obj.options[getColor(addonId, obj.source)];
+        return getColor(addonId, obj.options[getColor(addonId, obj.source)]) || getColor(addonId, obj.default);
```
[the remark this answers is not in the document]
apple502j: @mxmou I don't know what the default is supposed to do, is it a fallback for a failed lookup or is it a fallback for a failed resolution? For example, if the option's mapped result was a provider which ended up returning null, should it fall back? (I doubt so)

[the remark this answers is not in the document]
mxmou: The fallback is only meant to be used if the value of `source` isn't in `options`.

apple502j (approved):

PR #6332, opened 2023-07-02
Add a math extension

kokofixcomputers:
### Changes

I've added a addon named Math Extension it adds a extension to scratch named Math

### Reason for changes

I added these changes so that I could do more math on scratch. Warning: other people without this math extension cannot run this project please add this warning. Other people could also add a bookmark on this url: javascript:fetch('https://raw.githubusercontent.com/LoganAbel/ScratchMath/main/Math.js').then(r=>r.text()).then(t=>eval(t))

### Tests

Tested on Brave (Chrome)

### DIFF
--- a/addons/3d-extension/addon.json
+++ b/addons/3d-extension/addon.json
@@ -1,6 +1,12 @@
 {
     "name": "3D Addon",
     "description": "Loads a 3D extension.",
+    "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects", "https://scratch.mit.edu/",]
+    }
+  ],
     "tags": ["editor"],
     "enabledByDefault": true
   }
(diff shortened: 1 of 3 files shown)


CST1229:
We don't accept addons that do things like add new blocks that don't work for people not using the addon.

DNin01:
And as well, I don't think fetching JavaScript from a URL and then injecting it with a browser extension would be a good idea for security. I think this is called remotely hosted code? Instead, the JavaScript should be copied from the source and inserted into the file.

In the future, it would be a good idea to open an issue before opening a PR so we can discuss your ideas first.

WorldLanguages:
Anyway, adding extensions to the editor is out of scope. Most progress in that matter is happening on TurboWarp.

PR #6337, opened 2023-07-02
editor-dark-mode: change the color of icons in paint editor

### LINKED ISSUE #6299: editor-dark-mode does not change color of costume editor buttons
### Why this would be helpful

From feedback
Not sure if this was mentioned already in some issue these last few days, sorry

### My suggestion

image

> in the costume editor the buttons arent changed. 

Screenshot: "Scratch 3.0 default colors (blue)" preset

### Possible alternatives

_No response_

### Additional context

_No response_

mxmou:
Resolves #6299

### Changes

Changes paint editor icons based on editor dark mode settings.

### Reason for changes

This makes the icons blue in blue presets:
image

It also makes them look better in dark mode:
image

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/editor-dark-mode/paint_icons.js
+++ b/addons/editor-dark-mode/paint_icons.js
@@ -1,0 +1,70 @@
+import { textColor } from "../../libraries/common/cs/text-color.esm.js";
+
+const iconSelector = "[class*='labeled-icon-button_edit-field-icon_']";
+const dataUriRegex = new RegExp("^data:image/svg\\+xml;base64,([A-Za-z0-9+/=]*)$");
+
+export default async function ({ addon, console }) {
+  const recolorSvg = (svg) => {
+    return svg
+      .replace(/#855cd6/gi, addon.settings.get("highlightText"))
+      .replace(/#ffffff|#fff|white/gi, addon.settings.get("accent"))
+      .replace(/#575e75/gi, textColor(addon.settings.get("accent")));
+  };
+
+  const srcToSvg = async (src) => {
+    const match = dataUriRegex.exec(src);
+    if (match) return atob(match[1]);
+    return await (await fetch(src)).text();
+  };
+
+  const updateIcon = async (icon) => {
+    let oldSvg;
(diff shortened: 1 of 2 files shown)


WorldLanguages (approved):
Tested in Chromium. Also tested dynamicDisable

addons/editor-dark-mode/paint_icons.js
```diff
@@ -0,0 +1,70 @@
...
+      const oldSetAttribute = icon.setAttribute;
+      icon.setAttribute = function (name, value) {
+        if (name === "src") {
+          srcToSvg(value).then((svg) => {
+            icon.saOriginalSvg = svg;
+            updateIcon(icon);
+          });
+        }
+        return oldSetAttribute.call(this, name, value);
+      };
```
WorldLanguages, line 43: Maybe the `oldSetAttribute.call` line can be moved above the "if" so that the order of operations is more clear? Remember setAttribute always returns undefined.

WorldLanguages:
After this is approved/merged, we can release v1.33.3

Samq64 (approved):

WorldLanguages:
Potential bug (feedback):

> Super good! i like how we can change the things that were changed to purple back or to whatever! except in the ask block menu it wont change ***and it causes a rare bug in the custom editor where the reshape tool will replace some of the buttons with the paste icon.***

WorldLanguages:
Possibly React behaving differently in some edge case. I think this is the first time we receive this bug report so it probably isn't happening too often.

mxmou:
> Potential bug (feedback):
> 
> > Super good! i like how we can change the things that were changed to purple back or to whatever! except in the ask block menu it wont change _**and it causes a rare bug in the custom editor where the reshape tool will replace some of the buttons with the paste icon.**_

I found a way to reproduce this consistently - I used the network throttling option in the devtools to emulate a slow connection. The bug happens if you switch to the select tool, then very quickly to reshape. When the paste icon is loaded, it replaces the "pointed" icon.

PR #6353, opened 2023-07-03
Use non-deprecated API on share-through-mystuff

### LINKED ISSUE #6280: share-through-mystuff should not use a deprecated REST API
This addon indirectly uses the `scratch.mit.edu/site-api` API instead of `api.scratch.mit.edu` for sharing projects

WorldLanguages:
Resolves #6280

### Changes

Adds traps so that Scratch sends a request to the REST API instead of the deprecated one, which is never used by vanilla Scratch to share projects.
This means the addon continues to work exactly like before, but the request is going to the intended URL. Checking whether the request succeeded or failed is still handled by Scratch, just like before.

### Tests

Tested in Chromium

### DIFF
--- a/addons/share-through-mystuff/userscript.js
+++ b/addons/share-through-mystuff/userscript.js
@@ -20,7 +20,31 @@ export default async function ({ addon, console, msg }) {
     event.preventDefault();
     const confirmation = await addon.tab.confirm(msg("confirmation-title"), msg("confirmation"));
     if (confirmation) {
-      event.target.parentElement.querySelector(".media-share").click();
+      const xToken = await addon.auth.fetchXToken();
+      const projectId = event.target
+        .closest(".media-item-content")
+        .querySelector(".media-thumb > a")
+        .href.match(/\d+/)[0];
+
+      const xhrOpen = XMLHttpRequest.prototype.open;
+      const xhrSend = XMLHttpRequest.prototype.send;
+      XMLHttpRequest.prototype.open = function (method, path, ...args) {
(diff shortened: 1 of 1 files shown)


Samq64 (approved):
Works the same as before on Firefox

apple502j:
Is there a reason as to why we don't call the API ourselves, instead of patching？

mxmou:
> Is there a reason as to why we don't call the API ourselves, instead of patching？

Most likely this:

> Checking whether the request succeeded or failed is still handled by Scratch, just like before.

WorldLanguages:
> Is there a reason as to why we don't call the API ourselves, instead of patching？

@apple502j Easiest way to get everything to work, I doubt we'll ever modify this addon again until My Stuff becomes a scratch-www page. Also, what mxmou said above.

addons/share-through-mystuff/userscript.js
```diff
@@ -20,7 +20,31 @@ export default async function ({ addon, console, msg }) {
...
+          this.setRequestHeader("X-Requested-With", ""); // Do not send this header
+          xhrSend.call(this); // Send empty body
+        };
+        return undefined;
+      };
+
+      event.target.parentElement.querySelector(".media-share").click(); // .click() is synchronous
+      // By this point, the request has already been sent. Remove traps.
+      XMLHttpRequest.prototype.open = xhrOpen;
+      XMLHttpRequest.prototype.send = xhrSend;
```
WorldLanguages, line 47: This line (47) is unnecessary, as prototype.send is never changed. The `send` property of an object is changed in line 37, not of the prototype (or "class")

PR #6357, opened 2023-07-04
dark-www: add two 2.0 presets

### LINKED ISSUE #6318: 2.0 colors preset for website colors
### Why this would be helpful

People might want to be able to bring back the old 2.0 colors from before the update, maybe even on 3.0 pages.

### My suggestion

There could be a preset to load these easily, but here's some colors it could use (this is what I have as my settings currently):
Navbar color: #0F8BC0
Link color: #1AA0D8
Highlight color: #19b6fa, #169FDB

### Possible alternatives

2.0 <-> 3.0?

### Additional context

_No response_

mxmou:
Resolves #6318

### Changes

Adds two new presets to `dark-www`. They both look identical to "Scratch's default colors (blue)" on scratchr2 pages. "Scratch 2.0 (2013)" is based on the old scratchr2 colors:
image

"Scratch 2.0 (2015)" is the color scheme that was used by scratch-www before the 3.0 beta was released:
image

### Reason for changes

To make it easier to use 2.0 colors on scratch-www pages.

### Tests

Tested on Edge and Firefox.


addons/dark-www/addon.json
```diff
@@ -2333,6 +2355,44 @@
...
+        "blue": "#e6f3fa",
+        "input": "#fafafa",
+        "button": "#25aff4",
+        "link": "#25aff4",
+        "footer": "#f2f2f2",
+        "border": "#0000001a",
+        "darkBanners": "off",
+        "darkForumCode": false
+      }
+    },
```
DNin01, line 2395: I feel like some people may not remember this one. I remember that change, but it's pretty subtle and by now, I don't think anyone would prefer it over the deeper blue (the one before mid-2023). I would skip this preset as to not add too many, but that's just my opinion.

WorldLanguages: I agree about the 2013 preset being more important. Not sure if that means we shouldn't add this one, though.

addons/dark-www/addon.json
```diff
@@ -2333,6 +2355,44 @@
         "darkForumCode": true
       }
     },
+    {
+      "name": "Scratch 2.0 (2013)",
```
WorldLanguages, line 2359: Could we rename this to just `Scratch 2.0`?
These colors are used to this day on some pages, and if someone wants a 2.0 preset, this one is probably the one they want

WorldLanguages: The other one could be kept as `Scratch 2.0 (2015)` or possibly it could be renamed to `Scratch 2.0 (2015-2018)`

WorldLanguages: > These colors are used to this day on some pages, and if someone wants a 2.0 preset, this one is probably the one they want

Well, I'm not the one to judge this, but I assume someone looking for the "2.0 preset" will want the navbar in the front page to have the exact same color as the navbar on My Stuff?

addons/dark-www/addon.json
```diff
@@ -2333,6 +2355,44 @@
...
+        "button": "#18abeb",
+        "link": "#1aa0d8",
+        "footer": "#ececec",
+        "border": "#0000001a",
+        "darkBanners": "off",
+        "darkForumCode": false
+      }
+    },
+    {
+      "name": "Scratch 2.0 (2015)",
```
DNin01, line 2378: Shouldn't this be "Scratch 3.0 (2015)"? I get that these are the scratch-www colors during version 2.0, but considering that they take effect on scratch-www, they would be the Scratch 3.0 colors at the time.

WorldLanguages: These are the colors scratch-www used while Scratch 2.0 was the available online editor.... This is painful lol

medlans: > These are the colors scratch-www used while Scratch 2.0 was the available online editor.... This is painful lol

LOL
https://web.archive.org/web/20151027002031/https://scratch.mit.edu/

mxmou: This color scheme was indeed used in early concepts for Scratch 3.0, such as [this one](https://github.com/scratchfoundation/scratch-gui/issues/1#issuecomment-250742655) from 2016. But the 3.0 beta was released in 2018 and the scratch-www color scheme was updated at the same time, so I don't think it makes sense to call the old one "3.0".

addons/dark-www/addon.json
WorldLanguages, line 2359:
```suggestion
      "name": "Scratch 2.0",
```

DNin01:
Okay, there goes misconfigured Prettier again...

Better temporarily disable the workflow, revert the commit, pull from upstream, and then re-enable the workflow.

WorldLanguages (approved):

apple502j (approved):

PR #6362, opened 2023-07-05
Fix prettier

Samq64:
Apparently Prettier v3 was released recently and it changes the default trailing whitespace valeu to all. This PR sets it back to the previous defualt. Untested.


Samq64:
It's still formating the whole project for no reason.

.github/workflows/format-code.yml
```diff
@@ -1,5 +1,5 @@
 name: Format code using Prettier
-on: [push, pull_request]
+on: workflow_dispatch
```
DNin01, line 2: You can also disable Actions to prevent automatic formatting in the meantime.

Samq64: Yeah, I know I just fogot to. I'm making it manual because WorldLanguages said he doesn't mind if it isn't automatic on Discord. I can undo that through.

Samq64:
I'll revert those commits in an hour once I'm not stuck on mobile.

.github/workflows/format-code.yml
apple502j:
```suggestion
on: [push, pull_request]
```

apple502j (approved):

apple502j:
Prettier action has always formatted every file, even if only one file is changed.

.github/workflows/format-code.yml
RedGuy12: > Yeah, I know I just forgot to do it. I'm making it manual because WorldLanguages said he doesn't mind if it isn't automatic on Discord. I can undo that through.

He doesn't want the auto-commits, the check should still be there

RedGuy12 (approved):

Samq64:
> Prettier action has always formatted every file, even if only one file is changed.

No..?

apple502j:
@Samq64 Hm, really? I can't find the code that says so, and the logs indicate [it has always been like that](https://github.com/ScratchAddons/ScratchAddons/actions/runs/5446193494/jobs/9906566473).

Samq64:
It looks like this will still happen with the 100+ PRs with a branches older than this.

DNin01:
> It looks the Prettier issues will still happen on the 100+ PRs with branches older than this.

We can ask pull request authors to sync their branches to pull down the config change.

PR #6371, opened 2023-07-07
Fix projects can be paused with Alt+X after disabling

### LINKED ISSUE #6369: Projects can still be paused/resumed with Alt+X after dynamically disabling `pause`
### Description

Usually, you aren't able to pause a project after dynamically disabling the Pause button addon because the button disappears. However, it is still possible to pause the project with the recently added Alt+X keyboard shortcut even after the addon is disabled.

I would assume this happens because the [event listener](https://github.com/ScratchAddons/ScratchAddons/blob/5212da5170762759f3d410bc81f30c79985ee57b/addons/pause/userscript.js#L18-L23) for Alt+X is not deactivated on dynamic disable.

This is a minor, non-disruptive issue and may or may not be worth fixing.

### Steps to reproduce

1. Enable `pause` and open a project.
2. Disable `pause`.
3. Press Alt+X.

_No response_

### Environment

N/A

WorldLanguages:
Resolves #6369

### Changes

Fixes

>Usually, you aren't able to pause a project after dynamically disabling the Pause button addon because the button disappears. However, it is still possible to pause the project with the recently added Alt+X keyboard shortcut even after the addon is disabled.

### Tests

No

### DIFF
--- a/addons/pause/userscript.js
+++ b/addons/pause/userscript.js
@@ -16,7 +16,7 @@ export default async function ({ addon, console, msg }) {
   onPauseChanged(setSrc);
 
   document.addEventListener("keydown", function (e) {
-    if (e.altKey && e.code === "KeyX") {
+    if (e.altKey && e.code === "KeyX" && !addon.self.disabled) {
       e.preventDefault();
       setPaused(!isPaused());
     }


DNin01 (approved):
Not tested, but it looks like this will work.

Samq64 (approved):

PR #6399, opened 2023-07-15
Don't collapse search bar on icon click

DNin01:
### Changes

- When clicking the magnifying glass icon, which is actually a button to search, the search bar now stays expanded if it was expanded already. You still need to focus the search bar input for it to expand, to let you know when you can type into the search bar.
- A side effect of this change is that pressing Shift+Tab now moves to the Scratch logo instead of "About" on the navbar, so the way the links are hidden has been changed so pressing Shift+Tab (to move to the previous element) when navigating with a keyboard still moves focus to "About". Keyboard usability is useful not only for accessibility but also efficiency (I could imagine people Tabbing to the search bar all the time).

### Reason for changes

- Prevents accidents if someone decides to click the magnifying glass.

#### Without this change

1. Click the search bar. Optionally type in a search query.
2. Try to click the magnifying glass icon. The search bar will collapse before you can activate it.

#### With this change

1. Click the search bar and type in a search query.
2. Click the magnifying glass icon. The search bar will stay expanded and you will be sent to the search results page.
3. Press Tab until focus moves onto the search bar. The search bar will expand once the search input is focused.
4. Press Shift+Tab until focus moves onto the "About" button. The search bar will collapse once focus moves out of the _entire_ search bar element.

### Tests

Tested on Edge 114 in combination with `discuss-button`.

### DIFF
--- a/addons/expanding-search-bar/expanding-search-bar.js
+++ b/addons/expanding-search-bar/expanding-search-bar.js
@@ -51,7 +51,7 @@ export default async function ({ addon, console }) {
             ahref.style.setProperty("padding", "0px", "important");
             let span = exsearch_links[i].querySelector("span");
             span.style.display = "none";
-          } //Hide them. (Make them invisible but still keyboard-focusable)
+          } //Hide them (make them invisible but maintain keyboard-focusablility)
         }
       }
       function exsearch_clickOut() {
@@ -89,7 +89,7 @@ export default async function ({ addon, console }) {
       function exsearch_clickIn() {
         //Clicking into the search bar
         if (addon.self.disabled) return; //Don't expand if addon disabled
-        exsearch_siteNav.style.width = "0px"; //Hide the site navigation
+        exsearch_siteNav.style.width = "0px"; //Hide the site navigation (make them invisible but maintain keyboard-focusablility)
       }
       function exsearch_clickOut() {
         //Clicking out of  the search bar
--- a/addons/expanding-search-bar/expanding-search-bar.js
+++ b/addons/expanding-search-bar/expanding-search-bar.js
@@ -2,11 +2,19 @@ export default async function ({ addon, console }) {
   if (addon.tab.clientVersion === null) return; //if neither www or r2, exit
 
   var exsearch_searchBar; //The search bar element
+  var exsearch_searchBarInput; //The input portion of the search bar element
 
   //Wait for the search bar to load (we get the header links as we hide/show them, no need to worry about that)
   //Also we set the search bar value here too
   while (true) {
-    exsearch_searchBar = await addon.tab.waitForElement(
+    exsearch_searchBar = await addon.tab.waitForElement(".search", {
+      markAsSeen: true,
+      reduxCondition: (state) => {
+        if (!state.scratchGui) return true;
+        return state.scratchGui.mode.isPlayerOnly;
+      },
+    });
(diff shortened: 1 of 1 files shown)


addons/expanding-search-bar/expanding-search-bar.js
```diff
@@ -69,15 +89,19 @@ export default async function ({ addon, console }) {
       function exsearch_clickIn() {
         //Clicking into the search bar
         if (addon.self.disabled) return; //Don't expand if addon disabled
-        exsearch_siteNav.style.display = "none"; //Hide the site navigation
+        exsearch_siteNav.style.width = "0px"; //Hide the site navigation
```
WorldLanguages, line 92: What's the difference between display:none and setting the width to 0px?

DNin01: > A side effect of this change is that pressing Shift+Tab now moves to the Scratch logo instead of "About" on the navbar, so the way the links are hidden has been changed so pressing Shift+Tab (to move to the previous element) when navigating with a keyboard still moves focus to "About". Keyboard usability is useful not only for accessibility but also efficiency (I could imagine people Tabbing to the search bar all the time).

DNin01: To clarify, setting "display" to "none" makes an element not keyboard-focusable, but setting "width" to "0px" keeps its keyboard-focusability.

WorldLanguages: Maybe add a small comment to this line that says "allow keyboard focus with Tab key"?

DNin01: Done!

WorldLanguages (approved):
Fixes bug in Firefox

WorldLanguages:
Well, you didn't say it's a "bug".
Why wouldn't you consider it one?

DNin01:
> Well, you didn't say it's a "bug". Why wouldn't you consider it one?

Is it a "bug"? I thought it was more like an oversight - a missing feature, but it was definitely unintentional. You can change the label if you want.

WorldLanguages:
Good to merge?

DNin01:
@WorldLanguages 👍

WorldLanguages:
Potential regression (feedback):

> When searching something on the search bar, the search bar glitches/expands through the Create, Ideas, Explore, etc.. tabs.
OS: Windows 11
Browser: Microsoft Edge 115.0.1901.203 (Official build) (64-bit version)
Scratch Addons: 1.34.0

Ask me for the username if necessary, we can ask them for more info/details.

DNin01:
Not exactly sure what they're talking about, but it is worth noting that I didn't bother to actually "hide" the navbar buttons on `scratchr2` pages, as opposed to on `scratch-www` where they become zero-width. You can see [here](https://github.com/ScratchAddons/ScratchAddons/pull/6399/files#diff-7c51ce6b1b7c7bc630ad63c5dd03a20699740681d4268dac71d4b3b525635dd8R86) that only the navigation bar's width is set to 0px - the buttons stay the same. This doesn't cause any problems for me but I was able to hover over the Create button while the search bar was expanded only on `scratchr2` pages.

The next thing I would ask for is a screenshot or video.

WorldLanguages:
Firefox:

image

WorldLanguages:
Moved to #6567

Issue #2271, opened 2021-04-23
the hoverable profile picture card


FunctionalMetatable:
**Describe the new addon**

you know the cards that come up when you hover over a users profile picture, e.g
![image](https://user-images.githubusercontent.com/67440879/115899757-0cdf4180-a457-11eb-9608-d3386de5bf1e.png)
it could be implemented in scratch

TheColaber:
Duplicate of #1391 but since it's a discussion and not an issue, this can be left open.

mxmou:
I started working on this addon a while ago, not only for users but also for projects, comments and forum posts. I might finish it after the website dark mode rewrite.

Hans5958:
> 
> 
> I started working on this addon a while ago, not only for users but also for projects, comments and forum posts. I might finish it after the website dark mode rewrite.

What a coincidence that I also tried to create a similar concept. Guess we should create the issue earlier so no duplicate work is wasted.

```js
// ==UserScript==
// @name         Scratch Tooltips Concept
// @namespace    http://tampermonkey.net/
// @version      0.1
// @description  try to take over the world!
// @author       You
// @match        https://scratch.mit.edu/*
// @grant        none
// @require      https://cdn.jsdelivr.net/npm/@popperjs/core@2.7/dist/umd/popper.js
// @require      https://cdn.jsdelivr.net/npm/tippy.js@6/dist/tippy-bundle.umd.js
// ==/UserScript==

// TODO:
// Interactive tooltip (clickable project name and link)
// Adjustable text positions
// Use classes instead of inline style?
// Add support for studios and users
// Add info such as loves, favs, views, etc
// Options whether to also do it with thumbnails
// Update for new elements

const triggerTooltips = () => {

    var projects = [...document.querySelectorAll("a[href^='https://scratch.mit.edu/projects/']:not(.tooltips-read)"), ...document.querySelectorAll("a[href^='/projects/']:not(.tooltips-read)")]
    if (projects.length === 0) return
    console.log(projects)

    projects.forEach(element => {
        element.classList.add("tooltips-read")
        if (/\d+/.exec(element.href) === null) return
        if (element.children.length !== 0) return
        var id = /\d+/.exec(element.href)[0]
        // https://api.scratch.mit.edu/projects/
        var tippyOptions = {
            content: 'Loading...',
            allowHTML: true,
            popperOptions: {
                strategy: 'fixed',
                modifiers: [
                    {
                        name: 'flip',
                        options: {
[...]

PR #6400, opened 2023-07-15
New addon: Customizable link style

### LINKED ISSUE #6374: Colorblindness accommodations addon
### Why this would be helpful

There are a number of things on Scratch that can be improved to be more accessible to colorblind users.

### How this addon works

This addon would provide accommodations for colorblind users.

**Link style**
The most notable setting in this addon will the ability to change the style of links. Colorblind users can have a hard time differentiating links from normal text. There would be two options for link text: bold and underlined.

**Contrast enhancements**
Some elements can also be harder for colorblind users to see. For example, the toggled love and favorite icons aren’t the best in terms of contrast. This setting could darken certain elements so they are accessible.

Any more you can think of?

All of these features could be turned off as well, so if you don’t need a certain accommodation, you don’t have to apply it.

### Possible alternatives

Add these features to `dark-www`.

### Additional context

_No response_

DNin01:
Resolves #6374

### Changes

Starts a new addon that offers new style options for links. This can be useful for colorblind users or if you just like styled links.

### Tests and Issues

Tested in Edge 115.

- [ ] Test together and fix all conflicts with "Scratch 2.0 → 3.0" addon


WorldLanguages:
Related: ["As a colorblind person, I can't see links"](https://scratch.mit.edu/discuss/topic/694496/)

DNin01:
I added some new options, but I haven't figured out how to make it possible to customize the color of the underlined text with `dark-www` and the opacity at the same time. We might need to add new APIs.

WorldLanguages:
@DNin01 Could you elaborate on what exactly you are trying to make `dark-www` control and why it isn't working?

DNin01:
> @DNin01 Could you elaborate on what exactly you are trying to make `dark-www` control and why it isn't working?

I added a setting to adjust the opacity of the underline on links when Link style is set to Underline. There are no options to change the color, so I simply fill in A in HSLA with the setting ID, as you can see here:
https://github.com/ScratchAddons/ScratchAddons/blob/650954cf0e054cde8437fbd446b5265cd7272688/addons/colorblind/underline-links.css#L18
It works great when only this addon is enabled. However, things get more complicated when trying to implement the same opacity feature in combination of `dark-www`, a **different addon** that can change link colors.

I had to not only specify an alpha value for the underline, but also the entire color value, meaning that in order to change the color, you have to modify that HSLA value. You can't use a separate declaration, AFAIK. So, if I want to make it possible for `dark-www` to change this color while keeping `colorblind` in control of the opacity, I would have to somehow squish together a hex code from `dark-www` representing the color of links with the opacity setting from `colorblind`, all in CSS and the addon.json. At least, that's what I'm thinking.

If you know of a way to achieve that, that would be great. I don't know how to do that and I don't want to have to use a JS solution unless it is required.

WorldLanguages:
I guess the easiest way would be `dark-www` splitting the color setting into three different integer custom-CSS-variables, one for red, one for green, and one for blue. That's probably not yet possible with the `customCssVariables` API, though.

DNin01:
> I guess the easiest way would be `dark-www` splitting the color setting into three different integer custom-CSS-variables, one for red, one for green, and one for blue. That's probably not yet possible with the `customCssVariables` API, though.

That idea did go through my head, but it would probably be more feasible to think of a more compact solution.

WorldLanguages:
Well, it's weird to think that one addon is responsible for the RGB values and another one handles the transparency separately.
But as far as I understood, this only determines the transparency of the underline, and the use-case is different, so it makes sense to have this as a separate addon. You probably want to justify this in the original post after making this a non-draft though

Tacodiva:
Not sure if this is the right place to put this suggestion, but I too am colourblind. Scratch is generally very accessible and there's not much I note with my colourblindness using it, especially compared to playing some games. 

I do have one idea though it's not really a big deal. Is seems to me like the the "stop" button should be more clearly either clickable or not clickable. 

image
image

The flag gets a colored background which makes it very obvious, but the stop button does not which can mean that (for me at least) at a glance it's difficult to differentiate between the de saturated state and the full color state. Maybe a background could be added to the stop button too?

Secret-chest:
8-digit hexcodes with alpha exist.

mxmou:
> 8-digit hexcodes with alpha exist.

They aren't useful here because it isn't possible to use a variable for the alpha part. `color-mix()` would help, but it's only supported in Chrome 111+ and Firefox 113+:
```css
text-decoration-color: color-mix(in srgb, currentcolor calc(var(--colorblind-underlineAlpha) * 1%), transparent);
```

I don't think it's necessary to make the opacity customizable though - this feature is intended for people who need an underline to differentiate links from normal text, so an option to reduce the opacity isn't very useful.

DNin01:
Alright then, I'll probably remove the option entirely. Simple. We can always reconsider it.

WorldLanguages:
> I don't think it's necessary to make the opacity customizable though - this feature is intended for people who need an underline to differentiate links from normal text, so an option to reduce the opacity isn't very useful.

Yeah, a boolean setting would probably be sufficient. But we still need to deal with the color mixing problem anyway.

WorldLanguages:
> The flag gets a colored background which makes it very obvious, but the stop button does not which can mean that (for me at least) at a glance it's difficult to differentiate between the de saturated state and the full color state. Maybe a background could be added to the stop button too?

@Tacodiva To be honest I never really noticed the stop button changed color, very interesting. Though it looks like the stop button is only clickable only when the green flag has a background - it's not independent. So your suggestion would be to add a background to the stop button too, not just to the flag, in the same exact situations. Sounds good?

WorldLanguages:
> Yeah, a boolean setting would probably be sufficient. But we still need to deal with the color mixing problem anyway.

Nevermind, I think?

addons/colorblind/addon.json
```diff
@@ -0,0 +1,90 @@
...
+      "id": "contrast-switch",
+      "type": "boolean",
+      "default": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "updateUserstylesOnSettingsChange": true,
+  "versionAdded": "1.34.0",
+  "tags": ["theme"],
```
DNin01, line 88: Seems like I have to either specify that this is a website theme or an editor theme. Have we ever made an addon that applies to both the editor and the website?

Samq64: Yes, but not often and I think all of them have been split.

DNin01:
I was able to come up with a clever solution involving a CSS variable that overrides `text-decoration: none`[^1] when "link style" is set to "underlined", so I don't have to declare `text-decoration: underline !important` in this addon for things like hovered links.

[^1]: https://github.com/ScratchAddons/ScratchAddons/blob/f8585633b2cd73f99741fcca5110599a8fbe4ccf/addons/scratchr2/scratchr2.css#L45

apple502j:
Question, what does this addon do exactly, and why can't this just be a preset of editor-dark-mode/dark-www? I don't like this kind of "fix everything" style addon.

Also, would like opinions from actual colorblind people.

Tacodiva:
I would agree that this should just be a part of the editor theme, and a colorblind preset could be really cool.

Has anybody with colorblindess complained about the editor toggles? Obviously, I don't have every type of color-blindess but I can definitely say that for me, the links are hard to see but the editor buttons are pretty obvious.

DNin01:
This addon serves a different specific purpose than the customizable-colors addons, which is why I put these features in a new addon.

One could say that customizing colors is one of the first things a colorblind user would look for, though, which is also a good point.

DNin01:
I removed the high contrast switches option so now we don't have to worry about an addon for both the website and the editor.

WorldLanguages:
> * **Link style**: Change the style of links around the website.

I would suggest merging only this feature for this version in a few days (if possible) and leaving the high contrast love and favorite icons for later. (it is also quite confusing as Scratch itself has updated those icons for accessibility already, so I'm not sure what to think about it)

WorldLanguages:
Could we try to merge the link style settings (probably not the love/favorite icons feature), maybe with the "beta" tag? I think it would be a good opportunity to see how users react to it, and we may be able to get more feedback on potential accessibility enhancements.

DNin01:
What do you think of the underline style feature (choosing between solid, dashed, and dotted)? Should it be removed?

Secret-chest:
I think dotted underlines represent abbr, which Scratch does not currently use.

Secret-chest:
I think dotted underlines represent abbr, which Scratch does not currently use.

mxmou:
I think this feature will be useful for people who aren't colorblind too. Could the addon be called something like "customizable link style"?

WorldLanguages:
> I think this feature will be useful for people who aren't colorblind too. Could the addon be called something like "customizable link style"?

Yeah, we should rename it to something like that, at least for this release.

WorldLanguages:
> What do you think of the underline style feature (choosing between solid, dashed, and dotted)? Should it be removed?

I think they are okay, as long as solid is the default (when choosing "underlined")

addons/colorblind/addon.json
```diff
@@ -0,0 +1,92 @@
+{
+  "name": "Color blindness accommodations",
```
WorldLanguages, line 2:
```suggestion
  "name": "Customizable link style",
```

addons/colorblind/addon.json
```diff
@@ -0,0 +1,92 @@
...
+        {
+          "id": "bold",
+          "name": "Bold"
+        },
+        {
+          "id": "underline",
+          "name": "Underlined"
+        }
+      ],
+      "default": "bold"
```
WorldLanguages, line 60: I think "bold" is the option that makes the least sense.

We can choose "default" and make this behave like other addons named "customizable" (no changes by default), or have "underline" as a default which lets you easily notice what the addon changed after enabling it

addons/colorblind/addon.json
```diff
@@ -0,0 +1,92 @@
+{
+  "name": "Color blindness accommodations",
+  "description": "A few features to help make Scratch more accessible to colorblind users.",
```
WorldLanguages, line 3: We can keep the words "color blindness" in the description so that the old name of the addon is still searchable.

mxmou: Is that really necessary? The addon was never released with the previous name.

WorldLanguages: It is not necessary

DNin01:
I'm keeping the beta tag because we haven't tested this everywhere.

WorldLanguages (approved):
Looks good to me

addons/colorblind/addon.json
```diff
@@ -0,0 +1,98 @@
...
+  "credits": [
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
+    }
+  ],
+  "info": [
+    {
+      "id": "feedback",
+      "text": "Submit feedback about this addon (or other potential accessibility features) by going to scratchaddons.com/feedback or clicking \"send feedback\" on the sidebar."
```
DNin01, line 13:
```suggestion
      "text": "We want to expand this addon with other accessibility features in the future. You can send us your feedback by going to scratchaddons.com/feedback or clicking \"Send Feedback\" on the sidebar."
```

Secret-chest:
Why can't be both bold and underline be enabled?

WorldLanguages:
> Why can't be both bold and underline be enabled?

Huh good point

WorldLanguages:
Last day to modify this PR !

WorldLanguages:
Not very useful feedback:

> For the customizable link style, I would suggest adding a option to use the name of the project as the link, that would be SUPER useful!

Actually this is more related to #2271 than this

PR #6414, opened 2023-07-19
Find bar: Add F3 shortcut and use e.key

### LINKED ISSUE #6413: F3 to go to next in Editor find bar
### Why this would be helpful

F3 is the standard for "Find Next", and Shift+F3 is for "Find Previous".

### My suggestion

- Offering F3 and Shift+F3 as an alternative to arrow keys to find next/previous with the find bar.
- Being able to use F3 even when the find bar isn't focused, so the user can find next/previous without having to click the find bar first.

### Possible alternatives

_No response_

### Additional context

_No response_

DNin01:
Resolves #6413

### Changes

Updates Editor find bar:
- You can now use F3 to go to the next item and Shift+F3 to go to the previous item while using the find bar.
- Updated uses of `e.keyCode` to `e.key` because `keyCode` is deprecated.

### Tests

Tested in Edge 114.

### DIFF
--- a/addons/find-bar/userscript.js
+++ b/addons/find-bar/userscript.js
@@ -116,14 +116,12 @@ export default async function ({ addon, msg, console }) {
     inputKeyDown(e) {
       this.dropdown.inputKeyDown(e);
 
-      // Enter
-      if (e.keyCode === 13) {
+      if (e.key === "Enter") {
         this.findInput.blur();
         return;
       }
 
-      // Escape
-      if (e.keyCode === 27) {
+      if (e.key === "Escape") {
         if (this.findInput.value.length > 0) {
           this.findInput.value = ""; // Clear search first, then close on second press
           this.inputChange();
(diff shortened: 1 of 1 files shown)


Samq64:
F3 and Shift+3 seem to just bring up the browser's search box when the editor one isn't focused. All the other changes seem to work fine though, but I'm using QWERTY so someone else should try with a different layout just to be sure.

DNin01:
> F3 and Shift+3 seem to just bring up the browser's search box when the editor one isn't focused. All the other changes seem to work fine though, but I'm using QWERTY so someone else should try with a different layout just to be sure.

Yeah, I know. I would've liked to support F3 and Shift+F3 without selecting the Find bar first, but I decided to start here. Probably need to learn more about JS and Scratch internals before I can do that.

WorldLanguages:
> Resolves #6413

It does not, as you said F3 only works if the find bar is focused.

WorldLanguages:
The `eventKeyDown` function in the same file listens to key events for the whole page.

DNin01:
> The `eventKeyDown` function in the same file listens to key events for the whole page.

I found that but couldn't figure out how to write the code to jump between blocks while the find bar is closed. I am still learning JS.

WorldLanguages:
> I found that but couldn't figure out how to write the code to jump between blocks while the find bar is closed.

It might not be possible without refactoring a lot of code, depending on how it was implemented.

DNin01:
Partially resolved by #6923. I may start a new PR to reimplement my other plans if I feel like it.

PR #6205, opened 2023-06-14
Strip localhost permission and include DNRWHA in generated manifest

apple502j:
Resolves #6084

Three changes:

- localhost and `scratchfoundation.github.io` permissions are now removed during gen-manifest execution, from both the permissions section and the CS section.
- In Firefox, DNRWHA permission is no longer removed. FF113+ supports DNRWHA, and development build (without this being removed) works fine.
- Linter configuration changes.

PR #6421, opened 2023-07-21
Run addons locally on scratch-gui & `scratchfoundation.github.io`

WorldLanguages:
Resolves (the technical aspect of) #5281

### Changes

Most changes aren't new and they were part of #6053

### Reason for changes

Since #6205 was merged, we can keep all of these changes on the codebase. The `manifest.json` changes will not be released to extension store users.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -404,7 +404,7 @@ function userscriptMatches(data, scriptOrStyle, addonId) {
   let _parsedURL = new URL(_url);
   if (_parsedURL.origin === "https://scratchfoundation.github.io" || _parsedURL.port === "8601") {
     // Run addons on scratch-gui
-    _url = "https://scratch.mit.edu/projects/104/editor";
+    _url = "https://scratch.mit.edu/projects/editor/";
     _parsedURL = new URL(_url);
   }
   const url = _url;
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -401,11 +401,11 @@ function userscriptMatches(data, scriptOrStyle, addonId) {
   if (scriptOrStyle.if && !matchesIf(scriptOrStyle, scratchAddons.globalState.addonSettings[addonId])) return false;
 
   let _url = data.url;
-  let _parsedURL = new URL(url);
+  let _parsedURL = new URL(_url);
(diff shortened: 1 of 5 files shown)


background/get-userscripts.js
```diff
@@ -400,14 +400,21 @@ function matchesIf(injectable, settings) {
 function userscriptMatches(data, scriptOrStyle, addonId) {
   if (scriptOrStyle.if && !matchesIf(scriptOrStyle, scratchAddons.globalState.addonSettings[addonId])) return false;
 
-  const url = data.url;
-  const parsedURL = new URL(url);
+  let _url = data.url;
+  let _parsedURL = new URL(_url);
+  if (_parsedURL.origin === "https://scratchfoundation.github.io" || _parsedURL.port === "8601") {
+    // Run addons on scratch-gui
+    _url = "https://scratch.mit.edu/projects/104/editor";
```
apple502j, line 407: It makes more sense to run as a logged-out editor page.

content-scripts/prototype-handler.js
```diff
@@ -23,7 +23,9 @@ function injectPrototype() {
   };
 }
 
-if (!(document.documentElement instanceof SVGElement) && location.pathname.split("/")[1] === "projects") {
+const isLocal =
+  location.origin === "https://scratchfoundation.github.io" || location.origin.startsWith("http://localhost:");
+if ((!(document.documentElement instanceof SVGElement) && location.pathname.split("/")[1] === "projects") || isLocal) {
```
apple502j, line 28: Locally run WWW instances should not generally trigger this.

apple502j (changes requested):

background/get-userscripts.js
WorldLanguages: Good point! I will change this to `https://scratch.mit.edu/projects/editor/`

content-scripts/prototype-handler.js
WorldLanguages: What do you think this should be changed to?
Checking location.port (which should match GUI port) instead of the origin?

apple502j: Yeah, the code looks good now

apple502j (approved):

PR #6422, opened 2023-07-21
New addon: full screen project ediotr

WorldLanguages:
Supersedes PR #5464
Some of the code in this PR was authored by @ErrorGamer2000 

### Changes

- Adds a fullscreen button to the editor menu bar. Similarly to the "enhanced full screen" addon, this enables browser-level full screen, hiding the address bar.
- Optionally, fullscreen mode can hide the editor menu bar unless it's being hovered (similarly to the auto-hide block palette addon)

Would also be interesting to provide settings that automatically shrink the stage while in fullscreen mode.

### Tests

- [ ] Compatibility with "enhanced full screen" addon, which also toggles browser-level fullscreen. For example, going to the project player's fullscreen mode and back to the editor should not exit browser-level fullscreen.
- [ ] Firefox testing


ErrorGamer2000:
Thanks for the ping

WorldLanguages:
One difference between #5464 and this, is that here we try to remember the stage mode before the user enabled fullscreen, in order to restore it after the user exists fullscreen.

Issue #5419, opened 2022-12-07
Revamp "infinite scrolling" addon


WorldLanguages:
### Why this would be helpful/solve a problem

We should revisit this old addon which has been made like 2 years ago. I think we could improve it a lot.

### What this change does

- Add infinite scrolling to more places, such as "What's happening" (if full-areas is enabled) and project search / explore (#2520)
- ~~Remove weird settings such as "fix footer", think of better ways to let users access the footer~~
- We removed support for studios at #3462

### Demonstration(s) and additional context

We need to be careful, because it's easy to add a bug that clicks "load more" infinitely, such as #2980

Hans5958:
About the footer, it's just does not make sense to access the footer if what you want is an infinite stream of content. That setting is just doesn't make any sense to me.

Hans5958:
The logic doesn't work here. 

Without a infinite scrolling feture, the reason of someone to scroll to the bottom is to access the footer, but with it, the reason of someone to scroll to the bottom is to have more stuff; more content.

- You don't scroll down a Scratch studio gallery or a YouTube playlist to access the footer. You want more projects from it.
- You don't scroll down a comments section (Scratch, YouTube) to access the footer. You want more comments of it.
- You don't scroll down the front page of YouTube, Twitter, Facebook, Reddit, TikTok, etc to access the footer. You want more content from it. This is why some sites, like YouTube or TikTok, put the content that should be on the footer, to the sidebar.

Even Never Ending Reddit, the "infinite scrolling" of [Reddit Enhancement Suite](https://redditenhancementsuite.com/) for old Reddit, doesn't offer a fixed footer, probably with the reason I mentioned.

If someone is often to access the footer, then they shouldn't using an infinite scrolling feature. A hybrid is nice, but it should put it where it is appropriate and nicely (TikTok and YouTube is appropriate), and I don't think there is a way to do it appropriately and/or nicely, including having a conspicuous fixed footer.

RedGuy12:
related: https://github.com/ScratchAddons/ScratchAddons/issues/3071

WorldLanguages:
> Remove weird settings such as "fix footer"

+1 from feedback

WorldLanguages:
> Remove weird settings such as "fix footer", think of better ways to let users access the footer

Any ideas? @DNin01

WorldLanguages:
Possibly we could do this:
- Remove the "fix footer" setting
[...]

PR #6426, opened 2023-07-22
Infinite scrolling addon: remove "fix footer" setting

WorldLanguages:
Resolves https://github.com/ScratchAddons/ScratchAddons/issues/5419#issuecomment-1345267393

### Changes

Removes the "fix footer" setting from infinite-scroll. **This setting has been the default for a while and I have seen no evidence of appreciation for it.**
Just like before with the setting disabled, the footer is now accessible normally in most pages. In pages that infinitely scroll, the footer is not hidden, and you can only "use" it if there's no more content to load, or if your internet connection isn't fast enough and more content hasn't loaded yet.

We could possibly hide the footer in pages where we're dynamically adding infinite content to the main viewport, this PR does not try to do that.

If users send feedback about wanting this feature back, we might readd it as an off-by-default setting.

### Reason for changes

See https://github.com/ScratchAddons/ScratchAddons/issues/5419#issuecomment-1345267393


WorldLanguages:
We don't normally remove settings, but this setting is enabled by default at the moment, so disabling it for everyone and changing the default is pretty much equivalent to removing it.

Samq64 (approved):

BroJac5246 (approved):

apple502j (approved):

PR #6433, opened 2023-07-23
editor-messages: use fetch instead of `addon.account.getMsgCount()`

### LINKED ISSUE #5991: Remove/revisit addon.* API method: `addon.account.getMsgCount()`
### Description

This addon.* API method is very old and is only used by 1 addon (editor-messages).
Apparently, editor-messages fetches the message count every 5 seconds, which was never the intended behavior. This was probably a regression at some point in the past.
We should consider removing this method or fixing its behavior.

### Steps to reproduce

1. Enable editor-messages addon
2. Go to the project editor
3. Open the Network tab from devtools. The message count is requested every 5 seconds

Apparently, Scratch's `Cache-Control` header means that not every of these requests actually reach the internet, but this is still unintended behavior.

This was brought up on Discord

### Environment

Scratch Addons version: v1.31.2

WorldLanguages:
Resolves #5991

### Changes

- The only userscript that used this API now uses fetch every 30 seconds (instead of using addon.account.getMsgCount every 5 seconds)
- Old addon.account.getMsgCount code was not removed but it was made inaccesible.

### Reason for changes

See #5991

### Tests

Addon still works

### DIFF
--- a/addons/editor-messages/userscript.js
+++ b/addons/editor-messages/userscript.js
@@ -9,9 +9,9 @@ export default async function ({ addon, console, msg }) {
   messageCount.classList.add("sa-editormessages-count");
   messages.appendChild(messageCount);
   const setMessages = async () => {
-    const msgCount = Number(await addon.account.getMsgCount());
-    messageCount.innerText = msgCount;
-    if (msgCount === 0) {
+    const { count } = await (await fetch("https://api.scratch.mit.edu/users/World_Languages/messages/count")).json();
+    messageCount.innerText = count;
+    if (count === 0) {
       messageCount.setAttribute("style", `display: none;`);
     } else {
       messageCount.setAttribute("style", "");
@@ -22,12 +22,12 @@ export default async function ({ addon, console, msg }) {
   function createInterval() {
     if (addon.tab.editorMode === "editor") {
       setMessages();
-      interval = setInterval(setMessages, 5000);
+      interval = setInterval(setMessages, 30_000);
     } else {
       addon.tab.addEventListener("urlChange", function thisFunction() {
         if (addon.tab.editorMode === "editor") {
           setMessages();
-          interval = setInterval(setMessages, 5000);
+          interval = setInterval(setMessages, 30_000);
           addon.tab.removeEventListener("urlChange", thisFunction);
         }
       });
--- a/addons/editor-messages/userscript.js
+++ b/addons/editor-messages/userscript.js
@@ -9,7 +9,9 @@ export default async function ({ addon, console, msg }) {
   messageCount.classList.add("sa-editormessages-count");
   messages.appendChild(messageCount);
   const setMessages = async () => {
-    const { count } = await (await fetch("https://api.scratch.mit.edu/users/World_Languages/messages/count")).json();
+    const username = await addon.auth.fetchUsername();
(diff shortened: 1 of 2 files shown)


addons/editor-messages/userscript.js
```diff
@@ -9,9 +9,9 @@ export default async function ({ addon, console, msg }) {
   messageCount.classList.add("sa-editormessages-count");
   messages.appendChild(messageCount);
   const setMessages = async () => {
-    const msgCount = Number(await addon.account.getMsgCount());
-    messageCount.innerText = msgCount;
-    if (msgCount === 0) {
+    const { count } = await (await fetch("https://api.scratch.mit.edu/users/World_Languages/messages/count")).json();
```
DNin01, line 12: Uh... we're not fetching the currently logged-in user's message count and instead fetching yours?

DNin01 (changes requested):
Something's not quite right here, see my comment.

Also, I think it might take longer than 30 seconds for the API to refresh the message counter due to caching.

addons/editor-messages/userscript.js
WorldLanguages: lol 😄

WorldLanguages: It works on my machine !

WorldLanguages:
> Also, I think it might take longer than 30 seconds for the API to refresh the message counter due to caching.

This is okay, actually it looks like Scratch is forcing this to stay in cache for 145 seconds

DNin01 (approved):
Haven't tested yet, but this looks okay.

addons/editor-messages/userscript.js
```diff
@@ -9,9 +9,11 @@ export default async function ({ addon, console, msg }) {
   messageCount.classList.add("sa-editormessages-count");
   messages.appendChild(messageCount);
   const setMessages = async () => {
-    const msgCount = Number(await addon.account.getMsgCount());
-    messageCount.innerText = msgCount;
-    if (msgCount === 0) {
+    const username = await addon.auth.fetchUsername();
+    if (!username) return;
+    const { count } = await (await fetch(`https://api.scratch.mit.edu/users/${username}/messages/count`)).json();
```
DNin01, line 14: There's also a double `await`. It might be unnecessary?

DNin01: [‎background/load-addon-manifests.js](https://github.com/ScratchAddons/ScratchAddons/blob/d2a9ce32afd85340702c271d579b7dc6ac455212/background/load-addon-manifests.js#L43) does the same thing, though.

WorldLanguages: Yep - `fetch` returns a promise, and so does `json` (which is only available in the result of the promise, not the promise itself)

DNin01: And: https://github.com/ScratchAddons/ScratchAddons/blob/d2a9ce32afd85340702c271d579b7dc6ac455212/addons/turbowarp-player/userscript.js#L34-L40

Never mind.

Samq64 (approved):

Samq64:
I'm assuming `addon.account.getMsgCount()` will be removed separately?

apple502j:
Yeah, this method is already unimplemented in popup.

PR #6446, opened 2023-07-24
More robust solution for editor-compact/columns compat

### LINKED ISSUE #6444: columns too tall when switching back to code editor with editor-compact
### Description

When switching to another editor tab and back with both the two-column category menu and compact editor enabled the category selector takes up way too much vertical space.

### Steps to reproduce

1. Enable `columns` and `editor-compact`
2. Open the editor
3. Switch to the costumes tab
4. Switch back to the code tab

image

### Environment

Browser: Firefox 115
Operating system: Linux

WorldLanguages:
Resolves #6444

### Changes

Adds a class to the `<body>` element instead of adding it to the `gui_tab-panel`

### Reason for changes

Easiest way to make sure we're not missing any edge cases where the panel is re-created. The body element is always the same.

### Tests

Tested in Chromium

### DIFF
--- a/addons/columns/userscript.js
+++ b/addons/columns/userscript.js
@@ -148,21 +148,21 @@ export default async function ({ addon, msg, console }) {
     toolbox.populate_(workspace.options.languageTree);
     // Reposition the toolbox, since it's likely our addon moved it.
     toolbox.position();
-
-    addClass();
   }
 
-  function addClass() {
+  function updateClass() {
     // Add class to allow editor-compact to handle this addon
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Related commit: https://github.com/lisa-wolfgang/ScratchAddons/commit/672197be1c51730ace29697245b2ee13a74cc2ef

Unfortunately this does make the CSS selector quite less "simpler" but it's much more robust, as there's no need for the attribute/class to be on any particular element (except simplifying the CSS selector for performance purposes or to make it easier to the human eye)

PR #6450, opened 2023-07-25
Remove addon.account API

### LINKED ISSUE #6443: Remove `addon.account` object from API
The only property/method available on this object is `getMsgCount` which is no longer working on any context

apple502j:
Resolves #6443

Removes `addon.account` API, the implementing code, and `Account.js` file. Also fixes a lint error in editor-messages addon, which is complaining the use of underscore (old browser compat I believe?)


addons/editor-messages/userscript.js
```diff
@@ -24,12 +24,12 @@ export default async function ({ addon, console, msg }) {
   function createInterval() {
     if (addon.tab.editorMode === "editor") {
       setMessages();
-      interval = setInterval(setMessages, 30_000);
```
WorldLanguages, line 27: Note that this syntax has wide support, going back to Chrome 75 and Firefox 70

WorldLanguages (approved):

addons/editor-messages/userscript.js
WorldLanguages: If possible change the ESLint config to support this

PR #3912, opened 2021-12-12
New Addon: `stage-monitor`

Samq64:
Resolves #3590
Resolves #1680 (The transparent values part, custom fonts is a different issue)

### Changes

Adds a new addon that allows customizing the colour and border radius of stage monitors and their values.

### Reason for changes

Some people don't like the orange colour of variable monitors. This addon also allows customizing stage monitors to make them look pretty cool be adding transparency.

### Tests

Tested on Chromium 115.

image

PR #6464, opened 2023-07-29
Fixes for editor and website customizable colors addons

### LINKED ISSUE #6341: Website customizable colours should change `accent-color` of inputs
Currently, checkboxes use the browser defaults for colours.

image

It would be nice if they used a customizable colour. Perhaps there could be the option of either using the browser default or a customizable colour (maybe just the chosen link colour)

image

This can be set using the `accent-color` on the input

Valer100:
Resolves #6341

### Changes
Added support for colorizing some elements like checkboxes, radio buttons, inputs, etc..

Website checkboxes (before):
image

Website checkboxes (after):
image

Slider variables (before):
image

Slider variables (after):
image

Editor dialogs (before):
image

Editor dialogs (after):
image

Block palette checkboxes (before):
image

Block palette checkboxes (after):
image

Block palette categories (before):
image

Block palette categories (after):
image

Active green flag background color (before):
image

Active green flag background color (after):
image

Device searching dialog (before):
image

Device searching dialog (after):
image

Project loading screen (before):
image

Project loading screen (after):
image

Tutorial popup (before):
image

Tutorial popup (after):
image

**Note:** Changing the accent color of UI elements that use blue as their accent color is optional and disabled by default. You'll need to enable the following option:
image

### Reason for changes
I made these changes, because the GUI looks inconsistent without them.

### Tests
Tested on Google Chrome 115.0.5790.110


WorldLanguages:
> Editor dialogs (before):
> image

The CSS change for this one looks mostly fine. Not sure about the other ones.

addons/editor-dark-mode/experimental_editor.css
```diff
@@ -652,3 +652,18 @@ img[src*="c3RvcC1wbGF5YmFjayI+CiAgICAgICAgICAgICAgICAgICAgICAgIDx1c2UgZmlsbD0iYm
...
+  accent-color: var(--editorDarkMode-primary);
+}
+
+/* Checkboxes, radio buttons */
+label {
+  accent-color: var(--editorDarkMode-primary);
+}
+
+/* Input fields from dialogs */
+input.prompt_variable-name-text-input_1iu8- {
```
WorldLanguages, line 667:
```suggestion
input[class*=prompt_variable-name-text-input_] {
```

Samq64:
Could this also be applied to block palette checkboxes?

I like these changes but I think they were intentionally not included (Except the textbox border of course) before because Scratch doesn't apply styles to them.

Valer100:
@Samq64 Now it works for checkboxes from the block palette.

addons/editor-dark-mode/experimental_editor.css
```diff
@@ -652,3 +652,42 @@ img[src*="c3RvcC1wbGF5YmFjayI+CiAgICAgICAgICAgICAgICAgICAgICAgIDx1c2UgZmlsbD0iYm
...
+  stroke: var(--editorDarkMode-primary);
+}
+
+g.checked > path.blocklyFlyoutCheckboxPath {
+  stroke: var(--editorDarkMode-primary-text);
+}
+
+/* Block palette categories */
+div.scratchCategoryMenuItem:hover {
+  color: color-mix(in srgb, var(--editorDarkMode-primary) 80%, #FFFFFF) !important;
```
WorldLanguages, line 683: I don't think we can use `color-mix` due to browser support

WorldLanguages:
I hope @mxmou can provide a deeper review, but here are a few notes.

- The added CSS probably affects more UI elements than you'd expect. We rarely target all `label`s or `input`s at once.
- Color customization for variable monitors probably belongs in a new addon that doesn't yet exist (#3912)
- The `Scratch default colors (purple)` preset, ideally, should look exactly like vanilla Scratch (maybe browser defaults can be the exception, but this PR also changes some colors which are set by Scratch instead)

addons/dark-www/addon.json
```diff
@@ -14,6 +14,12 @@
       "note": "Dark WWW",
       "id": "darkwww",
       "link": "https://github.com/gducrash"
+    },
+    {
+      "name": "Valer100",
+      "note": "Some fixes",
+      "id": "fixes",
+      "link": "https://github.com/Valer100"
```
mxmou, line 22: I wouldn't call these changes "fixes" because the previous behavior wasn't wrong.
```suggestion
    {
      "name": "Valer100",
      "link": "https://github.com/Valer100"
```

addons/dark-www/experimental_scratchwww.css
```diff
@@ -809,3 +809,12 @@ p {
...
 }
+
+/* Fix the accent color of checkboxes and radio buttons */
+label {
+  accent-color: var(--darkWww-button);
+}
+
+input {
+  accent-color: #129804;
+}
```
mxmou, line 820: There's no reason to apply `accent-color` to both `input` and `label` elements.
```suggestion
input {
  accent-color: var(--darkWww-button);
}
```

addons/editor-dark-mode/addon.json
```diff
@@ -25,6 +25,12 @@
       "note": "TurboWarp Dark",
       "id": "twdark",
       "link": "https://github.com/GarboMuffin"
+    },
+    {
+      "name": "Valer100",
+      "note": "Some fixes",
+      "id": "fixes",
+      "link": "https://github.com/Valer100"
```
mxmou, line 33:
```suggestion
    {
      "name": "Valer100",
      "link": "https://github.com/Valer100"
```

addons/editor-dark-mode/experimental_editor.css
```diff
@@ -652,3 +652,42 @@ img[src*="c3RvcC1wbGF5YmFjayI+CiAgICAgICAgICAgICAgICAgICAgICAgIDx1c2UgZmlsbD0iYm
...
+}
+
+div.scratchCategoryMenuItem.categorySelected {
+  background-color: var(--editorDarkMode-primary-transparent15);
+}
```
mxmou, line 688: I think it's more important to be consistent with Scratch here. These are blue in vanilla Scratch.

mxmou (changes requested):
Changing things that Scratch doesn't normally use the purple color for, such as sliders and checkboxes, should be optional and disabled by default.

addons/dark-www/experimental_scratchwww.css
WorldLanguages: Shouldn't this be `body:not(.sa-body-editor) input`, at least?

PR #6469, opened 2023-07-30
Fix long preset names not looking correct

### LINKED ISSUE #6468: Long preset names look broken in the popup
### Description

If an addon has a preset with a long name, the button isn't displayed correctly in the popup.

### Steps to reproduce

There isn't a long enough preset name in the current version of SA (at least not in English), but #6467 adds one (see screenshot).

image

### Environment

Browser: Edge 117 dev
Scratch Addons version: 2f20ba0
Operating system: Windows

mybearworld:
Resolves #6468

(Don't mind the branch name, I originally solved this using a different strategy which didn't actually work)

### Changes

Change the way padding works on a .large-button to use actual padding.

### Reason for changes

So that buttons with more than one line look correct

### Tests

Tested in Edge, it would surprise me if it didn't work everywhere

the button looking correct

### DIFF
--- a/webpages/styles/components/buttons.css
+++ b/webpages/styles/components/buttons.css
@@ -2,10 +2,9 @@
   color: var(--content-text);
   background: var(--button-background);
   border: 1px solid var(--control-border);
-  padding: 0 12px;
+  padding: 7px 12px;
   border-radius: 4px;
   transition: 0.2s ease;
-  height: 32px;
   box-sizing: border-box;
   font-family: inherit;
   font-size: 12px;
--- a/webpages/styles/components/buttons.css
+++ b/webpages/styles/components/buttons.css
@@ -2,6 +2,7 @@
   color: var(--content-text);
   background: var(--button-background);
   border: 1px solid var(--control-border);
+  display: flex;
   padding: 7px 12px;
   border-radius: 4px;
   transition: 0.2s ease;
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Make sure this doesn't change the style of the buttons in the normal settings view.

mybearworld:
> Make sure this doesn't change the style of the buttons in the normal settings view.

It does not. Top is this PR, bottom is how it is currently.

image
image

DNin01 (commented):
Much better, but there is still an issue visible in your screenshot - on the button with the wrapping text, the color palette and the next line of text are shifted slightly.

mybearworld:
> Much better, but there is still an issue visible in your screenshot - on the button with the wrapping text, the color palette and the next line of text are shifted slightly.

Fixed.

webpages/styles/components/buttons.css
```diff
@@ -2,10 +2,10 @@
   color: var(--content-text);
   background: var(--button-background);
   border: 1px solid var(--control-border);
-  padding: 0 12px;
+  display: flex;
```
mxmou, line 5: This makes the preview and text incorrectly aligned unless `vertical-align` is set to `center`.

mybearworld: True, I didn't notice.
```suggestion
  display: flex;
  align-items: center;
```

mxmou (approved):

PR #6471, opened 2023-07-31
Fix editor-theme3 browser support (array.prototype.at)

WorldLanguages:
### Changes

Avoid `array.at(-1)`, specially inside a Blockly trap.
`Array.prototype.at` is only available in Chromium 92+ (released June 2021)

### Reason for changes

This was the root issue from a feedback report (Chromium 87).
image

### DIFF
--- a/addons/editor-theme3/theme3.js
+++ b/addons/editor-theme3/theme3.js
@@ -271,7 +271,7 @@ export default async function ({ addon, console, msg }) {
       }
     } else {
       const iconsToReplace = ["repeat.svg", "rotate-left.svg", "rotate-right.svg"];
-      const iconName = src.split("/").at(-1);
+      const iconName = src.split("/")[src.split("/").length - 1];
       if (iconsToReplace.includes(iconName)) {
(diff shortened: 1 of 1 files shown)


RedGuy12:
https://github.com/ScratchAddons/ScratchAddons/blob/2f20ba0497fbaf416ea8d9136e6f78bb3a40d087/addons/search-profile/userscript.js#L54
https://github.com/ScratchAddons/ScratchAddons/blob/2f20ba0497fbaf416ea8d9136e6f78bb3a40d087/addons/middle-click-popup/WorkspaceQuerier.js#L931

WorldLanguages:
@RedGuy12 Thanks! Haven't tried, but I think those break the feature but the website/editor are still functional. We don't want to guarantee all features work in 3 year old browser, but at least the settings page should be functional and addons shouldn't break the editor.

apple502j (approved):
Guess I have to configure eslint for this

WorldLanguages:
cc @mxmou this changed your code

PR #6481, opened 2023-08-03
Add v1.34.0 update notification

WorldLanguages:
We might still merge more PRs, but I think we can all agree that these 2 new addons deserve to be mentioned

Screenshot:

image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -242,15 +242,15 @@
   "extensionHasUpdated": {
     "message": "Scratch Addons has updated to v$1"
   },
-  "extensionUpdateInfo1_v1_33_2": {
-    "message": "Scratch just released an update that changes some of the colors around the website. If you'd like to update your settings to use these new colors, go back, or otherwise customize the colors, you can reset or change any themes from $1.",
+  "extensionUpdateInfo1_v1_34": {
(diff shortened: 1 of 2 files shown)


DNin01 (approved):

_locales/en/messages.json
```diff
@@ -242,15 +242,15 @@
   "extensionHasUpdated": {
     "message": "Scratch Addons has updated to v$1"
   },
-  "extensionUpdateInfo1_v1_33_2": {
-    "message": "Scratch just released an update that changes some of the colors around the website. If you'd like to update your settings to use these new colors, go back, or otherwise customize the colors, you can reset or change any themes from $1.",
+  "extensionUpdateInfo1_v1_34": {
+    "message": "New addon: \"arrow key incrementation in editor\", which lets you use the up and down arrow keys to increase or decrease number values in block inputs. Go to $1 to enable it.",
```
apple502j, line 246: One change I always wanted to make, but didn't have time, was to split the "Go to..." message into another string, since it is used frequently. Maybe now's the time.

apple502j (changes requested):

_locales/en/messages.json
WorldLanguages: It's such a small string, I'm not sure if it's worth it

mybearworld: I think there would be too many ways that "Go to $1 to enable it" could be changed slightly ("Go to $1 to change this" or something), where it's not really worth it.

PR #6489, opened 2023-08-04
New addon: preview project notes on own projects as seen by other users

### LINKED ISSUE #6487: Preview project Notes & Credits and Instructions
### Why this would be helpful

Sometimes I want to see what the text I've written in Instructions/Notes & Credits will look like to other users.

### How this addon works

This would add an `.action-button` on project pages that will replace the `textarea` with a rendered version.

### Possible alternatives

Open a private window, but this doesn't work on unshared projects.

### Additional context

Screenshot 2023-08-04 010159

(The above image is a working version([Steve0Greatness/ScratchAddons:Preview-Description](https://github.com/Steve0Greatness/ScratchAddons/tree/Preview-Description)). I just thought I should make an issue before, submitting a PR because I think them's the rules.)

Steve0Greatness:
Resolves #6487

### Changes

Added a new addon: Preview Project Descriptions(`preview-project-description`).

The addon adds a new button to `.action-buttons` that toggles between showing a preview of Notes & Credits and Instructions, and showing the textbox.

### Reason for changes

There was no way to preview unshared project's descriptions before sharing it, in which you then have to either logout, or open a private window.

### Tests

Tested on...

- FireFox
- Chrome
- Microsoft Edge


addons/preview-project-description/userscript.js
```diff
@@ -0,0 +1,68 @@
...
+  togglePreview(null, currentlyEnabled);
+
+  /**
+   * Will attempt to toggle between the off and on states, however an override can be passed which will completely override the current value
+   * @param {string} _
+   * @param {boolean} override
+   * @returns {void}
+   */
+  function togglePreview(_ = null, override = !currentlyEnabled) {
+    enableSwitcher.innerHTML = `<span>${override ? msg("Disable") : msg("Enable")}</span>`;
```
WorldLanguages, line 40: https://scratchaddons.com/docs/develop/userscripts/best-practices/#avoid-using-innerhtml

addons/preview-project-description/userscript.js
```diff
@@ -0,0 +1,68 @@
+export default async function ({ addon, msg }) {
+  const loggedInUser = await addon.auth.fetchUsername();
+  const projectOwner = (await addon.tab.waitForElement(".project-header .avatar")).alt;
```
WorldLanguages, line 3: I think it's better to wait until the notes exist above this line, and use this instead

```suggestion
  const projectOwner = addon.tab.redux.state.preview.projectInfo.author.username;
```

addons/preview-project-description/userscript.js
```diff
@@ -0,0 +1,68 @@
...
+   * @returns {void}
+   */
+  function setDisplayable(element, shown = true) {
+    if (shown) {
+      element.style.removeProperty("display");
+    } else {
+      element.style.display = "none";
+    }
+  }
+}
```
WorldLanguages, line 68: It might be better to create an userstyle CSS file and toggle a class from jS instead:

```css
.sa-preview-description-hidden {display:none;}
```

WorldLanguages:
> Allows you to preview the what will show up in the final notes and credits. It's not 100% accurate.

Why is it not 100% accurate? There's probably a way to make it actually accurate.

addons/preview-project-description/userscript.js
Steve0Greatness: I addressed it in a slightly different way

Steve0Greatness:
> Why is it not 100% accurate? There's probably a way to make it actually accurate.

I still need to add usernames and make it interact with the more links addon.

addons/preview-project-description/userscript.js
WorldLanguages: Hmm, I see. You can still use `element.classList.add` and `element.classList.remove` though.

WorldLanguages:
For the usernames, we can probably try to use Scratch's linkifier somehow. Unsure if it's really feasible.

WorldLanguages:
It might be possible to simply use Scratch's preview by somehow setting `isEditable` to false and forcing a React rerender: https://github.com/scratchfoundation/scratch-www/blob/59fa9cd12744ec6fcfc73d0d92c4fbbe5ca73d38/src/views/preview/project-view.jsx#L1032

WorldLanguages:
> It might be possible to simply use Scratch's preview by somehow setting `isEditable` to false

It's quite hard 😅

Secret-chest:
more-links compat too

Steve0Greatness:
Got usernames working! All it took was some dodgy regex.

Also, I did use `innerHTML`(I am aware that it's bad practice), however, I really don't want to try and figure out how to not use it when trying to work with that before was just not working, and there would have been basically no way to actually make both it and urls work.

I'm gonna add `more-links` compatability now 

Screenshot 2023-08-06 145801

Steve0Greatness:
I've just noticed an issue that I think is with the way addons are loaded: if you load the project editor from mystuff, the addon isn't able to load because none of the required DOM elements exist, and due to single page stuff the addon still isn't loaded when you click `See Project Page`. The only way I can see resolving this is adding a wait for element to every required element call.

Steve0Greatness:
@Secret-chest done.

WorldLanguages:
@Secret-chest See, for example, the project-notes-tabs addon source code to see how to run code each time the user gets into the editor.

WorldLanguages:
@Steve0Greatness See, for example, the project-notes-tabs addon source code to see how to run code each time the user gets into the editor.

addons/preview-project-description/userscript.js
```diff
@@ -0,0 +1,92 @@
...
+    let input = editor.value.replace(/\</g, "&lt;").replace(/\>/g, "&gt;").replace(/\&/g, "&amp;");
+    let rendered = input.replace(matchUsername, (matched) => {
+      return `<a href="/users/${matched.slice(1)}">${matched}</a>`;
+    });
+    if (moreLinksEnabled)
+      rendered.replace(matchHyperLinks, (matched) => {
+        if (matched.startsWith("https://") || matched.startsWith("http://")) matched = "//" + matched;
+        return `<a href="${matched}">${matched}</a>`;
+      });
+    preview.innerHTML = rendered;
```
WorldLanguages, line 76: This is probably vulnerable to XSS

Steve0Greatness: Maybe, but at the top of the function I'm matching for `<`, `>`, and `&`. I could probably also add `"` can't escape out of `href`, but otherwise I don't see how else it could be XSS unless you're talking about the site it brings you to.

WorldLanguages:
> It might be possible to simply use Scratch's preview by somehow setting `isEditable` to false and forcing a React rerender

I'm trying to get this to work with sketchy JavaScript hacks

WorldLanguages:
I think I got it to work. So no need to manually linkify URLs and `@mentions`.

@Steve0Greatness I'll push to your branch tomorrow

WorldLanguages:
Okay, that's it. No need to create the preview manually now, so more-links compat comes for free.

WorldLanguages:
Also, would be interesting if the way to toggle the preview was a bit more intuitive. It could be a toggle next to the tabs, if project-notes-tabs is enabled.

WorldLanguages:
> Okay, that's it. No need to create the preview manually now, so more-links compat comes for free.

Well, actually it looks like this is faking the username field more than it's supposed to, but not in Firefox (this time it's actually easier to implement in Firefox). This could lead to some bugs that may be very hard to find out about, so it's better to provide a fragile preview feature that's as restricted as possible.

Steve0Greatness:
No Preview Button

The ability to preview no longer works on unshared projects.

WorldLanguages:
Another funny bug:

brave_B06C4gp2fp

WorldLanguages:
> The ability to preview no longer works on unshared projects.

Luckily, that's only because of the position of the button. If we convert this to be a toggle somewhere else, it should work just fine in unshared projects.

Steve0Greatness:
> Luckily, that's only because of the position of the button. If we convert this to be a toggle somewhere else, it should work just fine in unshared projects.

How about when `project-notes-tabs` is on:

Screenshot 2023-08-08 124416

Although it may not be a great spot.

WorldLanguages:
This userscript is pretty crazy.

It involves:
- Prototype traps
- Javascript Proxy object (well, a getter function would have worked too)
- Differences between Chrome and Firefox
- Infinite loops
- Forcing React rerenders

I'm not sure if any other addon ever needed a combo of all of these lol.

WorldLanguages:
Well, I'm done for today (I'll dream about react rerenders tonight...)

Please tell me if you find any bugs while testing this in your shared projects. If not, we can work on moving the toggle to somewhere else and making this compatible with project-notes-tabs.

mybearworld:
It would be nice if the preview state was preserved when seeing inside.

WorldLanguages:
> Also, occasionally, this happens

@mybearworld It's quite hard trying to reproduce these bugs, unless it happened to you more than once in the same exact moment. Any other info?
Also, while searching for bugs we'll often try toggling the preview as soon as we enter the project, but I think most users will toggle it after at least 10 seconds after loading the project (unless you're going to the project just to see the preview). I guess this solution is a compromise between reliability and ensuring a one-to-one match between the preview the user sees and the one they would see if the project wasn't theirs.

mybearworld:
Would it be worth marking this as Beta? There's probably going to be some issues with it, from the very nature of the way this addon works

addons/preview-project-description/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Preview Project Descriptions",
+  "description": "Allows you to preview the what will show up in the final notes and credits.",
+  "credits": [
+    {
+      "name": "Steve0Greatness",
+      "link": "https://scratch.mit.edu/users/Steve0Greatness"
+    }
+  ],
+  "tags": ["community", "projectPage"],
```
WorldLanguages, line 10:
```suggestion
  "tags": ["community", "projectPage", "beta"],
```

addons/preview-project-description/addon.json
```diff
@@ -0,0 +1,21 @@
...
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"],
+      "runAtComplete": true
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.34.0"
```
WorldLanguages, line 20:
```suggestion
  "versionAdded": "1.35.0"
```

addons/preview-project-description/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Preview Project Descriptions",
+  "description": "Allows you to preview the what will show up in the final notes and credits.",
```
WorldLanguages, line 3: Maybe something like...

```suggestion
  "name": "Preview project instructions and notes",
  "description": "In your own projects, adds a toggle [somewhere TODO] that lets you preview the project instructions and notes as seen by other users.",
```

addons/preview-project-description/userscript.js
```diff
@@ -0,0 +1,171 @@
...
+  });
+
+  while (true) {
+    // Same waitForElement call as animated-thumb/userscript.js
+    await addon.tab.waitForElement(".flex-row.subactions > .flex-row.action-buttons", {
+      markAsSeen: true,
+      reduxCondition: (state) => state.scratchGui.mode.isPlayerOnly,
+    });
+
+    // TODO: also change animated-thumb/userscript.js
```
WorldLanguages, line 37: Reminder: TODO

WorldLanguages:
Looks like this could be merged, but these are pending I think:

- [ ] Making the UI better, probably better as a toggle than a button (any ideas welcome @BroJac5246 @DNin01 @Samq64 @mxmou on where to position the toggle/button)
- [ ] `project-notes-tabs` compatibility - and possibly adding a "preview" tab instead of a button/toggle
- [ ] The TODO in userscript.js
- [ ] I'll check the whole userscript.js to make sure this can't cause infinite loops or editor crashes - we prefer a more fragile feature that doesn't crash the page

Samq64:
Maybe on the right next to the instructions label?

> `project-notes-tabs` compatibility - and possibly adding a "preview" tab instead of a button/toggle

I like that idea.

BroJac5246:
> - [ ] Making the UI better, probably better as a toggle than a button (any ideas welcome @ BroJac5246 @ DNin01 @ Samq64 @ mxmou on where to position the toggle/button)

I think a toggle would look best, probably to the right of the Instructions tab (or the Notes and Credits tab, if `project-notes-tabs` is enabled).

> - [ ] `project-notes-tabs` compatibility - and possibly adding a "preview" tab instead of a button/toggle

Compatibility with this addon is important, IMO.

I think a tab that shows what it will look like for users without `project-notes-tab` _could_ be useful. But a toggle is more intuitive and should probably be used instead of a tab.

WorldLanguages:
image

image

image

WorldLanguages:
If `project-notes-tabs` is enabled, should the preview use the tabs or should it look like vanilla Scratch?
I was thinking vanilla Scratch, because you want to see how the project looks to others.

addons/preview-project-description/userscript.js
WorldLanguages: Maybe I should create a `waitForSelfProject()` utility that makes sure to consider going to your own project logged out and logging-in within the same session (Scratch does not reload in that case)

rgantzos:
> (We also need to be careful with adding the toggle next to the tabs - in some languages, there might not be enough space for all these words: `Instructions - Notes and Credits - Preview`)

Maybe an icon placed next to the tabs?

mybearworld:
German is particularly annoying here

The tabs taking up almost full width

WorldLanguages:
> German is particularly annoying here

Yeah, no idea where we could position the toggle in German...

I can only think of sending the word "Notes" for translation, and changing `Notes and Credits` for just `Notes` in all languages, if both addons are enabled at the same time.

mybearworld:
> I can only think of sending the word "Notes" for translation, and changing `Notes and Credits` for just `Notes` in all languages, if both addons are enabled at the same time.

This feels more like a hack than an actual fix.

WorldLanguages:
> This feels more like a hack than an actual fix.

I'm all ears, what else could we do? Decrease the font size of the tabs?

mybearworld:
> > This feels more like a hack than an actual fix.
> 
> I'm all ears, what else could we do? Decrease the font size of the tabs?

Not putting it next to the tabs would be an option. There'd be space next to the remix button area, for example.

WorldLanguages:
> There'd be space next to the remix button area, for example.

Yeah, but that's less intuitive.

Steve0Greatness:
How about next to the project title?

DNin01:
Or above the field(s) if there's not enough space, like where the remix attribution tags are?

WorldLanguages:
image

or:

image

WorldLanguages:
> Or above the field(s) if there's not enough space, like where the remix attribution tags are?

Unsure if there's an accurate way to know if there's "enough space" with JavaScript 🤔

BroJac5246:
> image
> 
> or:
> 
> image
> 

Nice! I like the second one more because IMO the first one draws too much attention to an element that you won't need most of the time.

Secret-chest:
The switch feels bulky and the padding should be reduced. Also, not all people use project notes tabs.

mxmou:
> image
> 
> or:
> 
> image

There's a lot of empty space in both of these versions. I think it's better to only move the switch above the tabs if there's not enough space next to the tabs.

image
image

WorldLanguages:
> There's a lot of empty space in both of these versions. I think it's better to only move the switch above the tabs if there's not enough space next to the tabs.

Agreed, I might need a bit of help with the CSS though.

Also, is there a reliable way to know if "there's enough space next to the tabs", with JavaScript I assume? I'm not aware of any. Has any addon had to tackle this problem before?

WorldLanguages:
Well, we're waiting on a toggle, so I'll try to get it somewhat done in the following days.

WorldLanguages:
> Also, is there a reliable way to know if "there's enough space next to the tabs", with JavaScript I assume?

I'm going to try:
1. Add the element next to the tab anyway
2. Comparing `.sa-tabs`'s clientWidth and scrollWidth to see if they match.
3. If they don't match, remove the element and re-add it above the tabs.

mxmou:
I think this can be done with CSS. You need a div containing both `.sa-tabs` and the preview switch with the following styles:
```css
margin: 0;
width: 100%;
display: flex;
flex-wrap: reverse;
```
The preview switch just needs `margin-inline-start: auto` to align it with the right edge of the description.

WorldLanguages:
@mxmou Thanks! I almost got it to work, I'm having trouble getting this new div you mention auto-adapt to the height depending on whether the element wrapped or not.

WorldLanguages:
German:
image
image

English:
image

WorldLanguages:
### Project tabs off:
brave_j7K5QtU8TH

### Project tabs on:
brave_3pUitfoQL9

addons/preview-project-description/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Preview project instructions and notes",
+  "description": "In your own projects, adds a toggle [somewhere TODO] that lets you preview the project instructions and notes as seen by other users.",
```
WorldLanguages, line 3:
```suggestion
  "description": "In your own projects, adds a toggle above the Instructions field to preview the project instructions and notes as seen by other users.",
```

Steve0Greatness:
I think I accidentally closed this 😅

addons/preview-project-description/addon.json
WorldLanguages: feedback?

WorldLanguages (approved):
I don't expect anyone else to approve my ugly code, but maybe you can test the functionality 1) with tabs on or off 2) with empty instructions, empty notes, both empty, both filled

This might not work if you first log out, then go to one of your projects, and log-in within the project page. The project note tabs addon also has similar bugs in those cases.

addons/preview-project-description/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Preview project instructions and notes",
+  "description": "In your own projects, adds a toggle above the Instructions field to preview the project instructions and notes as seen by other users.",
+  "credits": [
+    {
+      "name": "Steve0Greatness",
+      "link": "https://scratch.mit.edu/users/Steve0Greatness"
+    }
+  ],
+  "tags": ["community", "projectPage", "beta"],
```
WorldLanguages, line 10:
```suggestion
  "tags": ["community", "projectPage", "beta", "featured"],
```

DNin01 (commented):
Looks like the switch slightly pushes the Instructions field down.
Preview toggle visible VS invisible

PR #6495, opened 2023-08-05
Rewrite `my-ocular` on profiles and add new shared space

### LINKED ISSUE #6492: Fix `my-ocular` on profile pages
Currently, the `my-ocular` addon groups all of the elements within the country element on profile pages together as the innerText. However, this is not compatible with other extensions (including ScratchTools) or with any future addons that use the same space.

It should be added to a new shared space to fix this.

This would help with issues such as #4706

rgantzos:
Resolves #6492 

### Changes

This adds the new shared space `afterProfileCountry`, and appends the ocular status element at place 4 (to leave room in front of it, as it takes up a lot of space).

### Reason for changes

This improves compatibility with other extensions, as well as other addons.

### Tests

Tested on Chrome 115.0.5790.114 on MacOS Sonoma


addons/my-ocular/profile.js
```diff
@@ -1,46 +1,31 @@
...
   dot.style.backgroundColor = data.color;
+  span.appendChild(dot);
 
-  // Create location element
-  let locationElem = document.createElement("span");
-  locationElem.id = "my-ocular-old-location";
-  locationElem.innerText = container.innerText; // Set it to the old innerText
-  locationElem.fontStyle = "italic";
-  container.innerText = ""; // Clear the old location
+  if (document.querySelector(".sa-ocular-status")) return;
```
WorldLanguages, line 19: Not sure how this could be realistically possible.

```suggestion
  if (document.querySelector(".sa-ocular-status")) return;
```

addons/my-ocular/profile.js
```diff
@@ -1,46 +1,31 @@
...
   if (typeof data.userStatus !== "string") return;
 
-  // Create status element
-  let statusSpan = document.createElement("span");
-  statusSpan.title = msg("status-hover");
-  statusSpan.innerText = data.userStatus;
-  statusSpan.style.fontStyle = "italic";
-  statusSpan.style.setProperty("display", "inline-block", "important");
-  statusSpan.id = "my-ocular-span";
+  var span = document.createElement("span");
```
WorldLanguages, line 7:
```suggestion
  const span = document.createElement("span");
```

addons/my-ocular/profile.js
```diff
@@ -1,46 +1,31 @@
...
+  span.textContent = data.userStatus;
+  span.className = "sa-ocular-status";
+
+  span.style.fontStyle = "italic";
+  span.style.display = "inline-block !important";
 
-  // Create my-ocular-dot
-  let dot = document.createElement("span");
+  var dot = document.createElement("span");
```
WorldLanguages, line 14:
```suggestion
  const dot = document.createElement("span");
```

addon-api/content-script/Tab.js
```diff
@@ -613,6 +613,11 @@ export default class Tab extends Listenable {
         from: () => [],
         until: () => [],
       },
+      afterProfileCountry: {
+        element: () => q("p.profile-details > .location"),
+        from: () => [],
+        until: () => [],
+      },
```
WorldLanguages, line 620: I'm not sure how exactly this would work, as the `<span class="location">` already has a text node inside it (the user's country) 🤔
Maybe this space should create its own div and the space should be inside that new element?

WorldLanguages: Self reminder: some other addons might be caused by changing the location CSS, such as `better-featured-project`, `scratchr2`

addons/my-ocular/profile.js
CST1229: Probably only with multiple instances of SA running at the same time.
edit: how do i cancel a review

addon-api/content-script/Tab.js
rgantzos: `scratchr2` only changes minor styling such as color and shadow, but it does change the line height from `1.5385em` to `16px`, which doesn't make a difference.

`better-featured-project` adds `display: block !important;` to the element, which moves the new shared space element below the location. That's something that needs to be fixed.

I looked through the other addons. No other ones seem to affect the element, with the exception of the dark mode addon, but that just changes the colors.

addons/my-ocular/profile.js
```diff
@@ -1,46 +1,30 @@
...
-  // Create location element
-  let locationElem = document.createElement("span");
-  locationElem.id = "my-ocular-old-location";
-  locationElem.innerText = container.innerText; // Set it to the old innerText
-  locationElem.fontStyle = "italic";
-  container.innerText = ""; // Clear the old location
+  addon.tab.appendToSharedSpace({
+    space: "afterProfileCountry",
+    element: span,
+    order: 4, // set back due to width of element
```
WorldLanguages, line 22: I'm going to change it to 9, in case there's more elements in the future.

```suggestion
    order: 9, // set back due to width of element
```

addons/my-ocular/profile.js
```diff
@@ -1,46 +1,30 @@
...
+  span.textContent = data.userStatus;
+  span.className = "sa-ocular-status";
+
+  span.style.fontStyle = "italic";
+  span.style.display = "inline-block !important";
 
-  // Create my-ocular-dot
-  let dot = document.createElement("span");
+  const dot = document.createElement("span");
   dot.title = msg("status-hover");
```
WorldLanguages, line 15: The title attribute is in the wrong element now.
Hovering the whole status should let you see it, not just hovering the dot.

addons/my-ocular/status-dot.css
```diff
@@ -7,7 +7,7 @@
   display: inline-block;
 }
 
-#my-ocular-span {
+.sa-ocular-status {
   border-left: 1px solid rgb(204, 204, 204);
```
WorldLanguages, line 11: Looks like 2+ line statuses need `display: inline-block;` added here to behave the same way as before

There are more weird things going on though (this change won't be enough), go to CST's profile (https://scratch.mit.edu/users/CST1229/) with "profile page banner" enabled and see the difference between master branch and this branch.
The "profile page banner" (better-featured-project) also has some styles with !important on them which make debugging this more annoying

addons/my-ocular/profile.js
```diff
@@ -1,46 +1,30 @@
...
+  const span = document.createElement("span");
+  span.textContent = data.userStatus;
+  span.className = "sa-ocular-status";
+  span.title = msg("status-hover");
 
-  // Create my-ocular-dot
-  let dot = document.createElement("span");
-  dot.title = msg("status-hover");
-  dot.classList.add("my-ocular-dot");
-  dot.style.setProperty("display", "inline-block", "important"); // I have to do it like this because .style doesn't let me set prio, and featured project banner messes with this without !important
```
WorldLanguages, line 24: Why was this removed? I'll readd it through CSS now.

WorldLanguages (approved):

PR #6504, opened 2023-08-08
Better forum-post-countdown name and description

Samq64:
### Changes

Shortens the  `forum-post-countdown` name and adds a bit more to the description.

### Reason for changes

Because shorter addon names are better in most cases and the location feels like it should be part of the description.

### Tests

Tested on Chromium 115


DNin01 (approved):

WorldLanguages (changes requested):
This is going to be merged just before release (and translated versions of the name and description will continue to be old)

WorldLanguages (approved):

PR #6513, opened 2023-08-12
Create NFE Checker addon

### LINKED ISSUE #6484: Show if projects are NFE
### Why this would be helpful

I want to see whether projects are FE, NFE or not rated.

### How this addon works

It just adds a text next to the share date showing the status of the project.

### Possible alternatives

_No response_

### Additional context

ScratchTools has it.

mybearworld:
Resolves #6484

### Changes

Adds a "Project moderation status display" addon that adds some text next to the share date showing the status of a project.

### Reason for changes

So people can know whether their projects are NFE or not.

### Tests

Tested on Edge.

### DIFF
--- a/addons-l10n/en/nfe-checker.json
+++ b/addons-l10n/en/nfe-checker.json
@@ -4,7 +4,7 @@
   "nfe-checker/fe": "Unrestricted",
   "nfe-checker/title-fe": "The Scratch Team has reviewed this project and deemed it safe for everyone.",
   "nfe-checker/nfe": "Restricted",
-  "nfe-checker/title-nfe": "This project reviewed by a Scratch Team member and while considered safe for Scratch, won't be available by searching for it and can't appear on the search bar. This rarely happens and it happens with reason. Do not create forum topics or otherwise try to get the Scratch Team to reconsider. If you want to ask why this decision happened, use the Contact Us link in the footer.",
+  "nfe-checker/title-nfe": "This project has been reviewed by a Scratch Team member and was considered okay for Scratch but won't be listed in search results or the Explore page. This rarely happens and it happens with reason. Do not publicly appeal (e.g. create forum topics) or otherwise try to get the Scratch Team to reconsider. If you want to ask why this decision happened, use the Contact Us link in the footer.",
   "nfe-checker/unavailable": "Unavailable",
   "nfe-checker/title-unavailable": "This project is not part of a remix tree (it has no remix and is not a remix)."
 }
--- a/addons-l10n/en/nfe-checker.json
+++ b/addons-l10n/en/nfe-checker.json
@@ -1,0 +1,10 @@
+{
+  "nfe-checker/unreviewed": "Unreviewed",
+  "nfe-checker/title-unreviewed": "The Scratch Team has not looked at this project yet.",
+  "nfe-checker/fe": "Unrestricted",
+  "nfe-checker/title-fe": "The Scratch Team has reviewed this project and deemed it safe for everyone.",
+  "nfe-checker/nfe": "Restricted",
+  "nfe-checker/title-nfe": "This project reviewed by a Scratch Team member and while considered safe for Scratch, won't be available by searching for it and can't appear on the search bar. This rarely happens and it happens with reason. Do not create forum topics or otherwise try to get the Scratch Team to reconsider. If you want to ask why this decision happened, use the Contact Us link in the footer.",
+  "nfe-checker/unavailable": "Unavailable",
+  "nfe-checker/title-unavailable": "This project is not part of a remix tree (it has no remix and is not a remix)."
+}
--- a/addons-l10n/en/nfe-checker.json
+++ b/addons-l10n/en/nfe-checker.json
@@ -1,10 +1,8 @@
 {
-  "nfe-checker/unreviewed": "Unreviewed",
-  "nfe-checker/title-unreviewed": "The Scratch Team has not looked at this project yet.",
-  "nfe-checker/fe": "Unrestricted",
-  "nfe-checker/title-fe": "The Scratch Team has reviewed this project and deemed it safe for everyone.",
-  "nfe-checker/nfe": "Restricted",
-  "nfe-checker/title-nfe": "This project has been reviewed by a Scratch Team member and was considered okay for Scratch but won't be listed in search results or the Explore page. This rarely happens and it happens with reason. Do not publicly appeal (e.g. create forum topics) or otherwise try to get the Scratch Team to reconsider. If you want to ask why this decision happened, use the Contact Us link in the footer.",
-  "nfe-checker/unavailable": "Unavailable",
-  "nfe-checker/title-unavailable": "This project is not part of a remix tree (it has no remix and is not a remix)."
+  "nfe-checker/notreviewed": "Unreviewed",
+  "nfe-checker/title-notreviewed": "The Scratch Team has not looked at this project yet.",
+  "nfe-checker/safe": "Unrestricted",
+  "nfe-checker/title-safe": "The Scratch Team has reviewed this project and deemed it safe for everyone.",
+  "nfe-checker/notsafe": "Restricted",
+  "nfe-checker/title-notsafe": "This project has been reviewed by a Scratch Team member and was considered okay for Scratch but won't be listed in search results or the Explore page. This rarely happens and it happens with reason. Do not publicly appeal (e.g. create forum topics) or otherwise try to get the Scratch Team to reconsider. If you want to ask why this decision happened, use the Contact Us link in the footer."
 }
--- a/addons/nfe-checker/addon.json
+++ b/addons/nfe-checker/addon.json
@@ -1,13 +1,6 @@
 {
   "name": "Project moderation status display",
   "description": "Show the moderation status of a project.",
-  "info": [
-    {
-      "type": "notice",
-      "text": "The way we get this information requires the project to be part of a remix tree (either have a remix or be a remix).",
-      "id": "req-remix-tree"
-    }
-  ],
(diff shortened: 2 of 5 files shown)


WorldLanguages:
Which of the alternatives listed here #6484 does this implement?

addons/nfe-checker/addon.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Project moderation status display",
+  "description": "Show the moderation status of a project. This requires a project to be part of a remix tree (e.g. either have a remix or be a remix).",
```
DNin01, line 3:
```suggestion
  "description": "Show the moderation status of a project.",
  "info": [
    {
      "type": "notice",
      "text": "The way we get this information requires the project to be part of a remix tree (either have a remix or be a remix).",
      "id": "req-remix-tree"
    }
  ]
```

DNin01 (commented):
I would strongly suggest making this only display the status of the user's own projects. While it's technically possible to use this trick to get the moderation status of any project with a remix tree, I don't think it would be very useful to know this information for other projects.

addons-l10n/en/nfe-checker.json
```diff
@@ -0,0 +1,10 @@
+{
+  "nfe-checker/unreviewed": "Unreviewed",
+  "nfe-checker/title-unreviewed": "The Scratch Team has not looked at this project yet.",
+  "nfe-checker/fe": "Unrestricted",
+  "nfe-checker/title-fe": "The Scratch Team has reviewed this project and deemed it safe for everyone.",
+  "nfe-checker/nfe": "Restricted",
+  "nfe-checker/title-nfe": "This project reviewed by a Scratch Team member and while considered safe for Scratch, won't be available by searching for it and can't appear on the search bar. This rarely happens and it happens with reason. Do not create forum topics or otherwise try to get the Scratch Team to reconsider. If you want to ask why this decision happened, use the Contact Us link in the footer.",
```
DNin01, line 7:
```suggestion
  "nfe-checker/title-nfe": "This project has been reviewed by a Scratch Team member and was considered okay for Scratch but won't be listed in search results or the Explore page. This rarely happens and it happens with reason. Do not publicly appeal (e.g. create forum topics) or otherwise try to get the Scratch Team to reconsider. If you want to ask why this decision happened, use the Contact Us link in the footer.",
```

addons-l10n/en/nfe-checker.json
```diff
@@ -0,0 +1,10 @@
...
+  "nfe-checker/unavailable": "Unavailable",
+  "nfe-checker/title-unavailable": "This project is not part of a remix tree (it has no remix and is not a remix)."
```
DNin01, line 9: It doesn't really make sense to label the text element "Unavailable" out of context (_the project is unavailable? Huh?_)

There's gotta be a better way to make it obvious that the status is unavailable...

mxmou:
A way to get the status of a project without remixes was mentioned in the issue: https://github.com/ScratchAddons/ScratchAddons/issues/6484#issuecomment-1664400904

addons-l10n/en/nfe-checker.json
mybearworld: There's no longer an Unavailable option, so this is resolved

DNin01:
> A way to get the status of a project without remixes was mentioned in the issue: [#6484 (comment)](https://github.com/ScratchAddons/ScratchAddons/issues/6484#issuecomment-1664400904)

Oh, nice!

DNin01 (commented):
That's quite an improvement already! I would still consider my suggestion about only displaying the status for the user's own projects.

mybearworld:
> That's quite an improvement already! I would still consider my suggestion about only displaying the status for the user's own projects.

It's an artificial limitation, and I really don't like the idea of that.

DNin01:
> > That's quite an improvement already! I would still consider my suggestion about only displaying the status for the user's own projects.
> 
> It's an artificial limitation, and I really don't like the idea of that.

Yeah, I don't really like having to use artificial limitations either, but at the same time, what could you possibly get from seeing this status on someone else's project? It's kind of like Scratch Addons naming names in front of you when they make NFE content. I wouldn't feel good if that happened on me.

Hmm... one way around this problem might be a setting that is enabled by default called "only show for my own projects"...? What does everyone think of this solution and this problem in general?

DNin01:
> If the goal is avoiding people making fun of others because they make NFE content, this wouldn't help.

Yeah, that's another complication that comes up when creating a feature like this. I think this will require a bit more discussion. Does anyone else like @Secret-chest have thoughts on what we should do about displaying the status on projects that aren't yours? To recap, I don't think anyone would really find it "useful" to see whether projects from the community are FE or NFE but the trick we're using still works on them, and not allowing people to see it would be an artificial limitation.

BroJac5246:
> Hmm... one way around this problem might be a setting that is enabled by default called "only show for my own projects"...? What does everyone think of this solution and this problem in general?

I was thinking about this myself and I think it's a good idea.

> If the goal is avoiding people making fun of others because they make NFE content, this wouldn't help.

Will people actually "make fun of" those who make NFE content? I wouldn't think so, and it's - a violation of the community guidelines.

> Yeah, that's another complication that comes up when creating a feature like this. I think this will require a bit more discussion. Does anyone else like @Secret-chest have thoughts on what we should do about displaying the status on projects that aren't yours? To recap, I don't think anyone would really find it "useful" to see whether projects from the community are FE or NFE but the trick we're using still works on them, and not allowing people to see it would be an artificial limitation.

Really, my thoughts are that issues are unlikely to arise from this addon because... why would they?

This could definitely be helpful if you do come across a NFE project but wish to avoid such content. If you have this addon then you are forewarned.

Secret-chest:
> This could definitely be helpful if you do come across a NFE project but wish to avoid such content. If you have this addon then you are forewarned.

Good point.

SA is open source so someone could just remove the check in a few minutes, so I don't why we're adding such artificial limitations.

WorldLanguages:
> Will people actually "make fun of" those who make NFE content?

Probably not, but they could comment "your project is NFE" and it could create some drama even though the project author doesn't know about our addon or maybe doesn't even know what NFE is.

WorldLanguages:
> This could definitely be helpful if you do come across a NFE project but wish to avoid such content. If you have this addon then you are forewarned.

The problem is that these types of projects you are supposed to avoid are often taken down, simply. The Scratch Team usually uses NFE to silently make projects less discoverable. It's intentionally silent.

DNin01:
> > This could definitely be helpful if you do come across a NFE project but wish to avoid such content. If you have this addon then you are forewarned.
> 
> Good point.
> 
> SA is open source so someone could just remove the check in a few minutes, so I don't why we're adding such artificial limitations.

Not everyone is a developer.

WorldLanguages:
We would need to ask the Scratch Team moderation team to know how, ideally, everything should work according to them. Maybe, they would want to add a visible warning system, and maybe they can't because the developer team has other priorities. That might also explain why the remix tree still contains this information.

But we shouldn't try to reinterpret a moderator's intentions. Whether a project is NFE is only shown in the UI for admins, and they NFE'd projects that explain how to know whether a project is NFE. Then, my point is, it's not realistic to argue that it would be helpful to know if someone else's project is NFE or not. Maybe you could use another adjective, such as "informative", which really becomes important when it's your own project and wonder why it doesn't show up in search.

DNin01:
> it can be valuable information.

How?

> If this would be a problem in the ST's eyes, they'd remove the `moderation_status` key from appearing in their remix tree system.

I don't think they meant to disclose the status to us in doing so. It was probably just there so the page could hide it from the remix tree.

mxmou:
A former ST member said in [this forum post](https://scratch.mit.edu/discuss/post/4728749/):

> The FE/NFE tags were intended to be used silently without public knowledge, but then some Scratchers came by and made our lives a little more difficult for us.

WorldLanguages:
To be fair, Paddle2See publicly shared a lot of private knowledge regarding project moderation status which went to the Scratch Wiki around one decade ago. But regardless of the "responsibility", the intention is clear.

BroJac5246:
Should we reach out to the ST and see what their intentions are? If they ask us not to do this then we should respect that, but it's also possible their stance has changed— after all, they _could_ make it impossible to see if a project is NFE.

Samq64:
I say we don't ask the ST and just add it for the logged in user only.

WorldLanguages:
> Should we reach out to the ST and see what their intentions are?

I think their intentions are quite clear already, the NFE tags are intended to be silent. But I think it would be okay to add this addon as long as we don't heavily advertise it (we don't want to teach our users what NFE is, after all).

Chiroyce1:
> A former ST member said in [this forum post](https://scratch.mit.edu/discuss/post/4728749/):
> 
> > The FE/NFE tags were intended to be used silently without public knowledge, but then some Scratchers came by and made our lives a little more difficult for us.

That post was made in December 2020, if the status being public really is a *big* an issue for them, I don't think it should take more than a couple of years to make this info moderator only.  As W_L said,  as long as SA doesn't advertise this addon too much, it should be fine. The people who already know what it is can go ahead and enable it, and those who don't know about it will probably enable it and see that it is of no use to them.

WorldLanguages:
> That post was made in December 2020, if the status being public really is a _big_ an issue for them, I don't think it should take more than a couple of years to make this info moderator only.

Well, scratch-www is 8 years old already, and profiles are still server-side rendered with Django. Progress happens slowly. This staying unfixed for years doesn't mean much.

> As W_L said, as long as SA doesn't advertise this addon too much, it should be fine. The people who already know what it is can go ahead and enable it, and those who don't know about it will probably enable it and see that it is of no use to them.

Yeah, I agree.

PR #6535, opened 2023-08-17
Style reply button with `colorblind`

### LINKED ISSUE #6533: The reply button isn't styled by `colorblind`
### Description

The "reply" button in comments is not underlined even if the underline option in `colorblind` is selected.

### Steps to reproduce

1. Enable the underline option in `colorblind`
2. Go to a page with comments

The reply button having no underlines despite other links having one

### Environment

Browser: Edge 115
Scratch Addons version: 1.34.0
Operating system: Windows

mybearworld:
Resolves #6533

### Changes

Increases the specificity for scratchr2 reply buttons affected by the `scratchr2` addon and adds a new selector for scratch-www reply buttons.

A profile comment with an underlined reply button.

A studio comment with an underlined reply button.

### Reason for changes

To make `colorblind` more consistent.

### Tests

Tested on Edge.


Samq64 (approved):

addons/colorblind/bold-links.css
```diff
@@ -5,13 +5,15 @@ a:hover,
 a:active,
 a.black:hover,
 a.black:active,
+a.reply span,
```
DNin01, line 8: We can probably apply it to `a` instead of `span`, that's what Scratch 2.0 → 3.0 does while you're hovering over the reply button.
```suggestion
a.reply,
```

addons/colorblind/underline-links.css
```diff
@@ -5,11 +5,13 @@ a:hover,
 a:active,
 a.black:hover,
 a.black:active,
+a.reply span,
```
DNin01, line 8: (Same thing here)
```suggestion
a.reply,
```

addons/colorblind/bold-links.css
mybearworld: It looks better to have it on the span only, but sure, for consistency that makes sense.

DNin01 (approved):

Issue #3642, opened 2021-10-11
Use Drop-Downs in Setting Page


Samq64:
### Is your request related to a problem? Please describe.
I don't like the way these look and they use up a lot of space.

![Untitled](https://user-images.githubusercontent.com/81489795/136834765-f58b8059-27e1-4503-b20e-d9b6042a1850.png)

### Describe the solution you'd like

Use drop-downs instead.

### Describe alternatives you've considered

N/A

### Additional context

N/A

Platformer444:
> Some addons probably deserve but others could probably stay how they are. For instance, the first addon in your picture probably doesn't need the dropdowns as there are only 2 options. The one below probably could use a dropdown because it does take a lot of room and is pretty repetitive.

I think we can make an option for select whether it has to be a dropdown or the default.
Or
We can make the select with two options go with the default and the select with more than two options go with the dropdown.

PR #6547, opened 2023-08-18
Allow incrementing other input fields with `editor-number-arrow-keys`

### LINKED ISSUE #6546: Allow for increment via arrows also in sprite info area
### Why this would be helpful

Many users with discalculia problems do not understand numbers intuitively. Instead they can easily understand the meaning of increasing/decreasing the x/y/size/direction features of a sprite by using the arrow keys in the x/y/size/direction controls in the info area

### My suggestion

Allow users to increment/decrement values of the x/y/size/direction controls in info area by using the same arrow strategy adopted by the "Arrow key incrementation in editor" addon

### Possible alternatives

_No response_

### Additional context

_No response_

mybearworld:
Resolves #6546

### Changes

Added `input_input-form` to the potential matched input fields.

### Reason for changes

So that other numeric input fields can be incremented via arrow keys as well. 

### Tests

Tested in Edge.


Samq64:
Could the addon's description be updated to mention the other inputs?

mybearworld:
> Could the addon's description be updated to mention the other inputs?

Done

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -87,7 +87,11 @@ export default async function ({ addon }) {
 
   document.body.addEventListener("keydown", (e) => {
     if (addon.self.disabled) return;
-    if (!e.target.classList.contains("blocklyHtmlInput")) return;
+    if (
+      !e.target.classList.contains("blocklyHtmlInput") &&
+      !e.target.classList.contains(addon.tab.scratchClass("input_input-form"))
```
WorldLanguages, line 92:
```suggestion
      !e.target.className.includes("input_input-form_"))
```

WorldLanguages: (please test my suggestion to make sure it works)

WorldLanguages:
Also, some people suggested a `0.5` setting, I think we could add it.

DNin01:
Related if we're going to be adding more options to the single-select setting: #3642

Secret-chest:
> Also, some people suggested a 0.5 setting, I think we could add it.

I would also find 0.25 and 0.125 useful.

rgantzos:
> > Also, some people suggested a 0.5 setting, I think we could add it.
> 
> I would also find 0.25 and 0.125 useful.

I don't think it's worth getting into the thousandths, but a custom option would be very helpful.

Secret-chest:
> I don't think it's worth getting into the thousandths, but a custom option would be very helpful.

It's not really "into the thousandths", 0.125 is 1/8.

DNin01:
Let's talk about incrementation settings in #6555.

DNin01 (dismissed):
Doesn't seem to work properly in the other input fields. Only if you type in a number after incrementing it does the change stick.

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/e3aab3c3-881b-4e7e-be34-5b623ff1305c

addons/editor-number-arrow-keys/userscript.js
mybearworld: But why not just go with the more safe option?

mybearworld:
> Doesn't seem to work properly in the other input fields. Only if you type in a number after incrementing it does the change stick.

...Oh. That's not ideal.

WorldLanguages:
> Doesn't seem to work properly in the other input fields. Only if you type in a number after incrementing it does the change stick.

It's likely those inputs are handled by React, and React isn't getting the onchange event, so it doesn't update its internal state.

Try one of these:
```js
el.dispatchEvent(new Event("input", { "bubbles": true }));
// or
el.dispatchEvent(new Event("change", { "bubbles": true }));
```

mybearworld:
> It's likely those inputs are handled by React, and React isn't getting the onchange event, so it doesn't update its internal state.

Neither's working.

WorldLanguages:
Try this:

```js
el.dispatchEvent(new KeyboardEvent("keydown"));
el.value = 32;
el.dispatchEvent(new KeyboardEvent("keyup"));
```

WorldLanguages:
Looks like the Scratch component we should be looking for (in order to support, for example, changing the sprite size from the properties panel) is https://github.com/scratchfoundation/scratch-gui/blob/develop/src/components/forms/buffered-input-hoc.jsx

WorldLanguages:
Quite difficult to trigger the change event. Trying to find out why it doesn't work...

addons/editor-number-arrow-keys/userscript.js
mxmou: There are two different `input_input-form` classes - one from scratch-gui and one from scratch-paint. `scratchClass` will return the scratch-gui one.

WorldLanguages: > There are two different `input_input-form` classes

TBH, I didn't know. But it's always a bad idea to use scratchClass inside a querySelector.

WorldLanguages:
> Quite difficult to trigger the change event. Trying to find out why it doesn't work...

I'll need to look into some React internals with the browser debugger. Please ping me if I take too long

WorldLanguages:
Well, this was hard, but it's actually quite simple.
React registers `el.value` as a getter/setter and we have to circumvent that.

```js
const el = document.querySelector(`[class*=input_input-small]:not([placeholder])`); // just for testing
const newSizeValue = 80;

Object.getOwnPropertyDescriptor(el.constructor.prototype, "value").set.call(el, newSizeValue);
el.dispatchEvent(new Event("input", { bubbles: true }));
el.dispatchEvent(new Event("blur", { bubbles: true }));
```

WorldLanguages:
2 hours for this

brave_vN5bwTbSZL

WorldLanguages:
The editor has quite a lot of `input_input-form_` elements, even the find bar, sprite names, the project title, "ask and wait" prompts, etc.

We should probably be more restrictive with those to avoid bugs.

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -105,6 +106,39 @@ export default async function ({ addon }) {
...
+      if (!inputMap.has(e.target)) inputMap.set(e.target, { time: null, steps: -1 });
+      inputMap.get(e.target).time = currentTime;
+      const newNumOfSteps = (inputMap.get(e.target).steps += 1);
+
+      if (newNumOfSteps === FLUSH_AFTER_N_STEPS) {
+        flushInput();
+        inputMap.delete(e.target);
+      } else {
+        setTimeout(() => {
+          if (inputMap.get(e.target).time === currentTime) {
```
Samq64, line 133: `time` is sometimes undefined here.

WorldLanguages: Right, forgot to change this after I made it so entries are deleted.

`?.` would probably fix this, but I'd like to check

WorldLanguages:
> The editor has quite a lot of `input_input-form_` elements, even the find bar, sprite names, the project title, "ask and wait" prompts, etc.
> 
> We should probably be more restrictive with those to avoid bugs.

Anyone have a list of the inputs we should try to support? All other inputs wouldn't be supported.

I can think of these:
- Sprite properties (X/Y coordinates, direction, size, NOT sprite name)
- Number inputs in the costume editor (NOT costume name)

CST1229:
> Anyone have a list of the inputs we should try to support? All other inputs wouldn't be supported.

- The elements you said
- Addon elements:
	- Project video recorder duration/delay
	- Variable manager??

WorldLanguages:
> * Addon elements:
>   * Project video recorder duration/delay
>   * Variable manager values??

These sound OK, I doubt anyone will find it useful in the project video recorder number inputs, but it won't hurt.

> * List items???

Probably not.

WorldLanguages:
> * Variable manager values??

I will discard this idea for now, in case we want to add up/down to move between rows in the future.

addons/editor-number-arrow-keys/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Arrow key incrementation in editor",
-  "description": "Use the up and down arrow keys to increase or decrease number values in block inputs.",
+  "description": "Use the up and down arrow keys to increase or decrease number values in numeric input fields.",
```
WorldLanguages, line 3: Maybe we could add examples:

`numeric input fields, such as the input in the "wait () seconds" block, or the size in the sprite properties.`

WorldLanguages (approved):

DNin01 (approved):
Seems okay to me.

This doesn't work with the `color-inputs` addon at the moment, but that's okay, unless someone doesn't mind adding support for it?

addons/editor-number-arrow-keys/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Arrow key incrementation in editor",
-  "description": "Use the up and down arrow keys to increase or decrease number values in block inputs.",
+  "description": "Use the up and down arrow keys to increase or decrease number values in numeric input fields, such as the input in the \"wait () seconds\" block or the \"size\" field in the sprite properties.",
```
DNin01, line 3: I would suggest less specific examples:
```suggestion
  "description": "Use the up and down arrow keys to increase or decrease number values in numeric input fields such as those in blocks or the sprite properties area.",
```
Not sure if this suggestion is ideal, but to me it feels like the examples in the current description are a little too "specific" and don't cover every input field they should or something? To me, it should be clear that number inputs in _all_ blocks can be adjusted, and that _all_ numerical sprite properties can be adjusted too.

WorldLanguages:
> This doesn't work with the `color-inputs` addon at the moment, but that's okay, unless someone doesn't mind adding support for it?

Huh, good catch.

Issue #6048, opened 2023-05-14
Scratch Addons Broke(For me)


Ronxlotl:
### Description

Scratch addons v1.31.2 became broken the extention page and the dropdown appeared blank, and all addon weren't appearing.

### Steps to reproduce

I don't know any fixes.

### Environment

Browser: <!Brave>
Scratch Addons version: <1.31.2>
Operating system: <!Windows11>

WorldLanguages:
Can't help if we don't know your browser.

Ronxlotl:
> Can't help if we don't know your browser.

Brave Is a browser

Ronxlotl:
> > Can't help if we don't know your browser.
> 
> Brave Is a browser

Wait hold on i'm editing it for other devs to know my browser

WorldLanguages:
Oh I see, it was a markdown mistake.

WorldLanguages:
Try going to `chrome://extensions`, enable developer mode, find Scratch Addons in the list and inspect background/background.html

Ronxlotl:
Theres an "Find Errors" Feature For Brave I'm Screenshotting Two Errors That Brave Found

Ronxlotl:
Yeah Theres 90 Errors Name "ERR_FILE_NOT_FOUND"

WorldLanguages:
I think those can be ignored, anything else?
Post screenshots if you can

Ronxlotl:
Hold on

Ronxlotl:
The extenstion Page Blank
![Screenshot 2023-05-14 111501](https://github.com/ScratchAddons/ScratchAddons/assets/111940602/eb4646d8-7987-43a4-8be9-db2231ac1a6e)

WorldLanguages:
I meant screenshots of any errors in the inspect view

WorldLanguages:
Did you try restarting your browser? Type `brave://restart` into the address bar and press enter

Ronxlotl:
It still doesn't fix scratch addons

And heres the screenshots
![Screenshot 2023-05-14 112351](https://github.com/ScratchAddons/ScratchAddons/assets/111940602/7c627329-89f1-4c1e-bb51-6c578280de1b)
![Screenshot 2023-05-14 112417](https://github.com/ScratchAddons/ScratchAddons/assets/111940602/3c19e323-eaef-40c4-aa84-f3ec79db876f)
[...]

Issue #5306, opened 2022-11-01
Allow more control when Scratch Addons fails to load


DNin01:
### Why this would be helpful/solve a problem

Recently, someone had a situation: #5289 where their installation of SA would not load its setting page or its addons. That's not fun to experience, especially when you can't do anything. See this image from their issue:
![The setting page without any buttons or addons](https://user-images.githubusercontent.com/80355301/198852704-7776a088-b96f-4e03-b61e-8749479ab2eb.png)

### What this change does

There should be some way or guide to fix the problem, anything from a tip to restart the browser to a way to get your settings file, or maybe even a way to start over and reset Scratch Addons back to factory settings for those who wouldn't mind.
![A mockup. A warning message saying addons failed to load with a button below for recovery options.](https://user-images.githubusercontent.com/106490990/198857955-c9d3bb48-5818-4aa1-b51c-2847310e9e8f.png)

### Possible alternatives

We can discuss other ways to solve this on #5291.

### Demonstration(s) and additional context

Nothing else to say.

WorldLanguages:
My idea is:
- Creating a new page, such as `/webpages/error/index.html`
- It's not strictly necessary to use Vue for this page. It's also not necessary to add a lot of CSS to it (ugly page > no page)
- This new page would be localized through `chrome.i18n` and we should not test the case where that API is broken. If we ever use our own localization system for our webpages, we should make sure a localization library error doesn't crash this new page.
- This new page would be opened under the following cases:
  1. Unsupported browser (we would open this page instead of scratchaddons.com/unsupported-browser - details TBD. The idea is to allow users to export their settings even if the browser is unsupported)
  2. The settings page crashes (this might be a side effect of the background page crashing, of course)
  3. If `/popup/index.html` crashes, we could make sure to open the full-screen settings page automatically through `chrome.tabs.create()`. If the full-screen settings page crashes too, then the user will be redirected to the error page because of the point above.
- This new page would have a "clear all data" button, which would clear both `chrome.storage.sync` as well as `chrome.storage.local`. The user must understand that their settings will disappear from the universe if they haven't exported them.
- This new page would have an "export settings" button. We should make sure the code that exports the settings supports as many browsers as possible, and make sure it isn't buggy.

WorldLanguages:
This idea is easier said than done. I'll be looking at incremental ways to add error handling instead of trying to tackle everything at once.

WorldLanguages:
[...]

Issue #6448, opened 2023-07-24
Storing a daily backup of extension settings on scratch.mit.edu origin


WorldLanguages:
### Why this would be helpful

Apparently it's quite usual for Chrome extensions to get "corrupted" and the user needs to "repair" them. Then, the user may lose all of their settings. We're probably not causing this by ourselves, but we can still offer some type of solution.

Meanwhile, it's quite unusual to get logged out of a website you use frequently, unless on incognito mode or if you manually clear data.

### My suggestion

- If the user visits a Scratch page and it's been more than 24 hours, store a copy of the addon settings on IndexedDB on the scratch.mit.edu origin. (Other extensions may be able to read them)
- Provide a way to rollback to this copy. If storage is apparently empty and this copy exists, notify the user about it so we can be sure they are aware.

### Possible alternatives

Possibly storing a copy in extension IndexedDB (instead of scratch.mit.edu) might be equivalent. I don't have any way to know because it's hard to know why extensions get corrupted. I'm not sure if only `chrome.storage` is affected or if everything stored by the extension is.
I would imagine IndexedDB from all origins are stored in a single file by browsers, so I'd believe it would be equivalent and easier to implement, but I'm not sure.

### Additional context

This happened to griffpatch once (after a bad shutdown, apparently) and we get frequent feedback about it.
Haven't seen evidence of this happening to non-Windows users.

WorldLanguages:
It's quite likely Chromium decides an extension might be "corrupt" if it cannot verify the local version of it exactly matches the Chrome Web Store version. That verification can go wrong for various reasons.

Old screenshot I found the web:
![image](https://github.com/ScratchAddons/ScratchAddons/assets/17484114/27cbf40e-ee16-4c9f-9a96-7a177be913a6)

WorldLanguages:
Clicking "repair" probably installs the extension "from scratch" again, so that's probably why the storage is reset. It's possible IndexedDB on the extension origin is also cleared through this process.

![image](https://github.com/ScratchAddons/ScratchAddons/assets/17484114/f51f980e-49d9-4978-a682-00b82f1e5f4e)

DNin01:
That happened to me once, with a different extension. I forgot if I was accidentally messing around with it thinking it was a "from-source" version, though, but it probably wasn't that.

[...]

PR #6572, opened 2023-08-21
Create error page, redirect to it on typical settings page crashes

WorldLanguages:
Related to / mitigates #5306
Related to #6048
Related to #4949

### Changes

I posted this idea a while ago: https://github.com/ScratchAddons/ScratchAddons/issues/5306#issuecomment-1299223233

This PR creates an error page and displays it in the following situations:
- Invalid addon.json file during development (already existed since #4949, now also supports Firefox)
- The settings page crashed or doesn't work because `chrome.storage` is unavailable/broken (#6048) which happens quite often in Chrome

This page also provides a link/button to reset all extension settings. It also recommends restarting the browser.

This page won't be localized for now. I'm open to adding more CSS, but it's not very important. Also, this page won't support dark mode.

### Before: blank settings page

image

### After: redirect to error page

brave_dQHnZRIZ3X

### DIFF
--- a/webpages/error/export.js
+++ b/webpages/error/export.js
@@ -1,4 +1,0 @@
-document.querySelector("#export-settings-btn").onclick = () => {
-  // TODO
-  window.exportedSettings = true;
-};
--- a/webpages/error/export.js
+++ b/webpages/error/export.js
@@ -1,0 +1,4 @@
+document.querySelector("#export-settings-btn").onclick = () => {
+  // TODO
+  window.exportedSettings = true;
+};
--- a/webpages/error/reset-settings.js
+++ b/webpages/error/reset-settings.js
@@ -33,7 +33,7 @@ document.querySelector("#reset-settings-btn").onclick = () => {
 
     (async () => {
       const dbs = window.indexedDB.databases ? await window.indexedDB.databases() : null;
-      const dbNames = dbs ? IDB_DATABASES : dbs.map((db) => db.name);
+      const dbNames = !dbs ? IDB_DATABASES : dbs.map((db) => db.name);
       dbNames.forEach((name) => {
         window.indexedDB.deleteDatabase(name);
       });
--- a/webpages/error/reset-settings.js
+++ b/webpages/error/reset-settings.js
@@ -13,21 +13,34 @@ document.querySelector("#reset-settings-btn").onclick = () => {
 
   try {
     chrome.storage.local.clear();
-  } catch (err) {}
+  } catch (err) {
+    console.error(err);
+  }
   try {
     chrome.storage.sync.clear();
-  } catch (err) {}
+  } catch (err) {
+    console.error(err);
+  }
   try {
     localStorage.clear();
-  } catch (err) {}
+  } catch (err) {
+    console.error(err);
+  }
   try {
+    // We should try to list all IndexedDB databases here so that
+    // we can actually clear them all in Firefox.
+    const IDB_DATABASES = ["notifier", "messaging"];
+
     (async () => {
-      const dbs = await window.indexedDB.databases();
-      dbs.forEach((db) => {
-        window.indexedDB.deleteDatabase(db.name);
+      const dbs = window.indexedDB.databases ? await window.indexedDB.databases() : null;
+      const dbNames = dbs ? IDB_DATABASES : dbs.map((db) => db.name);
+      dbNames.forEach((name) => {
+        window.indexedDB.deleteDatabase(name);
       });
     })();
-  } catch (err) {}
+  } catch (err) {
+    console.error(err);
+  }
 
   setTimeout(() => {
     alert("All settings were reset. The extension will reload.");
--- a/webpages/error/reset-settings.js
+++ b/webpages/error/reset-settings.js
@@ -32,5 +32,5 @@ document.querySelector("#reset-settings-btn").onclick = () => {
   setTimeout(() => {
     alert("All settings were reset. The extension will reload.");
     chrome.runtime.reload();
-  }, 1500)
+  }, 1500);
 };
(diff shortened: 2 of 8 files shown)


WorldLanguages:
We could also add a button so that users can send feedback from this page. The "unsupported browsers" version of this page should also get the up-to-date minimum requirements, somehow.

Also, this is quite important as the manifest v3 upgrade might leave some users behind, and we want them to *at least* be able to export their settings.

DNin01:
Just so we know, clicking a link with an href to `chrome://restart` does not work. I just tested it. You must manually type in into the address bar.

webpages/error/reset-settings.js
```diff
@@ -0,0 +1,36 @@
+document.querySelector("#reset-settings-btn").onclick = () => {
+  if (!window.exportedSettings) {
+    const res = confirm(
+      "Are you sure you want to CLEAR ALL SETTINGS? If you haven't exported your settings first, click Cancel."
+    );
+    if (!res) return;
+    const res2 = prompt("Type 123 to confirm. You will lose all your Scratch Addons settings.");
```
DNin01, line 7: A better confirmation command might be "YES", which is what some command-line tools do.

WorldLanguages:
> Just so we know, clicking a link with an href to `chrome://restart` does not work. I just tested it. You must manually type in into the address bar.

Yeah, it's not a link. Also, the URL is different in Firefox.

Samq64:
Is it possible to display the error in the popup/settings page?

Samq64:
> The "unsupported browsers" version of this page should also get the up-to-date minimum requirements, somehow.

The easiest solution would be to link to https://scratchaddons.com/docs/getting-started/installing/#from-extension-stores.

webpages/error/reset-settings.js
BroJac5246: Or something that can't be ignored like "Delete my data," "Delete my settings," etc.

webpages/error/reset-settings.js
```diff
@@ -0,0 +1,36 @@
...
+  } catch (err) {}
+  try {
+    chrome.storage.sync.clear();
+  } catch (err) {}
+  try {
+    localStorage.clear();
+  } catch (err) {}
+  try {
+    (async () => {
+      const dbs = await window.indexedDB.databases();
```
apple502j, line 25: Not supported in Firefox.

webpages/error/reset-settings.js
WorldLanguages: I think I decided to use 123 because if we ever send this for translation, it would avoid issues such as translators accidentally translating "YES" into their language.

WorldLanguages:
> Is it possible to display the error in the popup/settings page?

Typically, the settings page is broken if `chrome.storage` malfunctions. We can add some code that redirects to this page if storage is clearly not working properly.

webpages/error/export.js
```diff
@@ -0,0 +1,4 @@
+document.querySelector("#export-settings-btn").onclick = () => {
+  // TODO
```
WorldLanguages, line 2: Note to self: it's possible chrome.storage will be broken here, so settings won't be able to be exported.

Related: #6448

WorldLanguages:
I suggest showing the error page in these 2 cases, for the 1.35.0 release:
- When going to the full screen settings page and `chrome.storage` is malfunctioning (see #6048) Also, I won't consider this to be blocked by the work in progress Vue 3 settings page PR.
- In dev mode, when an addon.json file couldn't be parsed properly. In the future, we can extend it to more types of errors.

I'm unsure whether this page will be localized. This is still better than the previous behavior (a broken settings page and an also unlocalized error message that didn't work in Firefox, respectively).

Possibly, the "export settings as file" button won't be available in the first version of this page. I should create an issue after this is merged without that button. Localizing this page also deserves an issue.

WorldLanguages:
If anyone wants to help with adding A BIT of CSS, go ahead. But this page will not distinguish light/dark mode, and it should be accessible and easily readable by anyone.

webpages/error/reset-settings.js
```diff
@@ -0,0 +1,49 @@
...
+    console.error(err);
+  }
+  try {
+    // We should try to list all IndexedDB databases here so that
+    // we can actually clear them all in Firefox.
+    const IDB_DATABASES = ["notifier", "messaging"];
+
+    (async () => {
+      const dbs = window.indexedDB.databases ? await window.indexedDB.databases() : null;
+      const dbNames = dbs ? IDB_DATABASES : dbs.map((db) => db.name);
```
CST1229, line 36:
```suggestion
      const dbNames = !dbs ? IDB_DATABASES : dbs.map((db) => db.name);
```
I think IDB_DATABASES is only supposed to be used if it coudln't get `dbs`.

webpages/error/reset-settings.js
DNin01: I think if it is a word like "YES" it should match the localized string too.

WorldLanguages: As a Spanish speaker I can see that go wrong as "si" and "sí" are different strings.

DNin01: Maybe in that case "123" would be the magic command, otherwise it would be a word?

WorldLanguages: Hmm, that sounds confusing.

mxmou: I don't see why "yes" would be better than "123". The only reason the user has to type anything is that they don't click OK without reading the message - the exact sequence of characters doesn't matter.

WorldLanguages:
Created issues:
- Export settings button: #6669
- Localization of this page: #6670

Not going to create an issue on making this page more visually pleasing for now...

Issue #6440, opened 2023-07-24
Potential additions to `.vscode` directory (default VSC settings)


WorldLanguages:
We already added a few default settings here: #6100

\
Here's a list of other things we could use this folder for (most of this are taken from #3968)
- Recommend VSCode extensions (Prettier, ESLint, etc.)
- Recommend VSCode extensions for Vue 3 development (possibly on `/webpages/.vscode` if that's possible)
- Check JSON schemas to more files
- File nesting
- Set Prettier as the default formatter

#3968 proposes we enable format on save, but I'm not sure if that's a good idea.

It would also be nice to find potential solutions to some annoyances, such as the great amount of subfolders on the addons directory.

PR #6593, opened 2023-08-25
Recommend VSCode extensions

Samq64:
Progress on #6440

### Changes

Adds all the extension recommendations from #3968 except for Vetur since #6272 recommends extensions in the webpages directory. Prettier is also set to format on save. I didn't recommend `vsc-focused-folder` because Apple said it wasn't well maintained.

### Reason for changes

Extensions make development easier and I think formatting on save is better than the Prettier commits, although I didn't touch those here.


.vscode/settings.json
```diff
@@ -31,5 +31,8 @@
         }
       }
     }
-  ]
+  ],
+
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true
```
WorldLanguages, line 37: Format on save is a bit opinionated, idk...

Samq64: My idea was this would replace the Prettier action but I don't think most poorly formatted PR authors are using VSCode anyway. I can remove it.

Samq64: No, I'll undo that.

Samq64:
I removed the EditorConfig and SVGO extensions so now it's just Prettier and ESLint but I couldn't seem to get ESLint to work out of the box for some reason.

Samq64:
Waiting for #7134 to fix ESLint.

Samq64:
Could this be merged now since the ESLint config was migrated a while ago?

WorldLanguages (approved):

PR #6598, opened 2023-08-27
Fix middle-click-popup not working if language is set to Turkish/Japanese

Tacodiva:
If the language is set to Turkish, the middle click popup doesn't work. This happens because I accidentally assumed that the first input of the `() of []` block was the variable and the second was the sprite. This is not true in Turkish and maybe other languages.


Tacodiva:
Uhhhh, sorry about that. I accidentally merged branches the wrong way around :sweat_smile:

Samq64:
> Idk if having those commits are gonna be a problem when the PR they're from gets merged and I don't think I can remove them.

I think PRs are almost always squashed.

WorldLanguages:
PRs are squashed so as long as you agree with the "files changed" tab of the PR we're fine.

Tacodiva:
Yes the diff of this PR is good to go

WorldLanguages (approved):
The popup now works properly in Turkish.

apple502j (approved):

Issue #5101, opened 2022-09-12
Option to record for however long you want


0K9090:
### Why this would be helpful/solve a problem

I find it weird that you can only record for a certain amount of time. Sure, you can set the seconds to 999999 but it still ends in a couple of minutes.

And i don't recall that my project stopped when it downloaded the recording

### What this change does

you can record for how ever long you want to

### Possible alternatives

download a screen recorder, but it will most likely have a watermark

WorldLanguages:
> Sure, you can set the seconds to 999999 but it still ends in a couple of minutes.

5 minutes if the maximum. We do not properly validate the second count if it is typed, instead of using the input controls.

https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/mediarecorder/userscript.js#L227

WorldLanguages:
Do we have a public statement on why 5 min is the limit? I think we could allow users to bypass it if they know what they're doing.

DNin01:
> Do we have a public statement on why 5 min is the limit? I think we could allow users to bypass it if they know what they're doing.

I don't know, but it might be that longer files take longer to render or process or take up more memory and that it just wouldn't be feasible after some amount of recording.

WorldLanguages:
So I think it's mainly RAM usage (if browsers don't optimize this, all the video would be stored in RAM before it's stopped, hopefully it ends up in swap/disk) and big video file sizes.

BroJac5246:
> So I think it's mainly RAM usage (if browsers don't optimize this, all the video would be stored in RAM before it's stopped, hopefully it ends up in swap/disk) and big video file sizes.

Could we add an option to override the limit and include a warning?

WorldLanguages:
I think we could, but I wouldn't be sure what the best way would be.

apple502j:
Yeah, the last thing we want is a 10GB recording because someone forgot to turn off recording...

0K9090:
> Yeah, the last thing we want is a 10GB recording because someone forgot to turn off recording...

the only solution to that would be a playstation-like-thing where if it's on for too long it will automatically turn off if the user doesn't close the prompt

BroJac5246:
> > Yeah, the last thing we want is a 10GB recording because someone forgot to turn off recording...
> 
> the only solution to that would be a playstation-like-thing where if it's on for too long it will automatically turn off if the user doesn't close the prompt

We could also check if the user is interacting with anything.

[...]

PR #6603, opened 2023-08-27
Increase mediarecorder (project recorder) limit from 5 min to 10

WorldLanguages:
Resolves part of #5101

### Changes

- This does not let you record as long as you want.
- This, still, does not warn you that the recording will stop after 10 minutes, even if you entered more than 600 seconds in the input. That will still be tracked as part of #5101
- I think we capped at 5 minutes because of potential RAM limitations or accidentally wasting hard drive space (for example see comment https://github.com/ScratchAddons/ScratchAddons/issues/5101#issuecomment-1247550130) but I think we can all agree 10 minutes, while being double, is still not a lot, and reasonable. We can always try to increase the limit later. This is a very simple change.

### Tests

Is testing necessary?

### DIFF
--- a/addons/mediarecorder/userscript.js
+++ b/addons/mediarecorder/userscript.js
@@ -35,7 +35,7 @@ export default async ({ addon, console, msg }) => {
       const recordOptionSecondsInput = Object.assign(document.createElement("input"), {
         type: "number",
         min: 1,
-        max: 300,
+        max: 600,
         defaultValue: 30,
         id: "recordOptionSecondsInput",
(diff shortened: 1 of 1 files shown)


WorldLanguages:
If this gets merged, we should add a latestUpdate to the addon before release.

DNin01 (commented):
This might be a good time to add something as simple as a note to the addon or edit the description to tell the user that 10 minutes is the maximum.

apple502j (approved):

PR #6611, opened 2023-08-31
New Addon: Larger "Save Now" button

### LINKED ISSUE #5189: Bigger save now button "hitbox"

A lot of the time when I want to manually save, I have to move the mouse pointer somewhat precisely to hit the Save Now button in the top-right corner.

### What this addon does

This addon increases the size of the Save Now button's clickable area by increasing its padding height. This means that even if your mouse pointer lands slightly above or below the Save Now button, it can still click it.

Size before:
The hit area is only on top of the text, not very forgiving

Size after:
The hit area stretches slightly above and below the text for a bigger clickable button area

I would use this addon every day if I used Scratch Addons daily.

### Possible alternatives

Get used to Ctrl+S.

I already have a working userstyle for this that I will share in a PR soon if you think this addon is worth it.

Also, any name suggestions?

BroJac5246:
Resolves #5189

### Changes

Makes the "Save Now" button in the editor larger. Also includes a setting to determine if it should be highlighted (this is off by default because I want it to be unobtrusive since this addon is enabled by default).

I also updated some credits on other addons I worked on because I decided I want them to be different.

**Screenshots:**
Before:
Screenshot 2023-08-30 22 16 27

After:
Screenshot 2023-08-30 22 16 58

After (with "Highlight on hover" enabled):
Screenshot 2023-08-30 22 17 42

### Reason for changes

This was highly requested and is useful to all users because the current "Save Now" button is quite small.

### Tests

Tested on Chrome 115 on ChromeOS 115.
This still needs testing on Firefox and ideally Edge.


BroJac5246:
Note: I was unable to get the setting to dynamically enable + disable. It's probably pretty simple, but I wasn't able to figure that out. If someone can help with that then that would be amazing :D

addons/big-save-button/addon.json
```diff
@@ -0,0 +1,36 @@
...
+  "versionAdded": "1.35.0",
+  "enabledByDefault": true,
+  "credits": [
+    {
+      "name": "BroJac5246",
+      "link": "https://github.com/BroJac5246/"
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
```
DNin01, line 14: It would be nice to also include [`updateUserstylesOnSettingsChange`](https://scratchaddons.com/docs/reference/addon-manifest/#updateuserstylesonsettingschange).

addons/big-save-button/userstyle.css
```diff
@@ -0,0 +1,13 @@
+.save-status_save-now_2shdk span {
```
DNin01, line 1: Probably want to use `[class*=...]` here. [Learn more](https://scratchaddons.com/docs/develop/userstyles/best-practices/#avoid-targeting-hashed-class-names)

This should work:
```suggestion
[class*="save-status_save-now"] span {
```

addons/big-save-button/highlight.css
```diff
@@ -0,0 +1,3 @@
+.menu-bar_menu-bar-item_oLDa-:hover {
```
DNin01, line 1: Same thing here.
```suggestion
[class*="menu-bar_menu-bar-item"]:hover {
```

addons/big-save-button/highlight.css
DNin01, line 1: Ideally, the button should only be highlighted if it is clickable. Currently, the background is shown even when there is no text content, and probably while saving too.

DNin01 (changes requested):
Some things need just a bit of tweaking.

addons/big-save-button/addon.json
DNin01: > Note: I was unable to get the setting to dynamically enable + disable. It's probably pretty simple, but I wasn't able to figure that out. If someone can help with that then that would be amazing :D

✨ [`updateUserstylesOnSettingsChange`](https://scratchaddons.com/docs/reference/addon-manifest/#updateuserstylesonsettingschange) ✨

BroJac5246: I tried that but I was having some issues with it... I might try again tomorrow.

addons/big-save-button/userstyle.css
BroJac5246: Ah! Yes! I thought of this one and then forgot about it 🤦

Thanks :D

BroJac5246:
I've been super busy recently but I'll try to work on this over the weekend.

Samq64:
I don't think having the hover on by default is a problem.

BroJac5246:
> I don't think having the hover on by default is a problem.

Do you think it makes more sense? I can turn it on if we decide that's better.

WorldLanguages:
> I don't think having the hover on by default is a problem.

I would personally enable it by default (and I'm not even sure if it should be an option!)

addons/big-save-button/addon.json
WorldLanguages: any details? it should work as expected

BroJac5246: > any details? it should work as expected

It was acting up for me. But maybe I was doing something wrong and didn't realize it. I'm pretty busy but I should be able to work on this tomorrow or Saturday 🤞

addons/big-save-button/highlight.css
```diff
@@ -0,0 +1,3 @@
+[class*="menu-bar_menu-bar-item"]:hover {
```
DNin01, line 1: > Ideally, the button should only be highlighted if it is clickable. Currently, the background is shown even when there is no text content, and probably while saving too.

I wish we could use this:
```css
[class*="menu-bar_menu-bar-item"]:has([class*="save-status_save-now"] span):hover
```
but [`:has()` needs to be enabled through a flag in Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility).

DNin01: Anyone have any ideas of how we could make that work across browsers? Or if maybe we should disable that option in Firefox or something?

BroJac5246: > Or if maybe we should disable that option in Firefox or something?

I don't want to do that considering this is going to be enabled by default, and also since W_L suggested making this enabled by default (which makes sense).

DNin01: Wait, could we just select `[class*="save-status_save-now"] span` instead? It looks identical to me that way.

Oh, but with Compact editor on, the span element dips down a little, so there is actually a difference by selecting the spam element. This might actually be worth tackling because it means the clickable area is being extended beyond the menu bar. Would you say that's intentional?

BroJac5246: > Wait, could we just select `[class*="save-status_save-now"] span` instead? It looks identical to me that way.
> 
> Oh, but with Compact editor on, the span element dips down a little, so there is actually a difference by selecting the spam element. This might actually be worth tackling because it means the clickable area is being extended beyond the menu bar. Would you say that's intentional?

I'm thinking that's a bug with the compact editor? I'll need to check it out first. It can probably just be fixed, though.

addons/big-save-button/addon.json
BroJac5246: I tried again today and it didn't work... but then I tried again and it did. I have no idea what happened. All I know is it did work in the end, so hopefully this is fixed.

DNin01: Changes to addon manifests require an extension reload. Maybe that's why?

BroJac5246: > Changes to addon manifests require an extension reload. Maybe that's why? Anyway, glad to hear it's working.

I did reload it both times it messed up, so I'm not sure what it was.

Hopefully it doesn't matter now.

addons/big-save-button/highlight.css
Samq64: Can't `[class*="save-status_save-now"] span:hover` be selected directly?

addons/big-save-button/highlight.css
```diff
@@ -0,0 +1,3 @@
...
+  background-color: hsla(0, 0%, 0%, 0.15);
```
Samq64, line 2:
```suggestion
  background-color: var(--editorDarkMode-menuBar-border), hsla(0, 0%, 0%, 0.15);
```

addons/big-save-button/userstyle.css
Samq64, line 1: This userstyle can be simplified by using flex and it's compatible with the compact editor.
```css
[class*="save-status_save-now"] {
  height: 100%;
  padding: 0 12px;
  display: flex;
  align-items: center;
}
```

Samq64:
May I push my suggestions?

addons/big-save-button/highlight.css
Samq64: `[class*="save-status_save-now"]:hover` seems to work fine with my styles below.

BroJac5246:
> May I push my suggestions?

Sure! I've been really busy recently but since you've made progress, go ahead.

Samq64:
Has it been decided if the setting will be removed or not?

BroJac5246:
> Has it been decided if the highlight will be always enabled or not?

4 people, including me, agreed that it makes sense, and no one disagreed, so I would say yes.

DNin01:
> Has it been decided if the highlight will be always enabled or not?

I think enabling the highlight by default makes sense and seems like a good thing to bundle in this addon.

Samq64 (approved):

DNin01 (approved):
I guess it's fine for highlighting to not be an option.

Anyway, it looks like the addon is working well!

BroJac5246:
Will this be merged as soon as 1.36.0 releases? (Or maybe a little later if 1.36.1 comes out as well?)

WorldLanguages:
This is going for v1.37.0 which shouldn't take long

BroJac5246:
@WorldLanguages should this be merged?

addons/big-save-button/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Larger \"Save Now\" button",
+  "description": "Extends the size of the \"Save Now\" button in the editor, making it easier to click.",
+  "tags": ["editor", "editorMenuBar", "recommended"],
+  "versionAdded": "1.36.0",
```
WorldLanguages, line 5:
```suggestion
  "versionAdded": "1.37.0",
```

addons/big-save-button/addon.json
```diff
@@ -0,0 +1,21 @@
...
+  "enabledByDefault": true,
```
WorldLanguages, line 6: !! just a reminder

BroJac5246: That it's enabled by default?

WorldLanguages (approved):

addons/big-save-button/addon.json
WorldLanguages: yep

WorldLanguages: the relevant discussion is in the issue
just wanted to point out that it's intentional

WorldLanguages:
> This addon could have been a part of Customizable menu bar.

See comments:
- https://github.com/ScratchAddons/ScratchAddons/issues/5189#issuecomment-1663202919
- https://github.com/ScratchAddons/ScratchAddons/issues/5189#issuecomment-1699636415

Secret-chest:
> This addon could have been a part of Customizable menu bar.

Then it couldn't be enabled by default.

Samq64:
> > This addon could have been a part of Customizable menu bar.
> 
> Then it couldn't be enabled by default.

It could since all the optional are optional and disabled by default but you're right that most enabled by default addons don't have settings.

Secret-chest:
> It could since all the optional are optional and disabled by default but you're right that most enabled by default addons don't have settings.

Addons with "Customizable" usually do nothing in the default state.

PR #6617, opened 2023-09-03
Specify end of line sequence in VS Code settings

DNin01:
### Changes

Sets `files.eol` in `.vscode/settings.json` to "\n", which makes newly created files in Visual Studio Code use LF as the default end of line sequence in the Scratch Addons directory.

### Reason for changes

To try to reduce the number of times the Prettier status check fails because the files were not using the end of line sequence we were asking for. Not sure how much it will help, but hey, anything that helps still helps.

### Tests

Tested in Visual Studio Code.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "files.eol": "\n",
-  
+
   "search.exclude": {
     "libraries/thirdparty/**": true,
     "libraries/licenses/**": true,
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,6 @@
 {
+  "files.eol": "\n",
+  
   "search.exclude": {
     "libraries/thirdparty/**": true,
     "libraries/licenses/**": true,


apple502j:
https://github.com/ScratchAddons/ScratchAddons/blob/master/.editorconfig#L4

WorldLanguages:
> https://github.com/ScratchAddons/ScratchAddons/blob/master/.editorconfig#L4

Does VSCode respect that?

DNin01:
> > https://github.com/ScratchAddons/ScratchAddons/blob/master/.editorconfig#L4
> 
> Does VSCode respect that?

Apparently you need an [extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig). We could migrate to `.vscode/settings.json`.

If we do that, we probably shouldn't just specify `files.eol`, I realize.

WorldLanguages:
Looks like VSC needs a plugin (had no idea) so we should duplicate all editorconfig values into `.vscode` when possible.

PR #6620, opened 2023-09-03
Update README image URL

rgantzos:
### Changes

Replaces the README popup image with the scratchaddons.com image.

### Reason for changes

The old one looked old and outdated, especially because it highlighted a 2.0 page.


README.md
```diff
@@ -19,7 +19,7 @@ Scratch Addons combines new and existing features and themes for the [Scratch](h
 
 <div align="center">
 
-![Scratch Addons screenshot](.github/readme-images/screenshot.png)
+![intro-hero](https://github.com/ScratchAddons/ScratchAddons/assets/86856959/3cd47267-df00-4f0c-babe-73b3e6c4e4f0)
```
DNin01, line 22: We should probably store the image in the repository, inside `.github/readme-images/`.

DNin01 (changes requested):

Samq64:
> The old one looked old and outdated, especially because it highlighted a 2.0 page.

That was kind of the point since I was trying to highlight Scratch 2 to Scratch 3.

README.md
WorldLanguages: Also why remove the descriptive alt text?

WorldLanguages:
I think we could keep both images in sync. Related to other issues I don't have the energy to search #for.

README.md
Samq64: Why not use the version from `website-v2` so they stay in sync?

DNin01 (approved):
Fetching from scratchaddons.com works, and means we only have to update the image in one place.

Samq64 (approved):

Samq64:
Actually, could the previous one be deleted?

DNin01:
> Actually, could the previous one be deleted?

Done.

Hans5958 (approved):
LGTM

PR #6622, opened 2023-09-03
Copy `.editorconfig` properties to VS Code settings

### LINKED ISSUE #6621: Copy `.editorconfig` configuration to `.vscode/settings.json`
### My suggestion

`.editorconfig` is kind of like a standard configuration file that is meant to be usable in all code editors and IDEs.

In Visual Studio Code, you need to install the [EditorConfig for VS Code extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) to use it, from what I know. We already have a [`.vscode/settings.json`](https://github.com/ScratchAddons/ScratchAddons/blob/master/.vscode/settings.json) file, so we might want to try specifying each configuration from `.editorconfig` in `.vscode/settings.json` as well so that no extension is required. **We would not remove `.editorconfig`.**

Related: #6617

### Possible alternatives

Recommend the EditorConfig extension: #6593

@apple502j @WorldLanguages

DNin01:
Resolves #6621
Supersedes #6617

### Changes

Copies the configuration from `.editorconfig` into `.vscode/settings.json`:

https://github.com/ScratchAddons/ScratchAddons/blob/ce838780c0471cbfd94e6980408974c1e9db2195/.editorconfig#L3-L12
↓↓↓

https://github.com/ScratchAddons/ScratchAddons/blob/df2daf23eed18cfdb4247b0d19d375513c7c1841/.vscode/settings.json#L2-L12

### Reason for changes

Visual Studio Code doesn't natively support `.editorconfig`; you have to install an extension.

This might even help with some of the Prettier status check failures.

### Tests

Tested in Visual Studio Code.


.vscode/settings.json
```diff
@@ -1,4 +1,16 @@
 {
+  "files.eol": "\n",
+  "editor.insertSpaces": true,
+  "editor.indentSize": 2,
+  "files.encoding": "utf8",
```
WorldLanguages, line 5: utf8 is probably the default (it's 2023 lol) maybe check which ones aren't needed bc they're already defaults?

DNin01: Sure, but better safe than sorry, IMO - it's always possible for someone to have adjusted their settings for some reason (maybe one of someone's own projects uses a different style), and we've had most of these in `.editorconfig` this whole time.

Are you sure we shouldn't explicitly specify settings that are already defaults?

WorldLanguages: I'm not sure how VSCode handles user defaults vs project defaults tbh, but yeah, better safe than sorry I guess

apple502j:
i'd rather recommend editorconfig ext than this, tbh

WorldLanguages:
> i'd rather recommend editorconfig ext than this, tbh

Why? It's not like we change the editorconfig very often. Doesn't hurt to change it twice.

DNin01:
> i'd rather recommend editorconfig ext than this, tbh

Related: #6593

DNin01:
> i'd rather recommend editorconfig ext than this, tbh

We _could_ do that, but the benefit of not requiring an extension to be installed is that newcomers who aren't familiar with extensions don't have to worry about it, though other extensions (like [Vue Language Support](https://marketplace.visualstudio.com/items?itemName=Vue.volar) if working with Vue becomes required) might be needed in the future so then that point doesn't really mean anything.

...Other than the fact that you don't have to install as much stuff, plus, why not use VS Code's built-in features anyway? Built-in is always better.

Hans5958:
Correct me if I wrong, but isn't this rendundant? Is there anything wrong that makes VS Code not using `.editorconfig`?

DNin01:
> Correct me if I wrong, but isn't this rendundant? Is there anything wrong that makes VS Code not using `.editorconfig`?

Although Git supports rendering EditorConfig, it surprisingly does not actually follow it. For that, you need to install the EditorConfig extension.

DNin01:
> > Although Git supports rendering EditorConfig
> 
> Do you mean VSC?

Yeah.

DNin01:
So yeah, @Hans5958, VS Code doesn't apply EditorConfig out of the box; to use it, you have to install the [EditorConfig extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig). We could do that, but it might be less beginner-friendly. `.vscode/settings.json` is more seamless, though it does require that you open the folder for the repository in VS Code.

DNin01:
So should we add these attributes to the `settings.json` or not? Is there any reason not to?

Samq64 (approved):

WorldLanguages (approved):

PR #6625, opened 2023-09-05
Add custom options to editor key incrementation

### LINKED ISSUE #6555: `editor-number-arrow-keys`: More options for how much to increment
### Why this would be helpful

The new `editor-number-arrow-keys` addon features a few options for how much to change a number by when pressing up/down. [It was decided](https://github.com/ScratchAddons/ScratchAddons/pull/6020#issuecomment-1620458388) that the options be part of a single-select setting, but there aren't very many options to choose from, there's just 0.01, 0.1, 1, 10, and of course, none.

### My suggestion

Some way to include more options such as larger or more precise units, namely 0.5 and 100.

### Possible alternatives

_No response_

### Additional context

Continuation of a conversation on #6547:
> @WorldLanguages
> Also, some people suggested a `0.5` setting, I think we could add it.

> @Secret-chest
> I would also find 0.25 and 0.125 useful.

> @rgantzos
> I don't think it's worth getting into the thousandths, but a custom option would be very helpful.

<sub>Yes, I am intentionally mentioning the commentors, if that even works in quotes.</sub>

Samq64:
Resolves #6555

### Changes

Adds a toggle to use custom values to `editor-number-arrow-keys`.

### Reason for changes

 Having more options could be useful.

### Tests

Tested on Brave.


WorldLanguages:
Do we have evidence of an actual user suggesting `100`?

Samq64:
> Do we have evidence of an actual user suggesting `100`?

Just DNin's original post I think.

DNin01:
#3642 would help with the clutter. At the moment, I don't know what to think of so many options one after another.

WorldLanguages:
Can we remove 100? :P

DNin01:
> Can we remove 100? :P

I think the only main reason not to include 100 would be due to space constraints.

WorldLanguages (approved):

DNin01 (approved):
Single-select overflow
Yeah.

Guess that's just life. We could maybe fix this later.

Samq64:
> Can we remove 100? :P

The popup still has two rows without the 100 option so I don't think it's big deal.

rgantzos:
> > Can we remove 100? :P
> 
> The popup still has two rows without the 100 option so I don't think it's big deal.

Even if it doesn't make a difference in terms of the UI, I don't know who would need to change a sprite position by 100, or brightness by 100, or rotation by 100 lol. It's a very large number that it would almost become an inconvenience for many uses.

Samq64:
> Even if it doesn't make a difference in terms of the UI, I don't know who would need to change a sprite position by 100, or brightness by 100, or rotation by 100 lol. It's a very large number that it would almost become an inconvenience for many uses.

I was thinking things like scores but sure I can remove it since it doesn't make much sense outside of variables.

BroJac5246:
> > Even if it doesn't make a difference in terms of the UI, I don't know who would need to change a sprite position by 100, or brightness by 100, or rotation by 100 lol. It's a very large number that it would almost become an inconvenience for many uses.
> 
> I was thinking things like scores but sure I can remove it since it doesn't make much sense outside of variables.

Yes, but if there are any use cases, why get rid of it? It's not like it takes any effort to have it present.

Secret-chest:
You can add 2, 4, 8, 16 to fix these :)

rgantzos:
There has been a lot of debate about the values. What's wrong with giving full customization and just allowing users to set the value themselves? There can still be a default. At this point the addon settings are limiting the capabilities of the addon.

Samq64:
> There has been a lot of debate about the values. What's wrong with giving full customization and just allowing users to set the value themselves? There can still be a default. At this point the addon settings are limiting the capabilities of the addon.

I think we wanted to avoid making a new setting type since we only support integers right now but I agree that would be better.

BroJac5246:
> > There has been a lot of debate about the values. What's wrong with giving full customization and just allowing users to set the value themselves? There can still be a default. At this point the addon settings are limiting the capabilities of the addon.
> 
> I think we wanted to avoid making a new setting type since we only support integers right now but I agree that would be better.

I don't think we should have arbitrary restrictions just because we don't support decimals in settings.

Secret-chest:
Unless we make it customisable, the requests will continue, because we can't satisfy everyone.

Samq64:
I can add float inputs into this PR but I think that needs more discussion and will probably cause the milestone to be bumped.

DNin01:
> I can add float inputs into this PR but I think that needs more discussion and will probably cause the milestone to be bumped.

Let's do that in another PR.

WorldLanguages:
> Unless we make it customisable, the requests will continue, because we can't satisfy everyone.

I did not see any requests from the feedback form, literally zero messages about this addon.

Personally I only observed some talk about an `0.5` option in the Discord server this addon was released.

WorldLanguages:
> > ...we only support integers right now...
> 
> Wait, why? What's hold it on so we can replace integers in favor of floats or something like that?

Adding more options right now does not mean we will never support non-integer number settings.
I don't really want to turn this PR into a discussion on whether those make sense.

WorldLanguages:
Also, let's not forget we do have "string" settings, so we can already support floats, but they won't be validated by the settings page as numbers.

Samq64:
Could 0.125 be removed so the settings fit on one row in the popup? It's also very close to 0.1.

WorldLanguages:
> Until we have full customization could 0.125 be removed so the each setting fits on one row in the popup? It's also very close to 0.1.

We could keep the "select" settings intact and add a single toggle for users that want to customize the three exact values through string/float settings.

Secret-chest:
> Until we have full customization could 0.125 be removed so the each setting fits on one row in the popup? It's also very close to 0.1.

Noooo! As the suggester of this, I was going to use it not because I like the number, but because it is a power of 2.

Samq64:
I removed the new options and added a custom option instead which currently uses a string setting. If a setting is invalid the block's value changes to NaN which can be undone with Ctrl+Z after unfocusing the input.

BroJac5246:
> > Until we have full customization could 0.125 be removed so the each setting fits on one row in the popup? It's also very close to 0.1.
> 
> Noooo! As the suggester of this, I was going to use it not because I like the number, but because it is a power of 2.

But the whole reason a setting should exist is that people will use it... otherwise why bother?

Also, it's in the thousandths, and that's pretty small.

Secret-chest:
> Also, it's in the thousandths, and that's pretty small.

Doesn't matter, it's 1/8.

DNin01 (approved):
Yeah, that's actually quite a bit better. I can't think of any other ways, so.

WorldLanguages:
Screen recording:

brave_hbmrXMqfdi

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -94,12 +94,21 @@ export default async function ({ addon }) {
...
-    const changeBy =
-      (e.shiftKey
+    let changeBy = e.code === "ArrowUp" ? 1 : -1;
+    if (addon.settings.get("useCustom")) {
+      // TODO: add proper float settings
+      changeBy *= e.shiftKey
+        ? parseFloat(addon.settings.get("shiftCustom"))
+        : e.altKey
+        ? parseFloat(addon.settings.get("altCustom"))
+        : parseFloat(addon.settings.get("regularCustom"));
```
WorldLanguages, line 104: Yeah, we need to be careful with NaN here. We should early-return or change by zero if the string setting can't be converted into a number.

addons/editor-number-arrow-keys/addon.json
```diff
@@ -92,7 +109,23 @@
...
+    {
+      "name": "Change on Alt+Key",
+      "id": "altCustom",
+      "type": "string",
+      "default": "0.1",
+      "if": { "settings": { "useCustom": true } }
+    },
+
+    {
+      "name": "Use custom values",
```
WorldLanguages, line 125: We can classify this as a "new" setting with latestUpdate.

WorldLanguages:
Note that the fact a number can fit into 8 character doesn't mean it is actually supported, but I decided to draw the line somewhere.

WorldLanguages (approved):

addons/editor-number-arrow-keys/addon.json
```diff
@@ -3,6 +3,12 @@
   "description": "Use the up and down arrow keys to increase or decrease number values in block inputs.",
   "tags": ["editor", "codeEditor", "recommended"],
   "versionAdded": "1.34.0",
+  "latestUpdate": {
+    "version": "1.35",
+    "isMajor": true,
+    "newSettings": ["useCustom", "regularCustom", "shiftCustom", "altCustom"],
```
WorldLanguages, line 9: I mean, technically all four are new, but...

```suggestion
    "newSettings": ["useCustom"],
```

WorldLanguages:
@Secret-chest Hope you're happy now 😂

BroJac5246:
> > Also, it's in the thousandths, and that's pretty small.
> 
> Doesn't matter, it's 1/8.

The same logic applies the other way around, too.

I'm not opposed to this, but only if people will use it. Will they?

Secret-chest:
Good.

Issue #6584, opened 2023-08-23
Buttons Not Aligned.  


MaterArc:
### Description

When you enable the Scratch 2.0 to Scratch 3.0 addon, on the 'my stuff' page, two of the buttons aren't properly aligned. 

### Steps to reproduce

1. Enable the Scratch 2.0 → 3.0 Addon
2. Go to the My stuff page
3. Look at the buttons with your eyes and process the information with your brain lol

### Demonstration(s) and additional context

fine lol
<img width="262" alt="Screenshot 2023-08-23 at 4 45 30 PM" src="https://github.com/ScratchAddons/ScratchAddons/assets/105017592/30d11019-de04-4eb3-9d63-28a6ccf511d8">

### Environment

Browser: Chrome version 116.0.5845.96
Scratch Addons version: 1.34.1
Operating system: macOS version 10.15.7

rgantzos:
It's supposed to be like that, just like on the normal My Stuff page. The Shared Projects and Unshared Projects tabs are subcategories of the All Projects tag. Here's what it looks like without the addon:
<img width="261" alt="Screenshot 2023-08-23 at 1 54 11 PM" src="https://github.com/ScratchAddons/ScratchAddons/assets/86856959/ee5947f6-54e6-415f-98d3-c6e6e1eb3774">

MaterArc:
Never really noticed that lol, someone can close it if they need. It just looks out of place to someone who didn't know that the original was indented to show categories. Anyways though, glad to hear it wasn't a bug or anything.

DNin01:
Works as intended.

Samq64:
I made an issue for this a while back too. There is an addon to use horizontal tabs on my stuff which I think makes it look either way.

Secret-chest:
The indent should be larger, because now it just looks misaligned.

WorldLanguages:
At some point, I was also confused at this. Unsure what could be done.

mxmou:
> At some point, I was also confused at this. Unsure what could be done.

Something like this?
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/cce3ef7d-698c-4529-bb0b-86e931227605)

WorldLanguages:
Yeah, that would be nice!

DNin01:
Fancy having the subtabs disappear when you aren't looking at a list of projects? (When My Studios or Trash is selected, the Shared Projects and Unshared Projects buttons would be hidden.)

[...]

PR #6630, opened 2023-09-06
scratchr2: improve My Stuff tabs

mxmou:
### Changes

* Implements the suggestion in https://github.com/ScratchAddons/ScratchAddons/issues/6584#issuecomment-1692206685
* That didn't look very good when the text on the tab had more than one line, so this also
  * makes it truncate instead of wrap;
  * reduces the padding, font size, and font weight to make that less likely.

image

### Reason for changes

With the previous design, it wasn't obvious that the indented tabs were subcategories.

### Tests

Tested in both browsers in different languages.


DNin01 (commented):
To me, this design looks a bit distracting - it doesn't completely fit the theme of Scratch's UI, IMO. Maybe making it blend in with the background more would help. Also, does the color change if you have a custom `dark-www` highlight color?

mxmou:
> To me, this design looks a bit distracting - it doesn't completely fit the theme of Scratch's UI, IMO. Maybe making it blend in with the background more would help.

Do you think this looks better?
image

> Also, does the color change if you have a custom `dark-www` highlight color?

It isn't currently affected by dark mode settings. I'll fix that.

DNin01 (approved):
Pretty good.

WorldLanguages:
brave_YmIjPA0UJz

WorldLanguages (approved):
Looks nice!

Secret-chest:
The dots should be thinner and the same colour as the unselected button border.

Samq64:
> The dots should be thinner and the same colour as the unselected button border.

But the unselected button are white like the background. Are you saying the dots should only be outlines?

mxmou:
> > The dots should be thinner and the same colour as the unselected button border.
> 
> But the unselected button are white like the background.

@Secret-chest wants the dots to be the same color as the button borders:
image

I think they look better if they're more visible.

Secret-chest:
OR make them thinner.

DNin01:
Yeah, I kind of feel like the dots a supplementary element and don't need to follow contrast guidelines as strictly.

I wonder what it would look like if they were more like dashes instead of thick dots, as suggested by @Secret-chest.

DNin01:
> I think they look better if they're more visible.

It's a tough decision, isn't it? I feel kinda torn too.

WorldLanguages:
> I think they look better if they're more visible.

I think I like the opposite, personally

GrahamSH-LLK:
> > > The dots should be thinner and the same colour as the unselected button border.
> > 
> > 
> > But the unselected button are white like the background.
> 
> @Secret-chest wants the dots to be the same color as the button borders: image
> 
> I think they look better if they're more visible.

This looks _way_ better

BroJac5246:
> This looks way better

"This" meaning the picture or the suggestion? I'm guessing the former?

GrahamSH-LLK:
> > This looks way better
> 
> "This" meaning the picture or the suggestion? I'm guessing the former?

The picture

WorldLanguages:
I think a majority would prefer this version, over the purple one in the OP (original post):

(reposting image from comment https://github.com/ScratchAddons/ScratchAddons/pull/6630#issuecomment-1713732728)

image

Secret-chest:
Let's vote:

* :+1: for light grey dots
* :-1: for dark grey dots
* :tada: for purple dots

DNin01:
Direct comparison:
image image image

DNin01:
I wonder if it could change depending on the state of the buttons (like hovered, active, etc.)?

BroJac5246:
> I wonder if it could change depending on the state of the buttons (like hovered, active, etc.)?

Interesting... but probably also overkill.

WorldLanguages (approved):

PR #6633, opened 2023-09-09
Re-add the fix footer setting as a new addon

### LINKED ISSUE #6542: Re-add fix footer option to infinite scrolling
### Why this would be helpful

One user said it made it more convenient to access the forums but they probably didn't know the navbar addon existed.

@mybearworld Has made a few good points though:

>> Removes the "fix footer" setting from infinite-scroll. This setting has been the default for a while and I have seen no evidence of appreciation for it.
>> ...
>> If users send feedback about wanting this feature back, we might readd it as an off-by-default setting.
>- https://github.com/ScratchAddons/ScratchAddons/pull/6426
>
> So, here's feedback! I used this feature a lot even outside of pages with infinite scrolling. It's a much more convenient way of accessing the footer and I think this addon shouldn't have been removed.
>
>> Without a infinite scrolling feture, the reason of someone to scroll to the bottom is to access the footer, but with it, the reason of someone to scroll to the bottom is to have more stuff; more content.
> - https://github.com/ScratchAddons/ScratchAddons/issues/5419#issuecomment-1345267393
>
> This logic seems flawed to me. Say you're scrolling through some comments, and someone asks something about the Terms of Use and you want to reference it. Without a "Fix Footer" feature, this requires going to a new tab, opening Scratch and going to the ToU there, while with such an option, you can just middle click that link.
>
>> Even Never Ending Reddit, the "infinite scrolling" of Reddit Enhancement Suite for old Reddit, doesn't offer a fixed footer, probably with the reason I mentioned.
>
> Just because someone else isn't doing it doesn't necessarily mean SA shouldn't be doing it.

### My suggestion

Re-add the fix footer setting, but with a different name and disabled by default. It's also probably be a good idea to make it only apply pages to with infinite scrolling this time.

### Possible alternatives

_No response_

### Additional context

_No response_

Samq64:
Resolves #6542

### Changes

Re-adds the fix footer setting from infinite scrolling as a new addon with a toggle to only apply to infinite scrolling pages.

### Reason for changes

A couple people wanted it back.

### Tests

Tested on Brave and Firefox. The setting isn't dynamic but neither is `infinite-scroll`.


addons/collapse-footer/addon.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "Auto-hiding footer",
+  "description": "Collpases the website footer and expends hovering over it. Prevously part of \"Infinite scolling\".",
```
DNin01, line 3: Typos:
```suggestion
  "description": "Collapses the website footer and expands while hovering over it. Prevously part of \"Infinite scolling\".",
```

addons/collapse-footer/addon.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "Auto-hiding footer",
+  "description": "Collapses the website footer and expands while hovering over it. Prevously part of \"Infinite scolling\".",
```
mxmou, line 3: This name and description would only make sense if the footer was always visible at the bottom of the screen without the addon.

addons/collapse-footer/userstyle.css
```diff
@@ -0,0 +1,31 @@
...
+  padding-top: 4px !important;
+  z-index: 20;
+  transition-duration: 0.5s;
+}
+.sa-collapse-footer #footer .lists dl {
+  margin-top: 0;
+}
+.sa-collapse-footer #footer:hover {
+  height: var(--footer-hover-height) !important;
+}
```
DNin01, line 23: The footer might look a little more pleasing if the top margin was consistent with the vanilla footer, only when the footer is expanded:
```suggestion
.sa-collapse-footer #footer .lists dl {
  margin-top: 0;
  transition-duration: 0.5s;
}
.sa-collapse-footer #footer:hover {
  height: var(--footer-hover-height) !important;
}
.sa-collapse-footer #footer:hover .lists dl {
  margin-top: 12px;
}
```

addons/collapse-footer/addon.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "Sticky footer",
+  "description": "Makes the website footer stick out slightly from the bottom of the window and makes it expand when hovering over it. Previously part of \"Infinite scolling\".",
```
DNin01, line 3: Potential description extension:
```suggestion
  "description": "Makes the website footer stick out slightly from the bottom of the window and makes it expand when hovering over it, allowing it to be quickly accessed from anywhere on the page. Previously part of \"Infinite scolling\".",
```

DNin01 (commented):
I actually really like the sound of this.

addons/collapse-footer/userstyle.css
DNin01: Wait, that doesn't cover `scratchr2`...

Samq64: I didn't bother changing the CSS from the old version except for the selectors.

DNin01 (approved):

addons/collapse-footer/userscript.js
```diff
@@ -0,0 +1,16 @@
+export default async function ({ addon, console }) {
+  if (!addon.settings.get("infiniteScroll")) {
```
WorldLanguages, line 2: We should check whether the addon is enabled in case the user disables infinite-scrolling, I think

WorldLanguages:
image

Did the "fix footer" setting from infinite-scrolling always look this bad? I didn't know the About/Community/etc. labels were visible all the time. Mandela effect or did the Scratch CSS ever change, causing this?
I went back to Scratch Addons v1.32.x and it also looks like this.

DNin01:
> image
> 
> Did the "fix footer" setting from infinite-scrolling always look this bad? I didn't know the About/Community/etc. labels were visible all the time. Mandela effect or did the Scratch CSS ever change, causing this? I went back to Scratch Addons v1.32.x and it also looks like this.

What's bad about it?

DNin01 (commented):
You know what might also be cool, an animation speed setting. We have this option in other addons.

Samq64:
> Did the "fix footer" setting from infinite-scrolling always look this bad? I didn't know the About/Community/etc. labels were visible all the time.

Maybe it could be an arrow similar to collapsing sprite properties? Can you think of any other alternatives?

BroJac5246:
> image
> 
> Did the "fix footer" setting from infinite-scrolling always look this bad? I didn't know the About/Community/etc. labels were visible all the time. Mandela effect or did the Scratch CSS ever change, causing this?
> I went back to Scratch Addons v1.32.x and it also looks like this.
> 
> EDIT: according to redguy it has always been this way.

Yes it was, this is why I am against re-adding this.

> > Did the "fix footer" setting from infinite-scrolling always look this bad? I didn't know the About/Community/etc. labels were visible all the time.
> 
> Maybe it could be an arrow similar to collapsing sprite properties?

It's not perfect, but that would be much better.

CST1229:
> Did the "fix footer" setting from infinite-scrolling always look this bad? I didn't know the About/Community/etc. labels were visible all the time. Mandela effect or did the Scratch CSS ever change, causing this?

IIRC it was like that on scratchr2 pages but they were not visible on scratch-www pages (or when using the `scratchr2` addon).

Samq64:
footer-arrow

Samq64:
I added an icon:

footer-arrow

> You know what might also be cool, an animation speed setting. We have this option in other addons.

Sure, I guess. I didn't add an option for instant though.

DNin01:
Do you think it should wait a fraction of a second before collapsing? Sometimes I find my mouse dipping into my system's taskbar briefly which causes the footer to stutter as it is hovered and immediately unhovered, plus, if you move the mouse up too fast it also collapses. The smoother and more forgiving it is, the better the UX will be. You can use JavaScript to do this:
```js
let collapseTimeout;

footer.addEventListener("mouseover", () => {
  footer.classList.add("expanded");
  if (collapseTimeout) {
    clearTimeout(collapseTimeout);
  }
});

footer.addEventListener("mouseout", () => {
  collapseTimeout = setTimeout(() => {
    footer.classList.remove("expanded");
  }, 200)
});
```

DNin01:
Another way is to add an invisible hover area (potentially done with `padding-top`) occupying the space the footer will take up once it has fully expanded, so if your mouse moves somewhere before the footer reaches it, it will stay expanded.

Those are some small reliability improvements that can add up.

DNin01 (commented):
I feel like maybe(...?) we should choose a different arrow icon, like the one used in `sprite-properties`. Or at least make the current one smaller.

Well, the current arrow is fine too.

Secret-chest:
> Sure, I guess. I didn't add an option for instant though.

Motion-sensitive people exist

DNin01 (commented):
> Motion-sensitive people exist

Oh yeah, I didn't think about that. Maybe there could be a separate switch to reduce motion or maybe we can just not have an animation at all depending on the user's reduced motion preference.

DNin01:
If we add an option to reduce motion, the animation could probably be replaced by a fade-in.

WorldLanguages:
Pending until we close translation strings: this PR and #6707

addons/collapse-footer/addon.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "Sticky footer",
+  "description": "Makes the website footer stick out slightly from the bottom of the window and makes it expand when hovering over it, allowing it to be quickly accessed from anywhere on the page. Previously part of \"Infinite scolling\".",
```
WorldLanguages, line 3: LGTM

PR #6651, opened 2023-09-12
Don't feature `longer-wiwo`

DNin01:
### Changes

Moves the 'Higher Character Limit in "What I'm Working on"' addon down from Featured to Others.

### Reason for changes

I don't like the sound of putting exploits higher up - advertising the method to extend the WIWO by 55 characters won't make the Scratch Team happy. Then again, we're an unofficial extension...

### Tests

Not tested.

### DIFF
--- a/addons/longer-wiwo/addon.json
+++ b/addons/longer-wiwo/addon.json
@@ -16,6 +16,6 @@
     }
   ],
   "versionAdded": "1.11.0",
-  "tags": ["community", "profiles", "featured"],
+  "tags": ["community", "profiles"],
   "enabledByDefault": false
 }


BroJac5246:
> I don't like the sound of putting exploits higher up

I can generally agree with this.
> advertising the method to extend the WIWO by 55 characters won't make the Scratch Team happy.

People do advertise this on Scratch itself and ST doesn't seem to care. And they could always fix it. Have they taken a public stance on it?
> Then again, we're an unofficial extension...

That said, it's in our best interest to remain on the ST's good side.

WorldLanguages (approved):
This addon reminds me of `Remix button in own projects` but you can already remix yourself by having 2 accounts and switching tabs, no extension/console needed...
That addon isn't featured though
We also have some addons that add "new" blocks but you can also get them through the backpack without extensions/console.

Samq64 (approved):

WorldLanguages:
Looks like we'll just un-feature this addon for now.

RedGuy12:
~~rip my addon~~

PR #6653, opened 2023-09-12
Better customizable link style settings UX + separate bold setting

### LINKED ISSUE #6503: Better customizable link style settings UX
### Why this would be helpful

The current way the customizable link style settings are displayed to the user have some issues:

1. There's a setting labelled default that isn't the default
1. That default setting is useless since it basically disables the addon
1. The bold setting can be confusing since links are already bold in a lot of places

The default setting was probably just left over from a last minute change.

### My suggestion

Replacing the link style setting with a toggle to make all links bold and an option to choose none as the underline style. This would also allow users to have bold underlined links.

### Possible alternatives

_No response_

### Additional context

image

Unfortunately I think strings are closed for this release.

Samq64:
Resolves #6503

### Changes

Changes the addon's settings to have just 2 select values, one for bold links and one for the underline style. Now link boldness can be changed separately from the underline and has the option to keep it as the default since Scratch has both bold and normal links.

### Reason for changes

To make the settings easier to understand.

### Tests

Tested on Firefox and Brave, but the non-bold links setting might still need a few selector adjustments. The transition code was also tested.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -144,7 +144,6 @@
   "featured-dangos-v2",
   "remember-collapsed-categories",
   "editor-number-arrow-keys",
-  "colorblind",
   "forum-post-countdown",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
@@ -162,6 +161,7 @@
   "material-forum-editor-buttons",
   "forum-quote-code-beautifier",
   "horizontal-mystuff-tabs",
+  "colorblind",
   "// For editor:",
   "editor-dark-mode",
   "editor-colored-context-menus",
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -144,6 +144,7 @@
   "featured-dangos-v2",
   "remember-collapsed-categories",
   "editor-number-arrow-keys",
+  "colorblind",
   "forum-post-countdown",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
@@ -161,7 +162,6 @@
   "material-forum-editor-buttons",
   "forum-quote-code-beautifier",
   "horizontal-mystuff-tabs",
-  "colorblind",
   "// For editor:",
   "editor-dark-mode",
   "editor-colored-context-menus",
--- a/addons/colorblind/addon.json
+++ b/addons/colorblind/addon.json
@@ -88,7 +88,7 @@
           "name": "Dotted"
         }
       ],
-      "default": "solid"
+      "default": "none"
     }
   ],
   "dynamicEnable": true,
--- a/addons/colorblind/addon.json
+++ b/addons/colorblind/addon.json
@@ -7,12 +7,6 @@
       "link": "https://github.com/DNin01"
     }
   ],
-  "info": [
-    {
-      "id": "feedback",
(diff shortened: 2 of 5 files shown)


addons/colorblind/addon.json
```diff
@@ -83,16 +88,12 @@
           "name": "Dotted"
         }
       ],
-      "default": "solid",
-      "if": {
-        "settings": { "links": "underline" }
-      }
+      "default": "solid"
```
mxmou, line 91: The default settings of "customizable XYZ" addons are usually the same as Scratch's defaults.
```suggestion
      "default": "none"
```

background/get-addon-settings.js
```diff
@@ -639,6 +639,18 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
...
+          // Transition v1.34 to v1.35
+          if (settings.links !== "underline") settings["underline-style"] = "none";
+          if (settings.links === "bold") {
+            settings.bold = "all";
+          } else {
+            settings["bold"] = "default";
+          }
+          delete settings.links;
+          madeAnyChanges = madeChangesToAddon = true;
+        }
```
mxmou, line 653: This should check if `settings.links` exists first, otherwise it will reset the settings on every browser restart.

addons/colorblind/non-bold-links.css
```diff
@@ -0,0 +1,25 @@
...
+a.black:hover,
+a.black:active,
+a.reply,
+input.link,
+input.link:hover,
+input.link:active,
+input.link.black:hover,
+input.link.black:active,
+.button > span,
+button,
```
mxmou, line 16: I don't think this addon should affect buttons.

mxmou (changes requested):

mxmou (approved):

addons/colorblind/addon.json
```diff
@@ -7,12 +7,6 @@
       "link": "https://github.com/DNin01"
     }
   ],
-  "info": [
-    {
-      "id": "feedback",
-      "text": "We want to expand this addon with other accessibility features in the future. You can send us your feedback by going to scratchaddons.com/feedback or clicking \"Send Feedback\" on the sidebar."
```
DNin01, line 13: Yeah, I'm pretty happy with how the addon is doing as just a link style customization. I could see other color blindness accommodations finding their home in new addons.

DNin01 (approved):
Nicely done.

addons/addons.json
```diff
@@ -162,6 +161,7 @@
   "material-forum-editor-buttons",
   "forum-quote-code-beautifier",
   "horizontal-mystuff-tabs",
+  "colorblind",
```
DNin01, line 164: Does moving this addon down here have any negative consequences?

Samq64: I think it was for something on a page with `scratchr2` enabled but moving it down also broke some other styles, so I switched to just using `!important`.

WorldLanguages:
This seems OK to merge right?

Issue #817, opened 2020-11-22
More blocks addon


wgyt735yt:
**Describe the new addon**

This addon adds custom blocks and hacked blocks to the Scratch editors under the Scratch addons category.
This addon could also allow other addons like the pause addon to add their blocks (sa-pause in this example) to this category.
![image](https://user-images.githubusercontent.com/68466727/99908012-0123ff00-2cae-11eb-9525-ee345f72f9ba.png)

apple502j:
I'd say it's rejected - we don't want to encourage Scratchers to install Scratch Addons to run some projects. Rules have changed since NitroBlock.

wgyt735yt:
> I'd say it's rejected - we don't want to encourage Scratchers to install Scratch Addons to run some projects. Rules have changed since NitroBlock.

No, I don't think you understand what I'm saying.
The pause addon takes advantage of a custom block called sa-pause, and that could go in this category. This would also make it easier to use common workarounds for commonly suggested blocks and make them easy to add.

apple502j:
Pause block is a custom block, and does not require new extension category, and it's only for debugging purposes.

wgyt735yt:
It only uses vanilla blocks, and it only makes it easy to drag in workarounds like the ones listed here https://scratch.mit.edu/discuss/topic/444061/

JeffaloBob:
> Pause block is a custom block, and does not require new extension category, and it's only for debugging purposes.

debugging purposes for project creators though.

i think this is a good idea

wgyt735yt:
Here is what needs to happen for this to become reality:

- [ ] We need to inject a new category to the sidebar

- [ ] We need a way to add blocks to the sidebar

- [ ] We need a way for other addons to add their own blocks

If anyone has suggestions let me know

wgyt735yt:
Progress on the prototyping side has been made :)
I made the category, the block design and I made a video demonstrating.
Due to GitHub not allowing me to upload the mp4 file, I have the video hosted on Google Drive and it is available [here](https://drive.google.com/file/d/1X2dDcryFL2PbYRx_r1iHSIxliXmXGKTW/view).

Explosion-Scratch:
Just the stuff like while loops, for i in range, etc would be really helpful. No support however for blocks that need scratchaddons to be run, that would attract ST attention and be really bad. Speaking of which, should the ```SA-pause``` be removed?

TheColaber:
[...]

PR #6676, opened 2023-09-22
New addon: Editor extra blocks

Tacodiva:
Adds a new addon which reveals several blocks present in the Scratch source code, but not accessible via the vanilla editor.

image

Needs someone to write a description (I've given up on writing addon descriptions lmao), but other than that should be ready for review.

### DIFF
--- a/addons/editor-extra-blocks/addon.json
+++ b/addons/editor-extra-blocks/addon.json
@@ -1,0 +1,34 @@
+{
+  "name": "Reveal hidden editor blocks",
+  "description": "TODO",
+  "tags": ["editor", "codeEditor"],
+  "credits": [
+    {
+      "name": "Tacodiva",
+      "link": "https://github.com/Tacodiva"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Show Counter blocks",
+      "id": "counterBlocks",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "name": "Show Useless blocks",
+      "id": "uselessBlocks",
+      "type": "boolean",
+      "default": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.36.0"
+}
--- a/addons/editor-extra-blocks/userscript.js
+++ b/addons/editor-extra-blocks/userscript.js
@@ -1,0 +1,219 @@
+export default async function ({ addon, msg }) {
+  if (!addon.tab.redux.state) return console.warn("Redux is not available!");
+  addon.tab.redux.initialize();
+
+  const UPDATE_TOOLBOX_ACTION = "scratch-gui/toolbox/UPDATE_TOOLBOX";
+
+  const xmlParser = new DOMParser();
+  const xmlSerializer = new XMLSerializer();
+
+  const vm = addon.tab.traps.vm;
+
+  function encodeXML(string) {
+    return string
+      .replace(/&/g, "&amp;")
+      .replace(/</g, "&lt;")
+      .replace(/>/g, "&gt;")
+      .replace(/"/g, "&quot;")
+      .replace(/'/g, "&apos;");
+  }
+
+  const reduxStateListener = async (e) => {
+    if (e.detail.action.type === UPDATE_TOOLBOX_ACTION && !e.detail.action.saExtraBlocks) {
+      const toolboxXML = xmlParser.parseFromString(e.detail.action.toolboxXML, "text/xml");
+
+      const insertAfter = (referenceBlockOpcode, seperator, ...newBlocks) => {
+        const referenceBlock = toolboxXML.querySelectorAll(`[type="${referenceBlockOpcode}"`)[0];
+
+        const newBlockXMLs = [];
+        if (seperator) {
+          const seperatorXML = toolboxXML.createElement("sep");
+          seperatorXML.setAttribute("gap", 36);
+          newBlockXMLs.push(seperatorXML);
+        }
+        for (const newBlock of newBlocks) {
+          const newBlockXML = toolboxXML.createElement("block");
+          newBlockXML.setAttribute("type", newBlock.type);
+          if (newBlock.innerHTML) newBlockXML.innerHTML = newBlock.innerHTML;
+          newBlockXMLs.push(newBlockXML);
+        }
+
+        referenceBlock.after(...newBlockXMLs);
+      };
+
+      let defaultVariableName = "my variable";
+      if (vm.editingTarget) {
+        let variableIDs = Object.keys(vm.editingTarget.variables);
+        if (variableIDs.length !== 0) {
+          defaultVariableName = vm.editingTarget.variables[variableIDs[0]].name;
+        } else {
+          const stage = vm.runtime.getTargetForStage();
+          if (stage) {
+            variableIDs = Object.keys(stage.variables);
+            if (variableIDs.length !== 0) {
+              defaultVariableName = stage.variables[variableIDs[0]].name;
+            }
+          }
+        }
+      }
+
+      insertAfter(
+        "control_repeat_until",
+        false,
+
+        { type: "control_while" },
+        {
+          type: "control_for_each",
+          innerHTML: `
+          <field name="VARIABLE">${encodeXML(defaultVariableName)}</field>
+          <value name="VALUE">
+              <shadow type="math_whole_number">
+                  <field name="NUM">10</field>
+              </shadow>
+          </value>`,
+        }
+      );
+
+      insertAfter(
+        "event_broadcastandwait",
+        true,
+
+        {
+          type: "event_whentouchingobject",
+          innerHTML: `
+        <value name="TOUCHINGOBJECTMENU">
+          <shadow type="event_touchingobjectmenu">
+          </shadow>
+        </value>`,
+        }
+      );
+
+      insertAfter("sensing_loudness", false, { type: "sensing_loud" });
+
+      if (addon.settings.get("counterBlocks")) {
+        insertAfter(
+          "control_delete_this_clone",
+          true,
+
+          { type: "control_get_counter" },
+          { type: "control_incr_counter" },
+          { type: "control_clear_counter" }
+        );
+      }
+
+      if (addon.settings.get("uselessBlocks")) {
+        insertAfter(
+          "motion_sety",
+          true,
+
+          { type: "motion_align_scene" }
+        );
+
+        insertAfter(
+          "motion_sety",
+          true,
+
+          {
+            type: "motion_scroll_right",
+            innerHTML: `
+            <value name="DISTANCE">
+              <shadow type="math_number">
+                  <field name="NUM">10</field>
+              </shadow>
+            </value>`,
+          },
+          {
+            type: "motion_scroll_up",
+            innerHTML: `
+            <value name="DISTANCE">
+              <shadow type="math_number">
+                  <field name="NUM">10</field>
+              </shadow>
+            </value>`,
+          }
+        );
+
+        insertAfter(
+          "motion_direction",
+          true,
+
+          { type: "motion_xscroll" },
+          { type: "motion_yscroll" }
+        );
+
+        insertAfter(
+          "looks_hide",
+          false,
+
+          { type: "looks_hideallsprites" }
+        );
+
+        insertAfter(
+          "looks_setsizeto",
+          true,
+
+          {
+            type: "looks_setstretchto",
+            innerHTML: `
+            <value name="STRETCH">
+              <shadow type="math_number">
+                  <field name="NUM">100</field>
+              </shadow>
+            </value>`,
+          },
+          {
+            type: "looks_changestretchby",
+            innerHTML: `
+            <value name="CHANGE">
+              <shadow type="math_number">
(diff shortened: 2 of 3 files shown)


RedGuy12:
resolves #817

Secret-chest:
please only enable the working ones

Tacodiva:
> please only enable the working ones
@Secret-chest 
I'm not sure what you mean? Do you mean this addon shouldn't include the non-working blocks at all? I figured hiding them behind a setting was a good compromise, and some of the "non-working" blocks do have (admittedly very obscure) use-cases, like "all at once { delete this clone }".

Secret-chest:
I didn't see the setting

addons/editor-extra-blocks/addon.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "Reveal hidden editor blocks",
+  "description": "TODO",
```
Samq64, line 3:
```suggestion
  "description": "Adds some blocks from previous versions of Scratch that were later removed from the block palette. Not all of them are still functional.",
```

addons/editor-extra-blocks/addon.json
```diff
@@ -0,0 +1,34 @@
...
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "settings": [
+    {
+      "name": "Show Counter blocks",
```
Samq64, line 19: Why do the counter blocks get its own setting? Is it because they're not very useful compared to the loops and when touching ones?

addons/editor-extra-blocks/addon.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "Reveal hidden editor blocks",
```
Samq64, line 2:
```suggestion
  "name": "Show hidden editor blocks",
```
Maybe even just Hidden editor blocks? The name is also fine the way it is.

DNin01 (commented):
I'm pretty sure that would appear to violate Scratch's TOU:

> 4.4 You may only submit user-generated projects that were created with (1) the Scratch website editor or (2) an unmodified copy of the Scratch editor compiled from the source code described in Section 5.3. You may not upload any projects that were created, by you or by anyone else, with a modified version of the Scratch editor.

Although it's up for debate whether that section is enforced, I don't think we want to look like we're violating it by allowing access to the "lost" blocks. Many Scratchers would wonder where we got them from and most of these blocks have workarounds that people will be willing to take, potentially leading to actions from the Scratch Team.

DNin01:
I would suggest that we discuss this idea and any potential pitfalls before reviewing the pull request.

CST1229:
> Although it's up for debate whether that section is enforced, I don't think we want to look like we're violating it by allowing access to the "lost" blocks. Many Scratchers would wonder where we got them from and most of these blocks have workarounds that people will be willing to take, potentially leading to actions from the Scratch Team.

The "Extra key options" addon already allows accessing blocks not accessible in vanilla Scratch.

DNin01:
> > Although it's up for debate whether that section is enforced, I don't think we want to look like we're violating it by allowing access to the "lost" blocks. Many Scratchers would wonder where we got them from and most of these blocks have workarounds that people will be willing to take, potentially leading to actions from the Scratch Team.
> 
> The "Extra key options" addon already allows accessing blocks not accessible in vanilla Scratch.

Right. 🤔

There is some gray area here.

Tacodiva:
> I'm pretty sure that would appear to violate Scratch's TOU

I think it's also important to add that Scratch Add-ons is a modification to the Scratch editor, so is against TOS by default.

These blocks can be used by non-sa users by backpacking them from another project, so this addon doesn't allow you to do anything that coukd not be done with the vanilla editor.

addons/editor-extra-blocks/addon.json
Tacodiva: Yeah, it was my opinion that they where kinda useless compared to the other blocks and there's a few of them so I thought it best to put them in their own setting.

WorldLanguages:
We should be really careful when adding blocks that do nothing, as it's not obvious which ones do just by looking at the blocks.

medlans:
Only the working ones should be included honestly.
> I'm pretty sure that would appear to violate Scratch's TOU:
> 
> > 4.4 You may only submit user-generated projects that were created with (1) the Scratch website editor or (2) an unmodified copy of the Scratch editor compiled from the source code described in Section 5.3. You may not upload any projects that were created, by you or by anyone else, with a modified version of the Scratch editor.
> 
> Although it's up for debate whether that section is enforced, I don't think we want to look like we're violating it by allowing access to the "lost" blocks. Many Scratchers would wonder where we got them from and most of these blocks have workarounds that people will be willing to take, potentially leading to actions from the Scratch Team.

That part of the TOU has confused me because it sounds like it blocks any of the 2.0 editors and the 3.0 offline editor though.

DNin01:
> Only the working ones should be included honestly.

Yeah, I think so too.

Samq64:
If the non-functional blocks are added there should be a notice about them so it stands out more than the last part of the description.

Tacodiva:
I agree there should be a notice about them. Would it be possible / a good idea to somehow link to the Scratch wiki page [here](https://en.scratch-wiki.info/wiki/Hidden_Blocks#List_of_hidden_blocks) which has a list of the hidden blocks and what they do (including weather their useless or not).

WorldLanguages:
Maybe the useless blocks could be a separate, easter-egg addon?

Secret-chest:
> Maybe the useless blocks could be a separate, easter-egg addon?

Would be fun. We need more easter eggs.

WorldLanguages:
> Maybe the useless blocks could be a separate, easter-egg addon?

@DNin01 Why the downvote?

DNin01:
> > Maybe the useless blocks could be a separate, easter-egg addon?
> 
> 
> 
> @DNin01 Why the downvote?

It's hard to justify that it's a good easter egg choice when it is quite literally hacking blocks in. While providing access to any useful hidden/lost blocks would still be considered hacking, I would personally try to minimize the extent to which we hack things in.

WorldLanguages:
> it is quite literally hacking blocks in

I was 10 years old and scratchers still obtained these blocks from the backpack back then though...

addons/editor-extra-blocks/addon.json
KingKittyTurnip: Kinda New To GitHub, So I'm Not Sure If I Should Say Stuff A Month Late, But I Will Say That Counter Blocks Have A Specifice Use Case. They Reset To 0 ANY-time A project Is Loaded, Reloaded, Remixed. I Use These In Every Project For A One-Time Intro That Changes If The Project Is Remixed. Somthing Not Possible With Variables. Yes, Specific. The Value Is The Only Value Not Saved In The Json.

DNin01: Yeah, I've seen it used that way before.

WorldLanguages: Remix detector? 💀

KingKittyTurnip: Yup. I Also Learned It's possible To Do The Same Thing With A Complicated Username Setup, But The Counter Blocks make it more Convenient... So IDK How These Blocks Should be handled For Scratch Addons.

KingKittyTurnip:
"x scroll", "y scroll", and "user id" Blocks Can Crash Scratch If Put Into A List That's Shown On The Stage.

addons/editor-extra-blocks/addon.json
BroJac5246: > Yup. I Also Learned It's possible To Do The Same Thing With A Complicated Username Setup, But The Counter Blocks make it more Convenient & Reliable... So IDK How These Blocks Should be handled For Scratch Addons.
> 
> It's More Like A On Project Start Detector, But Can be Used For Detecting Remixes.

But what would you do with a remix detector (other than blocking access to your project?)

KingKittyTurnip: Sense We Arn't Really Allowed To Block Access, I Use it As A Extra Failsafe in Case If Somewon Uploads A Project & It Doesn't Say The Remix, But Should be.

BroJac5246: > Sense We Arn't Really Allowed To Block Access, I Use it As A Extra Failsafe in Case If Somewon Uploads A Project & It Doesn't Say The Remix, But Should be.

That's a little extreme, though... is that really a reason for Scratch Addons to support counter blocks?
> One Thing I'm Tring To Figure Out Is The Most Effective Way Of Detecting Wether They broke The Failsafe, & Most Effective Contraption. At That Point It Would Block Access To The Project.

That's too complicated, probably not realistically possible, and risks violating Scratch's TOS.

(Also, might I suggest that you not capitalize every word you write?)

KingKittyTurnip: I guess it kinda would violate, but the remixer would be too.

IDK how Scratch Addons should handle these blocks. Even though I really want easier access to these blocks.

Counter blocks don't easily have a readying monitor, so they kinda also work as a "hidden" variable.

KingKittyTurnip:
This Should Include The Seceret Extention Blocks & Menus If This Would Be Accepted, (Including the extention menus).

addons/editor-extra-blocks/addon.json
BroJac5246: > I guess it kinda would violate, but the remixer would be too.

To cite the adage: two wrongs don't make a right
> IDK how Scratch Addons should handle these blocks. Even though I really want easier access to these blocks.

Yeah. It's been somewhat contentious, but I generally support adding them.
> Counter blocks don't easily have a readying monitor, so they kinda also work as a "hidden" variable.

That's their main advantage. Sometimes I'll use them, but I always ask how it's better then a variable 🤷

KingKittyTurnip: It would only really actually violate if the remixer violates. But yes thats true.

I Still Feel Like They Don't Need A Seprate Toggle, like the while block just saves 1 not, and the "loud?" block replaces "loudness > 10", while the counter blocks give new mechanics. As of now the hidden blocks just fulfill a niche, not nessecarly be great. the "for each [] in ()" block is probably the 1 exception.

Tacodiva: @KingKittyTurnip @BroJac5246 This conversation seems to be getting a bit sidetracked. I chose to hide the counter blocks by default because, unlike the other blocks, they don't have an 'intuitive' use. It's obvious what the 'while' block does, or the 'for each' or even '<loud ?>'. These have usecases that the average Scratcher can likely grasp very easily. But on a surface level, the counter is just a more restrictive variable with no real utility. Additionally, there are 3 different counter blocks which all take up valuable space in the 'control' category. 

I have the same opinion about the `all at once` block. It can be used for `all at once { delete this clone }` to continue script execution for the original sprite but not the clones, yet I don't think this use is obvious so it too is hidden by default.

All this being said, it's really not that hard to flip a switch in the settings menu. Unless someone else has something to add to this discussion, I stand by my decision.

addons/editor-extra-blocks/userscript.js
```diff
@@ -0,0 +1,219 @@
...
+        type: UPDATE_TOOLBOX_ACTION,
+        toolboxXML: xmlSerializer.serializeToString(toolboxXML),
+        saExtraBlocks: true,
+      });
+    }
+  };
+
+  const updateToolbox = () => {
+    if (vm.editingTarget) {
+      vm.emitWorkspaceUpdate();
```
pufferfish101007, line 200: This method clears the undo stack, which isn't an ideal side-effect of changon addon settings or enabling/disabling the addon.

pufferfish101007 (changes requested):

addons/editor-extra-blocks/userscript.js
mxmou: Related: https://github.com/ScratchAddons/ScratchAddons/pull/6727#issuecomment-1752023575

Joeclinton1:
The foreach block is very useful due to it's ability to simplify code and make projects slighty faster. It is a pain to keep getting from the backpack, and I would very much appreciate if this addon was merged just for the foreach block.

The users clearly want it too. It clearly doesn't violate TOS since you can get these from the backpack and if scratch cared enough they would make them unpackbackable. The blocks were only removed to reduce the number of blocks thus simplifing learning. It's not like they're dangerous.

addons/editor-extra-blocks/userscript.js
mxmou: [`updateToolboxXML()`](https://github.com/ScratchAddons/ScratchAddons/blob/9773b46/libraries/common/cs/update-toolbox-xml.js) now exists.

DNin01:
Yeah, this addon is fine. It probably shouldn't be recommended or featured, but it's fine to have. Not having it isn't going to stop the hundreds of creators who are already using these hidden blocks.

Hidden blocks that aren't functional or useful enough could be excluded from this addon.

DNin01:
> The foreach block is very useful due to it's ability to simplify code and make projects slighty faster. It is a pain to keep getting from the backpack, and I would very much appreciate if this addon was merged just for the foreach block.

This block even used to be shown on the block palette in TurboWarp, in addition to the `while` loop, which is still shown.

<sup>[Why it was removed](https://github.com/TurboWarp/scratch-gui/issues/769)</sup>

WorldLanguages:
Yes, it is fine to allow easier access to things that can already be achieved with the backpack.

PR #6684, opened 2023-09-23
Fix share confirm modal appearing on ideas button

### LINKED ISSUE #6683: Ideas page button triggers share conformation
### Description

When the "Choose a tutorial" button is pressed on [the ideas page](https://scratch.mit.edu/ideas), the "Are you sure you want to share?" popup appears.

### Steps to reproduce

1. Turn on the Conformations addon
2. Open https://scratch.mit.edu/ideas
3. Click on the "Choose a tutorial" button

_No response_

### Environment

Browser: Firefox 117
Operating system: macOS Ventura

mybearworld:
Resolves #6683

### Changes

Make the `.banner-button` selector more specific by also requiring a `.banner-text` next to it, which the ideas button doesn't have.

### Reason for changes

To fix the share modal appearing on the ideas button.

### Tests

Tested on Edge 116.

### DIFF
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -13,7 +13,7 @@ export default async function ({ addon, console, msg }) {
       let cancelMessage = null;
       if (
         addon.settings.get("projectsharing") &&
-        e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-button")
+        e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-text + .banner-button")
       ) {
         title = addon.tab.scratchMessage("project.share.shareButton"); // "Share"
         cancelMessage = msg("share");
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -13,7 +13,9 @@ export default async function ({ addon, console, msg }) {
       let cancelMessage = null;
       if (
         addon.settings.get("projectsharing") &&
-        e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-text + .banner-button")
+        e.target.closest(
+          "[class*='share-button_share-button']:not([class*='is-shared']), .banner-text + .banner-button"
+        )
       ) {
         title = addon.tab.scratchMessage("project.share.shareButton"); // "Share"
         cancelMessage = msg("share");


WorldLanguages:
> Make the `.banner-button` selector more specific by also requiring a `.banner-text` next to it, which the ideas button doesn't have.

Maybe we can change it to something even more specific to the Share button? Is there anything?

mxmou:
> > Make the `.banner-button` selector more specific by also requiring a `.banner-text` next to it, which the ideas button doesn't have.
> 
> Maybe we can change it to something even more specific to the Share button? Is there anything?

`.preview .banner-button` would match a `.banner-button` on the project page, but there's no "share banner" or "share button" class.

mybearworld:
> > Make the `.banner-button` selector more specific by also requiring a `.banner-text` next to it, which the ideas button doesn't have.
> 
> Maybe we can change it to something even more specific to the Share button? Is there anything?



> > Make the `.banner-button` selector more specific by also requiring a `.banner-text` next to it, which the ideas button doesn't have.
> 
> Maybe we can change it to something even more specific to the Share button? Is there anything?

It's very generic.

```html
<div class="banner-outer">
  <div class="flex-row inner banner-inner">
    <span class="banner-text">
      <span>
        This project is not shared — so only you can see it. Click share to let
        everyone see it!
      </span>
    </span>
    <button class="button banner-button"><span>Share</span></button>
  </div>
</div>
```

Something with `:has` would probably work, but that doesn't have good enough browser support.

MaterArc:
It seems scratch got a little lazy in naming their elements...

DNin01:
> It seems scratch got a little lazy in naming their elements...

They haven't really. The Scratch source code has more access to itself than a browser extension and knows variables the extension doesn't. A separate, unrelated module like our addon's script has to work with limited information (especially since Scratch uses things like [React](https://github.com/facebook/react)) and therefore may have trouble distinguishing a certain button from others. This is why updates to the editor can leave added buttons in the wrong place.

The parent elements provide more contextual information - selecting `.preview .banner-button` as pointed out by @mxmou should solve the problem. That or checking the URL.

WorldLanguages:
Adding `body:not(.sa-body-editor)` to the beginning of the selector would be pretty much equivalent to checking that the URL is a project page.

mybearworld:
> Adding `body:not(.sa-body-editor)` to the beginning of the selector would be pretty much equivalent to checking that the URL is a project page.

This seems to only get added when you're in the editor.

WorldLanguages:
> Adding `body:not(.sa-body-editor)` to the beginning of the selector would be pretty much equivalent to checking that the URL is a project page.

Response to myself: I don't think it is very useful in this case, though.

WorldLanguages (approved):
Works!

Samq64 (approved):

addons/confirm-actions/userscript.js
```diff
@@ -13,7 +13,9 @@ export default async function ({ addon, console, msg }) {
       let cancelMessage = null;
       if (
         addon.settings.get("projectsharing") &&
-        e.target.closest("[class*='share-button_share-button']:not([class*='is-shared']), .banner-button")
+        e.target.closest(
+          "[class*='share-button_share-button']:not([class*='is-shared']), .banner-text + .banner-button"
```
DNin01, line 17: I would have also selected `.preview` but this is good too.

DNin01 (approved):

PR #6687, opened 2023-09-23
Make `forum-live-preview` affected by toolbar buttons

### LINKED ISSUE #6685: `forum-live-preview` doesn't trigger on toolbar button click
### Description

When clicking on the bold button, one of the scratch blocks, etc., the live preview doesn't trigger.

### Steps to reproduce

1. Go to an open topic / create a new one.
2. Click on one of the toolbar icons.
3. See nothing happen

_No response_

### Environment

Browser: Edge 116
Operating system: Windows

mybearworld:
Resolves #6685

### Changes

Add a click to a toolbar button to the actions that trigger the event to show the preview.

### Reason for changes

To make the experience make more sense.

### Tests

Tested in Edge 116.

### DIFF
--- a/addons/forum-live-preview/userscript.js
+++ b/addons/forum-live-preview/userscript.js
@@ -40,7 +40,11 @@ export default async function ({ addon, console }) {
 
   textarea.addEventListener("input", delayedPreview);
   markItUpButtons.forEach((el) => {
-    el.addEventListener("click", delayedPreview);
+    el.addEventListener("click", () => {
+      if (textarea.value) {
+        delayedPreview();
+      }
+    });
   });
   addon.self.addEventListener("disabled", () => {
     showPreview();
--- a/addons/forum-live-preview/userscript.js
+++ b/addons/forum-live-preview/userscript.js
@@ -1,6 +1,7 @@
 export default async function ({ addon, console }) {
   const textarea = await addon.tab.waitForElement(".markItUpEditor");
   const previewButton = await addon.tab.waitForElement(".markItUpButton.preview");
+  const markItUpButtons = document.querySelectorAll(".markItUpButton:not(.markItUpDropMenu)");
   let previewIframe;
   let delay;
   switch (addon.settings.get("rate")) {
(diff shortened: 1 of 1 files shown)


Samq64 (approved):

WorldLanguages:
I'm not sure if this way of fixing this makes sense - see my comment https://github.com/ScratchAddons/ScratchAddons/issues/6685#issuecomment-1738360738

mybearworld:
> I'm not sure if this way of fixing this makes sense - see my comment [#6685 (comment)](https://github.com/ScratchAddons/ScratchAddons/issues/6685#issuecomment-1738360738)

Huh, good catch.

mybearworld:
Ok, so this seems to happen because vanilla Scratch updates the preview on its own when a forum toolbar button is clicked.

mybearworld:
The situation in this fix then is that the listeners on the `markItUpButtons` are completely redundant, *except* when the preview is currently hidden.

addons/forum-live-preview/userscript.js
```diff
@@ -1,6 +1,7 @@
 export default async function ({ addon, console }) {
   const textarea = await addon.tab.waitForElement(".markItUpEditor");
   const previewButton = await addon.tab.waitForElement(".markItUpButton.preview");
+  const markItUpButtons = document.querySelectorAll(".markItUpButton:not(.markItUpDropMenu)");
```
WorldLanguages, line 4: No guarantee the element exists yet

WorldLanguages: Actually, nevermind, the preview button is the last one and the line above waits until it exists
Not sure how this would play with other addons that add their own buttons to the toolbar though

mybearworld: > Not sure how this would play with other addons that add their own buttons to the toolbar though

True, an event listener on body that checks the `classList` would probably be better here

mybearworld: Never mind, Scratch seems to call `e.stopPropagation()` or something to that effect when a button is clicked

PR #6688, opened 2023-09-23
`editor-theme3`: Style define blocks correctly

### LINKED ISSUE #6686: `editor-theme3` doesn't change color of define blocks correctly in the forums
### Description

The top line of blocks changed by `editor-theme3` in the forums always will have the regular color.

### Steps to reproduce

1. Enable `editor-theme3`.
2. Change the color of the define block.
3. Enable "Change color of blocks in the forums".
4. Go to a forum page with a define block (https://scratch.mit.edu/discuss/post/7529163/)

A dark mode screenshot of a forum post with a scratch define block in it that has its color changed by editor-theme3, but not the color of the upper line.

### Environment

Browser: Edge 116
Operating system: Windows

mybearworld:
Resolves #6686

### Changes

Style the `.sb-define-hat-cap` class.

A define scratch block properly styled with `editor-theme3`.

### Reason for changes

To make `editor-theme3` affect define blocks.

### Tests

Tested in Edge 116.

### DIFF
--- a/addons/editor-theme3/forums/base.css
+++ b/addons/editor-theme3/forums/base.css
@@ -13,6 +13,10 @@
 .sb-pen {
   fill: var(--sa-block-color);
 }
+.sb-define-hat-cap {
+  fill: var(--sa-block-color);
+  stroke: var(--sa-block-color);
+}
 .sb3-motion,
 .sb3-looks,
 .sb3-sound,
@@ -137,7 +141,8 @@
 .sb-custom-arg ~ .sb-label,
 .sb3-custom-arg,
 .sb3-custom-arg-dark,
-.sb3-custom-arg ~ .sb3-label {
+.sb3-custom-arg ~ .sb3-label,
+.sb-define-hat-cap {
   --sa-block-color: var(--editorTheme3-customColor);
 }
 .sb-extension,


mybearworld:
> This fixes the issue where the cap isn't affected. However, there are still some existing issues: the text is still white when text color is set to Black, and with the settings "Colored on white/black background", it should have a white/black cap instead of a colored one. We might want to fix those while we're at it.

That makes sense.

mybearworld:
The color of the text inside of the block outline can be easily selected with `.sb-outline ~ .sb-label` - the "define" text is more tricky, though.

Samq64 (approved):

addons/editor-theme3/forums/base.css
```diff
@@ -137,7 +141,11 @@
 .sb-custom-arg ~ .sb-label,
 .sb3-custom-arg,
 .sb3-custom-arg-dark,
-.sb3-custom-arg ~ .sb3-label {
+.sb3-custom-arg ~ .sb3-label,
+.sb-outline,
+.sb-outline ~ .sb-label,
+.sb-label:only-of-type:not(:last-child),
```
mxmou, line 147: This also matches the labels of some other blocks:
image
(blocks from [this post](https://scratch.mit.edu/discuss/post/7535381/))

mxmou (changes requested):

addons/editor-theme3/forums/base.css
mybearworld: Fixed - looks like this now, which is still incorrect, but that's the same behavior as SA has without this PR

image

Samq64:
The forums now use Scratch 3 blocks (see https://github.com/ScratchAddons/ScratchAddons/issues/7910).

PR #6690, opened 2023-09-23
Create "Button to hide monitors" addon

### LINKED ISSUE #6530: Option to hide monitors on projects
### Why this would be helpful

Oftentimes, monitors can get in the way of things in projects that are not well thought through. A feature that hides adds a button that allows temporary hiding of all monitorswould be helpful.

### How this addon works

This addon adds a button to the top bar (e.g. where the green flag, stop sign, etc are) that toggles the visibility state of monitors.

### Possible alternatives

_No response_

### Additional context

_No response_

mybearworld:
Resolves #6530

### Changes

Add a "Button to hide monitors" addon that does what its name says.

Button to hide monitors

### Reason for changes

Oftentimes, monitors can get in the way of things in projects that are not well thought through. A feature that hides adds a button that allows temporary hiding of all monitorswould be helpful.

### Tests

Tested in Edge

### To Do

- [ ] **Use an icon, and not text for this.** I'm not sure what a good icon for this would be.
- [x] **You can still interact with these when they're invisible.** There are likely solutions to this, though.


addons/hide-monitors/addon.json
```diff
@@ -0,0 +1,30 @@
...
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["projects"]
+    }
+  ],
+  "info": [
+    {
+      "id": "no-unintended",
+      "text": "You are not able to click through the places where the monitors would normally be. This is done so this addon can't be used to cheat in games, or using projects in other unintended ways."
```
Samq64, line 20: Do you have an example of a project where you can cheat if the monitors are gone but not if they're just invisible?

addons/hide-monitors/userstyle.css
```diff
@@ -0,0 +1,3 @@
+.sa-hide-monitors {
+  opacity: 0;
```
DNin01, line 2: I wonder if they should be just barely visible so you can tell where they are if you're trying to click and it's not working.

addons/hide-monitors/addon.json
mybearworld: Not really - would this wording be better?
```suggestion
      "text": "You are not able to click through the places where the monitors would normally be. This is done so you can't use the project in unexpected ways, like clicking on things you aren't supposed to."
```

addons/hide-monitors/userstyle.css
mybearworld: Maybe it could be a three state toggle? `Monitors shown` -> `Monitors barely shown` -> `Monitors hidden`?

DNin01: Whether they were hidden or barely visible could also be set in the addon's settings.

WorldLanguages:
Is the button only available in the editor?

mybearworld:
> Is the button only available in the editor?

No, it's available everywhere.

BroJac5246:
The button takes up too much space. But if it's an icon, it won't look like a button.

mybearworld:
> The button takes up too much space. But if it's an icon, it won't look like a button.

I just used that as a placeholder, because I really don't know what icon would be good for this.

BroJac5246:
> > The button takes up too much space. But if it's an icon, it won't look like a button.
> 
> I just used that as a placeholder, because I really don't know what icon would be good for this.

A button, even one with an icon, looks out of place where it is currently located.

mybearworld:
> > > The button takes up too much space. But if it's an icon, it won't look like a button.
> > 
> > 
> > I just used that as a placeholder, because I really don't know what icon would be good for this.
> 
> A button, even one with an icon, looks out of place where it is currently located.

The button would just have its padding, background, border, etc. removed.

BroJac5246:
> > > > The button takes up too much space. But if it's an icon, it won't look like a button.
> > > 
> > > 
> > > I just used that as a placeholder, because I really don't know what icon would be good for this.
> > 
> > A button, even one with an icon, looks out of place where it is currently located.
> 
> The button would just have its padding, background, border, etc. removed.

... at which point it looks like a status icon.

PR #6108, opened 2023-05-26
editor-theme3: adapt to "color mode" settings in menu bar

WorldLanguages:
Mostly resolves #5905

### Changes

If "customizable block colors" is enabled, the menu will look like this:
image

### Tests

Needs testing in Firefox and RTL languages.

PR #6247, opened 2023-06-21
editor-theme3: compatibility with Scratch high contrast theme (attempt 2)

mxmou:
Resolves #5314

### Changes

Same as #6146, the only differences are in this commit: 1ba72a9

### Reason for changes

Compatibility with the Scratch update.

### Tests

Tested locally and on scratch.mit.edu. Switching languages works correctly, #5314 is fixed and the crash described in #6212 doesn't happen.

Issue #6175, opened 2023-06-09
High Cotrast Zoom Controls


Samq64:
### Why this would be helpful

Scratch will soon change the zoom control style when using the high contrast theme:
|Original|High Contrast|
|------------|--------------------|
|![image](https://github.com/ScratchAddons/ScratchAddons/assets/81489795/b48c15e1-2b43-4a29-b469-6e34a6deee6d) |![image](https://github.com/ScratchAddons/ScratchAddons/assets/81489795/a8b72ce9-eb38-4aad-9c68-82524b39fe8d)|

The problem is the customizable block colors addon disables Scratch's colour mode menu, so the high contrast zoom controls can't be enabled when using that addon.

### My suggestion

Adding an option to the customizable code area zoom addon that enables the high contrast zoom controls so they can be used with customizable block colors enabled.

### Possible alternatives

Making it a new addon.

### Additional context

This is a pretty minor thing, so maybe it's not worth adding.

mxmou:
I think editor dark mode would be a better place for this setting.

WorldLanguages:
> I think editor dark mode would be a better place for this setting.

Maybe "customizable code area zoom" can be responsible for changing the SVG, and editor dark mode should only change the color? Or were you referring to something completely different?

WorldLanguages:
> The problem is the customizable block colors addon disables Scratch's colour mode menu, so the high contrast zoom controls can't be enabled when using that addon.

Right now it's pretty bad, the zoom controls look different depending on which theme Scratch thinks is selected, and the only way the user can change it is by disabling "customizable block colors", changing the Scratch theme, and then reenabling it back.

WorldLanguages:
> Maybe "customizable code area zoom" can be responsible for changing the SVG, and editor dark mode should only change the color?

To clarify, we don't currently provide a way to customize the exact colors of the icons, but editor dark mode sometimes applies `invert(1) hue-rotate(180deg)` to them.

DNin01:
> The problem is the customizable block colors addon disables Scratch's colour mode menu, so the high contrast zoom controls can't be enabled when using that addon.

Another option instead of graying out the options on the color mode selector is simply adding a note saying, "colors may be overridden by Scratch Addons", so users would be able to change the color mode if necessary while still being informed that some colors may still be overridden.

WorldLanguages:
[...]

PR #6694, opened 2023-09-24
editor-theme3: monitor colors

### LINKED ISSUE #4033: Customizable block colors: affect monitors

The "customizable block colors" addon doesn't change the colors of stage monitors.

Modify the addon so that it changes the appearance of stage monitors.

mxmou:
Resolves #4033

### Changes

* Adds a setting to the customizable block colors addon that changes the colors of stage monitors.
image
* Fixes the bug described in https://github.com/ScratchAddons/ScratchAddons/pull/6108#issuecomment-1572329064

### Reason for changes

Monitors normally have the same color as blocks.

### Tests

Tested on Edge and Firefox.


medlans:
Would this allow users to change like the outside part? I mean like this:
MONITORS
Yes, this isn't 2.0. Though I wasn't sure if many wanted this.

mxmou:
> Would this allow users to change like the outside part to something custom?

That could be a part of #3912.

addons/editor-theme3/addon.json
```diff
@@ -328,6 +328,12 @@
         }
       ]
     },
+    {
+      "name": "Change the colors of variables and lists on the stage",
+      "id": "monitors",
+      "type": "boolean",
+      "default": false
```
WorldLanguages, line 335: Any particular reason to default to false?

WorldLanguages:
Something to consider is that, if this setting is off, Scratch will color monitors based on which color mode is selected - original or high-contrast - but this addon intentionally disables the color mode selector in the editor menu bar.
It's not very nice to have a "non-deterministic" UI depending on the values of settings that cannot be changed while the addon is enabled. It only depends on which mode was selected the last time the addon was disabled. This problem can also be seen with the zoom controls and we discussed that on #6175

WorldLanguages:
Regarding my previous comment, I think we could consider adding the following behavior:
If Scratch thinks it's in high-contrast mode and the setting is disabled, the `updateMonitorColors()` function should still run, but it would set monitors to the vanilla (original) colors, not the user-selected colors.
What do you think?

Another alternative is providing a 3 option select if this new setting is disabled:
Color pallete to use for monitors: `Unspecified`(?) / `Original` / `High contrast`

I also wonder if we should try something similar for the zoom controls.
I don't want to delay this PR due to these edge cases, but we should keep them in mind.

addons/editor-theme3/addon.json
mxmou: The `affectStage` setting in editor dark mode, which also changes monitors, is off by default.

mxmou:
> Regarding my previous comment, I think we could consider adding the following behavior:
> If Scratch thinks it's in high-contrast mode and the setting is disabled, the `updateMonitorColors()` function should still run, but it would set monitors to the vanilla (original) colors, not the user-selected colors. What do you think?

That's probably better than the current behavior.

DNin01:
> Something to consider is that, if this setting is off, Scratch will color monitors based on which color mode is selected - original or high-contrast - but this addon intentionally disables the color mode selector in the editor menu bar.
> It's not very nice to have a "non-deterministic" UI depending on the values of settings that cannot be changed while the addon is enabled. It only depends on which mode was selected the last time the addon was disabled. This problem can also be seen with the zoom controls and we discussed that on #6175

Another thing we could do instead of disabling the options is simply adding a note saying, "colors may be overridden by Scratch Addons".

addons/editor-theme3/addon.json
WorldLanguages: Hmm, I wouldn't expect the "editor dark mode" addon to affect the stage by default, but having this by default seems reasonable enough.

WorldLanguages:
> Another thing we could do instead of disabling the options is simply adding a note saying, "colors may be overridden by Scratch Addons".

Well, at that point I wouldn't call them "color modes"... Figuring out how to make this intuitive is a bit messy. We went with the easiest solution which was disabling the menu and linking to addon settings if `editor-theme3` is enabled.

addons/editor-theme3/addon.json
WorldLanguages, line 335:
```suggestion
      "default": true
```

WorldLanguages: (also needs latestUpdate.newSettings)

mxmou: The editor and block colors are related though. For example, dark mode and dark blocks will usually be used together.

WorldLanguages: @mxmou So you think this setting should stay as disabled by default?

In that case, the high-contrast presets need to set this to true... Of course, we want setting high-contrast mode from our settings page to work the same as enabling it in vanilla Scratch.
IMO, it would be simpler to enable the setting by default though

WorldLanguages: Discussion in code comment above

mxmou: If this setting is enabled by the default, I think the one in editor dark mode should be too.

WorldLanguages: I would agree if Scratch had a native dark mode that also affected the stage

WorldLanguages:
We will merge this for next version!

@mxmou It's reasonable to make this new setting off-by-default, but at least the "high contrast" preset should include it enabled. The intention is to have the "high contrast" preset look exactly like the high contrast mode if the addon was disabled.

Also:

> > Regarding my previous comment, I think we could consider adding the following behavior:
> > If Scratch thinks it's in high-contrast mode and the setting is disabled, the `updateMonitorColors()` function should still run, but it would set monitors to the vanilla (original) colors, not the user-selected colors. What do you think?
> 
> That's probably better than the current behavior.

I can work on implementing this behavior, if you confirm that's what we actually want to do

mxmou:
> @mxmou It's reasonable to make this new setting off-by-default, but at least the "high contrast" preset should include it enabled. The intention is to have the "high contrast" preset look exactly like the high contrast mode if the addon was disabled (exception: zoom controls)

Makes sense.

> > > Regarding my previous comment, I think we could consider adding the following behavior:
> > > If Scratch thinks it's in high-contrast mode and the setting is disabled, the `updateMonitorColors()` function should still run, but it would set monitors to the vanilla (original) colors, not the user-selected colors. What do you think?
> > 
> > 
> > That's probably better than the current behavior.
> 
> I can work on implementing this behavior, if you confirm that's what we actually want to do

I think it is - ideally nothing would depend on Scratch's color mode setting when this addon is enabled.

WorldLanguages:
We're not going to migrate the existing users of the preset, I think.

addons/editor-theme3/theme3.js
```diff
@@ -457,6 +457,52 @@ export default async function ({ addon, console, msg }) {
...
+          value.style.backgroundColor = primaryColor(category);
+          value.style.color = isColoredTextMode() ? tertiaryColor(category) : uncoloredTextColor();
+          // Border color for list items
+          if (textMode() === "colorOnBlack") value.style.borderColor = "rgba(255, 255, 255, 0.15)";
+          else value.style.removeProperty("border-color");
+        } else {
+          value.style.backgroundColor = originalColors[category.colorId].primary;
+          value.style.color = originalColors.text;
+          value.style.removeProperty("border-color");
+        }
```
WorldLanguages, line 490: Regarding this:

> If Scratch thinks it's in high-contrast mode and the setting is disabled, the updateMonitorColors() function should still run, but it would set monitors to the vanilla (original) colors, not the user-selected colors. What do you think?


Looks like the code already does this, but `originalColors` is defined to be the colors used by Scratch on page load (which misses the point). Would it hurt to hardcode these colors to be the **actual** original colors, which I assume is the original intention of this object?
Note: they were previously hardcoded, but [a commit which predates the Scratch theme selector](https://github.com/ScratchAddons/ScratchAddons/commit/b169b4461dd5755ad00bbdf7bfe3e2d81f7bd4b2#diff-58d2bc5ff8a91b2133df32c4c832575a79c3904cdc98cb7a3642158302794909R70) made the code shorter by getting the original colors from the Blockly object instead.

WorldLanguages:
- [ ] Decide whether to attempt to always use "Original" theme colors for monitors if the setting is disabled, in case High-Contrast is the internally selected option of the editor
- [ ] latestUpdate

addons/editor-theme3/addon.json
```diff
@@ -357,7 +363,8 @@
         "sa-color": "#34E4D0",
         "comment-color": "#FEF49C",
         "input-color": "#FFFFFF",
-        "text": "black"
+        "text": "black",
+        "monitors": true
```
mxmou, line 367: Other presets should have `"monitors": false`.

addons/editor-theme3/theme3.js
mxmou: > Also, I think the decision not to hardcode `originalColors` might cause bugs if during the same tab session, the user toggles the addon and changes the Scratch theme, as `originalColors` will no longer match `Blockly.Colours`

This was fixed in PR #6247. [This part](https://github.com/ScratchAddons/ScratchAddons/blob/c967343/addons/editor-theme3/theme3.js#L457-L460) of the code updates `originalColors` when the Scratch theme changes.

> I believe my suggestion is to use `Blockly.Colours` when appropriate (for example, if the addon is disabled) and rename `originalColors` to reflect it's the colors of the "Original" theme, which will be useful for this usecase. We could either hardcode the JSON object, or find a way to get it directly from the page somehow (we should also consider TurboWarp)

This wouldn't work - the addon works by changing `Blockly.Colours`, so it has to change them back when disabled, which is what the `originalColors` object is for. The monitor code could use a separate hardcoded object.

WorldLanguages:
@mxmou I will merge as admin, if you're okay with that.

Samq64 (approved):
This doesn't work on the project page but that's intentional, right?

Samq64 (approved):

WorldLanguages:
Thanks samq64 for reporting the bug

WorldLanguages:
Bug from feedback: #7604

PR #6693, opened 2023-09-24
"Show profile pictures next to usernames" addon

mybearworld:
Resolves #6270

### Changes

Create a "Show profile pictures next to usernames" addon that does what it says.

### Reason for changes

PFPs are in most parts of the website.

### Tests

Tested in Edge.

### To Do

**Make it respond to changes in the DOM.** Currently, it doesn't work with most scratch-www things and also not for any new comments. I have tried doing this with `MutationObserver` and crashed my browser every time.
Also, I have probably missed some selectors, which means profile pictures there appear twice.

Issue #6139, opened 2023-06-01
Require waitForElement optimizations for userscripts that run in project pages


WorldLanguages:
Setting this issue as a reminder to research this and think of ideas.
Most addons apply these optimizations but I will check if some of them don't.

WorldLanguages:
We should also consider a re-thought of the waitForElement options API, which might be confusing for new contributors. Possibly `{markAsSeen:true}` should be the default?

WorldLanguages:
**Some suggestions:**

- Remove the `markAsSeen` option, default it to true, and encourage the use of `addon.tab.waitForElementOnce` (doesn't exist yet) instead.
- Avoid reusing the same set of `reduxEvents` each time. Give a name to each set and expose them as `addon.tab.waitForElement.EDITOR_DOM_INITIALIZED`, for example.

Regarding project pages:
- Each `waitForElement` call should indicate whether it plans to be executed in a project page (indicating true or false would be mandatory).
- If above is true, then each call should indicate if it is intended for the project page, the project editor, or both. This replaces `isPlayerOnly` optimizations.
- Optimizations such as "currently in the code/costume/sound " (`editorTab.activeTabIndex`) will continue to look the same, but maybe they could also be made more expressive. We should consider this, as there's 14 files that check if the tab index matches exactly a number.

WorldLanguages:
Related comment: https://github.com/ScratchAddons/ScratchAddons/pull/6695#issuecomment-1734230346

WorldLanguages:
Related comment 2: https://github.com/ScratchAddons/ScratchAddons/issues/6880#issuecomment-1823304648

WorldLanguages:
Related comment 3: https://github.com/ScratchAddons/ScratchAddons/issues/6956#issuecomment-1858406128

WorldLanguages:
I'm thinking, maybe we could consider the editor and project pages completely separate concepts - and make it so that every userscript that wants to run on `/projects/*` needs to explicitly ask for it in the addon manifest. The "projects" match would cease to exist, and no amount of asterisks would be enough to run in project pages.

WorldLanguages:
Also related:
If a waitForElement call includes `reduxEvents`, it should still run the selector for the first time, right? Looks like that is not the case, according to this addon: https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/disable-auto-save/userscript.js#L10

mxmou:
[...]

PR #6695, opened 2023-09-24
"Show profile pictures next to usernames" addon

### LINKED ISSUE #6270: pfp (profile picture) to username
### Why this would be helpful

As you know we can not see pfp on pretty much any username on scratch, except on messages, What's Happening? and Explore

### How this addon works

Adding images to the username on: Messages (fav, love and follow), home page rows, profile page rows and studio activity (please say if you know about more)

### Possible alternatives

_No response_

### Additional context

This is a part of the scratch 4.0 addon
How it could look home page:
image

mybearworld:
Resolves #6270
Supersedes #6693 <small>(which I closed and deleted for... some reason)</small>

### Changes

Create a "Show profile pictures next to usernames" addon that does what it says.

### Reason for changes

PFPs are in most parts of the website.

### Tests

Tested in Edge.

### To Do

**Make it respond to changes in the DOM.** Currently, it doesn't work with most scratch-www things and also not for any new comments. I have tried doing this with `MutationObserver` and crashed my browser every time.
Also, I have probably missed some selectors, which means profile pictures there appear twice.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -149,6 +149,7 @@
   "forum-post-countdown",
   "paint-skew",
   "preview-project-description",
+  "more-pfps",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/more-pfps/addon.json
+++ b/addons/more-pfps/addon.json
@@ -1,0 +1,42 @@
+{
+  "name": "Show profile pictures next to usernames",
+  "description": "Shows the user's profile picture next to their username in several places on the website.",
+  "tags": ["community"],
+  "userscripts": [
+    {
+      "matches": "*",
+      "url": "userscript.js"
+    }
+  ],
+  "userstyles": [
+    {
+      "matches": "*",
+      "url": "userstyle.css"
+    },
+    {
+      "matches": "*",
+      "url": "in-href.css",
+      "if": {
+        "settings": { "in-href": false }
+      }
+    }
+  ],
+  "settings": [
+    {
+      "id": "in-href",
+      "name": "Show profile pictures even if the user is only linked",
+      "type": "boolean",
+      "default": false
+    }
+  ],
+  "versionAdded": "1.35.0",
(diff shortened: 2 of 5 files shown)


mybearworld:
I really hope there wasn't some really logical reason I closed the original that I'll remember when suddenly waking up at 3am lol

WorldLanguages:
> I have tried doing this with `MutationObserver` and crashed my browser every time.

This was probably happening:

1. Your observer gives your callback function an event
2. In response to the event, you modified the DOM
3. Your DOM modification called your callback function again, in this case you should not modify the DOM unless you want to end up in an infinite loop. You should ignore the event in this case.

Still, I wouldn't recommend using MutationObserver in addons, maybe just use waitForElement with a different selector that changes depending of the URL?

mybearworld:
> Still, I wouldn't recommend using MutationObserver in addons, maybe just use waitForElement with a different selector that changes depending of the URL?

I thought `waitForElement` shouldn't be ran multiple times, which is why I tried doing this with `MutationObserver`  - but this would be much more straight forward with `waitForElement`

WorldLanguages:
> I thought `waitForElement` shouldn't be ran multiple times

Not sure what you mean by that. Lots of addons use `waitForElement` inside a while(true) loop with markAsSeen:true.

mybearworld:
> > I thought `waitForElement` shouldn't be ran multiple times
> 
> Not sure what you mean by that. Lots of addons use `waitForElement` inside a while(true) loop with markAsSeen:true.

Thinking about it, I think I've just misinterpreted https://github.com/ScratchAddons/ScratchAddons/pull/6166#issuecomment-1586153746.

WorldLanguages:
If you're interested, we can talk about the waitForElement API on issue #6139 
I admit the `while(true)` pattern is a bit annoying and it would be way more intuitive to use an event-like system where you can subscribe to the event and cancel it. It's for historical reasons, keep in mind markAsSeen didn't exist when this API was first added.

PR #6704, opened 2023-09-27
Fix duplicate latestUpdate key in addon manifest

Samq64:
### Changes

Merges a duplicate `latestUpdate` key in `editor-number-arrow-keys`.

### Tests

Tested on Chromium.

PR #6707, opened 2023-09-28
Highlight mediarecorder update and feature new text-size setting

### LINKED ISSUE #6612: Update mediarecorder addon description
See comment https://github.com/ScratchAddons/ScratchAddons/pull/6603#pullrequestreview-1597239807
or we may end up merging more mediarecorder changes this time, so we shouldn't write until we know all the updates

WorldLanguages:
Resolves #6612

### DIFF
--- a/addons/custom-block-text/addon.json
+++ b/addons/custom-block-text/addon.json
@@ -49,7 +49,8 @@
   "versionAdded": "1.24.0",
   "latestUpdate": {
     "version": "1.35.0",
-    "newSettings": ["size"]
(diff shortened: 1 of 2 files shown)


Samq64 (approved):
Should `editor-number-arrow-keys` get `isMajor too?` since the values are customizable and it works outside blocks? I forgot to ask on #6704

WorldLanguages:
> Should `editor-number-arrow-keys` get `isMajor` too?

IMO no
I would even says mediarecorder would deserve isMajor instead

PR #6709, opened 2023-09-28
Fix CSS regression in settings page

### LINKED ISSUE #6706: `enablepopup` has buttons ordered vertically instead of horizontally
### Description

Open full screen settings and Cancel buttons misaligned
They're supposed to be side to side.

I think I know when this happened - see Environment.

### Steps to reproduce

1. Try to enable `scratch-notifier` from the popup

_No response_

### Environment

Browser: Edge 117
Operating system: Windows 11

I believe this is when it started: **[`50b7f15`](https://github.com/ScratchAddons/ScratchAddons/commit/50b7f157cfc02424b22fcb5a63d84d0a4e81359a)** (#6469). Reverting it fixes the problem.

Samq64:
Resolves #6706

Tested on Chromium

### DIFF
--- a/webpages/styles/components/enablepopup.css
+++ b/webpages/styles/components/enablepopup.css
@@ -43,6 +43,10 @@
   margin-bottom: 30px;
 }
 
+.popup .large-button {
+  display: inline-block;
+}
+
 .popup .large-button:not(:last-child) {
   margin-inline-end: 20px;
 }
--- a/webpages/styles/components/enablepopup.css
+++ b/webpages/styles/components/enablepopup.css
@@ -44,7 +44,7 @@
 }
 
 .popup .large-button {
-  display: inline-block;
+  display: inline-flex;
 }
 
 .popup .large-button:not(:last-child) {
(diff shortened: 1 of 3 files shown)


webpages/styles/components/enablepopup.css
```diff
@@ -43,6 +43,10 @@
   margin-bottom: 30px;
 }
 
+.popup .large-button {
+  display: inline-block;
```
WorldLanguages, line 47: Maybe inline-flex would be better in case we change CSS in the future?

DNin01 (commented):
Are there any other potential better long-term solutions? Like maybe only certain buttons should've used `display: flex` and all others should use `display: inline-flex` or `display: inline-block`.

WorldLanguages:
Personally, I believe this is simple as-is.
At some point CSS was going to make "inline" a separate property (it's unnecessarily merged with block/flex/etc into 1 single property) but I'm not sure where that suggestion ended up.

PR #6719, opened 2023-09-30
Create "Costume numbers in block dropdowns" addon

### LINKED ISSUE #4992: An addon that shows costume numbers in the costumes dropdown
### What this addon does
This addon would add the costume number before the costume name in the dropdowns where you select a costume so that you can tell... well, what the costume number of a certain costume is from the dropdown. I would use this occasionally.

### How it works
Self-explanatory

I don't know of any

A dropdown with the following options: 1. costume 1; 2. costume 2

mybearworld:
Resolves #4992

### Changes

Create a new "Costume numbers in block dropdowns" addon that shows costume numbers in block dropdowns.

A screenshot of a backdrop dropdown with costume numbers.

### Reason for changes

It can be helpful sometimes.

### Tests

Tested in Edge. It's compatible with `editor-searchable-dropdowns`.

### DIFF
--- a/addons/dropdown-costume-numbers/userscript.js
+++ b/addons/dropdown-costume-numbers/userscript.js
@@ -29,7 +29,7 @@ export default async function ({ addon, console }) {
       const cutThree = COSTUME_BLOCKS[block.type];
       allItems.forEach((el, i) => {
         if (cutThree && i >= allItems.length - 3) return;
-        el.lastChild.dataset.index = i;
+        el.lastChild.dataset.index = i + 1;
       });
     } else {
       dropDownDiv.classList.remove("sa-dropdown-costume-numbers");
--- a/addons/dropdown-costume-numbers/userscript.js
+++ b/addons/dropdown-costume-numbers/userscript.js
@@ -7,7 +7,7 @@ export default async function ({ addon, console }) {
   };
 
   const dropDownDiv = await addon.tab.waitForElement(".blocklyDropDownDiv");
-  const workspace = (await addon.tab.traps.getBlockly()).getMainWorkspace();
+  const workspace = await addon.tab.traps.getWorkspace();
 
   // This will show the original text without the numbers for a bit.
   // I'm not aware of any better way to detect dropdown close, though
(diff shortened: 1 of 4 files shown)


WorldLanguages:
Nice use case for CSS :before.

This addon calls getMainWorkspace once so it's probably going to have some known bugs. Remember `addon.tab.traps.getWorkspace()` already exists, but only a few addons use it for now.

mybearworld:
> This addon calls getMainWorkspace once so it's probably going to have some known bugs. Remember `addon.tab.traps.getWorkspace()` already exists, but only a few addons use it for now.

I'll use that - I really have no idea about how the traps stuff works, I just used the first thing that worked in the documentation :P

addons/dropdown-costume-numbers/userscript.js
```diff
@@ -0,0 +1,39 @@
...
+
+  const dropDownDiv = await addon.tab.waitForElement(".blocklyDropDownDiv");
+  const workspace = await addon.tab.traps.getWorkspace();
+
+  // This will show the original text without the numbers for a bit.
+  // I'm not aware of any better way to detect dropdown close, though
+  const waitForDropdownClose = () => {
+    const currentStyle = dropDownDiv.getAttribute("style");
+    return new Promise((resolve) =>
+      setInterval(() => {
```
WorldLanguages, line 17: Looks like we're looking for a way to run code only when there's a new dropdown that should be affected. Of course, a setInterval would do the trick, but it's pretty inefficient. Using waitForElement can also hurt performance depending on how it's used.

The best way to avoid setInterval (or other ugly ways to check if there's a new dropdown) is probably taking inspiration from the editor-searchable-dropdown addon:

Instead of while(true) waitForElement etc...
```js
const Blockly = await addon.tab.traps.getBlockly();
const oldDropDownDivShow = Blockly.DropDownDiv.show;
Blockly.DropDownDiv.show = function (...args) {
  try {
    // there's probably a new dropdown in the page
    // do stuff with the dropdown normally here...
  } catch(e) {console.error(e);}
  return oldDropDownDivShow.call(this, ...args);
}

mxmou: I think it would make more sense to override `Blockly.FieldDropdown.prototype.showEditor_` in this case. This would also make checking the dropdown type easier (`this.name === "COSTUME"`, `this.name === "BACKDROP"`).

PR #6727, opened 2023-10-01
Fix workspace undo bugs after workspace reload

### LINKED ISSUE #6691: Variables and lists can be deleted by undoing after changing settings in/toggling some addons
### Description

When I do something with the "Customizable block text style", "Customizable block shape", and possibly other addons, a dangerous action gets added to the undo history - if you click undo after enabling, disabling, or adjusting the settings in the aforementioned addons, it makes the variable and list blocks disappear from the block palette. What's more, when I undid in a sprite containing these variables, it asked me if I wanted to **quite literally DELETE THE USES OF THEM**. Also, if there is a comment attached to a block, it disappears if you undo once, and then variables and lists get deleted on the second undo, which sounds similar to a Scratch bug.

Redoing looks like it restores the variables in the block palette, but the block inputs are still messed up. Also, any variables that are also in other sprites seem to be unaffected.

Also, this happened somehow:
Blank field equals 42 block
Broadcast blank and wait, broadcast blank
(Yes, literal message dropdowns got deleted - there weren't variable blocks inserted there.)

This is the most hilarious destructive bug I have seen here yet.

Testers, make sure you try this out in a project you don't care about or somebody else's project so that nothing... like, auto-saves or something.

### Steps to reproduce

1. Toggle or reconfigure "Customizable block text style"
2. Right-click > Undo

_No response_

### Environment

Browser: Edge 117
Scratch Addons version: I was able to reproduce this all the way back to some version of v1.33.0 and TurboWarp (its addons haven't updated in a long while), and this bug is probably even older than that.
Operating system: Windows 11

WorldLanguages:
Resolves #6691

### Changes

Disables workspace undo after reloading the workspace. The whole undo stack gets cleared.
There might be a way to keep the old "undo" actions available, but I'm not aware of them.

### Tests

Fixes the bug in Chromium, addons custom-block-shape and custom-block-text
cc @DNin01 We probably need to check other addons that use `clearWorkspaceAndLoadFromXml`

### DIFF
--- a/addons/cat-blocks/userscript.js
+++ b/addons/cat-blocks/userscript.js
@@ -335,10 +335,12 @@ export default async function ({ addon, console }) {
     }
     const flyout = workspace.getFlyout();
     if (flyout) {
+      Blockly.Events.disable();
       const flyoutWorkspace = flyout.getWorkspace();
       Blockly.Xml.clearWorkspaceAndLoadFromXml(Blockly.Xml.workspaceToDom(flyoutWorkspace), flyoutWorkspace);
       workspace.getToolbox().refreshSelection();
       workspace.toolboxRefreshEnabled_ = true;
+      Blockly.Events.enable();
     }
   }
 }
--- a/addons/custom-block-shape/update-all-blocks.js
+++ b/addons/custom-block-shape/update-all-blocks.js
@@ -1,4 +1,7 @@
-export function updateAllBlocks(vm, workspace) {
+export function updateAllBlocks(vm, workspace, blockly) {
+  const eventsOriginallyEnabled = blockly.Events.isEnabled();
+  blockly.Events.disable(); // Clears workspace right-click→undo (see SA/SA#6691)
+
   if (workspace) {
     if (vm.editingTarget) {
       vm.emitWorkspaceUpdate();
@@ -14,4 +17,8 @@ export function updateAllBlocks(vm, workspace) {
       workspace.toolboxRefreshEnabled_ = true;
     }
   }
+
+  // There's no particular reason for checking whether events were originally enabled.
+  // Unconditionally enabling events at this point could, in theory, cause bugs in the future,
+  if (eventsOriginallyEnabled) blockly.Events.enable(); // Re-enable events
 }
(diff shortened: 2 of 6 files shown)


WorldLanguages:
Note that some other addons already use `ScratchBlocks.Events.disable();`

DNin01:
On the `master` branch, I have been able to replicate the bug with all the addons using `clearWorkspaceAndLoadFromXml()` in any way that don't require a refresh. Tested this pull request and the problem no longer occurs in the addons that were fixed so far.

All related code AFAIK: [Code search](https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons+%28clearWorkspaceAndLoadFromXml+OR+updateAllBlocks%29+NOT+path%3A%2F%5Eaddons%5C%2Fcustom-block-shape%5C%2Fupdate-all-blocks.js%2F&type=code). Mark as a draft while we work on the other addons?

WorldLanguages:
@DNin01 Can you check for any remaining bugs?

DNin01 (approved):
All of the addons are working fine - the undo glitches are no more! This change also seems to prevent dynamically enabling/disabling/configuring from starting the auto save timer (which means "Save Now" no longer appears in the corner afterward) in all addons except `cat-blocks` and `editor-theme3`. Might be worth researching that more in the future.

After dynamically enabling/disabling/configuring one of these addons, comments attached to blocks flicker and are still part of the undo history afterward, which means if you have a block comment in a sprite and you dynamically enable/disable/configure an addon, you can undo to delete it. That's partly a Scratch bug, so it might not be fixable.

The "Make a Block" modal still has weird undo behavior, but I don't think anyone undoes in it.

Oh yeah, all I did to get `cat-blocks` to dynamically enable was to add `"dynamicEnable": true` to the `addon.json` and it seems to just work.

mxmou:
> There might be a way to keep the old "undo" actions available, but I'm not aware of them.

I think this is the simplest way:
```js
if (vm.editingTarget) {
  const clearUndo = workspace.clearUndo;
  workspace.clearUndo = () => {};
  vm.emitWorkspaceUpdate();
  workspace.clearUndo = clearUndo;
}
```
The code for reloading the main workspace needs to be moved after `Blockly.Events.disable()` for this to work.

WorldLanguages:
@mxmou If you're interested, it would be nice to have a single "workspace reload" module/library that is used by all addons, and that can be customized through different functions and/or parameters.
Right now, some modules are shared and some are not, which is a bit messy.

PR #6728, opened 2023-10-01
Reconsider usage of timestamps in URLs to bypass cache

### LINKED ISSUE #6717: Check Scratch API endpoints where we use `nocache` or `timestamp` in the URL
The requests with those URL parameters are being intentionally throttled, which makes Scratch Messaging slow, we should check whether we want to bypass the cache at all

WorldLanguages:
Resolves #6717

### Changes

- Removes `nocache=${Date.now()}` and `timestamp=${Date.now()}` from Scratch Messaging code. These requests may now get cached by the browser (disk cache) or by the Scratch API.
- When an uncached version of the request is really needed, rename the URL paramater to be descriptive.

### Reason for changes

The "nocache" keyword is filtered by the Scratch API and is intentionally throttled.

### DIFF
--- a/addons/turbowarp-player/userscript.js
+++ b/addons/turbowarp-player/userscript.js
@@ -32,11 +32,14 @@ export default async function ({ addon, console, msg }) {
     if (addon.tab.redux.state?.preview?.projectInfo?.public === false) {
       let projectToken = (
         await (
-          await fetch(`https://api.scratch.mit.edu/projects/${projectId}?current_time_to_get_updated_project_token=${Date.now()}`, {
-            headers: {
-              "x-token": await addon.auth.fetchXToken(),
-            },
-          })
+          await fetch(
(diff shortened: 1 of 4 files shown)


mybearworld:
Does that mean Scratch Messaging will now act like scratch-www with its messages?

mybearworld:
> > Does that mean Scratch Messaging will now act like scratch-www with its messages?
> 
> Depends on what you mean.

Scratch-www often takes a bit of time before showing a message - but more importantly, it also shows "ghost messages" a lot, e.g. messages that are already read.

WorldLanguages:
> it also shows "ghost messages" a lot, e.g. messages that are already read.

I think that's because they don't set the message count to 0 after they were marked as read. I made it so after marking messages as read from the popup, the message count is obtained again bypassing the cache, so there should be no ghost messages after reopening the popup.

> Scratch-www often takes a bit of time before showing a message

It's likely Scratch Addons will need more time to react to message count changes.

Issue #5666, opened 2023-01-25
Limit the number of list items that can be stored inside an addon's settings


DNin01:
### Why this would be helpful

Lists are one of the biggest offenders to Scratch Addons' storage, which is [limited](https://github.com/ScratchAddons/ScratchAddons/issues/5190). I mean, look at how many items I could add if I wanted to.
![30 list items in the settings of a single addon](https://user-images.githubusercontent.com/106490990/214464709-41df142e-0863-4f7e-bc1f-094acea39452.png)

### My suggestion

If this is becoming a problem, it might be worth considering limiting how many items that the user can add to each list in addon settings and the length of each of them. I was able to prevent SA from saving my settings by filling up lists with items, and if there are people who add many items to the Cloud games or Customizable navigation bar addons, it could happen to them too.

For Customizable navigation bar, you're probably not going to need any more than 8 links, and for Cloud games, I don't know.

### Possible alternatives

- Find a way to compress the data. Cloud games definitely doesn't need the `https://scratch.mit.edu/` at the start of _every_ list item if you're only going to modify the part after that. We could do this in addition to adding a limit, too.
- Related: #5190

WorldLanguages:
Good idea, we could have a default limit of 10 items and allow it to be customized on each addon.json

penguinmoose:
> Good idea, we could have a default limit of 10 items and allow it to be customized on each addon.json

Do we just hide the "plus" icon when the number of items exceeds 10 (or whatever limit is imposed for that addon)? Or should we gray it out?

BroJac5246:
> > > Good idea, we could have a default limit of 10 items and allow it to be customized on each addon.json
> > 
> > Do we just hide the "plus" icon when the number of items exceeds 10 (or whatever limit is imposed for that addon)? Or should we gray it out?
> 
> I was thinking of an alert when you try to add an 11th item, but that works too. Might want a tooltip to let the user know that they can only add up to 10 items in that case.

I think we should grey it out, since that's less obtrusive but still understandable.

WorldLanguages:
Related: #5849

PR #6734, opened 2023-10-02
Create "Report templates in the forums" addon

### LINKED ISSUE #6733: Report templates for the forums
### Why this would be helpful

Often, you don't need much variety with report messages, for example if the topic is a guide, spam or a duplicate.

### How this addon works

This addon would add some report templates to the forum page. These could be 100% customizable with the `table` setting type.

### Possible alternatives

_No response_

### Additional context

Something like this:

The report page with a couple of templates added

mybearworld:
Resolves #6733

### Changes

Create a "Report templates in the forums" addon that does what it says.

Report templates after the report reason field with one in use

### Reason for changes

To make reporting posts easier.

### Tests

Tested in Edge.


mybearworld:
It might be nice to have a note that appears only when the user uses a language that is not English that says that these reasons should ideally be kept in English - is that possible in some way?

WorldLanguages:
This will have higher chances of being merged if the templates aren't customizable. For that level of customization, general-purpose extensions can help populate textareas with user-defined templates.

mybearworld:
> This will have higher chances of being merged if the templates aren't customizable. For that level of customization, general-purpose extensions can help populate textareas with user-defined templates.

True, that makes sense. I don't see myself ever really customizing these anyway.

CST1229:
> This will have higher chances of being merged if the templates aren't customizable

I would expect some people to suggest making it customizable if you don't make them customizable.

Secret-chest:
Guides may still be allowed.

mybearworld:
> Guides may still be allowed.

Where?

Secret-chest:
> Where?

See "Why most suggestions fail"

Secret-chest:
Why hardcode things when it's easier not to?

mybearworld:
> Why hardcode things when it's easier not to?

What do you mean? How would I not have them hardcoded?

DNin01:
> This will have higher chances of being merged if the templates aren't customizable. For that level of customization, general-purpose extensions can help populate textareas with user-defined templates.

Yeah, maybe. It would save some storage too. But some users may not agree with the built-in messages, and customizable messages can have more personality.

At that point, maybe it's better for everyone to use a general-purpose extension for this, no need for Scratch Addons?

BroJac5246:
> > This will have higher chances of being merged if the templates aren't customizable. For that level of customization, general-purpose extensions can help populate textareas with user-defined templates.
> 
> Yeah, maybe. It would save some storage too. But some users may not agree with the built-in messages, and customizable messages can have more personality.

Perhaps multiple options with different styles could be added?
> At that point, maybe it's better for everyone to use a general-purpose extension for this, no need for Scratch Addons?

What extension would that be? And how would that be communicated with users?

Samq64:
> Perhaps multiple options with different styles could be added?

Is it really necessary? The submissions aren't public and users are still free to just type their own message into the textbox.

Secret-chest:
> Perhaps multiple options with different styles could be added?

That's overkill, just have it be customisable.

BroJac5246:
> > Perhaps multiple options with different styles could be added?
> 
> That's overkill, just have it be customisable.

How is it not overkill, then, to have it be fully customizable?

Secret-chest:
> How is it not overkill, then, to have it be fully customizable?

Because you have more flexibility than some useless "styles".

BroJac5246:
> > How is it not overkill, then, to have it be fully customizable?
> 
> Because you have more flexibility than some useless "styles".

Offering multiple styles would offer more flexibility than offering no choices while solving DNin01's points. It probably doesn't matter that much, but I also think that the moderators might go insane if they keep getting near-identical reports 😂

Also, why do you put "styles" in quotes? It's a valid literary term.

Secret-chest:
> more flexibility

No it's not. I have many other templates I want, but if they aren't customisable we will start uselessly negotiating them.

BroJac5246:
> > more flexibility
> 
> No it's not. I have many other templates I want, but if they aren't customisable we will start uselessly negotiating them.

By the very definition of "more" it is more flexibility... an amount of options greater than one is indisputably more options than one!

@DNin01 has brought up the point that it might not be practical to store large custom strings that the user inputs.

That's why I think we should have a few styles. One could be formal, one could be thankful, one could be playful and full of emojis, etc. So people could pick whatever style best fits their personality.

Or, of course, we could determine that it doesn't matter and make it just have a to-the-point request. But either way, custom messages are overkill.

mybearworld:
> That's why I think we should have a few styles. One could be formal, one could be thankful, one could be playful and full of emojis, etc. So people could pick whatever style best fits their personality.

I feel like no prebuilt style would actually match a user's personality exactly, and there'd be tons of requests of adding a style that matches someone's personality better.

> and full of emojis,

Probably not - we don't know if their report system handles emoji better than the rest of the forums do.

BroJac5246:
> > That's why I think we should have a few styles. One could be formal, one could be thankful, one could be playful and full of emojis, etc. So people could pick whatever style best fits their personality.
> 
> I feel like no prebuilt style would actually match a user's personality exactly, and there'd be tons of requests of adding a style that matches someone's personality better.

True. Do we know how we would implement custom messages? Like, regarding storage limits?

> > and full of emojis,
> 
> Probably not - we don't know if their report system handles emoji better than the rest of the forums do.

Good point. I guess emoticons could maybe be used, but that would need consideration.

mybearworld:
> True. Do we know how we would implement custom messages? Like, regarding storage limits?

There's an issue for this: #5190

DNin01:
Sure, if customization doesn't make the cut, having options for styles seems nice at first glance, but if that's the point, users will still want to have more control over the messages they send, and if they don't end up liking one of our template styles, it's not ideal/efficient to have to modify it every time they use a template.

I don't know if there's any one solution that balances UX, expression, and stability. Even I don't know what to do; there are so many ups and downs. And at that point, is any automated option better than just having the user write the report reasoning themselves?

Honestly though, I don't think the team really cares about how you style your report. Report reasonings are private messages - styles still seem overkill to me from a UX standpoint. If we make nice, constructive templates, I think a handful of users would still use this and we can always reconsider.

DNin01:
Regarding storage concerns, I once suggested [separating certain addon setting namespaces into separate storage keys](https://github.com/ScratchAddons/ScratchAddons/issues/5190#issuecomment-1296328270). Distributing storage use across multiple keys helps up stay below [the quota for individual storage keys](https://developer.chrome.com/docs/extensions/reference/storage/#property-sync), which might be enough to fit user-defined templates, considering that currently the limit of our string field setting is 100 characters (I believe?) and [table length limits are always an option](https://github.com/ScratchAddons/ScratchAddons/issues/5666).
> **QUOTA_BYTES_PER_ITEM**
> `8192`

Secret-chest:
> But either way, custom messages are overkill.

It's. Not. About. The. Style. It's. About. The. Actual. Meaning.

Secret-chest:
Also, why are we translating them?

mybearworld:
> > But either way, custom messages are overkill.
> 
> It's. Not. About. The. Style. It's. About. The. Actual. Meaning.

What other meanings would you use? They could probably be added here.

> Also, why are we translating them?

Just the names - the actual reasons are always in English.

BroJac5246:
> > **QUOTA_BYTES_PER_ITEM**
> > `8192`
> 
> ...which might be enough to fit user-defined templates, considering that currently the limit of our string field setting is 100 characters (I believe?) and [table length limits are also an option](https://github.com/ScratchAddons/ScratchAddons/issues/5666).

I agree with you now, but is 100 characters enough? Can that limit be easily increased?

BroJac5246:
> > But either way, custom messages are overkill.
> 
> It's. Not. About. The. Style. It's. About. The. Actual. Meaning.

Then why not just provide one descriptive option that can have all of the necessary meaning? In that case the whole problem of implementing custom messages would be unnecessary.

(Though I think DNin01 has a good idea and I now think we should go with that)

Secret-chest:
> What other meanings would you use? They could probably be added here.

If you want to fit ten meanings then sure:

* Please Move To
* Correct Spelling
* More Descriptive Title
* Remove Mention of Browser Extensions
* Close as Rejected
* Close as Solved/Implemented
* Please Reopen
* Remove Spam

mybearworld:
> * Correct Spelling
> * More Descriptive Title

Should these be in a report?

> * Remove Mention of Browser Extensions
> * Remove Spam

Those are already there.

> * Please Move To
> * Close as Rejected
> * Close as Solved/Implemented
> * Please Reopen

I'll add these.

Secret-chest:
Should the messages be polite? They should be short and easy to edit.

Secret-chest:
What happened to SA? Now we reject settings because IT MiGhT ConFUsE SoME UsERs???

Secret-chest:
Also, I only tolerate British English. Food for thought.

mybearworld:
These messages are one way private messages between you and the Scratch Team. Keeping the writing style 100% accurate to your writing style isn't important, but I'd say being polite is. This is actually one of the reasons I wanted to make this addon - too many of my report reasons are just "spam, please close" which I don't really like.

Secret-chest:
Less "Could you" though. Keep it shortm

BroJac5246:
> Also, I only tolerate British English. Food for thought.

You can always try avoid all instances of American English and see how it goes 🤷

But Scratch is American, hence why we use American English.

mybearworld:
> Less "Could you" though. Keep it shortm

Why? What else would you use?

mybearworld:
> Just "Please".

How would that make it easier to edit?

WorldLanguages:
I would hope we first focus on providing good defaults for each type of report. Even if the messages aren't customizable when this is merged, that doesn't mean they will never be! (addons aren't frozen when they're first merged!)
I'm pretty sure Scratch would add this to the Report pages themselves, but it's probably hard to customize DjangoBB, and the Scratch Team wouldn't write a "userscript" on top of their own website, like we typically do (it would be beneficial to them IMO, but whatever).

We also need to be careful with the templates we provide. And we don't want to affect the amount of report messages the Scratch Team gets.

mybearworld:
> And we don't want to affect the amount of report messages the Scratch Team gets.

How would this addon influence that?

WorldLanguages:
> How would this addon influence that for the worse?

I don't have the answer. I'm saying it should be kept in mind.

MaterArc:
> > And we don't want to affect the amount of report messages the Scratch Team gets.
> 
> How would this addon influence that for the worse?

People tend to be lazy and therefore tend to not report things if it requires them to write a report without any template or in a paragraph form. With this addon, more people might start reporting things that they see, as it is much easier when you have a button that can pretty much write out the majority of the report for you (which can get hectic, seeing as ScratchAddons has over 200k users and there are bound to be many more duplicate reports) Which in tern, just causes more work for the ST to read. Obviously, this isn't science or backed data, and as World_Languages said above, it's just another thing to add on the the discussion when talking about adding an addon like this. On the flip side though, I really like the idea of this. TheForumHelpers had something similar to this on their old website called QuickReply (which unfortunately got removed). I can say that QuickReply saved me a lot of time personally, but to each his own 😅

Secret-chest:
Making the templates empty by default and customisable could help this.

mybearworld:
I actually didn't/don't like the QuickReply-like things much, because those are very much encouraging you to post only very generic messages, which often don't apply. This is different for report messages, where the amount of different report messages is pretty tiny. I do imagine they'd get more reports, yes, but those would mostly actually be fitting. People that report posts incorrectly do have the option of just using two buttons without this addon.

PR #6744, opened 2023-10-04
Add and use decimal setting type

apple502j:
See also ScratchAddons/manifest-schema#65
Please, at least use `untranslated`...

### DIFF
--- a/addons/editor-number-arrow-keys/addon.json
+++ b/addons/editor-number-arrow-keys/addon.json
@@ -50,6 +50,7 @@
       "type": "decimal",
       "default": 1,
       "min": 0,
+      "max": 100,
       "step": 0.1,
       "if": { "settings": { "useCustom": true } }
     },
@@ -89,6 +90,7 @@
       "type": "decimal",
       "default": 10,
       "min": 0,
+      "max": 100,
       "step": 0.1,
       "if": { "settings": { "useCustom": true } }
     },
@@ -128,6 +130,7 @@
       "type": "decimal",
       "default": 0.1,
       "min": 0,
+      "max": 100,
       "step": 0.01,
       "if": { "settings": { "useCustom": true } }
     },
--- a/addons/editor-number-arrow-keys/addon.json
+++ b/addons/editor-number-arrow-keys/addon.json
@@ -47,9 +47,10 @@
     {
       "name": "Change on regular key press",
       "id": "regularCustom",
-      "type": "string",
-      "default": "1",
-      "max": 8,
+      "type": "decimal",
+      "default": 1,
+      "min": 0,
+      "step": 0.1,
       "if": { "settings": { "useCustom": true } }
     },
     {
@@ -85,9 +86,10 @@
     {
       "name": "Change on Shift+Key",
       "id": "shiftCustom",
-      "type": "string",
-      "default": "10",
-      "max": 8,
+      "type": "decimal",
+      "default": 10,
+      "min": 0,
+      "step": 0.1,
       "if": { "settings": { "useCustom": true } }
     },
     {
@@ -123,9 +125,10 @@
     {
       "name": "Change on Alt+Key",
       "id": "altCustom",
-      "type": "string",
-      "default": "0.1",
-      "max": 8,
+      "type": "decimal",
+      "default": 0.1,
+      "min": 0,
+      "step": 0.01,
       "if": { "settings": { "useCustom": true } }
     },
 
--- a/addons/editor-number-arrow-keys/userscript.js
+++ b/addons/editor-number-arrow-keys/userscript.js
@@ -122,14 +122,9 @@ export default async function ({ addon }) {
         : e.altKey
         ? addon.settings.get("altCustom")
         : addon.settings.get("regularCustom");
-      if (settingValue === "") settingValue = 0;
-      let valueAsFloat = parseFloat(settingValue);
-      if (valueAsFloat < 0) valueAsFloat *= -1; // If user typed a negative number, we make it positive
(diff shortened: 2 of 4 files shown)


WorldLanguages:
> Please, at least use `untranslated`...

?

DNin01:
> > Please, at least use `untranslated`...
> 
> 
> 
> ?

I think they meant setting the type of the "numerical" string settings to "untranslated" to specify that the default value is not translatable. Whoops!
> Setting default values can be translated if the `type` is `string`:
- `setting.default` - `addon-id/@setting-default-SETTINGID`

WorldLanguages:
> setting the type of the "numerical" string settings to "untranslated"

Do we even have an "untranslated" type?

DNin01:
> > setting the type of the "numerical" string settings to "untranslated"
> 
> 
> 
> Do we even have an "untranslated" type?

[Yup.](https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons%20untranslated&type=code)

Samq64:
This has the same problem as #3880 which is probably fine but the step will need to be decreased for @Secret-chest's use case.

Secret-chest:
The step should be 0.0001, we're not going to encourage using the up/down buttons to change the setting. Also, there's no reason to make it smaller for Alt.

DNin01 (changes requested):
I feel like not having stepping at all and just being able to set them to whatever you want might be better than being able to change these settings with the arrow keys or buttons. Alternatively, the types could be switched to "untranslated" for a string with an untranslated default value.

I would prefer we get the customizability improvement right the first time so we don't have to review this again for v1.36, the next next major update, or maybe revert this customizability for now until we can get it right.

apple502j:
It doesn't make sense to support arbitrary-precision decimal. There is not much difference `.01` would make in most use cases.

DNin01 (commented):
> It doesn't make sense to support arbitrary-precision decimal. There is not much difference .01 would make in most use cases.

Alright, yeah, maybe it doesn't make sense. `0.1` increments and one `0.01` is still okay, though I would encourage discussion around how precise it should get, `.005` at the least.

Personally, I understand how getting it more precise than the thousandth isn't important.

This also reminds me that this addon can't currently increment numbers that are too large or too precise, so we might also want to add **maximums** for these settings. Probably `100` or `1000` tops?

Secret-chest:
Shut up. 0.125 is not any random multiple of 1/1000 - it's 1/8. It's not about the difference from 0.12, it's about it being a power of 2.

Secret-chest:
Alright then, I'll just hack the JSON.

Samq64 (approved):
I don't think discussion about minimum values should block this from being merged for 1.35.

DNin01 (commented):
As I said before, we also want to add value maximums while we're at it, and my proposal would be 100, maybe 1,000.

BroJac5246:
> Shut up.

😕
>0.125 is not any random multiple of 1/1000 - it's 1/8. It's not about the difference from 0.12, it's about it being a power of 2.

I would also like to see thousands. Hundredths seem low for a decimal input.

mxmou:
The decimal input should at least accept all values of the select setting. It's currently only possible to set "change on regular key press" to 0.01 if "use custom values" is disabled.

I think `step="any"` is the best option - someone who decides to use these settings probably wants to specify a smaller or larger increment than the predefined options allow.

Secret-chest:
Also, all settings should have the same step.

WorldLanguages:
I think you see why I just went for string settings. It didn't make sense to block customization efforts due to all of this :P

DNin01 (commented):
Another thing, the userscript probably doesn't need to parse the values anymore since they're always going to be a float.

DNin01:
> I don't think discussion about minimum values should block this from being merged for 1.35.

I just think that if we decide to upgrade this again in the future, it would be kind of strange to put up another update notification to say the settings got even more precise. Once we announce our addition of these custom value fields, they'll already feel like they're in their final form.

Any further changes/reconsiderations (even delaying the release of the addition) _before v1.35_ are acceptable, though.

DNin01 (approved):

Secret-chest:
Fix the steps.

WorldLanguages:
Can we move this to the v1.36 milestone?

WorldLanguages:
Nevermind, this has 2 approvals

Secret-chest:
Please don't merge. Make the steps smaller, or even better, any.

apple502j:
@Secret-chest That will be left for another PR.

Secret-chest:
NOOOOO!

WorldLanguages:
How come you can't enter more than 2 decimal digits? You can't even enter `0.01` which is a valid option with "custom values" off! What's up?

WorldLanguages:
> It doesn't make sense to support arbitrary-precision decimal.

Then why did we add the setting?

WorldLanguages:
Why was this merged? String settings worked just fine and were actually customizable and functional

mxmou said:
> The decimal input should at least accept all values of the select setting. It's currently only possible to set "change on regular key press" to 0.01 if "use custom values" is disabled.

DNin01 said:
> it would be kind of strange to put up another update notification to say the settings got even more precise

Secret-chest:
Guess I'm hacking the JSON because of some arbitrary decision made to NoT COnFuSe uSerS!

WorldLanguages:
We should either revert this PR or possibly revert #6625 instead (and wait until v1.36 is released to allow for custom values in this addon).
Or as secret-chest said, "fix the steps" (tip: I had no idea what you meant by that, try being more clear such as saying 0.01 can't be entered)

PR #6747, opened 2023-10-07
Custom nicknames addon

### LINKED ISSUE #5317: make nickname

**(notice)** My English isn't very good, so I'll simple briefly.
### What this addon does

1.Select the person you want to add a nickname to
2.enter a nickname
### request
Setting 1 and 2 should be simple

### Possible alternatives

_No response_

スクリーンショット 2022-11-03 211355
In this way, users can freely add other's nicknames

mybearworld:
Resolves #5317

### Changes

Create a "Custom Nicknames" addon that allows you to set custom nicknames for anyone.

The header of JLdude33's profile with a "Set nickname" button

The header of JLdude33's profile with the nickname set as JLnickname33 - the "Set nickname" button is now a "Change nickname" button

A comment chain involving JLdude33, where every link to JLdude33 mentions "JLnickname33"

### Reason for changes

Easier recognition of users

### Tests

Tested in Edge.

### DIFF
--- a/addons-l10n/en/nicknames.json
+++ b/addons-l10n/en/nicknames.json
@@ -1,0 +1,5 @@
+{
+  "nicknames/change-nickname": "Change nickname",
+  "nicknames/set-nickname": "Set nickname",
+  "nicknames/which-nickname": "What should {username}'s nickname be?"
+}
--- a/addons-l10n/en/nicknames.json
+++ b/addons-l10n/en/nicknames.json
@@ -1,5 +1,5 @@
 {
   "nicknames/change-nickname": "Change nickname",
   "nicknames/set-nickname": "Set nickname",
-  "nicknames/which-nickname": "What should {username}'s nickname be?"
+  "nicknames/which-nickname": "What should {username}'s nickname be? Leave this blank to remove the nickname."
 }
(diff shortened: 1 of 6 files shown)


Samq64:
It might make sense to wait until there's a proper addon storage API.

DNin01 (commented):
We'd also have to consider how this will affect communication.

- What happens to mentions in comments? Do they mention the user's username or the nickname you set?
- What happens if _you_ want to mention a user you assigned a nickname? What would come out when you post your comment or write their nickname anywhere else?
- What happens if you copy text from a comment with a nickname mention?

Although it's not impossible, problems like this make we wonder if this whole addon is worth our time and effort.

WorldLanguages:
> We'd also have to consider that people will probably be trying to mention people by their nickname.

That could end up being problematic.

mybearworld:
> > We'd also have to consider that people will probably be trying to mention people by their nickname.
> 
> That could end up being problematic.

Potential solution: Whenever somebody types a nickname, offer a dropdown list of their actual usernames.

Something like this (*very* rough mockup):
A rough mockup of a dropdown that would appear when someone begins to type a nickname.

Samq64:
> It might make sense to wait until there's a proper addon storage API.

Did I misunderstand something here? Local storage works fine for now right?

WorldLanguages:
> Did I misunderstand something here? Local storage works fine for now right?

Yes, but it's not exportable and it doesn't sync between devices.

Issue #6522, opened 2023-08-15
Check if all or any conditions are met with `if` for addon manifest


DNin01:
### Why this would be helpful

Sometimes we want to hide addon settings or deactivate userstyles unless all of a few conditions are met, and sometimes any of a few.

### My suggestion

Here's an attempt at a syntax:
```json
{
  "if": [
    "any",
    [ { "setting": "one" }, true ],
    [ { "setting": "two" }, true ],
    { "addon": "example" }
  ]
}
```

Or:
```json
{
  "if": {
    "requirement": "any",
    "settings": {
      "one": true,
      "two": true
    },
    "addonsEnabled": [
      "example"
    ]
  }
}
```

These syntaxes would both work the same but are structured differently - in this example, if any of the notated conditions are met or if the `example` addon is enabled, the setting will be visible.

We could also try to come up with more powerful syntaxes so we can evaluate multiple separate conditions instead of just all of everything or any of everything, which would be useful for mixing different logic gates.

### Additional context

Related: #5692

DNin01:
I noticed VS Code seems to use the JS condition syntax for the "when" property for keyboard shortcuts, which you can see in the "when" column in keyboard shortcuts settings. If VS Code can, we might be able to do something similar.

```json
{
  "if": "addonEnabled('example') || self.settings('one') || self.settings('two')"
}
```

BroJac5246:
> @DNin01 The problem is we can't use eval() and it's probably overkill to write a small JS parser ourselves

Could [this](https://developer.chrome.com/docs/extensions/mv3/sandboxingEval/) be used to allow us to take advantage of `eval()`? I'm not familiar with it, so the answer could very much be "no," but I am curious.

Also, how complicated is a "small JavaScript parser"? Since logic is part of JavaScript by default, is it really that hard? (Once again, I'm not super familiar with this, and the answer may be "yes.")

WorldLanguages:
> Could [this](https://developer.chrome.com/docs/extensions/mv3/sandboxingEval/) be used to allow us to take advantage of `eval()`? I'm not familiar with it, so the answer could very much be "no," but I am curious.

I think it's not supported in Firefox, unfortunately.

> Also, how complicated is a "small JavaScript parser"?

As long as someone builds it in a way that's easy to maintain, it can be as complicated as you want :P

PR #6760, opened 2023-10-11
Math-in-inputs

Jazza-231:
Resolves a suggestion in the community server (https://discord.com/channels/806602307750985799/1160942498248663120)

### Changes

If you input, for example "100/2" into any number input (wip) and press enter or click off (wip) it will evaluate your input to 50

### Reason for changes

just nice to be able to do math, i know griffpatch for example uses the actual blocks to do math so this way he doesnt have to

### Tests

ive tested it in brave, it seems to work except clicking off of the sprite properties doesnt evaluate it, and the costume outline width input isnt supported either (hence why its a draft pr)


addons/math-in-inputs/userscript.js
```diff
@@ -0,0 +1,72 @@
...
+
+  function parseMath(value) {
+    if (!/^[0-9+\-*/().^]+$/.test(value)) {
+      return value;
+    }
+
+    const processedValue = value.replace(/(\d+)\^(\d+)/g, power);
+
+    try {
+      return eval(processedValue) || 0;
```
WorldLanguages, line 36: Whoever is reading this, we will not use eval. But it's fine while this stays as a draft.

addons/math-in-inputs/userscript.js
```diff
@@ -0,0 +1,72 @@
+export default async function ({ addon }) {
+  function getRegexFromSettings() {
+    let textInInputs = addon.settings.get("textInInputs");
+    return textInInputs ? /^[0-9a-z+\-*^/().]+$/ : /^[0-9+\-*^/().]+$/;
```
pufferfish101007, line 4:
```suggestion
return textInInputs ? /^[0-9a-z+\-*^/().]+$/i : /^[0-9+\-*^/().]+$/;
```

pufferfish101007 (changes requested):

addons/math-in-inputs/addon.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "Allow math in number inputs",
+  "description": "Use math symbols (*, /, -, +, etc) in number inputs and have them automatically solve.",
```
mybearworld, line 3: These should probably be phrased more similarly to other addons.

```suggestion
  "name": "Mathematical expressions in number inputs",
  "description": "Evaluates mathematical expressions (e.g. 2+4*3) in number inputs in the editor automatically.",
```

addons/math-in-inputs/userscript.js
```diff
@@ -0,0 +1,72 @@
...
+    else if (el.matches(".mediaRecorderPopupContent input[type=number]")) {
+      // Number inputs in `mediarecorder` addon modal
+      return true;
+    } else if (el.className.includes("input_input-form_")) {
+      if (el.matches("[class*=sprite-info_sprite-info_] [class*=input_input-small_]")) {
+      }
+      // Sprite X/Y coordinates, size and direction (excludes sprite name)
+      return true;
+    } else if (el.matches("[class*=paint-editor_editor-container-top_] input[type=number]")) {
+      // Number inputs in costume editor (note that browsers already provide up/down clickable buttons for these)
```
mybearworld, line 19:
```suggestion
```

addons/math-in-inputs/userscript.js
Jazza-231: Ah darn I thought the regex check would be good enough, ill make a custom parser

addons/math-in-inputs/userscript.js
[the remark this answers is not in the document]
Jazza-231: Uhm yes I think capitilisation is needed for some inputs so it does have to be case insensitive, my bad 🤣

addons/math-in-inputs/userscript.js
```diff
@@ -0,0 +1,72 @@
...
+
+  // I 100% stole this part of the code from "editor-number-arrow-keys"
+  const isSupportedElement = (el) => {
+    if (el.classList.contains("blocklyHtmlInput")) return true;
+    else if (el.matches(".mediaRecorderPopupContent input[type=number]")) {
+      // Number inputs in `mediarecorder` addon modal
+      return true;
+    } else if (el.className.includes("input_input-form_")) {
+      if (el.matches("[class*=sprite-info_sprite-info_] [class*=input_input-small_]")) {
+      }
```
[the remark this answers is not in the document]
Jazza-231: I just recopied it from the thingy, no clue what was happening there

addons/math-in-inputs/userscript.js
Jazza-231: I did in fact not make it but its done

Jazza-231:
Thanks to everyone who reviewd and suggested, now just gotta finish the to do

Jazza-231:
I've spent a while trying to figure out to do items 2 and 3 and can not for the life of me figure out how to get those working, looking through so much code and nothing is working so if anyone got ideas plz help otherwise this gonna die lol

addons/math-in-inputs/userscript.js
```diff
@@ -149,9 +141,9 @@ export default async function ({ addon }) {
         mutation.addedNodes.forEach(function (node) {
           if (
             node instanceof HTMLElement &&
-            (node.className.includes("input-group_input-group_plJaJ") ||
-              node.classList.contains("input_input-form_l9eYg") ||
-              node.classList.contains("mediaRecorderPopupContent"))
+            (node.classList.contains(addon.tab.scratchClass("input-group_input-group")) ||
+            node.classList.contains(addon.tab.scratchClass("input_input-form")) ||
```
WorldLanguages, line 145: scratchClass isn't supposed to use this way. Use className.includes until the last underscore

Jazza-231: I have grasped more knowledge of how this addon should work (through experimentation) and plan to fix everything I see wrong with it. Will do tmr (it is 3am)

Jazza-231:
I doubt anyone is reading this but this is a list of current problems:

- [ ] Outline width input doesn't switch to type=text
- [ ] Pressing enter on sprite properties does not keep the value (but clicking off does)
- [ ] Paint snapping and onion skinning inputs don't switch to type=text

I think that's all the current issues.

Jazza-231:
If I'm not lying to myself, absolutely everything should be working now, pressing enter on ALL number inputs work, pressing ctrl + enter on ALL inputs work, and clicking off of ALL number inputs works. text can be written in the inputs it needs to be, it parses everything perfectly.

Jazza-231:
I broke a different addon :( ill look into it

Jazza-231:
I've said this like 10 times now but I think its actually fully ready (watch someone find more problems fr)

addons/editor-number-arrow-keys/addon.json
```diff
@@ -4,9 +4,8 @@
   "tags": ["editor", "codeEditor", "recommended"],
   "versionAdded": "1.34.0",
   "latestUpdate": {
-    "version": "1.35.0",
-    "newSettings": ["useCustom"],
-    "temporaryNotice": "You can now increment input fields that are not in blocks, like the sprite properties, and fully customize the values."
+    "version": "1.36.0",
+    "temporaryNotice": "More inputs are now supported, such as color dropdowns, and onion skinning inputs."
```
mybearworld, line 8: This should probably be in a separate PR, unless it's required by the addon to work (which it probably isn't).

Jazza-231:
I have separated the more inputs into #6810

WorldLanguages:
This is ready to review then, right? Does it work in all types of inputs it's supposed to?

Jazza-231:
Yeah i think it all works perfectly

Jazza-231:
Yeah i think it all works perfectly

WorldLanguages:
> This is ready to review then, right? Does it work in all types of inputs it's supposed to?

I think I brought up on Discord that this is problematic, because what if you want a sprite to say `2+2` (literally)?

DNin01:
> > This is ready to review then, right? Does it work in all types of inputs it's supposed to?
> 
> I think I brought up on Discord that this is problematic, because what if you want a sprite to say `2+2` (literally)?

We could just differentiate between numerical block inputs and text block inputs? (Might also be helpful in `editor-number-arrow-keys`) The problem with that is that the `() = ()`, `() < ()` and `() > ()` blocks are text inputs.

Jazza-231:
Due to issues with block inputs that are actually text inputs, and the fact I could not find a way to differentiate between them based on the currently focused input on the blocks, the addon now only evaluates on ctrl+enter, and the addon.json reflects this. A setting has been clarified, and input types are still changed when they are needed to. enter now does nothing. hopefully this is all good and people can use it.

Jazza-231:
Floating point errors are now accounted for and fixed, using a newly added library because both math-in-inputs and editor-number-arrow-keys use it, all math works with it (power, multiply, divide, addition, subtraction) besides exponents being decimals which I don't bother trying to account for floating point error because holy fuck it hards, hopefully this is actually done, this has been quite the ride

Jazza-231:
i think this is ok? no one has found anything wrong with it

addons/math-in-inputs/userscript.js
```diff
@@ -0,0 +1,163 @@
+import { shiftDecimalPointToRight, shiftDecimalPointToLeft } from "../../libraries/common/cs/math-on-decimals.js";
+
+export default async function ({ addon }) {
+  function getRegexFromSettings() {
+    let textInInputs = addon.settings.get("textInInputs");
+    return textInInputs ? /^.*$/i : /^[0-9+\-*^/(). ]+$/;
+  }
+  const checkElForTypeChange = (el, focus) => {
```
Joeclinton1, line 8: Instead of using lots of if/elses for each condition to check if to return true or false. Here's some cleaner code you could use:

```js
 const checkElForTypeChange = (el, focus) => {
    if (!el.classList) return false;
    const type = focus === "in" ? "input[type=number]" : "input[type=text]";
    const selectors = [
      "[class*=mediaRecorderPopupContent]",
      "[class*=input_input-form_]",
      "[class*=Popover-body]",
      "[class*=sa-paint-snap-settings]",
      "[class*=sa-onion-settings]",
    ];
    return selectors.some(selector => el.matches(`${selector}${type}`));
  };```

Joeclinton1 (changes requested):
The addon works. I tested dynamic enable/disable that works too.

My main concern is that I think it would be much better to use an existing expression parser and evaluator rather than making your own one. Your current one is limited to simple expressions, and vulnerable to possible bugs that are hard to detect. By relying on an existing well tested one you could increase functionality alot and improve robustness. I recommend using this: https://www.npmjs.com/package/expr-eval it's only 146kb so shouldn't be a problem.

addons/math-in-inputs/userscript.js
Jazza-231: I haven't tested it but from my reasoning of it, this wont work. When we match to "[class*=input_input-form_]" this in itself doesnt return true, but then allows more specific matches to continue. If we don't do this, I *think* it will start allowing math in inputs that we dont want it to, which is how you have implemented it. Currently if any input matches "[class*=input_input-form_]" in your suggested changes, it will immediately return true, which we do NOT want, because there are some input that match this that we do not want to do math in.

Jazza-231:
Does anyone else have opinions on using the external math parser? I feel like it's not entirely necessary but in this case, its not really up to me.

addons/math-in-inputs/userscript.js
Jazza-231: For example, the costume name input (unless we somehow decide that it should do math in ALL inputs no matter how...odd it might seem?)
image

BroJac5246:
> Does anyone else have opinions on using the external math parser? I feel like it's not entirely necessary but in this case, its not really up to me.

As long as it works well and we credit it correctly, why not? This is why it exists.

Samq64:
> I recommend using this: https://www.npmjs.com/package/expr-eval, it's only 146kb so shouldn't be a problem for size.

That's probably fine but are there any other smaller options that do the same job without using eval? A 150kb library is still relatively large for a 6mb extension.

DNin01:
It also just occurred to me, if we add an expression evaluation library, does that mean we could add proper logic to `if` properties in addon manifests, like shown in #6522?

Jazza-231:
Well, I'm open to adding this if its got two good uses, but as I've never used packages before (I am still in the learning stage) I might need instruction on how to properly do it (If you can help, maybe DM me or ping me in the dev server?) Once its set up or whatever I can set it up for this addon, but if we do want the proper logic for the if property, if you want me to do that (in a new pr ofc) ill need a lot of instruction (annoying for someone else, good learning for me) or someone else can do that themself (work for someone else, less learning fot me but still ok)

Joeclinton1:
> > I recommend using this: https://www.npmjs.com/package/expr-eval, it's only 146kb so shouldn't be a problem for size.
> 
> That's probably fine but are there any other smaller options that do the same job without using eval? A 150kb library is still relatively large for a 6mb extension.

Looking at the code I might be wrong but I think bundle.min.js has everything needed. That's only 25.3 kB, which is much more appealing :)

WorldLanguages:
If the maintained library doesn't use eval() / new Function() internally I wouldn't think about it, let's just use it. Does the library you're linking to use it?

Joeclinton1:
> If the maintained library doesn't use eval() / new Function() internally I wouldn't think about it, let's just use it. Does the library you're linking to use it?

The whole purpose of that library is avoid using eval(), so no it doesn't.

WorldLanguages:
Sounds like a great idea to use the library then.

Jazza-231:
> Sounds like a great idea to use the library then.

Did you just volunteer to teach me how to do it?

Jazza-231:
> > Did you just volunteer to teach me how to do it?
> 
> I could be wrong but I think you copy [unpkg.com/expr-eval@2.0.2/dist/bundle.min.js](https://unpkg.com/expr-eval@2.0.2/dist/bundle.min.js) to `/libraries/thirdparty/cs/eval-expr.min.js`, add an entry to `/libraries/README.md` (it's MIT) and then `import {function1, function2} from "/libraries/thirdparty/cs/eval-expr.min.js"` at the top of your script.

I have copied the code, updated the readme (as seen in the latest commit) but i dont actually know how to use it, obviously "function1" and "function2" are placeholders, but i can not read the code at all and so dont know what i do need to import, so if anyone could help that would be great

Jazza-231:
Frankly, that caused a lot more trouble than it did good.
We now have...what 2 more functions? other than that, this library has a lot of issues for our use case, and it also is now impossible to mitigate the floating point errors, because i am NOT touching that...i did the work, up to you all to decide if the % and ! was worth it...

Jazza-231:
> > Did you just volunteer to teach me how to do it?
> 
> I could be wrong but I think you copy [unpkg.com/expr-eval@2.0.2/dist/bundle.min.js](https://unpkg.com/expr-eval@2.0.2/dist/bundle.min.js) to `/libraries/thirdparty/cs/eval-expr.min.js`, add an entry to `/libraries/README.md` (it's MIT) and then `import {function1, function2} from "/libraries/thirdparty/cs/eval-expr.min.js"` at the top of your script.

Oh and for future reference, what I actually had to do was make a script element with the source being the package, and then wait for the script element to load before accessing the function through window

GrahamSH-LLK:
> > > Did you just volunteer to teach me how to do it?
> > 
> > 
> > I could be wrong but I think you copy [unpkg.com/expr-eval@2.0.2/dist/bundle.min.js](https://unpkg.com/expr-eval@2.0.2/dist/bundle.min.js) to `/libraries/thirdparty/cs/eval-expr.min.js`, add an entry to `/libraries/README.md` (it's MIT) and then `import {function1, function2} from "/libraries/thirdparty/cs/eval-expr.min.js"` at the top of your script.
> 
> Oh and for future reference, what I actually had to do was make a script element with the source being the package, and then wait for the script element to load before accessing the function through window

addon.tab.loadScript

Jazza-231:
> > > > Did you just volunteer to teach me how to do it?
> > > 
> > > 
> > > I could be wrong but I think you copy [unpkg.com/expr-eval@2.0.2/dist/bundle.min.js](https://unpkg.com/expr-eval@2.0.2/dist/bundle.min.js) to `/libraries/thirdparty/cs/eval-expr.min.js`, add an entry to `/libraries/README.md` (it's MIT) and then `import {function1, function2} from "/libraries/thirdparty/cs/eval-expr.min.js"` at the top of your script.
> > 
> > 
> > Oh and for future reference, what I actually had to do was make a script element with the source being the package, and then wait for the script element to load before accessing the function through window
> 
> addon.tab.loadScript

Oh. Well that's not the problem that needs to be discussed here

Jazza-231:
> Frankly, that caused a lot more trouble than it did good. We now have...what 2 more functions? other than that, this library has a lot of issues for our use case, and it also is now impossible to mitigate the floating point errors, because i am NOT touching that...i did the work, up to you all to decide if the % and ! was worth it...

@WorldLanguages this is important, and i need to know if its worth it

WorldLanguages:
You can not use a library if you think that's better.

Jazza-231:
Personally I do not think it's worth it

Jazza-231:
Personally I do not think it's worth it

PR #6764, opened 2023-10-12
Revert "Bump version to 1.35.0 stable"

WorldLanguages:
Reverts ScratchAddons/ScratchAddons#6763

What was the point of merging #6744 as part of this release?

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3,7 +3,7 @@
   "name": "__MSG_extensionName__",
   "description": "__MSG_extensionDescription__",
   "version": "1.35.0",
-  "version_name": "1.35.0",
+  "version_name": "1.35.0-prerelease",
   "default_locale": "en",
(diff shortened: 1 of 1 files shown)


DNin01:
> What was the point of merging #6744 as part of this release?

Feeling that change was a bit rushed, perhaps? I think we should just postpone the customizability aspect of that addon (that means reverting https://github.com/ScratchAddons/ScratchAddons/commit/e6465f39bac516c2878943960d50756f46b5a7e3 (#6625)) until v1.36. That'll give us plenty of time to work out a better implementation.

But we can keep the decimal setting type available for if we ever need to use it in the future.

WorldLanguages:
So you suggest reverting the new `editor-number-arrow-keys` settings? Why not keep the string settings?

PR #6767, opened 2023-10-13
Revert "Add and use decimal setting type"

WorldLanguages:
Reverts ScratchAddons/ScratchAddons#6744


DNin01:
https://github.com/ScratchAddons/manifest-schema/commit/151d4b9ce1f71137fc72aa384f607040e738696b still needs to be reverted since we dropped support for decimal setting types as a result of this reversion. Opened a pull request for that: https://github.com/ScratchAddons/manifest-schema/pull/66, but I don't know if we're going to re-add the setting type sooner or later. Another option is to just re-add the setting type even if we don't use it.

PR #6775, opened 2023-10-15
`cloud-games`: Add character limit to `url` setting in table

DNin01:
### Changes

Adds a character limit of 44 characters to the URL string fields in the Cloud games addon. AFAIK, 44 characters is the most you'll ever need:
```
https://               =  8
scratch.mit.edu/  + 16 = 24
projects/         +  9 = 33
1234567890/       + 11 = 44
```

### Reason for changes

I don't know why, it just seems like a good precaution nonetheless.

### Tests

Tested.

### DIFF
--- a/addons/cloud-games/addon.json
+++ b/addons/cloud-games/addon.json
@@ -37,6 +37,7 @@
           "name": "Project/Studio URL",
           "id": "url",
           "type": "untranslated",
+          "max": 44,
           "default": ""
         }
       ],


mxmou (changes requested):
Many project and studio URLs accepted by the addon are longer than that:
* https://scratch.mit.edu/projects/905650648/#comments-360987454 (62 characters)
* https://scratch.mit.edu/studios/33858471/curators (49 characters)
* https://scratch.mit.edu/studios/33858471/comments#comments-237564077 (68 characters)

If a limit is added, it should be longer than the expected length of a URL - maybe 100 characters.

addons/cloud-games/addon.json
```diff
@@ -37,6 +37,7 @@
           "name": "Project/Studio URL",
           "id": "url",
           "type": "untranslated",
+          "max": 44,
```
WorldLanguages, line 40:
```suggestion
          "max": 100,
```

Secret-chest:
Also, let's not forget we're soon getting to 10-digit project IDs

mxmou:
> Also, let's not forget we're soon getting to 10-digit project IDs

The example link in the OP has a 10-digit ID.

addons/cloud-games/addon.json
DNin01: 100 is the default limit.

DNin01:
> Many project and studio URLs accepted by the addon are longer than that:
> 
> * https://scratch.mit.edu/projects/905650648/#comments-360987454 (62 characters)
> * https://scratch.mit.edu/studios/33858471/curators (49 characters)
> * https://scratch.mit.edu/studios/33858471/comments#comments-237564077 (68 characters)
> 
> 
> 
> If a limit is added, it should be longer than the expected length of a URL - maybe 100 characters.

Just because certain URLs are accepted doesn't mean they're expected/intended to be. Such URLs would be cut off anyway and would still technically work cut off.

addons/cloud-games/addon.json
WorldLanguages: Oh

WorldLanguages:
Well, it's true that the URLs would also work cut-off.

apple502j:
Hey, let's add limits to settings it will be fine (says a person whose very commit was reversed for this)

WorldLanguages:
@apple502j Which commit are you referring to?

DNin01:
> @apple502j Which commit are you referring to?

#6744.

DNin01:
Realistically, though, I don't imagine anyone is going to _actually want_ to input a URL longer than 44 characters - the worst-case scenario for direct project page URLs with up to 10 ID digits (when it reaches a billion) as I described above. No one's losing anything by being limited to 44 characters here.

From my testing, longer values are cut off and still work, which means that if a longer URL was accidentally pasted, it would be truncated, saving storage space. That's the only realistic advantage I can think of.

DNin01:
An alternative to this would be parsing the value in the input field to omit `https://scratch.mit.edu/` and anything after the ID.

[https://scratch.mit.edu/<b>projects/905650648</b>/#comments-360987454](https://scratch.mit.edu/projects/905650648/#comments-360987454)

That would save much more space! Just limiting the length like this doesn't help at all the vast majority of the time.

PR #6780, opened 2023-10-16
Fix ghost messages after marking as read (scratch-messaging)

### LINKED ISSUE #6777: `scratch-messaging`: Ghost messages after marking as read
### Description

When marking messages as read, the message counter will go to 0. However, after some time, it'll go back to the count it was before - when checking the messages, it goes back to 0.

### Steps to reproduce

1. Get a message
2. Mark it as read
3. Go out of the popup
4. Wait for a bit
5. See the ghost message
6. Click on the popup
7. See it disappear

Very likely caused by #6728, but unintentional: 

> I made it so after marking messages as read from the popup, the message count is obtained again bypassing the cache, so there should be no ghost messages after reopening the popup.

### Environment

Browser: Edge 118
Scratch Addons version: 1.35.0
Operating system: Windows 11

WorldLanguages:
Resolves #6777

### Changes

- Continues to add URL parameters to the URL when bypassing the cache (after marking messages as read)
- Store last seen `X-Amz-Cf-Id` header in IDB, as an attempt to ignore server-side cached responses to the message count endpoint
- Do not send cookies to the message count endpoint, as it's not necessary

### Tests

Tested in Chromium

### DIFF
--- a/background/message-cache.js
+++ b/background/message-cache.js
@@ -112,20 +112,9 @@ chrome.alarms.onAlarm.addListener(async (alarm) => {
     }
     case BADGE_ALARM_NAME: {
       if (scratchAddons.globalState.auth.isLoggedIn) {
-        const { count, resId } = await MessageCache.fetchMessageCount(scratchAddons.globalState.auth.username);
-        const db = await MessageCache.openDatabase();
-        try {
-          const lastResId = await db.get("countResId", scratchAddons.cookieStoreId);
-          if (lastResId && lastResId === resId) {
-            // No need to replace the count cache. We've already seen this request.
-            // We don't want to override the count as it might potentially be more up-to-date, if cache was recently bypassed by other code.
-            console.log("Ignored cached request for message count endpoint (in alarm).");
-          } else {
-            await db.put("count", count, scratchAddons.cookieStoreId);
-          }
-        } finally {
-          await db.close();
-        }
+        const msgCountData = await MessageCache.fetchMessageCount(scratchAddons.globalState.auth.username);
+        const count = MessageCache.getUpToDateMsgCount(scratchAddons.cookieStoreId, msgCountData);
+        await db.put("count", count, scratchAddons.cookieStoreId);
       }
       await updateBadge(scratchAddons.cookieStoreId);
     }
--- a/background/message-cache.js
+++ b/background/message-cache.js
@@ -113,7 +113,7 @@ chrome.alarms.onAlarm.addListener(async (alarm) => {
     case BADGE_ALARM_NAME: {
       if (scratchAddons.globalState.auth.isLoggedIn) {
         const msgCountData = await MessageCache.fetchMessageCount(scratchAddons.globalState.auth.username);
-        const count = MessageCache.getUpToDateMsgCount(scratchAddons.cookieStoreId, msgCountData);
+        const count = await MessageCache.getUpToDateMsgCount(scratchAddons.cookieStoreId, msgCountData);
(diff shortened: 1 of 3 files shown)


mybearworld:
This doesn't really fix it for me. The ghost messages still appear, but they *don't* go away when going into the popup, instead they just appear normally again.

mybearworld:
> This doesn't really fix it for me. The ghost messages still appear, but they don't go away when going into the popup, instead they just appear normally again. Marking as read a second time works as intended, though.

Oh wait, this also happens in regular SA, but with a ghost message at the end. So, this does work.

WorldLanguages:
@mybearworld Nice to hear it fixed the issue, could you elaborate what you mean though?

mybearworld:
When you click "Mark messages as read", chances are, your messages won't actually be marked as read. Instead, the message count will just be set to 0, and the next time the popup is clicked, the messages will still be there.

mybearworld:
Wait, now it doesn't happen on regular Scratch Addons? This is weird

WorldLanguages:
> the message count will just be set to 0, and the next time the popup is clicked, the messages will still be there.

You say there are situations where the badge doesn't match the amount of displayed messages? That's weird

WorldLanguages:
Also, now that the URL parameter was removed, there might be some server-side caching in play here.

WorldLanguages:
> > the message count will just be set to 0, and the next time the popup is clicked, the messages will still be there.
> 
> You say there are situations where the badge doesn't match the amount of displayed messages? That's weird

That is very weird, because as far as I know, we use the message count to decide how many messages to display! Messages don't have a "read/unread" status, both Scratch itself and Scratch Addons just enumerate the amount of unread messages starting at the top to decide which messages are considered unread.

mybearworld:
I can't replicate a lot of what I commented here anymore for some reason. It's possible that this was a side effect of testing this on both this fork and regular SA in quick succession.

However, what does happen is this:

- Get a message
- Mark it as read
- Go out of the popup
- Go into the popup again
- See the messages there again, **with the message count reflecting the popup**.
- Mark it as read
- ...
- At some point, marking it as read actually works.

This happens **only on this fork, not regular SA.**

WorldLanguages:
Well, it could be server-side caching. @mybearworld Could you try adding `?${Date.now()}` at the end of the `const url` (see lines changed in this PR) and see if that fixes it? (we probably don't want to release a change like that, but it may let us know why you're experiencing that)
Also remember to disable other SA instances and reload the extension

WorldLanguages:
I think I'm going to assume there's server-side caching and release to users. There's little risk it will cause bugs and it shouldn't cause more load to Scratch servers.
If the root cause of the issue mybearworld is experiencing is different we'll need more time to diagnose it.

WorldLanguages:
Actually, instead I'll try to ignore cached responses we've already seen

mybearworld:
Adding `?${Date.now()}` doesn't fix it.

Here's line 78, just in case I did something wrong:

```js
  const url = `https://api.scratch.mit.edu/users/${username}/messages/count?${Date.now()}`;
```

WorldLanguages:
> Adding `?${Date.now()}` doesn't fix it.

Oh... weird

WorldLanguages:
@mybearworld I would also appreciate if you could test this PR (pull my latest commit!)

I recommend removing the extension and loading it unpacked again so that IndexedDB is cleared

mybearworld:
This now works correctly for me!

PR #6786, opened 2023-10-17
Settings page: show addon IDs when hovered

just-a-hriday:
### Changes

I added a tooltip to the addon names which displays the internal id.
<strike>This is supposed to be toggleable but I'm too stupid to understand how the settings work.</strike>

tooltip screenshot from middle-click-popup

### Reason for changes

A lot of addon ids don't do a very good job of describing their purpose, and are very different from the display names.
This makes it a real pain to find addons in the code while contributing, which is why this <strike>toggleable</strike> tooltip showing their ids should be useful.

<sup>I actually thought of this while I was looking for the _Running block border_ addon in the code, and you can imagine how surprised I was when my tooltip told me it was `editor-stepping` - perfect example of why this is useful</sup>

### Tests

Works on Edge 118.
Basically identical code is used in other places in the codebase as well, so it shouldn't be a problem.


webpages/styles/components/tooltips.css
```diff
@@ -17,7 +17,8 @@
   z-index: 10;
   left: 100%;
   margin-inline-start: 10px; /* distance from parent */
-  width: 200px;
+  width: max-content;
+  max-width: 200px;
```
WorldLanguages, line 21: Most likely, this broke *something*

WorldLanguages: Actually, this might be fine, but making all addon IDs a "tooltip" (and changing the HTML structure) broke the loading indicator and possibly more CSS selectors.

just-a-hriday: That sounds pretty bad

just-a-hriday:
You have to get prereleases from github right?

webpages/settings/components/addon-body.html
```diff
@@ -7,7 +7,10 @@
         </div>
         <div class="addon-name">
           <img :src="addonIconSrc" class="icon-type" />
-          <span @click="devShowAddonIds">{{ addon._displayedAddonId || addon.name }}</span>
+          <span class="tooltip">
+            {{ addon.name}}
+            <span v-if="devMode" class="tooltiptext">{{ addon._addonId }}</span>
+          </span>
         </div>
```
mxmou, line 14: To fix the loading indicator:
```html
<div class="addon-name tooltip">
  <img :src="addonIconSrc" class="icon-type" />
  <span>{{ addon.name}}</span>
  <span v-if="devMode" class="tooltiptext">{{ addon._addonId }}</span>
</div>
```

mxmou (approved):

WorldLanguages:
Thanks mxmou for the review. I think CSS issues were the only blocker for this.

WorldLanguages:
I wonder why the tooltips don't show in the small popup-view settings. Is this a side-effect of unrelated CSS code? It would be nicer to explicitly disable these tooltips in the popup view.

By the way, I just tested to make sure `chrome.management.getSelf` works as expected in Firefox, and it does.

mxmou:
> I wonder why the tooltips don't show in the small popup-view settings. Is this a side-effect of unrelated CSS code? It would be nicer to explicitly disable these tooltips in the popup view.

[All tooltips are disabled in the popup.](https://github.com/ScratchAddons/ScratchAddons/blob/ed10b3c169750b67b104afdee48d4f9c2662cfc7/webpages/settings/style.css#L181-L183) If a tooltip that should be visible in the popup is ever added, the selector could be changed to something like
```css
body.iframe .addon-name .tooltiptext,
body.iframe .badge .tooltiptext
```

DNin01:
> @DNin01 As a developer what do you think of this? I guess it will be useful
> 
> In the future we could extend it to setting IDs

I really like the sound of it. Makes it more efficient to tell what an addon's ID is just by finding its name without having to memorize them, which can be helpful when mentioning an addon by ID or trying to find one. I'm sure it will come in really handy for new contributors.

(Wow, I didn't know `chrome.management.getSelf.installType` was a thing!)

Samq64:
Does it make sense to show the tooltip in the popup though? I feel like it would get in the way with longer addon names sometimes.

mxmou:
> Does it make sense to show the tooltip in the popup though? I feel like it would get in the way with longer addon names sometimes.

It currently isn't visible in the popup.

WorldLanguages:
We don't want it to show in the popup. I just said that it is already not visible in the popup, without any new CSS making it explicit, so it's just a side-effect of existing styles for tooltips.

Samq64 (approved):

WorldLanguages (approved):

PR #6787, opened 2023-10-17
New addon: "pop-out stage" (draft)

### LINKED ISSUE #1993: "Pop out stage" (make stage appear in floating browser window/tab)
**Describe the new addon**

Pulling from LLK/scratch-gui#2225: an addon which adds (probably near the fullscreen/presentation/small/normal-stage buttons) a button to make the stage into a separate browser tab. Plus, a button to move the stage back (or simply have the stage reappear when the popup window/tab is closed). This would allow for more particular split-screen views, or moving the stage to a different monitor/display.

I'm not sure how complicated this would be to implement, but I'm guessing it can be made a bit simpler by just making the stage area invisible (but still present) on the main page, and sending the canvas data over to the other tab. There would need to be a way to transfer events on that canvas (i.e. mouse/touch & keyboard input) over to the editor tab (with the VM etc), too!

WorldLanguages:
Resolves #1993

### Changes

- [ ] Stream the project stage canvas into a different tab
- [ ] The project FPS is not throttled when the main editor window is backgrounded
- [ ] General compatibility with `60fps` addon
- [ ] `<mouse down?>` block works properly in the pop-out tab
- [ ] Other mouse events
- [ ] Keyboard events
- [ ] Mouse coordinates
- [ ] Green flag, pause button and stop button from the popout tab
- [ ] Variable and list monitors
- [ ] "Ask-and-wait" prompts

### Tests

Not until it's ready.

### DIFF
--- a/addons/60fps/userscript.js
+++ b/addons/60fps/userscript.js
@@ -1,8 +1,9 @@
+import { polluteRuntimeStart, setVmIntervalDelay, restartStepInterval } from "./vm-stepping-interval-module.js";
+
 export default async function ({ addon, console }) {
   // TODO: test whether e.altKey is true in chromebooks when alt+clicking.
   // If so, no timeout needed, similar to mute-project addon.
 
-  let global_fps = 30;
   const vm = addon.tab.traps.vm;
   let mode = false;
   let monitorUpdateFixed = false;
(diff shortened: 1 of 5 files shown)


DNin01:
Also worth testing `gamepad` with this if applicable.

PR #6791, opened 2023-10-18
Re-add decimal settings + Remove key incremention select settings

### LINKED ISSUE #6769: Remove select settings from `editor-number-arrow-keys`
> Maybe we should also remove the select settings and "use custom values" switch and only have input fields for numbers?

I was actually going to make an issue about that. I think the main reason the select settings were kept was because we didn't have proper decimal settings but now we do.

For the step I suggest decreasing all of them to 0.005 (or maybe 0.001?)

_Originally posted by @Samq64 in https://github.com/ScratchAddons/ScratchAddons/issues/6765#issuecomment-1760610220_

Samq64:
Resolves #6769

### Changes

- Re-adds decimal settings (#6744)
- Sets the setting steps to `any` as suggested in https://github.com/ScratchAddons/ScratchAddons/pull/6744#issuecomment-1751738617
- Removes the select settings

### Reason for changes

It allows full value customization while still being validated by the settings page and simplifies the addon's code.

### Tests

The addon works but transitioning from the select settings to decimal ones doesn't right now.

### DIFF
--- a/addons/editor-number-arrow-keys/addon.json
+++ b/addons/editor-number-arrow-keys/addon.json
@@ -50,6 +50,7 @@
       "type": "decimal",
       "default": 1,
       "min": 0,
+      "max": 100,
       "step": 0.1,
       "if": { "settings": { "useCustom": true } }
     },
@@ -89,6 +90,7 @@
       "type": "decimal",
       "default": 10,
       "min": 0,
+      "max": 100,
       "step": 0.1,
       "if": { "settings": { "useCustom": true } }
     },
@@ -128,6 +130,7 @@
       "type": "decimal",
       "default": 0.1,
       "min": 0,
+      "max": 100,
       "step": 0.01,
       "if": { "settings": { "useCustom": true } }
     },
--- a/addons/editor-number-arrow-keys/addon.json
+++ b/addons/editor-number-arrow-keys/addon.json
@@ -17,73 +17,13 @@
   "settings": [
     {
       "name": "Change on regular key press",
-      "id": "regular",
-      "type": "select",
-      "default": "one",
-      "potentialValues": [
-        {
-          "id": "none",
-          "name": "None"
-        },
-        {
-          "id": "hundredth",
-          "name": "0.01"
-        },
-        {
-          "id": "tenth",
-          "name": "0.1"
-        },
-        {
-          "id": "one",
-          "name": "1"
-        },
-        {
-          "id": "ten",
-          "name": "10"
-        }
-      ],
-      "if": { "settings": { "useCustom": false } }
-    },
-    {
-      "name": "Change on regular key press",
       "id": "regularCustom",
       "type": "decimal",
       "default": 1,
       "min": 0,
       "max": 100,
-      "step": 0.1,
-      "if": { "settings": { "useCustom": true } }
+      "step": "any"
     },
-    {
-      "name": "Change on Shift+Key",
-      "id": "shift",
-      "type": "select",
-      "default": "ten",
-      "potentialValues": [
-        {
-          "id": "none",
-          "name": "None"
-        },
-        {
-          "id": "hundredth",
-          "name": "0.01"
-        },
-        {
-          "id": "tenth",
-          "name": "0.1"
-        },
-        {
-          "id": "one",
-          "name": "1"
-        },
-        {
-          "id": "ten",
-          "name": "10"
-        }
-      ],
-      "if": { "settings": { "useCustom": false } }
-    },
-
     {
       "name": "Change on Shift+Key",
       "id": "shiftCustom",
@@ -91,39 +31,8 @@
       "default": 10,
       "min": 0,
       "max": 100,
-      "step": 0.1,
-      "if": { "settings": { "useCustom": true } }
+      "step": "any"
     },
-    {
-      "name": "Change on Alt+Key",
-      "id": "alt",
-      "type": "select",
-      "default": "tenth",
-      "potentialValues": [
-        {
-          "id": "none",
-          "name": "None"
-        },
-        {
-          "id": "hundredth",
-          "name": "0.01"
-        },
-        {
-          "id": "tenth",
-          "name": "0.1"
-        },
-        {
-          "id": "one",
-          "name": "1"
-        },
-        {
-          "id": "ten",
-          "name": "10"
-        }
-      ],
-      "if": { "settings": { "useCustom": false } }
-    },
-
     {
       "name": "Change on Alt+Key",
       "id": "altCustom",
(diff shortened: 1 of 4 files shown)


addons/editor-number-arrow-keys/addon.json
```diff
@@ -18,122 +18,29 @@
...
-      "id": "regularCustom",
-      "type": "string",
-      "default": "1",
-      "max": 8,
-      "if": { "settings": { "useCustom": true } }
+      "type": "decimal",
+      "default": 1,
+      "min": 0,
+      "max": 100,
+      "step": "any"
```
DNin01, line 25: Currently, users can enter up to 8-digit long numbers and 6 decimal digits (7 might be possible if you use `.` instead of `0.`). Ideally, `max` should be 10,000,000 to remain backwards compatible with what we currently have. In addition, there should be some sort of decimal precision limit if we want to be consistent with what we already have, but step also adds the ability to increment by the set increment with the input's arrow buttons and that would be super slow depending on how precise we get, but maybe that doesn't matter all that much. While I value backwards compatibility, I think we could make the limits just ever so slightly stricter in a way 99.9% of people won't care about, like maybe dropping just one digit of significance.

If I were to suggest a set max and step to try and maximize flexibility while trying to not make stepping with the arrow buttons outright impractical, I'd go with `1,000,000` and `0.0001` respectively. Most people will probably select numbers within that range, right?

...Is there seriously no other way to disable stepping and the arrow buttons? When step is as low as I'm talking about, it doesn't really make sense to me.

Or we could just make incrementing numbers in block inputs not work if the setting value is too precise, with code in the userscript, but that just seems not nice. Or... we could use JS to "validate" the setting values??? If that were the case, I'd personally allow 5 or 6 decimal digits.

DNin01:
I've set the maximum values to 10,000,000 for backwards compatibility. We might adjust it slightly later depending on the discussion.

DNin01:
Oh, you know why this check is failing? It's because step is expected to be a number, but we set it to the string `"any"`.

Secret-chest:
Please allow it to be as precise as possible. There's no reason not to.

Samq64:
Now that I think about it `step` could be added to the integer type but at that point it would be the same as decimal settings since JavaScript only has a Number type. Would it make sense to merge them?

addons/editor-number-arrow-keys/addon.json
Samq64: At that point should the max be removed?

DNin01: > At that point should the max be removed?

Currently, the addon can't work with numbers that are too large.

DNin01:
> Now that I think about it `step` could be added to the integer type but at that point it would be the same as decimal settings since JavaScript only has a Number type. Would it make sense to merge them?

In that case, we could create a "number" type (where we support the step property) to rule them all.

WorldLanguages:
(off-topic, but I can't believe so many keyboard presses around the world were spent to talk about a setting type that is only useful for 1 addon out of 150+ :P)

DNin01:
> The addon works but transitioning from the select settings to decimal ones doesn't right now.

Did some debugging and it looks like the if-then condition on line 167 in `get-addon-settings.js` is corrupting them before the actual migration happens, probably due to the fact that the type changed. I'll attempt a fix now...

DNin01:
Did that fix the problem?

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -116,30 +109,11 @@ export default async function ({ addon }) {
...
-        ? addon.settings.get("shiftCustom")
-        : e.altKey
-        ? addon.settings.get("altCustom")
-        : addon.settings.get("regularCustom");
-      if (settingValue === "") settingValue = 0;
-      let valueAsFloat = parseFloat(settingValue);
-      if (valueAsFloat < 0) valueAsFloat *= -1; // If user typed a negative number, we make it positive
-      if (Number.isNaN(valueAsFloat)) {
-        return;
-      } else if (valueAsFloat === 0 || (valueAsFloat < 100000000 && valueAsFloat > 0.00000099)) {
```
DNin01, line 130: This excluded numbers that were too precise back when the settings were string inputs. Not sure if we still want to do something about that. This safeguard helps prevent accidental precision loss when incrementing by precise values.

Samq64: I assume #6794 will be merged at the same time as this?

Samq64:
I renamed the decimal setting type to just number and will open another PR to 

> > The addon works but transitioning from the select settings to decimal ones doesn't right now.
> 
> Did some debugging and it looks like the if-then condition on line 167 in `get-addon-settings.js` is corrupting them before the actual migration happens, probably due to the fact that the type changed. I'll attempt a fix now...

Thanks! It worked.

addons/editor-number-arrow-keys/userscript.js
DNin01: > I assume #6794 will be merged at the same time as this?

Well, that PR doesn't change the precision of the setting values, but the number inputs seem to automatically truncate them, and using scientific notation makes incrementing throw an error because it can't be converted to BigInt, which just means nothing happens.

DNin01:
You might have noticed I made it so that the string values are converted to numbers, undoing your adjustment. The reason is that the code that was previously breaking migration also has the (positive) consequence of already converting the settings with a string value to numbers (that's why we didn't need to convert them to numbers right as they are migrated), but converting them there 1. simplifies the code and 2. allows the migration code to be independent; if we ever remove the other code block for some reason, we don't want it to break the migration we added (turns out, string values do work just fine from my testing, but it's ideal for them to be numbers).

DNin01 (approved):

DNin01 (approved):
Works well! Since one side of the calculation uses BigInt and the Scratch Addons settings page normalizes entered numbers, encountering precision loss shouldn't be a problem.

Secret-chest:
The ideal thing would be to make the arrows change by the step but still allow other values to be typed.

Samq64:
> The ideal thing would be to make the arrows change by the step but still allow other values to be typed.

The standard HTML `<input>` doesn't seem to allow it and I don't think anyone wants to make a custom component just for that.

DNin01:
> The ideal thing would be to make the arrows change by the step but still allow other values to be typed.

It would be, but it is difficult with the way the HTML `step` property is designed. The most we could probably do is code custom arrow key incrementation for the inputs in the settings page.

Secret-chest:
> The most we could probably do is code custom arrow key incrementation for the inputs in the settings page.

I don't know Vue, but I don't think it would take too much. We can look at `editor-stepping`.

Samq64:
> > The most we could probably do is code custom arrow key incrementation for the inputs in the settings page.
> 
> I don't know Vue, but I don't think it would take too much. We can look at `editor-stepping`.

That could work but that would be for a different PR.

PR #6794, opened 2023-10-20
Use BigInt in `editor-number-arrow-keys`

DNin01:
### Changes

`editor-number-arrow-keys` now supports incrementation with infinitely significant numbers, thanks to BigInt, although I kept a limit of ~~100~~ 30 digits for performance reasons and to help prevent accidentally corrupting codes made of numbers. As it turned out, this was a lot easier than I thought!

Incrementing an extremely precise number with no precision loss

The step increments (the values set in the addon settings) are still limited in precision.

### Reason for changes

There are a couple cases where you might need incrementation with larger or more precise numbers.

### Tests

Tested in Edge 118 to the best of my ability.


DNin01:
A bit lost by the diff? I intentionally split up my changes into two commits - try viewing them one at a time.

addons/editor-number-arrow-keys/userscript.js
```diff
@@ -44,46 +44,33 @@ export default async function ({ addon }) {
...
 
-    return (isNumberNegative ? "-" : "") + normalizedNum;
+    return numStr ? (isNumberNegative ? "-" : "") + numStr : 0;
   };
 
   const isValidNumber = (numStr) => {
-    if (numStr.length > 10) return false;
-    if (amountOfDecimals(numStr) > 5) return false;
-    return normalizeNumber(numStr) === Number(numStr).toString();
+    if (numStr.length > 100) return false;
```
DNin01, line 67: You know, are 100-digit numbers really fine-tuned every day? Not sure if the limit needs to be _this_ high. And such numbers will still lose precision when used in the project anyway. Maybe somewhere closer to 30 would be sensible?

WorldLanguages:
Nice! Make sure to test all the edge cases again...

mxmou (approved):

Samq64:
This line:
```js
} else if (valueAsFloat === 0 || (valueAsFloat < 100000000 && valueAsFloat > 0.00000099)) {
```
looks like it wasn't updated to support more than 8 characters. Was it forgotten or will that be done when the maximum setting value is changed.

DNin01:
> This line:
> 
> ```js
> } else if (valueAsFloat === 0 || (valueAsFloat < 100000000 && valueAsFloat > 0.00000099)) {
> ```
> 
> looks like it wasn't updated to support more than 8 characters. Was it forgotten or will that be done when the maximum setting value is changed.

Yeah, I decided not to make any changes to the precision of setting values. It's fine, pretty much nobody is going to be incrementing by units more precise than that. We would also need to use a string-type setting for that to be able to be infinitely precise, which we already do, but we're considering [changing it](https://github.com/ScratchAddons/ScratchAddons/pull/6791).

Samq64 (approved):

PR #6799, opened 2023-10-21
`copy-message-link`: fix incompatibility, remove unnecessary css

just-a-hriday:
### Changes

Removes unnecessary stylesheet, and removes the displayNoneWhileDisabled thing.
Also has the unfortunate side effect of no longer removing the _copy link_ button when the addon is disabled. I can't imagine why.

- [x] Fix displayNoneWhileDisabled thing

### Reason for changes

The stylesheet is completely useless - the only reason it's needed is because of the displayNoneWhileDisabled thing. Since that modifies the display property, it means that you can't do display:none while the mouse isn't hovered over the buttons. So you need to make them transparent instead, which makes them retain their dimensions. This causes conflict with `emoji-picker` (and possibly something else too) by being visible on top of the emoji picker dialogs.

<sub>I can't test now, and I won't be able to check github until tuesday</sub>

### DIFF
--- a/addons/copy-message-link/profile.js
+++ b/addons/copy-message-link/profile.js
@@ -2,40 +2,47 @@ export default async function ({ addon, console, msg }) {
   addon.self.addEventListener("disabled", () => {
     document.querySelectorAll(".sa-copy-link-btn").forEach(element => element.remove());
   })
-  
-  let amtOfComments = 0;
-  let pass = 0;
-  while (true) {
-    const newAmtOfComments = document.querySelectorAll("div.comment").length;
-    if (amtOfComments !== newAmtOfComments) {
-      pass++;
-      amtOfComments = newAmtOfComments;
+  addon.self.addEventListener("reenabled", () => {
+    createButtons;
+  })
+
+  createButtons();
+
+  function createButtons() {
+    let amtOfComments = 0;
+    let pass = 0;
+    while (true) {
+      const newAmtOfComments = document.querySelectorAll("div.comment").length;
+      if (amtOfComments !== newAmtOfComments) {
+        pass++;
+        amtOfComments = newAmtOfComments;
+      }
+      const comment = await addon.tab.waitForElement(`div.comment:not([data-sa-copy-link-pass='${pass}'])`);
+      comment.dataset.saCopyLinkPass = pass;
+      if (comment.querySelector(".sa-copy-link-btn")) {
+        // This will to all comments after posting a new one,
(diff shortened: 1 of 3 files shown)


WorldLanguages:
> The stylesheet is completely useless - the only reason it's needed is because of the displayNoneWhileDisabled thing. Since that modifies the display property, it means that you can't do display:none while the mouse isn't hovered over the buttons.

I think it's okay to use !important in that case, according to the docs: https://scratchaddons.com/docs/reference/addon-api/addon.tab/#using-addontabdisplaynonewhiledisabled-dynamicdisable

WorldLanguages:
Maybe the solution is to restore the stylesheet, only target the "copy comment link" button, and use `display: none !important`

just-a-hriday:
> Maybe the solution is to restore the stylesheet, only target the "copy comment link" button, and use `display: none !important`

If you put !important on display:none, then it'll override whatever you put in the ::hover styles.

WorldLanguages:
What about adding `!important` to both?

just-a-hriday:
I guess that should work because `::hover` is a more specific selector.

WorldLanguages:
Yes, you could opt to do that

just-a-hriday:
How?

addons/copy-message-link/profile.js
```diff
@@ -1,4 +1,8 @@
 export default async function ({ addon, console, msg }) {
+  addon.self.addEventListener("disabled", () => {
+    document.querySelectorAll(".sa-copy-link-btn").forEach(element => element.remove());
+  })
+
```
WorldLanguages, line 5: This is not enough
You need to make sure the "copy link" button is no longer added if the addon is disabled (boolean `addon.self.disabled`) and also check that reenabling the addon adds the buttons back.

just-a-hriday: > You need to make sure the "copy link" button is no longer added if the addon is disabled

What do you mean? The buttons are deleted when it's disabled and they don't come back.

> and also check that reenabling the addon adds the buttons back

I'll fix that real quick

WorldLanguages: > and they don't come back.

Have you tried loading more pages of comments after disabling the addon?

just-a-hriday: Just tried, works perfectly

just-a-hriday:
> What about adding `!important` to both?

Clearly this is the better solution, but I'm having trouble reverting my commits so I'll create a new PR which does this.

PR #6802, opened 2023-10-22
Fix addon userscripts using wrong language for localized strings

WorldLanguages:
### Changes

When there's no `scratchlanguage` cookie, Scratch uses `navigator.language` as the language, and we want to match the language used by Scratch.

Remember that in modern browsers, `navigator.language = navigator.languages[0]` which doesn't necessarily match with the browser UI language, which is used for the popup and settings page through `chrome.i18n`.

### Tests

Tested in Chromium


DNin01 (approved):
Haven't tested but looks good to me.

PR #6809, opened 2023-10-26
Better-arrow-key-incrementation

Jazza-231:
Just allows the editor-number-arrow-keys addon work with the colour picker inputs, paint snapping inputs, and onion skinning inputs. Used to be a part of the math-in-inputs pr but it is not separated.

### Changes

See above. + Does include changes I made to the way it checks through the elements, meaning that the math-in-inputs pr has to be merged first and then this (I think?) (this addon will NOT work if just the math-in-inputs is merged and this pr isnt.)

### Reason for changes

Consistency, even if it's arguable not that useful, it should work in absolutely every input.

### Tests

Well when this was a part of my original changes it all works so yeah.


Jazza-231:
Crap I created this branch from the math-in-inputs branchhhh ignore this ill try again

PR #3960, opened 2021-12-20
Improve Scratch 2.0 → 3.0 addon

mxmou:
Resolves #642
Resolves #1699
Resolves #3136
Resolves #1092
Resolves #3733
Related to #3072

### Changes

* Changes the border radius of most buttons to 8px to match scratch-www.
* Makes most white buttons either blue or gray
* Makes the footer and profile comments much more similar to their scratch-www equivalents. (Is there a good way to make the avatar outlines affected by the addon for customizing them?)
* Improves the top part of the profile page:
image
image
* Replaces the outdated box design of forum posts with scratch-www style comment bubbles:
image
* scratch-www style pagination
* Replaces weird navigation in My Stuff and Account Settings with this:
image
* Makes the pages that appear when clicking various "View All" links more similar to the search page.
* Changes the user-id addon slightly and adds dynamic enable/disable to it.
* Makes loading spinners look like the scratch-www ones.

### Reason for changes

Consistency with scratch-www

### Tests

To make testing loading spinners easier:
* Comments: go to https://scratch.mit.edu/users/gdpr0000001/
* My Stuff: block the request to /site-api/projects/all/​
* posting a comment, following/unfollowing a forum topic: create a network throttling profile with maximum latency.

PR #6811, opened 2023-10-27
Minor `emoji-picker` bugfix

just-a-hriday:
### Changes

Just adds a few `!important`s in the stylesheet to fix the margins below the show more emojis button on profile pages.

### Reason for changes

image
<sup>Weird margins</sup>

### Tests

Tested in Edge 118.

---

<strike>Also since I've opened a PR for emoji-picker anyways, do you guys think it would be a good idea to replace the emoji on the button with one of the cat emojis?</strike> This is way too much work, I'm sure someone will get to it later.

### DIFF
--- a/addons/emoji-picker/userstyle.css
+++ b/addons/emoji-picker/userstyle.css
@@ -9,7 +9,6 @@
   box-shadow: 0px 0px 5px #00000022;
   z-index: 1;
   padding: 5px;
-  padding-bottom: 0;
   overflow: hidden;
   line-height: 0px;
   font-size: 15px;
@@ -88,12 +87,11 @@
 .compose-comment .compose-bottom-row .sa-emoji-picker-see-more {
   width: 100%;
   margin-top: 5px;
-  margin-bottom: 5px;
 }
 #comments #comment-form .sa-emoji-button-r2 .sa-emoji-picker-see-more,
 #comments .comment form .sa-emoji-button-r2 .sa-emoji-picker-see-more {
   width: calc(100% - 2px);
-  margin-bottom: 7px;
+  margin-bottom: 2px;
   margin-top: 4px;
 }
 
--- a/addons/emoji-picker/userstyle.css
+++ b/addons/emoji-picker/userstyle.css
@@ -85,15 +85,16 @@
   width: 20px;
   height: 20px;
 }
-.sa-emoji-picker-see-more {
+.compose-comment .compose-bottom-row .sa-emoji-picker-see-more {
   width: 100%;
-  margin-top: 5px !important;
-  margin-bottom: 5px !important;
+  margin-top: 5px;
+  margin-bottom: 5px;
 }
-.sa-emoji-button-r2 .sa-emoji-picker-see-more {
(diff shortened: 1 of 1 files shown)


WorldLanguages:
> Just adds a few `!important`s in the stylesheet.

Can you clarify why? What other styles is this overriding?

WorldLanguages:
> do you guys think it would be a good idea to replace the emoji on the button with one of the cat emojis?

I think it would be fine, as long as it's clear it's an emoji

just-a-hriday:
> > Just adds a few `!important`s in the stylesheet.
> 
> Can you clarify why? What other styles is this overriding?

Edited the post.

WorldLanguages:
But why is !important needed in the first place? What other styles win if we don't include it? CSS from other addons or vanilla Scratch CSS? If it's another addon please mention which one it is

just-a-hriday:
> But why is !important needed in the first place? What other styles win if we don't include it? CSS from other addons or vanilla Scratch CSS? If it's another addon please mention which one it is

Lemme check...
Turns out there's no bottom margin in the userstyle at all, so I had to add that. The !important on it is to stop it getting overridden by vanilla scratch styles.
The top margins were also getting overridden by vanilla scratch.

addons/emoji-picker/userstyle.css
```diff
@@ -87,12 +87,13 @@
 }
 .sa-emoji-picker-see-more {
   width: 100%;
-  margin-top: 5px;
+  margin-top: 5px !important;
+  margin-bottom: 5px !important;
 }
 .sa-emoji-button-r2 .sa-emoji-picker-see-more {
   width: calc(100% - 2px);
   margin-bottom: 7px;
```
MaterArc, line 95: Hmm If you test it out, all you need to do it play with this value instead of adding a whole new property for `margin-top` and `margin-bottom`. In my opinion, the top margin doesn't really need to be changed, all you really need is to make the box a tad bit bigger to accomodate the button, which can be done without adding new lines to the file.

MaterArc: That way, you wouldn't need the `!important` property, either. Since none of it gets overridden by vanilla scratch, in my testing.

mxmou (changes requested):
Increasing the specificity of selectors is usually better than adding `!important`.

MaterArc:
> Increasing the specificity of selectors is usually better than adding `!important`.

Yeah this change is better than all of ours. I agree though.

mxmou:
> But why is !important needed in the first place? What other styles win if we don't include it? CSS from other addons or vanilla Scratch CSS? If it's another addon please mention which one it is

Vanilla Scratch on project and studio pages, Scratch 2.0 → 3.0 on profile pages. The problematic CSS was added in #3960 and in scratchfoundation/scratch-www#6646.

DNin01:
More about specificity: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

just-a-hriday:
Sure, not a problem with me - I'm not really working on this kind of stuff these days.

Samq64 (approved):

mxmou (approved):

PR #6812, opened 2023-10-27
Allow stage header to overflow to a side

### LINKED ISSUE #3030: Moving stage
Hi! I found a bug.

When several addons (clones counter, mouse pointer coordinates, gamepad support, debugger) and "30 FPS with interpolation" are activated, the stage move when the mouse pointer move (because of coordinates text).

It works on TurboWarp and Scratch.

https://user-images.githubusercontent.com/80577386/125785330-fa0d24dd-4145-4ee0-a6e0-83edced2b360.mp4

WorldLanguages:
Resolves #3030

### Changes

Adds the CSS styles suggested by mxmou (in comment https://github.com/ScratchAddons/ScratchAddons/issues/3030#issuecomment-1657084455) when using appendToSharedSpace with the stageHeader space.

image

Having `appendToSharedSpace` inject styles on its own is not ideal, but the overflow bug is bad enough on its own. We can find better ways to handle shared CSS between addons in the future.

Also worth discussing: now that's space, the debugger and gamepad buttons can stay visible even if the stage is in small-mode. We currently hide them to save space, but there's enough space now.

### Tests

- [ ] Test in RTL
- [ ] Test with addons enabled such as `editor-stage-left` and `editor-buttons-reverse-order` (in LTR)
- [ ] Item above but RTL

### DIFF
--- a/addon-api/content-script/stage-header-styles.js
+++ b/addon-api/content-script/stage-header-styles.js
@@ -1,0 +1,20 @@
+let injected = false;
+
+export function applyStyles() {
+  if (injected) return;
+  injected = true;
+
+  const css = `
+/* appendToSharedSpace CSS for stageHeader space - see SA/SA#3030 */
+[class*="stage-header_stage-header-wrapper_"] {
+    height: 2.75rem;
+}
+  
+[class*="stage-header_stage-menu-wrapper_"] {
+    position: absolute;
+    right: 0;
+    min-width: 100%;
+}
+`.trim();
+  document.head.appendChild(Object.assign(document.createElement("style"), { textContent: css }));
+}
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -632,6 +632,13 @@ export default class Tab extends Listenable {
       },
     };
 
+    if (space === "stageHeader") {
+      // Load stageHeader space CSS
+      import("./stage-header-styles.js").then((m) => {
+        m.applyStyles();
+      });
+    }
+
     const spaceInfo = sharedSpaces[space];
     const spaceElement = spaceInfo.element();
     if (!spaceElement) return false;


addon-api/content-script/stage-header-styles.js
```diff
@@ -0,0 +1,20 @@
...
+
+  const css = `
+/* appendToSharedSpace CSS for stageHeader space - see SA/SA#3030 */
+[class*="stage-header_stage-header-wrapper_"] {
+    height: 2.75rem;
+}
+  
+[class*="stage-header_stage-menu-wrapper_"] {
+    position: absolute;
+    right: 0;
```
mxmou, line 15: This won't work in RTL.

mxmou (commented):
This is better than the current behavior, but still doesn't fully fix the overflow issue. Some controls can be covered by the find bar in a small window:
image

The stage header still overflows if `editor-stage-left` is enabled. I'm not sure how that could be fixed - with a scrollbar?
image

mxmou:
> We can find better ways to handle shared CSS between addons in the future.

Maybe a CSS file in `libraries/common/cs` that's imported by all affected addons, like the existing [spinner.css](https://github.com/ScratchAddons/ScratchAddons/blob/master/libraries/common/cs/spinner.css).

Secret-chest:
Make sure all the dynamic items have a fixed width to prevent layout shifting.

WorldLanguages:
> The stage header still overflows if editor-stage-left is enabled. I'm not sure how that could be fixed - with a scrollbar?

Uhh, yeah I forgot that in that case there's no more space to use...

Maybe we should first try to find the best way to use the limited space (such as shrinking the turbo mode indicator, etc.), and think of this overflow as just a way of showing more UI buttons in some situations where it's possible. I think I'll close this PR

PR #5629, opened 2023-01-16
Add support for explore page to infinite-scroll

Jazza-231:
Resolves #2520

### Changes
In addition to the new setting for the explore page, I reorganized the studio settings to be in the same order as the tabs on scratch, and fixed the addon spamming the messages API for no reason. 
(This is my first time making a pr or anything to do with GitHub so please say if I do anything wrong)

### Reason for changes
The suggestion to add infinite scrolling to the explore page got 31 upvotes on the SA discord
(Also redguy said it would be easy)

### Tests
Tested every setting again after changes, on chromium 109.0.5414.87. The network log shows no API spamming.

PR #6813, opened 2023-10-27
Infinite scrolling: support Explore and Search pages

### LINKED ISSUE #2520: Infinite scrolling in Explore

Yes! Have you ever just went to a project on the explore tab, then go back to see you're at the top of the page again? It sucks for me.

Infinite scrolling on the Explore page to avoid losing where you were on the page.

n/a

### Additional context

I checked and this shouldn't be a dupe. ✌️

WorldLanguages:
Superseeds #5629
Resolves #2520
Included at #5419

### Changes

Adds infinite scrolling support to project/studio search and project/studio explore.

PR #5629 also attempted this, but it sent too many requests. Scratch's "load more" button sends one request each time it's clicked, and we don't want to be sending more than one request concurrently. This is not the case in other infinite scrolling areas where the button is disabled/hidden while a new page of content is loading.

### Tests

Tested in Chromium

### DIFF
--- a/addons/infinite-scroll/addon.json
+++ b/addons/infinite-scroll/addon.json
@@ -63,6 +63,12 @@
       "id": "messageScroll",
       "type": "boolean",
       "default": true
+    },
+    {
+      "name": "Search and Explore",
+      "id": "searchExploreScroll",
+      "type": "boolean",
+      "default": true
     }
   ],
   "userscripts": [
@@ -79,7 +85,9 @@
         "projects",
         "https://scratch.mit.edu/studios/*",
         "https://scratch.mit.edu/users/*",
-        "https://scratch.mit.edu/messages"
+        "https://scratch.mit.edu/messages",
+        "https://scratch.mit.edu/search/*",
+        "https://scratch.mit.edu/explore/*"
       ]
     }
   ],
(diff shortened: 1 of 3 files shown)


WorldLanguages:
I must admit buttonScroll.js is a bit messy. Studios have a different way to fix this using event stopPropagation.

Also, I wonder if other already-supported pages are also doing unnecessary fetches and we haven't noticed. I mentioned this at https://github.com/ScratchAddons/ScratchAddons/pull/5629#issuecomment-1401067425.

WorldLanguages:
Worth checking there aren't any project/studio duplicates when scrolling. There shouldn't be, but there might be a bug.

WorldLanguages:
> Worth checking there aren't any project/studio duplicates when scrolling

Looks like that's a Scratch issue caused by caching Explore aggressively (server-side cached for 12 hours)

WorldLanguages:
I'm pretty sure this works like it's supposed to, but can I get an approval on the manifest changes?

addons/infinite-scroll/addon.json
```diff
@@ -63,6 +63,12 @@
       "id": "messageScroll",
       "type": "boolean",
       "default": true
+    },
+    {
+      "name": "Search and Explore",
+      "id": "searchExploreScroll",
+      "type": "boolean",
+      "default": true
```
WorldLanguages, line 71: Enabling this new setting by default sounds reasonable.

PR #6814, opened 2023-10-28
`copy-message-link`: fix `emoji-picker` incompatibility

just-a-hriday:
Continuation of #6799, I messed that one up and couldn't revert the commits so here we are.

### Changes

Fixes incompatibility with `emoji-picker`.

### Reason for changes

Since the displayNoneWhileDisabled thing modifies the display property of the button, you can't do display:none while the mouse isn't hovered over the buttons. So you need to make them transparent instead, which makes them retain their dimensions. This causes conflict with `emoji-picker` (and possibly something else too) by being visible on top of the emoji picker dialogs.
screenshot of the bug

---

Works on Edge 118.

### DIFF
--- a/addons/copy-message-link/profile.css
+++ b/addons/copy-message-link/profile.css
@@ -1,7 +1,6 @@
 #comments .comment .report {
-  display: block;
-  opacity: 0;
+  display: none !important;
 }
 #comments .comment:hover .report {
-  opacity: 1;
+  display: block !important;
 }


Samq64:
> This causes conflict with emoji-picker (and possibly something else too) by being visible on top of the emoji picker dialogs.

This only seems happen on profile pages with `scratchr2` enabled. This solution is fine but setting the z-index could also work.

mxmou:
This doesn't fully fix the issue - the buttons are still visible if the mouse is over the comment but outside the emoji picker popup. Adding a z-index to the popup like @Samq64 suggested would be a better solution.

DNin01:
I opened a pull request: #6890

This bug actually doesn't depend on `copy-message-link`.

WorldLanguages:
We just merged #6890, does this PR still make sense?

mxmou:
> We just merged #6890, does this PR still make sense?

I don't think so - the bug that this PR tries to fix was fixed by #6890.

PR #6817, opened 2023-10-28
Rename speed options

### LINKED ISSUE #6352: Change how animation speed settings in addons are shown to the user
### Why this would be helpful

Our current way of allowing animation speeds to be customized is by offering 4 speed settings: "Slow", "Default", "Quick", and "Instant". This works great, but I think in the future with design changes like #6317, it will make more sense if this changes.

In the issue I just mentioned, I changed the design of the single-select input to a slider with three tick marks indicating the possible switch states. Each one would correspond to one of the settings above.

Also, it just doesn't feel quite right to have the slider or the single-select input be able to go from Quick to Instant. I don't know if it's just because it makes the slider feel less linear or if it's because Instant is not really an animation speed.

Also, at one point I renamed these settings, and [this](https://github.com/DNin01/ScratchAddons/commit/e390425cfeb621d345c73aa461756791b9bcc3c0#diff-9b4b42b4716b30db3810c308df36a6633c39fd1e17336d408d77365efd120149R89-R109) is one thing I did. Yup, they used to be "Long", "Default", "Short" instead of "Slow", "Default", "Quick". This is okay and I like the change from it being called a duration to a speed, but in general, "Fast" is used more often than "Quick", so I think that could be changed. Also, we might want to change "Default" to "Normal" or "Medium".

### My suggestion

So here are the changes we can make:
- Split up the Instant setting into its own on/off switch called "Enable animation".
- Use a slider instead of buttons.
- Rename the speed settings to "Slow", "Normal", and "Fast".

### Possible alternatives

Any ideas?

### Additional context

_No response_

DNin01:
Resolves #6352?

### Changes

Changes animation speed settings' names as follows:

- "Instant", "Quick", "Default", "Slow" to "Off", "Slow", "Normal", "Fast" (which is also swapped from before)
- "Animation speed" to "Animation" wherever "Off" is an option
- `forum-live-preview`:
  - "Refresh rate" to "Refresh delay"
  - "Slow", "Default", "Fast" to "Short", "Default", "Long" (which is also swapped from before)

(Isn't it something that for all of these, whether they are durations or speeds are now the opposite from what they were originally?)

### Reason for changes

To make them sound more natural.

### Tests

Tested on Edge 118.

### DIFF
--- a/addons/forum-live-preview/addon.json
+++ b/addons/forum-live-preview/addon.json
@@ -24,21 +24,21 @@
   ],
   "settings": [
     {
-      "name": "Refresh rate",
+      "name": "Refresh delay",
       "id": "rate",
       "type": "select",
       "potentialValues": [
         {
-          "id": "slow",
-          "name": "Slow"
+          "id": "fast",
+          "name": "Short"
         },
         {
           "id": "default",
           "name": "Default"
         },
         {
-          "id": "fast",
-          "name": "Fast"
+          "id": "slow",
+          "name": "Long"
         }
       ],
       "default": "default"
--- a/addons/collapse-footer/addon.json
+++ b/addons/collapse-footer/addon.json
@@ -34,16 +34,16 @@
       "type": "select",
       "potentialValues": [
         {
-          "id": "short",
-          "name": "Quick"
+          "id": "long",
+          "name": "Slow"
         },
         {
           "id": "default",
-          "name": "Default"
+          "name": "Normal"
         },
         {
-          "id": "long",
-          "name": "Slow"
+          "id": "short",
+          "name": "Fast"
         }
       ],
       "default": "default"
--- a/addons/custom-zoom/addon.json
+++ b/addons/custom-zoom/addon.json
@@ -62,25 +62,25 @@
       "default": false
     },
     {
-      "name": "Autohide Animation Speed",
+      "name": "Autohide Animation",
       "id": "speed",
       "type": "select",
       "potentialValues": [
         {
           "id": "none",
-          "name": "Instant"
-        },
-        {
-          "id": "short",
-          "name": "Quick"
-        },
-        {
-          "id": "default",
-          "name": "Default"
+          "name": "Off"
         },
         {
           "id": "long",
           "name": "Slow"
+        },
+        {
+          "id": "default",
+          "name": "Normal"
+        },
+        {
+          "id": "short",
+          "name": "Fast"
         }
       ],
       "default": "default",
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -86,25 +86,25 @@
       "default": "cathover"
     },
     {
-      "name": "Animation speed",
+      "name": "Animation",
       "id": "speed",
       "type": "select",
       "potentialValues": [
         {
           "id": "none",
-          "name": "Instant"
-        },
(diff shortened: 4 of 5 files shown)


Secret-chest:
Please add the no animation option in collapse footer.

DNin01:
> Please add the no animation option in collapse footer.

Maybe in a different PR.

Secret-chest:
> Please add the no animation option in collapse footer.
> 
> Maybe in a different PR.

Consistency

addons/forum-live-preview/addon.json
```diff
@@ -24,21 +24,21 @@
...
       "potentialValues": [
         {
-          "id": "slow",
-          "name": "Slow"
+          "id": "fast",
+          "name": "Short"
         },
         {
           "id": "default",
           "name": "Default"
```
Samq64, line 37: Is there a reason this wasn't also changed to normal?

Samq64:
> Please add the no animation option in collapse footer.

Maybe a fade? The problem is the footer can take up half the window when expanded so it can be pretty jarring to accidentally hover the cursor over it. I know this is also kind of a problem as elsewhere too but at least the areas are smaller.

Samq64:
> I thought we had renamed these already 😅

It looks like it was just `hide-flyout`?
- #5245

Secret-chest:
> Maybe a fade? The problem is the footer can take up half the window when expanded so it can be pretty jarring to accidentally hover the cursor over it. I know this is also kind of a problem as elsewhere too but at least the areas are smaller.

A fade would have the same problems.

addons/forum-live-preview/addon.json
DNin01: It's a different kind of "Default"; all the other options are speeds, except for these ones which are phrased like durations. I'm still thinking about it.

DNin01:
> > Maybe a fade? The problem is the footer can take up half the window when expanded so it can be pretty jarring to accidentally hover the cursor over it. I know this is also kind of a problem as elsewhere too but at least the areas are smaller.
> 
> A fade would have the same problems.

Same with having it pop into existence with no animation…

But there's no motion in a fade animation, so it should be a little less jarring.

But it would be nice to talk about this elsewhere. I'll plan to create an issue.

DNin01:
> I thought we had renamed these already 😅 

Yeah, we did. And we're doing it again…

Samq64 (approved):

WorldLanguages:
> > I thought we had renamed these already 😅
> 
> Yeah, we did. And we're doing it again…

As long as there's consensus on the new names, it's fine :P

addons/forum-live-preview/addon.json
mxmou: I don't think this distinction makes sense - "Default" and "Normal" could both refer to a speed or a duration.

mxmou:
> **"Instant", "Quick", "Default", "Slow"** to **"Off", "Slow", "Normal", "Fast"** (which is also swapped from before) `4 files`

I like the new names but not the order - "fast" being next to "off" is the more natural choice.

WorldLanguages:
> I like the new names but not the order - "fast" being next to "off" is the more natural choice.

Reminds me of https://electronics.stackexchange.com/questions/404819/fan-speed-switch-why-off-high-med-low 😄

PR #6819, opened 2023-10-29
Create "Text-to-speech subtitles" addon

### LINKED ISSUE #5483: TTS Subtitles addon
### Why this would be helpful

Many projects do not have subtitles for the deaf or hard of hearing, making those projects inaccessible to those who can't hear or even just turned their volume down.

### How this addon works

This addon would be able to help in projects using the Text to Speech extension. It would show subtitles below the project player for anything the TTS voice says.

A mockup of the subtitles area

### Possible alternatives

_No response_

### Additional context

_No response_

mybearworld:
Resolves #5483

### Changes

Create an addon called "Text-to-speech subtitles" which adds subtitles to projects using text to speech. This is done by trapping calls to `Promise.race`, which the "speak" block uses.

The Scratch project page with subtitles below the project
The Scratch editor with subtitles below the stage

### Reason for changes

Accessibility - it's also a relatively widely requested feature on Scratch

### Tests

Tested in Edge.

### DIFF
--- a/addons/tts-subtitles/addon.json
+++ b/addons/tts-subtitles/addon.json
@@ -1,0 +1,31 @@
+{
+  "name": "Text-to-speech subtitles",
+  "description": "Adds subtitles to every project that uses text-to-speech.",
+  "tags": ["community", "projectPage"],
+  "userscripts": [
+    {
+      "matches": ["projects"],
+      "url": "userscript.js"
+    }
+  ],
+  "userstyles": [
+    {
+      "matches": ["projects"],
+      "url": "userstyle.css"
+    }
+  ],
+  "versionAdded": "1.36.0",
+  "info": [
+    {
+      "id": "add-subtitles",
+      "text": "Only people that have this addon enabled will be able to see these subtitles. In your projects, make sure to add subtitles yourself, so that anyone can use these subtitles.",
+      "type": "notice"
+    }
+  ],
+  "credits": [
+    {
+      "link": "https://scratch.mit.edu/users/mybearworld",
+      "name": "mybearworld"
+    }
+  ]
+}
--- a/addons/tts-subtitles/addon.json
+++ b/addons/tts-subtitles/addon.json
@@ -14,6 +14,14 @@
       "url": "userstyle.css"
     }
   ],
+  "settings": [
+    {
+      "id": "show-in-editor",
+      "name": "Show subtitles in editor",
+      "type": "boolean",
+      "default": false
+    }
+  ],
   "versionAdded": "1.36.0",
   "info": [
     {
--- a/addons/tts-subtitles/addon.json
+++ b/addons/tts-subtitles/addon.json
@@ -35,5 +35,7 @@
       "link": "https://scratch.mit.edu/users/mybearworld",
       "name": "mybearworld"
     }
-  ]
+  ],
+  "dynamicDisable": true,
+  "dynamicEnable": true
 }
--- a/addons/tts-subtitles/userscript.js
+++ b/addons/tts-subtitles/userscript.js
@@ -3,6 +3,7 @@ export default async function ({ addon, console }) {
   const createSubtitleBox = async () => {
     subtitleBox?.remove?.();
     if (
+      addon.self.disabled ||
       (addon.tab.editorMode !== "projectpage" && addon.tab.editorMode !== "editor") ||
       (addon.tab.editorMode === "editor" && !addon.settings.get("show-in-editor"))
     ) {
@@ -17,6 +18,8 @@ export default async function ({ addon, console }) {
     player.insertAdjacentElement("afterend", box);
     subtitleBox = box;
   };
+  addon.self.addEventListener("disabled", createSubtitleBox);
+  addon.self.addEventListener("reenabled", createSubtitleBox);
   addon.settings.addEventListener("change", createSubtitleBox);
   addon.tab.addEventListener("urlChange", createSubtitleBox);
   createSubtitleBox();
--- a/addons/tts-subtitles/userscript.js
+++ b/addons/tts-subtitles/userscript.js
@@ -1,7 +1,11 @@
 export default async function ({ addon, console }) {
   let subtitleBox;
   const createSubtitleBox = async () => {
-    if (addon.tab.editorMode !== "projectpage" && addon.tab.editorMode !== "editor") {
+    subtitleBox?.remove?.();
+    if (
+      (addon.tab.editorMode !== "projectpage" && addon.tab.editorMode !== "editor") ||
+      (addon.tab.editorMode === "editor" && !addon.settings.get("show-in-editor"))
+    ) {
       return null;
     }
     const player =
@@ -11,12 +15,11 @@ export default async function ({ addon, console }) {
     const box = document.createElement("div");
     box.classList.add("sa-tts-subtitles-box");
     player.insertAdjacentElement("afterend", box);
-    return box;
+    subtitleBox = box;
   };
-  addon.tab.addEventListener("urlChange", async () => {
-    subtitleBox = await createSubtitleBox();
-  });
-  subtitleBox = await createSubtitleBox();
+  addon.settings.addEventListener("change", createSubtitleBox);
+  addon.tab.addEventListener("urlChange", createSubtitleBox);
+  createSubtitleBox();
 
   const addToSubtitlesBox = (text) => {
     if (!subtitleBox) {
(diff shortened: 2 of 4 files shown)


addons/tts-subtitles/userscript.js
```diff
@@ -0,0 +1,41 @@
...
+  const addToSubtitlesBox = (text) => {
+    if (!subtitleBox) {
+      return;
+    }
+    subtitleBox.textContent += text + "\n";
+    subtitleBox.scrollTop = subtitleBox.scrollHeight;
+  };
+
+  // It needs to be on window.Promise for some reason
+  // Otherwise, there's an error that when googled yields zero results
```
WorldLanguages, line 30: Hehe, welcome to the world of traps and polluting 😅 

It's likely you forgot to use "bind". It's also possible you messed up with the promises. I don't think it's a good idea to use await inside a function that returns a promise. I might change this code a bit before merging

mybearworld:
> * Specifying the name of the sprite that's talking. Does it make sense?

You wouldn't be able to see which sprite is talking with regular Scratch. You also don't necessarily know which sprite is called what.

> * Specifying when the voice changes? (Scratch supports multiple types of voices)

The only thing encoded in requests is `male` (tenor, giant) or `female` (alto, squeak, kitten, tenor if language has no male voice, giant if language has no male voice), so this would have to be done by editing the behavior in the VM, which would also not work for dynamic values.

> * Hans made an comment on the style of the captions: comment https://github.com/ScratchAddons/ScratchAddons/pull/5514#issuecomment-1368908223

I'm not sure how much can be done here.

> * The previous PR did not work for "dynamic values like joins and variables"

This one does as it uses a different approach.

> * I believe the other PR did not add the subtitles to the editor. It's not a bad idea to include them too, but it will take more effort to maintain and fix incompatibilities with other editor addons fighting for the same space.

I'm really not sure how to handle this. On one hand, the editor really does not have much space available - you can see in my screenshot that the sprite area barely has room. On the other - what if someone who needs/wants those subtitles is trying to remix a project with text to speech?

addons/tts-subtitles/userscript.js
mybearworld: I don't think I used `bind` - I'll see if that fixes things.

DNin01:
> Some things that were brought up there were:
> 
> - Specifying the name of the sprite that's talking. Does it make sense?

You can't tell just from the voice, so I don't think this matters.

> - Specifying when the voice changes? (Scratch supports multiple types of voices)

That would be worth supporting if possible. It's hard to find a good solution, though. Maybe user-settable? You could choose between no distinction, voice name in brackets, and color-coding.

> - I believe the other PR did not add the subtitles to the editor. It's not a bad idea to include them too, but it will take more effort to maintain and fix incompatibilities with other editor addons fighting for the same space.

One possible fix for that would be to display them on the stage. This would be optional, of course.

addons/tts-subtitles/userscript.js
mybearworld: It doesn't like the `this` value.

```js
  // Uncaught TypeError: [object Module] is not a constructor
  const oldRace = Promise.race;
  window.Promise.race = (args) => {
    const result = oldRace.bind(this)(args);
```

WorldLanguages: Arrow functions do not have a defined `this` value. Try a normal function.

WorldLanguages: > const result = oldRace.bind(Promise)(args);

Well, I guess that works too. But now you know why "this" wasn't working.

WorldLanguages:
> I'm really not sure how to handle this. On one hand, the editor really does not have much space available - you can see in my screenshot that the sprite area barely has room. On the other - what if someone who needs/wants those subtitles is trying to remix a project with text to speech?

Could we make it a setting and by default, not show the subtitles in the editor?

addons/tts-subtitles/addon.json
```diff
@@ -0,0 +1,31 @@
...
+    {
+      "matches": ["projects"],
+      "url": "userstyle.css"
+    }
+  ],
+  "versionAdded": "1.36.0",
+  "info": [
+    {
+      "id": "add-subtitles",
+      "text": "Only people that have this addon enabled will be able to see these subtitles. In your projects, make sure to add subtitles yourself, so that anyone can use these subtitles.",
```
WorldLanguages, line 21: > make sure to add subtitles yourself

What is this supposed to mean?

DNin01: Probably creating your own subtitles feature for projects you create. Might want to reword that part. It is a good point - it reminds users of this addon who create projects that this isn't necessarily a substitute for making proper built-in subtitles.
```suggestion
      "text": "Only users who have this addon enabled will get subtitles. If you want to offer subtitles to all users in your projects, you should include built-in subtitles.",
```

mybearworld: That is what I meant, thanks!

WorldLanguages:
> Could we make it a setting and by default, not show the subtitles in the editor?

What do you think?

mybearworld:
> > Could we make it a setting and by default, not show the subtitles in the editor?
> 
> What do you think?

That sounds good - you could dynamically enable them from the popup if you need them in the editor, then.

mybearworld:
Text-to-speech pronounces some emoji and emoticons as text. However, this isn't unique - `😃`, `😆` and `😂` are all said as `smiley face`. This could theoretically lead to someone having an unfair advantage in a project. This is also treated slightly different by different languages - the set of emojis that are differentiated, whether and if so which emoticons get pronounced, if anything is even pronounced at all, is all different. Should this addon do anything about that?

DNin01:
> Text-to-speech pronounces some emoji and emoticons as text. However, this isn't unique - `😃`, `😆` and `😂` are all said as `smiley face`. This could theoretically lead to someone having an unfair advantage in a project.

Maybe that's theoretically possible, but most creators don't use emoji in text-to-speech prompts because of how they're handled, and of those creators, probably very few will discover this. I would also think it would be a violation of Scratch's terms of use for someone to try to deliberately split SA users from non-SA users like this when they can't necessarily talk about it because of Scratch's policies and the users must have this addon available and enabled to harness this difference (reducing your project's accessibility). Plus, anyone can just see inside. I don't think it's a huge deal.

And what unfair advantages could result, anyway?

DNin01:
It _could_ be problematic if you need to know what is spoken when a certain emoji is used (say, in some sort of puzzle that relies on you knowing the names of emojis as they are spoken. But then again, I'm not sure how often we see projects like that). That could possibly be an accessibility issue, but it is better than not having the addon at all.

mybearworld:
> It _could_ be problematic if you need to know what is spoken when a certain emoji is used (say, in some sort of puzzle that relies on you knowing the names of emojis as they are spoken. But then again, I'm not sure how often we see projects like that). That could possibly be an accessibility issue, but it is better than not having the addon at all.

That's what I meant. I know it's not very likely to happen, but it is possible.

WorldLanguages:
> Text-to-speech pronounces some emoji and emoticons as text. However, this isn't unique - `😃`, `😆` and `😂` are all said as `smiley face`. This could theoretically lead to someone having an unfair advantage in a project. This is also treated slightly different by different languages - the set of emojis that are differentiated, whether and if so which emoticons get pronounced, if anything is even pronounced at all, is all different. Should this addon do anything about that?

Wow, would have never considered this...

It's still quite easy to know which text is being spoken by looking at the Network tab on the browser devtools, right? This addon is nice for accessibility and preventing this problem is near impossible. So I don't think we should do anything about it.

(Related: the pause button doesn't behave very well in some cloud data projects, but as far as I know putting your computer to sleep also achieves the same effect of "pausing" a project and resuming it later, and pause is a very useful feature...)

mybearworld:
> It's still quite easy to know which text is being spoken by looking at the Network tab on the browser devtools, right? This addon is nice for accessibility and preventing this problem is near impossible. So I don't think we should do anything about it.

That makes sense.


> (Related: the pause button doesn't behave very well in some cloud data projects, but as far as I know putting your computer to sleep also achieves the same effect of "pausing" a project and resuming it later, and pause is a very useful feature...)

...Which is probably applicable to much more projects than the emoji thing. It should really be fine.

DNin01:
> It's still quite easy to know which text is being spoken by looking at the Network tab on the browser devtools, right?

Not everyone understands their way around developer tools.

WorldLanguages:
> Not everyone knows how to use developer tools. It's less user-friendly to have to observe network logs in a separate window or sidebar to tell how an emoji is read by the TTS voice.

Yeah, but the combination makes this not so important: it's already possible, and it's very hard for us to do anything about it without sacrificing accessibility features

mxmou:
What exactly was wrong with the previous PR (#5514)? It does appear to support variables.
image
image

WorldLanguages:
@mxmou I think more complex data didn't work in the previous, such as joins ("x" + "y"). But we'd need to ask the PR author to be sure.

Issue #6779, opened 2023-10-16
Have line endings auto-converted after committing


DNin01:
Line endings have probably slipped past many contributors' minds. Since different operating systems use different end-of-line sequences (Windows uses CR + LF characters and all other notable modern operating systems use a LF character, to mark the end of each line), we like to make sure that code in the repo maintains a consistent line ending format through Prettier to prevent line endings from getting mixed.

### So, what's the problem?

Historically, we have used Prettier on the CI/CD end to ensure that all line endings are encoded with LF and set the preferred line ending to LF in `.editorconfig`. However, this greatly increases the likelihood that LF will appear on Windows machines as firstly, Prettier will auto-format everything to LF if used, and secondly, EditorConfig will cause all new files to use LF as well. 99% of the time, the programs developers are using are smart enough for this not to matter, however, it's **possible** and appreciated to enforce LF in a way that doesn't have the slight potential to cause issues (or degrade performance - IDK if that's something that happens) during local development:

### My proposal

- Set line endings to `auto` in `.prettierrc`. This will make Prettier automatically set the line endings (based on the file's surrounding line endings, I think?), and I believe we can still have Prettier format the line endings by running it from the command line with the `--end-of-line lf` flag, which we'd do when running it through GitHub Actions, just in case CRLF line endings end up getting pushed (like if Git is misconfigured).
- Set `end_of_line` in `.editorconfig` to `auto`, which will make newly created files (right?) use the line ending sequence native to the operating system used during development. <sub>If we ever add style prefs to `.vscode/settings.json`, don't specify the end of line style in there - I think the user's EOL sequence setting should be respected when possible and it defaults to the native sequence.</sub>
- Use `.gitattributes` to tell Git how to handle line endings. There should be a way to auto-convert line endings when committed to the repo and to the OS's native sequence when checked out. From what I can tell, it would even work for contributors that forgot to configure `core.autocrlf` in Git.

### Downsides

The problem with the above proposal is that it's a lot of work to solve a **mostly nonexistent** or at least not huge issue, but we can still learn from line ending problems and perhaps just do some of these things, or different things, as long as the source code continues to be consistently formatted with LF-line endings.

Thoughts, @ScratchAddons/admins?

### Related lines

https://github.com/ScratchAddons/ScratchAddons/blob/dafd2ba7d66b4c676db174c9bc29cb095d5a74a5/.prettierrc.json#L3
[...]

PR #6821, opened 2023-10-30
Add text=auto and more to .gitattributes

DNin01:
A step towards #6779

### Changes

- Sets the `text=auto` Git attribute, which makes Git convert the line endings in files that it thinks are text to LF on checkin (when committing) and to the system's native line ending on checkout.
  - Does not affect Prettier or EditorConfig.
  - Contributors can still use LF during development if they want.
- Marks the most significant file types in the repository as either text or binary as appropriate.
- Tells Linguist (GitHub's renderer) that .vscode/*.json are JSON with Comments files.

### Reason for changes

Often, new contributors can be committing inconsistent line endings if they have not set `core.autocrlf` correctly. This setting performs line ending conversion when checking in changes, and optionally during checkout as well. This means additions and modifications to text files that are committed always use LF characters to mark the ends of lines, which is what we want. Setting `text=auto` in the `.gitattributes` file replicates the desired behavior for checkout (assuming `core.eol` is set to `native`) and checkin.

Learn more:
- https://git-scm.com/docs/gitattributes#_checking_out_and_checking_in
- https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreeol
- https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreautocrlf

Please correct me if I'm wrong.

### Notes

***Tested as of https://github.com/ScratchAddons/ScratchAddons/pull/6821/commits/caddeb53976ea013a41575141534f1caeb6f7840***

Tested in Git version 2.41.0.windows.1 with the following config/attribute combinations:
- `core.autocrlf` = `true`, `* text=auto`
- `core.autocrlf` = `false`, `* text=auto`
- `core.autocrlf` = `false`

Confirmed the following (with `text=auto` unless otherwise specified):
- The index stays the same no matter how `core.autocrlf` is configured (tested with true and false)
- No changes were made when running `git add --renormalize .`
- Before, when this attribute was not set and `core.autocrlf` = `false`, running `git add --renormalize .` can cause CRLF to appear in the index
- Theoretically, this would work on macOS or Linux too, only with the working directory having LF line endings (haven't tested, why didn't setting `core.autocrlf=false` and `core.eol=lf` then switching branches in GitHub Desktop replicate this?)
- I have read the Git documentation and I think I know what I'm doing

Tests:
- [x] Successful test on Windows
- [ ] Successful test on macOS/Linux


Hans5958 (approved):
General convention, so LGTM. Probably put it on the top instead on the bottom of `export-ignore`.

DNin01:
I should check what file extensions this applies to.

DNin01:
I haven't found anything that suggests that Git takes file extensions into account when determining whether a file is text or binary, so it would probably be sensible to explicitly mark each file type that appears in the repository as either text or binary.

TheColaber (approved):

DNin01:
@Hans5958 Some [changes](https://github.com/ScratchAddons/ScratchAddons/pull/6821/commits/89761afa3ee894a879039bb28a1494fe3234e31d) were made since your last review. Does this still look OK? If it is, I think we're ready to merge.

Hans5958 (commented):
All good in me.

Hans5958 (approved):

PR #6822, opened 2023-10-30
Prefer native line endings locally

### LINKED ISSUE #6779: Have line endings auto-converted after committing

DNin01:
Resolves #6779
- [x] Prerequisite: #6821

### Changes

It's nice when the system gets to read its native line ending, so this change strives to apply the following:
- Prefer the native line ending of the developer's OS when making changes locally. EditorConfig now uses the native EOL sequence and Prettier doesn't care about line endings anymore.
- Prettier DOES care about the line endings when run through GitHub Actions; the workflow sets Prettier to prefer LF line endings. This makes sure that if any CRLF gets accidentally committed, it is corrected.

As listed earlier, this should ideally be combined with #6821 to reduce the likelihood of CRLF appearing in commits. Well, I guess it's not strictly necessary as long as people have Git configured to handle line endings properly, but this change means that Prettier and EditorConfig won't be able to help with using consistent line endings before committing.

### Tests

Tested in github.dev except for EditorConfig and fully tested the workflow. This should apply to status checks too, once this is merged.


Hans5958:
If, in the end, the line breaks are then converted by the formatter, then I would rather to have it on the start of my development process then let my formatter do it later.

PS. Heck, let's set up pre-commit hooks if we want more consistency. Would make Chromebook developers sad, though.

DNin01:
Git would still convert them to LF as you commit, right? Especially with the PR I mentioned in my description? No need for pre-hooks?

Was also going to say again that I don't know if this has ever been a problem. It doesn't seem like having LF line endings locally causes any issues at this time, other than showing warning messages on `git add .` if you use auto-CRLF. If that's not enough to warrant this change and no one's ever experienced issues with having LF locally, then none of these changes are necessary. Personally, I would still apply it anyway because the changes are simple and can be combined with the `.gitattributes` addition linked above, and it makes the most sense to me to use system-native line endings wherever possible.

DNin01:
@Hans5958 It doesn't have to go through the formatter. Git auto-converts CRLF in the working directory to LF when it is committed as long as it's configured properly (with auto-CRLF, or text in `.gitattributes` as seen in #6821).

WorldLanguages (approved):

Hans5958 (approved):
I'm still uncertain about this, but since the activity here is quite low, I would give an approval just to see how things pans out.

Samq64:
Is this ready to merge after fixing the conflict?

PR #6824, opened 2023-10-30
Stop window scrolling when you scroll over a comment

Jazza-231:
Resolves bug report in SA server

### Changes

Just added pointer-events: none; to scroll.css

### Reason for changes

Comments should be ignored and should just scroll the code area

### Tests

Works on brave

### DIFF
--- a/addons/fix-editor-comments/scroll.css
+++ b/addons/fix-editor-comments/scroll.css
@@ -4,4 +4,5 @@
 .scratchCommentTextarea {
   overflow-y: auto;
   background-color: transparent;
+  pointer-events: none;
 }


CST1229:
Wouldn't this also prevent the textarea from being focused or selecting anything in it?

WorldLanguages:
> Wouldn't this also prevent the textarea from being focused or selecting anything in it?

Yeah, I think we'd need to fix this with JavaScript instead (event capture and stop-propagation)
CSS will only let you disable all mouse events, it doesn't let you disable just scroll.

WorldLanguages:
For context, apparently this is the bug report

> Intended behavior: Holding CTRL/CMD and scrolling while hovering in the code editor zooms Scratch's code editor in and out.
>
> However, while using the Better Editor Comments addon and zooming while hovering over a code comment, it uses the browser's built-in zoom feature instead of Scratch's to zoom in and out. Happens on SA v1.35.1, both Chrome and Firefox

Jazza-231:
The second I submitted this I knew it didn't work 😭😭 aaaaaaaaaaa

Jazza-231:
See https://github.com/ScratchAddons/ScratchAddons/issues/6829

PR #6828, opened 2023-10-30
Adjust "custom block text style" addon description

MaterArc:
### Changes

Tiny grammar fix to make it sound more fluent.

### Reason for changes

Fluency

### DIFF
--- a/addons/custom-block-text/addon.json
+++ b/addons/custom-block-text/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Customizable block text style",
-  "description": "Allows you to adjust the size and thickness of block text or add a shadow to make it easier to read.",
+  "description": "Allows you to adjust the size and font weight of block text or add a shadow to make it easier to read.",
   "tags": ["editor", "codeEditor", "featured"],
   "credits": [
     {
--- a/addons/custom-block-text/addon.json
+++ b/addons/custom-block-text/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Customizable block text style",
-  "description": "Allows you to adjust the size and font weight of block text or add a shadow to make it easier to read.",
+  "description": "Lets you customize the style of text on blocks, including adjusting the size, using bold text, or adding a shadow.",
   "tags": ["editor", "codeEditor", "featured"],
   "credits": [
     {


addons/custom-block-text/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Customizable block text style",
-  "description": "Allows you to adjust the size and thickness of block text or add a shadow to make it easier to read.",
+  "description": "Allows you to adjust the size and font weight of block text or add a shadow to make it easier to read.",
```
DNin01, line 3: "Size and font weight" sounds slightly inconsistent. I wonder if saying "font size and weight" or just "size and weight" would be better. I don't know, what do some other folks think?

Wait for some conversation before changing anything.

WorldLanguages:
Isn't "thickness" good enough?

just-a-hriday:
> Isn't "thickness" good enough?

Yeah, it sounds better to me too.

MaterArc:
Thickness just sounds off to me, that could just be me. It sounds choppy. It took me more than once reading the description to understand what they were actually trying to say. Again, this could just be a me thing...

BroJac5246:
> Thickness just sounds off to me, that could just be me. It sounds choppy. It took me more than once reading the description to understand what they were actually trying to say. Again, this could just be a me thing...

I'm not a fan of "thickness," either. But do people know what font weight is?

WorldLanguages:
Font weight is probably the right term, but it's hard to tell how many users will understand what it means. As a non-native English speaker "thickness" sounds more intuitive, I basically had to memorize the "font-weight" CSS property due to its frequency.

DNin01:
Maybe it would be better to say "use bold text" instead of "adjust the font weight".

WorldLanguages:
> Maybe it would be better to say "use bold text" instead of "adjust the font weight".

Yeah, that would reflect the current customization options.

addons/custom-block-text/addon.json
DNin01, line 3:
```suggestion
  "description": "Lets you customize the style of text on blocks, including adjusting the size, using bold text, or adding a shadow.",
```

MaterArc:
Thanks, @DNin01 ! It looks much better now :)

MaterArc:
Nope, I think it's merge-ready

PR #5927, opened 2023-04-15
Turn forum toolbar icons white in dark mode

mybearworld:
Resolves #5923

### Changes

Remove the background of the forum toolbar icons in dark mode, and instead replace them with white alternatives.

### Reason for changes

To use dark buttons in dark mode.

### Tests

<s>Tested locally on Edge 114.0.1788.0 (Official build) dev (64-bit)</s> I'm not sure how to download branches, so I can't test it properly.

PR #6836, opened 2023-11-04
Markdown in forums addon

### LINKED ISSUE #6557: Markdown in forum posts
### Why this would be helpful

Most platforms use Markdown for their comments/forum posts. Remembering to use BBCode for Scratch can be annoying.

### How this addon works

Before submitting, this addon would convert all Markdown into BBCode. This would be a complete replacement for BBCode (as BBCode and Markdown are incompatible).

Quotes could have a `bbcode` directive of sorts, so quotes can still work:

```md
> bbcode
> This is a quote with [u]BBCode[/u], which is literally put in so that you can quote posts using BBCode without needing to [i]convert[/i] the BBCode to Markdown.

i agree
```

Of course, we wouldn't support Markdown features that aren't supported in BBCode, so no tables, headings, etc.

### Possible alternatives

_No response_

### Additional context

Related: #184

mybearworld:
Resolves #6557

### Changes

Create a new addon "Markdown in forums" that allows you to use Markdown in the forums by converting the Markdown to BBCode on the fly. This also adds a new library, [marked](https://marked.js.org/).

### Reason for changes

Most platforms use Markdown for their comments/forum posts. Remembering to use BBCode for Scratch can be annoying. Also, Markdown is generally more readable more BBCode and can just be more fun to write.

### Tests

Tested in Edge.

## To Do

This PR is a draft. There are still some important things to be done for this addon to be usable.

- How to handle features that Markdown has but the forums don't?
  - [ ] Tables - currently it just stops a user from posting anything with tables
  - [X] Inline code - this is implemented by just outputting what the user entered - ``the `print` function`` -> ``the `print` function``
  - [X] Task lists - this is implemented by 
- How to handle tags that the forums have but Markdown doesn't?
  - [ ] [u]
  - [ ] [big], [small]
  - [X] [scratchblocks] - this is implemented by a special `scratchblocks` language on code blocks
  - [ ] [color]
  - [ ] [center]
  - [ ] [wp], [wiki], [google], [dictionary]
- Compatibility with quotes
  - [X] A way for quotes to use BBCode - this is implemented by a special `raw-bbcode` language on code blocks
  - [x] Outputting in Markdown when clicking on the Quote button
- [ ] Editing posts where just using `raw-bbcode` would not be ideal
- Compatibility with toolbar
  - [X] Handle toolbar buttons that Markdown supports that are there
  - [ ] Handle toolbar buttons that BBCode has that Markdown doesn't have (notably list item)
  - [X] Handle toolbar buttons that don't exist but Markdown has (e.g. headings)
  - [ ] Keyboard shortcuts
  - [ ] Scratchblocks
- [ ] Compatibility with preview panel
- Compatibility with other addons
  - [ ] `better-quoter`
  - [ ] `forum-toolbar`
  - [x] `forum-id`
  - [x] `dark-www`
  - [x] `material-forum-editor-buttons`
- [ ] Explanation of Markdown syntax?
- [ ] BBCode is provided for editing when 60 second rule applies


mybearworld:
> * [ ]  Editing posts where just using `raw-bbcode` would not be ideal

A way to do this would be to include the original markdown, maybe base64 encoded, in the original post. This would make the post very long and annoying to quote, though.

Secret-chest:
> A way to do this would be to include the original markdown, maybe base64 encoded, in the original post. This would make the post very long and annoying to quote, though.

Find a way to convert BBCode back to Markdown and add some special code to shos that each post is Markdown.

mybearworld:
> > A way to do this would be to include the original markdown, maybe base64 encoded, in the original post. This would make the post very long and annoying to quote, though.
> 
> Find a way to convert BBCode back to Markdown and add some special code to shos that each post is Markdown.

There's technically multiple ways to do certain things - `*italic*` versus `_italic_` comes to mind: *italic* _italic_.
There may also be cases where you just really want to keep your particular formatting, which this method wouldn't allow.

Secret-chest:
Heading suggestions:

- h1: big and bold
- h2: big and underlined
- h3: big
- h4: bold and underlined
- h5: underlined
- h6: bold

mybearworld:
> Heading suggestions:
> 
> * h1: big and bold
> * h2: big and underlined
> * h3: big
> * h4: bold and underlined
> * h5: underlined
> * h6: bold

These are pretty good, I've added them.

mybearworld:
I can't get the icons for `dark-www` to be consistent with the other ones :P
I really should've written down what I did for #5927...

mybearworld:
> Also I think 2 lines for paragraphs make sense.

The problem is that paragraphs are every line of text, so this:
```md
> **User wrote:**
> (#1)
> ```raw-bbcode
> This is a quote!
> ```
```
gets turned into this:
```
[quote User](#1)

This is a quote![/quote]
```
Which doesn't look too good.

mybearworld:
An idea for wiki/dictionary/etc links: maybe the syntax could be something like `[Link text](wiki:Target Page)`? One problem with that is that `<wiki:Target Page>` would be unintuitive - should it output `[wiki Target Page]Target Page[/wiki]` or `[wiki Target Page]wiki:Target Page[/wiki]`? Arguably the former would be more useful, but the latter might be more in line with expectations.

mybearworld:
Is there a way to not have Format code mess things up for preexisting PRs?

WorldLanguages:
> Is there a way to not have Format code mess things up for preexisting PRs?

In the future, the action will not run automatically.

addons/image-uploader/userscript.js
```diff
@@ -136,7 +136,12 @@ export default async function ({ addon, msg, console }) {
       const data = await res.json();
 
       if (data.status === "ok") {
-        insert(textBox, `[img]https://assets.scratch.mit.edu/get_image/.%2E/${data["content-name"]}[/img]`);
+        insert(
+          textBox,
+          (await addon.self.getEnabledAddons()).includes("markdown-forums")
+            ? `![](https://assets.scratch.mit.edu/get_image/.%2E/${data["content-name"]})`
+            : `[img]https://assets.scratch.mit.edu/get_image/.%2E/${data["content-name"]}[/img]`
```
WorldLanguages, line 143: I think this looks perfectly OK, but do we want to allow dynamic enable & disable of the new `markdown-forums` addon? Won't it be a mess to revert these syntax changes back?

mybearworld: It currently allows dynamic enable/disable without converting anything back and forth. This is probably safer than trying to accurately convert between BBCode and Markdown.

WorldLanguages: Yeah that sounds like a mess

mybearworld:
I accidentally deleted my fork. New PR is at #7970

Issue #5652, opened 2023-01-20
Slowmo mode


reemlife1:
### Why this would be helpful

This would be helpful for checking if variable are working correctly, if you want a fun challange and more!

### My suggestion

When you click on the Green Flag while holding the Alt key,it will turn on/off Slowmo mode that slows the game down by how much is needed (maybe even make another addon that makes turbo and (maybe) slowmo have sliders for how fast/slow they will go?).

### Possible alternatives

You can also make one slider when you turn on turbo mode that first in the middle, you can slide it to the left to make it slower, and to the right to make it faster!

DNin01:
We could lower the minimum frame rate setting in the `60fps` addon, too, which might work better. However, this could allow some games to be cheesed.

fractioneater:
> However, this could allow some games to be cheesed.

It's easy to create a turbo mode detector in Scratch, and it shouldn't be any harder to create a slow motion detector. If the creator of the project really cares, they only have to put the slightest bit of effort into preventing it.

BroJac5246:
> > However, this could allow some games to be cheesed.
> 
> It's easy to create a turbo mode detector in Scratch, and it shouldn't be any harder to create a slow motion detector. If the creator of the project really cares, they only have to put the slightest bit of effort into preventing it.

A few issues:

1. Not everyone is aware SA exists or could have said feature.
2. Lag could be misinterpreted as a slow-mo feature.
3. SA should not adversely affect those using vanilla Scratch.

I'm not against this feature, but those are some considerations.

WorldLanguages:
Looks like a duplicate of #2760 (please tell me if you disagree)

DNin01:
> Looks like a duplicate of #2760 (please tell me if you disagree)

That suggestion looks like a general speed adjustment feature, while this one seems to be geared towards just slowing down the project, so I don't think it's necessarily a duplicate, but it's related for sure.

penguinmoose:
@BroJac5246 There can be a clear warning right above the project canvas that makes the user aware of slow-mode.

BroJac5246:
> @BroJac5246 There can be a clear warning right above the project canvas that makes the user aware of slow-mode.

Why? We don't display such a warning for 60FPS.

penguinmoose:
> > @BroJac5246 There can be a clear warning right above the project canvas that makes the user aware of slow-mode.
> 
> Why? We don't display such a warning for 60FPS.

[...]

PR #6848, opened 2023-11-09
[String change] `60fps`: Rewrite the title and description for better clearness

iqnite:
Many people don't know what FPS means on Scratch, and the title "60FPS project player mode" is partially incorrect because you can change the frame rate. Adding an explanation and changing the setting's name might help those users.

### DIFF
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -1,10 +1,15 @@
 {
-  "name": "60FPS project player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Custom project framerate",
+  "description": "Alt+Click the green flag to toggle between Scratch's default 30FPS and a custom framerate.",
   "info": [
     {
+      "type": "info",
+      "text": "The framerate (FPS) determines how many blocks are run per second. For example, running at 60FPS doubles the execution speed.",
+      "id": "fpsExplanation"
+    },
+    {
       "type": "notice",
-      "text": "Most projects will not behave properly when running at 60FPS, because increasing the frame rate also increases the speed at which scripts run. This should only be used on projects that support 60FPS.",
+      "text": "Most projects will not behave properly when running at different FPS, because increasing the frame rate also increases the speed at which scripts run. This should only be used on projects that support 60FPS.",
       "id": "projectRunsFasterNotice"
     }
   ],
@@ -28,7 +33,7 @@
   ],
   "settings": [
     {
-      "name": "Alt+GreenFlag FPS",
+      "name": "FPS",
       "id": "framerate",
       "type": "integer",
       "min": 31,


DNin01 (commented):
Does anyone know how common laptop screens with refresh rates over 60 Hz are today?

In some ways, focusing on 60 FPS leads to simpler explanations and may even make the addon easier to find for those who just want 60 FPS, but of course, that wouldn't make sense if _everyone_ suddenly started owning screens that refreshed faster. I'm currently unsure on what it should say.

Secret-chest:
> Does anyone know how common laptop screens with refresh rates over 60 Hz are today?.

Most gamer laptops are 165 Hz.

Samq64:
> > Does anyone know how common laptop screens with refresh rates over 60 Hz are today?.
> 
> Most gamer laptops are 165 Hz.

Some are also 90, 120, 144 and 240. 60 is still the most common though especially on the types of devices I assume most Scratchers are using (Chromebooks, low-mid range laptops).

addons/60fps/addon.json
```diff
@@ -1,10 +1,15 @@
...
-  "name": "60FPS project player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Custom project framerate",
+  "description": "Alt+Click the green flag to toggle between Scratch's default 30FPS and a custom framerate.",
   "info": [
+    {
+      "type": "info",
+      "text": "The framerate (FPS) determines how many blocks are run per second. For example, running at 60FPS doubles the execution speed.",
+      "id": "fpsExplanation"
+    },
```
DNin01, line 9: This note is probably not important when we already warn the user in the note below this one that projects may run faster.

```suggestion

```

To comment on this explanation, it is a little too simplified and needs some extra details, IMO. Changing the framerate mainly just changes how fast the screen refreshes (and users already know that). Raising it gets us smoother animations but doing so has the _side effect_ of changing how fast projects run because Scratch automatically iterates certain code each frame, which is stated in the other notice below this one.

The framerate also does not have an exact impact on how many blocks run in a given time period; there are _specific_ things that Scratch iterates or evaluates once per frame.

Maybe if we ever make a Scratch Addons wiki or user guide we'll get deeper into this. 🙂

DNin01:
Even though >60 Hz screens are not that common on low/mid-range laptops AFAIK, I guess we could still say custom framerate instead. We have the setting, after all.

addons/60fps/addon.json
```diff
@@ -1,10 +1,15 @@
...
     {
       "type": "notice",
-      "text": "Most projects will not behave properly when running at 60FPS, because increasing the frame rate also increases the speed at which scripts run. This should only be used on projects that support 60FPS.",
+      "text": "Most projects will not behave properly when running at different FPS, because increasing the frame rate also increases the speed at which scripts run. This should only be used on projects that support 60FPS.",
```
DNin01, line 12: I have a rewrite for this:
```suggestion
      "text": "Changing the framerate of Scratch projects also affects the rate at which certain code (such as loops) runs, so most projects won't behave properly when run at a different framerate. This addon is meant to be used on projects that are designed to support a higher framerate.",
```

addons/60fps/addon.json
```diff
@@ -1,10 +1,15 @@
 {
-  "name": "60FPS project player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Custom project framerate",
```
DNin01, line 2: Custom or customizable sounds kind of like you can set it to whatever you want whenever you want, and, I mean, you can kind of, you can go into the addon settings and change it, but you can only _quickly_ switch between 30 FPS and one custom framerate, so maybe the name shouldn't over-express the customizability aspect.
```suggestion
  "name": "Higher project framerate",
```

addons/60fps/addon.json
```diff
@@ -1,10 +1,15 @@
...
+  "description": "Alt+Click the green flag to toggle between Scratch's default 30FPS and a custom framerate.",
```
DNin01, line 3:
```suggestion
  "description": "Lets you run projects at higher framerates so they look smoother. Hold Alt and click the green flag to toggle between Scratch's default 30 FPS and a custom framerate set below.",
```

addons/60fps/addon.json
```diff
@@ -1,10 +1,15 @@
...
   "info": [
+    {
+      "type": "info",
+      "text": "The framerate (FPS) determines how many blocks are run per second. For example, running at 60FPS doubles the execution speed.",
+      "id": "fpsExplanation"
+    },
     {
       "type": "notice",
```
DNin01, line 11: I'd even understand if the `projectRunsFasterNotice` note was marked as a red warning. That might catch attention a bit _too_ much depending on who you ask, though.

addons/60fps/addon.json
```diff
@@ -28,7 +33,7 @@
   ],
   "settings": [
     {
-      "name": "Alt+GreenFlag FPS",
+      "name": "FPS",
```
DNin01, line 36: Almost forgot, I think clarifying that this is the custom framerate you can set might not hurt.
```suggestion
      "name": "Custom frames per second (FPS)",
```

DNin01: "Custom frames per second" sounds a little weird to say compared to "custom FPS" (because nouns),
```suggestion
      "name": "Custom FPS",
```
but saying the full thing instead of just the abbreviation sounds like a good thing. Maybe we can say "frames per second" somewhere in the description or a note instead of here if that feels better?

addons/60fps/addon.json
iqnite: > Maybe if we ever make a Scratch Addons wiki or user guide we'll get deeper into this. 🙂

The TurboWarp docs already have an article about that. Your suggestion makes sense, I wrote this PR at 10 pm :)

addons/60fps/addon.json
DNin01: Anyone have opinions on what the name should be? It's true that it's customizable, but whatever custom framerate you've set needs to be adjusted through the settings, and also requires a gesture to activate.

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "60FPS project player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Customizable project framerate",
```
WorldLanguages, line 2: In general, addons that start with "customizable" don't change anything unless you change the settings. I thought it was worth mentioning.

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
...
+  "description": "Lets you run projects at higher framerates so they look smoother. Hold Alt and click the green flag to toggle between Scratch's default 30 FPS and a custom framerate set below.",
```
WorldLanguages, line 3: I'm not sure how important this is, but previously `Alt+Click` was part of the addon name, then it became part of the first sentence of the description, and now it was moved to the second. After each of these changes there's probably more people that don't know how to toggle this

WorldLanguages:
Looks like an enhancement to me, I don't have strong opinions in favor of keeping the current strings nor I strictly think they need to change. Mentioning the word "framerate" is nice though.

addons/60fps/addon.json
DNin01: My idea was this:
```suggestion
  "name": "Higher project framerate",
```

iqnite: > In general, addons that start with "customizable" don't change anything unless you change the settings. I thought it was worth mentioning.

It's like halfway, because the settings aren't the Scratch default, but they're not applied until the user applies them by Alt+clicking on the green flag. That puts the addon in a strange position. It would be resolved if we just

> discontinue this addon and have people use the TurboWarp button addon instead

and considering that all those suggestions about adding Interpolation, HQ Pen, etc. are being rejected, this would be fair enough, and it would abide by the DRY principle. The problem is (as always) that previous users will be confused, and people who design projects for 60FPS might want to enable the feature in the editor, which is impossible with TurboWarp mode.

addons/60fps/addon.json
iqnite: Maybe changing the title to "Alt+Click green flag to increase FPS" might do the job, like "Ctrl+Enter to post".

DNin01: > Maybe changing the title to "Alt+Click green flag to increase FPS" might do the job, like "Ctrl+Enter to post".

The issue is we prefer noun phrases.

iqnite: > The issue is we prefer noun phrases.

"Do not run duplicated blocks"

DNin01: > > The issue is we prefer noun phrases.
> 
> "Do not run duplicated blocks"

It's not a strict rule but it is in the best practices.
> Use nouns instead of verbs (for example, `Customizable new sprite position` instead of `Change new sprite position`) unless doing so would result in the addon name being overly complicated.

I think using a noun phrase here works better and may be shorter as well. It's not exactly the same as "Ctrl+Enter to post".

iqnite: OK that's true. But how should we get around that problem then?

iqnite: "Alt+Click green flag for higher FPS"

WorldLanguages: I would try to keep `Alt+Click the green flag` within the first 10 words of the description

WorldLanguages:
I would personally go for

- Addon name: `Higher project framerate` or `Higher project framerate mode`
- Description: `Hold Alt while clicking the green flag to run the project at more FPS. You can toggle between Scratch's default 30 FPS and a custom framerate set below.`  (I removed the mention of "smoothing")

Samq64:
> - Description: Hold Alt while clicking the green flag to run the project at more FPS. You can toggle between Scratch's default 30 FPS and a custom framerate set below. (I removed the mention of "smoothing")

Maybe more FPS could be changed to higher frames per second?

DNin01:
> > * Description: Hold Alt while clicking the green flag to run the project at more FPS. You can toggle between Scratch's default 30 FPS and a custom framerate set below. (I removed the mention of "smoothing")
> 
> Maybe more FPS could be changed to higher frames per second?

Honestly, I don't find either way sounding quite right to me. I would use the word "framerate" except that we already use that in the name and once in the description.

Secret-chest:
Name: _Faster project framerate toggle_

Description:
> Hold Alt while clicking the green flag to toggle between your FPS set here and the default 30 FPS

iqnite:
The concept is:
- The title contains basic info that allows you to understand what the addon does at a glance. For example, an experienced user can understand "Alt+Click green flag for higher framerate" immediately. For people who don't know what framerate means, they can look at the addon's description where it explains what FPS is and they can change it below.

WorldLanguages:
Relevant comment: https://github.com/ScratchAddons/ScratchAddons/issues/5652#issuecomment-1524076419

addons/60fps/addon.json
mybearworld: "Higher FPS by Alt+Click on green flag", or something to that avail might work.

Samq64 (commented):
I also prefer higher project framerate for the addon name. "Alt+click green flag to change project framerate" is very long.

Secret-chest:
The setting should mention it's per second.

DNin01:
> The setting should mention it's per second.

We should at least say "per second" somewhere, even if it's not on the setting name.

addons/60fps/addon.json
```diff
@@ -28,7 +28,7 @@
   ],
   "settings": [
     {
-      "name": "Alt+GreenFlag FPS",
+      "name": "Custom frames per second (FPS)",
```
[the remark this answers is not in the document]
DNin01: > @Secret-chest
> 
> The setting should mention it's per second.

We should at least say "per second" somewhere, even if it's not on the setting name.

DNin01, line 31: Not sure if a setting name really needs to include "frames per second" and its abbreviation. It could be moved into the description, perhaps.
```suggestion
      "name": "Custom frames per second",
```

Or we could use only an abbreviation:
```suggestion
      "name": "Custom FPS",
```

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "60FPS project player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Alt+click green flag to change project framerate",
+  "description": "Lets you run projects at higher framerates so they look smoother. Hold Alt and click the green flag to toggle between Scratch's default 30 FPS and a custom framerate set below.",
```
DNin01, line 3: To abbreviate "frames per second" in the description:
```suggestion
  "description": "Lets you run projects at higher framerates so they look smoother. Hold Alt and click the green flag to toggle between Scratch's default 30 frames per second (FPS) and a custom framerate set below.",
```

DNin01 (commented):
Regarding the use of the abbreviation "FPS", I have a suggestion...

iqnite:
I'm sorry for the wait. I've now taken care of most of the abovementioned issues.

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "60FPS project player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Higher project framerate mode",
+  "description": "Lets you run projects at more frames per second (FPS) so they look smoother. Hold Alt and click the green flag to toggle between Scratch's default 30 FPS and a custom framerate set below.",
   "info": [
     {
-      "type": "notice",
-      "text": "Most projects will not behave properly when running at 60FPS, because increasing the frame rate also increases the speed at which scripts run. This should only be used on projects that support 60FPS.",
+      "type": "warning",
```
Samq64, line 6: This might have been brought up already but does this need to be a warning? I don't think it does since it's still not the default when loading a project.

iqnite: Yes, but notices tell the user in which situations the addon might not work, and warnings say how the addon may hinder the user experience.

mxmou: > Yes, but notices tell the user in which situations the addon might not work, and warnings say how the addon may hinder the user experience.

Not necessarily. The following are all notices:
* `cat-blocks`: The "watch mouse cursor" setting may impact performance when the editor is open.
* `better-img-uploads`: Avoid using the HD upload button if you plan to edit the image after uploading.
* `turbowarp-player`: "Replace player" mode changes the Scratch player to a TurboWarp embed, which doesn't support seeing inside projects.

mxmou: I think a warning is appropriate if simply enabling the addon has negative effects. In this case, enabling the addon isn't enough to make projects behave incorrectly - 60 FPS mode needs to be activated.

addons/60fps/addon.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "60FPS project player mode",
-  "description": "Alt+Click the green flag to toggle 60FPS.",
+  "name": "Higher project framerate mode",
+  "description": "Lets you run projects at more frames per second (FPS) so they look smoother. Hold Alt and click the green flag to toggle between Scratch's default 30 FPS and a custom framerate set below.",
   "info": [
     {
       "type": "notice",
-      "text": "Most projects will not behave properly when running at 60FPS, because increasing the frame rate also increases the speed at which scripts run. This should only be used on projects that support 60FPS.",
+      "text": "Changing the framerate of Scratch projects also affects the rate at which certain code (such as loops) runs, so most projects won't behave properly when run at a different framerate. Higher framerates are only meant to be used on projects that are designed to support them.",
```
Samq64, line 7:
```suggestion
      "text": "Changing the framerate of Scratch projects also affects the rate at which certain code such as loops run, so most projects won't behave properly when run at a different framerate. Higher framerates are only meant to be used on projects that are designed to support them.",
```

Samq64: Never mind, it's fine the way it is.

WorldLanguages:
@DNin01 If you approve this we can merge it

DNin01 (approved):
This is actually pretty good.

PR #6868, opened 2023-11-14
Format code

DNin01:
It looks like Prettier has made some changes again and is formatting some things that were fine before. I just opened this pull request so those changes can happen.

I recommend all contributors merge `upstream/master` into their branches once this pull request has been merged to avoid headaches.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -36,8 +36,8 @@ export default class Tab extends Listenable {
       this._clientVersion = document.querySelector("meta[name='format-detection']")
         ? "scratch-www"
         : document.querySelector("script[type='text/javascript']")
-        ? "scratchr2"
-        : null;
+          ? "scratchr2"
+          : null;
     return this._clientVersion;
   }
   /**
--- a/addons/editor-number-arrow-keys/userscript.js
+++ b/addons/editor-number-arrow-keys/userscript.js
@@ -120,8 +120,8 @@ export default async function ({ addon }) {
       let settingValue = e.shiftKey
         ? addon.settings.get("shiftCustom")
         : e.altKey
-        ? addon.settings.get("altCustom")
-        : addon.settings.get("regularCustom");
+          ? addon.settings.get("altCustom")
+          : addon.settings.get("regularCustom");
       if (settingValue === "") settingValue = 0;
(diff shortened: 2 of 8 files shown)


DNin01:
This doesn't make any changes other than updating the formatting, so this can be merged without needing a thorough review.

WorldLanguages:
Let's change the Prettier GitHub action so that it never commits automatically unless you manually run it.

DNin01:
> Let's change the Prettier GitHub action so that it never commits automatically unless you manually run it.

Sure. It won't fix the status check failure, but it would remove some headaches of having to sync the branch and revert format-code commits.

DNin01:
> Can't the action be set to only use a specific version of prettier?

Can we do that with the VS Code extension (and the Node.js module if we ever use that for development)?

DNin01:
But let's start by merging this. This issue is disrupting our contributors' pull requests.

WorldLanguages:
I was thinking about 2 separate workflows because contributors could still want to manually format their code by manually triggering the action, but I bet it's also possible to do it with one, somewhat more confusing though

Hans5958:
> I was thinking about 2 separate workflows because contributors could still want to manually format their code by manually triggering the action, but I bet it's also possible to do it with one, somewhat more confusing though

There's also a way by using `if:` on a step, so the formatting step can be run only on dispatch.

DNin01:
> Let's change the Prettier GitHub action so that it never commits automatically unless you manually run it.

The root problem, though, is that Prettier sometimes changes the way things are formatted. Lots of things naturally change overtime and disrupt workflows when they do. Making this process manual won't solve it completely. It'll only make the situation controllable. Even if we do this, most contributors will have this happen at least once per update because they don't know what they're about to run into. And then they have to wait for us to fix it (because our [rules](https://github.com/ScratchAddons/ScratchAddons/settings/branches) don't allow our action to push to our `master` branch) and then merge upstream into their branches.

Hardcoding a version might help as it would give us control over the problem, but we may still want to update occasionally.

WorldLanguages:
@DNin01 I haven't think of it much, but doesn't hardcoding the version cause similar problems? What would we need to do when we want to increase the version manually? Still, I guess manually updating it is better than it happening automatically with no prior notice.

PR #6873, opened 2023-11-15
`scratchr2`: Style comment warnings correctly

### LINKED ISSUE #6859: Wrong comment warning position with scratchr2
### Description

> Dunno if it could really be fixable (other than attempting to swap it to the message thing like with studio pages or project pages), so I'd rather send it through the feedback page here (and since I'm not really a good explainer). The mute reason overlaps a bit of comments on 2.0 pages (eg. user profiles, etc) with 2.0 -> 3.0 enabled (Firefox 119.0.1 32 bit here) - It's not really a big problem but thought I'd report anyways.
(Apologies for using assets.scratch.mit.edu for this incase it's inconvenient in this regard)
https://assets.scratch.mit.edu/get_image/.%2E/ba823cc902c18cef4d76f224c73cefed.png
Does block a bit of view until you dismiss it which is kind of annoying though.

### Steps to reproduce

This also happens with our message.

1. Enable `scratchr2`
2. Go to your profile
4. Comment "Scratch Addons"

ba823cc902c18cef4d76f224c73cefed

This doesn't happen on scratch-www because it doesn't use a tooltip and on profiles without the addon the textbox isn't as wide.

### Environment

Browser: Firefox 119
Operating system: Linux

mybearworld:
Resolves #6859

### Changes

Styles comment warnings on profiles so they look as they do in scratchr2. This also makes the issue linked above no longer apply.

An empty comment with a warning saying "You can't post an empty comment!" on a scratchr2 page with the scratchr2 addon enabled, with the warning looking just as with scratch-www.

### Reason for changes

Making it more consistent, and fixing a bug.

### Tests

Tested in Edge.

### DIFF
--- a/addons/scratchr2/scratchr2_comments.css
+++ b/addons/scratchr2/scratchr2_comments.css
@@ -180,17 +180,17 @@
   /* emoji-picker compatibility */
   width: auto;
 }
-#comments #comment-form .control-group.tooltip {
+#comments .control-group.tooltip {
   display: flex;
   flex-direction: column;
 }
-#comments #comment-form .tooltip textarea {
+#comments .control-group.tooltip textarea {
   margin-bottom: 0.5rem;
 }
-#comments #comment-form .tooltip.error textarea {
+#comments .control-group.tooltip.error textarea {
   box-shadow: none;
 }
-#comments #comment-form .tooltip.error #comment-alert {
+#comments .control-group.tooltip.error .hovertext {
   position: static;
   box-shadow: none;
   order: -1;
@@ -202,10 +202,10 @@
   padding: 0.25rem;
   width: 100%;
 }
-#comments #comment-form .tooltip.error textarea:focus + #comment-alert {
+#comments .control-group.tooltip.error textarea:focus + .hovertext {
   display: block !important;
 }
-#comments #comment-form .tooltip.error #comment-alert::after,
-#comments #comment-form .tooltip.error .arrow {
+#comments .control-group.tooltip.error .hovertext::after,
+#comments .control-group.tooltip.error .arrow {
   display: none;
 }
--- a/addons/scratchr2/scratchr2_comments.css
+++ b/addons/scratchr2/scratchr2_comments.css
@@ -202,6 +202,9 @@
   padding: 4px;
   width: 100%;
 }
+#comments .control-group.tooltip.error .hovertext .text {
+  font-size: 13px;
+}
 #comments .control-group.tooltip.error textarea:focus + .hovertext {
   display: block !important;
 }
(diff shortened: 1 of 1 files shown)


mybearworld:
Just realized this probably won't affect replies.

mybearworld:
> Just realized this probably won't affect replies.

Fixed:

An empty reply with a warning saying "You can't post an empty comment!" on a scratchr2 page with the scratchr2 addon enabled, with the warning looking just as with scratch-www.

addons/scratchr2/scratchr2_comments.css
```diff
@@ -180,3 +180,32 @@
   /* emoji-picker compatibility */
   width: auto;
 }
+#comments .control-group.tooltip {
+  display: flex;
+  flex-direction: column;
+}
+#comments .control-group.tooltip textarea {
+  margin-bottom: 0.5rem;
```
mxmou, line 188:
```suggestion
  margin-bottom: 8px;
```

`1rem` on scratchr2 pages isn't the same size as in scratch-www.

addons/scratchr2/scratchr2_comments.css
```diff
@@ -180,3 +180,32 @@
...
+}
+#comments .control-group.tooltip.error textarea {
+  box-shadow: none;
+}
+#comments .control-group.tooltip.error .hovertext {
+  position: static;
+  box-shadow: none;
+  order: -1;
+  box-sizing: border-box;
+  margin-bottom: 0.5rem;
```
mxmou, line 198:
```suggestion
  margin-bottom: 8px;
```

addons/scratchr2/scratchr2_comments.css
```diff
@@ -180,3 +180,32 @@
...
+  border: 1px solid rgba(0, 0, 0, 0.1);
+  border-radius: 5px;
+  background-color: #ffab1a;
+  padding: 0.25rem;
```
mxmou, line 202:
```suggestion
  padding: 4px;
```

mxmou (changes requested):
The error text is currently larger than in scratch-www. I think it should be the same size as the comment text:
```css
#comments form .tooltip .hovertext .text {
  font-size: 13px;
}
```

mxmou:
The font size is still too large. Everything else looks good.

WorldLanguages:
@mybearworld Any reasons not to set the font size to 13px, as mxmou suggested?

mybearworld:
Sorry, I missed it - done

Font size adjusted

mxmou (approved):

Samq64 (approved):

PR #6889, opened 2023-11-25
discuss-button: hide username using CSS

### LINKED ISSUE #6880: `discuss-button`: "Compact user dropdown" doesn't always work on the My Stuff page
### Description

"Compact user dropdown" doesn't always work on the My Stuff page. The padding is still removed, though.

### Steps to reproduce

1. Go to My Stuff
2. Do some stuff quickly
3. If nothing happened, go back to another page and repeat

Image:

The right part of the header with no padding between username and profile picture

Error in the console:

```js
Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'textContent')
    at add (r2.js:17:54)
```

### Environment

Browser: Edge 119
Operating system: Windows

mxmou:
Resolves #6880

### Changes

Uses CSS instead of JavaScript to hide the username when the "compact user dropdown" setting is enabled.

### Reason for changes

To fix a race condition and make the code simpler.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/discuss-button/compact-nav.css
+++ b/addons/discuss-button/compact-nav.css
@@ -5,6 +5,10 @@
   display: none;
 }
 
+#topnav ul.account-nav .logged-in-user .dropdown-toggle {
+  font-size: 0 !important;
+}
+
 .sa-profile-name {
   display: block;
   font-size: 0.75rem;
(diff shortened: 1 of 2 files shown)


DNin01 (changes requested):
Great to see there's a way to do that with CSS! Even setting a minimum font size in the browser doesn't override it (because it's 0, I guess?).

However, the `scratchr2` addon overrides this on scratchr2 pages. This could be fixed by using `!important`, ordering this addon after `scratchr2` in `addons.json`, or increasing the specificity.

DNin01 (approved):

WorldLanguages (approved):

WorldLanguages:
Thanks DNin01 for noticing the conflict with the scratchr2 addon

PR #4166, opened 2022-02-02
scratchr2: fix z-index

apple502j:
Resolves #4161

Adds `z-index: 1;` to `.actions-wrap` so that it is over the commenter name (but below the modals)

PR #6890, opened 2023-11-25
Layer emoji picker over Report and Delete buttons

DNin01:
### Changes

Fixes a bug where the "Report" and "Delete" buttons are displayed on top of the emoji picker on scratchr2 pages when the `scratchr2` addon is enabled. This happens because the `scratchr2` addon sets their `z-index` property to 1, causing them to be layered over the emoji picker.

Emoji picker window with Report and Delete buttons from a comment displaying on top

This change increases the emoji picker's z-index to 2, fixing the problem.

### Tests

Tested in Edge 119.

### DIFF
--- a/addons/emoji-picker/userstyle.css
+++ b/addons/emoji-picker/userstyle.css
@@ -7,7 +7,7 @@
   border-radius: 8px;
   border: 2px solid rgba(0, 0, 0, 0.15);
   box-shadow: 0px 0px 5px #00000022;
-  z-index: 1;
+  z-index: 2;
   padding: 5px;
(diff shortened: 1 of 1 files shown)


DNin01:
Thanks @just-a-hriday for bringing this bug to my attention.

mxmou (approved):

Samq64 (approved):

mxmou:
> Do we have a link to the commit/PR that caused this regression?

#4166

WorldLanguages:
Oh, that's quite old.

PR #6894, opened 2023-11-26
editor-compact: fix backpack on costume tab

### LINKED ISSUE #6862: compact-editor: not enough room for backpack when on costumes tab
### What should happen
Backpack item names show as normal in costumes (via compact editor enabled)
### What actually happens
Item icons (costume, etc) in backpack are cut into half (roughly) and titles are not visible, only workarounds to it are zooming out (90%) or entering full screen.

### Steps to reproduce

1. Enable compact editor (if not enabled yet) and go into the project editor,
2. Go to costumes,
3. Open backpack (Items should be cut in half).

backpack
<sub>forgot i should use a screenshot of not just the costume pane probably</sub>
### Environment

Browser: Firefox (v119.0.1, 32 bit)
Scratch Addons version: v1.35.1
Operating system: Windows 10 (1366 x 768 screen resolution)

mxmou:
Resolves #6862

### Changes

Only changes the paint editor tool selector to a single column if there's enough space.

image image

### Reason for changes

To make space for the backpack on small screens.

### Tests

Tested on Edge and Firefox with different zoom levels and window sizes.

### DIFF
--- a/addons/editor-compact/userstyle.css
+++ b/addons/editor-compact/userstyle.css
@@ -364,8 +364,10 @@ body:not(.sa-columns-enabled) [class*="gui_tab-panel"] [class*="gui_extension-bu
 [class*="paint-editor_top-align-row"] {
   padding-top: 0.5rem;
 }
-[class*="paint-editor_mode-selector"] {
-  width: min-content;
+@media (min-height: 710px) {
+  [class*="paint-editor_mode-selector"] {
+    width: min-content;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


WorldLanguages (approved):
Tested succesfully in Chromium.
With my default screen size and zoom, I still have enough space for one column of vector tools.
Using Chrome Devtools' "device toolbar" set at Responsive 1300x640 the bug was fixed, as the vector tools now used 2 columns instead which left enough space for the backpack.

PR #6897, opened 2023-11-27
Specify version for Prettier action

### LINKED ISSUE #6882: Specify a version for Prettier
### Why this would be helpful

Our ["Format code with Prettier" workflow](https://github.com/ScratchAddons/ScratchAddons/blob/3a342fa4a27cc79f138df666a03d72b4a548a5ec/.github/workflows/format-code.yml) automatically checks the formatting of all the code in a branch when it is updated. On forks with actions enabled, any inconsistencies are corrected and committed automatically.

That's super useful to make sure the code that contributors are proposing is consistent with Prettier's unified style, but it isn't all that fun when the style changes and causes files that were consistent style-wise before to be modified with no forewarning, even ones that weren't touched by the pull request author. Take a moment to imagine hundreds of unrelated files being modified by Prettier automatically the moment you push commits to your pull request. Wouldn't you be mystified? I sure was when [that happened to me](https://github.com/ScratchAddons/ScratchAddons/pull/6259#commits-pushed-75ded5d).

In some cases, pull request authors have to wait until we fix the problem (I believe it's because our [branch protection rules](https://github.com/ScratchAddons/ScratchAddons/settings/branches) don't allow our action to push to our `master` branch, but I don't know for sure) and then merge `master` from upstream into their branch to fix the problem.

I have seen an instance of this not [one](https://github.com/ScratchAddons/ScratchAddons/pull/6362) but [two](https://github.com/ScratchAddons/ScratchAddons/pull/6868) times.

### My suggestion

To help us control these problems, we might want to specify a version of Prettier to use. This would prevent updates from causing all these headaches, though we might still want to update it from time to time. The nice thing about that is we might be able to inform contributors about the changes so they're not confused when this happens.

### Possible alternatives

_No response_

### Additional context

_No response_

DNin01:
Resolves #6882

### Changes

Sets the version of Prettier to use to 3.1.0. This will keep us on the current version of Prettier even if it is updated. We can always update manually.

### Reason for changes

Currently, Prettier updates can cause our code formatting style to change unexpectedly, causing disruptions for contributors and us. This doesn't mean we'll never update, but if we do:
1. Currently open pull requests will still have status checks fail, but since the version will also be hardcoded on their branches, their code's formatting won't change automatically until their authors merge `upstream/master` into their branches, which updates the formatting anyway.
2. We might be able to take precautions (like notifying contributors) before updating the version to reduce the number of headaches.

### Tests

Tested with GitHub Actions.

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -24,6 +24,7 @@ jobs:
         uses: ScratchAddons/prettier_action@master
         with:
           prettier_options: --write .
+          prettier_version: 3.1.0
           commit_message: Format code
         env:
           GITHUB_TOKEN: ${{ steps.generate-token.outputs.token }}
@@ -33,6 +34,7 @@ jobs:
         uses: ScratchAddons/prettier_action@master
         with:
           prettier_options: --write .
+          prettier_version: 3.1.0
           commit_message: Format code
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


DNin01:
This okay with you, @WorldLanguages @Hans5958?

3.1.0 is the [latest stable version](https://prettier.io/versions) at this time.

Hans5958 (approved):
LGTM. Would need another look by another reviewer.

WorldLanguages (approved):
We'll still have problems with new versions but this is still better.

PR #6900, opened 2023-11-28
Create a separate AudioContext for the sound library

### LINKED ISSUE #6847: [Minor] Project audio unpauses when opening sound library
### Description

If I have paused a project with music and open the sound library, the project audio resumes. Note that the rest is still paused, and the Pause button keeps the triangle shape; The only thing that unpauses is the audio.

I don't know if this is semi-intentional, anyways it's annoying.

### Steps to reproduce

1. Open and start a project containing music in the editor (like a "forever: play sound until the end" loop)
2. Click on Pause
3. Switch to the Sounds tab, hover on Add sound and click Choose a sound

_No response_

### Environment

Browser: Vivaldi 6.4.3160.41 (Stable channel) (64-bit) 
Scratch Addons version: 1.36.0-prerelease [8d22730](https://github.com/ScratchAddons/ScratchAddons/commit/8d2273099d8e261f4638458cb5339e7143f0b4e6)
Operating system: Windows 11 Pro 23H2 22635.2700

mxmou:
Resolves #6847

### Changes

Makes the sound library use a separate audio context if `debugger` or `pause` is enabled.

### Reason for changes

This allows the library to play sounds while the project is paused without resuming the project's audio.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/debugger/module.js
+++ b/addons/debugger/module.js
@@ -5,6 +5,8 @@ const STATUS_YIELD = 2;
 const STATUS_YIELD_TICK = 3;
 const STATUS_DONE = 4;
 
+const REACT_INTERNAL_PREFIX = "__reactInternalInstance$";
+
 let vm;
 
 let paused = false;
@@ -349,12 +351,12 @@ export const singleStep = () => {
   eventTarget.dispatchEvent(new CustomEvent("step"));
 };
 
-export const setup = (_vm) => {
+export const setup = (addon) => {
   if (vm) {
     return;
   }
 
-  vm = _vm;
+  vm = addon.tab.traps.vm;
 
   const originalStepThreads = vm.runtime.sequencer.stepThreads;
   vm.runtime.sequencer.stepThreads = function () {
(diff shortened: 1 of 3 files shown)


addons/debugger/module.js
```diff
@@ -407,4 +409,37 @@ export const setup = (_vm) => {
     }
     return count;
   };
+
+  // All instances of AudioEngine use the same AudioContext by default,
+  // which means that opening the sound library resumes the VM's context. See #6847.
+  // This can be fixed by creating a separate context from the sound library.
+  addon.tab.waitForElement("[class*='play-button_play-button_']").then(() => {
```
WorldLanguages, line 416: This needs to be optimized to avoid running `document.querySelector()` on every DOM changes.
We should check if any redux events are dispatched when opening the sound library. If not, we can pollute a relevant method that is called when the sound library is opened, and directly use querySelector there.
(I assume this waits until the sound library is opened, right? I haven't checked)

mxmou: `scratch-gui/modals/OPEN_MODAL` is probably the right Redux event.

WorldLanguages (approved):
Tested in Chromium

addons/debugger/module.js
WorldLanguages: I added that reduxEvent.

WorldLanguages:
@mxmou Possibly check one last time that everything works as expected, and merge this

PR #6902, opened 2023-11-29
editor-theme3: fix transparent inputs

### LINKED ISSUE #6806: Block input transparcy causes text glitch on focus
### Description

> When applying transparency to "block entry" with the block theme addon, when editing an entry, the text is glitched because of a small font size/spacing issue. It seems the bug is only present on custom blocks.
>
>How to reproduce:
>1. Activate the "custom block theme" addon and give it ~20% opacity
>2. Open the Scratch editor and create a new custom block, with an input entry
>3. Take an instance of the custom block and write a long string inside
>
>What should happen:
>* Nothing should go wrong
>
>What is happening:
>* There are two texts with the same characters overlapping each other - Probably the html input and the background text >when unselected
>
>Possible fix:
>* Switch "display: none" CSS attribute between the two so one is hidden when it's inactive
>* Keep the input and the text as the same element

### Steps to reproduce

I can reproduce without making a custom block.

1. Enable `editor-theme3`
2. Set the block input color to 20% opacity
3. Click a block input

20231024_09h13m16s_grim

### Environment

Browser: Chromium 117
Operating system: Linux

mxmou:
Resolves #6806

### Changes

Makes the HTML part of a focused block input invisible.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/editor-theme3/black_text.css
+++ b/addons/editor-theme3/black_text.css
@@ -6,9 +6,6 @@
 .blocklyEditableText > text {
   fill: var(--editorTheme3-inputColor-blackText);
 }
-.blocklyHtmlInput {
-  color: var(--editorTheme3-inputColor-blackText);
-}
 
 .blocklyDropDownDiv .goog-menuitem,
 .sa-contextmenu-colored .blocklyContextMenu .goog-menuitem .goog-menuitem-content {
--- a/addons/editor-theme3/theme3.css
+++ b/addons/editor-theme3/theme3.css
@@ -2,8 +2,11 @@
   fill: var(--editorTheme3-inputColor-text);
 }
 .blocklyHtmlInput {
-  background-color: var(--editorTheme3-inputColor);
-  color: var(--editorTheme3-inputColor-text);
+  /* Hide the HTML input so that only the SVG part is visible.
+     This is needed for transparent inputs. */
+  background-color: transparent;
+  color: transparent;
+  caret-color: var(--editorTheme3-inputColor-text);
 }
 
 /* Override Scratch's high contrast theme */


DNin01 (commented):
This makes the text selection highlight be layered over the text, causing a subtle text color change.

With invisible `.blocklyHtmlInput`:
Text with selection highlight layered above

With visible `.blocklyHtmlInput`:
Text with selection highlight layered below

`color: transparent` here is also overridden by [`black_text.css:9`](https://github.com/ScratchAddons/ScratchAddons/blob/22608151e3f3b5ebb2ca72c59e08053a7372aaba/addons/editor-theme3/black_text.css#L9-L11), which happens when the text color is set to Black.

mxmou:
> This makes the text selection highlight be layered over the text, causing a subtle text color change.
> 
> With invisible `.blocklyHtmlInput`: Text with selection highlight layered above
> 
> With visible `.blocklyHtmlInput`: Text with selection highlight layered below

Hiding the SVG part instead of the HTML input would fix that, but I don't think it's possible with just CSS.

> `color: transparent` here is also overridden by [`black_text.css:9`](https://github.com/ScratchAddons/ScratchAddons/blob/22608151e3f3b5ebb2ca72c59e08053a7372aaba/addons/editor-theme3/black_text.css#L9-L11), which happens when the text color is set to Black.

Fixed.

DNin01 (approved):

DNin01:
> Hiding the SVG part instead of the HTML input would fix that, but I don't think it's possible with just CSS.

Yeah, I figured it would be tricky. It's not a necessity.

WorldLanguages (approved):

PR #6906, opened 2023-11-29
custom-block-shape: fix empty inputs

### LINKED ISSUE #6667: custom-block-shape: does not support blocks with 'null' inputs
### Description

Blocks with 'null' inputs normally look like this:
image

Compared to blocks with normal inputs:
image

When 'Customizable block shape' is applied with the '2.0 Blocks' preset, this is what they look like:
image
image

Notice that the null inputs (1st image) do not scale with the block, making them stick out horrendously. 

### Steps to reproduce

1. Turn off all addons.
2. [Go to the editor of this project.](https://scratch.mit.edu/projects/560198467/editor)
3. Note the perfectly fine-looking blocks. (Aside from the obviously JSON edited inputs)
4. Turn on 'Customizable block shape' and apply the '2.0 Blocks' preset.
5. Observe the ugliness of the null inputs.
6. Refresh and realize it doesn't fix it.
7. Sob 

Fixing this would be useful to TurboWarp extension creators as null inputs are often used for blocks that should only accept changing values such as reporters and booleans.

### Environment

Browser: 
Scratch Addons version: 
Operating system:

mxmou:
Resolves #6667

### Changes

Makes the padding setting of the `custom-block-shape` addon affect empty round inputs.

### Reason for changes

Some TurboWarp extensions use this type of input:
image

It currently looks like this with `custom-block-shape`:
image

### Tests

Tested by opening [this project](https://scratch.mit.edu/projects/560198467/editor) on Edge and Firefox.
image

### DIFF
--- a/addons/custom-block-shape/userscript.js
+++ b/addons/custom-block-shape/userscript.js
@@ -123,24 +123,24 @@ export default async function ({ addon, console }) {
       BlockSvg.INPUT_SHAPE_HEXAGONAL_WIDTH = 12 * GRID_UNIT * multiplier;
       BlockSvg.INPUT_SHAPE_ROUND =
         "M " +
-        4 * GRID_UNIT +
+        4 * GRID_UNIT * multiplier +
         ",0" +
         " h " +
-        4 * GRID_UNIT +
+        4 * GRID_UNIT * multiplier +
         " a " +
-        4 * GRID_UNIT +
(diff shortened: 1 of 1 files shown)


WorldLanguages (approved):
Interesting way to fix it.

Samq64 (approved):

PR #6909, opened 2023-11-29
Option to execute double-clicked scripts

### LINKED ISSUE #3958: double-click to run scripts

Although sometimes people accidentally run scripts when they don't want to, forcing them to press `ctrl` every time they do want to is just too impractical.

Replacing crtl+clicking to run scripts with double-clicking.

Making it an option for the current addon or making it a new addon.

### Additional context

N/A

DNin01:
Resolves #3958

### Changes

Adds a setting to the `ctrl-click-run-scripts` addon that enables the ability to double click scripts to execute them, instead of having to hold Ctrl. This option is on by default.

Here's how it works:
- When you click on a block stack, a 400ms timer is started and a "double click" flag is raised. If the timer ends, the "double click" flag is dropped. The script does not run yet.
- When you click a block stack while the flag is up (double-click), the script runs. Also, the timer is restarted but the flag stays raised, which allows you to repeatedly execute script as fast as you normally would.

The behavior can be disabled through the setting and works in tandem with Ctrl-clicking.

### Reason for changes

Not allowing scripts to be executed when clicked can be useful to prevent accidents, but requiring the Ctrl key to be used is a bit cumbersome. Requiring a double click is a simple and more user-friendly solution.

### Tests

Tested with:
- Edge 119
- Scratch Addons v1.36.0-pre (https://github.com/ScratchAddons/ScratchAddons/commit/8d2273099d8e261f4638458cb5339e7143f0b4e6)
- scratch.mit.edu

### DIFF
--- a/addons/ctrl-click-run-scripts/addon.json
+++ b/addons/ctrl-click-run-scripts/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Ctrl+Click to run scripts",
-  "description": "Only run scripts on click if also holding the Ctrl key.",
+  "description": "Only execute clicked blocks and scripts if also holding the Ctrl key or, optionally, if double clicked.",
   "info": [
     {
       "text": "On macOS, use the Cmd key instead of the Ctrl key.",
@@ -15,6 +15,10 @@
     {
       "name": "apple502j",
       "link": "https://github.com/apple502j"
+    },
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
     }
   ],
   "dynamicEnable": true,
@@ -25,7 +29,20 @@
       "matches": ["projects"]
     }
   ],
+  "settings": [
+    {
+      "name": "Execute on double click",
+      "id": "double-click",
+      "type": "boolean",
+      "default": true
+    }
+  ],
   "tags": ["editor", "codeEditor"],
   "versionAdded": "1.15.0",
+  "latestUpdate": {
+    "version": "1.36.0",
+    "newSettings": ["double-click"],
+    "temporaryNotice": "Scripts can now be executed by double clicking."
+  },
   "enabledByDefault": false
 }
--- a/addons/ctrl-click-run-scripts/userscript.js
+++ b/addons/ctrl-click-run-scripts/userscript.js
@@ -17,10 +17,30 @@ export default async function ({ addon, console }) {
     }
   );
 
-  // Limits all script running to CTRL + click
+  let doubleClick = false;
+  let doubleClickTimeout;
+  function startDoubleClick() {
+    if (addon.settings.get("double-click")) {
+      doubleClick = true;
+      doubleClickTimeout = setTimeout(() => {
+        doubleClick = false;
+      }, 400);
+    }
+  }
+
+  // Capture clicks on scripts
   const newBlocklyListen = function (e) {
     if (!addon.self.disabled && e.element === "stackclick" && !ctrlKeyPressed) {
-      return;
+      if (doubleClick) {
+        // Allow script to execute on double click
+        clearTimeout(doubleClickTimeout);
+        startDoubleClick();
+        originalBlocklyListen.call(this, e);
+      } else {
+        // Start a double-click timeout and prevent script execution
+        startDoubleClick();
+        return;
+      }
     } else {
       originalBlocklyListen.call(this, e);
     }


addons/ctrl-click-run-scripts/userscript.js
```diff
@@ -17,10 +17,30 @@ export default async function ({ addon, console }) {
...
+      }, 400);
+    }
+  }
+
+  // Capture clicks on scripts
   const newBlocklyListen = function (e) {
     if (!addon.self.disabled && e.element === "stackclick" && !ctrlKeyPressed) {
-      return;
+      if (doubleClick) {
+        // Allow script to execute on double click
```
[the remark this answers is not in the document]
WorldLanguages: Yeah, it's good enough

BroJac5246:
> But as we suggested to add the danger tag to this addon, an alternative we could consider is to change it so that it doesn't change any Scratch editor nothing by default, unless you tweak its settings.

That would probably confuse people.

WorldLanguages:
> That would probably confuse people.

Okay then, I went a little off-topic, so forget about my suggestion for now. Let's focus on the double-clicking.

Secret-chest:
Why 400ms? Please use the system delay or make it a setting (in milliseconds, not with select).

Samq64:
It looks like there is a dedicated [double-click event](https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event) which I would assume does follow the system's setting. Would it make sense to use that here?

WorldLanguages:
> Please use the system delay

I'm not sure if there's a way to know the system delay as a number. I think our only option would be using the doubleclick event.

DNin01:
Mmm...yeah, I don't know if we can rely on `dblclick` events because the blocks kind of just swallow it and the event doesn't get triggered. It's interesting - you have to double click outside the workspace, on an empty area in the workspace, or on a block on the block palette for a `dblclick` listener on the document to fire. There might be a way around that that I don't know of, though.

addons/ctrl-click-run-scripts/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Ctrl+Click to run scripts",
-  "description": "Only run scripts on click if also holding the Ctrl key.",
+  "description": "Only execute clicked blocks and scripts if also holding the Ctrl key or, optionally, if double clicked.",
```
Samq64, line 3: Ctrl+Click still works with the setting enabled.
```suggestion
  "description": "Only execute clicked blocks and scripts if also holding the Ctrl key and, optionally, if double clicked.",
```

addons/ctrl-click-run-scripts/addon.json
```diff
@@ -25,7 +29,20 @@
       "matches": ["projects"]
     }
   ],
+  "settings": [
+    {
+      "name": "Execute on double click",
```
Samq64, line 34: Does this need to be a setting? How often will someone accidentally double-click something? I think having multiple ways to activate it without any settings is fine and the middle-click popup already does it. If this is done then I think it would make sense to also rename the addon to use double-click.

DNin01: One way making a setting for double-clicking to run is that it could make the double-click feature more discoverable.

Samq64:
I think Ctrl+Clicking could be dropped completely in favour of double-clicking.

BroJac5246:
> I think Ctrl+Clicking could be dropped completely in favour of double-clicking.

If we do that then there should be a temporary pop-up/alert when someone tries the old way.

BroJac5246:
Has this been tested for mobile compatibility?

DNin01:
> I think Ctrl+Clicking could be dropped completely in favour of double-clicking.

While double clicking may be the better option, I think some people would still prefer to use clicking while holding Ctrl as the way to execute scripts, so I would keep that as an _alternative_ option, while primarily suggesting to double click to execute scripts.

We could even make a setting so users could choose between double clicking or Ctrl-clicking. The latter would be the default behavior for existing users and it would default to double clicking for users who haven't used the addon before.

Secret-chest:
If reading the system delay is not an option, can we make it customisable?

addons/ctrl-click-run-scripts/userscript.js
```diff
@@ -13,15 +13,35 @@ export default async function ({ addon, console }) {
...
+        clearTimeout(doubleClickTimeout);
+        startDoubleClick();
+        originalBlocklyListen.call(this, e);
+      } else {
+        // Start a double-click timeout and prevent script execution
+        startDoubleClick();
+        return;
```
DNin01: Unnecessary `return`?

addons/ctrl-click-run-scripts/addon.json
DNin01, line 1: This addon should probably be renamed to "Double-click to run scripts" and then holding the Ctrl key could be an option that defaults to on for existing users.

DNin01:
I came up with a few possible presentations of this addon. Anyone have a favorite?

Customizable action to run scripts addon settings
Double click to run scripts addon settings (1)
Disable click to run scripts addon settings

DNin01:
> @WorldLanguages: let's change this addon to "customizable method to run blocks" (just kidding [...])

Just like I did in that screenshot!

Anyway, any thoughts on what to name it and what the settings should be, @Samq64 and @WorldLanguages? I will at least be renaming it to _something_, because Ctrl-clicking is more awkward than double clicking.

DNin01:
> > But as we suggested to add the danger tag to this addon, an alternative we could consider is to change it so that it doesn't change any Scratch editor nothing by default, unless you tweak its settings.
>
> That would probably confuse people.

The "customizable action to run scripts" proposal I just made could be coded to have the following behavior when enabled and set to "run by clicking once or twice":

- On a single click, the addon waits for a double click, but if no second click happens, the script runs anyway.
- On a double click, the script runs.

That behavior is consistent with Scratch 1.4 and 2.0. In Scratch 3.0, double-clicking a script activates it twice instead.

Samq64:
In general I like the wording of the second one with the reporter setting from the first.

DNin01:
I have done quite a bit of work here. Check the PR description for details on the new behavior.

I may change the name and description again later. I'll also try to add a setting to allow lone reporter blocks to be activated with any type of click, as I teased in one of my screenshots.

addons/ctrl-click-run-scripts/addon.json
```diff
@@ -25,7 +23,43 @@
...
+        {
+          "id": "double",
+          "name": "Clicking twice"
+        },
+        {
+          "id": "ctrl",
+          "name": "Ctrl-clicking"
+        }
+      ],
+      "default": "single"
```
DNin01: I'm not sure how to default this setting to `"ctrl"` for existing addon users who enabled this addon. The `chrome.runtime.onInstalled` event's details have a `previousVersion` property which we could leverage, but I'm not quite sure how to work it into our setting migration procedure.

We could just make the default option `"ctrl"` for the next release and then change it to `"single"` again for the release afterward. That way, the behavior doesn't change for existing users of this addon (but also for existing non-users).

WorldLanguages: > I'm not sure how to default this setting to `"ctrl"` for existing addon users who enabled this addon.

In the past we've hardcoded some stuff like that in the get-addon-settings.js file

mxmou: #7632 would probably make this easier to implement.

mxmou:
> * Click once: Scripts run on a single click or if clicked three times or more quickly.

Since this is the default, I think it would make more sense for it to behave the same way as Scratch, where clicking the script more than once runs it that many times. Allowing multi-clicks makes sense if users don't know that a single click is enough, but I believe most of our users have been using Scratch 3 for a while and therefore already know that.

DNin01:
> > * Click once: Scripts run on a single click or if clicked three times or more quickly.
> 
> Since this is the default, I think it would make more sense for it to behave the same way as Scratch, where clicking the script more than once runs it that many times. Allowing multi-clicks makes sense if users don't know that a single click is enough, but I believe most of our users have been using Scratch 3 for a while and therefore already know that.

This would be fine except that the default setting would then have no purpose—functioning as if the addon wasn't enabled.

Some users who still heavily use old versions of Scratch (I don't know why they would, but there's gotta be someone) would find it nice to double click and have a script run once.

Another possible default option would be to have scripts not be able to be triggered at all... which is probably not the best idea...

DNin01:
What if none of the settings were the default option? Then the user would have to explicitly choose an option for the addon to do anything. Do we... support select settings without a default?

mxmou:
> This would be fine except that the default setting would then have no purpose—functioning as if the addon wasn't enabled.

I think that's fine. Most "customizable XYZ" addons work like that.

mxmou:
I've just realized that the simplest way to present this addon might be with no select setting - when the addon is enabled, single clicks are disabled but double click and Ctrl+click both work. This way, one can use their preferred method without choosing it explicitly in the settings.

This implementation would also be the best for touch screen users: currently, a user with a touch screen device can't use Ctrl+click, even on other devices, because of sync.

DNin01:
I've removed the options and now you can either Ctrl-click or double click to activate scripts, but the focus of the addon is on double clicking.

Secret-chest:
I like the removal of options, but you should still mention that control-click works.

PR #6915, opened 2023-11-30
New-ish addon: Redirect to standard forums (from mobile forums 404)

### LINKED ISSUE #6910: Remove the "redirect to standard forums" addon (&/or inclusions of the mobile forums in addons)
### Why this would be helpful

The mobile forums were removed (and even asked about in the forums) after the 4-5 day long maintenance lap, the addon is basically useless now since [any attempt to go to the mobile forums now gives you a 404](scratch.mit.edu/discuss/m/).

### My suggestion

Just remove the addon since it has no use anymore, although it might be best to ask contact-us before doing so incase it was only removed temporarily for whatever reason that exists.
EDIT:
> Reminder: I think one or more addons try to run on the mobile forums, that no longer exist

Since heavily related I think I can still edit and include this, however if not just let me know, but either way removing them if the removal of the mobile forums isn't temporary (kinda weird of the mobile forums did get removed only temporarily) could probably optimize a bit since they basically run on nothingness (eg. necropost highlighter is just gonna uselessly run on /m/* (disincluding the mobile page), however it has nothing to highlight anymore unless I'm also reading the code wrong lol)
### Possible alternatives

1. Move it to konami code (though that'd be an odd choice)
2. Wait and see if it's only a temporary removal (best guess would be asking contact-us, though I doubt this is temporal)
3. Add a note to addons including it that it no longer functions and may be deleted sometime soon (basically the same as 2 but with adding a note to all the addons that also run on the mobile forums)

### Additional context

https://scratch.mit.edu/discuss/topic/726596/ ("Were the mobile forums supposed to be removed?")
mobile leads to 404

iqnite:
Resolves #6910
Closes #6917 

### Changes

- ~~Remove the `redirect-mobile-forums` addon~~
- **New!** Attempt to redirect to the standard forums from 404 pages caused by `discuss/m/`
- Remove scripts that were exclusively for mobile devices
- Remove CSS rulesets that were exclusively for mobile devices
- **New!** Make the addon enabled by default
- ~~**New!** Change the addon's ID to `smart-404`~~

### Reason for changes

Scratch no longer supports the `discuss/m/` forums, so why should Scratch Addons support them? Still, there are links to the mobile forums, which cause 404 errors.

### Tests

Redirecting from 404 pages works.
I broadly tested the other addons where I removed the rulesets and scripts and I didn't notice any bugs. (I wrote this addon at 10 pm, so if anybody else could take a closer look at them?)
For the same reason, I kept the `dark-www` settings for mobile forums, as I'm not sure if it would still work.


DNin01:
> Just as a note: It should probably be worth seeing if it was only a temporary removal, though it might be unlikely

Yeah, there is a small chance that they might come back soon, but my guess is they are gone. Wait a few weeks, perhaps?

Chiroyce1:
If they have decided on removing it entirely, it only makes this addon more useful by redirecting any old mobile links either on Scratch itself or more commonly Google search, see #6917

iqnite:
I know, I set up this PR to be ready then. But the addon could also be modified to redirect to the forums instead of showing a 404 page when people click on a mobile forum link.

Samq64:
Could `redirect-mobile-forums` be re-added? Also the [forum match](https://github.com/ScratchAddons/ScratchAddons/blob/26dac9a5f4d3dba31fa1c436129de65a9548b3d6/background/get-userscripts.js#L344) can be simplified.

iqnite:
**Update:** Based on user feedback, I've added back the addon, and augmented its functionality so that it automatically redirects from 404 pages that were invoked from mobile forum links.
It has the potential to become a "Smart 404" addon... (flashbacks from SA Nightly)

addons/redirect-mobile-forums/addon.json
```diff
@@ -1,6 +1,13 @@
 {
   "name": "Redirect to standard forums",
   "description": "Redirects you to the standard (\"main\") forums when you click a link to the mobile forums.",
+  "info": [
+    {
+      "id": "mobile-forum-deprecation-warning",
```
Samq64, line 6: Is a notice necessary I think updating the description should be enough.

Samq64 (commented):
#6917I was actually a PR for the `redirect-mobile-forums` changes, I just wanted it added back the way it was so there wouldn't be any conflicts, but now that it's done it could also just be closed when this is merged instead.

addons/redirect-mobile-forums/userscript.js
```diff
@@ -1,3 +1,4 @@
 export default async function ({ addon, console, msg }) {
-  window.location.replace(window.location.href.replace("m/", ""));
+  if (window.location.href.includes("https://scratch.mit.edu/discuss/m/"))
```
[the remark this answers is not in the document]
iqnite: Is scratch.mit.edu automatically supposed?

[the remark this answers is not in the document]
iqnite: I tried it and it doesn't work...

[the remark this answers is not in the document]
iqnite: The problem is that 404 pages have a weird URL. They're shown as `/discuss/m`, but they're actually `/404`.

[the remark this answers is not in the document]
DNin01: Try this:
```suggestion
  if (window.location.pathname.startsWith("/discuss/m/"))
```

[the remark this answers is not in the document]
Samq64: Yeah I just accidentally forgot to type `pathname`.

iqnite:
Should we change the addon's ID to `smart-404`? That way, it would be enabled by default for new and existing users.

WorldLanguages:
> Should we change the addon's ID to `smart-404`? That way, it would be enabled by default for new and existing users.

I wouldn't name it `smart-404` unless we actually plan to implement a smart 404 feature. You can consider just adding `-v2` at the end of the addon name.

iqnite:
But if somebody wants to build a smart 404 mechanism for other pages, they can it on this addon.

BroJac5246:
> But if somebody wants to build a smart 404 mechanism for other pages, they can it on this addon.

If an addon's scope is expanded, it generally keeps the ID that refers to the old function. Anticipating a Smart 404s addon and changing the ID to that would not follow this custom, so the ID shouldn't be changed unless it is determined (before this or other similar PRs are merged) that the addon's scope should be expanded.

Samq64 (approved):

addons/redirect-mobile-forums-v2/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Redirect to standard forums",
-  "description": "Redirects you to the standard (\"main\") forums when you click a link to the mobile forums.",
+  "description": "Redirects you to the standard (\"main\") forums when you click a link to the mobile forums, avoiding Page not found errors.",
```
DNin01, line 3: A description suggestion I proposed on #6917:

```suggestion
  "description": "Redirects you to the standard (\"main\") view of the forums when you follow a link to the mobile view of the forums (which doesn't work anymore)",
```

iqnite: It should be clear that you get an error if you don't enable this addon.

DNin01: Like this?
```suggestion
  "description": "Redirects you to the standard (\"main\") view of the forums when you follow a link to the mobile view of the forums, avoiding Page not found errors.",
```

DNin01: > I think it should be visit instead of follow

"Visit" could work but may as well be changed from "visit a link to" to just "visit". Other alternative verbs include "open" and "go to".

addons/necropost-highlighter/userscript.js
```diff
@@ -29,13 +29,11 @@ export default async function ({ addon, global, console, msg }) {
...
-  const isMobileSite = /^\/discuss\/m\//.test(location.pathname);
-  // tableBodyNode will be null on mobile and non-table search page:
+  // tableBodyNode will be null on non-table search page:
   // https://scratch.mit.edu/discuss/search/?action=show_user&show_as=posts
-  const tableBodyNode = document.querySelector("tbody"); // null on mobile
+  const tableBodyNode = document.querySelector("tbody");
 
-  if (!tableBodyNode && !isMobileSite) {
+  if (!tableBodyNode) {
     // Standard site page without a table. Don't run on this page.
```
DNin01, line 37: This comment could be simplified as we don't have to account for the mobile forums anymore:
```suggestion
    // No table on this page. Don't run.
```

DNin01 (approved):
I should've probably tested this a little more, but I don't want to keep us waiting. This looks good enough.

WorldLanguages:
Sorry I forgot to move the milestone. Strings are closed right now, so we can't add new addons (technically it's not new, but the addon ID is new, so all translations are lost)

Samq64:
Is this ready to merge then?

WorldLanguages:
I think we forgot to update the regexes on addons/necropost-highlighter/addon.json that still consider the mobile forums

WorldLanguages:
> Could `redirect-mobile-forums` be re-added? Also the [forum match](https://github.com/ScratchAddons/ScratchAddons/blob/26dac9a5f4d3dba31fa1c436129de65a9548b3d6/background/get-userscripts.js#L344) can be simplified.

That too

PR #6917, opened 2023-12-01
Enable redirect-mobile-forums by default

### LINKED ISSUE #6910: Remove the "redirect to standard forums" addon (&/or inclusions of the mobile forums in addons)

Samq64:
Closes #6910

### Changes

Enables the Redirect to standard forums addon by default and adjusts the name and description to indicate that the links are broken. Also fixes a race conditon in the addon by also running on `/404`.

### Reason for changes

All links to the mobile forums such as from google search results are broken so there's no reason not to redirect users to the correct URL by default.

### Tests

Tested on Firefox 120

### DIFF
--- a/addons/redirect-mobile-forums/addon.json
+++ b/addons/redirect-mobile-forums/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "Redirect to standard forums",
-  "description": "Redirects you to the standard (\"main\") forums when you click a link to the mobile forums.",
+  "name": "Fix mobile forum links",
+  "description": "Redirects you to the standard (\"main\") forums when you click a broken link to the mobile forums.",
   "credits": [
     {
       "name": "Weredime (9gr)",
@@ -10,11 +10,11 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["https://scratch.mit.edu/discuss/m/*"],
+      "matches": ["https://scratch.mit.edu/discuss/m/*", "https://scratch.mit.edu/404"],
       "runAtComplete": false
     }
   ],
   "versionAdded": "1.16.0",
   "tags": ["community", "forums"],
-  "enabledByDefault": false
+  "enabledByDefault": true
 }
--- a/addons/redirect-mobile-forums/addon.json
+++ b/addons/redirect-mobile-forums/addon.json
@@ -1,6 +1,6 @@
 {
-  "name": "Fix mobile forum links",
-  "description": "Redirects you to the standard (\"main\") forums when you click a broken link to the mobile forums.",
+  "name": "Redirect to standard forums",
+  "description": "Redirects you to the standard (\"main\") view of the forums when you follow a link to the mobile forums (which doesn't work anymore).",
   "credits": [
     {
       "name": "Weredime (9gr)",
--- a/addons/redirect-mobile-forums/userscript.js
+++ b/addons/redirect-mobile-forums/userscript.js
@@ -1,3 +1,5 @@
-export default async function ({ addon, console, msg }) {
-  window.location.replace(window.location.href.replace("m/", ""));
+export default async function ({ addon, console }) {
+  if (window.location.pathname.startsWith("/discuss/m/")) {
+    window.location.replace(window.location.href.replace("m/", ""));
+  }
 }
--- a/addons/redirect-mobile-forums/userscript.js
+++ b/addons/redirect-mobile-forums/userscript.js
@@ -1,3 +1,3 @@
-export default async function ({ addon, console }) {
+export default async function ({ addon, console, msg }) {
   window.location.replace(window.location.href.replace("m/", ""));
 }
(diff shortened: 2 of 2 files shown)


addons/redirect-mobile-forums/addon.json
```diff
@@ -10,11 +10,11 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["https://scratch.mit.edu/discuss/m/*"],
+      "matches": ["https://scratch.mit.edu/discuss/m/*", "https://scratch.mit.edu/404"],
```
WorldLanguages, line 13: If the userscript runs in 404, it needs to be changed so that it only does stuff if the URL path starts with `/discuss/m/` or is exactly `/discuss/m`

DNin01:
Nice - I was just about to suggest this idea on one of the active PRs!

addons/redirect-mobile-forums/userscript.js
```diff
@@ -1,3 +1,5 @@
-export default async function ({ addon, console, msg }) {
-  window.location.replace(window.location.href.replace("m/", ""));
+export default async function ({ addon, console }) {
+  if (window.location.pathname.startsWith("/discuss/m/")) {
```
WorldLanguages, line 2:
```suggestion
  if (window.location.pathname.startsWith("/discuss/m/") || window.location.pathname === "/discuss/m") {
```

addons/redirect-mobile-forums/addon.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "Redirect to standard forums",
-  "description": "Redirects you to the standard (\"main\") forums when you click a link to the mobile forums.",
+  "name": "Fix mobile forum links",
+  "description": "Redirects you to the standard (\"main\") forums when you click a broken link to the mobile forums.",
```
DNin01, line 3: Not completely sure about changing the name/description to reflect the broken mobile view. This wording seems to imply that at least some links are broken when in fact, _all_ links to the mobile forums are broken. Plus, the point of this addon is to redirect you to the standard forums nonetheless, so it makes sense to focus on that functionality. I agree on everything else, though.

addons/redirect-mobile-forums/userscript.js
Samq64: `startsWith()` also works if it's the full string and since it replaces `m/` checking for a path without a trailing slash isn't very useful without updating the replace.

addons/redirect-mobile-forums/addon.json
DNin01: > mention that all links to the mobile forums don't work anymore?

Yeah, it would actually probably be a good idea to still note that. Here's my proposal:
```suggestion
  "name": "Redirect to standard forums",
  "description": "Redirects you to the standard (\"main\") view of the forums when you follow a link to the mobile forums (which don't work anymore).",
```

WorldLanguages: @DNin01 Sounds good but you mixed singular and plural in the last sentence

addons/redirect-mobile-forums/userscript.js
WorldLanguages: Right

addons/redirect-mobile-forums/addon.json
WorldLanguages: ugh good point

DNin01 (commented):
Do we need to do anything else for this to be automatically enabled for _existing_ users?

WorldLanguages:
> Do we need to do anything else for this to be automatically enabled for _existing_ users?

Yes, the easiest way is to change the addon ID.

RedGuy12:
> Do we need to do anything else for this to be automatically enabled for _existing_ users?

relevant to #5431

iqnite:
Possible transfer to #6915

Samq64:
Since it's been a few weeks it looks like search engines have already removed the dead links and I can find one anymore.

PR #6920, opened 2023-12-01
`collapse-footer`: Option to click to expand/collapse

### LINKED ISSUE #6833: `collapse-footer`: Option to click to expand/collapse
### Why this would be helpful

Sometimes the user can accidentally hover over the footer area, causing it to expand unexpectedly, which can be jarring.

It can also hinder touch screen compatibility to rely on hovering to hide or show a visual element.

### My suggestion

A solution for users who are getting tired of this might be to add an option to change how you expand the footer. The setting could be called "expand on" with the options "hover" and "click". When "click" is selected, the footer would be able to be expanded by clicking it and collapsed by clicking outside of it, instead of having the state change when the mouse enters or leaves it.

### Possible alternatives

Require the mouse to be hovering over the footer area for a moment before it expands.

Touches could also be accounted for in tandem with hover detection.

### Additional context

_No response_

iqnite:
Resolves #6833

### Changes

Add a setting to the Sticky footer addon that allows to control whether the footer should toggle on hover (current behavior) or on click.

### Reason for changes

When you accidentally hover over the footer area, the big footer shows up, and you need to move the mouse half the screen up to make it disappear. Also, you might accidentally click on a footer link and leave the current page.

### Tests

Tested on Vivaldi 6.4.3160.47, also with touchscreen emulation.

### DIFF
--- a/addons/collapse-footer/userscript.js
+++ b/addons/collapse-footer/userscript.js
@@ -10,7 +10,7 @@ export default async function ({ addon, console }) {
     }, 200);
   }
 
-  function instantCollapseFooter(event) {
+  function instantCollapseFooter(event, footer) {
     // Only hide if the click is outside the footer
     if (!footer.contains(event.target)) footer.classList.remove("expanded");
   }


addons/collapse-footer/addon.json
```diff
@@ -56,6 +56,12 @@
         "addonEnabled": ["infinite-scroll"]
       },
       "default": false
+    },
+    {
+      "name": "Toggle on click",
```
Samq64, line 61: Would it make sense to turn this into a select?

iqnite: It only has 2 options, it's not subject to get more, and it's consistent with the other settings.

Secret-chest:
Toggle on click should be turned into the opposite, Open on hover.

mxmou:
I don't think a Boolean setting makes sense for this - it isn't clear what the behavior is when it's turned off. There's nothing wrong with a select setting with two options.

addons/collapse-footer/addon.json
```diff
@@ -56,6 +56,22 @@
...
+        {
+          "id": "hover",
+          "name": "Hover"
+        },
+        {
+          "id": "click",
+          "name": "Click"
+        }
+      ],
+      "default": "hover"
```
DNin01, line 74: Should this setting be placed above `Only on infinite scrolling pages`?

addons/collapse-footer/userscript.js
```diff
@@ -29,16 +35,19 @@ export default async function ({ addon, console }) {
...
     }
   });
 
-  footer.addEventListener("mouseout", () => {
-    collapseTimeout = setTimeout(() => {
-      footer.classList.remove("expanded");
-    }, 200);
-  });
+  if (addon.settings.get("mode") === "click") {
+    document.addEventListener("click", (event) => {
```
DNin01, line 46: This could be changed to a `mousedown` event to collapse the footer as soon as you press your mouse button somewhere outside the footer, which might be more responsive if, for example, you long click or drag. It also triggers on right click and middle click.
```suggestion
    document.addEventListener("mousedown", (event) => {
```

_Expanding_ on `click` seems fine, though.

addons/collapse-footer/userscript.js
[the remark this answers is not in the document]
iqnite: The other settings don't support it either, so...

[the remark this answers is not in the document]
DNin01: The thing with the "only on infinite scrolling pages" setting is that the Infinite Scrolling addon is not dynamically toggleable, so we can't have that setting dynamically adjustable, or at least it wouldn't make sense. The setting for the input method doesn't have this problem.

addons/collapse-footer/addon.json
iqnite: Done!

addons/collapse-footer/userscript.js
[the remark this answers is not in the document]
iqnite: OK, I'm working on it.

Samq64 (approved):

PR #6923, opened 2023-12-02
Use `event.key` instead of `code` and `keyCode`

### LINKED ISSUE #6455: Keyboard shortcuts may not work on alternate keyboard layouts
### Description

I haven't tested this because I only use a QWERTY keyboard, but apparently, the `code` property of `KeyboardEvent` is not affected by your keyboard layout. Instead, `code` basically returns the QWERTY-equivalent character of the key you pressed on your keyboard, which may useful if you are making a web game, since you can tie controls to physical positions on the keyboard instead of characters on the keyboard.

You can learn more about it on MDN: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

Potentially affected scripts: [Code search (7 items)](https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons+%28e.code+OR+event.code%29&type=code)

### Steps to reproduce

Paste this in your browser console:
```js
window.addEventListener(
  "keydown",
  (e) => {
    console.log(e.code);
  }
);
```

Then, with an alternate keyboard layout, press a key that is different on a QWERTY keyboard (such as 'A' on AZERTY. You will see "KeyQ" in the console).

_No response_

### Environment

Browser: Edge 115
Operating system: Windows 11

### LINKED ISSUE #6456: Some keyboard shortcuts do not work while Caps Lock is on
### Description

Here's an example:

When Caps Lock is on, you can't expand the Find bar from `editor-find-bar` with Ctrl+F.

This is probably because holding down Shift or enabling Caps Lock changes the value reported by the `key` property of `KeyboardEvent`, and since JavaScript is case-sensitive, the condition for the if statement for focusing the find bar isn't fulfilled.

[There are also more places where we use `e.key`.](https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons+%28e.key+OR+event.key%29&type=code)

We could fix this by using `e.code` but I would recommend against that because it will cause more problems: #6455

### Steps to reproduce

1. Enable `editor-find-bar`.
2. Turn on Caps Lock.
3. Press Ctrl+F.

Fun fact, Scratch actually has [a similar bug](https://scratch.mit.edu/discuss/topic/540257/).

### Environment

Browser: Edge 115
Operating system: Windows 11

mxmou:
Resolves #6455
Resolves #6456

### Changes

Replaces most uses of `event.code` and `event.keyCode` with `event.key` (`event.key.toLowerCase()` for letters). Code that interacts with Blockly wasn't changed.

### Reason for changes

`event.code` represents the physical key, not the character associated with it. `event.keyCode` is deprecated and depends on the browser and platform.

### Tests

Tested on Edge and Firefox. The find bar shortcut now works with Caps Lock. The pause shortcut now works on keyboard layouts other than QWERTY - I tested this by changing the layout to Dvorak in Windows settings.

### DIFF
--- a/addons/2d-color-picker/paint-editor.js
+++ b/addons/2d-color-picker/paint-editor.js
@@ -129,10 +129,10 @@ export default async ({ addon, console, msg }) => {
     saColorLabel.appendChild(saColorLabelName);
     saColorLabel.appendChild(saColorLabelVal);
 
-    let keyPressed = -1;
+    let keyPressed = null;
     let originalPos = { x: 0, y: 0 };
-    window.addEventListener("keydown", (e) => (keyPressed = e.keyCode));
-    window.addEventListener("keyup", () => (keyPressed = -1));
+    window.addEventListener("keydown", (e) => (keyPressed = e.key));
+    window.addEventListener("keyup", () => (keyPressed = null));
 
     let origHue = 0;
     let el = null;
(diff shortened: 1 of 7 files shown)


WorldLanguages (approved):
Made sure all key codes are correct by using keycode.info

WorldLanguages:
It could be worth mentioning in the docs that we prefer `event.key` over all other ways to detect which key was pressed.

Samq64:
> It could be worth mentioning in the docs that we prefer `event.key` over all other ways to detect which key was pressed.

https://github.com/ScratchAddons/website-v2/pull/405

PR #6924, opened 2023-12-02
scratchr2: resize profile rows after dynamic enable/disable

### LINKED ISSUE #6379: Loading a profile page with `scratchr2` on and then dynamically disabling it cuts off studio rows
### Description

When Scratch 2.0 → 3.0 is enabled and you open a profile page and then dynamically disable the addon, the content on the "Studios I Curate" and "Studios I'm Following" rows get cut off.

Also, if you open a profile page with Scratch 2.0 → 3.0 disabled and then dynamically enable it, extra padding is added between the scroll bar and the content, so it seems that entire area doesn't take Dynamic Enable or Dynamic Disable into account like everything else.

I found that element.style.height on an element with the class name `box-content slider-carousel horizontal sliderCarousel` is what causes this to happen. To fix this, the height should be updated on the fly.

### Steps to reproduce

1. Enable `scratchr2`.
2. Open a profile page.
3. Dynamically disable `scratchr2`. The content on the "Studios I Curate" rows are cut off.
4. Reload the profile page.
5. Dynamically enable `scratchr2`. There's an unnatural padding below the content, pushing the scroll bar lower than it should be.

Scratch studio rows cut off

### Environment

Browser: Edge 114
Scratch Addons version: 1.34.0-pre (https://github.com/ScratchAddons/ScratchAddons/commit/ba107cdf6fcc8518d857d979f8d81ca5a9303978)
Operating system: Windows 11

mxmou:
Resolves #6379

### Changes

Resizes rows on the profile page when Scratch 2.0 → 3.0 is enabled or disabled.

### Tests

Tested on Edge and Firefox.


WorldLanguages (approved):
Tested in Chromium.
Sorry for the wait

PR #6931, opened 2023-12-06
New addon: Download project from project page

Jazza-231:
Resolves issue in my brain and I think a few times its been suggested in community server

### Changes

Adds button to download the sb3 from the project page
### Reason for changes

yes
### Tests

Works fine on brave

I would like to get opinions on how it should be styled, colour, just icon, etc, bc its a bit wide rn


addons/save-in-project-page/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Save the project from the project page",
```
Samq64, line 2:
```suggestion
  "name": "Project page download button",
```

addons/save-in-project-page/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "Allows you to save the sb3 file of the project from the project page, without pressing \"See inside\"",
```
Samq64, line 3:
```suggestion
  "description": "Allows you to save the sb3 files of projects from the project page, without opening the editor.",
```

addons/save-in-project-page/userscript.js
```diff
@@ -0,0 +1,21 @@
...
+
+  const downloadButton = document.createElement("button");
+  downloadButton.innerHTML = "Download project";
+  downloadButton.onclick = download;
+  downloadButton.classList = "button sa-download-button";
+
+  const container = document.querySelector(".project-buttons");
+
+  const seeInsideButton = await addon.tab.waitForElement(".see-inside-button");
+  container.insertBefore(downloadButton, seeInsideButton);
```
Samq64, line 20: There's an API that can be used here to ensure the elements always load in the same order.
```suggestion
  addon.tab.appendToSharedSpace({ space: "beforeRemixButton", element: downloadButton, order: 2 });
```

Samq64 (commented):
I think the button needs to be re-added when clicking back to the project page from the editor.

addons/save-in-project-page/userscript.js
```diff
@@ -0,0 +1,21 @@
+import downloadBlob from "../../libraries/common/cs/download-blob.js";
+
+export default async function ({ addon, console }) {
+  const vm = addon.tab.traps.vm;
+
+  async function download() {
+    const project = await vm.saveProjectSb3();
+    const title = await addon.tab.waitForElement(".project-title");
+    downloadBlob(`${title.innerHTML}.sb3`, project);
```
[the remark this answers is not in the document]
DNin01: `addon.tab.redux.state.scratchGui.projectTitle` works, but seemingly only when you're in the editor.

Jazza-231:
@Samq64 Nice catch with it needing to be added back 👍

addons/save-in-project-page/userscript.js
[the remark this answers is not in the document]
WorldLanguages: innerText seems good enough as a default name of the file

addons/save-in-project-page/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Project page download button",
+  "description": "Allows you to download projects directly from the project page, without having to open the editor first.",
+  "tags": ["community", "projectPlayer", "recommended"],
+  "versionAdded": "1.35.0",
```
Samq64, line 5:
```suggestion
  "versionAdded": "1.36.0",
```

addons/save-in-project-page/userscript.js
WorldLanguages: Remember own projects don't have a remix button, unless the dedicated "remix own projects" addon is enabled

addons/save-in-project-page/userscript.js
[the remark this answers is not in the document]
DNin01: > innerText seems good enough as a default name of the file

Yeah, it's not like we've done this before, have we?

> what is wrong with using innertext? it works fine for me?

As I [said](https://github.com/ScratchAddons/ScratchAddons/pull/6931#discussion_r1417983667)...

> innerHTML returns a string of the HTML for the project title input box when you have created the project. innerText and textContent both return an empty string in the same case.

Let me show you:

innerHTML of read-only project title
innerHTML of editable project title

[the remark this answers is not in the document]
Jazza-231: oh, my bad I didnt realise it was an input, well is there an API for checking if you are the project creator? because to get around this ill just get the value of the input instead

addons/save-in-project-page/userscript.js
Jazza-231: well the no matter what i think it should be next to the see inside project button, that way if there's no remix button, its fine, and if there is a remix button (someone else's project or with the addon) the two purple buttons are next to each other

Jazza-231: working on that now 👍

addons/save-in-project-page/userscript.js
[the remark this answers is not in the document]
Jazza-231: fixed by [commit](https://github.com/ScratchAddons/ScratchAddons/pull/6931/commits/d60ac75a2484524d00d93ec38f1cc0761d3ecb3f)

DNin01 (dismissed):
I tested downloading from the project page versus the editor and it seems fine - project files appeared to match exactly from my inspection testing!

There is, however, a bug where the addon is sometimes unable to recognize the project creator that we should resolve.

addons/download-button/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+  const vm = addon.tab.traps.vm;
+  const { redux } = addon.tab;
+
+  await redux.waitForState((state) => typeof state.session.session?.user === "object");
+  const username = await addon.auth.fetchUsername();
+
+  const projectAuthor =
+    (await redux.state.preview.status.project) === "FETCHED"
+      ? redux.state.preview.projectInfo.author?.username
+      : "Can not find project author";
```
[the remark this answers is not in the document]
DNin01: Oh, I see. At some point, `redux.state.preview.status.project` returns "FETCHING", which continues execution.

DNin01:
Tested on an unshared project.

addons/download-button/userscript.js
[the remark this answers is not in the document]
Jazza-231: It could try a set amount of times with a short delay between each after it has returned FETCHING to check when it has returned FETCHED?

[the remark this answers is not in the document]
mxmou: You can use `waitForState()`:
```js
await redux.waitForState((state) => state.preview.status.project === "FETCHED");
```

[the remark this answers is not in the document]
Jazza-231: Ooh sweet thanks

DNin01 (approved):
Looks good to me!

addons/download-button/userscript.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const downloadButton = document.createElement("button");
+  downloadButton.innerText = "Download";
+  downloadButton.onclick = download;
+  downloadButton.classList = "button action-button sa-download-button";
+
+  function addbutton() {
+    addon.tab.waitForElement(".flex-row .subactions", { markAsSeen: true });
+    console.log(shared);
```
Samq64, line 32:
```suggestion
```

addons/download-button/userscript.js
```diff
@@ -0,0 +1,46 @@
...
+  const { redux } = addon.tab;
+
+  await redux.waitForState((state) => typeof state.session.session?.user === "object");
+  const username = await addon.auth.fetchUsername();
+
+  await redux.waitForState((state) => state.preview.status.project === "FETCHED");
+  const projectAuthor = redux.state.preview.projectInfo.author?.username;
+
+  const isOwn = username === projectAuthor;
+  const shared = (await addon.tab.redux.state.preview.projectInfo.is_published) == true ? true : false;
```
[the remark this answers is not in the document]
Jazza-231: I cant remember why i did this, i think in my testing it seemed to return both "true" and true

addons/download-button/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Project page download button",
+  "description": "Allows you to download projects directly from the project page, without having to open the editor first.",
```
WorldLanguages, line 3: Would be nice to say where the button is located.
`next to the Copy Link button`?

addons/download-button/userscript.js
```diff
@@ -0,0 +1,45 @@
...
+  async function download() {
+    const project = await vm.saveProjectSb3();
+    const title = isOwn
+      ? await addon.tab.waitForElement(".project-title input")
+      : await addon.tab.waitForElement(".project-title");
+    downloadBlob(`${isOwn ? title.value : title.innerText}.sb3`, project);
+  }
+
+  const downloadButton = document.createElement("button");
+  downloadButton.innerText = "Download";
```
WorldLanguages, line 26: We should attempt to localize this string!
https://scratchaddons.com/docs/localization/localizing-addons/

WorldLanguages (changes requested):
https://github.com/ScratchAddons/ScratchAddons/pull/6931#pullrequestreview-1853983585

addons/download-button/userscript.js
Jazza-231: on it.

Jazza-231:
@WorldLanguages done

Jazza-231:
I added a spinner to show that something is happening, which is helpful when you try to download bigger projects.

Samq64 (changes requested):
When clicking the button while the project is still loading this happens:
Screenshot_20240302_083836
Ideally the button would only be added after the project finishes loading but before it starts.

Jazza-231:
> When clicking the button while the project is still loading this happens: Screenshot_20240302_083836 Ideally the button would only be added after the project finishes loading but before it starts.

Aw shit... Uhm is there a Redux event for when it's finished loading or a pre-existing way to detect that? If not, what should I do, look for a green flag or something?

Samq64:
> > When clicking the button while the project is still loading this happens: Screenshot_20240302_083836 Ideally the button would only be added after the project finishes loading but before it starts.
> 
> Aw shit... Uhm is there a Redux event for when it's finished loading or a pre-existing way to detect that? If not, what should I do, look for a green flag or something?

@CST1229 gave it to me on the Dev Discord. The whole line would be:
```js
await redux.waitForState((state) => state.scratchGui.projectState.loadingState.startsWith("SHOWING"));
```

Jazza-231:
hey @Samq64 how's this? it blocks the clicks until the project is loaded...
Download project 2

(im using my phone data for this, thats how much it matters to me)

addons/download-button/userscript.js
```diff
@@ -0,0 +1,57 @@
+import downloadBlob from "../../libraries/common/cs/download-blob.js";
+
+export default async function ({ addon, console, msg }) {
+  const vm = addon.tab.traps.vm;
+  const { redux } = addon.tab;
+
+  await redux.waitForState((state) => typeof state.session.session?.user === "object");
+  const username = await addon.auth.fetchUsername();
```
Samq64, line 8: Why use both of these methods? The first just cases the addon to never do anything when logged out.

addons/download-button/userscript.js
```diff
@@ -0,0 +1,57 @@
...
+
+  async function download() {
+    const downloadButton = document.querySelector(".sa-download-button");
+    if (downloadButton.classList.contains("waiting")) return;
+    downloadButton.classList.add("loading");
+    try {
+      const project = await vm.saveProjectSb3();
+      const title = isOwn
+        ? await addon.tab.waitForElement(".project-title input")
+        : await addon.tab.waitForElement(".project-title");
```
Samq64, line 24: Correct me if I'm wrong but there elements are guaranteed to exist by this point, right?
```suggestion
        ? document.querySelector(".project-title input")
        : document.querySelector(".project-title");
```

addons/download-button/userscript.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  addbutton();
+  addon.tab.addEventListener("urlChange", () => {
+    addbutton();
+  });
+
+  addon.tab.displayNoneWhileDisabled(downloadButton);
+
+  vm.runtime.on("PROJECT_LOADED", () => {
```
Samq64, line 52: I was thinking of using the other method near the beginning of the file and while it will cause the button to only be added later on it also prevents some edge cases since this event is only fired once while the other way checks for a condition that stays true.

addons/download-button/userscript.js
Jazza-231: Uhm yeah they should be

addons/download-button/userscript.js
Jazza-231: If I put the await at the end then I can still put the button in and make it do the same as is it does now I think

addons/download-button/userscript.js
Jazza-231: genuinely no idea, obviously we want them to still be able to download other's projects when they are logged out... ill deal with it

Samq64 (approved):

DNin01:
> hey @Samq64 how's this? it blocks the clicks until the project is loaded...

Personally, I would dim/gray out the button while the project is loading instead of saying "Loading…". Scratch's Remix button works this way.

addons/download-button/loading.svg
DNin01, line 1: For consistency, the spinner should probably be the same as Scratch's (the one that appears when saving projects).

I believe [this](https://github.com/ScratchAddons/ScratchAddons/blob/master/libraries/common/cs/spinner.css) is the one. To import the CSS, just include `@import url("../../libraries/common/cs/spinner.css");` in your userstyle.

Jazza-231:
> > hey @Samq64 how's this? it blocks the clicks until the project is loaded...
> 
> Personally, I would dim/gray out the button while the project is loading instead of saying "Loading…". Scratch's Remix button works this way.

Alright, I'll do that

addons/download-button/loading.svg
Jazza-231: good point

Jazza-231: uhm...im using :before for the spinner, but the css targets .sa-spinner, but i can not add a class to the :before psuedo class, any recommendations? quite annoying tbh

Jazza-231: I hope its ok i dont use the library directly, because i need to make quite a few changes to it i just copied the css i needed and modified it, if thats not ok then

Samq64:
I was going to say just @import the styles at the top of the class and add that class with JS but then I realized it's a `:before` so it's not that simple.

addons/download-button/loading.svg
DNin01, line 1: If this spinner isn't being used anymore, it can be deleted.

DNin01 (approved):

addons/download-button/loading.svg
Jazza-231: oops

addons/download-button/style.css
```diff
@@ -0,0 +1,49 @@
...
+@keyframes spin {
+  0% {
+    transform: rotate(0);
+  }
+  100% {
+    transform: rotate(359deg);
+  }
+}
+
+#sa-download-button.loading:before {
```
Samq64, line 40:
```suggestion
#sa-download-button.loading:before,
#sa-download-button.waiting:before {
```

DNin01: Displaying a spinner on the button during a download is fine, but if we show the spinner while the project is loading, it might make it seem like the addon is doing something.

addons/download-button/userscript.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  const downloadButton = document.createElement("button");
+  downloadButton.innerText = msg("download");
+  downloadButton.onclick = download;
+  downloadButton.classList = "button action-button sa-download-button waiting";
+  downloadButton.id = "sa-download-button";
+
+  function addbutton() {
+    addon.tab.waitForElement(".flex-row .subactions", { markAsSeen: true });
```
WorldLanguages, line 34: This line does nothing, you need to `await` it.

addons/download-button/addon.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Project page download button",
+  "description": "Allows you to download projects directly from the project page, without having to open the editor first. The Download button is added next to the Copy Link button.",
+  "tags": ["community", "projectPlayer", "recommended"],
```
WorldLanguages, line 4: Recommended? Probably not...

addons/download-button/userscript.js
WorldLanguages: Solved by my commit https://github.com/ScratchAddons/ScratchAddons/pull/6931/commits/98a00b400f52af9952c518c1cc32af4b8b828c9e

WorldLanguages (approved):
LGTM!

But we should probably remove the recommended tag before merging
@Jazza-231 Any other comments?

addons/download-button/addon.json
Samq64: Probably should be `featured`.
```suggestion
  "tags": ["community", "projectPlayer", "featured"],
```

Jazza-231:
I'll do those when I get home

Jazza-231:
> @Jazza-231 Any other comments?

I added a line you removed back in to add the button while the project is not yet loaded

BroJac5246:
> what does lgtm mean

[Looks good to me](https://letmegooglethat.com/?q=lgtm+meaning)

Jazza-231:
> > what does lgtm mean
> 
> [Looks good to me](https://letmegooglethat.com/?q=lgtm+meaning)

Oh lol fsr i was thinking "lets get this merged"

WorldLanguages:
> > @Jazza-231 Any other comments?
> 
> I added a line you removed back in to add the button while the project is not yet loaded

Well, it's not ideal as the button won't be added if it's the editor instead of the project page, but it will work anyway, and will not log any errors. (appendToSharedSpace returns false if the element wasn't added, it doesn't log or throw any errors)

Issue #6592, opened 2023-08-25
Local variables by default


Samq64:
### Why this would be helpful

When working with big projects new variables are usually local, but Scratch defaults to global ones.

### How this addon works

An addon that changes the default selected variable scope in the create a variable dialog to local.

### Possible alternatives

Maybe this could be part of `swap-local-global`?

### Additional context

Apparently ScratchTools has this feature.

WorldLanguages:
> Maybe this could be part of `swap-local-global`?

Yeah, it's either a setting there, or a new addon.

medlans:
Yea st has it, I have that extension and I checked. I think that it could be added here or the extension could be made compatible with ST since I need that to add some stuff and scratch-www -> scratchr2.

> > Maybe this could be part of `swap-local-global`?
> 
> Yeah, it's either a setting there, or a new addon.

I think maybe a local var Addon could be made to combine the 2, and turn it on.

WorldLanguages:
Related: #6594

penguinmoose:
I think it's better if instead, this addon makes it retain the previous local/global selection (also applies to the cloud variable checkbox).

WorldLanguages:
> I think it's better if instead, this addon makes it retain the previous local/global selection (also applies to the cloud variable checkbox).

Maybe we could instead allow you to create multiple variables within the same New Variable modal? (through a custom UI or maybe comma separated)

WorldLanguages:
Also, I think `data-category-tweaks-v2` could provide separate "Create Global Variable" and "Create Local Variable" if the setting to separate them is enabled.

Samq64:
Closing in favor of #6940.

PR #6932, opened 2023-12-06
Press ctrl+enter to make a variable for this sprite only

jazzathedev:
Resolves my brain and suggestion (can u sense a pattern)

### Changes

On ctrl key press it selects the local thingy, and ctrl+enter clicks the ok button cuz it doesn't actually just work with ctrl too
### Reason for changes

convenience
### Tests

works on brave

### DIFF
--- a/addons/ctrl-enter-ftso/addon.json
+++ b/addons/ctrl-enter-ftso/addon.json
@@ -12,7 +12,7 @@
   "credits": [
     {
       "name": "Jazza",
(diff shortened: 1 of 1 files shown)


Samq64:
Related: #6592

Jazza-231:
I need to remake this so ignore it for now

Jazza-231:
> does it really need an addon of its own?

It will once I actually make it good, see https://github.com/ScratchAddons/ScratchAddons/issues/6940#issuecomment-1849388224

PR #6952, opened 2023-12-12
Don't escape BBCode tags with `better-quoter` and `show-bbcode`

### LINKED ISSUE #6948: `better-quoter`: escaping BBCode always happens while viewing BBCode with `show-bbcode` in pre-release
### Description
Unsure if I should use /feedback or just make an issue on pre-release issues (especially since the PR that added this is merged, so I think it'd be odd if I reported the bug there)
### What should happen
Quotes post like normal (BBCode included)
### What actually happens
Adds un-needed double that prevents the BBCode from properly registering ([]).

### Steps to reproduce

1. Enable `better-quoter` & `show-bbcode`
2. Go to any post in the scratchblocks testing topic
3. View bbcode
4. Highlight from start to finish and try quoting.
5. See issue (hopefully)

used in example: https://scratch.mit.edu/discuss/post/7631503/
kép

### Environment

Browser: Firefox 12.0.1 (32 bit)
Scratch Addons version:  v1.36.0-pre (using commit 66d5782 's version which was just a transifex update (EDIT: nevermind thank you to DNin01 I forgot that's the commit hash which I'm supposed to give lol)
Operating system: Windows 10 (32 bit)

mybearworld:
Resolves #6948

### Changes

Add a data attribute to the element added by `show-bbcode`, so that `better-quoter` can see it and not escape brackets.

### Reason for changes

To make the interop make more sense and be useful.

### Tests

Tested in Edge.

### DIFF
--- a/addons/better-quoter/module.js
+++ b/addons/better-quoter/module.js
@@ -209,7 +209,7 @@ function setup() {
   if (isSetup) return;
   isSetup = true;
   const originalCopyPaste = window.copy_paste;
-  window.copy_paste = function (id) {
+  window.copy_paste = async function (id) {
     const post = $("#" + id);
     const username = post.find(".username").text();
     const idText =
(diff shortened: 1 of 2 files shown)


WorldLanguages:
As I don't frequently use the forums, I'm not sure if I want to waste energies trying to understand what is going on here 😂

WorldLanguages:
@Zydrolic Could you check if this fixes the bug?

WorldLanguages:
This PR caused a regression: #6964

PR #6979, opened 2023-12-19
Change "argument" to "parameter"

DNin01:
### Changes

In the `block-switching` addon, changes occurrences of the word "argument" to "parameter". To my knowledge, the computer science terms "argument" and "parameter" are used interchangeably, but in Scratch, they are officially called parameters, according to the Scratch Team's latest [tutorial video](https://youtu.be/596LEqy6p7M). IMO, I also think "custom block parameter" might be understood by more users.

### Reason for changes

Consistency with the official terminology.

### Tests

Not tested.

### DIFF
--- a/addons/block-switching/addon.json
+++ b/addons/block-switching/addon.json
@@ -85,24 +85,24 @@
       }
     },
     {
-      "name": "Custom block arguments",
+      "name": "Custom block parameters",
       "id": "customargs",
       "type": "boolean",
       "default": true
     },
     {
-      "name": "Shown custom block arguments options",
+      "name": "Shown custom block parameters options",
       "id": "customargsmode",
       "type": "select",
       "default": "defOnly",
       "potentialValues": [
         {
           "id": "defOnly",
-          "name": "Arguments in own custom block"
+          "name": "Parameters in own custom block"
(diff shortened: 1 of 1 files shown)


BroJac5246:
> To my knowledge, the computer science terms \"argument\" and \"parameter\" are used interchangeably

I believe an argument is the value passed into a parameter, if I'm not mistaken. They are nevertheless used interchangeably, though.

DNin01:
Ohhhh!

https://developer.mozilla.org/en-US/docs/Glossary/Parameter#parameters_versus_arguments
> Note the difference between parameters and arguments:
> - Function [parameters](https://developer.mozilla.org/en-US/docs/Glossary/Parameter) are the names listed in the function's definition.
> - Function [arguments](https://developer.mozilla.org/en-US/docs/Glossary/Argument) are the real values passed to the function.
> - Parameters are initialized to the values of the arguments supplied.

DNin01:
So the reporter blocks that represent the value (argument) passed into a custom block are, in fact, classified as parameters.

Samq64 (approved):

mxmou (approved):

PR #6993, opened 2023-12-23
Allow opening /source in a new tab with `show-bbcode`

### LINKED ISSUE #6937: `show-bbcode`: option to open /source in new tab (or directly open/copy url) instead of replacing post
### Why this would be helpful

It's much easier to find BBCode or sections by just viewing from the /source page rather than using the view source button (due to the post being focused) and having the post itself get replaced.
Not to mention a lot easier to link as well to show the source to others, and to mention make the feature a bit more well known since it's really useful.
One example would be TOLoRS (The Official List of Rejected Suggestions, Top Sticky by Za-Chary), where people complain about slow load times ([the topic](https://scratch.mit.edu/discuss/topic/343602/) - [the source](https://scratch.mit.edu/discuss/post/3466932/source/) - compare the two and the source loads much quicker due to the post being focused, it's also easier to search despite some oddities (presumably issues supporting a few characters)) - The addon does not mention the existance of /source either, so all you really get is being able to paste the full entire source somewhere.

### My suggestion

Add an option to either make the addon open the source of the post (via discuss/post/(id)/source - already achieved in the addon code, I'm just too bad to make a PR and try) rather than loading and replacing the post.
Something like this (Note: made this with inspect):
mock-up
Or this via an option to just directly link to the source (not open in a new tab), though I don't really think this is neccassary
direct-url

### Possible alternatives

1. get used to getting the post id and adding source
2. don't implement
3. just have these two options rather than replacing the post contents (locally) with the source.
4. (added in edit) have an option to copy the source url instead of any of these
5. (added in another edit) Mention the existance of it (/discuss/post/*/source) in the addon's description.

### Additional context

Context: This is mainly due to me realizing /source can load much faster if you just view it there, especially for TOLoRS which has had complaints of slow load times.

Nothing else.

mybearworld:
Resolves #6937

### Changes

Add a setting to `show-bbcode` allowing opening the source in a new tab.

The "DISPLAY TYPE" setting with the options "Show source in post" and "Open in new tab".

### Reason for changes

It's easier to go through.

### Tests

Tested in Edge.

### DIFF
--- a/addons/show-bbcode/userscript.js
+++ b/addons/show-bbcode/userscript.js
@@ -1,46 +1,3 @@
-function viewSource(post, msg, addon) {
-  return function (event) {
-    event.preventDefault();
-    const body = post.querySelector(".postmsg");
-    if (event.target.getAttribute("data-state") === "post") {
-      const sourceLink = "https://scratch.mit.edu/discuss/post/" + post.id.substring(1) + "/source/";
-      if (addon.settings.get("display-type") === "new-tab") {
-        open(sourceLink);
-        return;
-      }
-      event.target.innerText = msg("source-button-active");
-      event.target.removeAttribute("title");
-      if (event.target.originalHTML === undefined) {
-        event.target.originalHTML = body.firstElementChild;
-      }
-      body.removeChild(body.firstElementChild);
-      const source = document.createElement("div");
-      body.insertBefore(source, body.firstElementChild);
-      source.className = "post_body_html";
-      source.dataset.showBbcode = "";
-      if (event.target.sourceText !== undefined) {
-        event.target.setAttribute("data-state", "source");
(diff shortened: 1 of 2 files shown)


addons/show-bbcode/userscript.js
```diff
@@ -1,8 +1,13 @@
-function viewSource(post, msg) {
+function viewSource(post, msg, addon) {
```
Samq64, line 1: Could this function be moved inside the main one instead? That would allow it to use `msg` and `addon` without needing to pass it in.

mybearworld: Done

Samq64 (approved):
When dynamically changing to viewing in a new tab with a post source already open the button needs to clicked twice to actually open it in a new tab but that's very minor.

Issue #340, opened 2020-09-27
Upload to Mac App Store for Safari support


scratchusernamemrtbts:
**Is your feature request related to a problem? Please describe.**

some people that only use safari can't use this addon
**Describe the solution you'd like**

upload this to mac appstore to be a safari extension
**Describe alternatives you've considered**

**Additional context**

JeffaloBob:
does safari support the same kind of extensions as chrome and firefox? @WorldLanguages

scratchusernamemrtbts:
Yes

scratchusernamemrtbts:
But you’ll need to make a home page for it as an app

Thingyman-DEV:
I don't think they can. To make it actually on the mac store, you need an apple developer account which is _99 dollars a year_

GrahamSH-LLK:
Yeah. I don't think we have that kind of money. If you really want it, just download chrome.

WorldLanguages:
Two things missing:
- $99 a year
- Support for some APIs on Safari 14, like `chrome.cookies.addListener("change" ...` (source: [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged))

scratchusernamemrtbts:
I already try to convert it to safari
but there is still a bug
the popup page don't show the info it shows only the menu bar
but it works in the editor or when I click the popout
<img width="2560" alt="ภาพถ่ายหน้าจอ 2563-11-01 เวลา 07 53 43" src="https://user-images.githubusercontent.com/66320961/97792926-663c8580-1c17-11eb-9b5d-ebbdbf2ef335.png">
<img width="2560" alt="ภาพถ่ายหน้าจอ 2563-11-01 เวลา 07 54 18" src="https://user-images.githubusercontent.com/66320961/97792930-75233800-1c17-11eb-8d01-d3a27c3a3c11.png">


<img width="2560" alt="ภาพถ่ายหน้าจอ 2563-11-01 เวลา 07 44 57" src="https://user-images.githubusercontent.com/66320961/97792824-2628d300-1c16-11eb-87d6-7562a9184022.png">

<img width="2560" alt="ภาพถ่ายหน้าจอ 2563-11-01 เวลา 07 42 06" src="https://user-images.githubusercontent.com/66320961/97792797-c29ea580-1c15-11eb-85a8-3e1072281132.png">

scratchusernamemrtbts:
[...]

PR #7006, opened 2023-12-27
Safari

GarboMuffin:
This is a proof of concept Safari port.

The native parts of the app are unmodified stuff from Apple's porting script. I don't predict any need to change those in the near future. Some notes for development are over here: https://github.com/GarboMuffin/ScratchAddons/tree/safari/safari#readme

I've made one pass through almost every extension to look for bugs; here's what I've got for now:

 - [ ] addon.tab.traps.vm is sometimes null
 - [ ] cookies.onChanged not supported (the function exists but does nothing)
 - [ ] cant apply presets in popup (confirm() just returns false immediately)
 - [ ] mediarecorder: broken; Safari's MediaRecorder only does video/mp4
 - [ ] resizable-comment-input: broken; seems like `resize: vertical` doesn't work if `padding:` is specified?
 - [ ] scratch notifier: broken; chrome.notifications is not supported. window.Notification seems to require re-requesting permission each time Safari starts. [Not aware of any workaround.](https://stackoverflow.com/questions/75660781/can-macos-safari-web-extensions-issue-local-notifications-without-asking-for-per) I speculate there may be some workaround with Safari's proprietary push notification system.
 - [ ] zebra striping: broken; `filter:` might not work in SVG
 - [X] gamepad: unreliable; adjusted existing unreliable browser detection to include Safari
 - [ ] bitmap-copy: should support Cmd+C on macOS
 - [ ] blocks2image: test images were being truncated; unknown if that's Safari's fault or the addon's fault
 - [ ] remove-search-bar-autocomplete: the problem this fixes does not seem to exist in Safari
 - [ ] need proper [macOS-style icons](https://developer.apple.com/design/human-interface-guidelines/app-icons)

See commit history for explanation behind the changes. Here's some of the highlights:

 - `chrome.i18n.getMessage()` gets stuck in an infinite loop if $1 is replaced with $1
 - some `promisify(chrome.*)` calls need an explicit bind() to work properly
 - `chrome.runtime.reload()` does not automatically close the current tab. Not sure if calling close() immediately after reload() has any reliability issues on slow computers.
 - `new URL(import.meta.url).origin` is not usable because it returns `safari-web-extension://lowercase-uuid` which doesn't work; it needs to be `safari-web-extension://UPPERCASE-UUID`
 - `chrome.browserAction` badge APIs don't return a Promise and don't use callbacks
 - `chrome.management` is not supported
 - `sender.url` does not exist when the popups fetch their data


sfederici:
My usual newbie question, sorry: how do I checkout the code in order to test it?

Samq64:
> My usual newbie question, sorry: how do I checkout the code in order to test it?

https://github.com/GarboMuffin/ScratchAddons/blob/safari/safari/README.md#development

GarboMuffin:
@sfederici re: https://github.com/ScratchAddons/ScratchAddons/issues/340#issuecomment-1870229116 -- trying to keep the discussion over here

What do you see in this screen?

image

sfederici:
I managed to install the extension. I can see the addon button in Safari and the settings page. I see only a few addons (the ones that are supposed to work?). I tried the "mouse position" addon, but the mouse position doesn't show up.

sfederici:
> And what version of Safari do you have installed? (Top right by the apple icon > Safari > About Safari)

17.1

GarboMuffin:
> I managed to install the extension. I can see the addon button in Safari and the settings page. I see only a few addons (the ones that are supposed to work?). I tried the "mouse position" addon, but the mouse position doesn't show up.

There's something really weird here. I'm not sure what's up

sfederici:
> When you say "I see only a few addons" do you mean that the full settings page only has a few of them listed at all?

For example, for the editor addons I only see the following ones:
immagine

Samq64:
immagine

What version of the extension are you running, it looks like it might be a couple years old.

sfederici:
> What version of the extension are you running? It looks like it's a couple years old.

If I'm not wrong I downloaded the code from https://github.com/GarboMuffin/ScratchAddons/tree/safari

CST1229:
> If I'm not wrong I downloaded the code from [GarboMuffin/ScratchAddons@`safari`](https://github.com/GarboMuffin/ScratchAddons/tree/safari?rgh-link-date=2024-01-05T18%3A02%3A00Z)

You might have accidentally downloaded the code from GarboMuffin's `master` branch instead (which is a couple years old).

sfederici:
> You might have accidentally downloaded the code from GarboMuffin's `master` branch instead 

This branch doesn't include the safari folder that I used to compile the extension, right?

GarboMuffin:
try this instead: https://testflight.apple.com/join/OrodhadK

sfederici:
> try this

Everything worked really well. Almost all the addons I'm interested in worked very well except for one that allows to drag images from a folder to the GUI. When I drag the images, the Scratch editor tab moves to a different page showing just that image.

I liked a lot the new export format for screen recordings, mp4. It is very handy. Do you think this could be added to the standard ScratchAddons for Safari and Chrome too?

CST1229:
> I liked a lot the new export format for screen recordings, mp4. It is very handy. Do you think this could be added to the standard ScratchAddons for Safari and Chrome too?

It's only there because Safari only supports recording in mp4 (while all the other browsers only support webm).

sfederici:
I dont' know if it is supposed to work, but I tested the full-signature addon for the website (that I'd like a lot) and at least for the "what's happening" area when I click the show more button nothing happens.

GarboMuffin:
This has been stalled because until Apple fixes bugs like https://bugs.webkit.org/show_bug.cgi?id=253696 it just isn't possible to port large parts of Scratch Addons

WorldLanguages:
> This has been stalled because until Apple fixes bugs like https://bugs.webkit.org/show_bug.cgi?id=253696 it just isn't possible to port large parts of Scratch Addons

Well, "Scratch Messaging" isn't really the most important part of Scratch Addons nowadays. But it is true that Safari has lots of these tiny bugs everywhere in the extension API space.

GarboMuffin:
This will get looked at again in the new year.

Supposedly declarativeNetRequest is still broken for an extension's own requests, but supposedly cookie API now lets extensions access HTTP-only cookies, so it would be theoretically possible (and very scary!) to send the cookies over native messaging to some native code that will do the requests.

It's also possible that their MV3 is less buggy than their MV2 and maybe they've been fixing bugs without updating the public bug tracker

DNin01:
> It's also possible that their MV3 is less buggy than their MV2

Hasn't Safari already dropped support for MV2 extensions?

GarboMuffin:
I don't think so, but my understanding is modern Scratch Addons is mv3 so its not needed for us

sfederici:
> This will get looked at again in the new year.

Thanks for letting us know. I really look forward to it

PR #7017, opened 2023-12-30
Add some code comments to `confirm-actions`

DNin01:
### Changes

Added comments to the source code of the `confirm-actions` script, mainly to label each if statement.

This does not affect the functionality of the addon, only the source code.

Also renames two variables.

### Reason for changes

It makes the source code easier to understand.

### Tests

Tested.

PR #7018, opened 2023-12-30
Confirm clicking Reply while writing a comment

### LINKED ISSUE #7000: Confirmations Addon: Confirm when you press the "reply" button
### Why this would be helpful

When I am trying to write a comment, I sometimes accidentally press the "reply" button which removes the comment you are writing. It would be helpful to warn the user when there is content in the comment and you press the "reply" button.

### My suggestion

Warn the user about clicking the "reply" button when the comment has text in it and add the feature to the "Confirmations" addon. If the comment you are writing is empty, don't warn the user about clicking the "reply" button.

### Possible alternatives

_No response_

### Additional context

comment

DNin01:
Resolves #7000

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/ae2df17a-fc3f-4d0a-97a6-645e59fa00c2

### Changes

Each comment on Scratch has a reply button that shows the comment box when clicked. If you click on the reply button again, the comment box is removed and your comment gets discarded. This update shows a confirmation before this action, if you've enabled "Confirm canceling comments", of course.

### Reason for changes

To prevent accidentally discarding comments.

### Tests

Tested on profile pages, projects, and studios, only a little bit.

### DIFF
--- a/addons/confirm-actions/userscript.js
+++ b/addons/confirm-actions/userscript.js
@@ -48,14 +48,21 @@ export default async function ({ addon, console, msg }) {
       } else if (addon.settings.get("closingtopic") && e.target.closest("dd form button")) {
         title = msg("closetopic-title");
         cancelMessage = msg("closetopic");
-      } else if (
-        addon.settings.get("cancelcomment") &&
-        e.target.closest("div[data-control='cancel'] > a, .compose-cancel")
-      ) {
-        // Do not ask to confirm canceling empty comments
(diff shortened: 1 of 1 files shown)


mxmou (approved):

WorldLanguages (approved):
Nice! LGTM

WorldLanguages:
I will merge #7017 after this one gets merged

Samq64 (approved):

PR #7019, opened 2023-12-30
Added the "Invert stage" plugin

gamemake-eng:
### Changes

I added a plugin that allows people to invert the stage with a click of a button!

[Screen recording 2023-12-30 12.33.42 PM.webm](https://github.com/ScratchAddons/ScratchAddons/assets/63882211/90ca2f67-ab39-4525-bb42-9817465d088e)

### Reason for changes

I made this plugin mainly to allow users to test their projects for accessibility without having to change system settings.

### Tests

I only have a chromebook so I was only able to test it on chrome. Everything seems to work as normal even in the editor.


gamemake-eng:
Should I close this and make another PR on a separate branch?

mybearworld:
Does inverting colors actually help with color blind testing? The colors are just different like that, there's not any colors merging.

addons-l10n/en/invert-stage.json
mybearworld, line 1: This file is not being used. Addon names and similar get translated automatically, you don't need to create a file for it

addons/addons.json
```diff
@@ -6,6 +6,7 @@
   "scratch-notifier",
   "discuss-button",
   "editor-stepping",
+  "invert-stage",
```
mybearworld, line 9: This should be added before the comment that says `"// NEW ADDONS ABOVE THIS ↑↑"`, not in the middle of the addon list..

addons/invert-stage/addon.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "Invert Stage",
+  "description": "Inverts the stage. Useful for testing contrast and people with colorblindness.",
+  "tags": ["community"],
```
mybearworld, line 4: This is an editor addon, so it should have matching tags.

```suggestion
  "tags": ["editor", "projectPlayer"],
```

addons/invert-stage/addon.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "Invert Stage",
+  "description": "Inverts the stage. Useful for testing contrast and people with colorblindness.",
+  "tags": ["community"],
+  "versionAdded": "1.36.0",
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects", "projectEmbeds"]
```
mybearworld, line 9: If this is an addon for testing projects for colorblindness, does it need to be on project embeds as well? Those are mostly used by people viewing the project, not making them.

addons/invert-stage/addon.json
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["projects", "projectEmbeds"]
+    }
+  ],
+  "credits": [
+    {
+      "name": "michealtheratz (@batmangreen123 on scratch)",
```
mybearworld, line 20: Should this contain two names?

addons/invert-stage/style.css
mybearworld, line 1: `.btn-container` and `.invert-btn` are quite generic names that may conflict with something in the future. It's general practice to prefix them (so something like `sa-invert-stage-btn-container` and `sa-invert-stage-button`).

https://scratchaddons.com/docs/develop/userstyles/best-practices/#begin-addon-defined-class-names-with-sa-

addons/invert-stage/style.css
```diff
@@ -0,0 +1,26 @@
...
+    border: 1px solid hsla(0, 0%, 0%, 0.15);
+    border-radius: calc(0.5rem / 2);
+    width: calc(2rem + 2px);
+    height: calc(2rem + 2px);
+    background: hsla(0, 100%, 100%, 1);
+    padding: 0.375rem;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
```
mybearworld, line 12: Are all these prefixes necessary for the browsers SA supports?

addons/invert-stage/userscript.js
```diff
@@ -0,0 +1,37 @@
+export default async function ({ addon, console, msg }) {
+  // INVERT STAGE ADDON BY @BATMANGREEN123 (MICHEALTHERATZ)
+  const vm = addon.tab.traps.vm;
+  const canvas = vm.renderer.canvas;
```
mybearworld, line 4: This part of the code errors for me, with `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'canvas')`.

addons/invert-stage/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+
+  // CREATING THE BUTTON (Using material icons bc i'm lazy)
+  const img = document.createElement("img");
+  img.className = "invert-btn btn-container";
+  img.draggable = false;
+  img.src = addon.self.dir + "/contrast.svg";
+
+  // EVENT LISTENERS
+  img.addEventListener("click", () => invert = !invert);
+  addon.tab.displayNoneWhileDisabled(img);
```
mybearworld, line 18: Dynamic disable/enable is not actually enabled in the `addon.json` file, so this doesn't do anything.

addons/invert-stage/userscript.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+  window.requestAnimationFrame(loop);
+
+  while (true) {
+    // ADDING THE BUTTON INTO THE SCRATCH-GUI
+    await addon.tab.waitForElement("[class^='stage-header_stage-menu-wrapper']", {
+      markAsSeen: true,
+      reduxEvents: ["scratch-gui/mode/SET_PLAYER", "fontsLoaded/SET_FONTS_LOADED", "scratch-gui/locales/SELECT_LOCALE"],
+    });
+    addon.tab.appendToSharedSpace({ space: "fullscreenStageHeader", element: img, order: 2 });
```
mybearworld, line 35: I'm not sure because I can't test this, but does this also add the button in non-full-screen mode?

mybearworld (changes requested):
I couldn't actually test this because it errored for me, but this is what I found when reading the code

BroJac5246:
> Does inverting colors actually help with color blind testing? The colors are just different like that, there's not any colors merging.

Inverting colors may help color blind people themselves (since weak colors will get mapped to other colors that the eye _can_ distinguish), but it definitely won't help with testing because colors don't become less distinctive, just different — like if I replace red and yellow with orange and green, a regular person could still tell them apart because they're still different colors. If we invert the colors, it won't affect testing because they'll still be distinct to the creator, but it may help colorblind people because, say red and green get mapped to yellow and cyan, they may be able to see the difference now. That said, no one would actually watch an entire project like this because you can tell it's inverted — instead one might use their operating system's built-in color correction feature.

If you want to test for colorblind-friendliness, you'll need a filter that weakens (as opposed to changes) certain colors — this is hard to explain but Android has a developer option for this (Simulate color space) if you're able to check that out to see what I mean. I would assume there are Chrome extensions that do the same.

I'm not saying we shouldn't do this, I'm just not sure it makes sense for its current purpose.

DNin01:
I would agree that inverting colors is not sufficient for testing color blindness support. We should instead have filters for specific types of color blindness.

Also worth noting is Chrome(?) DevTools has a feature to simulate different types of vision deficiency. I found it through the device emulation feature in Edge, and I would expect it works in Chrome. I used it on one of my recent projects.

addons/invert-stage/addon.json
gamemake-eng: It could be userful for people who are colorblind themselves so I will keep this as is

addons/invert-stage/userscript.js
gamemake-eng: what browser are you using?

mybearworld: Edge 120 - that runs on Chromium, so if it works on Chrome, it should work here

addons/invert-stage/userscript.js
gamemake-eng: I tested it on my computer and it seems to apply to both.

mybearworld: Alright, then. I wasn't sure.

addons/invert-stage/userscript.js
gamemake-eng: that's weird, maybe try it on regular chrome.

mybearworld: I can't, because of parental controls - addons should work on Edge anyways.

addons-l10n/en/invert-stage.json
mybearworld: This isn't resolved

addons/invert-stage/userscript.js
gamemake-eng: You have parental controls? I might have to look into this. Maybe it's just a you problem.

gamemake-eng: because the VM object should have the canvas

WorldLanguages:
Nice first addon!

I'm not sure if this will actually get into the extension. Hopefully it was fun to implement it, no matter if it gets into a release or not :-)

addons/invert-stage/userscript.js
mybearworld: The error message seems to suggest that the VM does not have a renderer. Maybe it only exists once the project has loaded? Does this addon work for you on, say, Appel?

addons-l10n/en/invert-stage.json
mybearworld, line 1: This is still unused

gamemake-eng:
> Nice first addon!
> 
> I'm not sure if this will actually get into the extension. Hopefully it was fun to implement it, no matter if it gets into a release or not :-)

Thanks! I partially made this to showcase how people can manipulate the stage itself in an addon. I also found out you can get the webGL context from the renderer object so I can do stuff with shaders too. There is a low chance that this will get accepted, but it was worth a try.

TheColaber:
Well done on this addon, it looks like it was fun to make! Look into the Issues tab and see if any open addon suggestions also sound interesting to implement.

PR #7020, opened 2023-12-30
New addon: reorder-custom-inputs

### LINKED ISSUE #6073: Improved Function Editing: reorder custom block parameters
### Why this would be helpful

Oftentimes, I would like to add a new input to a function (aka "my blocks"). For example, I might have a function like this:
```
create sprite at x: (x) y: (y) costume: (costume)
```
I might want to add another value, z, for the z-position of the sprite. However, unless I want to delete some of the inputs and rewrite my code, my only option is to add it to the end:
```
create sprite at x: (x) y: (y) costume: (costume) z: (z)
```
This is frustrating, because logically, it should be after x and y. Especially in complicated functions, rewriting it for an aesthetic change simply isn't worthwhile, but it still makes for a bad user experience (especially if you are programming an engine that other Scratchers are meant to use).

A more practical problem is if you created a function with unlabeled inputs, like this:
```
create sprite at (x) (y) (costume)
```
Then, you might want to add labels to your function, to make it easier to use. However, inserting labels between inputs is not possible, and you'd have to delete and re-create the inputs, and rewrite the function.

### How this addon works

When you edit a function, you can click on inputs to rename them. In addition, a "delete" icon appears above the inputs when selected. This addon would add "move left" and "move right" icons above the inputs, allowing you to rearrange them within the function. In addition, if you clicked the "add input" or "add label" buttons while selecting an input or label, it would add the new input or label right after your selection, rather than at the end of the function.

### Possible alternatives

I'm not sure if there are any good alternatives to this--except maybe an option to substitute all instances of one function input for another? That would be able to do the same job, but it would be more complicated and less intuitive.

### Additional context

_No response_

Joeclinton1:
Resolves #6073 

### Changes

Created a new addon called: reorder-custom-inputs

reorder custom inputs

### Reason for changes

To make it possible to more easily change the order of inputs in a custom block.

### Tests

It works fine in Chrome.
- tested on the different input types
- tested trying to shift inputs when they're on the edge
- tested how shifting inputs affects existing argument usage within the block itself. 
- Update: tested dynamicEnable/Disable
- Update: tested merging label input into another label input

All works.


addons/reorder-custom-inputs/userscript.js
```diff
@@ -0,0 +1,73 @@
...
+
+    Blockly.WidgetDiv.DIV.insertAdjacentHTML(
+      "beforeend",
+      `
+            <svg width="20px" height="40px" 
+                 style="left: ${direction === "left" ? "calc(50% - 20px)" : "calc(50% + 20px)"}" 
+                 class="blocklyTextShiftArrow">
+                <path d="${path}" fill="none" stroke="#FF661A" stroke-width="2"></path>
+            </svg>`
+    );
```
CST1229, line 15: I'm not sure if doing this is allowed in addons (as I know `innerHTML` isn't, and this similarly involves adding HTML from a string).

Joeclinton1: The innerHTML thing is to prevent css attacks however here It is allowed because it doesn't take in user provided data and add it to the page. It's no different to just adding an svg file.

addons/reorder-custom-inputs/addon.json
```diff
@@ -0,0 +1,28 @@
...
+  ],
+  "userstyles": [
+    {
+      "url": "arrows.css",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.36.0",
+  "tags": ["editor"],
+  "enabledByDefault": true,
```
WorldLanguages, line 25:
```suggestion
  "enabledByDefault": false,
```

Sorry, judging by existing addons this wouldn't get enabled by default

GarboMuffin (commented):
The reordering part works great

A couple nitpicks:

 - above breaks dynamic enable in the edge case where the procedure editor is already open when the addon is enabled
 - dont see a need to be adding things to ProcedureUtils and the actual Block objects when you can use local function (resolving this fixes above)
 - claims dynamicDisable: true but does not actually support that

GarboMuffin (approved):

Samq64 (approved):
The addon and dynamic enable/disable work as expected.

Samq64:
Have have a couple suggestions:
2. When moving text inputs around they get merged. I'm not sure if this is a bug or not but it's probably ticky to prevent.
1. When creating a new input with one already highlighted could it be moved after the highlighted one?

Joeclinton1:
@Samq64 

1. Scratch doesn't allow there to be two labels next to each other which is what we'd need for it to not get merged, so we can't really do anything about it.
2. The behaviour of inserting new inputs after the current one should be done as a second PR imo. This current PR provides a work around by adding an input to the end and then reordering it to come after, so it's a step in the right direction.

WorldLanguages (approved):

Samq64:
Here's a description suggestion: 
```json
"Allows rearranging custom block inputs with the editor's \"Make a block\" screen.",
```

WorldLanguages:
@Samq64 I was thinking of using the word "parameter". See PR #7094

PR #7028, opened 2024-01-02
New Addon: Backpack-place-at-cursor

### LINKED ISSUE #1559: Import scripts from the backpack to the location in the code they're dropped to, not to the origin
Scratch currently creates them at the top-left corner, which isn't very helpful when I drop them to the place I want them to be, I have to scroll there and drag them back, one screen-width at a time (which is multiple in large programs, even when zoomed out).

If it's difficult to put the top-left part of it where the mouse-pointer drops it, an alternative intermediate solution could be determining it with the same function that determines the hat-block goes when defining a new function (which goes to the top-left of the current screen position, not of the entire program).

Joeclinton1:
Resolves #1559 

### Changes

Created an addon which when enabled makes code blocks position their top left corner at the mouse cursor when dragged in from the backpack.

place backpack code at mouse

### Reason for changes

Fixes a bug in the scratch code where code blocks just go to the origin which causes many users annoyance. Going to the mouse position is the expected behaviour.

### Tests

Tested in Chrome. 

Tested placing code with:
- different locations
- various different scroll distances in both horizontal and vertical
- different zooms of both browser and workspace
- RTL = true/false
- autohide enabled/disabled and two column category menu enabled/disabled
- the stage on left side addon enabled/disabled
- the stage in large/small/hidden mode
- different sizes of code blocks
- the target being a hovered Sprite or stage.
- the target being a hovered Sprite or stage AND this target is the current editingTarget
- different scroll levels in the browser window

Tested the dynamicEnable and dynamicDisable

There is a small bug: If the code block, when placed, goes beyond the edge of the screen the code area will auto scroll which makes the position a little off but it's so minor an issue that solving that is not worth the effort.

Update: added fix to make it work with RTL and autohide disabled.
Update 2: added fix to make it work with sharing code to a sprite/stage instead of directly to the main workspace
Update 3: added fix for shared code not being vertically aligned if the browser window has a vertical scroll bar and is not at the top.


Joeclinton1:
The name and description are as good as I could come up with but if you have any better ideas please suggest them

CST1229:
For me (on Firefox) it places scripts way off to the side (the mouse cursor in this screenshot is where I released the script):
image

Enabling the "Auto-hiding block palette" addon fixes it, so I assume you just didn't test with that addon disabled.
image

Joeclinton1:
@CST1229 You've assumed right. I wonder why that makes a difference. Hmm, how best to solve this?

BroJac5246:
Can I propose making this enabled by default? It's universally useful.

WorldLanguages:
> Can I propose making this enabled by default? It's universally useful.

I think this would be the first enabled-by-default addon that pollutes the VM. I'll make sure that we catch all reasonable potential exceptions. Remember that throwing an error inside a polluted VM method may cause the project page to crash and may lead to data loss (for example unsaved costumes).

Joeclinton1:
@WorldLanguages 
> I think this would be the first enabled-by-default addon that pollutes the VM.

I replaced the `vm.shareBlocksToTarget` function, but I only did this because I couldn't figure out how to replace the `TargetPane.shareBlocks` function which is part of the react component: TargetPane.jsx. However I've seen in the docs that you have an api called `addon.tab.traps.getInternalKey` which could maybe be used to get the TargetPane. Would you prefer this GUI function is "polluted" instead of the VM (if it's even possible to do it this way of course)?

Joeclinton1:
Now if you place the code inside a hovered sprite in the targetPane, which is also the current open sprite. It will revert to the default functionality.

addons/place-backpack-code-at-cursor/userscript.js
```diff
@@ -0,0 +1,45 @@
+export default async function ({ addon }) {
+  const Blockly = await addon.tab.traps.getBlockly();
+  const {
+    traps: { vm },
+    redux,
+  } = await addon.tab;
```
Samq64, line 6: This syntax seems confusing but that could just be me.

Samq64 (approved):
I'm not very familiar with Scratch's internals it works with all editor addons enabled and disabled. Dynamic enable/disable also work fine.

addons/place-backpack-code-at-cursor/userscript.js
Joeclinton1: It was this nice one liner before then the auto formatter ruined it. Either way, whilst it is maybe a little more confusing it has the benefits of only using one await, and less duplicate code. I'd say it doesn't matter.

addons/place-backpack-code-at-cursor/addon.json
```diff
@@ -0,0 +1,20 @@
...
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.36.0",
+  "tags": ["editor"],
```
Samq64, line 16:
```suggestion
  "tags": ["editor", "codeEditor", "recommended"],
```

medlans:
> Can I propose making this enabled by default? It's universally useful.

Scratch 2.0 actually did what this PR did, I have no idea why it was changed in 3.0 (I just found out that the behavior was changed) but the current behavior makes no sense. At least for me, it just puts it in the top left corner of the scripts area in the same spot.

Joeclinton1:
> > Can I propose making this enabled by default? It's universally useful.
> 
> I think this would be the first enabled-by-default addon that pollutes the VM. I'll make sure that we catch all reasonable potential exceptions. Remember that throwing an error inside a polluted VM method may cause the project page to crash and may lead to data loss (for example unsaved costumes).

There are very few errors that you could create in a VM function that might crash the project, and since all this code does is access some properties and set the x,y property of the blocks objects it's impossible for any error created to crash the VM. The only bugs that could cause a crash would be infinite loops, and erroneously setting important data the project relies on. As such I believe it would be an overeaction to begin adding any try/catches to this.

GarboMuffin (commented):
Hi, some small things you may want to address:

 - Does not work when someone closes the procedure editor then immediately attempts to drag code in from the backpack (use addon.tab.traps.getWorkspace)
 - Does not work when I try to drop the code behind the flyout (no flyout-related addons. The code still gets dropped in the corner)
 - Because loading things from the backpack is async, the user's mouse can move to all sorts of places before shareBlocksToTarget finally gets called, especially on the really cheap Chromebooks that schools buy their students. Mouse may not even be over the code editor anymore, so the blocks could end up in the corner.
 - I didn't mention this on your other PR because it really doesn't matter, but it's more common to do `if (addon.self.disabled) { /* dont */ }` instead of using the disabled and reenabled events to apply and unapply the trap.
 - addon.tab is not a Promise so you dont need to await it

GarboMuffin:
Also, regarding the crashing concerns above, even if shareBlocksToTarget throws an error, all that happens is the blocks don't get dropped. This isn't one of the methods where an error breaks all of scratch-gui. Backpack loading is async, so this function doesn't get called from a place where React would be able to detect the error.

Joeclinton1:
@GarboMuffin Thank you for reviewing.

1. I have fixed the "await" issue, the bug where it doesn't work after the procedure editor, and made code placeable below flyout
2. I think the current logic for reenable/disable is better than `if (addon.self.disabled) { /* dont */ }` as it removes all trace of itself, unlike the other which continues to polllute.
3. I fixed the issue with mouse moving by simply using mouseup instead of mousemove as the event listener for moving the mouse. I am 99% sure this won't lead to a race condition.

GarboMuffin (approved):

GarboMuffin:
I guess mouseup might be insufficient if someone manages to get another entire click in while the backpack is loading. That seems quite hard to do though. If someone's computer is slow enough then maybe they can drag and drop another entire piece of code from the backpack before the current one loads. I don't think that's worth worrying about.

Joeclinton1:
> I guess mouseup might be insufficient if someone manages to get another entire click in while the backpack is loading. That seems quite hard to do though. If someone's computer is slow enough then maybe they can drag and drop another entire piece of code from the backpack before the current one loads. I don't think that's worth worrying about.
> 
> unrelated: might want to check your computer's clock, those commits are dated 3 days in the future, lol

Oh, lol,  I forgot that changing the computer clock would affect the commits. My mistake. I'm currently trying out an experimental idea of mine where if all my clocks were set to the day of my coursework deadline, would I feel more inclined to actually work on the coursework. Considering I just submitted commits for a PR instead of doing my coursework I think the result will be negative.

WorldLanguages:
> unlike the other which continues to polllute.

Keep in mind it's often best practice not to unpollute. Removing all traces of the pollution may cause side-effects to other features, in cases where the polluted method itself polluted a polluted method by another feature. Yes, it sounds like a tongue twister.
See https://scratchaddons.com/docs/develop/userscripts/best-practices/#do-not-unpollute-functions

WorldLanguages:
This feature is already available on TurboWarp :)
It will be enabled by default to all Scratch Addons users on the upcoming release

addons/place-backpack-code-at-cursor/addon.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.36.0",
+  "tags": ["editor", "codeEditor", "recommended"],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "enabledByDefault": false
```
Samq64, line 19:
```suggestion
  "enabledByDefault": true
```

Samq64 (approved):

addons/place-backpack-code-at-cursor/addon.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.36.0",
+  "tags": ["editor", "codeEditor", "recommended"],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "enabledByDefault": true
```
WorldLanguages, line 19: 🎉

WorldLanguages:
> As such I believe it would be an overeaction to begin adding any try/catches to this.

I just added 2 checks to make sure other extension features (Redux and getWorkspace) can break without causing the backpack to become unusable for 99% of users.
Try/catch was unnecessary because Scratch calls shareBlocksToTarget inside a promise chain, so any potencial exceptions wouldn't trigger any crashes or data loss.

addons/place-backpack-code-at-cursor/addon.json
WorldLanguages: Thanks samq for the reminder.

WorldLanguages (approved):

PR #7030, opened 2024-01-02
New addon: "delete other assets" (`delete-others`)

### LINKED ISSUE #4692: "Delete others" button on costumes and sounds context menu
This option is useful for deleting all costumes/sounds from a sprite. To delete all costumes, you create a new one, then right click it and choose "delete others" (sprites must have at least 1 costume). To delete all sounds, you right click any and choose "delete others", then delete the remaining one manually.

Old implementation available at #3406 (before the existence of `addon.tab.createEditorContextMenu` #3429 and `addon.tab.createModal` - we should use those now)

Jazza-231:
Resolves #4692 

### Changes

Adds a button to the right click menu for costumes, backdrops, and sounds to delete all the others.

### Reason for changes

Apparently people want this.

### Tests

Tested on brave, restore deleted costumes etc work, i think its all good. Please suggest changes to addon.json name and desc etc


OneShot-Niko:
Shouldn't this be marked dangerous?

Joeclinton1 (changes requested):
The functionality works well, I just have some nitpick comments on the code:

- Can you use a single decrementing for loop to delete the costumes? This is a common and recommended method for modifying an array that you are also iterating over.
- Instead of having two separate parts of logic for sound and costume, can you combine them into a unified loop by using a ternary operations for the functions that differ.
-- E.g `let deleteFunc = ctx.type === "costume" ? target.deleteCostume.bind(target) : target.deleteSound.bind(target);`
`let getLengthFunc = ctx.type === "costume" ? () => target.getCostumes().length : () => target.getSounds().length;` and then you can use these functions in the unified code
- can you unify the getRestoreCostumeFun and getRestoreSoundFun by making it take "costume" or "sound" as an input and performing the right logic.

Jazza-231:
I'll do it later™

Samq64:
> Shouldn't this be marked dangerous?
> EDIT: Oh wait, there's a confirmation. Didn't notice.

I'm not saying it needs to be but disable auto-save also has a confirmation but it's still marked as dangerous.

Jazza-231:
@Joeclinton1 done, done and done

WorldLanguages:
> Tested on brave, restore deleted costumes etc work,

So does it restore all of the deleted costumes/sounds? I think this is quite important and we didn't mention it in #4692. It's sometimes easy to forget about undo.

Jazza-231:
It sure does!

Secret-chest:
> I'm not saying it needs to be but disable auto save doesn't remove the confirmation but it's still marked as dangerous.

It's a browser native confirmation, and it should be dangerous because you will forget to save, here you need to invoke it manually.

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,64 @@
...
+
+  addon.tab.createEditorContextMenu(
+    (ctx) => {
+      if (window.confirm(`Are you sure you want to delete all other ${ctx.type}s?`)) {
+        const type = ctx.type === "costume" ? "Costume" : "Sound";
+
+        deletedItems = [];
+        target = vm.editingTarget;
+
+        const loop = ctx.type === "costume" ? target.getCostumes().length - 1 : target.getSounds().length - 1;
```
Joeclinton1, line 30: "loop" is a strange naming choice.

Joeclinton1 (approved):
Works well, and code is much cleaner now. Well done!

addons/delete-others/userscript.js
Jazza-231: I felt it was easier than "numberOfAssets" imo it's fine

Jazza-231: If you need me to, I'll change it, it's not like I've typed a full variable name in vsc after its declared

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,64 @@
...
+        vm.addSound(sound, target.id);
+      });
+    }
+    deletedItems = [];
+  }
+
+  addon.tab.createEditorContextMenu(
+    (ctx) => {
+      if (window.confirm(`Are you sure you want to delete all other ${ctx.type}s?`)) {
+        const type = ctx.type === "costume" ? "Costume" : "Sound";
```
Samq64, line 25: What's the capitalization for?

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,64 @@
...
+
+        deletedItems = [];
+        target = vm.editingTarget;
+
+        const loop = ctx.type === "costume" ? target.getCostumes().length - 1 : target.getSounds().length - 1;
+
+        for (let i = loop; i > -1; i--) {
```
Samq64, line 32: Using a decrementing loop here seems a little weird. I see a `for...in` loop wasn't used instead though.

Samq64 (commented):
Could the menu item be made orange by marking it as dangerous? It would also be nice if the restore message was made plural but that isn't really necessary.

addons/delete-others/userscript.js
Samq64: Actually you could use a `for..of` loop here.

addons/delete-others/userscript.js
mxmou: This should be translatable, with separate strings for costumes and sounds. `addon.tab.confirm` is the preferred way to ask for confirmation.

Jazza-231:
> Could the menu item be made orange by marking it as dangerous

I added the dangerous property and set it to true as a test and it seemed to not change the colour? I'm not sure why, perhaps I did something wrong, could someone suggest the code change so that I can see if I'm doing it right?

> It would also be nice if the restore message was made plural but that isn't really necessary.

Yeah that annoyed me too, I'll see if it's decently easy for me to do.

addons/delete-others/userscript.js
Jazza-231: > What's the capitalization for?

Scratch's Redux state for the deletedItem is capitalised 

> This should be translatable, with separate strings for costumes and sounds. `addon.tab.confirm` is the preferred way to ask for confirmation.

Alrighty

addons/delete-others/userscript.js
Jazza-231: Funny, @Joeclinton1 was adamant it was the preferred solution here, and after using it, I do agree. There is no array of assets to loop over as that just doesn't make sense, and due to the fact that deleting assets before the asset you don't want deleted changes the index of the target asset, it's a lot simpler to use the decrementing loop

Joeclinton1:
I don't think this should be marked as dangerous, since you can just restore the costumes, there's nothing dangerous about it. This also doesn't remove functionality it adds functionality.

DNin01:
> I don't think this should be marked as dangerous, since you can just restore the costumes, there's nothing dangerous about it. This also doesn't remove functionality it adds functionality.

I think they were talking about setting the button as dangerous, to make it use the color orange when hovered over.

Jazza-231:
image
image
It seems that even with dangerous: true, the button does not change colour when hovered. Tbh, I don't think it needs to be dangerous, there is a confirmation, and you can restore the costumes easily

mxmou:
> It seems that even with dangerous: true, the button does not change colour when hovered.

That's a bug in editor dark mode. It does become orange on hover if dark mode is disabled.

Jazza-231:
Oh ok, I'll add it then

Jazza-231:
The restore button now properly reflects how many assets are being restored (aka it's pluralised)

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,83 @@
...
+
+  function showDeleteOthers(ctx) {
+    if (ctx.type === "costume") {
+      return vm.editingTarget.getCostumes().length > 1;
+    } else {
+      return vm.editingTarget.getSounds().length > 1;
+    }
+  }
+
+  window.addon = addon;
```
Samq64, line 71:
```suggestion
```

Jazza-231: Oops

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+    deletedItems = [];
+  }
+
+  addon.tab.createEditorContextMenu(
+    async (ctx) => {
+      if (
+        await addon.tab.confirm(
+          msg(ctx.type === "costume" ? "confirmCostume" : "confirmSound"),
+          msg(ctx.type === "costume" ? "infoCostume" : "infoSound")
+        )
```
mxmou, line 28:
```suggestion
        await addon.tab.confirm(
          msg(ctx.type === "costume" ? "confirmCostume" : "confirmSound"),
          msg(ctx.type === "costume" ? "infoCostume" : "infoSound"),
          { useEditorClasses: true }
        )
```
This will make the popup more similar to other popups in the editor and affected by editor dark mode. Editor popups are smaller, so the title is too long - you could change it to "Delete Others" and move the current title to the description.

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+  const vm = await addon.tab.traps.vm;
+
+  const types = ["sound", "costume"];
+  let deletedItems = [];
+  let deleted;
+  let target;
+
+  function getRestoreFun(type) {
+    if (type === "costume") {
+      deletedItems.forEach((costume) => {
```
mxmou, line 11: This restores the items in the wrong order.

addons/delete-others/addon.json
mxmou, line 1: I think you can add `dynamicEnable` and `dynamicDisable`. The menu option will be automatically hidden when the addon is disabled.

mxmou (changes requested):

addons/delete-others/addon.json
Jazza-231: From my testing, it seems not.

Jazza-231: Dynamic enable works but not disable

Jazza-231: Nvm, see https://github.com/ScratchAddons/ScratchAddons/pull/7030/commits/2dde30912f014584f65e63074c099a284a793101

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+    deletedItems = [];
+  }
+
+  addon.tab.createEditorContextMenu(
+    async (ctx) => {
+      const deleteOthersStr = msg("deleteOthers");
+      if (
+        await addon.tab.confirm(
+          deleteOthersStr.charAt(0).toUpperCase() + deleteOthersStr.slice(1),
```
mxmou, line 28: I think this should be a separate string. In English, it should be "Delete Others", not "Delete others".

Jazza-231: Fair

mxmou (approved):

Joeclinton1 (changes requested):
It needs a few changes:

1. Change the variable name `loop` to something more descriptive like `numAssets`
1. Add a redux condition to the waitForElement that checks if the edit menu is open. The redux state is found at `Redux.state.scratchGui.menus.editMenu`. This is an easy optimisation.
1. The order of items isn't kept the same when restoring sounds. There must be some kind of bug there.

Jazza-231:
1. done
2. done
3. i have no idea what could be causing this

Joeclinton1:
I looked into the "sounds being out of order" bug and it seems that the function addSounds has different logic to addCostumes, so it's kind of a bug on Scratch's part. You might not be able to fix this.

Joeclinton1:
> I looked into the "sounds being out of order" bug and it seems that the function addSounds has different logic to addCostumes, so it's kind of a bug on Scratch's part. You might not be able to fix this.

I think the solution is clear. You should pollute the addSounds and fix their bug.

Jazza-231:
> > I looked into the "sounds being out of order" bug and it seems that the function addSounds has different logic to addCostumes, so it's kind of a bug on Scratch's part. You might not be able to fix this.
> 
> I think the solution is clear. You should pollute the addSounds and fix their bug. It's literally the only thing left for this addon to be finished.

Frankly both my motivation and spare time have plummeted, and I'm resigned to no addons of mine being added, if you or anyone else wants to do this please do, otherwise it will take me a while to do, if it all.

Jazza-231:
Ok, i have some spare time so ill work on this

Jazza-231:
@Joeclinton1 @WorldLanguages i dont know who else wants to test but im pretty sure this is all good now, if it could be merged that would make me happi uwu

Jazza-231:
This should be ready for final reviews and approvals

Joeclinton1 (approved):
LGSTM!

addons-l10n/en/delete-others.json
```diff
@@ -0,0 +1,8 @@
+{
+  "delete-others/deleteOthers": "delete others",
+  "delete-others/confirmTitle": "Delete Others",
+  "delete-others/infoCostume": "Are you sure you want to delete all other costumes?\nThe deleted costumes can be restored by going to Edit > Restore Costumes",
+  "delete-others/infoSound": "Are you sure you want to delete all other sounds?\nThe deleted sounds can be restored by going to Edit > Restore Sounds",
```
mxmou, line 5:
```suggestion
  "delete-others/infoCostume": "Are you sure you want to delete all other costumes?\nThe deleted costumes can be restored by going to Edit \u2192 Restore Costumes.",
  "delete-others/infoSound": "Are you sure you want to delete all other sounds?\nThe deleted sounds can be restored by going to Edit \u2192 Restore Sounds.",
```
`\u2192` is a right arrow (→).

addons/delete-others/addon.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Delete other assets",
+  "description": "Adds an option to the right click menu to easily delete all other costumes, backdrops, or sounds.",
+  "tags": ["editor", "costumeEditor", "recommended"],
+  "versionAdded": "1.37.0",
```
mxmou, line 5:
```suggestion
  "versionAdded": "1.39.0",
```

mxmou (approved):

WorldLanguages:
I'm not 100% sure if Jaza is active nowadays, so I'll make a few changes myself.

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,83 @@
...
+
+  while (true) {
+    const restoreButton = await addon.tab.waitForElement(
+      '[class*="menu-bar_menu-bar-item_"]:nth-child(4) [class*="menu_menu-item_"]:first-child > span',
+      { markAsSeen: true, reduxCondition: (state) => state.scratchGui.menus.editMenu }
+    );
+
+    const deletedItem = addon.tab.redux.state.scratchGui.restoreDeletion.deletedItem;
+
+    if (deletedItem && deletedItem.endsWith("s")) restoreButton.innerText = msg(`multi${deletedItem}`);
```
WorldLanguages, line 81: This seems like a very weird way to support localization here.

Are we absolutely sure zero languages have a sentence for "restore costume/backdrop/sound" which already ends with an S?

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,83 @@
+export default async function ({ addon, console, msg }) {
+  const vm = await addon.tab.traps.vm;
```
WorldLanguages, line 2:
```suggestion
  const vm = addon.tab.traps.vm;
```

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,83 @@
...
+            restoreFun: getRestoreFun.bind(this, ctx.type),
+            deletedItem: `${type}${deletedItems.length > 1 ? "s" : ""}`,
+          },
+        });
+      }
+    },
+    {
+      types,
+      position: "assetContextMenuAfterDelete",
+      order: 1,
```
WorldLanguages, line 57: This order number is already taken by `move-to-top-bottom`.

WorldLanguages: Nevermind, that's another shared space. This is below the delete button.

addons/delete-others/userscript.js
WorldLanguages: I changed it to use a variable instead. If it was necessary to change Redux state for some reason, please tell me. I already contacted the PR author.

addons/delete-others/userscript.js
```diff
@@ -0,0 +1,99 @@
...
+  const TYPES = ["sound", "costume"];
+  let deletedItems = [];
+  let target;
+  let shouldChangeRestoreButtonText = false;
+
+  function getRestoreFun(type) {
+    deletedItems.reverse();
+    if (type === "costume") {
+      deletedItems.forEach((costume) => {
+        vm.addCostume(costume.md5, costume, target.id, "");
```
WorldLanguages, line 13: I'm getting an error on this line.
`Uncaught TypeError: Cannot read properties of undefined (reading 'indexOf')`

See:
- https://github.com/scratchfoundation/scratch-vm/blob/f638db00447ce13b4a506068e52aef0ab1a9b641/src/import/load-costume.js#L359
- https://github.com/scratchfoundation/scratch-vm/blob/f638db00447ce13b4a506068e52aef0ab1a9b641/src/util/string-util.js#L34

WorldLanguages: The root cause is that `costume.md5` is undefined. I don't know if it's all costumes or only the ones I'm playing with.

WorldLanguages: Suggestion by GarboMuffin

```suggestion
        vm.addCostume(costume.assetId + "." + costume.dataFormat, costume, target.id, "");
```

addons/delete-others/userscript.js
WorldLanguages: Jazza confirmed using a local variable is fine.

addons/delete-others/addon.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Delete other assets",
+  "description": "Adds an option to the right click menu to easily delete all other costumes, backdrops, or sounds.",
```
mxmou, line 3: From Transifex:
image

WorldLanguages: Solved by #7579

addons/delete-others/addon.json
```diff
@@ -0,0 +1,15 @@
...
+  "tags": ["editor", "costumeEditor", "recommended"],
```
WorldLanguages, line 4: Recommended? @mxmou @Joeclinton1

WorldLanguages: I think we forgot to look at the tags

mxmou: It should probably be featured but not recommended, like `move-to-top-bottom`.

WorldLanguages: Ok, I opened PR #7733

PR #7050, opened 2024-01-05
Change the category of a few addons

### LINKED ISSUE #6367: Change Scratch Notifier Category
### Why this would be helpful

Scratch Notifier is in the website category which seems a little strange.

### My suggestion

Renaming the "Extension popup features" category and moving Scratch Notifier there.

### Possible alternatives

_No response_

### Additional context

_No response_

Samq64:
### Changes

- Adds the theme tag to `cat-blocks`, `columns` and `custom-block-text`.
- Removes the theme tag from `resizable-comment-input`.
- ~~Renames the extension popup features category and moves `scratch-notifier` there.~~

### Reason for changes

More consistent addon categorization.

### Tests

Tested on Firefox.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -114,7 +114,7 @@
     "message": "Website Themes"
   },
   "popupFeatures": {
-    "message": "Extension Features"
+    "message": "Extension Popup Features"
   },
   "theme": {
     "message": "Theme:"
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -114,7 +114,7 @@
     "message": "Website Themes"
   },
   "popupFeatures": {
-    "message": "Extension Popup Features"
+    "message": "Extension Features"
   },
   "theme": {
     "message": "Theme:"
--- a/addons/cat-blocks/addon.json
+++ b/addons/cat-blocks/addon.json
@@ -32,7 +32,7 @@
     }
   ],
   "versionAdded": "1.14.0",
-  "tags": ["editor", "theme", "featured"],
+  "tags": ["editor", "codeEditor", "theme", "featured"],
   "enabledByDefault": false,
   "libraries": ["scratch-blocks"]
 }
--- a/addons/cat-blocks/addon.json
+++ b/addons/cat-blocks/addon.json
@@ -32,7 +32,7 @@
     }
   ],
   "versionAdded": "1.14.0",
-  "tags": ["editor", "codeEditor", "featured"],
+  "tags": ["editor", "theme", "featured"],
   "enabledByDefault": false,
   "libraries": ["scratch-blocks"]
 }
--- a/addons/columns/addon.json
+++ b/addons/columns/addon.json
@@ -7,7 +7,7 @@
       "link": "https://scratch.mit.edu/TheColaber"
     }
   ],
-  "tags": ["editor", "codeEditor", "featured"],
+  "tags": ["editor", "theme", "featured"],
   "dynamicEnable": true,
   "dynamicDisable": true,
   "userscripts": [
--- a/addons/columns/addon.json
+++ b/addons/columns/addon.json
@@ -7,7 +7,7 @@
       "link": "https://scratch.mit.edu/TheColaber"
     }
   ],
-  "tags": ["editor", "theme", "featured"],
+  "tags": ["editor", "codeEditor", "theme", "featured"],
   "dynamicEnable": true,
   "dynamicDisable": true,
   "userscripts": [
--- a/addons/custom-block-text/addon.json
+++ b/addons/custom-block-text/addon.json
@@ -1,7 +1,7 @@
 {
   "name": "Customizable block text style",
   "description": "Lets you customize the style of text on blocks, including adjusting the size, using bold text, or adding a shadow.",
-  "tags": ["editor", "codeEditor", "featured"],
+  "tags": ["editor", "codeEditor", "featured", "theme"],
   "credits": [
     {
       "name": "Secret-chest",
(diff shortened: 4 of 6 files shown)


_locales/en/messages.json
```diff
@@ -114,7 +114,7 @@
     "message": "Website Themes"
   },
   "popupFeatures": {
-    "message": "Extension Popup Features"
+    "message": "Extension Features"
```
DNin01, line 117: I like the idea of this because the one addon that runs in the background, Scratch Notifier, in addition to all of the popup addons, perfectly fits in, but I wonder if there's a clearer way of putting this. Technically one could say that everything is an "extension feature", and "Extension Popup Features" was more self-explanatory about how those kinds of addons work.

But like @BroJac5246 said, if you think about it, if each category is a subset of all of Scratch Addons' addons and there's already an "All" tab, then logically this category won't contain every feature and therefore must be about something else (in this case, it's related to the non-website/editor features - those that are part of the extension's inner workings and the popup), so I think this works. I'd just say that if Scratch Notifier ever disappears, we could change this back to "Extension Popup Features".

addons/columns/addon.json
```diff
@@ -7,7 +7,7 @@
       "link": "https://scratch.mit.edu/TheColaber"
     }
   ],
-  "tags": ["editor", "codeEditor", "featured"],
+  "tags": ["editor", "theme", "featured"],
```
DNin01, line 10: This could have both `"codeEditor"` and `"theme"`, if I understand correctly.

addons/cat-blocks/addon.json
```diff
@@ -32,7 +32,7 @@
     }
   ],
   "versionAdded": "1.14.0",
-  "tags": ["editor", "codeEditor", "featured"],
+  "tags": ["editor", "theme", "featured"],
```
DNin01, line 35: This one too?

addons/columns/addon.json
Samq64: Yeah that works, I just removed them because themes aren't currently subcategoized, but they might be in the future.

DNin01 (approved):

DNin01:
We might also want to update the addon manifest documentation after releasing this.

_locales/en/messages.json
BroJac5246: @DNin01 did you mean to tag me? (Apparently the GitHub app's notifications aren't working for me so sorry for the late reply)

BroJac5246: I agree that referring to the popup directly isn't technically the most accurate, however "extension features" sounds way too ambiguous. As pointed out, everything, including all of the addons, are features of the extension. Also, many users won't determine the difference between "addon" and "extension," especially given that the two are pretty much identical, just one is browser-level while one is extension-level.

I would change it back unless a better wording can be found, which would be ideal. "Extension/Scratch Addons internal addons" or something similar might work, but what do you think?

Samq64: Would "Browser features" be any less confusing?

BroJac5246: > Would "Browser features" be any less confusing?

No because they're still not the browser's features.

Samq64: The only othee things I can think of are "Background features" and "Other features" which is already used to categorize non-featured addons. Your suggestion describes it well but I think it's a little too long.

mxmou: > "Extension/Scratch Addons internal addons"

The word "internal" makes it sound like these addons aren't intended for the user.

DNin01: A few alternative options:
- Separate the categories into two groups, "Scratch website" and "Scratch Addons", or something like that
- Don't list extension popup features as addons. Instead, make them toggleable through More Settings. I would imagine most people would want to keep these enabled anyway.
  - Scratch Notifier could then stay in Website Features or in some other category, potentially "Miscellaneous"/"Other".
  - And you'd probably configure what games display on the Cloud Games addon through the popup.
  - Wonder if the message count badge (or even Scratch Notifier) could be turned on/off from the Scratch Messaging addon in addition to More Settings, to make it easier to disable for those who don't need it?

BroJac5246: > > "Extension/Scratch Addons internal addons"
> 
> The word "internal" makes it sound like these addons aren't intended for the user.

> The only othee things I can think of are "Background features" and "Other features" which is already used to categorize non-featured addons. Your suggestion describes it well but I think it's a little too long.

These might make sense together, by replacing "internal" with "background"... though "Scratch Addons background addons" sounds a little weird.

Samq64: Would just "Background addons" be enough?

DNin01: > Would just "Background addons" be enough?

That doesn't sound right for popup features.

WorldLanguages: As far as I remember most of what you said I had as internal dialogue in my head when trying to name the category.
I would suggest against trying to propose changes to this category name unless it's the only change in the PR.

Samq64: Should I undo the Scratch notifier category change then?

WorldLanguages:
Changing the category of scratch-notifier to match other addons that don't change anything about the website does make some sense.

_locales/en/messages.json
DNin01: Yeah, probably the best thing to do right now would be to separate the category name change from this PR. Maybe even open an issue or discussion for it first.

DNin01 (approved):

mxmou (approved):

WorldLanguages (approved):

WorldLanguages:
@DNin01 It's an interesting idea, but I'm not sure how users would be able to figure out what a *widget* is.

Secret-chest:
> @DNin01 It's an interesting idea, but I'm not sure how users would be able to figure out what a widget is.

Extension Tabs and Features would be fine.

PR #7059, opened 2024-01-08
Shiftable anchor point

Jazza-231:
Resolves community suggestion with 112 upvotes: https://discord.com/channels/806602307750985799/1061421076019687495

### Changes

This is a draft pr. Please feel free to continue on my work or give me pointers on how to further achieve this.

### Reason for changes

People want it, its cool.

### Tests

In it's current and very limited state, it works fine in brave.

### More
There are a lot of things to do. Some I know how to do, some I may need help with, and all are welcome to add to it.
- Compatibility with the snapping addon (do we snap to the anchor point or the center or both?)
- Check that current selection actually has a set anchor point, if not, use its center. (should selection be a single item or group only? it may cause problems if this is not a restriction)
- Make the pivot point not move visually with the object when it is not locked relative to the object
- Allow said locking relative to the object
- Make the pivot point draggable, and also snappable with the snapping addon (if the snapping addon is disabled, still snap to the object center.)
- Theres probably more. This list will be updated accordingly

Good luck to me and anyone else who decides to take this on!

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -150,6 +150,7 @@
   "paint-skew",
   "preview-project-description",
   "collapse-footer",
+  "anchor-point",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/anchor-point/addon.json
+++ b/addons/anchor-point/addon.json
@@ -1,0 +1,18 @@
+{
+  "name": "Shiftable anchor point",
+  "description": "Allows you to rotate the selection around a point other than the center.",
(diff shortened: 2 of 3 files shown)


ItsEnderium:
Yall just forgot about this

Joeclinton1:
> Yall just forgot about this

Does anyone even want it?

ItsEnderium:
> > Yall just forgot about this
> 
> Does anyone even want it?

Me, it would help significantly with developing maps for 360 platformers
Maybe that’s too specific but I’m sure there are other use cases

PR #7065, opened 2024-01-11
Link to commit history on dev versions

### LINKED ISSUE #6451: Link to latest commit on pre-release vsersions
### Why this would be helpful

When working with pre-release versions of the extension it can sometimes be useful to see the latest commit or commit history on that branch without switching windows. The bug report template also asks users to fill in the commit hash and this might make that a little easier to do.

### My suggestion

In pre-release versions of the extension replace the link to the changelog in the popup with one to https://github.com/ScratchAddons/ScratchAddons/commits/commit-hash to easily see the state of the repo on that branch.

### Possible alternatives

Linking to commit directly or somehow replacing the pre-release text with the shortened commit hash.

### Additional context

I'm pretty sure this was mentioned somewhere but I don't think it had an issue yet.

Samq64:
Resolves #6451

### Changes

When installed logically the changelog link in the popup links to the commit history instead. Also I was wrong about `ORIG_HEAD`, fetching the actual latest commit requires two fetches as far as I can tell.

### Reason for changes

Linking to the release notes doesn't make much sense for development versions and it makes it easier to know exactly which version is being run.

### Tests

Tested on Brave. Maybe I'm just missing something obvious but I couldn't get fetch to work inside 
`computed` so I stored the hash in a global variable.

### DIFF
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -14,6 +14,9 @@ function calculatePopupSize() {
   document.body.classList.remove("loading");
 }
 
+const res = await fetch("../../.git/ORIG_HEAD");
+const commitHash = await res.text();
+
 window.addEventListener("load", () => setTimeout(calculatePopupSize, 0));
 
 const vue = new Vue({
@@ -53,10 +56,14 @@ const vue = new Vue({
   },
   computed: {
     changelogLink() {
-      const uiLanguage = chrome.i18n.getUILanguage();
-      const localeSlash = uiLanguage.startsWith("en") ? "" : `${uiLanguage.split("-")[0]}/`;
-      const utm = `utm_source=extension&utm_medium=popup&utm_campaign=v${chrome.runtime.getManifest().version}`;
-      return `https://scratchaddons.com/${localeSlash}changelog/?${utm}#v${chrome.runtime.getManifest().version}`;
+      if (chrome.runtime.getManifest().version_name.includes("-prerelease")) {
+        return `https://github.com/ScratchAddons/ScratchAddons/commits/${commitHash}`;
+      } else {
+        const uiLanguage = chrome.i18n.getUILanguage();
+        const localeSlash = uiLanguage.startsWith("en") ? "" : `${uiLanguage.split("-")[0]}/`;
+        const utm = `utm_source=extension&utm_medium=popup&utm_campaign=v${chrome.runtime.getManifest().version}`;
+        return `https://scratchaddons.com/${localeSlash}changelog/?${utm}#v${chrome.runtime.getManifest().version}`;
+      }
     },
     version() {
       const prerelease = chrome.runtime.getManifest().version_name.includes("-prerelease");
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -15,11 +15,9 @@ function calculatePopupSize() {
 }
 
 const commitHash = await fetch("../../.git/HEAD")
-.then(res => res.text())
-.then(data => data.replace("ref: ",""))
-.then(branch => fetch(`../../.git/${branch}`)
-.then(res => res.text())
-)
+  .then((res) => res.text())
+  .then((data) => data.replace("ref: ", ""))
+  .then((branch) => fetch(`../../.git/${branch}`).then((res) => res.text()));
 
 window.addEventListener("load", () => setTimeout(calculatePopupSize, 0));
 
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -14,8 +14,12 @@ function calculatePopupSize() {
   document.body.classList.remove("loading");
 }
 
-const res = await fetch("../../.git/ORIG_HEAD");
-const commitHash = await res.text();
+const commitHash = await fetch("../../.git/HEAD")
+.then(res => res.text())
+.then(data => data.replace("ref: ",""))
+.then(branch => fetch(`../../.git/${branch}`)
+.then(res => res.text())
+)
 
 window.addEventListener("load", () => setTimeout(calculatePopupSize, 0));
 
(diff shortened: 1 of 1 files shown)


webpages/popup/index.js
```diff
@@ -53,15 +59,18 @@ const vue = new Vue({
...
+        const utm = `utm_source=extension&utm_medium=popup&utm_campaign=v${chrome.runtime.getManifest().version}`;
+        return `https://scratchaddons.com/${localeSlash}changelog/?${utm}#v${chrome.runtime.getManifest().version}`;
+      }
     },
     version() {
-      const prerelease = chrome.runtime.getManifest().version_name.includes("-prerelease");
       const ver = chrome.runtime.getManifest().version;
-      return prerelease ? ver + "-pre" : ver;
+      return this.devMode ? ver + "-pre" : ver;
     },
```
DNin01, line 74: Not sure if this should base off of `this.devMode`. What does this function do?

webpages/popup/index.js
```diff
@@ -14,6 +14,11 @@ function calculatePopupSize() {
   document.body.classList.remove("loading");
 }
 
+const commitHash = await fetch("../../.git/HEAD")
+  .then((res) => res.text())
+  .then((data) => data.replace("ref: ", ""))
+  .then((branch) => fetch(`../../.git/${branch}`).then((res) => res.text()));
```
DNin01, line 20: If a branch name contains dots or slashes, would that become a problem here? Just making sure.

DNin01 (commented):
Looks good. Works fine on forks too.

A few questions though...

webpages/popup/index.js
Samq64: It displays `-pre` after the version number in the popup, so yeah it might make sense to revert that.

webpages/popup/index.js
Samq64: Git stores branches with slashes in subfolders so there's no problem there, I haven't tested dots yet.

DNin01 (commented):
Oh, also, the feature is kind of hidden away, and the tooltip for the version number still says "View changelog". To help with that, we could do one (or more) of a few things:
- Find another place to link to the commit history, like on a dedicated button in More Settings.
- On the bug report issue form and/or some sort of documentation (e.g. contributing guidelines), tell the user how they can access this feature.

And as well, it might be more efficient to directly copy the hash to the clipboard, instead of opening the commit history page, since the point is to supply it in bug reports.

webpages/popup/index.js
DNin01: Potentially relevant documentation: https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names

Samq64:
> - Find another place to link to the commit history, like on a dedicated button in More Settings.

I forgot more settings also links to the changelog and I haven't updated that one yet.

webpages/popup/index.js
DNin01: Seems like it's not even possible to create a branch with `..`:
Creating a branch named foo forward slash dot dot forward slash bar

Samq64:
I would also like to display the shortened hash instead of `-pre` but I'm not sure how to do that with just JavaScript.

webpages/popup/index.js
DNin01: But this probably could only fetch within Scratch Addons' own directory anyway, as the protocol of the fetch is `chrome-extension://`.

DNin01:
> I would also like to display the shortened hash instead of `-pre` but I'm not sure how to do that with just JavaScript.

Like, in place of the version number? The following should work, although I don't know if we want to show a commit hash _instead of_ a version number. Personally, I'd rather see the version number there.

`webpages/popup/index.js`
```diff
@@ -70,7 +70,7 @@ const vue = new Vue({
     },
     version() {
       const ver = chrome.runtime.getManifest().version;
-      return this.devMode ? ver + "-pre" : ver;
+      return this.devMode ? commitHash.substring(0, 7) : "v" + ver;
     },
   },
 });
```
`webpages/popup/index.html`
```diff
@@ -21,7 +21,7 @@
         <img src="../../images/icon-transparent.svg" id="logo" alt="Logo" />
         <span id="title-text" v-cloak>
           {{ msg("extensionName") }}
-          <a id="version" :href="changelogLink" target="_blank" title="{{ msg('changelog') }}">v{{ version }}</a>
+          <a id="version" :href="changelogLink" target="_blank" title="{{ msg('changelog') }}">{{ version }}</a>
         </span>
       </div>
       <div id="settings" @click="openSettingsPage()">
```

But if you were thinking of doing something slightly different with the version number, now you know how to modify it.

Samq64:
> Like, in place of the version number?

I was thinking something like v1.x.x (abc1234) but now that I think about it the -pre should probably be kept.

> `return this.devMode ? commitHash.substring(0, 7) : "v" + ver;`

Shorting the string like that isn't the problem, I just didn't realize that's how GitHub displays them for some reason.

Hans5958:
I don't think this would work well for cases when someone exports the ZIP to test the extension (which is a lot).

I think the only way to do this properly is using a build script so that it fetches the hash and puts it nicely for the extension to use.

DNin01:
> I don't think this would work well for cases when someone exports the ZIP to test the extension (and there's a lot of people who do that).

It could perhaps just not tell you the hash outside of a Git repository. (Wait, does that cause a crash with this PR?)

Regardless, this wouldn't have the same advantages on a fork or other branch since the commit hash is important for determining the "version" of the extension. The hash would be most meaningful on unmodified `ScratchAddons/ScratchAddons:master`, where people's own modifications aren't at play.

There's probably another way to get the commit hash, too. Personally, I access it via the "History" tab in GitHub Desktop. There's probably a way to do it with the command line too.

> I think the only way to do this properly is using a build script so that it fetches the hash and puts it nicely for the extension to use.

An interesting idea. Probably not worth it just for this, but if we ever create a build script for the whole project, this could be a bonus feature.

Samq64:
I might figure out a proper solution after Vue 3.

PR #7069, opened 2024-01-12
Add "Remove minimum height from signatures" addon

### LINKED ISSUE #7068: Remove the max height from signatures with `full-signature`
### Why this would be helpful

No matter how few space a signature actually needs, it'll *always* take up 168px with `full-signature`. This can be very annoying, especially for 1 line signatures (like mine).

### How this works

This would remove the minimum height from signatures.

### Additional context

**Before**
The signature takes up a lot of unnecessary space - it's only a line long

**After**
The signature only takes up one line

mybearworld:
Resolves #7068

### Changes

Add "Remove minimum height from signatures" addon.

<sup>(It's four lines of CSS, it can be moved into another addon pretty easily if need be)</sup>

### Reason for changes

> No matter how few space a signature actually needs, it'll always take up 168px. This can be very annoying, especially for 1 line signatures (like mine).

### Tests

Tested on Edge

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -151,6 +151,7 @@
   "paint-skew",
   "preview-project-description",
   "collapse-footer",
+  "signature-min-height",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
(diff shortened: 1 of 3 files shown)


mybearworld:
i'm stupid

PR #7071, opened 2024-01-12
`full-signature`: Don't force signature height to be 168px

### LINKED ISSUE #7068: Remove the max height from signatures with `full-signature`

mybearworld:
Resolves #7068

### Changes

`max-height: 168px;`

### Reason for changes

It makes more sense

### Tests

Tested in Edge

### DIFF
--- a/addons/full-signature/signature.css
+++ b/addons/full-signature/signature.css
@@ -1,7 +1,5 @@
 .postsignature {
   overflow-y: auto !important;
   resize: vertical;
-  max-height: -moz-fit-content !important;
-  max-height: fit-content !important;
-  height: 168px;
+  max-height: 168px;
 }


Secret-chest:
Should be a Force uniform height option

mxmou:
This removes the ability to make signatures taller than 168px by resizing. I don't know if there's a way to fix the bug while keeping the resize functionality with just CSS.

> Should be a Force uniform height option

Why would that be useful?

WorldLanguages:
I guess we'll need to add JS to the mix... Unfortunate but I'm not sure if we can do better

WorldLanguages:
> This removes the ability to make signatures taller than 168px by resizing

@mxmou Should we merge this anyway?
The minimum height issue is annoying, and it's enabled by default to all users.

mxmou:
> > This removes the ability to make signatures taller than 168px by resizing
> 
> @mxmou Should we merge this anyway? The minimum height issue is annoying, and it's enabled by default to all users.

If we do this, the resize functionality should be removed entirely - showing a resize handle doesn't make sense if the height can't actually be increased.

mybearworld:
I've removed the resizing from this PR. I doubt that was ever really used.

Samq64 (approved):

mxmou (approved):

Issue #5821, opened 2023-03-19
hide-flyout default to "open palette" 


sfederici:
### Why this would be helpful

I'm using a lot the hide-flyout addon. It works really well and it is extremely useful. As my students usually move from basic-Scratch to ScratchAddons they are surprised for not seeing the block palette (they are using a version preinstalled in the lab with all the necessary addons already enabled). 

### My suggestion

Change (or select by option) default for hide-flyout addon to palette open, instead of closed. Store the palette status -closed/open- in the browser, otherwise closing the palette every time could be annoying.

### Possible alternatives

None

WorldLanguages:
This new feature would need to be a setting. The palette is hidden by default by design, as seen in old pull requests. So if this is implemented, you'd need to enable a setting,

WorldLanguages:
> Store the palette status -closed/open- in the browser, otherwise closing the palette every time could be annoying.

@sfederici Could you clarify what you meant here?

sfederici:
> @sfederici Could you clarify what you meant here?

Sure. If I close the palette, that means that I acknowledged that it can be closed. So, next time I open Scratch, it is reasonable that the palette is closed. Storing the palette status (closed/open) in the browser it would be to me a good addition to the addon.

WorldLanguages:
@sfederici Hmm, I see what you mean. However, this will be added as a setting, so storing whether the user closed the palette at least once wouldn't make much sense, since the user can simply toggle the setting again.

sfederici:
ok

PR #7076, opened 2024-01-13
hide-flyout: new setting "open by default" (auto-hide block pallete addon)

callumjt:
adds a new option to Auto-hiding block palette which makes the flyout lock when the page is loaded 

tested on chrome and firefox :)

(sorry for low framerate, for some reason obs decided to die)

https://github.com/ScratchAddons/ScratchAddons/assets/116295240/e93ebaf6-017d-46e5-8193-a5f20e165a07

### DIFF
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -19,7 +19,10 @@
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "latestUpdate": "1.36.0",
+  "latestUpdate": {
+    "version": "1.36.0",
+    "newSettings": ["lockLoad"]
+  }.
   "customCssVariables": [
     {
       "name": "lockDisplay",
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -81,7 +81,11 @@
         {
           "id": "category",
           "name": "Category click"
-        }
+        },
+        {
+          "id": "pageload",
+          "name": "page load",
+        },
       ],
       "default": "cathover"
     },
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -19,6 +19,7 @@
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
+  "latestUpdate": "1.36.0",
   "customCssVariables": [
     {
       "name": "lockDisplay",
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -19,10 +19,6 @@
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
-  "latestUpdate": {
-    "version": "1.36.0",
-    "newSettings": ["lockLoad"]
-  }.
   "customCssVariables": [
     {
       "name": "lockDisplay",
@@ -122,8 +118,7 @@
   ],
   "tags": ["editor", "codeEditor", "recommended"],
   "latestUpdate": {
-    "version": "1.27.0",
-    "temporaryNotice": "This addon was revised and many bugs were fixed.",
-    "isMajor": true
+    "version": "1.36.0",
+    "newSettings": ["lockLoad"]
   }
 }
--- a/addons/hide-flyout/addon.json
+++ b/addons/hide-flyout/addon.json
@@ -110,7 +110,7 @@
       "default": "default"
     },
     {
-      "name": "Lock by default",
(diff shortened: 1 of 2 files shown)


CubesterYT:
I remember this was asked for in TurboWarp, I'm so stoked this is coming, I need it bad.

LilyMakesThings:
A dream come true, hope to see this in TurboWarp ASAP <3

DNin01 (changes requested):
Pretty cool, but this does need testing with the "Category click" setting. There's no lock button when you use that setting, but it could instead be programmed to start out open, requiring a category click to close. Related: #5821. What actually happens in this PR is **you can't close it**.

I like to set this addon to toggle on category click and I've always wanted it to start out open.

callumjt:
> Pretty cool, but this does need testing with the "Category click" setting. There's no lock button when you use that setting, but it could instead be programmed to start out open, requiring a category click to close. Related: #5821. What actually happens in this PR is **you can't close it**.
> 
> I like to set this addon to toggle on category click and I've always wanted it to start out open.

was making this and tried to set "motion" to enabled on game start but "Blockly.getMainWorkspace().getToolbox().selectedItem_.setSelected(false);" doesnt seem to work

DNin01:
@mxmou, do you want to take a look?

callumjt:
the code i used is 

`const option = addon.settings.get("lockLoad");
    if (option) {
      if (getToggleSetting() === "category") {
        Blockly.getMainWorkspace().getToolbox().selectedItem_.setSelected(true);
        onmouseleave(null, 0);
        toggle = true;
      } else {
        flyoutLock = option;
        updateLockDisplay()
      }
      flyOut.classList.remove("sa-flyoutClose");
      scrollBar.classList.remove("sa-flyoutClose");
    };`

mxmou:
> was making this and tried to set "motion" to enabled on page load but "Blockly.getMainWorkspace().getToolbox().selectedItem_.setSelected(true);" doesnt seem to work

That's because of the `setSelected(false)` call on line 191. Try adding `&& !toggle` to the `if (!shouldScroll)` condition. You can then remove line 53.

DNin01 (approved):
Seems to be working!

Only thing is the block category [doesn't initially appear highlighted](https://github.com/ScratchAddons/ScratchAddons/assets/106490990/6b86a6a9-0c81-4dc1-81e9-e82adb1c2eb1) if you've scrolled the block palette to the very top and then dynamically enable the addon using the "Category click" and "Lock by default" settings, but that's very minor.

addons/hide-flyout/addon.json
```diff
@@ -108,6 +108,12 @@
         }
       ],
       "default": "default"
+    },
+    {
+      "name": "Lock by default",
```
Samq64, line 113: The category click setting doesn't have a lock.
```suggestion
      "name": "Open by default",
```

Samq64 (commented):
Could `latestUpdate` be added to highlight the new setting?

callumjt:
Changed the option name and added latestUpdate

addons/hide-flyout/addon.json
```diff
@@ -19,6 +19,7 @@
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
+  "latestUpdate": "1.36.0",
```
Samq64, line 22:
```suggestion
  "latestUpdate": {
    "version": "1.36.0",
    "newSettings": ["lockLoad"]
  }.
```

Samq64 (approved):

addons/hide-flyout/userscript.js
```diff
@@ -46,6 +46,21 @@ export default async function ({ addon, console, msg }) {
     lockIcon.src = addon.self.dir + `/${flyoutLock ? "" : "un"}lock.svg`;
   }
 
+  function autoLock() {
+    const option = addon.settings.get("lockLoad");
+    if (option) {
+      if (getToggleSetting() === "category") {
+        onmouseleave(null, 0);
```
mxmou, line 53:
```suggestion
```
`onmouseleave()` closes the palette. There's no reason to call it here.

mxmou (changes requested):
Lines 151-155 (in the `"reenabled"` event listener) close the palette if the addon is dynamically enabled and the mode is category click. This should only happen if the new setting is disabled.

When switching from category click to a different mode, the palette should be automatically locked if the new setting is enabled.

mxmou (approved):

Issue #7034, opened 2024-01-03
Firefox injects content scripts dynamically


Chiroyce1:
### Description

-

### Steps to reproduce

- Make sure a few addons are enabled, then turn off SA entirely in Firefox
- Visit a Scratch project or any Scratch page
- Turn on SA from the browser addon menu, go back to the project
- SA's scripts are injected dynamically (+ certain project addons don't work as intended due to them being run after page load, ex: vol-slider)

### Environment

Should happen on most Firefox versions with any SA version

Issue #4374, opened 2022-03-08
Finer zoom on costume editor


astrosticks:
### Is your request related to a problem? Please describe.

The costume editor is always either too big (doesn't fit in the frame) or too little (only half as tall as the frame) for me. Screenshots below: 

Default zoom:
![image](https://user-images.githubusercontent.com/32315760/157141019-53b986c5-0ded-40d6-b4ed-50c928f6b27d.png)

After zooming in once:
![image](https://user-images.githubusercontent.com/32315760/157141031-11da10cd-99c8-464a-96f4-833f6c03d222.png)
 

### Describe the solution you'd like

I'd like the zoom in/out buttons to allow for some finer control of the zoom level.

### Describe alternatives you've considered

A zoom percentage input could also work?

Chiroyce1:
Doesn't the scroll wheel on a mouse or zooming on a trackpad add finer control?

DNin01:
Also nice would be to allow zooming in/out to be faster, for those who often zoom in huge amounts. Scratch 2 increased the zoom level exponentially with every `+` click, so you'd go from 100% to 200% to 400%... up to 1,600%.

WorldLanguages:
> The costume editor is always either too big (doesn't fit in the frame) or too little (only half as tall as the frame) 

I can relate to this

WorldLanguages:
From feedback:
> In the costume creation, when hovering over what you're working with, I think there should be an option to zoom using the mouse scroll. That would be extremely helpful.

I replied in their profile telling them about holding the Ctrl key. But I don't know if I was aware about this until trying it out 2 minutes ago.

Samq64:
This could be part of #7079.

PR #7079, opened 2024-01-14
Add Zoom slider addon

### LINKED ISSUE #7040: Replace zoom buttons with a slider
### Why this would be helpful

The zoom buttons themselves aren't really helpful (constantly changing `custom-zoom` can get annoying after a long while) as they are in 2 seperate positions as well rather than just using a slider (though the implementation of that is probably difficult so its fair)

### How this addon works

Replaces the 2 zoom buttons with a singular slider.

Possibly also display the zoom percentage, though that's not really needed

Could be added to `custom-zoom` rather than being a new addon possibly

### Possible alternatives

None that I could think of.

### Additional context

no

mybearworld:
Resolves #7040

### Changes

Add a "Zoom slider" addon that replaces the built in zoom buttons with a slider.

Some Scratch blocks and a slider at the bottom right, with an input field above that says "68%". Next to the slider is a reset button.

### Reason for changes

Oftentimes, it's more convenient.

### Tests

Tested with Edge. It's (hopefully) compatible with `custom-zoom`.


CubesterYT:
There was a button that reset zoom, seems like it was removed here?

OneShot-Niko:
> There was a button that reset zoom, seems like it was removed here?

Clicking the % (eg. 200%) will reset it, mybearworld is gonna add a note i think
currently I'm testing as well and its going up by 5% on librewolf (though its step is 1 and we talked about this on meower currently)

mybearworld:
Would this work?
The regular slider with the = button next to it

CubesterYT:
That's perfect

CubesterYT:
Ok this new push is everything we need.

OneShot-Niko:
> Ok this new push is everything we need. I'm in love.

Though off-topic, same here
this will literally resolve the one thing that I wish existed in vanilla scratch even in 1.4

Joeclinton1:
Instead of a new addon, I think it should be combined with "Customizable code area zoom". Reduces clutter.

mybearworld:
According to @OneShot-Niko, the autohide compatibility is currently not working on Firefox

OneShot-Niko:
> According to @OneShot-Niko, the autohide compatibility is currently not working on Firefox

(Note: turned out this was just #7034 and I forgot to reload)

addons/zoom-slider/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Zoom slider",
+  "description": "Replaces the regular zoom buttons with a slider.",
+  "tags": ["editor"],
```
Samq64, line 4:
```suggestion
  "tags": ["editor", "codeEditor"],
```

Samq64 (commented):
Everything seems to work as expected but is there an easy way to prevent the zoom controls from hiding while the slider is being dragged? Also I wonder if the animation should be vertical with this addon since it's shorter and wider with the slider.

Samq64:
Also does the size input look slightly better to the right of the slider?

mybearworld:
> Also I wonder if the animation should be vertical with this addon since it's shorter and wider with the slider.

That does look better.

mybearworld:
> Also does the size input look a little better if it's to the right of the slider?

I don't really like this, personally.

The input next to the slider

Samq64:
> > Also does the size input look a little better if it's to the right of the slider?
> 
> I don't really like this, personally.
> 
> The input next to the slider

I was thinking with the reset button on above it but that won't quite line up either.

OneShot-Niko:
> > Also does the size input look a little better if it's to the right of the slider?
> 
> I don't really like this, personally.
> 
> The input next to the slider

I guess that could be allowed to get changed by the user, if someone does like [it] that way?
Personally I'm not a fan of it either, though it might be a good thing to keep in mind.

OneShot-Niko:
> I don't think that level of customization is necessary. (Especially as that isn't just a CSS change, but requires extra JS to make work.)

Fair.

mybearworld:
> is there an easy way to prevent the zoom controls from hiding while the slider is being dragged? 

Only with `:has` (which isn't available cross-browser), I think.

mybearworld:
> is there an easy way to prevent the zoom controls from hiding while the slider is being dragged?

`<input type="range">` elements don't seem to lose their focus when they're being released, which makes doing that not particularly simple

addons/zoom-slider/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Zoom slider",
+  "description": "Replaces the regular zoom buttons with a slider.",
```
Samq64, line 3: It would probably make sense to also mention the number input.
```suggestion
  "description": "Replaces the code area zoom buttons with a slider and number input.",
```

addons/zoom-slider/userscript.js
```diff
@@ -0,0 +1,79 @@
+export default async function ({ addon, msg }) {
+  const addSlider = async () => {
+    if (addon.tab.editorMode !== "editor" || addon.self.disabled) {
+      return;
+    }
+
+    const zoom = await addon.tab.waitForElement(".blocklyZoom");
+    const blockly = await addon.tab.traps.getBlockly();
+    window.blockly = blockly;
```
Samq64, line 9: I'm pretty sure this isn't necessary? Correct me if I'm wrong.

addons/zoom-slider/userscript.js
```diff
@@ -0,0 +1,79 @@
...
+    foreignObject.setAttribute("width", "120");
+    foreignObject.setAttribute("x", "-126");
+    foreignObject.setAttribute("y", "79");
+    const wrapperDiv = document.createElement("div");
+    wrapperDiv.classList.add("sa-zoom-slider-wrapper");
+    const slider = document.createElement("input");
+    slider.type = "range";
+    slider.min = blockly.mainWorkspace.options.zoomOptions.minScale;
+    slider.max = blockly.mainWorkspace.options.zoomOptions.maxScale;
+    slider.step = "0.01";
```
Samq64, line 22: Maybe I'm missing something but if you're already rounding the scale why not set the step to an integer?

addons/zoom-slider/userscript.js
```diff
@@ -0,0 +1,79 @@
...
+
+    addon.self.addEventListener("disabled", () => {
+      foreignObject.remove();
+    });
+  };
+
+  const percentScale = (scale) => Math.round(parseFloat(scale) * 100);
+
+  addSlider();
+  addon.tab.addEventListener("urlChange", addSlider);
```
Samq64, line 77: This will also try to add the slider to the project. The easiest way to fix this is probably to put the `addon.tab.waitForElement` in a `while (true)` loop like other editor addons.

addons/zoom-slider/userscript.js
mybearworld: Oh, you're right, I put that there for debugging.

addons/zoom-slider/userscript.js
mybearworld: I don't think the scale's being rounded to an integer - that's only for displaying it as a percentage.

addons/zoom-slider/userscript.js
mybearworld: It won't add it to the project page, though:
https://github.com/mybearworld/fork-ScratchAddons/blob/06d9d6564eed83f8a17a6300b2569477c53fea7a/addons/zoom-slider/userscript.js#L3-L5
It does probably make sense to make it have the same pattern as other addons, though.

Samq64: Right, I'm not sure if it's getting removed properly though because of these errors:
image

DNin01 (changes requested):
The act of resizing the workspace seems to cause the right-click context menu to disappear.

DNin01 (commented):
Design-wise, I think it might be better if the zoom level input field was to the right or left of the slider instead of above. If possible, a percent symbol would also be nice to show, the input field might look better smaller, and the input could be adjustable with the arrow keys.

DNin01:
> The script repeatedly checks the workspace's zoom level to update the slider and input with the current zoom level. In doing so, it dispatches an `input` event on the slider, updating the actual zoom level at this line.

> The act of resizing the workspace seems to cause the right-click context menu to disappear.

Wait, why isn't the if-then condition on line 62 stopping this from happening unnecessarily?

Console message reading, "0.68, number; 0.68, string"

Oh.

DNin01:
By the way, fixing the if-then condition mistake would only fix the context menu problem, not the zoom jitter.

DNin01:
Give me a minute...

addons/zoom-slider/userscript.js
```diff
@@ -0,0 +1,78 @@
...
+      }
+      slider.value = count.value / 100;
+      slider.dispatchEvent(new Event("input"));
+      countInput = false;
+    });
+
+    setInterval(() => {
+      slider.min = blockly.mainWorkspace.options.zoomOptions.minScale;
+      slider.max = blockly.mainWorkspace.options.zoomOptions.maxScale;
+      if (parseFloat(slider.value) !== blockly.mainWorkspace.scale) {
```
DNin01, line 62: Either of these would fix that.
```suggestion
      if (slider.value !== blockly.mainWorkspace.scale) {
```
```suggestion
      if (parseFloat(slider.value) !== parseFloat(blockly.mainWorkspace.scale)) {
```

addons/zoom-slider/userscript.js
```diff
@@ -0,0 +1,78 @@
...
+    count.addEventListener("blur", async () => (countInput = false));
+    count.addEventListener("change", async () => {
+      const newScale = parseFloat(count.value / 100);
+      const { maxScale, minScale } = blockly.mainWorkspace.options.zoomOptions;
+      if (newScale > maxScale) {
+        count.value = maxScale * 100;
+      }
+      if (newScale < minScale) {
+        count.value = minScale * 100;
+      }
```
DNin01, line 53: These can be condensed if you want.
```suggestion
      if (newScale > maxScale) count.value = maxScale * 100;
      if (newScale < minScale) count.value = minScale * 100;
```

DNin01:
I have a more complex change waiting once the above two suggestions have been resolved.

addons/zoom-slider/userscript.js
DNin01: Actually, this can be fixed by not setting the zoom level to a string.

addons/zoom-slider/userscript.js
```diff
@@ -0,0 +1,78 @@
...
+    count.value = percentScale(blockly.mainWorkspace.scale);
+    wrapperDiv.append(count, slider);
+    foreignObject.append(wrapperDiv);
+    zoom.append(foreignObject);
+
+    slider.addEventListener("mousedown", (e) => {
+      e.stopPropagation();
+    });
+    slider.addEventListener("input", async () => {
+      blockly.mainWorkspace.setScale(slider.value);
```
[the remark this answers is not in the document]
DNin01: I fixed the unnecessary repeated zoom level updating which means the context menu bug is now fixed, but the zoom jitter is still there because it's still rounding off.

mybearworld:
> Design-wise, I think it might be better if the zoom level input field was to the right or left of the slider instead of above. If possible, a percent symbol would also be nice to show, the input field might look better smaller, and the input could be adjustable with the arrow keys.

This doesn't look that great.

The input next to the slider

OneShot-Niko:
> `editor-number-arrow-keys` should probably do that - not sure if it's compatible, though.

Incompatible, as far as my testing goes on LibreWolf - However it does work with the slider (though that was honestly obvious by a long shot)
Up goes to the start and down goes to the end instead:


https://github.com/ScratchAddons/ScratchAddons/assets/150537842/7e4a9641-d155-4bea-8787-14e2ad7fdf95

addons/zoom-slider/userscript.js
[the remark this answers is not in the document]
mybearworld: What do you mean by zoom jitter? I don't think I notice anything like that when using it.

mybearworld:
The addon should now be compatible with `editor-number-arrow-keys`.

OneShot-Niko:
> The addon should now be compatible with `editor-number-arrow-keys`.

Appears that it won't change until the user submits - Unsure if intentional, though.


https://github.com/ScratchAddons/ScratchAddons/assets/150537842/fb9e7739-043e-4c12-aab4-ed70600992cb

mybearworld:
> Appears that it won't change until the user submits - Unsure if intentional, though.

You'd think I'd have noticed that 🤦
Fixing...

mybearworld:
> > Appears that it won't change until the user submits - Unsure if intentional, though.
> 
> You'd think I'd have noticed that 🤦 Fixing...

Fixed. The behavior is a bit choppy, but that seems to be intended.

DNin01:
> `editor-number-arrow-keys` should probably do that - not sure if it's compatible, though.

Arrow key incrementation could be built-in, without requiring another addon. The outline and brush size inputs in the costume editor have this ability.

`editor-number-arrow-keys` could still be compatible just so you can use the Shift+arrow key and Alt+arrow key shortcuts on it.

addons/zoom-slider/userscript.js
mybearworld: I don't see these errors. What did you do to get them?

Samq64: Switch between the editor and project page.

mybearworld: Weird, I don't get these errors when doing that. What browser are you using where that occurs?

Samq64: A recent commit fixed it.

addons/zoom-slider/userscript.js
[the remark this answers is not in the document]
Samq64: That seems to have also been fixed.

Samq64 (approved):

addons/zoom-slider/userscript.js
mybearworld: Oh, that's good

addons/zoom-slider/userscript.js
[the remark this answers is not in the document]
mybearworld: Also good

OneShot-Niko:
Could this also be enabled by default? Asking out of curiousity, since it's a lot easier to use than the two arrows.

BroJac5246:
> Could this also be enabled by default? Asking out of curiosity, since it's a lot easier to use than the two arrows.

I'm not a big fan of the current design and wouldn't be very happy if that replaced the current system.

mybearworld:
> > Could this also be enabled by default? Asking out of curiosity, since it's a lot easier to use than the two arrows.
> 
> I'm not a big fan of the current design and wouldn't be very happy if that replaced the current system.

Do you have any idea for improvement?

DNin01:
> Could this also be enabled by default? Asking out of curiousity, since it's a lot easier to use than the two arrows.

It's a really cool addon, but we don't usually default to enabling most types of addons. This one in particular replaces a familiar feature and is something that a user would probably want the choice to configure first. We could recommend it, though.

Secret-chest:
Can we have separate recenter and rescale buttons?

BroJac5246:
> > > Could this also be enabled by default? Asking out of curiosity, since it's a lot easier to use than the two arrows.
> > 
> > I'm not a big fan of the current design and wouldn't be very happy if that replaced the current system.
> 
> Do you have any idea for improvement?

Maybe make the text field fade out when you're not hovering in that area — most of the time people don't need the exact number.

mybearworld:
When the autohide setting from `custom-zoom` is enabled and `zoom-slider` is then dynamically enabled/disabled, the zoom area doesn't update. This should no longer happen if #7258 gets merged.

WorldLanguages:
Related feedback: https://github.com/ScratchAddons/ScratchAddons/issues/4374#issuecomment-2037490624

mybearworld:
Currently not compatible with the hide backpack addon

mybearworld:
> Currently not compatible with the hide backpack addon

Should be fixed with https://github.com/ScratchAddons/ScratchAddons/pull/7079/commits/05ff12c46e0371525608b83a5832f5c1f86460d2

mybearworld:
Merging #7258 into this breaks it right now, I'll look at that tomorrow

mybearworld:
> Merging #7258 into this breaks it right now, I'll look at that tomorrow

Fixed, this PR should be fine now

PR #7090, opened 2024-01-21
gamepad: Mark Firefox 123 and later on Linux as not broken

GarboMuffin:
With https://bugzilla.mozilla.org/show_bug.cgi?id=1680982 being resolved in Firefox 123 (currently nightly), Firefox on Linux has a pretty good gamepad API implementation, so we can hide the warning that "This browser and operating system have known bugs that may make this addon difficult to use. Try another browser if you encounter problems"


Samq64 (approved):
Tested on Firefox stable and nightly on Linux. The notice is displayed as expected.

WorldLanguages (approved):

PR #7091, opened 2024-01-21
Blockly Developer Tools Addon

ZXMushroom63:
This addon makes modifying block data and creating hacked blocks easier. It allows power-users to do almost anything they like within their Scratch projects, which is quite fun.

### Changes
It adds an edit button next to every top level block, which when clicked opens an editor showing the block's data as XML.

### Tests

It works flawlessly in Chrome, in Firefox the buttons have trouble rendering correctly.


OneShot-Niko:
Considering that hacked blocks already have high warrant of getting NFE'd, I don't think making this would be a really good idea unless it's an easter-egg addon possibly.

Just mentioning.

CST1229:
> EDIT2: Nevermind - [Found one example of an NFE hacked blocks project](https://scratch.mit.edu/projects/315757409).
[Another one.](https://scratch.mit.edu/projects/382631312)

CubesterYT:
Oh my god, I've been thinking of this idea ever since I joined Scratch Addons a long time ago, but I never thought it was possible. You are amazing!

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,402 @@
+export default async function ({ addon, console }) {
+    await addon.tab.waitForElement("svg.blocklySvg>g.blocklyWorkspace");
+    if (!Blockly) { //If the Blockly global is not exposed, the script cannot do anything useful.
+        return "Failed to find Blockly instance!";
+    }
+    var workspace = Blockly.getMainWorkspace(); //Get the main Blockly workspace
```
pufferfish101007, line 6: you should use `addon.tab.getWorkspace()` instead

CST1229: and probably also `const Blockly = await addon.tab.getBlockly()` instead of waiting for an element then trying to find window.Blockly

ZXMushroom63: Ok. When I wrote the addon I wrote it to work on any instance of blockly, Scratch or not. (It worked on Scratch, TurboWarp and Google's blockly demo with no modification). I just sort of did a copypaste for the addon. :)

I'll try to rewrite it a bit to work better with ScratchAddons. (Or, at least, look neater.)

ZXMushroom63:
> Considering that hacked blocks already have high warrant of getting NFE'd, I don't think making this would be a really good idea unless it's an easter-egg addon possibly.
> 
> Just mentioning. EDIT: Atleast from what I remember, most hacked-block projects can get NFE'd. May be remembering wrong here... EDIT2: Nevermind - [Found one example of an NFE hacked blocks project](https://scratch.mit.edu/projects/315757409).

ScratchAddons already allows for creating "hacked" key blocks. Either way, this is still a valid concern, but I think that marking it as "dangerous" probably warns people enough.

ZXMushroom63:
> Oh my god, I've been thinking of this idea ever since I joined Scratch Addons a long time ago, but I never thought it was possible. You are amazing!

Thanks!

ZXMushroom63:
I've figured out why firefox is not working.
The `overflow: visible` property I'm using still won't show content in Firefox unless the element has a positive size. 🙄
Not impressed with mozilla.

ZXMushroom63:
I fixed firefox support. Turns out that somewhere there is an optimisation to only render elements with a positive size.

addons/blocklydevtools/userscript.js
ZXMushroom63: Done

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,419 @@
...
+        container.style.maxHeight = "100vh";
+        container.style.overflowY = "scroll";
+        container.style.caretColor = "white";
+
+        //Disable spellcheck and autocomplete.
+        container.setAttribute("autocomplete", "false");
+        container.setAttribute("spellcheck", "false");
+
+        //Set the editor's contents to the formatted original XML.
+        container.innerHTML = formatXml(originalXml, true);
```
Samq64, line 169: Ideally all uses of `innerHTML` would be replaced with `innerText` or `textContext` where possible.

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,419 @@
...
+            }
+            xmlStr = xmlStr.replace(originalXml, container.textContent); //Replace original XML with modified XML.
+            //xmlStr = xmlStr.replace(/\u2003/g, "").replace(/\n/g, "");
+            xmlStr = xmlStr.replace(/\u00A0/g, "\u0020"); //Replace any non-breaking spaces with normal ones.
+            dom = (Blockly.Xml.textToDom || Blockly.utils.xml.textToDom)(xmlStr); //Update the DOM variable.
+            workspace.clear(); //Clear the workspace
+            Blockly.Xml.domToWorkspace(dom, workspace); //Load the DOM
+            if (Blockly.getMainWorkspace().getToolbox() //If the blockly instance has a toolbox, it needs to be refreshed,
+            ) {
+                Blockly.getMainWorkspace().getToolbox().refreshSelection();
```
Samq64, line 195: Can't the existing `workspace` here?
```suggestion
            if (workspace.getToolbox() //If the blockly instance has a toolbox, it needs to be refreshed,
            ) {
                workspace.getToolbox().refreshSelection();
```

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,419 @@
...
+
+            btnWrapper.style.cursor = "auto"; //Fix for cursor being the grab hand
+            btnWrapper.style.width = "max-content";
+
+            var btn = document.createElement("span"); //Create the edit button
+            btn.style.cursor = "pointer";
+            btn.style.zIndex = "999";
+            btn.style.lineHeight = "1rem";
+            btn.setAttribute("data-is-blocklydev-btn", "true");
+            btn.innerHTML = "✏️"; //📝✏️
```
Samq64, line 244: Ideally all these emojis would be proper icons instead I don't think that has to be done now.

Samq64 (commented):
We try to define variables with `const` and `let` in most places.

Samq64:
How did I not notice this was a draft?

WorldLanguages:
Screenshot looks very cool.

> I don't think making this would be a really good idea unless it's an easter-egg addon possibly.

Yeah, we could consider this as an easter-egg I think.

addons/blocklydevtools/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Blockly Developer Tools",
+    "description": "Adds buttons next to blocks in the editor that allow you to collapse, force delete and modify their XML data. Only enable if you know what you're doing.",
```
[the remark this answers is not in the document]
ZXMushroom63: Yep! I didn't know this was possible to add warnings, but it is in fact in the doc.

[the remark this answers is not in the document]
ZXMushroom63: Done.

addons/addons.json
```diff
@@ -176,5 +176,6 @@
   "editor-buttons-reverse-order",
   "columns",
   "editor-compact",
-  "editor-theme3"
+  "editor-theme3",
+  "blocklydevtools"
```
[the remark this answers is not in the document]
ZXMushroom63: Done.

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,419 @@
+export default async function ({ addon, console }) {
+    var isScratchAddons = true; //Should the script use ScratchAddons APIs.
```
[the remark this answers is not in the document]
ZXMushroom63: I would prefer not to remove the checks, but if it is important enough, I might just leave a note as a comment explaining how to use the script without the ScratchAddons APIs.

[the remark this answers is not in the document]
ZXMushroom63: Done

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,419 @@
...
+            return blockElem.querySelector("g.blocklyDraggable[data-id]:not(.blocklyInsertionMarker)>path.blocklyPath[d]");
+        }
+    }
+    function isNewBlocklyBlock(node) { //Is node a block svg that has not yet been hooked.
+        return node && node instanceof Element && node.tagName === "g" && node.classList.contains("blocklyDraggable") && !node.classList.contains("blocklyInsertionMarker") && node.hasAttribute("data-id") && !node.hasAttribute("data-blocklydev-hooked");
+    }
+    function makeBlockEditor(blockId) { //Make the XML editor for a blockId
+        function getXmlFromBlockId(id, dom) { //Utility function to get the block's XML string from a Blockly serialised DOM.
+            for (let i = 0; i < dom.children.length; i++) {
+                const element = dom.children[i];
```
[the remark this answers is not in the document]
ZXMushroom63: Done.

addons/blocklydevtools/userscript.js
ZXMushroom63: Yes, I know, but I've found innerText and textContent to be very hard to work with (in terms of getting indentation to work). However, I think I've mitigated any XSS by replacing any `<` or `>` characters with `&lt;` and `&gt;` in `formatXml()`.

addons/blocklydevtools/userscript.js
ZXMushroom63: Done.

addons/blocklydevtools/userscript.js
ZXMushroom63: I have found replaced some unnecessary usage of innerHTML with innerText when the 'icons' are set for the buttons.

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,419 @@
+export default async function ({ addon, console }) {
+    var isScratchAddons = true; //Should the script use ScratchAddons APIs.
+    if (isScratchAddons) {
+        await addon.tab.waitForElement("svg.blocklySvg>g.blocklyWorkspace");
```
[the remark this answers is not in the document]
ZXMushroom63: In the changes to remove checks from the userscript, I've made it so `getBlockly()` only runs if it needs to.

Joeclinton1:
> Screenshot looks very cool.
> 
> > I don't think making this would be a really good idea unless it's an easter-egg addon possibly.
> 
> Yeah, we could consider this as an easter-egg I think.

I think either include it as a full addon or don't include at all. It's not an easter egg. The addon is clearly dangerous in that a user can break their project's json with it. I say decide whether that is a problem and make a hard decision on it. The addon itself already is NFE so don't let that get in the way of your thinking. My personal opinion is that this addon should've been voted on in addon suggestion issue before development was ever started. I think an addon which lets you hack the blocks without potentially breaking your project should be the starting point. And if this addon is ok, then of course the addon suggestion for hacked blocks which has been permanently sidelined must also be ok.

ZXMushroom63:
> My personal opinion is that this addon should've been voted on in addon suggestion issue before development was ever started.

I made it for a different project, the addon bit came later.

> The addon is clearly dangerous in that a user can break their project's json with it. I say decide whether that is a problem and make a hard decision on it.
My opinion is that with the addon being marked as dangerous, having a big red warning on it and asking users to confirm before enabling is probably enough.
At any rate, while it can corrupt your projects' XML (json is when serialised), and occasionally crash the client entirely, reloading the page should start back from an autosave. It would be a very rare occasion that any actual data loss would occur. (Not saying that it can't happen though)

Joeclinton1:
Anyway you could make it check for errors caused by a broken xml and revert with the error displayed to the user?

ZXMushroom63:
Very probably! Blockly saves all error messages when parsing a workspace dom from a string, which is how the editor's save feature works. I could very easily just check for any errors before loading in the dom.

ZXMushroom63:
Added error catching:
Screenshot 2024-01-23 174431

OneShot-Niko:
> ScratchAddons already allows for creating "hacked" key blocks. Either way, this is still a valid concern, but I think that marking it as "dangerous" probably warns people enough.

Fair.

ZXMushroom63:
Alright, I think its ready for review.

ZXMushroom63:
Btw, how do I get Prettier to not fail?

Samq64:
> Btw, how do I get Prettier to not fail?

You can enable it from the actions tab on your fork.

addons/addons.json
```diff
@@ -1,5 +1,6 @@
 [
   "editor-searchable-dropdowns",
+  "blocklydevtools",
```
Samq64, line 3: I meant between `reorder-custom-inputs` and `NEW ADDONS ABOVE THIS` but it's not a big deal as long as it's not all the way at the bottom.

addons/blocklydevtools/style.css
```diff
@@ -0,0 +1,26 @@
...
+
+/* Hide editor buttons on blocks that aren't top level */
+g.blocklyDraggable[data-id]:not(.blocklyInsertionMarker):not(g.blocklyBlockCanvas>g.blocklyDraggable[data-id]:not(.blocklyInsertionMarker)) [data-is-blocklydev-editor-btn] {
+    display: none !important;
+}
+
+/* Hide the XML editor on blocks that aren't top level */
+g.blocklyDraggable[data-id]:not(.blocklyInsertionMarker):not(g.blocklyBlockCanvas>g.blocklyDraggable[data-id]:not(.blocklyInsertionMarker)) [data-isblocklydeveditor] {
+    display: none !important;
+}
```
Samq64, line 14: Could these be given classes that start with `sa-`? These can probably be combined into 1 rule. Also, have you checked if `!important` is actually necessary?

addons/blocklydevtools/userscript.js
```diff
@@ -0,0 +1,488 @@
...
+        editor.style.color = "yellowgreen";
+        editor.style.border = "2px solid white";
+        editor.style.zIndex = "998";
+        editor.style.borderRadius = "0.8rem";
+        editor.style.width = "max-content";
+        editor.style.backgroundColor = "rgb(0,0,30)";
+        editor.style.overflowX = "hidden";
+        editor.style.maxHeight = "100vh";
+        editor.style.overflowY = "scroll";
+        editor.style.caretColor = "white";
```
Samq64, line 103: Could these and styles for other elements the addon adds also be moved to the CSS file?

addons-l10n/en/forum-id.json
Samq64, line 1: I'm not sure what happened here.

Samq64 (commented):
The buttons aren't added when re-entering the editor. Most editor addons handle this by putting an `addon.tab.waitForElement()` and injecting the elements in a while true loop. Dynamic enable/disable could also be considered but I'm not sure if it makes sense for an addon marked dangerous.

addons/blocklydevtools/style.css
ZXMushroom63: It is, because the buttons use inline styles, which by default are considered more important than seperate style tags

addons-l10n/en/forum-id.json
ZXMushroom63: Either my git or vscode keeps changing the line endings of files, and even though the content isn't modified, it shows hp as a cange.

addons/blocklydevtools/userscript.js
ZXMushroom63: Yep, probably.

addons/blocklydevtools/style.css
ZXMushroom63: I can't rename the blockly internal classes, but I will store the button and editor css as a class soon.

addons/blocklydevtools/userscript.js
ZXMushroom63: Done.

addons/blocklydevtools/style.css
ZXMushroom63: Done merging css rules.

addons/addons.json
ZXMushroom63: Ok, moved it.

ZXMushroom63:
Any updates on getting this merged?

Samq64:
> Any updates on getting this merged?

I'm not very familiar with how Blockly works internally so I we'll need someone who does which can take some time. Either way this is unlikely to make it into the next release since we're behind schedule.

ZXMushroom63:
> > Any updates on getting this merged?
> 
> I'm not very familiar with how Blockly works internally so I we'll need someone who does which can take some time. Either way this is unlikely to make it into the next release since we're behind schedule.

Ok.

Samq64:
Scratch is rewriting `scratch-blocks` to use Blockly as a library instead of forking it (#7768). Could this PR be updated with compatibility for it? The new editor is deployed [here]( https://scratchfoundation.github.io/scratch-editor/spork/scratch-gui/) but it is currently using an older version React than the website so you might have to temporarily revert #8611 after this PR's branch is up to date.

ZXMushroom63:
It should just work, but I will test and update if necessary

ZXMushroom63:
@Samq64 got everything working
Note: the link to the updated branch was down, so I ended up running it locally but everything should work

Issue #6954, opened 2023-12-13
Overflown project action buttons


DNin01:
### Description

This is what the row of project action buttons usually looks like:
![Project action buttons next to project statistics](https://github.com/ScratchAddons/ScratchAddons/assets/106490990/9fd85f5d-608b-4a66-a3c8-45684d694a5a)

But due to space constraints, sometimes the row of buttons for project actions like copying the link, reporting, and adding a project to a studio (among other actions part of addons from Scratch Addons) can overflow into another row, like this:
![Overflown project action buttons row](https://github.com/ScratchAddons/ScratchAddons/assets/106490990/406a0f48-eecc-4c28-8dfd-ffb510e9de87)
This is usually caused by addons that add buttons to this row, and happens when the view counter and the share timestamp overlap, causing the entire row of action buttons, which are all in the same divider element, to wrap. Here's a video where I have revealed the flexboxes as I resize a button until it is wide enough to cause the row to wrap:

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/aaeb454c-83f2-4c86-8e1c-90c9d6fd7a68

However, in rare cases, it can also happen in unmodified Scratch if a project gets so much attention that the love, favorite, and view counters take up enough space to overlap with the share timestamp's usual position, as is the case here:
![Overflown project action buttons row on popular project](https://github.com/ScratchAddons/ScratchAddons/assets/106490990/6c6b3933-1d49-4445-86f2-49dab691104a)

With enough buttons, the row of buttons itself can also span multiple rows:
![Overflown project action buttons row on own project spanning two rows](https://github.com/ScratchAddons/ScratchAddons/assets/106490990/ff68b338-d9cc-4638-abda-25fe57634c3e)
<sup>Download button from #6931</sup>

I was just wondering if there was anything we could do about this? I know, I know, here I am again complaining about things taking up too much space, but I think it's worth at least knowing about this.

Having the buttons placed under where they usually go is not a _big_ deal - you learn to live with it, but there are a couple other reasons I'm creating this issue:
1. Image 4 where the button row itself is crowded and the wrapping button rows are touching each other. This could be fixed by preventing the button row from overflowing and wrapping, even if they take up too much-- let me just show you what I mean:
![Overflown project action buttons row extending past project statistics](https://github.com/ScratchAddons/ScratchAddons/assets/106490990/6b827c02-ed28-4785-916f-642a1d1d1484)
[...]

PR #7129, opened 2024-02-04
New addon: Play Later

DinosuGithub:
### Changes

I created an addon that allows people to add projects to be played later:
Play Later button

These projects are accessed in the bottom section of the home page:
Play Later section

### Reason for changes

Often, Scratchers briefly log onto the Scratch website and find a very cool project that they don't have time to explore at the moment. This addon offers a way to find it again easily when they have more time.

### Tests

No known issues in Chrome.

Note that the extension is intended to remove projects from the Play Later section when they are clicked in the home page section, but not when the user visits them elsewhere. This is so that users do not forget to remove them when they want to, but they do not lose track of the project if they happen to visit it somewhere else.


WorldLanguages:
Pretty cool!
I don't think we ever discussed this idea, so I can't say for sure if this will ever get added into the extension. For example, using a studio or browser bookmarks is an easy workaround. But still, pretty cool feature you developed!

DinosuGithub:
Thanks for the review! With this addon I was focusing on replicating YouTube's "Watch Later" option, about which the same points could be made (since you could bookmark the video or add it to a playlist). I understand if this doesn't match the criteria for new addons though, since I didn't discuss it first.

DNin01:
It's an improvement from browser bookmarks for sure, since this displays project thumbnails.

This is related to the suggestion #1758.

The only problem at first look is that the "Play Later" button could exacerbate the theoretical issue where the project statistics row overlaps with the buttons and causes them to overflow into a different row (#6954).

addons/addons.json
```diff
@@ -1,4 +1,5 @@
 [
+  "play-later",
```
DNin01, line 2: Newly added addons are usually added above `"// NEW ADDONS ABOVE THIS ↑↑",`, not at the top.

DinosuGithub: Oh, I see! I moved it to right above that comment.

DinosuGithub:
> The only problem at first look is that the "Play Later" button could exacerbate the theoretical issue where the project statistics row overlaps with the buttons and causes them to overflow into a different row (#6954).
Ah, is there a better place I should put the button?

DNin01:
> > The only problem at first look is that the "Play Later" button could exacerbate the theoretical issue where the project statistics row overlaps with the buttons and causes them to overflow into a different row (#6954).
> 
> Ah, is there a better place I should put the button?

#6931 put it next to the remix button. We might want to discuss placement options before moving the "Play Later" button from where it is now.

DinosuGithub:
Yeah, that's a possible option. It's funny you would say that because that's where I originally put it, but I moved it down to the lower button section because I was afraid it was cutting off the project title too much.

DNin01:
> Yeah, that's a possible option. It's funny you would say that because that's where I originally put it, but I moved it down to the lower button section because I was afraid it was cutting off the project title too much.

It might be okay if it was up there but didn't have a text label, instead showing a hovernote, which can be done by setting the `title` attribute. However, a button without a label (especially this one) could be a bit ambiguous at first with just an icon. But we'll probably keep discussing.

DNin01:
Furthermore, at the moment, I personally like the "Play Later" button on the bottom where it is now. The download button in that other PR felt like the kind of button that fit well next to the project title. It's different for each button.

DNin01:
Guess we'll have to find some other ways to prevent overflow, now that that button row is getting more crowded than ever (#6954, again)

DinosuGithub:
I agree that it’s most appropriate in the bottom, but if we do ever decide to move it to the top again, the icon can be the clock icon from the bottom.

addons/play-later/homePage.css
Samq64, line 1: Could the styles that don't rely on variables be moved back to this file?

addons-l10n/en/play-later.json
```diff
@@ -0,0 +1,6 @@
+{
+    "play-later/play-later-title": "Play Later",
+    "play-later/add": "Play Later",
```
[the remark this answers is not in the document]
DinosuGithub: I separated them because I thought that they might sometimes translate differently, since one refers to the action of adding a project to be played later, while the other refers to the section title.

addons/play-later/addToListButton.js
```diff
@@ -0,0 +1,52 @@
+import { addToList, getList, removeFromList } from './list.js';
+
+function projectId() {
+    return window.location.pathname.split('projects/')[1].split('/')[0];
```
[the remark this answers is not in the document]
DinosuGithub: Assigned this to a constant 👍

addons/play-later/addToListButton.js
```diff
@@ -0,0 +1,52 @@
...
+            }
+            
+            createButton();
+        });
+
+        addon.tab.appendToSharedSpace({space: 'afterCopyLinkButton', element: button});
+    }
+
+    await createButton();
+    addon.tab.addEventListener('urlChange', () => {
```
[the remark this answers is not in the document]
DinosuGithub: Ah, that makes sense. Is this required though? I ask because this case seems different from the other ones, since the button content and function change each time it is pressed. I can make this change if it fits a standard better, but I don't think it will be less code.

addons/play-later/homePage.css
DinosuGithub: Do you have any lines of JS in mind? From what I could find, the two style lines that don't depend on anything (`slickTrack.style.transform = 'translate3d(0px, 0px, 0px)';` and `slickTrack.style.width = '0px';`) are changed later in the same file. (I can move these into homePage.css if you think that would be better though.)

Samq64: Now that I look at it again there isn't much so the file can just be deleted.

PR #7130, opened 2024-02-04
Hide sprites via context menu addon

### LINKED ISSUE #7077: new addon: show/hide sprites via context menu
### Why this would be helpful

Moving from one sprite to another just to hide something can get annoying on large projects and doing some workaround such as via broadcasts will just take up memory and that's still not really a full solution

### How this addon works

adds an extra button to context menus to show or hide the sprite
either just include both or dynamically change to the opposite of the sprite's current state (shown/hidden)

### Possible alternatives

A new tab in sprites just for hiding/showing with all sprites listed (though this would be probably too big)
no idea for alternatives

One possible setting could be including both show and hide instead of replacing the option like with `block-switching`'s "show option to change block to self".

### Additional context

<sub>I used turbowarp and the scratch editor for so long and this has always been bugging me with the lag of going to another sprite and then switching back
no other context</sub>

mybearworld:
Resolves #7077

### Changes

Adds a new addon that adds a context menu entry that hides/shows a sprite.

A screenshot of the context menu with the option "hide".

### Reason for changes

Convenience, OneShot-Niko accidentally wanting to make duplicates 3 times

### Tests

Tested in Edge.

### DIFF
--- a/addons/hide-show-context-menu/addon.json
+++ b/addons/hide-show-context-menu/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Hide sprites via context menu",
+  "name": "Hide sprites with context menu",
   "description": "Adds an option to a sprite's right click menu to hide or show it.",
   "tags": ["editor"],
   "versionAdded": "1.36.0",
--- a/addons/hide-show-context-menu/addon.json
+++ b/addons/hide-show-context-menu/addon.json
@@ -1,0 +1,20 @@
+{
+  "name": "Hide sprites via context menu",
+  "description": "Adds an option to a sprite's right click menu to hide or show it.",
+  "tags": ["editor"],
+  "versionAdded": "1.36.0",
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "credits": [
+    {
+      "name": "mybearworld",
+      "link": "https://scratch.mit.edu/users/mybearworld"
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true
+}
--- a/addons/hide-show-context-menu/userscript.js
+++ b/addons/hide-show-context-menu/userscript.js
@@ -6,7 +6,7 @@ export default async function ({ addon, msg }) {
   const sharedOptions = {
     types: ["sprite"],
     position: "assetContextMenuAfterDelete",
-    order: 12, // after folders
+    order: 9, // before folders - it's unpredictable how many slots folders takes up
     border: true,
   };
   window.vm = vm;
--- a/addons/hide-show-context-menu/userscript.js
+++ b/addons/hide-show-context-menu/userscript.js
@@ -1,0 +1,28 @@
+export default async function ({ addon, msg }) {
+  const vm = addon.tab.traps.vm;
+  const getSprite = (name) => {
+    return vm.runtime.targets.find((target) => target.sprite.name === name);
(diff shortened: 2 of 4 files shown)


OneShot-Niko:
Could this be enabled by default?

> OneShot-Niko accidentally wanting to make duplicates 3 times

in my defense my memory is absolutely horrible, atleast I managed to remember each time lol

addons/hide-show-context-menu/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Hide sprites via context menu",
```
Samq64, line 2:
```suggestion
  "name": "Hide sprites from context menu",
```

addons/hide-show-context-menu/userscript.js
```diff
@@ -0,0 +1,26 @@
...
+
+  const stateFromBool = (bool) => (bool ? "show" : "hide");
+  const sharedOptions = {
+    types: ["sprite"],
+    position: "assetContextMenuAfterDelete",
+    order: 9, // before folders - it's unpredictable how many slots folders takes up
+    border: true,
+  };
+
+  [true, false].forEach((visibility) => {
```
Samq64, line 13: This works but it's confusing.

addons/hide-show-context-menu/userscript.js
```diff
@@ -0,0 +1,26 @@
...
+    addon.tab.createEditorContextMenu(
+      (ctx) => {
+        getSprite(ctx.itemId).setVisible(visibility);
+      },
+      {
+        ...sharedOptions,
+        className: `sa-hide-show-context-menu-${stateFromBool(visibility)}`,
```
Samq64, line 20: This addon doesn't have a userstyle so I don't think a class is necessary.

Samq64 (approved):

Samq64:
I didn't mean to approve but the addon does work as expected.

addons/hide-show-context-menu/addon.json
mybearworld: This sounds to me like regularly, the sprites would appear in a context menu, and this addon hides them from it.

addons/hide-show-context-menu/userscript.js
mybearworld: The alternative would be code duplicating, wouldn't it?

addons/hide-show-context-menu/userscript.js
mybearworld: It's required by the function according to the docs.

addons/hide-show-context-menu/addon.json
```diff
@@ -0,0 +1,20 @@
...
+    }
+  ],
+  "credits": [
+    {
+      "name": "mybearworld",
+      "link": "https://scratch.mit.edu/users/mybearworld"
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true
```
Samq64, line 19: Dynamic disable isn't handled right now.

addons/hide-show-context-menu/addon.json
Samq64:
```suggestion
  "name": "Hide sprites with context menu",
```

addons/hide-show-context-menu/addon.json
mybearworld: The function seems to handle that too.

PR #7131, opened 2024-02-04
pause: fix keyboard shortcut on macOS

GarboMuffin:
On macOS, option+x types ≈ and shift+option+x types ˛

### DIFF
--- a/addons/pause/userscript.js
+++ b/addons/pause/userscript.js
@@ -19,8 +19,11 @@ export default async function ({ addon, console, msg }) {
   onPauseChanged(setSrc);
 
   document.addEventListener("keydown", function (e) {
-    // On macOS, option+x types ≈ and shift+option+x types ˛ pretty universally across keyboard layouts
-    if (e.altKey && (e.key.toLowerCase() === "x" || e.key === "≈" || e.key === "˛") && !addon.self.disabled) {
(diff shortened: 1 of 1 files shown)


DNin01:
So this bug is a regression caused by #6923 that I didn't even know about until now.

`e.code` will cause the keyboard shortcut to change on other keyboard layouts as explained in the above PR, but is that okay?

GarboMuffin:
I checked like 10 different keyboard layouts and option+x always seems to be ≈ so I can just hardcode that instead

DNin01 (commented):
Looks okay now. Will need someone on macOS to test.

GarboMuffin:
Honestly using the "deprecated" keyCode is probably the most reasonable way to go about this (off topic: it's rather silly for that to be "deprecated" -- it's never going to actually be removed). It's always 88 regardless of qwerty vs dvorak and holding option does not change it.

DNin01:
Maybe we could check if `keyCode` is 88 OR `key` is ≈ or x.

DNin01 (approved):
I don't have a Mac, but this looks good to me.

GarboMuffin:
I have tested this in Firefox, Brave, and Safari on macOS 14.3.1 using the default US keyboard layout and Dvorak

Issue #7116, opened 2024-01-31
Automatically clean up old `tx-...` branches/PRs


DNin01:
Sometimes "Translation update" pull requests are closed without merging, which leaves behind their branches. This could be resolved by automatically deleting those branches.

Stale pull requests could also be closed automatically when a new one is created, and their branches could be deleted at the same time.

@WorldLanguages @Hans5958 Does that sound nice? It could save us time. Or does it not really matter?

<sub>Also, I would call the workflow "_Sync_ strings with Transifex" to be more concise.</sub>

Hans5958:
Ah, yes. I was thinking about it for a while. 

One thing is to have a single branch (e.g. `tx`), so that it is possible to delete the remote branch, closing the PR after it, then recreate the branch and creating the new PR.

Another thing is instead of closing the PR, the PR is adjusted (e.g. the title). Of course when there is no PR present, a new PR will be created.

DNin01:
Related: https://github.com/ScratchAddons/l10n-script/issues/11

Hans5958:
Let's see if that works. If it isn't, I will revert it.

PS. Didn't knew it will also automatically close.

Hans5958:
Some testing done and I can confirm that it works. For now, I'm doing the first way of it.

DNin01:
🚀

PR #7134, opened 2024-02-05
Multiple updates on workflow

### LINKED ISSUE #7133: GitHub Actions: Upgrade to Node 20 soon
Node 16 is no longer being supported and GitHub will deprecate it in Actions soon.

https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20

> Our plan is to transition all actions to run on Node 20 by Spring 2024. We will actively monitor the migration's progress and gather community feedback before finalizing the transition date.

We will want to transition to Node 20 or higher by updating workflows and actions across our repositories soon, or GitHub will transition for us.

@Hans5958, do you want to self-assign?

### LINKED ISSUE #7658: ESLint workflow not working
Error message on every run since... oh, I don't know when:
```
Run npx eslint . --config .eslintrc.json --ext .js
  npx eslint . --config .eslintrc.json --ext .js
  shell: /usr/bin/bash -e {0}
Invalid option '--ext' - perhaps you meant '-c'?
You're using eslint.config.js, some command line flags are no longer available. Please see https://eslint.org/docs/latest/use/command-line-interface for details.
Error: Process completed with exit code 2.
```
The check still succeeds, it just can't evaluate anything. (That's probably why we haven't noticed.)

It might also be a good time to revise the rules, see if we should be more restrictive about certain things like unused variables.

@WorldLanguages @Hans5958

Hans5958:
Resolves #7133 
Resolves #7658  
Relates to #7116

### Changes

Some updates on the workflows, especially on bumping the dependencies. All bumps are considered by looking at each and every changelog.

### Reason for changes

Housekeeping

### Tests

It should've worked, but there may some tests that would need to be done, especially on our in-house scripts.


Hans5958:
@apple502j Would it be possible to run unittest for every push/PR, or are there any considerations?

I was thinking we can turn off the "create issue when failed," and just give a failure when it fails.

DNin01 (commented):
I'm not super familiar with the language of workflows or actions so I can't tell if these upgrades will break anything, but everything else looks good.

- Nice dependency updates.
- Updating Prettier shouldn't be a huge deal for contributors because we added `prettier_version`. For us, just remember to re-run Prettier on the source repo so it's all green.
- Nice new, concise name for the Transifex workflow.

Hans5958:
I had an idea to update the l10n workflow so it beautifies the JSON files like what I did on [website-i18n](https://github.com/ScratchAddons/website-i18n/commit/c0674f4e4701da4da99859a7da79798b633282e7), instead of it being minifed, but then I would increase the extension package quite a bit.

> Note that apple502j is "on inedible hiatus with regard to SA for a while", so he might not be able to respond.

Hmm, fair. Should I just wing it and do that? I saw that it only checks l10n files, maybe we should expand it more...

DNin01:
> I had an idea to update the l10n workflow so it beautifies the JSON files like what I did on [website-i18n](https://github.com/ScratchAddons/website-i18n/commit/c0674f4e4701da4da99859a7da79798b633282e7), instead of it being minifed, but then I would increase the extension package quite a bit.

Well, would this reduce the file size of the diffs? It would make them more readable for sure.

I did this with one of the repositories I store locally, but I also trimmed all the indentation which reduced the file size quite a bit after formatting. Of course, the file itself is not as small as it would be all in one line, and the trick might be more effective on JSON with lots of nested objects, but it's a good compromise for having small, readable diffs.

What I mean is something like this:
```json
{
"a": {
"b": "c"
}
}
```

.github/workflows/transifex.yml
```diff
@@ -1,20 +1,21 @@
-name: Pull and push strings via Transifex
+name: Sync strings via Transifex
```
DNin01, line 1: Nice change. But usually there's something to synchronize data _with_.
```suggestion
name: Sync strings with Transifex
```
I don't know if that really matters.

WorldLanguages:
> I had an idea to update the l10n workflow so it beautifies the JSON files

I would try to avoid changes like this. They're not really necessary.

Samq64:
> > I had an idea to update the l10n workflow so it beautifies the JSON files
> 
> I would try to avoid changes like this. They're not really necessary.

Being minified also discourages newer contributors from modifying them directly.

.github/workflows/transifex.yml
Hans5958: Yeah, I forgot that you said to use "with" after I pushed the change, but I'm still being undecided.

Now you mentioned it, I would change it to it.

Hans5958:
> I would try to avoid changes like this. They're not really necessary.

If your concern is it being too minor, it's more of because "since we're doing a bundle of updates, might as well include more updates."

.github/workflows/format-code.yml
DNin01, line 1: Is it possible to only run Prettier formatting when certain paths are updated? We already ignore generated files like locales other than English, so running formatting when those are changed won't do anything.

I don't know if there's a way to make a _blacklist_ of paths.

Hans5958: Yes, you can, with `paths-ignore`. I have did it, and also a message to sync both paths.

Hans5958:
By the way, this PR is _really_ ready for review. I hope this can be merged in a not so long time.

DNin01:
Can `actions/upload-artifact` also be updated to v4?

Hans5958:
I mean, this MR is way, way, **way** overdue.

Hans5958:
Also updated the ESLint config to be able to be used on ESLint v9.x. Now this PR resolves #7658.

DNin01 (commented):
I'm not well versed in this sort of stuff, but this looks good AFAIK.

DNin01:
We have until June 30th, unless we opt into continuing to use Node 16.

> On June 30th 2024, we will change the default from Node16 to Node20.
> To opt out of this and continue using Node16 while it is still available in the runner, you can choose to set `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true`
> 
> We will then continue to monitor Node16 usage and will communicate a timeline for the removal of Node16 at the start of October, based on the volume of continued use. This means that customers who use the environment variable to continue to use Node16 now have until October to complete their migrations.

https://github.blog/changelog/2024-05-17-updated-dates-for-actions-runner-using-node20-instead-of-node16-by-default/

.github/workflows/eslint.yml
```diff
@@ -1,28 +1,31 @@
...
-      - name: Install ESLint
-        run: |
-          npm install eslint
+        uses: actions/checkout@v4
 
       - name: Run ESLint
-        run: npx eslint .
-          --config .eslintrc.json
-          --ext .js
+        run: npx eslint . --config .eslintrc.json --ext .js
```
DNin01, line 30: Was this supposed to be updated to refer to `eslint.config.mjs`?

DNin01: Also, I don't know if we even have to specify `--config`.

Hans5958:
I forgot to test if the ESLint is working. Turns out I still have adjustments to do. I also have to add the `npm install` bodge, unless I can do `package.json`. This should work for now.

@WorldLanguages, should I add `package.json`?

> We have until June 30th, unless we opt into continuing to use Node 16.

Am I missing something? It's been a while and the workflows have been updated to v20.x. For the note, since June 30 is the past, the workflows that uses it will use Node 20 regardless.

.github/workflows/eslint.yml
Hans5958: `--config` is not really needed, but I just left it *just in case*.

DNin01:
> > We have until June 30th, unless we opt into continuing to use Node 16.
> 
> 
> 
> Am I missing something? It's been a while and the workflows have been updated to v20.x. For the note, since June 30 is the past, the workflows that uses it will use Node 20 regardless.

Oh, it's not June anymore 😄

eslint.config.mjs
```diff
@@ -0,0 +1,147 @@
...
+        Chart: "readonly",
+        Comlink: "readonly",
+        idb: "readonly",
+        Scratch: "readonly",
+        Blockly: "readonly",
+        Clipboard: "readonly",
+        ClipboardItem: "readonly",
+      },
+
+      ecmaVersion: 2020,
```
Samq64, line 43: The oldest browser we now support is Chrome 96 from November 2021.
```suggestion
      ecmaVersion: 2021,
```
The comment near the end of the file should also be updated.

Hans5958: Is there the information for the minimum version of Firefox, just so I can do it in one go?

Samq64: It's much more recent, 109 from January 2023.

DNin01:
Deadline to update `actions/upload-artifact`: November 30th.

DNin01 (approved):
I'm not familiar enough with CI to do a thorough review, but I couldn't find any glaring issues, and this is long overdue, so I think we're good if you agree.

Samq64:
Could ESLint be moved to a separate PR? It does not work without `globals` installed and the easiest proper fix is to introduce a `package.json` which will need some more discussion.

DNin01:
The date has changed:
> Artifact actions v3 will be closing down by **December 5, 2024**. To raise awareness of the upcoming removal, we will temporarily fail jobs using v3 of `actions/upload-artifact` or `actions/download-artifact`. Builds that are scheduled to run during the brownout periods will fail. The brownouts are scheduled for the following dates and times:
> – November 14, 12pm – 1pm EST
> – November 21, 9am – 5pm EST

https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/#artifacts-v3-brownouts

Hans5958:
The ESLint fix is for #7658. Also, I don't expect for people to use ESLint out of the box (esp. when we don't define the version to be used and the latest ESLint won't work AFAIK), so that's why I upgrade it.

DNin01:
@Hans5958 We don't have much longer. Should we just merge?

Hans5958:
I hope I have said this, but `@actions/upload-artifact` is only used on a rarely used workflow action, so even if we lapse it, the risk is low.

DNin01:
> I hope I have said this, but `@actions/upload-artifact` is only used on a rarely used workflow action, so even if we lapse it, the risk is low.

We're getting ready to release v1.40.0, though...?

DNin01:
It'd be nice if we had Dependabot here.

DNin01:
@WorldLanguages, we're cutting it kinda close here if you're planning to release v1.40.0 soon. If we don't upgrade `upload-artifact` by merging this PR before December 5th, it'll stop working and we won't be able to release any more updates after that date.

Hans5958:
> > I hope I have said this, but `@actions/upload-artifact` is only used on a rarely used workflow action, so even if we lapse it, the risk is low.
> 
> We're getting ready to release v1.40.0, though...?

CMIIW, but I thought that is done manually, not here?

DNin01:
> > > I hope I have said this, but `@actions/upload-artifact` is only used on a rarely used workflow action, so even if we lapse it, the risk is low.
> > 
> > 
> > We're getting ready to release v1.40.0, though...?
> 
> CMIIW, but I thought that is done on WL's PC, not here?

I saw [this action](https://github.com/ScratchAddons/ScratchAddons/actions/workflows/packer.yml) which WL runs every release to generate builds to attach to the [releases](https://github.com/ScratchAddons/ScratchAddons/releases).

Hans5958:
Huh, didn't saw it is last run on v1.39.x.

Okay, I would use my discretion and will merge this in two days in the earliest, assuming no significant changes are being requested.

Samq64 (approved):
Untested

DNin01:
> Can't seem to do this since I don't have a power to bypass it, though.

You don't have permission to bypass it? Is that because you aren't on the @ScratchAddons/admins team? But you're an organization owner, so…?

But Samq just approved, so it should be able to be merged now.

DNin01:
> > If I can, I would use my discretion and will merge this by myself. Can't seem to do this since I don't have a power to bypass it, though.
>
> You don't have permission to bypass it? Is that because you aren't on the @ScratchAddons/admins team? But you're an organization owner, so…?

Psst... this might be something you could fix, @WorldLanguages.

<details><summary>Possible solutions</summary>
<p>

1. Check that the "Do not allow bypassing the above settings" option in the branch protection isn't checked.
2. Add Hans to the team.
3. Create a ruleset — which are like branch protections 2.0 — that can be bypassed by organization owners (let me know if you need help with this, I've used rulesets before).

</p>
</details>

Hans5958:
I will merge it after checking the files again on my PC. I'm typing this on my phone.

IIRC The admin team is just a team with additional perms to ping every admin, that is people who has Owner access to the organization. I'm just left the group to avoid any notifications since I'm pretty inactive at that time, but I still have the Owner role, and somehow I don't have access on bypass the check. It's weird. I may check it sometime later, I guess.

Hans5958:
Somehow Prettier workflow fails on master? I will take a look.

DNin01:
ESLint seems to have detected some errors also. They have probably been accumulating from when the scanning wasn't working.

Even then, the check was still a success, somehow. Might need to do some configuration.

WorldLanguages:
@DNin01 We figured out why Hans couldn't merge this as admin - he had the perms to add himself to the 'allowed list', but he wasn't actually in the list.

By the way, I added @mxmou and @Samq64 to that list, so that they can merge pull requests where they are the PR authors and there is 1 additional approval, which we already consider to equal 2 org approvals, even if GitHub doesn't agree.

WorldLanguages:
Unfortunately, the 'Pack extensions' action is still not working as expected.

`No files were found with the provided path: .dist/*.zip. No artifacts will be uploaded.`

I will pack the extension manually this time

Samq64:
> `No files were found with the provided path: .dist/*.zip. No artifacts will be uploaded.`

Most likely this is caused by https://github.com/actions/upload-artifact/issues/602.

DNin01:
> I will pack the extension manually this time

Phew. Why not always do it manually anyway?

PR #7145, opened 2024-02-10
Add 1.36 update notification

WorldLanguages:
Edit: screenshot is outdated

image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -243,14 +243,14 @@
     "message": "Scratch Addons has updated to v$1"
   },
   "extensionUpdateInfo1_v1_36": {
-    "message": "New addon: \"place backpack code at mouse\", which lets you drag-and-drop code blocks to the mouse position instead of to the code area corner. Go to $1 to enable it.",
+    "message": "New addon: \"rearrangeable custom block inputs\", which allows rearranging custom block parameters on the \"Make a block\" screen. Go to $1 to enable it.",
     "description": "The placeholder is \"Scratch Addons settings\"."
   },
   "scratchAddonsSettings": {
     "message": "Scratch Addons settings"
   },
   "extensionUpdateInfo2_v1_36": {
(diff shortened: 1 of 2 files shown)


WorldLanguages:
cc @DNin01 @Samq64 @mxmou

sorry for the hiatus :-) we will be releasing very shortly, and probably once again on the start of March

_locales/en/messages.json
```diff
@@ -242,15 +242,15 @@
   "extensionHasUpdated": {
     "message": "Scratch Addons has updated to v$1"
   },
-  "extensionUpdateInfo1_v1_35": {
-    "message": "New addon: \"skewing in costume editor\", which lets you hold Ctrl while scaling an object on its sides to skew it. Go to $1 to enable it.",
+  "extensionUpdateInfo1_v1_36": {
+    "message": "New addon: \"place backpack code at mouse\", which lets you drag-and-drop code blocks to the mouse position instead of to the code area corner. Go to $1 to enable it.",
```
Samq64, line 246: Coukd one of the other addons go here? This one is already enabled by default and doesn't sound very exciting.

WorldLanguages: Which one would you highlight instead?
Personally I think it's fine to highlight an addon even if we already turned it on automatically.

Samq64: I was thinking rearrangeable custom block inputs but any of them are fine.

Joeclinton1: Was going to say that "place backpack code at mouse" is enabled by default. So the message doesn't make sense.

DNin01: Furthermore, using Place Backpack Code at Mouse creates a very subtle change - the user may not be thinking about it, or even know that there was undesired behavior, nor would they have anything to say, most likely. It's kind of like how Faster Project Loading is enabled by default and no one really notices that anything is happening unless they read the addon's description.

Issue #6929, opened 2023-12-04
Set developer mode based off of install type instead of version name


DNin01:
### Why this would be helpful

1. Sometimes developers end up installing stable releases of the extension, which opens the welcome page even though they are a developer and loaded it the developer way.
2. If a developer updates from a prerelease version to a non-prerelease version, my understanding is that removing the extension will send you to the farewell page, which is also kind of funny.

Yes, users can install the extension from source, but it's intended to be used only for development purposes[^1].
> [Unpacked extensions](https://developer.chrome.com/docs/extensions/mv3/getstarted#unpacked) should only be used to load trusted code during the development process.

Although Firefox doesn't support `version_name`, it should still be returned by `chrome.runtime.getManifest()`, so what we're doing now does work on Firefox (though, I have only tested that with an unrecognized manifest property in Chrome).

[^1]: https://developer.chrome.com/docs/extensions/mv3/hosting/

### My suggestion

Replace [`version_name` checks](https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons+version_name+language%3AJavaScript&type=code&l=JavaScript) with `installType` checks:
```js
let developerMode = false;
chrome.management.getSelf((info) => {
  developerMode = info.installType === "development";
});
```

### Possible alternatives

We could add our own manifest property to the source code that told Scratch Addons whether it was loaded from source and then get it with `chrome.runtime.getManifest()`, but it would create a warning in the Errors tab, albeit one that can be ignored, and the key `browser_specific_settings` also causes that warning to appear anyway. We wouldn't include the key in the packed extensions - only in the source code.

WorldLanguages:
> We could add our own manifest property to the source code that told Scratch Addons whether it was loaded from source and then get it with chrome.runtime.getManifest(), but it would create a warning in the Errors tab

I believe for MV3 we'll have no choice to support both browsers (Chrome and Firefox), so we won't need to add our own manifest property. So we could take advantage of the existing differences between local development and store releases without creating an explicit property for this usecase.

Also, just a comment, in Chrome the extension ID of the Chrome Web Store release is fixed, but it varies during development. Not sure about Firefox, I don't think that's the case there.

WorldLanguages:
> We could add our own manifest property to the source code that told Scratch Addons whether it was loaded from source and then get it with chrome.runtime.getManifest(), but it would create a warning in the Errors tab

[...]

Issue #3834, opened 2021-11-21
1.23.0 ""prerenease""


kccuber-scratch:
### Description

Shows up as 1.23.0-prerenease in manifest
### Repoduction steps

1. download latest code
2. load it into browser
3. see that it says prerenease 

### Expected behaviour
it should say prerelease

### Actual behaviour
it says prerenease

### Screenshots
![image](https://user-images.githubusercontent.com/92891642/142747931-eab02d7b-e5b4-4688-be7f-0883c4e78e2f.png)

### Environment

- Browser:  FF 94
- Scratch Addons version:  1.23.0 ""prerenease""
- Operating system: 

### Additional context

n/a

Mbrick2:
#3836

mxmou:
Fixed by #3837?

CST1229:
it's spelled prereñease not prerenease

kccuber-scratch:
> Fixed by #3837?

Yep.

WorldLanguages:
lol

PR #7154, opened 2024-02-12
Automatically omit `-prerelease` when packing

DNin01:
Related to #6929

Not sure if this is really necessary, but...

### Changes

Modifies the `gen-manifest` script used when packing the extension for releases to automatically omit `-prerelease` from `version_name`, by setting it to `version`.

### Reason for changes

- If this is done automatically, we won't have to bump the version to a non-prerelease before releasing the extension or bump it to a prerelease when starting work on the next version. We would still have to change the version before releasing minor updates, but at least no worrying about whether it's a prerelease or not.
- It benefits developers who install releases through GitHub (if `version_name` is always a prerelease, the extension will always remain in developer mode).
- To avoid [prereñease moments](https://github.com/ScratchAddons/ScratchAddons/issues/3834). 😄

Our workflow before and after:
```diff
 We're on version 1.36.0 stable

 ⬆️ Bump to version 1.37.0 prerelease
 🛠️ Updates, updates, updates...
-✅ Bump to version 1.37.0 stable
 🚀 Release as 1.37.0 stable

 ⬆️ Bump to version 1.38.0 prerelease
 🐞 Oops! A bug or two.
-✅ Bump to version 1.37.1 stable
+✅ Bump to version 1.37.1 prerelease
 🚀 Release as 1.37.1 stable

 ⬆️ Bump to version 1.38.0 prerelease again
 🛠️ Everything else...
-✅ Bump to version 1.38.0 stable
 🚀 Release as 1.38.0 stable
```

### Tests

Not tested. Might want to do a test run after merging.

### DIFF
--- a/.github/gen-manifest.mjs
+++ b/.github/gen-manifest.mjs
@@ -18,6 +18,7 @@ const PERMISSIONS_ALWAYS_IGNORED = [
 export default (env, manifest) => {
   // Deep-clone
   manifest = JSON.parse(JSON.stringify(manifest));
+  manifest.version_name = manifest.version;
   manifest.icons["1024"] = "images/icon.png";
   manifest.icons["32"] = "images/icon-32.png";
   manifest.icons["16"] = "images/icon-16.png";


Hans5958:
I don't think this is needed. I would say it is a good idea to keep records of when something is stable or not, so I'd rather see the "bump" commits stay.

Samq64:
Will this change will also effect GitHub releases?

DNin01:
> Will this change will also effect GitHub releases?

Yes, or that's what I expect it to do.

WorldLanguages:
I think I agree with Hans.

PR #7157, opened 2024-02-12
Check installType instead of version_name in some places

DNin01:
Part of #6929

### Changes

Changes how `loadVueComponent` and `transition.js` check whether you're a developer or not. Instead of checking the `version_name` key in the manifest, they will now check the `ExtensionInfo.installType` property, which is `"development"` if the extension is loaded unpacked. This means if a developer installs Scratch Addons from source, the welcome or farewell webpages will never open, even on releases, and CSS normalization is always performed on Vue components, even on releases.

### Reason for changes

Usually, we refer to the version name, which has a `-prerelease` suffix for work-in-progress versions. If you're on one of these versions, you're assumed to be a developer, and we have features and differences specific to these versions, like a blue accent color and not being sent to a webpage on extension installs or removals.

However, this isn't foolproof, as sometimes you happen to be running a non-prerelease. To solve this, the `management` API can be used to get the reason the extension was installed instead, which tells us if the extension was installed unpacked or not, which would only be used by developers and maybe some power users.

### Tests

Tested `loadVueComponent`.


background/transition.js
```diff
@@ -1,15 +1,17 @@
 const utm = `utm_source=extension&utm_medium=tabscreate&utm_campaign=v${chrome.runtime.getManifest().version}`;
 const uiLanguage = chrome.i18n.getUILanguage();
 const localeSlash = uiLanguage.startsWith("en") ? "" : `${uiLanguage.split("-")[0]}/`;
+const developerMode = await new Promise((resolve) => {
+  chrome.management.getSelf((result) => {
```
TheColaber, line 5: you can use the return value of chrome.managment.getSelf, it's Chrome 88+

libraries/common/load-vue-components.js
```diff
@@ -17,19 +17,22 @@ export default (filenames) =>
...
-              const linesToAdd = normalizedText.split("\n").findIndex((line) => line === cssFirstLine);
-              const newLines = Object.assign(Array(linesToAdd), {
-                0: "/*",
-                [linesToAdd - 1]: "<style> */",
-              }).join("\n");
-              styles[filename] = `${newLines}\n${normalizedCss}/* \n</style> */\n/*# sourceURL=${htmlUrl} */`;
-            } else {
-              styles[filename] = css;
-            }
+            chrome.management.getSelf((result) => {
```
TheColaber, line 20: check this one time at the top of the file. it should never change without the entire extention reloading.

TheColaber (dismissed):

Samq64 (dismissed):
Localhost still works fine with the permission.

mxmou (approved):
Tested both changes on Edge and Firefox.

Samq64 (approved):
I'm not sure how to test a non-dev version. Also stable unpacked versions are still orange which I assume is intentional.

Samq64:
> stable unpacked versions are still orange which I assume is intentional.

Actually #7464 removed the popup icon switching so it's always blue unless `gen-manifest.mjs` is run so maybe the blue should rely on `installType`? But then again there are GitHub releases.

DNin01:
> > stable unpacked versions are still orange which I assume is intentional.
> 
> Actually, #7464 removed the popup icon switching so it's always blue unless `gen-manifest.mjs` is run. Maybe the blue should also rely on `installType`? But then again there are GitHub releases.

Power users might install the extension through GitHub to "use" it, in which case the action icon could be orange.

But I don't think we can change the icon on Chrome's Extensions page, and that icon is blue in the source code, so for consistency, I'd keep the icon blue on all unpacked installations. This would also let users easily differentiate store and source code installations.

PR #7163, opened 2024-02-14
Improvements to "re-orderable custom inputs" addon

### LINKED ISSUE #7151: Improvements to "re-orderable custom inputs" addon
### My suggestion

Have been using the addon from #7020 myself a bit and I think the improvements that were previously mentioned could indeed improve the user experience alot:
1. Inputs added after the selected instead of at the end
2. Disable label merging behaviour

I think these could be options that can be enabled.

The label merging is indeed particularly infuriating, and makes it horrible to use when you have labels. The code that handles the merging is technically [updateDeclarationProcCode_](https://github.com/scratchfoundation/scratch-blocks/blob/f210e042988b91bcdc2abeca7a2d85e178edadb2/blocks_vertical/procedures.js#L565), but the merging occurs because if in the procCode two strings are next to each other, even separated by a space, they'll be merged as one input. So really what you'd need is a delimeter in the proc code which is made invisible everywhere else, but is used to display the function as separate label parts when updateDisplay's [CreateAllInputs_](https://github.com/scratchfoundation/scratch-blocks/blob/f210e042988b91bcdc2abeca7a2d85e178edadb2/blocks_vertical/procedures.js#L205) is called. It could cause problems later on though.

Making the inputs insert after the selected requires replacing [addLabelExternal](https://github.com/scratchfoundation/scratch-blocks/blob/f210e042988b91bcdc2abeca7a2d85e178edadb2/blocks_vertical/procedures.js#L614), [addBooleanExternal](https://github.com/scratchfoundation/scratch-blocks/blob/f210e042988b91bcdc2abeca7a2d85e178edadb2/blocks_vertical/procedures.js#L626) and [addStringNumberExternal](https://github.com/scratchfoundation/scratch-blocks/blob/f210e042988b91bcdc2abeca7a2d85e178edadb2/blocks_vertical/procedures.js#L641)

(I thought I'd make this issue before anyone else complains about it)

Joeclinton1:
Resolves #7151

### Changes

- Replaces the updateDeclarationProcCode and modifiedCreateAllInputs inorder to prevent label merging when shifting inputs, adding inputs, or deleting inputs
- Wraps all three addInput functions with an insertInputAfter function that automatically shifts new inputs to be after currently selected ones for improved UX
- Add an option to disable adding inputs after selected.

### Reason for changes

I have been using the re-orderable custom inputs addon for a while now, and it's been bothering me that labels merge, as it makes it infuriating to use. 
The adding inputs after selected functionality was suggested in the original issue, so this seemed like a good time to implement it.

### Tests

On Chrome I've tested:
- Adding a new input with nothing selected
- Adding a new string/number, boolean and label input at the end
- Adding a new string/number, boolean and label input after a non-end input
- Tested that labels don't merge after:
  - Adding labels after existing labels
  - Adding labels between two existing labels
  - Deleting inputs that separate two labels
- Tested that it correctly refocuses no matter what I do
- Tested the dynamic enable/disable of the addon and the InsertInputsAfter setting
- (New) Tested dynamic enable/disable when the CustomProcedures editor is already open


Joeclinton1:
it has some bugs I've noticed:

 - The %l becomes visible when closing the procedure editor
 - If you add an input without a selected field and close the procedure editor, the bug where "%" is in procCode causes things to break

GarboMuffin:
Not the end of the world, but the label merging fix doesn't work when dynamic enabled with the block editor already open

Joeclinton1:
> Not the end of the world, but the label merging fix doesn't work when dynamic enabled with the block editor already open
> 
> I think it's fine otherwise

The insertAfterInput thing won't work either. It's kind of hard to fix. Currently I replace the functions in their Blocks object, but this only changes the function when the block is created. it doesn't update existing blocks. Maybe I could force a refresh of the procedures editor?

Joeclinton1:
@GarboMuffin Forcing a refresh, or directly modifying the CustomProcedures class seem impossible. So I suggest we just don't support it.

Joeclinton1:
Via directly modifying the customProcedures editor the fix for the dynamic enable/disable was possible. For future reference we got the current instance of the CustomProcedures block with `Blockly.getMainWorkspace().getAllBlocks()[0]`

addons/reorder-custom-inputs/userscript.js
```diff
@@ -78,21 +122,78 @@ export default async function ({ addon, console }) {
...
+    }
+  }
 
   function enableAddon() {
+    // pollute the procedures_declaration prototype with a modified version that prevents merging, and allows inserting after
+    polluteProcedureDeclaration(Blockly.Blocks["procedures_declaration"]);
+
+    // if custom procedures modal is already open we also directly pollute the existing procedures_declaration block
+    if (addon.tab.redux.state.scratchGui.customProcedures.active) {
+      polluteProcedureDeclaration(Blockly.getMainWorkspace().getAllBlocks()[0], false);
```
WorldLanguages, line 165: I've read a message by GarboMuffin on Discord where he mentions that `Blockly.getMainWorkspace()` is used intentionally here, and that it should not be changed to `addon.tab.traps.getWorkspace()` in the future. Could you add this as a code comment?

WorldLanguages: For some context, the behavior of getWorkspace is usually preferred (that's why it exists). Some addons weren't updated yet to use it, and updating this usage of getMainWorkspace() would cause bugs, so I want to avoid future issues.

Joeclinton1: I have added a comment

GarboMuffin:
that was the wrong branch, whoops

GarboMuffin (approved):

addons/reorder-custom-inputs/addon.json
```diff
@@ -20,7 +20,15 @@
...
-  "versionAdded": "1.36.0",
+  "settings": [
+    {
+      "name": "Insert new inputs after selected input",
+      "id": "InsertInputsAfter",
+      "type": "boolean",
+      "default": false
+    }
+  ],
+  "versionAdded": "1.37.0",
```
WorldLanguages, line 31:
```suggestion
  "versionAdded": "1.36.0",
```

WorldLanguages: Solved by my own commit

WorldLanguages:
image

Issue #7147, opened 2024-02-11
New settings aren't usually invisible


DNin01:
_Where's my new setting?_

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/553c1618-8ea7-41f8-924f-7e3207c3c877

Not sure what we can do about this, but it's best if new settings can be seen without having to toggle other settings.

This wouldn't be a problem if dependent settings were grayed out instead of completely hidden.

WorldLanguages:
I've also noticed this, maybe all settings could be visible (but grayed out if necessary) like you said, but I would try to look at other software UIs for inspiration.

mxmou:
For this addon, a select setting might be better than two Boolean settings:
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/cfc74ccb-bb15-4b07-ad74-dc43069548b3)

> I've also noticed this, maybe all settings could be visible (but grayed out if necessary) like you said, but I would try to look at other software UIs for inspiration.

This is what PowerToys does:
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/8ff13a63-c5b5-48ea-9d72-8fce4c64733f)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/39e9439b-3fc7-4e71-b2e2-104638f7590f)

If we choose this solution, the UI should clearly show which settings are related.

DNin01:
What PowerToys does is similar to #5719 and the screenshot in #6352.

WorldLanguages:
In general, I agree that a three-option select is superior to 2 booleans where one depends on the other.

DNin01:
Strings are closed, so unfortunately, we can't really solve this problem in a timely manner. We can and should still turn these two switches into a selection for clarity, but by the time we merge it, many people will have already checked out what's new, or we'll have updated to v1.37.0 where we'll have different new feature highlights.

RedGuy12:
> maybe all settings could be visible (but grayed out if necessary)

I brought this up when I implemented this feature, but I think it was either @mxmou or @lisa-wolfgang who said it would be weird since it's not immediately obvious what other setting it depends on - especially for cross-addon dependencies.

CST1229:
> I brought this up when I implemented this feature, but I think it was either @mxmou or @lisa-wolfgang who said it would be weird since it's not immediately obvious what other setting it depends on - especially for cross-addon dependencies.

[...]

PR #7165, opened 2024-02-15
Convert `fullscreen` toolbar toggles to a select setting

Samq64:
Suggested in this comment: https://github.com/ScratchAddons/ScratchAddons/issues/7147#issuecomment-1939420029

### Changes

Merges the "Hide toolbar in full screen" and "Show toolbar when hovered" settings into one select setting.

### Reason for changes

To keep all 3 toolbar visibility options together and avoid hidden settings.

### Tests

tested on Firefox and Brave.

### DIFF
--- a/addons/fullscreen/addon.json
+++ b/addons/fullscreen/addon.json
@@ -93,7 +93,7 @@
   "latestUpdate": {
     "version": "1.36.0",
     "isMajor": false,
-    "newSettings": ["toolbar"]
+    "newSettings": ["hoverToolbar"]
   },
   "enabledByDefault": false
 }
--- a/addons/fullscreen/addon.json
+++ b/addons/fullscreen/addon.json
@@ -10,7 +10,7 @@
   "info": [
     {
       "type": "notice",
-      "text": "If you choose to hide the toolbar, remember that you can use the Esc key to exit the project player's full screen mode.",
+      "text": "If you choose to always hide the toolbar, remember that you can use the Esc key to exit the project player's full screen mode.",
       "id": "hideToolbarNotice"
     }
   ],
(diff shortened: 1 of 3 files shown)


addons/fullscreen/addon.json
```diff
@@ -10,7 +10,7 @@
   "info": [
     {
       "type": "notice",
-      "text": "If you choose to hide the toolbar, remember that you can use the Esc key to exit the project player's full screen mode.",
+      "text": "If you choose to always hide the toolbar, remember that you can use the Esc key to exit the project player's full screen mode.",
```
DNin01, line 13: The setting controls when to _show_ the toolbar, so this might be a better description:
```suggestion
      "text": "If you choose to never show the toolbar, remember that you can use the Esc key to exit the project player's full screen mode.",
```

DNin01 (approved):
Looks okay to me!

I would rather merge this for v1.36.0, because if we merge it for v1.37.0, we can't really solve the problem of the new setting requiring another setting to be enabled to show up.

mxmou (approved):

DNin01:
@WorldLanguages As long as strings are still open, we could merge this for v1.36.0.

WorldLanguages:
Strings are already closed, so I don't think merging now is a good idea.

Samq64:
I adjusted the migration code for 1.37 but it should still work with pre-1.36 settings.

Samq64:
This also got two approvals just after the strings closed for the last release.

WorldLanguages:
I will double check everything shortly.

background/get-addon-settings.js
```diff
@@ -674,6 +674,20 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
           delete settings.links;
           madeAnyChanges = madeChangesToAddon = true;
         }
+
+        if (addonId === "fullscreen" && settings.hideToolbar !== undefined) {
+          // Transition v1.36 to v1.37
+          if (!settings.hideToolbar) {
+            settings.toolbar = "show";
+          } else if (settings?.hoverToolbar) {
```
WorldLanguages, line 682: Why is optional chaining used here? Isn't `settings` guaranteed to be an object?

Issue #6665, opened 2023-09-19
Research progress-bar compatibility with faster-project-loading 


WorldLanguages:
### Description

When going to an editor URL directly (for example, https://scratch.mit.edu/projects/104/editor) Scratch has a bug where it tries to requests assets from the wrong server, meaning the project loads slower and takes more requests than usual. See #5800 for more info.

The progress-bar addon monitors requests to know how far the project has loaded.

### Steps to reproduce

I don't have the exact steps, but because of feedback we've received, these are worth checking:

- The progress-bar addon should detect that the project has finished loading, if the faster-project-loading addon is disabled and the bug has occurred .
- Same as above, but enabling the faster-project-loading addon to fix the bug should also mean the loading/progress bar should work properly.

From feedback:

> The project loading bar in the editor shows as not finished loading when the project is loaded, and stays until you save the project.

Worth checking: it's possible both addons, when both enabled, trap window.XMLHttpRequest or window.fetch and what addon replaces these functions first could affect whether a bug happens or not.

### Environment

Scratch Addons version: v1.34.x

iqnite:
Detecting when the project has been loaded and then dismissing the progress bar would be good enough, e. g. by simulating the event of a completed save. (Clicking on File > Save now resets the progress bar.)

Tacodiva:
## The Problems

I have finished understanding this interaction. Firstly, a recap on how progress-bar works:

- When a project starts loading, Scratch queues every asset which needs to be loaded up onto a web worker.
- The progress bar traps this enqueuing and uses it to figure out how many assets the project has total.
- It then listens for when the assets actually load and updates the progress bar accordingly.
https://discord.com/family-center
I have looked into this interaction and there's 3 different things that can happen:

### 1. progress-bar without faster-project-loading [Buggy]

Here, Scratch queues all the assets to load, but from the wrong URL. The progress bar initially counts the correct number of assets, but as the bad requests fail Scratch re-enqueues them from the correct URL, but the progress bar sees that as new assets being requested so the number of assets in the project seems to go up once the bar reaches the end. 

### 2. progress-bar when faster-project-loading executes first [Buggy]

[...]

PR #7162, opened 2024-02-14
faster-project-loading: Fix dropping code into sprite-pane

WorldLanguages:
Resolves #7036

### Changes

Never blocks URLs ending in `.json` (as suggested by the bug report issue). Not aware of any better alternatives. Searched on Discord where this issue was brought up with no luck.

### Reason for changes

Fixes a bug we have missed for months.

### Tests

Can now drag backpack code into specific sprites in Chromium.

PR #7169, opened 2024-02-16
Improve progress-bar compatibility with faster-project-loading

Tacodiva:
See https://github.com/ScratchAddons/ScratchAddons/issues/6665#issuecomment-1947595991 for the problems these changes solve and the idea behind fixing them.

Tested manually, fixes the issues with and without faster-project-loading enabled, and works as normal on things unaffected but the bug.

### DIFF
--- a/addons/faster-project-loading/module.js
+++ b/addons/faster-project-loading/module.js
@@ -1,8 +1,5 @@
 export const BACKPACK_URL = "https://backpack.scratch.mit.edu/";
 
-// Inserting sprites from the backpack requests a ZIP archive from backpack.scratch.mit.edu, so we want to allow those
-export const SPRITE_FILE_EXTENSION = ".zip";
-
 /**
  * Returns true if a web worker request is being caused by the Scratch bug
  * https://github.com/scratchfoundation/scratch-gui/issues/8805.
@@ -15,8 +12,6 @@ export function isBadRequest(message) {
   if (typeof message.id !== "string" || typeof message.url !== "string") return false;
 
   if (!message.url.startsWith(BACKPACK_URL)) return false;
-
-  if (message.url.endsWith(SPRITE_FILE_EXTENSION)) return false;
 
   return true;
 }
--- a/addons/faster-project-loading/module.js
+++ b/addons/faster-project-loading/module.js
@@ -1,10 +1,7 @@
-
-
 export const BACKPACK_URL = "https://backpack.scratch.mit.edu/";
 
 // Inserting sprites from the backpack requests a ZIP archive from backpack.scratch.mit.edu, so we want to allow those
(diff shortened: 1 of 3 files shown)


addons/faster-project-loading/userscript.js
```diff
@@ -1,13 +1,12 @@
...
 
+import { BACKPACK_URL, SPRITE_FILE_EXTENSION, isBadRequest } from "./module.js";
+
 export default async function ({ addon }) {
-  const BACKPACK_URL = "https://backpack.scratch.mit.edu/";
-  // Inserting sprites from the backpack requests a ZIP archive from backpack.scratch.mit.edu, so we want to allow those
-  const SPRITE_FILE_EXTENSION = ".zip";
   // Dropping a code item from the backpack into a specific sprite within the sprite-pane (NOT into the code area)
   // requests a JSON file (https://backpack.scratch.mit.edu/{hash}.json) which we shouldn't block
   const CODE_FILE_EXTENSION = ".json";
```
DNin01, line 10: Why is `CODE_FILE_EXTENSION` outside the module?

DNin01: Oh, was `SPRITE_FILE_EXTENSION` supposed to be outside the module too?

DNin01: Sorry, I'm still trying to figure out what's happening here.

Tacodiva: Sorry just changed some things, it's confusing but the bad backpack requests are made twice, once using a webworker and the other using XHR. When trapping all of XHR, we have to deal with all the requests Scratch is making with XHR and thus there's a lot more edge cases. The web worker only deals with assets, thus the ZIP and JSON edge cases are not ever sent to the web worker, so they don't need to be check for.

DNin01: Oh yeah, because the file extension conditions don't need to be accounted for during project loading.

addons/faster-project-loading/module.js
```diff
@@ -0,0 +1,17 @@
+export const BACKPACK_URL = "https://backpack.scratch.mit.edu/";
+
+/**
+ * Returns true if a web worker request is being caused by the Scratch bug
+ * https://github.com/scratchfoundation/scratch-gui/issues/8805.
```
WorldLanguages, line 5:
```suggestion
 * https://github.com/scratchfoundation/scratch-gui/issues/8805
```

Tacodiva:
Upon request of wl I've taken another look at this and it seems good to go. I've checked that #7162 and #6063 have not been regressed. I've tried loading several projects both into the editor and into the project page and the loading bar is now working correctly on the editor and is still working correctly on the project page.

PR #7173, opened 2024-02-16
Remember settings from last recording

### LINKED ISSUE #6438: Make default settings for `mediarecorder` customizable
### Why this would be helpful

Project video recorder currently has a non-customizable set of default settings, and it looks to me like you can't change them, meaning you have to set all of the settings to what you want every time you hit record.

### My suggestion

It would be nice to be able to customize the defaults from the addon's settings. `onion-skinning` and `paint-snap` already do this.

### Possible alternatives

_No response_

### Additional context

_No response_

DNin01:
Probably resolves #6438

### Changes

- The addon now restores the settings you used for your last recording when you reopen the recording options modal, so you don't have to set them again if you decide to redo your recording. Note that clicking "Cancel" in the options modal does not save your settings and only the _last_ configuration is stored in storage.

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/fc1ab39d-a419-41ee-9a82-b6b3ae6c788f

- Made some code readability improvements:
  - Formatting changes for increased clarity
  - Added some comments
  - Used constant variable for recording duration hard limit

### Tests

Tested in Edge 121.


WorldLanguages:
Should we extend it to store the last settings in localStorage?

DNin01:
> Should we extend it to store the last settings in localStorage? (so that the settings survive a reload) Ideally we should store the preference for each project ID if we go this route, but it's probably overkill

Maybe we could store the settings in local storage for the current project, but only for one project ID at a time. If the user started a recording in a different project, the recording options used in the last project they recorded would be discarded.

DNin01:
I will add local storage support soon.

DNin01:
@WorldLanguages The addon now stores recording options in storage.

addons/mediarecorder/userscript.js
```diff
@@ -11,6 +12,39 @@ export default async ({ addon, console, msg }) => {
...
+    try {
+      return JSON.parse(localStorage.getItem("sa-record-options"));
+    } catch {
+      return {};
+    }
+  };
+
+  const getProjectId = () => {
+    const id = parseInt(window.location.pathname.split("/")[2]);
+    return id ? id : 0;
```
WorldLanguages, line 27:
```suggestion
    return id ?? 0;
```

WorldLanguages:
> Maybe we could store the settings in local storage for the current project, but only for one project ID at a time. If the user started a recording in a different project, the recording options used in the last project they recorded would be discarded.

I'm not sure if this makes sense. Are our defaults better than the user's last settings, even if its a different project?

I was actually referring to storing the last settings from multiple projects at the same time in storage - but that also comes with its challenges, because for example we don't want the storage to increase in size forever as the user records more and more projects.

addons/mediarecorder/userscript.js
DNin01: `parseInt` returns `NaN` when parsing nothing or a string without a number. If I'm correct, using a nullish coalescing operator (??)[^1] here would always return `id` as long as it's not nullish, and NaN is **not nullish**, whereas the conditional operator (? ... : ...) [^2] that we are currently using returns `id` as long as it's _truthy_, and NaN is, of course, falsy.

Is there another way this could be done?

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
[^2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator

mxmou: `id || 0`?

WorldLanguages: I should have suggested `||` instead to keep the same behavior, nice catch.

DNin01:
> > Maybe we could store the settings in local storage for the current project, but only for one project ID at a time. If the user started a recording in a different project, the recording options used in the last project they recorded would be discarded.
> 
> I'm not sure if this makes sense. Are our defaults better than the user's last settings, even if its a different project?

Last settings are probably better in any case. I just changed it.

DNin01:
If we can merge this soon, I can start working on #7460.

hueychen27:
Will this be added to the `addons.json` file so that we can keep these settings in the file too?

mxmou:
> Will this be added to the `addons.json` file so that we can keep these settings in the file too?

I don't think that's possible - these are saved in Scratch's local storage which the settings page doesn't have access to.

addons/mediarecorder/userscript.js
```diff
@@ -201,6 +233,7 @@ export default async ({ addon, console, msg }) => {
 
       return optionPromise;
     };
+
```
Samq64, line 236:
```suggestion
```

Samq64 (approved):

DNin01:
@WorldLanguages I changed it so that the settings used in the last recording are always the default **even across projects** (a.k.a. global). What do you think now?

WorldLanguages:
I think it's better than forgetting everything, sure

TheColaber (approved):
My commit added constants at the top of the file, prevented the addon from erroring if the JSON.parse somehow fails, made the naming of values consistent across the file, and adds the localstorage entry after the logic for accessing the user's mic is ran.

WorldLanguages:
DNin01 you can merge this if you approve the changes.

DNin01 (commented):
Haven't tried yet but looks good. 👍

Issue #7175, opened 2024-02-17
New setting in "Customizable Nav bar", "Expand the Width"


Waakul:
### Why this would be helpful

it adds more space for the search bar

### My suggestion

New setting in "Customizable Nav bar", "Expand the Width" that increases the width of scratch's nav bar breaking its width limit
Coming soon!

Waakul:
I made a pull request adding this feature pls check @WorldLanguages @pufferfish101007

OneShot-Niko:
> I made a pull request adding this feature pls check (ping)

patience

pufferfish101007:
does this not already exist (`expanding-search-bar`)?

CST1229:
> does this not already exist (`expanding-search-bar`)?

I think they mean making the actual navigation bar content extend to the edges of the screen.

DNin01:
I think it might be interesting to also be able to customize exactly how wide the content area on the navigation bar is.

PR #7176, opened 2024-02-17
New setting in "Customizable navgation bar" addon, "Expand the width"

Waakul:
https://github.com/ScratchAddons/ScratchAddons/issues/7175

Resolves # Adds more space for the search bar by shattering scratch's width limit (for the nav bar)

### Changes

Added a new setting in the "Customizable Navigation bar" addon, whose id is "discuss-button", named "Expand the width" when enabled, shatters scratch width limit (in the nav bar) and gives more space for the search bar

### Reason for changes

Increases width of the search bar thus incrementing the truncate limit and makes text in the search bar easier to see.

### Tests

Tested on: MS Edge

### DIFF
--- a/addons/discuss-button/addon.json
+++ b/addons/discuss-button/addon.json
@@ -29,6 +29,15 @@
       }
     },
     {
+      "url": "expand-width.css",
+      "matches": ["https://scratch.mit.edu/*"],
+      "if": {
+        "settings": {
+          "expand-width": true
+        }
+      }
+    },
+    {
       "url": "stick.css",
       "matches": ["https://scratch.mit.edu/*"],
       "if": {
@@ -97,6 +106,12 @@
       "default": false
     },
     {
+      "name": "Expand the width",
+      "id": "expand-width",
+      "type": "boolean",
+      "default": false
+    },
+    {
       "name": "Stick to",
       "id": "stick",
       "type": "select",
@@ -117,8 +132,8 @@
   "dynamicDisable": true,
   "versionAdded": "1.0.0",
   "latestUpdate": {
-    "version": "1.32.0",
-    "newSettings": ["stick"],
+    "version": "1.36.0",
+    "newSettings": ["expand-width"],
     "isMajor": false
   },
   "tags": ["community", "featured"],
--- a/addons/discuss-button/expand-width.css
+++ b/addons/discuss-button/expand-width.css
@@ -1,0 +1,3 @@
+#navigation > .inner{
+	width: 97%;
+}


addons/discuss-button/expand-width.css
```diff
@@ -0,0 +1,3 @@
+#navigation > .inner{
```
Samq64, line 1: This doesn't work on older (`scratchr2`) pages which need different selectors.
```suggestion
#navigation > .inner,
#topnav .container {
```

Samq64:
I don't think the search bar should get any wider than the width of the normal navigation bar or the page content. It's just gets wide on higher resolution screens.

CST1229:
> I don't think the search bar should get any wider than the width of the normal navigation bar or the page content. It just gets too wide on higher resolution screens. Ideally it would also stay centered with the page.

The addon does this:
image

Samq64:
> > I don't think the search bar should get any wider than the width of the normal navigation bar or the page content. It just gets too wide on higher resolution screens. Ideally it would also stay centered with the page.
> 
> The addon does this:
> (image)

What I meant is that the search bar can get too wide:
image

It's doesn't look that bad on `scratch-www` though.

Waakul:
image
It might be a bit too big on scratch-2.0 pages, but some people like me find this addon useful

Waakul:
Also this addon is essential on smaller devices to stop scratch from removing the search bar

Samq64 (approved):
I couldn't think of a way to easily shrink the search bar without causing problems. Other than that this look good.

WorldLanguages:
Unfortunately, I had no idea what "expand the width" meant until I saw some screenshots, so we would need to think of a better name.

CST1229:
> Unfortunately, I had no idea what "expand the width" meant until I saw some screenshots, so we would need to think of a better name.

"Expand content to screen edges"?

Samq64:
> > Unfortunately, I had no idea what "expand the width" meant until I saw some screenshots, so we would need to think of a better name.
> 
> "Expand content to screen edges"?

Maybe window instead of screen?

Waakul:
Expand Contents To Edges

DNin01:
I have formatted the code for you to match our preferred code style. If you want formatting to be done automatically, you can [enable workflows on your fork](https://github.com/Waakul/ScratchAddons/actions).

addons/discuss-button/addon.json
```diff
@@ -96,6 +105,12 @@
       "type": "boolean",
       "default": false
     },
+    {
+      "name": "Expand contents to the edges",
```
DNin01, line 109: I have some name ideas:
```suggestion
      "name": "Span the entire page's width",
```
```suggestion
      "name": "Extend to page edges",
```

DNin01:
I think it might be interesting to also be able to customize exactly how wide the content area on the navigation bar is.

OneShot-Niko (commented):
The title is probably gonna have a lot of renames

addons/discuss-button/addon.json
Samq64: I think the second suggestion here is the best so far.

Waakul:
Wait, maybe make it customizable

Waakul:
anyways i feel like this feature is lame so I'm closing

Samq64:
> ahh i accidentally commited the new pull request to this fork!!

That's why each PR should be on a different branch on your fork.

Assuming you're using GitHub desktop: https://github.com/ScratchAddons/ScratchAddons/pull/7176#issuecomment-1961058733

PR #7183, opened 2024-02-18
Add beta tag to deprecated/broken forum addons

WorldLanguages:
Not the ideal solution, but works for now


DNin01:
In the meantime, another quick patch we could make is adding an error message (to the userscript) when the user tries to use the feature, explaining that it no longer works. It's kind of rude, but better than the user submitting their post only to realize their image can't be used in it. Obviously, that would be for v1.36.1 or v1.37.0 because the translation team won't have time to translate for v1.36.0.

We could also add a warning note to the addon.

PR #7186, opened 2024-02-18
Add deprecation warning to image-uploader

WorldLanguages:
### Changes

- The userscript is still in the source, but I removed references to it from addon.json
- Adds an (untranslated in the release) warning to the addon stating that it no longer works

cc @DNin01

### DIFF
--- a/addons/image-uploader/addon.json
+++ b/addons/image-uploader/addon.json
@@ -6,10 +6,11 @@
       "name": "Jeffalo"
     }
   ],
-  "userscripts": [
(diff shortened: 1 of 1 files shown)


DNin01 (approved):
Having a warning message only in English is probably better than not having a warning message at all. I assume this kind of thing has happened before.

DNin01:
But you could translate it to Spanish before anyone else! 😄

Samq64 (approved):

PR #7187, opened 2024-02-18
custom-menu-bar: rewrite compact-dropdown userscript

Samq64:
See the post-merge reviews on #6297.

### Changes

Rewrites `custom-menu-bar`'s userscript to use the Redux as part of the `addon.tab.waitForElemet`.

### Reason for changes

To stop using a hashed selector and using Redux and `waitForElement` separately is unnecessary.

### Tests

Tested on Firefox

### DIFF
--- a/addons/custom-menu-bar/username.js
+++ b/addons/custom-menu-bar/username.js
@@ -1,6 +1,6 @@
 export default async function ({ addon }) {
   while (true) {
-    const menuItem = await addon.tab.waitForElement('[class^=account-nav_user-info_] + div > ul > :first-child', {
+    const menuItem = await addon.tab.waitForElement("[class^=account-nav_user-info_] + div > ul > :first-child", {
       markAsSeen: true,
       reduxEvents: ["scratch-gui/menus/OPEN_MENU"],
       reduxCondition: (state) => state.scratchGui.menus.accountMenu,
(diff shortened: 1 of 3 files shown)


addons/custom-menu-bar/username.js
```diff
@@ -1,14 +1,13 @@
...
-    addon.tab.redux.initialize(); // Start listening to Redux events
-    addon.tab.redux.addEventListener("statechanged", async (e) => {
-      if (!addon.settings.get("compact-username")) return; // Skip if the option is disabled
-      if (e.detail.action.type !== "scratch-gui/menus/OPEN_MENU" || e.detail.action.menu !== "accountMenu") return; // Skip if another Redux event occurs
-      const profileSpans = await addon.tab.waitForElement(".menu_menu-item_3EwYA.menu_hoverable_3u9dt"); // Do this if the menu is opened
-      const profile = profileSpans.appendChild(document.createElement("span")); // Add username to "Profile" menu option
-      profile.textContent = username;
-      profile.className = "sa-profile-name";
+  while (true) {
+    const menuItem = await addon.tab.waitForElement("[class^=menu_menu_] > :first-child", {
```
mxmou, line 3:
```suggestion
    const menuItem = await addon.tab.waitForElement("[class^=menu-bar_account-info-group_] [class^=menu_menu_] > :first-child", {
```
`reduxCondition` is only meant to be used as an optimization - the selector should still work without the condition. Currently, this happens if the condition is removed:
image

WorldLanguages: Indeed, it shouldn't match more elements than needed if reduxCondition theoretically always evaluates to true.

Samq64:
Just a minute the username styling broke when the compact user setting is disabled.

mxmou (approved):

PR #5331, opened 2022-11-06
New addon: Square block text and color inputs

CST1229:
Resolves #3698

### Changes

Adds the "square block text and color inputs" addon, which makes text and/or color inputs rectangular, with additional settings to affect only text or color inputs.

### Reason for changes

This is another addon that makes the editor more like Scratch 2.0, but it can also be useful for differentiating text and number inputs (since in 3.0, they look identical but number inputs only accept numbers, `.`, `-` and `e`)

### Tests

Tested on Chrome and Firefox, with dynamic enable/disable, options enable/disable and seeing outside then inside.

PR #7205, opened 2024-02-22
editor-square-inputs: add number option

CST1229:
Resolves??? [#Square block text inputs doesn't apply to number inputs](https://discord.com/channels/806602307750985799/1209864567430062131/1209864567430062131)
Also addresses [feedback](https://github.com/ScratchAddons/ScratchAddons/pull/5331#issuecomment-1960402896)

### Changes

Adds a (disabled-by-default) option to the Square block text inputs addon to affect number inputs, and also adjusts the description. I'm not sure if the description change is needed, as it might seem a bit contradictory with the number option existing.

The number option also affects all the other inputs that only accept numbers (like direction and note inputs), meaning you can enable both the text and number options to make all typable inputs square.

### Reason for changes

Some people got confused about how the addon doesn't affect number inputs.

### Tests

Only tested a bit on Firefox,  but it's a fairly small thing.
image
image
image

### DIFF
--- a/addons/editor-square-inputs/addon.json
+++ b/addons/editor-square-inputs/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Square block text inputs",
-  "description": "Makes text and color inputs on blocks rectangular instead of round, like in Scratch 2.0.",
+  "description": "Makes text (not number) and color inputs on blocks rectangular instead of round, like in Scratch 2.0.",
   "credits": [
     {
       "name": "CST1229",
@@ -22,6 +22,12 @@
       "default": true
     },
     {
+      "name": "Number inputs",
+      "id": "number",
+      "type": "boolean",
+      "default": false
+    },
+    {
       "name": "Color inputs",
       "id": "color",
       "type": "boolean",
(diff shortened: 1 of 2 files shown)


addons/editor-square-inputs/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Square block text inputs",
-  "description": "Makes text and color inputs on blocks rectangular instead of round, like in Scratch 2.0.",
+  "description": "Makes text (not number) and color inputs on blocks rectangular instead of round, like in Scratch 2.0.",
```
Samq64, line 3:
```suggestion
  "description": "Makes certain block input types such as text or color rectangular instead of round, like in Scratch 2.0.",
```

addons/editor-square-inputs/addon.json
```diff
@@ -21,6 +21,12 @@
       "type": "boolean",
       "default": true
     },
+    {
+      "name": "Number inputs",
+      "id": "number",
+      "type": "boolean",
+      "default": false
```
Samq64, line 28: Tiny nitpick but the default is a disabled setting in between 2 enabled ones. Could it be moved down so the enabled ones go first?

addons/editor-square-inputs/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Square block text inputs",
```
mxmou, line 2:
```suggestion
  "name": "Square block inputs",
```

mxmou (approved):

Samq64 (approved):

WorldLanguages:
Nice work!

PR #7216, opened 2024-02-23
`hide-flyout`: Remove code that shrinks Scratch coding workspace

Hexagonl:
### Changes

There was some code in the Auto-hiding block palette (`hide-flyout`) addon that explicitly shrinks the width of the code workspace by 250 pixels if the toggle setting is "Category hover" or "Category click" (unless two-columns category menu addon is enabled). This causes problems on devices with smaller screens where the code palette blocks the user from easily accessing blocks underneath it by panning. This PR removes that code.

Video of **CURRENT BEHAVIOR** with an example work environment that triggers the problem below:

https://github.com/ScratchAddons/ScratchAddons/assets/61996651/4fc702a9-a79f-491a-a424-f08c755f6519

With the changes, the user can now actually pan to access the blocks underneath the block palette.

Video of **CHANGED BEHAVIOR** below:

https://github.com/ScratchAddons/ScratchAddons/assets/61996651/27128303-3c84-443b-9935-b82a32241b96

### Reason for changes

It is frustrating when you try to pan to a block input hidden underneath the block palette only to realize the workspace is fenced, so you then need to fidget around with zooming in and out... hiding the block palette... resizing the window... hiding the stage... considering buying a larger monitor... lmao not ideal

### Tests

- Tested on Brave and Firefox using a 1920x1080 display at 150% Windows display scale
- Tested both with and without Two-column category menu (`columns`) addon

I accidentally left my microphone on in the videos so enjoy hearing my mouse slam on my desk.

### DIFF
--- a/addons/hide-flyout/userscript.js
+++ b/addons/hide-flyout/userscript.js
@@ -226,24 +226,6 @@ export default async function ({ addon, console, msg }) {
       }
       oldStepScrollAnimation.call(this);
     };
-
-    // add flyout size to the workspace dimensions
-    const oldGetMetrics = Blockly.WorkspaceSvg.getTopLevelWorkspaceMetrics_;
-    Blockly.WorkspaceSvg.getTopLevelWorkspaceMetrics_ = function () {
-      const metrics = oldGetMetrics.call(this);
-      if (addon.self.disabled || getToggleSetting() === "hover" || this.RTL) return metrics;
-      if (this.getToolbox()?.flyout_?.getWidth() === 310) {
-        // columns is enabled
-        return metrics;
-      }
-      return {
-        ...metrics,
-        absoluteLeft: metrics.absoluteLeft - 250,
-        viewWidth: metrics.viewWidth + 250,
-      };
-    };
-    if (Blockly.getMainWorkspace())
-      Blockly.getMainWorkspace().getMetrics = Blockly.WorkspaceSvg.getTopLevelWorkspaceMetrics_;
   }
 
   while (true) {
--- a/addons/hide-flyout/userscript.js
+++ b/addons/hide-flyout/userscript.js
@@ -297,9 +297,6 @@ export default async function ({ addon, console, msg }) {
 
     doOneTimeSetup();
     autoLock();
-    if (getToggleSetting() !== "hover") {
-      // update workspace dimensions
-      Blockly.svgResize(Blockly.getMainWorkspace());
-    }
+    Blockly.svgResize(Blockly.getMainWorkspace());
   }
 }
--- a/addons/hide-flyout/userscript.js
+++ b/addons/hide-flyout/userscript.js
@@ -143,8 +143,6 @@ export default async function ({ addon, console, msg }) {
     }
     addon.self.addEventListener("disabled", () => {
       Blockly.getMainWorkspace().getToolbox().selectedItem_.setSelected(true);
-      // update workspace dimensions
-      Blockly.svgResize(Blockly.getMainWorkspace());
     });
     addon.self.addEventListener("reenabled", () => {
       if (getToggleSetting() === "category" && !addon.settings.get("lockLoad")) {
@@ -152,8 +150,6 @@ export default async function ({ addon, console, msg }) {
         onmouseleave(null, 0);
         toggle = false;
       }
-      // update workspace dimensions
-      Blockly.svgResize(Blockly.getMainWorkspace());
     });
 
     addon.settings.addEventListener("change", () => {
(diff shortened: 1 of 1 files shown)


Samq64:
I think the reason for shrinking the workspace is so blocks don't accidentally get deleted when they're dragging while still overlapped with the palette.

Hexagonl:
> I think the reason for shrinking the workspace is so blocks don't accidentally get deleted when they're dragged but still overlap the palette.

I'm not able to reproduce this, unless I interpreted your words wrong. Everything seems to work correctly for me. Is there a PR or issue I can't find related to the thing you're talking about, so that it can be compared with my PR?

Hexagonl:
> I think the only time the view area should be shrunken is when you collapse the block palette in category click mode, because then it stays out of your way.

I typically use category click mode, but the shrunken view area is still a bother for me (as I leave the palette open by default). I don't see why not remove the shrinking for all options, as the more space the merrier lol

Samq64:
> > I think the reason for shrinking the workspace is so blocks don't accidentally get deleted when they're dragged but still overlap the palette.
> 
> I'm not able to reproduce this, unless I interpreted your words wrong. Everything seems to work correctly for me. Is there a PR or issue I can't find related to the thing you're talking about, so that it can be compared with my PR?

Never mind. It was an assumption. The code was added in it's own commit on #4506 so there's obviously a reason for it. @mxmou?

DNin01:
> > > I think the reason for shrinking the workspace is so blocks don't accidentally get deleted when they're dragged but still overlap the palette.
> > 
> > 
> > I'm not able to reproduce this, unless I interpreted your words wrong. Everything seems to work correctly for me. Is there a PR or issue I can't find related to the thing you're talking about, so that it can be compared with my PR?
> 
> Never mind. It was an assumption. The code was added in it's own commit on #4506 so there's obviously a reason for it. @mxmou?

Well, I think the area is shrunken when using `hide-flyout` because normally, in vanilla Scratch, the block palette persistently covers up part of the left side of the view, so extra empty space that you can scroll onto is added to the left edge of the code area to compensate for the overlaid block palette. When `hide-flyout` is active, the block palette can be hidden, therefore we don't need this extra padding, assuming the palette is hidden when you're trying to interact with something under its usual area.

code-block-palette-areas-diagram

However, this doesn't account for when the palette is locked or set to show when using category click to toggle.

Waakul:
Nice idea!

mxmou:
> The code was added in it's own commit on #4506 so there's obviously a reason for it. @mxmou?

The main reason was to extend the horizontal scrollbar to the left edge. The same metrics also affect how far it's possible to scroll, but I assumed that wouldn't be a problem because the palette would be closed most of the time. Now that there's a setting to open the palette by default, I think it makes sense to remove this code.

mxmou (changes requested):
The code added by #4871 should also be removed - its only purpose is to update the workspace dimensions.

Waakul:
no one would put the code there without a reason

mxmou:
> no one would put the code there without a reason

Not necessarily with a good reason.

Waakul:
> > no one would put the code there without a reason
> 
> Not necessarily with a good reason.

k

Waakul (approved):
Pretty useful ig

Hexagonl:
I also removed some similar looking code added in #4506 at [this pull](https://github.com/ScratchAddons/ScratchAddons/pull/4506/commits/bb75b888fde5e2964f855d5df0c2be1cd439ce6d). I couldn't think of another use for it, since the `columns` addon and RTL work fine without it.

DNin01 (commented):
Now when I open the editor with `hide-flyout` enabled, the workspace is extended to the right too far and the vertical scroll bar becomes hidden.

image

This happens on all setting combinations, but does not seem to happen on dynamic enable, and it goes back to normal after switching tabs in the editor or resizing the window.

mxmou (approved):

Hexagonl:
> Now when I open the editor with `hide-flyout` enabled, the workspace is extended to the right too far and the vertical scroll bar becomes hidden.

I've been able to reproduce it a couple times myself; let me try removing the "possible unnecessary" commit

mxmou:
> Now when I open the editor with `hide-flyout` enabled, the workspace is extended to the right too far and the vertical scroll bar becomes hidden.

Interestingly, this also happens on the stable version if the mode is palette area hover. Re-adding the "possibly unnecessary" code, but without the `if (getToggleSetting() !== "hover")` condition, fixes it.

DNin01 (approved):
OK now!

The only other UX thing I'd consider looking into is shrinking the code area view _while the flyout is hidden_ if the mode is set to "category click", as it was before on all modes except for "palette area hover" before these changes, but it's alright either way.

WorldLanguages:
Is this ready to merge?

Waakul:
> Is this ready to merge?

Probably

PR #7223, opened 2024-02-24
search-profile: consider atypical short and long usernames

### LINKED ISSUE #7222: Add support for unusual usernames
### Why this would be helpful

Some usernames are greater than 20 characters or shorter than 3, even though that's invalid to register. The username search addon filters these out even though they exist.

### My suggestion

Add an exceptions list for existing 20+ char or less than 3 char usernames.

### Possible alternatives

Check every username entered, even if it's not valid. This would mean every search on Scratch would have to get checked for validity though.

### Additional context

_No response_

Oren-Lindsey:
Resolves #7222

### Changes
Changes the regex to accept 2 character usernames as well as up to 30 character usernames (previously 20).

### Reason for changes
So that the unusual/"invalid" usernames that exist can be searched.

### Tests
Tested as working in Firefox.


Waakul (approved):
Minor change but meaningful, approved! :)

addons/search-profile/userscript.js
```diff
@@ -6,6 +6,126 @@ export default async function ({ addon, console, msg }) {
...
+    "scratchteacheralpha073",
+    "scratchteacheralpha074",
+    "scratchteacheralpha075",
+    "scratchteacheralpha076",
+    "scratchteacheralpha077",
+    "Spongebobrectanglepants",
+    "gdpr533f604550b2f20900645890",
+    "gdpr6d746b4e0e92767d4d76ca39",
+    "gdpr86fe38473d91f98d8cf9d247",
+    "gdpr8e0f8747d5b59e360814647f",
```
Samq64, line 127: These probably shouldn't be here, they were renamed for a reason.

addons/search-profile/userscript.js
```diff
@@ -6,6 +6,126 @@ export default async function ({ addon, console, msg }) {
...
+    "tg",
+    "tu",
+    "tv",
+    "vh",
+    "ya",
+    "yt",
+    "gasstationwithoutpumps",
+    "programmingwithpurpose",
+    "scratchcoder123forlife",
+    "scratchteacheralpha001",
```
Samq64, line 48: Could all the `scratchteacheralpha`s be a regex?

addons/search-profile/userscript.js
Waakul: > These probably shouldn't be here, they were renamed for a reason.

even though they are deleted accounts, it's visible so it'd be ok to have it their

Oren-Lindsey: Yeah, we're not revealing their identities or anything. And just because there's nothing on their profiles doesn't mean we shouldn't include them. We're here to make the extension and let people use it for whatever they want, not to gatekeep what they can use it for (if that makes sense).

addons/search-profile/userscript.js
Samq64: It doesn't have to be Regex just something to reduce then to a few lines.

addons/search-profile/userscript.js
```diff
@@ -6,6 +6,53 @@ export default async function ({ addon, console, msg }) {
...
+  'scratchteacheralpha058', 'scratchteacheralpha059',
+  'scratchteacheralpha060', 'scratchteacheralpha061',
+  'scratchteacheralpha063', 'scratchteacheralpha064',
+  'scratchteacheralpha065', 'scratchteacheralpha066',
+  'scratchteacheralpha067', 'scratchteacheralpha068',
+  'scratchteacheralpha069', 'scratchteacheralpha070',
+  'scratchteacheralpha071', 'scratchteacheralpha072',
+  'scratchteacheralpha073', 'scratchteacheralpha074',
+  'scratchteacheralpha075', 'scratchteacheralpha076',
+  'scratchteacheralpha077', 'Spongebobrectanglepants',
```
OneShot-Niko: Wouldn't doing `'scratchteacheralpha*',` be far more optimal here?

OneShot-Niko (changes requested):

mxmou:
I think changing `{3,20}` in the regex to, for example, `{2,30}` would be a better solution than hardcoding every known exception.

WorldLanguages:
> I think changing `{3,20}` in the regex to, for example, `{2,30}` would be a better solution than hardcoding every known exception.

Yeah

Samq64:
> I think changing {3,20} in the regex to, for example, {2,30} would be a better solution than hardcoding every known exception.

The longest one in the list is 28 so it makes sense to set that as the max.

PR #7226, opened 2024-02-24
Prevent Alt+X (pause) from cutting blocks

### LINKED ISSUE #7225: Alt+X pause shortcut may delete blocks
### Description

Related: #6804

+1 from feedback

### Steps to reproduce

See comment https://github.com/ScratchAddons/ScratchAddons/issues/6804#issuecomment-1779891960 and below.

_No response_

### Environment

Browser: any
Scratch Addons version: any since Alt+X was introduced as a shortcut

DNin01:
Resolves #7225

### Changes

When the `pause` addon is enabled, pressing <kbd>Alt</kbd>+<kbd>X</kbd> will now stop immediate propagation, to avoid activating other shortcuts (specifically, cutting blocks).

### Reason for changes

The Scratch editor's Blockly workspace cuts the focused block (usually the last one that was touched) when pressing <kbd>Ctrl</kbd>+<kbd>X</kbd>, <kbd>Cmd</kbd>+<kbd>X</kbd> (Mac), <kbd>Alt</kbd>+<kbd>X</kbd>, or <kbd>Option</kbd>+<kbd>X</kbd> (Mac), two of which conflict with our pause keyboard shortcut.

I wonder why part of this shortcut was bound to the <kbd>Alt</kbd> key in the first place.

### Tests

Tested with Edge 121.

Should probably check that this doesn't cause any other problems, like preventing other keyboard shortcuts from working.

### DIFF
--- a/addons/pause/userscript.js
+++ b/addons/pause/userscript.js
@@ -25,6 +25,7 @@ export default async function ({ addon, console, msg }) {
     // Because keyCode is deprecated we'll still check e.key in case keyCode is not as reliable as we think it is
     if (e.altKey && (e.key.toLowerCase() === "x" || e.keyCode === 88) && !addon.self.disabled) {
       e.preventDefault();
+      e.stopImmediatePropagation();
       setPaused(!isPaused());
     }
   });
--- a/addons/pause/userscript.js
+++ b/addons/pause/userscript.js
@@ -18,17 +18,21 @@ export default async function ({ addon, console, msg }) {
   setSrc();
   onPauseChanged(setSrc);
 
-  document.addEventListener("keydown", function (e) {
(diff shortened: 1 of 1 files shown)


GarboMuffin:
this didn't work in Firefox, you might need to make the event `{capture:true}` to ensure it always runs before Scratch's listener

DNin01:
> this didn't work in Firefox, you might need to make the event `{capture:true}` to ensure it always runs before Scratch's listener

I was also able to replicate when dynamically enabling the addon. That probably happens if the addon adds the event listener after the Blockly workspace does.

I fixed it by adding `capture: true` to prioritize the addon's keyboard shortcut.

mxmou (approved):

Samq64 (approved):

DNin01:
@WorldLanguages, we're ready to merge!

Issue #6876, opened 2023-11-17
Manifest V3: Master branch is now Manifest V3


WorldLanguages:
# See pull request: #7387

### The news

Google [recently announced](https://developer.chrome.com/blog/resuming-the-transition-to-mv3/) that they will begin disabling Manifest V2 extensions in pre-stable versions of Chrome (Dev, Canary, and Beta) as early as June 2024, in Chrome 127 and later. Users impacted by the rollout will see Manifest V2 extensions automatically disabled in their browser and will no longer be able to install Manifest V2 extensions from the Chrome Web Store. "We expect it will take at least a month to observe and stabilize the changes in pre-stable before expanding the rollout to stable channel Chrome, where it will also gradually roll out over time."
They have announced this already one year ago, but most likely this will actually happen this time.

### What does this mean for Scratch Addons?

**We will upload a Manifest V3 version of the extension to the Chrome Web Store before June 2024.**
As an indirect consequence, our minimum browser requirements will be affected as well. We currently support, to some extent, browsers versions as old as Chrome 80, released in 2020. We will increase the minum requirement to, at least, Chrome 92 (mid-2021). Older browsers won't be able to run the extension, as they don't know what a manifest three extension is, so it's not possible to show a "unsupported browser" screen in those cases.

### Plans on end-user browser support

This is the plan I had in mind to keep up with Chrome's requirements.

1. **Add the "export settings" and "view settings as file" buttons to the Unsupported Browser page.**
This will allow users running old browser versions to export their settings before the extension goes away.

2. **We will release a typical MV2 update which will increase the minimum browser requirements to match the MV3 requirements.**
We will not change our minimum browser requirements as we upgrade to MV3. We will do it earlier. As I said, we can't show any screens or buttons to Chrome versions that don't support basic MV3 functionality, so we have to block them from using the extension earlier.

### Plans on local development and browser support

Our current MV2 extension has a `manifest.json` file at root which is accepted by both Chrome and Firefox. They both log warnings, but the raw extension folder as uploaded to GitHub is accepted by both.
We plan to change the `manifest_version` of our manifest file to `3`, so all local development will use an MV3 extension, both in Chrome and Firefox.

[...]

PR #7231, opened 2024-02-25
Move de- & serializeSettings into its own module

WorldLanguages:
### Changes

The `serializeSettings` and `deserializeSettings` functions were moved into its own module.
These functions are used in the settings page to import and export settings (the buttons inside More Settings).
The extension should continue to work the same after these changes.

### Reason for changes

In preparation for #6876.
Users will be able to export their settings from their unsupported-browser page soon, so the `serializeSettings` method will be imported from a second page.
Most likely, `deserializeSettings` will only ever be useful to the settings page, but it was modularized anyway.

### Tests

Importing, exporting, and viewing settings as a file should continue to work normally from the settings page.

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -8,7 +8,7 @@ import categories from "./data/categories.js";
 import exampleManifest from "./data/example-manifest.js";
 import fuseOptions from "./data/fuse-options.js";
 import globalTheme from "../../libraries/common/global-theme.js";
-import minifySettings from "../../libraries/common/minify-settings.js";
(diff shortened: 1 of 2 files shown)


Samq64 (approved):
I guess it wasn't put in `libraries` because it's only useful in the settings page?

WorldLanguages:
> I guess it wasn't put in `libraries` because it's only useful in the settings page?

This is a MV3 blocker, so I'm not really thinking about stuff like that a lot. Sorry.

Samq64:
There still needs to be an iframe for this to work, right?

Issue #5083, opened 2022-09-05
Replace costume/sound button


Chikorita-Lover:
### Why this would be helpful/solve a problem

Sometimes, I want to update a costume or sound in my sprite. Right now, I have to delete the costume/sound, upload a new one, move it back to the original position in the list, then sometimes rename it.

### What this addon does

This addon adds a button (in the costumes/sounds panels) to quickly replace the currently selected costume/sound, preserving the location in the list and the costume/sound name.
In the costumes panel, this button could be to the right of the "convert to bitmap/vector" button, with the same appearance as the zoom buttons. In the sounds panel, this button could be to the right of the "delete" button, with the same appearance as the buttons to the left.

### Possible alternatives

We could also name this button the reupload button instead of the replace button.

mxmou:
The Scratch 2.0 costume editor had two buttons: "clear" to erase the current costume" and "import" to load a new one. This made it easy to combine multiple pictures into a single costume.

Chikorita-Lover:
Oh, I like that idea, too! It's been so long since Scratch 2.0, I forgot those even existed 🙃

DNin01:
> The Scratch 2.0 costume editor had two buttons: "clear" to erase the current costume" and "import" to load a new one. This made it easy to combine multiple pictures into a single costume.

With that in mind, maybe we could make an addon that lets you do just that and then we could use that as an alternative: Step 1. Delete everything in the current costume. Step 2. Click the Import button added by the addon to insert your new costume.

Chikorita-Lover:
That's what I was thinking, too

WorldLanguages:
Renamed the issue.
I also miss the import feature. But I got used to uploading as a new costume, copying, deleting, then pasting into the sprite I wanted to import.
The import feature also makes it easier to replace sprites, like the original issue post mentioned, without having a dedicated replace button.

WindowsUserInTheHouse:
Maybe it could be renamed to "Replace?" That just sound more like what it does.

Chikorita-Lover:
This was the original idea, but the issue's name was changed by World Languages after further discussion.

Secret-chest:
Why not "add" too?

Chikorita-Lover:
What would an "add" button do?

> > The Scratch 2.0 costume editor had two buttons: "clear" to erase the current costume" and "import" to load a new one. This made it easy to combine multiple pictures into a single costume.
> 
[...]

PR #7232, opened 2024-02-25
New Addon: Asset Conflict Resolution Dialog

### LINKED ISSUE #7177: Prompt when importing conflicting asset names
### Why this would be helpful

When importing a bunch of assets into a project the user might not want them to be renamed automatically.

### How this addon works

The addon would prompt the user to cancel/skip, replace or rename the asset with a textbox that is auto filled with what it would have been renamed to without the addon. Importing multiple files would add a checkbox to apply to all.

Basically just a file manager conflict dialog.

### Possible alternatives

_No response_

### Additional context

Closed PR: #7168

Joeclinton1:
Resolves #7177

### Changes

https://github.com/ScratchAddons/ScratchAddons/assets/48254978/cfdd0d6a-e7a5-4cca-bdef-bb2567be1337

image

### Reason for changes

For projects with assets created via external editors, it is often neccessary to replace the existing costumes with updated versions. Currently you need to delete the old ones, then import the new ones. This is tedious, so this accelerates that process.

### Tests

I have tested:
- Importing a single costume and renaming/replacing/skipping
- Importing multiple costumes and renaming/replacing WITHOUT "do this for all conflicts"
- Importing multiple costumes and renaming/replacing/skipping WITH "do this for all conflicts"
- Importing a single/multiple sounds(s) and renaming/replacing (with for all conflicts if multiple)
- Closing the modal for costume/sounds (desired behaviour is to skip all)
- It's compatible with 'file-drag-and-drop' and'dark-mode' and 'folders' addons
- works with dynamic enable/disable
- works with dragging an asset from one sprite to another
- when an asset is replaced it gets selected
- works when dragging costume/sound from backpack


addons/asset-conflict-dialog/style.css
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.conflictDialog-actions {
+  display: flex;
+  flex-direction: row;
+  margin-bottom: 1rem;
+  gap: 0.5rem;
+}
+
+.conflictDialog-button {
```
Samq64, line 27: Using `addon.tab.scratchClass` you should be able to apply Scratch's button classes which are automatically compatible with editor dark mode.

addons/asset-conflict-dialog/userscript.js
```diff
@@ -0,0 +1,181 @@
...
+            <div class = 'conflictDialog-footer'>
+                <input type="checkbox" id="applyToAll" name="applyToAll">
+                <label for="applyToAll">${safeMsg("applyToAll")}</label>
+            </div>
+          </div>
+      </div>
+    `;
+
+    // Create the modal
+    const { remove, content, closeButton, container } = addon.tab.createModal("Rename, Replace or Skip Assets", {
```
Samq64, line 27: This string should also be translatable.

addons/asset-conflict-dialog/userscript.js
```diff
@@ -0,0 +1,181 @@
+export default async function ({ addon, console, safeMsg }) {
+  function createAssetConflictDialog(fileName, actionClickCallback) {
+    // HTML content for the modal
+    const modalContent = `
```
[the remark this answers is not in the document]
Joeclinton1: But I like the readability of this format.

medlans:
Related thing that used to exist in 2.0 (basically you could just clear and then import into the costume): https://github.com/ScratchAddons/ScratchAddons/issues/5083

Joeclinton1:
> Related thing that used to exist in 2.0 (basically you could just clear and then import into the costume): #5083

Does this possibly also resolve that issue? It's not the desired design but it solves the same issue and potentially better as long as the names are the same.

medlans:
> > Related thing that used to exist in 2.0 (basically you could just clear and then import into the costume): #5083
> 
> Does this possibly also resolve that issue? It's not the desired design but it solves the same issue and potentially better as long as the names are the same.

The import button in 2.0 would add the uploaded thing to the costume instead of replacing, but you had to clear it first to do this, so they're a bit different. There was also the add button in 2.0, which was similar but you would select costumes/backdrops from the library instead.

addons/asset-conflict-dialog/style.css
```diff
@@ -0,0 +1,33 @@
...
+  flex-direction: row;
+  margin-bottom: 1rem;
+}
+
+.conflictDialog-actions > button:hover {
+  background-color: hsla(260, 60%, 60%, 0.15);
+}
+
+.conflictDialog-footer {
+  border-top: solid 1px #aaa;
```
Samq64, line 31:
```suggestion
  border-top: solid 1px var(--editorDarkMode-border, rgba(0, 0, 0, 0.15));
```
For consistency with other borders and dark mode support.

addons/asset-conflict-dialog/userscript.js
```diff
@@ -0,0 +1,178 @@
...
+function createAndAppendElement(type, parent, attrs = {}) {
+  const element = document.createElement(type);
+  Object.assign(element, attrs);
+  parent.appendChild(element);
+  return element;
+}
+
+export default async function ({ addon, console, msg }) {
+  function createAssetConflictDialog(fileName, actionClickCallback) {
+    // HTML content for the modal
```
Samq64, line 11: Old comment?

Joeclinton1:
Noticed a bug which breaks things. Must've been created from one of the new changes

Joeclinton1:
Btw, anyone who does review please actually do some tests, to confirm that it works. Due to the way the code works there are many opportunities for bugs, and there might be a bug that slipped past my testing.

mybearworld:
When a costume called "costume1" exists, clicking on the "New Costume" button does this:
An asset named costume1.undefined already exists.

mybearworld:
The dialog doesn't occur when dragging an asset from one sprite to another.

mybearworld:
When choosing to Import and Replace, the replaced costume/sound isn't selected, but it is when selecting Import and Rename.

Joeclinton1:
> Could this also apply to sprites?

The code for adding sounds and costumes are similar, but adding sprites uses an entirely different method. It would be alot of work. I don't know how useful it would be for people though.

Joeclinton1:
> When a costume called "costume1" exists, clicking on the "New Costume" button does this: An asset named costume1.undefined already exists.

This also happens when dragging from the backpack. Basically I have code which will look for the assets extension and concatenate that to the file name. There must be edge cases where the extension isn't yet defined.

mybearworld:
> This also happens when dragging from the backpack. Basically I have code which will look for the assets extension and concatenate that to the file name. There must be edge cases where the extension isn't yet defined.

Does it even need to show the file name? An asset called "costume.png" and "costume.svg" can't exist at the same time.

mybearworld:
> Well it looks better design wise, as most file conflict editors show the extension

But it doesn't really make sense from a Scratch perspective. File conflict editors show the file extension because it matters - I can have an `image.png`, an `image.svg` and an `image.docx` in the same folder. Can't say the same about `image`, `image` and `image`.

Joeclinton1:
> Was [48d23c6](https://github.com/ScratchAddons/ScratchAddons/commit/48d23c618639eb47b0967dc79b7be7b06484c935) supposed to do [#7232 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/7232#issuecomment-1966889173)? If so, it doesn't seem to do anything.

It sets the costume just doesn't update the costumeTab state so you won't see that it's been set. Until someone can tell me how to actually get the costume tab I can't set it's state. If you look at the stage you'll see it will update to the replaced costume.

Joeclinton1:
There's absolutely no way i'm doing this for sprites too. Way too much work. For now just deleting the old sprite then importing the new one, should be the go to method for that. Since scratch does everything by name and not reference nothing breaks.

GarboMuffin (commented):
The only thing I can find missing is full compatibility with the folders addon. I made it a bit  better, but the folders addon traps the RenderedTarget.addCostume/addSound that  vm.addCostume/addSound internally calls. These are different functions so the addons aren't overwriting each other, but this addon won't be aware of the folders addon inserting `folder name//` at the start to put new assets into the same folder as what the user has open.

Joeclinton1:
> The only thing I can find missing is full compatibility with the folders addon. I made it a bit better, but the folders addon traps the RenderedTarget.addCostume/addSound that vm.addCostume/addSound internally calls. These are different functions so the addons aren't overwriting each other, but this addon won't be aware of the folders addon inserting `folder name//` at the start to put new assets into the same folder as what the user has open.

Is there any actual problem, now that you've made these two fixes? Is there an elegant solution to this?

Joeclinton1:
@GarboMuffin 

```
// finding the right item is made complicated by the folders addon
const selectorItemsNumbers = tab.querySelectorAll("[class*='sprite-selector-item_number_']");
for (const selectorItemNumber of selectorItemsNumbers) {
  const index = +selectorItemNumber.textContent - 1;
  if (index === duplicateIndex) {
    setTimeout(() => {
      selectorItemNumber.parentNode.click();
    });
    break;
  }
}
```
is selecting by the index defined by some text content really the best way?

GarboMuffin:
I think I fixed all the folders compatibility bugs. Also made the duplicate item selecting code use React hacking instead of fake click events.

Another unrelated bug I found is that if there is already a "costume1" costume in the sprite (or whatever the localized equivalent is), adding a new blank costume by clicking on the paint brush icon will ask the user to choose what to do with the new blank costume. Similar happens when recording a new sound.

It's 2AM now so probably should be stopping here, not sure what the best way to resolve that would be.

Joeclinton1:
> I think I fixed all the folders compatibility bugs. Also made the duplicate item selecting code use React hacking instead of fake click events.
> 
> Another unrelated bug I found is that if there is already a "costume1" costume in the sprite (or whatever the localized equivalent is), adding a new blank costume by clicking on the paint brush icon will ask the user to choose what to do with the new blank costume. Similar happens when recording a new sound. I think it should just rename without asking.
> 
> It's 2AM now so probably should be stopping here, not sure what the best way to resolve that would be.

Awesome thank you, I'm going to move the assetSelect code to a utils file, so that other addons can use it, and so that the main code is a bit cleaner. The new costume thing is hard to fix. The two types of calls are almost identical at the lowest level. I think perhaps though the assetObj might give away what type it is will check.

GarboMuffin:
I haven't looked at the objects but a simple solution that should work most of the time would be checking if `asset.name` is `addon.tab.redux.state.locales.messages['gui.sharedMessages.costume'].replace('{index}', '1')` or `"recording1"` (always English) before the folders addon compatibility runs. obviously that's not ideal - but if there's nothing better

Joeclinton1:
> I haven't looked at the objects but a simple solution that should work most of the time would be checking if `asset.name` is `addon.tab.redux.state.locales.messages['gui.sharedMessages.costume'].replace('{index}', '1')` (and also check `gui.sharedMessages.backdrop`) or `"recording1"` (always English) before the folders addon compatibility runs. obviously that's not ideal - but if there's nothing better

The assetObj has a very tell tale signature when it's a new costume, just checking if it's the same for sounds right now

addons/folders/userscript.js
```diff
@@ -1,5 +1,65 @@
...
+const ensureNotReserved = (name) => {
+  if (name === "") return "2";
+  if (RESERVED_NAMES.includes(name)) return `${name}2`;
+  return name;
+};
+
+let currentSpriteFolder = null;
+let currentAssetFolder = null;
+
+export const addDefaultAssetFolderIfMissing = (asset) => {
```
WorldLanguages, line 57: Could we add a code comment here that explains when other addons should call this?

Joeclinton1: @GarboMuffin

Joeclinton1:
I tested this more extensively and found no bugs when using the sprite folders addon

GarboMuffin:
No dialog if I drag in a conflicting costume from the backpack

WorldLanguages:
> No dialog if I drag in a conflicting costume from the backpack

Oof... nice catch
RIP when this addon was less than 200 lines long

Joeclinton1:
> > No dialog if I drag in a conflicting costume from the backpack
> 
> Oof... nice catch RIP when this addon was less than 200 lines long

As Garbo hinted at this is probably a side effect of the previous fix. I tested quite a few ways of adding the costume and thought the signature was specific to the new costume, but it's possible dragging from backpack has the same signature.

GarboMuffin (approved):

Samq64:
Dropping files onto the backdrop pane or uploading them outside the costume editor doesn't trigger the dialog.

Joeclinton1:
> Dropping files onto the backdrop pane or uploading them outside the costume editor doesn't trigger the dialog.

Will fix when I have some free time

Joeclinton1:
> Dropping files onto the backdrop pane or uploading them outside the costume editor doesn't trigger the dialog.

I can reproduce the first thing, but I don't understand what you mean by the second. Could you please clarify this?

Joeclinton1:
Making it work for the backdrop was quite easy. It is unfortunate Scratch chose to have a separate add backdrop function.

Samq64 (approved):
I haven't taken a full look at all the code but drag and drop, sounds, backdrop and HD uploads all work fine.

Issue #5674, opened 2023-01-28
Rename some addon GitHub (code) names


penguinmoose:
Addons have titles, but they also have (shorter) names for referencing them in code and GitHub. For example, "Emoji picker for comments" is `emoji-picker` and "Messages in editor" is `editor-messages`. Most of these names are fine, but some are a bit confusing and different from the addon title, or what it actually does. This is probably because these names are set early on, but remain unchanged even when new features are added to the addon, or its primary purpose changes.

**Examples**
Scratch 2.0->3.0 is `scratchr2`, which is unclear (to most people)
Customizable block colors is `editor-theme3`, which is also unclear.
Insert blocks by name is `middle-click-popup`. While this is one of its features, it doesn't describe the entire addon well. I think that something like `insert-blocks` or `type-blocks` or `block-select-popup` would be better.
There are some other addons that have somewhat ambiguous names, like `dark-www` and others.

Samq64:
I tried opening a PR a while back but it got rejected: #5114 

Renaming an addon ID doesn't just mean renaming the addon's folder and the entry in `addons.json`, but also every reference to the addon in the whole extension. We would also need to add migration code so that users don't lose their settings for that addon. Multiply that by 10 and it becomes a giant mess.

mxmou:
> Insert blocks by name is `middle-click-popup`. While this is one of its features, it doesn't describe the entire addon well. I think that something like `insert-blocks` or `type-blocks` or `block-select-popup` would be better.

I think this one is fine. Middle click isn't the only way to access the popup but it's the most common.

If we ever start changing IDs, here are a few suggestions:
* `columns` → `toolbox-columns`
* `copy-message-link` → `copy-comment-link`
* `dark-www` → `website-dark-mode`
* `discuss-button` → `custom-navbar`
* `editor-theme3` → `custom-block-colors`
* `fix-pasted-scripts` → `fix-duplicated-scripts`
* `full-signature` → `full-areas`
* `my-ocular` → `ocular`
* `scratchr2` → `scratch2to3`
* `search-profile` → `user-search`? (see #5461)
* `tutorials-button` → `hide-tutorials-button`

CST1229:
also: `color-picker` -> `hex-color-picker`

mxmou:
> also: `color-picker` -> `hex-color-picker`

Quite a few addon IDs are names of a Scratch feature and don't actually describe what the addon does:
* `clones` (→ `clone-counter`)
[...]

PR #7240, opened 2024-03-01
Create "Addons By Name" table

DNin01:
Partially addresses #5674

### Changes

Adds a new "Addons By Name" document as `/addons/README.md`. It's a table of addon names (in alphabetical order) along with their IDs.

The table is automatically generated via the `gen-addon-table.py` Python script. Every time new addons are added or addon names are changed, this script must be re-run to update the addon table.

I'm new to Python, so let me know if there's something I can improve!

I'm also not sure if we need this, but I'll leave this open.

### Reason for changes

It can be hard to find the addon IDs of certain addons because sometimes they do not closely correspond to their addon's name or are ambiguous. The table solves that problem. It's also useful if you want to reference an addon by its ID but don't know it.

### Tests

Tested on Windows with Python 3.11.

### DIFF
--- a/addons/README.md
+++ b/addons/README.md
@@ -1,0 +1,178 @@
+# Addons By Name
+
+| Addon Name | Addon ID |
+|---|---|
+| 2D color picker | [2d-color-picker](/addons/2d-color-picker) |
+| 3.0 Scratchblocks on forums | [scratchblocks](/addons/scratchblocks) |
+| Alternating nested block colors | [zebra-striping](/addons/zebra-striping) |
+| Always show number pad | [number-pad](/addons/number-pad) |
+| Arrow key incrementation in editor | [editor-number-arrow-keys](/addons/editor-number-arrow-keys) |
+| Auto-hiding block palette | [hide-flyout](/addons/hide-flyout) |
+| Automatically add extensions | [load-extensions](/addons/load-extensions) |
+| BBCode source button | [show-bbcode](/addons/show-bbcode) |
+| Better editor comments | [fix-editor-comments](/addons/fix-editor-comments) |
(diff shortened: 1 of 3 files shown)


addons/README.md
```diff
@@ -0,0 +1,178 @@
+# Addons By Name
```
DNin01, line 1: Should "By" be uppercase or lowercase? I couldn't decide.

dev/gen-addon-table.py
DNin01, line 1: Is the `dev` directory a good place for this script?

DNin01, line 1: Not sure if we should have some forms of sanitization or directory checking (making sure we don't go outside the intended directories).

DNin01:
I wonder if a script like this could be used to build the settings page's addon list (so our extension doesn't have to repeatedly spend time generating it), or if it could generate a single file containing all the addon manifests in one file (would that be useful at all?).

Requiring contributors to run a script could raise the floor for contributors, though. But more software is starting to use Node.js on the development side, so...

Joeclinton1:
Disapprove. The best way to handle this is a util function that takes an addon name and returns an ID. This function would perform the same internal logic as this addon, but would mean there is no need for a static file that needs to be mantained and updated. There should not be two sources of truth.

DNin01:
I was going to say, why not just make a page part of the Scratch Addons extension that does the same thing?

I don't really like unnecessarily complicating the development process.

DNin01:
But you can also just hover over addon names to see their IDs in the settings page (#6786), so this is already sort of solved?

WorldLanguages:
Yeah, this is a quite similar concept to https://scratchaddons.com/addons/.

PR #4515, opened 2022-04-19
New "bugfix" tag for addons

Norbiros:
Resolves https://github.com/ScratchAddons/ScratchAddons/pull/2568#issuecomment-876833246

### Changes

Adds new tag to addons: `don't run duplicated blocks` and `fix SVG upload`
file

### Tests

Firefox newest.

PR #7242, opened 2024-03-03
Minor addons group

### LINKED ISSUE #2617: Minor addons section/tag
I am still not satisfied of how the new tag works. I thought it will solve this exact issue, but it isn't, so I am proposing this change.

How about a tag for minor addons? This tag is for addons that only does very little change (or unnoticable), such as the curator links, bug fix addons, and that small addon that somehow assigned as an easter egg addon.

### LINKED ISSUE #7152: All-in-one bug-fix addon
### Why this would be helpful

Sometimes there's a bug or undesired behavior in Scratch that we want to fix with an addon. If I got them all, these include:
- Faster project loading
- Resizable comment input?
- Place backpack code at mouse
- Better editor comments
- Do not automatically run duplicated blocks
- Fix sprite pane glitching
- Redirect to standard forums (now that the old mobile view of the forums has been taken down)
- Remember collapsed forum categories
- SVG uploads fix
- The `Use previous color instead of resetting after switching tools` setting in Customizable default costume editor colors

The main issue with these is that the user has to be looking out for them when they're a great opportunity to have bug fixes enabled by default. Surprisingly, most of these are not enabled by default.

Some of these are also so minor they almost don't warrant a whole addon.

### How this addon works

This addon would just be a collection of bug fixes and other subconscious UX enhancements that objectively, everyone would use. It'd be enabled by default and each bug fix would be a setting, so you could see what is being patched, and turn any patches off if for some reason you want to.

Some of the above addons could be moved into this addon.

Perhaps this addon could have a confirmation when being disabled, so no one accidentally turns it off and ruins parts of their experience.

### Possible alternatives

We could also add a "Bug fixes" category of addons and separate these fixes into their own addons which would only show up in this category.

### Additional context

Related:
- #6409

DNin01:
Resolves #2617, resolves #7152

### Changes

Adds a group to the addon list in the settings page called "Bug fixes" (we can change this as we see fit). These addons...
- do not appear under Enabled, but can still be searched
- show a confirmation before disabling

### Reason for changes

Some addons fix bugs and those addons deserve to be enabled by default, but that would also clutter the "Enabled" group in the list of addons. The "bug fixes" group allows us to set them aside, making more room for addons that users would actually want to take a look at.

### Tests

Tested on Edge 122.

### To-Dos

- [ ] Update the manifest schema
- [ ] What should we do with these addons in the extension popup's "Addons on this tab" group?
- [ ] What should we do with `fix-uploaded-svgs`?

We might also want to address any confusion around why these addons aren't showing up under "Enabled".

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -50,6 +50,12 @@
   "beta": {
     "message": "Beta"
   },
+  "minor": {
+    "message": "Bug fixes"
+  },
+  "minorDisableWarning": {
+    "message": "The \"$1\" addon fixes undesireable behavior in Scratch and we strongly recommend having it enabled. Are you sure you want to disable it?"
+  },
   "danger": {
     "message": "Dangerous"
   },
--- a/addons/disable-sprite-wobble/addon.json
+++ b/addons/disable-sprite-wobble/addon.json
@@ -48,8 +48,8 @@
       "default": "none"
     }
   ],
-  "tags": ["editor", "recommended"],
-  "enabledByDefault": false,
+  "tags": ["editor", "recommended", "minor"],
+  "enabledByDefault": true,
   "versionAdded": "1.24.0",
   "latestUpdate": {
     "version": "1.35.0",
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -92,7 +92,7 @@ async function transitionToNewStorageKeys(addonSettings) {
 }
 
 const ADDON_SETTINGS_KEYS = ["addonSettings", "addonSettings1", "addonSettings2", "addonSettings3"];
-chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems) => {
+chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled", "lastVersion"], (storageItems) => {
   const isSettingsStorageTransitionPending = storageItems.addonSettings && !storageItems.addonSettings3;
   if (isSettingsStorageTransitionPending) {
     transitionToNewStorageKeys(storageItems.addonSettings);
@@ -103,6 +103,8 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
   const addonSettings = areAddonSettingsEmpty
     ? {} // Default value
     : { ...storageItems.addonSettings1, ...storageItems.addonSettings2, ...storageItems.addonSettings3 };
+  const lastVersionStr = storageItems.lastVersion;
+  const lastVersion = lastVersionStr?.split(".") || ["0", "0", "0"];
   const func = () => {
     // Start by migrating settings (sometimes we add new settings or make changes to
     // the available settings in some addons between versions)
@@ -686,10 +688,16 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
 
     // Finally, minify the settings and store them in the scratchAddons object
     const prerelease = chrome.runtime.getManifest().version_name.endsWith("-prerelease");
+    const currentVersion = chrome.runtime.getManifest().version;
+    console.log(`Transitioned from ${lastVersionStr} \u2192 ${currentVersion}`);
+    if (currentVersion !== lastVersionStr) {
+      madeAnyChanges = true;
+    }
     if (madeAnyChanges)
       chrome.storage.sync.set({
         ...minifySettings(addonSettings, prerelease ? null : scratchAddons.manifests),
         addonsEnabled,
+        lastVersion: currentVersion,
       });
     scratchAddons.globalState.addonSettings = addonSettings;
     scratchAddons.localState.addonsEnabled = addonsEnabled;
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -163,6 +163,13 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled", "lastVersion"]
       }
     }
 
+    if (lastVersion[1] < 37) {
+      // Enable addons that became enabled by default starting v1.37.0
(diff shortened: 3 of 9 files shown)


webpages/settings/index.js
DNin01, line 1: I also have written code for enabling `fix-uploaded-svgs` for 10% of users, but I'm not sure if I want to include that in this PR without some discussion first.

Samq64:
Related open PR: #4515

addons/disable-sprite-wobble/addon.json
```diff
@@ -48,8 +48,8 @@
       "default": "none"
     }
   ],
-  "tags": ["editor", "recommended"],
-  "enabledByDefault": false,
+  "tags": ["editor", "recommended", "minor"],
```
Samq64, line 51: This is the only addon with a setting and the new tag.

DNin01: TODO:
- [ ] If the `disable-sprite-wobble` addon gets enabled automatically, the `"mode"` setting should be reset.

background/get-addon-settings.js
```diff
@@ -686,10 +695,16 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
 
     // Finally, minify the settings and store them in the scratchAddons object
     const prerelease = chrome.runtime.getManifest().version_name.endsWith("-prerelease");
+    const currentVersion = chrome.runtime.getManifest().version;
+    console.log(`Transitioned from ${lastVersionStr} \u2192 ${currentVersion}`);
```
DNin01, line 699: Maybe should only log this if changes were made.

Samq64:
What should we do with Better editor comments?

There are also a couple forum bugfix addons:
- Redirect to standard forums
- Remember collapsed forum categories

DNin01:
> What should we do with Better editor comments?

I know, right?

> There are also a couple forum bugfix addons:
> 
> * Redirect to standard forums

I'll include that once #6915 is merged. It changes the addon ID.

> * Remember collapsed forum categories

That one I think might be best left for the user to decide on, but maybe.

webpages/settings/index.js
DNin01: @WorldLanguages, how do you feel about this?

WorldLanguages: For ten percent of users? What does that mean?

DNin01: > For ten percent of users? What does that mean?

Just using Math.random() and if it's less than 0.1, enable it if it's currently not, otherwise don't. (As part of the migration logic for transitioning to v1.37 or whenever)

The point being, it would reduce the impact of any undetected bugs in the addon. (After what happened with Faster project loading)

Anyway, do you have any opinions on enabling an addon by default for a subset of users? Should we do that at all or just enable it for everyone?

WorldLanguages: > Just using Math.random() and if it's less than 0.1

Hmm... Sounds weird

WorldLanguages:
Does this actually resolve #7152? I thought that issue was about using setting descriptions and a single addon for most bug fixes.

webpages/settings/index.js
DNin01: > > Just using Math.random() and if it's less than 0.1
> 
> Hmm... Sounds weird

Probably should open an issue first for that, then.

Hans5958:
> Does this actually resolve #7152? I thought that issue was about using setting descriptions and a single addon for most bug fixes.

It doesn't resolve the issue but will close the issue nonetheless as it won't be implemented because other alternatives has been chosen.

Secret-chest:
> forum bugfix

This would also be the right time to change forums to a website subcategory and not a group.

DNin01:
> > forum bugfix
> 
> This would also be the right time to change forums to a website subcategory and not a group.

I think the main reason the forum addons are in a separate group is because not everyone uses the forums and the forums are a smaller area of the entire website.

mxmou:
> not everyone uses the forums

Almost no one uses the forums.

DNin01:
> > not everyone uses the forums
> 
> Almost no one uses the forums.

Yeah, that's a better way of putting it. 😄

Secret-chest:
> I think the main reason the forum addons are in a separate group is because not everyone uses the forums and the forums are a smaller area of the entire website.

We could just make it list them at the bottom.

mxmou:
> > I think the main reason the forum addons are in a separate group is because not everyone uses the forums and the forums are a smaller area of the entire website.
> 
> We could just make it list them at the bottom.

The way addons are sorted is already confusing: #3012

Secret-chest:
Then let's just put forum bugfix addons as a bugfix.

_locales/en/messages.json
```diff
@@ -50,6 +50,12 @@
   "beta": {
     "message": "Beta"
   },
+  "minor": {
+    "message": "Bug fixes"
+  },
+  "minorDisableWarning": {
+    "message": "The \"$1\" addon fixes undesireable behavior in Scratch and we strongly recommend having it enabled. Are you sure you want to disable it?"
```
mxmou, line 57:
```suggestion
    "message": "The \"$1\" addon fixes undesirable behavior in Scratch and we strongly recommend having it enabled. Are you sure you want to disable it?"
```

mxmou:
I don't think addons in the bug fixes group need the "recommended" tag.

> What should we do with these addons in the extension popup's "Addons on this tab" group?

There could be a separate "bug fixes on this tab" group below "addons on this tab".

WorldLanguages:
Just a comment: I haven't seen this PR in detail yet, but if there's consensus in favor of this I will just merge it :)

WorldLanguages:
> Adds a group to the addon list in the settings page called "Bug fixes" (we can change this as we see fit). These addons...
> 
> * do not appear under Enabled, but can still be searched
> * show a confirmation before disabling

Personally, I think this is a good idea. As these bug-fix addon usually don't have settings and there's no good reason to disable them, I don't see the point of listing them in the "enabled" group. These should, however, still be listed within the "addons (running) on this tab" group when using the settings page from the popup.

PR #7243, opened 2024-03-03
Recommend `auto-follow-topics` addon

DNin01:
The "Follow topics after posting" addon is very cool. I think many forums these days notify participants of forum threads of new activity, but Scratch's doesn't - only automatically for topics you create. That's why I would recommend this addon. A lot of us would probably want to know when someone responds to a topic they're participating in, most of the time.

### To Do

- [ ] Figure out whether this addon should be in the "Recommended" group or not


Samq64:
I'm not sure if it's intentional but it doesn't look like any forum exclusive addons have the recommended tag right now.

Chiroyce1 (approved):

DNin01:
> I'm not sure if it's intentional but it doesn't look like any forum exclusive addons have the recommended tag right now.

We should test if there's anything that doesn't work right in the settings page due to this change. I might have just made this addon disappear from the Forums group.

OneShot-Niko:
> We should test if there's anything that doesn't work right in the settings page due to this change. I might have just made this addon disappear from the Forums group.

Nope, just tried on LibreWolf:
reccomended

Waakul (approved):

Waakul (approved):

DNin01:
> > We should test if there's anything that doesn't work right in the settings page due to this change. I might have just made this addon disappear from the Forums group.
> 
> Nope, just tried on LibreWolf:
> 
> reccomended

I meant the section in the list, not the category.

OneShot-Niko:
> I meant the section in the list, not the category.

Oh, my apologies then!

DNin01:
It does not show up under Forums, but instead Recommended. I suspected this because the condition for putting addons there is inside an else-if condition.

But wait, we already have some forum addons that are recommended or featured and also don't appear under Forums. However, those are all addons that also work on other areas of Scratch.

@WorldLanguages @Hans5958 What should we do about this? Should we ever recommend or feature forum-exclusive addons? These addons are probably separated into their own group for a reason, so if we were to recommend anything and keep that separation, we could place them under Forums instead of Recommended if they only do anything on forum pages.

WorldLanguages:
> It does not show up under Forums, but instead Recommended.

I think that's fine.

> I suspected this because the condition for putting addons there is inside an else-if condition.

We could change the grouping logic so that it shows in both groups, if we want (recommended forum addons would be listed twice).

DNin01:
Another way to keep the separation but solve the problem of having forum addons not shown under the Forums group would be to place forum addons under other addons in the list, kind of like how we place website addons under editor addons, and then move everything under the Forums group into other groups.

WorldLanguages (approved):
I think we should merge this, and consider changing the grouping algorithm later

Samq64 (approved):

DNin01:
I was experimenting with changing the grouping algorithm and honestly, I think we're okay to leave it as is where an addon can only be under one group, at least for a while.

WorldLanguages:
At some point I will no longer participate in settings page development, so my opinion no longer matters

PR #5904, opened 2023-04-08
New addon: "Move sprite to front layer"

scratchforsccom:
## See original PR: #5673 (sfederici)

New addon "Move to Top Layer" to move a sprite to the top layer when shift-clicking its thumbnail in the sprite list. I extracted the code for this addon from a larger addon created by Norbiros that wasn't fully accepted by ScratchAddons.

Resolves #4231

### Changes

Created new addon "move-to-top-layer"

### Reason for changes

When a sprite is covered by other sprites on the Stage so that it is not visible, to make it visible you have to go to the palette and run a "go to front layer" block. By enabling this extension it can be quickly brought to the front layer by simply shift-clicking its thumbnail in the sprite list

### Tests

I tested this extension locally as an unpacked extension.

PR #7244, opened 2024-03-03
move-to-top-layer: Add context menu and support folders addon

### LINKED ISSUE #5998: Enhance "move sprite to front layer" addon
- Do not bring sprite to the front if the trash icon or a context menu item was Shift+Clicked (to avoid potential issues)
- Add a "move to front layer" context menu item

_Originally posted by @WorldLanguages in https://github.com/ScratchAddons/ScratchAddons/issues/5904#issuecomment-1527704431_

Addon ID: move-to-top-layer

Samq64:
Resolves #5998
Resolves https://github.com/ScratchAddons/ScratchAddons/pull/5904#issuecomment-2634174144

### Changes

Adds a context menu item to the move sprite to front layer addon and always shows the sprite when used. The message is not included here since it already exists from the original PR. This also fixes compatibility with the folders addon.

### Reason for changes

It's more intuitive and mobile-friendly.

### Tests

Tested on Firefox.


WorldLanguages:
> I shift clicked the delete and info buttons and I didn't notice any problems.

Hmm, I wonder what I meant by that then...

addons/move-to-top-layer/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Move sprite to front layer",
-  "description": "Shift+Click a sprite within the sprite pane to move it to the front (top layer) of the stage.",
+  "description": "Right click or shift+click a sprite within the sprite pane to move it to the front (top layer) of the stage.",
```
mxmou, line 3: This sounds like right-clicking and shift-clicking do the same thing.

Samq64:
> > I shift clicked the delete and info buttons and I didn't notice any problems.
> 
> Hmm, I wonder what I meant by that then...

Shift-clicking the buttons still brings the sprite to the front though when it probably shouldn't

addons/move-to-top-layer/userscript.js
```diff
@@ -1,15 +1,37 @@
...
         const spriteName = parentDiv.querySelector("div[class^='sprite-selector-item_sprite-name']").innerText;
-        // move the sprite with that name to front
-        vm.runtime.getSpriteTargetByName(spriteName).goToFront();
+        spriteToFront(spriteName);
       }
     }
   });
+  addon.tab.createEditorContextMenu(
+    (ctx) => {
+      spriteToFront(ctx.name);
```
mxmou: This doesn't work if `folders` is enabled because it replaces the sprite's name with an object. The best way to fix this is probably in the addon API: [here](https://github.com/ScratchAddons/ScratchAddons/blob/833fff4/addon-api/content-script/contextmenu.js#L94), `props.name.realName` should be used instead of `props.name` if it exists.

Samq64: How should I handle compatibility when shift-clicking?

mxmou: One way - I don't know if it's the simplest - is to read the name from the React component props instead of the DOM. If you do that, you can also read the ID instead of the name, which is simpler because you won't need additional logic for when folders is enabled. The context menu code could also be simplified by using `ctx.itemId` instead of `ctx.name`.

Issue #7245, opened 2024-03-03
Custom palette scroll speed


LilyMakesThings:
### Why this would be helpful

The palette's scrolling can be disorienting; if you're building a script really quickly it's very easy to drag out or misclick on the wrong blocks.

### How this addon works

You can select the speed you want the palette to scroll at; you can either slow it down, speed it up or remove scrolling entirely (in which case it snaps to the category you select).

BlueGoat11:
I think that would be one of the best addons for me…On Mar 2, 2024, at 10:25 PM, LilyMakesThings ***@***.***> wrote:
Why this would be helpful
The palette's scrolling can be disorienting; if you're building a script really quickly it's very easy to drag out or misclick on the wrong blocks.
How this addon works
You can select the speed you want the palette to scroll at; you can either slow it down, speed it up or remove scrolling entirely (in which case it snaps to the category you select).
Possible alternatives
No response
Additional context
No response

DNin01:
I'm not sure most people would care about changing the speed of the scroll animation. 🤔

The option to disable it entirely for people who are set off by it would be nice, though. 👍

But I suppose it might not hurt to have speed options. On the off-chance that someone wants to customize the scroll speed, this could be a nice thing to have around. We'll need to discuss this and decide whether we want to do that. I personally don't like to overload users with unnecessary customization options.

If we decide on just having an on/off switch for the scroll animation, we could even include this in an addon about reducing motion.

BlueGoat11:
I agree with the customization overload ideology, but I do like the idea of a custom scroll speed, or even just a don’t scroll option. When I’m speed-scratching that little boost of a scroll speed would make a difference. For me, I can never by memory remember where each block is in each section, so just being able to scroll faster might increase productivity. Plus, it probably wouldn’t really be hard to implement…

DNin01:
If there's enough consensus, I think this feature would be a nice addition.

LilyMakesThings:
I've been working on a project to test this (as well as other unrelated things). I tested faster scroll speeds and slower scroll speeds. Both have their own advantages.

BroJac5246:
I like the idea, but are there any existing addons this could be made part of? This feels too small IMO.

BlueGoat11:
I don’t know…

Samq64:
@BlueGoat11 If you don't have anything productive to add please just leave a comment reaction on the website instead.

LilyMakesThings:
[...]

PR #7246, opened 2024-03-03
Custom palette scroll speed

LilyMakesThings:
Resolves https://github.com/ScratchAddons/ScratchAddons/issues/7245

### Changes

Changes the speed that the palette scrolls when selecting a category.

### Reason for changes

The scrolling animation within the palette can be rather disorienting, and being able to change the speed or disable it entirely would be beneficial.

### Tests

Works on Opera, Chrome and Firefox.


addons/editor-palette-scroll/addon.json
```diff
@@ -0,0 +1,48 @@
...
+  "credits": [
+    {
+      "name": "LilyMakesThings",
+      "link": "https://scratch.mit.edu/users/LilyMakesThings/"
+    }
+  ],
+  "enabledByDefault": false,
+  "userscripts": [
+    {
+      "url": "userscript.js"
```
Samq64, line 14: This never runs because it's missing `matches` which is required since not all addons run in  the editor.

Samq64 (changes requested):
The extension crashes and after fixing that the settings do the opposite of what's expected.

LilyMakesThings:
Honestly thinking this would work better as part of a different addon. As others have pointed out, it's a little too small.

PR #7247, opened 2024-03-03
New addon: sort assets by alphabetical order

jazzathedev:
Resolves suggestion from the [community server](https://discord.com/channels/806602307750985799/1212743716301832332)

### Changes

Adds an option to the	edit menu to sort the assets by alphabetical order.

### Reason for changes

In a future version of this when I make it, it could be used to fix the order of broken gifs? will add settings for different sorting methods, current is basic js array sort.

### Tests

works on brave for both costumes and sounds, sprites are not done yet, and i need pointers on how to go about that. if its not feasible, sprites will be ommited (they arent that useful to order anyway)


Jazza-231:
Alphabetical
Current implementation allows costumes and sounds to be sorted using .sort()

Jazza-231:
I have implemented a better sorting system, which uses [natural sort](https://snipplr.com/view/36012/javascript-natural-sort). this is just...better, but it is slower, and when a lot of things are sorting, it can take a while. I am not sure if this can be improved. besides that, sprites still need to be done, and for consistency i would like to find a way to make the menu close when the item is clicked on while it is greyed out (which you will see when you click on the grayed out "restore" button) i am not sure how to do either of those, and believe me i have tried. if you have ideas, please tell me, and if you think those things are fine as is, i guess no changes to them need to be made

Jazza-231:
https://github.com/ScratchAddons/ScratchAddons/assets/81956724/913d45d6-f2f9-4e07-b39e-f187ed71e8dc

oh, and anyone know why it doesnt come back when the addon is reenabled while the menu is open and it was disabled before the menu was opened, but it does come back when it was enabled before the menu was opened, disabled when the menu was open, and reenabled when the menu was still open?

Joeclinton1:
For sprite sorting a function exists similar to reorderSound or reorderCostumes, it's called [reorderTarget](https://github.com/scratchfoundation/scratch-vm/blob/develop/src/virtual-machine.js#L1441).

Jazza-231:
> For sprite sorting a function exists similar to reorderSound or reorderCostume, it's called [reorderTarget](https://github.com/scratchfoundation/scratch-vm/blob/develop/src/virtual-machine.js?rgh-link-date=2024-03-03T12%3A03%3A37Z#L1441).

oh my god thank you how did i not see that

Jazza-231:
(latest commit does not work)

Jazza-231:
ok im pretty sure it works, sprites are done, the performance is decent, now just need to figure out how to close that menu

Joeclinton1:
Chatgpt thinks your natural sort can be done by using localCompare, which would make it alot more concise. Could you test?
```js
Array.prototype.naturalSort = function() {
    return this.sort((a, b) => 
        String(a).localeCompare(String(b), undefined, {numeric: true, sensitivity: 'base'})
    );
};
```

Jazza-231:
> Chatgpt thinks your natural sort can be done by using localCompare, which would make it alot more concise. Could you test?
> 
> ```js
> Array.prototype.naturalSort = function() {
>     return this.sort((a, b) => 
>         String(a).localeCompare(String(b), undefined, {numeric: true, sensitivity: 'base'})
>     );
> };
> ```

well ill be damned it seems to sort the same

Joeclinton1:
> ok im pretty sure it works, sprites are done, the performance is actually really good, now just need to figure out how to close that menu

See [here](https://github.com/scratchfoundation/scratch-gui/blob/34604ea3cf3ea511f1b2ed3b753023a4e325eb32/src/components/menu-bar/menu-bar.jsx#L272)

You need to get the menu-bar and call `this.props.onRequestCloseEdit()`

BlueGoat11:
I like thatOn Mar 3, 2024, at 9:13 AM, Joe Clinton ***@***.***> wrote:﻿

ok im pretty sure it works, sprites are done, the performance is actually really good, now just need to figure out how to close that menu

See here
You need to get the menu-bar and call this.props.onRequestCloseEdit()

Jazza-231:
> > ok im pretty sure it works, sprites are done, the performance is actually really good, now just need to figure out how to close that menu
> 
> See [here](https://github.com/scratchfoundation/scratch-gui/blob/34604ea3cf3ea511f1b2ed3b753023a4e325eb32/src/components/menu-bar/menu-bar.jsx#L272)
> 
> You need to get the menu-bar and call `this.props.onRequestCloseEdit()`
> 
> Edit: if you look at what onRequestCloseEdit calls, you can probably close the menu with a redux dispatch

How do you find this shit, I looked so hard...I can't do this tonight so remind me to do it tmr

Jazza-231:
I am happy to have figured out that redux event on my own, redux is one of my weaknesses so that felt good, anyway i think this is all done now!

Joeclinton1 (changes requested):
The actual reordering works perfectly for sprites, costumes and sounds in my tests. Edit menu works fine.

1. Dynamic/enable works properly in all cases except the edge case that the menu is already open. Can you make it check if the menu is open when enabled and insert the option?
1. The originally selected asset is no longer selected after sorting, can you fix this? Take a look at the code in #7232 for how to select an asset. You can probably copy and paste it. You'll need to do the sprites a little bit differently though, but it's really just a case of adding another selector.
1. I would've preferred a context menu option over an edit menu option, but i guess since this is applied to all costumes instead of one it kind of makes sense for it to not need context. Having to go to the code tab to sort sprites is a little awkward but sprites are kind of an after thought anyway.
1. I don't like the While True loop. Addons should not use while loops unless absolutely neccessary for performance reasons. There is an event triggered when that menu is opened why not just trigger your code on that? Otherwise the code is good.

Jazza-231:
1. Done
2. Done
3. See discord reply
4. Done

Jazza-231:
> Has a bug
> 
> image
> 
> Edit Menu doesn't close when option is clicked due to this bug. Tried to reorder sprites for context, but also happens with the other tabs.

you were on the editor tab yeah? i think i know why this is happening

Jazza-231:
should be fixed now, i cant repo the edit menu not closing as it always does for me, but this should be working now.

Joeclinton1 (approved):
Looks good to me!

Jazza-231:
I think this just needs reviews and approvals

DNin01:
> I think this just needs reviews and approvals

Joeclinton1 has already approved this pull request.

addons/alphabetical-order/userscript.js
```diff
@@ -0,0 +1,126 @@
+export default async function ({ addon, msg, console }) {
+  const vm = addon.tab.traps.vm;
+
+  Array.prototype.naturalSort = function () {
```
WorldLanguages, line 4: Is it necessary to add this to the prototype? Doesn't sound like the best idea. Why not just make it a normal function that receives the array as an argument?

WorldLanguages:
What's the expected behavior of Ctrl+Z/undo after sorting alphabetically?

addons/alphabetical-order/userscript.js
Jazza-231: It's not necessary but I find it much more convenient in this scenario and it has no negative effects as far as I can tell.

Jazza-231:
> What's the expected behavior of Ctrl+Z/undo after sorting alphabetically?

Well one would assume it will undo the reordering.. I'm not sure if this actually happens, I think all I need to do is call some function in the VM and that will save it as a snapshot that can be undone.

DNin01:
> What's the expected behavior of Ctrl+Z/undo after sorting alphabetically?

Are you talking about the undo buttons in the costume editor?

Ctrl+Z can't usually undo costume deletions, additions, or reorders, AFAIK.

Jazza-231:
> Ctrl+Z can't usually undo costume deletions, additions, or reorders, AFAIK.

That is correct, but perhaps in this case it should? Or maybe change the restore button to be restore <asset> order?

addons/alphabetical-order/userscript.js
WorldLanguages: Well, https://flaviocopes.com/javascript-why-not-modify-object-prototype/

WorldLanguages:
> Ctrl+Z can't usually undo costume deletions, additions, or reorders, AFAIK.

Well, I meant `Edit > Restore Costume` then.
Of course, Scratch won't let you undo reorders, because you can only do one reorder at a time, so it's trivial to undo it manually. But if we're going to do multiple reorders with a click of a button, it sounds sensible to add a way to undo it...

DNin01:
Related addon idea: notification with undo button when deleting things or doing other destructive actions, including alphabetizing a sprite's costumes or sounds.

Jazza-231:
> > Ctrl+Z can't usually undo costume deletions, additions, or reorders, AFAIK.
> 
> Well, I meant `Edit > Restore Costume` then. Of course, Scratch won't let you undo reorders, because you can only do one reorder at a time, so it's trivial to undo it manually. But if we're going to do multiple reorders with a click of a button, it sounds sensible to add a way to undo it...

I am about to add functionality to the edit menu to restore the previous order using restore

Joeclinton1:
The while loop which keeps checking if it isSorted is really questionable. You should just sort it once, then directly sort it with reorder, there's no need to check if it's actually sorted if you could just sort it correctly in one pass. There absolutely should not be anything checking iteration counts.

Joeclinton1 (changes requested):
The while loop which keeps checking if it isSorted is really questionable. You should just sort it once, then directly sort it with reorder, there's no need to check if it's actually sorted if you could just sort it correctly in one pass. There absolutely should not be anything checking iteration counts.

I probably shouldn't have approved this, I missed the while loop logic.

Jazza-231:
> The while loop which keeps checking if it isSorted is really questionable. You should just sort it once, then directly sort it with reorder, there's no need to check if it's actually sorted if you could just sort it correctly in one pass. There absolutely should not be anything checking iteration counts.
> 
> I probably shouldn't have approved this, I missed the while loop logic.

Feel free to commit changes, ill give you contributor access. I seem to be having trouble with the whole idea or sorting and I'm not sure why

Joeclinton1:
> > The while loop which keeps checking if it isSorted is really questionable. You should just sort it once, then directly sort it with reorder, there's no need to check if it's actually sorted if you could just sort it correctly in one pass. There absolutely should not be anything checking iteration counts.
> > I probably shouldn't have approved this, I missed the while loop logic.
> 
> Feel free to commit changes, ill give you contributor access. I seem to be having trouble with the whole idea of sorting and I'm not sure why

If I'm making changes i'm going to redo most of your logic

Jazza-231:
> If I'm making changes i'm going to redo most of your logic

Fair enough, if you want to do that then go ahead, feel free to add yourself to the credits too, I am genuinely fully stumped as to why its not working

Joeclinton1:
I did the rewrite, seems to work now.

Perhaps an option could be added to settings for descending vs ascending sort?

Jazza-231:
Any ideas why the prettier check is failing?

Joeclinton1:
> Any ideas why the prettier check is failing?

It's been failing on things recently

Joeclinton1:
I'm going to list the things that need to be tested for this addon:

- Sort alphabetically appears if the assets are not sorted alphabetically for costumes, sounds,sprites. When clicked it sorts them correctly, and sorts the correct assets depending on which target pane you're on.
- Sort alphabetically is grayed out if the assets in the target pane are already sorted
- If sort alphabetically is applied, a button to restore the original order replaces the normal undo button
- when the `restore order` button is clicked, the order returns to what it was before the sort was applied
- if `restore order` is clicked and new assets have been created between the original sort and the restore order, those assets remain in the same order but are placed at the end of the asset array
- if `restore order` button is clicked, and assets have been renamed, those assets still return to their original order. Renaming other assets than the ones on the target pane for the target that was reordered has no effect.

WorldLanguages:
> Also the addon should be renamed to "sort assets by name" since "sort by alphabetical order" implies the existence of other sorts

Have we considered adding the ability to sort numerically? That is, [1,3,10] instead of [1,10,3].

Jazza-231:
> > Also the addon should be renamed to "sort assets by name" since "sort by alphabetical order" implies the existence of other sorts
> 
> Have we considered adding the ability to sort numerically? That is, [1,3,10] instead of [1,10,3].

This addon does that. We chose this method because sorting numerically is most useful in every single case, no matter what.

Joeclinton1:
Is this still in development, I feel like it's done.

kpohogenic:
has compatibility with the existing 'Sprite Folders' addon been considered? what if sprites in one folder and not ordered next to eachother after sorting?

PR #7252, opened 2024-03-06
Add "export settings" button to unsupported-browser page

### LINKED ISSUE #7125: Unsupported browser page: add Export Settings button
### Why this would be helpful

Related to #6669
Part of MV3 roadmap: #6876

> This will allow users running old browser versions to export their settings before (and after) the extension goes away.

Nowadays, each user's settings is like a piece of art, that is hard to memorize, and basically impossible to replicate exactly if lost. It would be very rude if the Unsupported Browser page not only kicked you out from using the extension, but also made it impossible for users to export their selection of settings.

### My suggestion

The "unsupported browser" page will live in `chrome-extension://`, but the page will be a full-screen iframe of the current page (scratchaddons.com/unsupported-browser).

Clicking the new "Export Settings" / "View settings as file" buttons will run `parent.postMessage()`, giving control back to the extension.

If the user has no internet connection, they won't be able to export their settings. It's also a good opportunity to display a simple "unsupported browser" text in that case.

### Possible alternatives

There might be other ways to let users export settings if their browser is unsupported, but we need to quickly add one before we upgrade to Manifest V3.

### Additional context

_No response_

WorldLanguages:
Progress on #6876
Resolves #7125

### Changes

The unsupported-browser page is now a full-screen iframe. The user can now export and view their settings if their browser is unsupported.

### Reason for changes

We will be increasing our minimum browser version requirements very soon, and we don't want existing users to permanently lose access to their settings, which technically wouldn't be gone, but only developers would know how to access them if we changed nothing.

### Tests

You can emulate an unsupported browser by hardcoding `const checkIfUnsupported = () => true` on file background/handle-unsupported-version.js

- [x] Tested in old supported Chrome version
- [x] Tested in old supported Firefox version

### DIFF
--- a/webpages/error/unsupported-browser.js
+++ b/webpages/error/unsupported-browser.js
@@ -12,7 +12,7 @@ const IFRAME_ORIGIN = "https://scratchaddons.com";
 
 window.onmessage = async (e) => {
   if (e.origin !== IFRAME_ORIGIN) return;
-  
+
   const message = e.data;
   console.log("[SA] Message received by extension", message);
 
--- a/webpages/error/unsupported-browser.js
+++ b/webpages/error/unsupported-browser.js
@@ -1,0 +1,52 @@
+const iframe = document.querySelector("iframe");
+
+try {
+  const uiLanguage = chrome.i18n.getUILanguage();
+  const localeSlash = uiLanguage.startsWith("en") ? "" : `${uiLanguage.split("-")[0]}/`;
+  iframe.src = `https://scratchaddons.com/${localeSlash}unsupported-browser/`;
+} catch (err) {
+  iframe.src = "https://scratchaddons.com/unsupported-browser/";
+}
+
+const IFRAME_ORIGIN = "https://scratchaddons.com";
+
+window.onmessage = async (e) => {
+  if (e.origin !== IFRAME_ORIGIN) return;
+  
+  const message = e.data;
+  console.log("[SA] Message received by extension", message);
+
+  if (message.msgType) {
+    if (message.msgType === "getVersionInfo") {
+      // Reply back to the website with the extension version
+      iframe.contentWindow.postMessage({ versionInfo: chrome.runtime.getManifest().version }, IFRAME_ORIGIN);
+    }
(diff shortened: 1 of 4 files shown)


WorldLanguages:
@DNin01 As far as I can see, we only show the unsupported-browser page if the user tries to interact with the extension. We never open it automatically.

webpages/error/unsupported-browser.js
```diff
@@ -0,0 +1,52 @@
...
+      document.title = message.msgContent;
+    }
+
+    if (["exportSettingsAsFile", "openSettingsAsTab"].includes(message.msgType)) {
+      // See webpages/settings/index.js for main import/export settings code
+      const { serializeSettings } = await import("../settings/settings-utils.js");
+      const downloadBlob = (await import("../../libraries/common/cs/download-blob.js")).default;
+
+      // Export extension settings as a .json file
+      if (message.msgType === "exportSettingsAsFile") {
```
Samq64, line 35: On Chromium 79 I got an unexpected token error on this line. Firefox 85 and Chromium 104 both work as expected.

WorldLanguages: Do you have more info on the error? Possibly async/await not properly supported?

Samq64: It's actually on line 35 of `minify-settings.js`, not this file. I haven't checked what's causing it yet.

Samq64: https://github.com/WorldLanguages/ScratchAddons/blob/029665784935ba0681d9e2efa0658f6ada2644ab/libraries/common/minify-settings.js#L35

Samq64: Chrome 80 is also unaffected so this isn't a problem.

Samq64 (approved):
Tested on Chromium 80 and Firefox 85. The settings were imported successfully into up-to-date browser versions. Firefox does block viewing the settings the first time but it works after allowing the popup.

WorldLanguages:
Issue #7125 also mentioned offline support for this page, but I didn't implement it.

PR #7257, opened 2024-03-09
Change minimum browser requirements in preparation for MV3

WorldLanguages:
See #6876

### Changes

Increases the minimum browser requirements.

- Chrome 96: released in November 2021, it supports `declarativeNetRequestWithHostAccess` and most MV3 features.
- Firefox 109: released in January 2023, it's the first stable Firefox version to support MV3.

Page https://scratchaddons.com/unsupported-browser will need to be updated.

### Reason for changes

See #6876 for the reasoning on increasing the requirements before the actual MV3 upgrade.
By the way, #7252 was recently merged, which allows users to export their extension settings even if their browser is unsupported.

### Tests

Working on it

### DIFF
--- a/background/handle-unsupported-version.js
+++ b/background/handle-unsupported-version.js
@@ -6,7 +6,11 @@ const checkIfUnsupported = () => {
   };
 
   let { browser, version } = getVersion();
-  return (browser === "Chrome" && version < 80) || (browser === "Firefox" && version < 86);
+  const MIN_CHROME_VERSION = 96;
+  const MIN_FIREFOX_VERSION = 109;
+  return (
+    (browser === "Chrome" && version < MIN_CHROME_VERSION) || (browser === "Firefox" && version < MIN_FIREFOX_VERSION)
+  );
 };
 
 if (checkIfUnsupported()) {


WorldLanguages:
Curious fact: Chrome 96 does support Windows 7, so we're not dropping that OS just yet 😅

DNin01:
I'll do some research on the minimum version for the Firefox extension APIs.

Samq64:
Firefox could be bumped it to 115 ESR (July 2023) which is still supported and the last version to run on Windows 7 and MacOS 10.12.

DNin01 (commented):
Wow are there a lot of things to look at, but I'll at least say, as I've said before, the `storage.session` API is supported in Firefox 115+, so if we were to use that, we'd need to bump the minimum version.

WorldLanguages:
Chrome 96 also lacks that session storage API, @DNin01 
It's only available since Chrome 102

WorldLanguages:
Note that Chrome 96 and Firefox 109 do support, at least, basic MV3 features, so it would be OK to bump the requirements again in the actual MV3 upgrade if needed.

Remember that our goals are:
- Users should not get stuck on the latest supported version for their browser - instead, they should get locked out.
- If the browser is unsupported, users should be able to export their extension settings.

DNin01:
> * Users should not get stuck on the latest supported version for their browser - instead, they should get locked out.

I'll remind us that this done is for security reasons (we don't want to be unable to patch security bugs).

WorldLanguages:
We also want to avoid bug reports from unsupported browsers. Volunteer time is better used by actual bugs.

WorldLanguages:
Website PR: https://github.com/ScratchAddons/website-v2/pull/424

Samq64 (approved):
We can always increase the minimum requirements further later if we need to.

DNin01:
Should we write a post about this on the blog somewhere or… just the changelog, or…?

WorldLanguages:
I guess we should list this in the changelog

PR #7258, opened 2024-03-09
`custom-zoom` autohide: Use CSS `:hover` instead of `getBoundingClientRect`

mybearworld:
### Changes

Currently, `custom-zoom` detects if the cursor is hovering over the area with `getBoundingClientRect` and `mouseover` event. This PR makes it just use the `:hover` selector.

### Reason for changes

Removes most of the logic for this. It's also a bit more robust when the window is being resized.

### Tests

Tested in Edge.


Waakul:
No one's as stupid to not use ":hover" when needed, It's probably done for a reason.

Jazza-231:
> No one's as stupid to not use ":hover" when needed, It's probably done for a reason.

1. This is an entirely needed comment.
2. It's generally not a good idea to call anyone stupid.
3. People make oversights all the time
4. Please try to be more respectful overall, I believe this is not the first time something like this has happened before

Waakul:
> > No one's as stupid to not use ":hover" when needed, It's probably done for a reason.
> 
> 1. This is an entirely needed comment.
> 2. It's generally not a good idea to call anyone stupid.
> 3. People make oversights all the time
> 4. Please try to be more respectful overall, I believe this is not the first time something like this has happened before

i didnt call anyone stupid i said "no one is as stupid to do it without a reason"

mybearworld:
> No one's as stupid to not use ":hover" when needed, It's probably done for a reason. EDIT: woops!, it seems you made that addon! sorry.

I didn't make that addon, and I don't think the person that originally made it is stupid.
The way this addon worked before was entirely reasonable; I had to make some changes in order to get `:hover` to actually work in this context.

Waakul:
> > No one's as stupid to not use ":hover" when needed, It's probably done for a reason. EDIT: woops!, it seems you made that addon! sorry.
> 
> I didn't make that addon, and I don't think the person that originally made it is stupid. The way this addon worked before was entirely reasonable; I had to make some changes in order to get `:hover` to actually work in this context.

i didnt call him stupid, i said he most likely did it for some reasons.

mybearworld:
> i didnt call him stupid, i said he most likely did it for some reasons. You'll need to do lots of testing to clarify everything works (cause using :hover might break other things)

Yeah, the reasons seem to be:
- Moving the zoom area before the SVG makes it go behind the SVG, but is also necessary in order to select it with CSS.
- Having `:hover` work also means that `pointer-events` can't be set to `none`, but it also has to be because otherwise, the zoom area will receive all clicks in its area, and the zoom buttons will stop working.

And I have addressed them.

addons/custom-zoom/style.css
```diff
@@ -4,11 +4,11 @@
   right: 0;
   width: 60px;
   height: 148px;
-  pointer-events: none;
+  z-index: 21;
```
WorldLanguages, line 7: Can you explain the choice of 21?

mybearworld: Not sure why I set it to that, it works perfectly fine if it's lower than that.

```suggestion
  z-index: 1;
```

WorldLanguages:
We usually try to avoid mousemove events on the document, so this is also related to performance.

mxmou:
This change makes it impossible to scroll or zoom using the mouse wheel when the mouse is above the zoom buttons.

mybearworld:
> This change makes it impossible to scroll or zoom using the mouse wheel when the mouse is above the zoom buttons.

Should be fixed with https://github.com/ScratchAddons/ScratchAddons/pull/7258/commits/83e3feeffaf9d98348ea6777927fcf59ffbf5bd9

mxmou (approved):

PR #7259, opened 2024-03-10
New addon: Default to bitmap mode

### LINKED ISSUE #7201: New addon: default new costumes to bitmap, instead of vector
### Why this would be helpful

Saves a click each time a new costume is created

### How this addon works

> Could you make an option to default to bitmap when creating a new costume?

This was suggested twice on Discord, and also on the feedback form.

### Possible alternatives

_No response_

### Additional context

_No response_

mybearworld:
Resolves #7201

### Changes

When the empty vector is added, an empty bitmap is added instead.

### Reason for changes

Saves a click for people who often use bitmap

### Tests

Tested in Edge.

### DIFF
--- a/addons/default-bitmap/userscript.js
+++ b/addons/default-bitmap/userscript.js
@@ -5,7 +5,7 @@ export default async function ({ addon, console }) {
     if (md5 === "cd21514d0531fdffb22204e0ec5ed84a.svg" && !addon.self.disabled) {
       return _addCostume.call(
         vm,
-        "8ac8569502a1fc1a33aad0ae5786f20b.png",
+        emptyBitmap.md5,
         {
           ...emptyBitmap,
           name: costume.name,
--- a/addons/default-bitmap/userscript.js
+++ b/addons/default-bitmap/userscript.js
@@ -1,0 +1,27 @@
+export default async function ({ addon, console }) {
+  const vm = addon.tab.traps.vm;
+  const _addCostume = vm.addCostume;
+  vm.addCostume = (md5, costume, ...args) => {
+    if (md5 === "cd21514d0531fdffb22204e0ec5ed84a.svg" && !addon.self.disabled) {
+      return _addCostume.call(
+        vm,
+        "8ac8569502a1fc1a33aad0ae5786f20b.png",
+        {
+          ...emptyBitmap,
+          name: costume.name,
+        },
+        ...args
+      );
+    }
+    return _addCostume.call(vm, md5, costume, ...args);
+  };
+}
+
+const emptyBitmap = {
+  md5: "8ac8569502a1fc1a33aad0ae5786f20b.png",
+  assetId: "8ac8569502a1fc1a33aad0ae5786f20b",
+  bitmapResolution: 2,
+  rotationCenterX: 1,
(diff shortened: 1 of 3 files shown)


CST1229:
I think this might also trigger when importing an empty vector? Though that's extremely minor and something no one will ever do unless they're intentionally triggering the bug.

mybearworld:
> I think this might also trigger when importing an empty vector? Though that's an extremely minor bug and something no one will ever do unless they're intentionally triggering the bug.

I'm not sure why, but when uploading the empty SVG yourself, the hash is `4b7a854341abd530a6780bfb6a68627c.svg`, which doesn't trigger the replacing.

WorldLanguages:
> I think this might also trigger when importing an empty vector?

Even if it did, it's not a big deal. This looks like the simplest way to implement the addon.

addons/default-bitmap/userscript.js
```diff
@@ -0,0 +1,27 @@
+export default async function ({ addon, console }) {
+  const vm = addon.tab.traps.vm;
+  const _addCostume = vm.addCostume;
+  vm.addCostume = (md5, costume, ...args) => {
+    if (md5 === "cd21514d0531fdffb22204e0ec5ed84a.svg" && !addon.self.disabled) {
+      return _addCostume.call(
+        vm,
```
WorldLanguages, line 7: We usually just write `this` here, but in this case it's equivalent.
For a second, I thought the VM was a parameter, though. It's a bit more confusing.

Also, if you want to follow conventions, the `_addCostume` variable should be named `originalAddCostume`.

WorldLanguages:
@mybearworld Thanks for the PR, I'll quickly review it. Prioritizing a bit, as this was directly requested through user feedback multiple times, and it's a quite simple implementation.

WorldLanguages:
Should we extend this so that it also defaults to bitmap when creating a new sprite/backdrop with the "paint" button?

mybearworld:
> Should we extend this so that it also defaults to bitmap when creating a new sprite/backdrop with the "paint" button?

That would make sense.

addons/default-bitmap/userscript.js
mxmou: > Also, if you want to follow conventions, the `_addCostume` variable should be named `originalAddCostume`.

There's no convention - some addons use the `original` prefix, some use `old` and a few use an underscore.

WorldLanguages: yeah, it's not strictly enforced

mybearworld:
When trapping `vm.addSprite` to make it add the bitmap instead of the vector, the bitmap gets zoomed in for some reason:
image

WorldLanguages:
> When trapping `vm.addSprite` to make it add the bitmap instead of the vector, the bitmap gets zoomed in for some reason

Maybe we should wait until the user goes to the Costumes tab, and just emulate a button click on the "convert to bitmap" button 🤔

mybearworld:
That's not an instant conversion, though. It's noticable.

WorldLanguages:
> That's not an instant conversion, though. It's noticable.

Hmm, good point.

medlans:
This was closed too.

PR #7261, opened 2024-03-10
`custom-menu-bar`: change addon name, added word "editor"

Samq64:
### Changes

Mentions the word editor in "Customizable menu bar".

### Reason for changes

Someone mentioned the navigation and meubar addons could be confused with each other without reading the description.

### Tests

Tested on Chromium.


WorldLanguages (approved):
I don't think this was ever suggested in the original PR. This name is superior I think.

Issue #6914, opened 2023-11-30
`forum-cooldown` addon (occasionally) doesn't display cooldown


OneShot-Niko:
### Description
EDIT: Originally I thought this was because of the mobile forums being removed, however it appears I was wrong as it only occasionally doesn't occur. Apologies for not testing thrice instead of twice.

The forum cooldown display addon occasionally does not display cooldown.

### Steps to reproduce

1. Enable the addon
2. Post something (Reccomended to do this in the official scratchblocks testing topic)
3. Try to instantly post something else, you'll get the small note but the cooldown will not display itself.

### Demonstration(s) and additional context

None ((for now))

### Environment

Browser: Firefox/FF 120.0 32 bit
Scratch Addons version: v1.35.1 (stable/public)
Operating system: Windows 10

Secret-chest:
Still works for me, but not all the time.

OneShot-Niko:
Trying to replicate this and just can't anymore

This might've been just an addon conflict that got resolved but I honestly have no idea since I just cannot replicate this anymore, so i'll just close

PR #7273, opened 2024-03-16
Make `forum-post-cooldown` use `auto-follow-topic`'s logic

### LINKED ISSUE #6926: `forum-post-countdown` also starts the countdown on other errors
### Description

When any error occurs before posting, `forum-post-countdown` starts the 60 second countdown.

### Steps to reproduce

1. Try to post an empty post or `assets.scratch.mit.edu/get_image`

_No response_

### Environment

Browser: Edge 119
Operating system: Windows

### LINKED ISSUE #7269: `forum-post-countdown` is inaccurate
Reported by a forumer on my Scratch profile:

> It never displays the timer until I trigger the 60 second alert, and it starts over at 60 when I TRIGGER the 60 second rule, not when I last made my post.

mybearworld:
Resolves #6926
Probably resolves #7269
Resolves (possibly) #6914 (not sure about this one, I can't seem to replicate that issue even on stable)

### Changes

Turn `auto-follow-topic` into a module, and make `forum-post-cooldown` use it.

### Reason for changes

Improving `forum-post-cooldown`.

### Tests

Tested in Edge.

### DIFF
--- a/addons/auto-follow-topics/module.js
+++ b/addons/auto-follow-topics/module.js
@@ -1,3 +1,4 @@
+// This module should run after page load - only use this with `runAtComplete:true`
 const getTopics = () => JSON.parse(sessionStorage.getItem("auto-follow-topics") || "[]");
 const setTopics = (topics) => sessionStorage.setItem("auto-follow-topics", JSON.stringify(topics));
 const postError = () => Boolean(document.querySelector(".errorlist"));
--- a/addons/auto-follow-topics/module.js
+++ b/addons/auto-follow-topics/module.js
@@ -1,0 +1,30 @@
+const getTopics = () => JSON.parse(sessionStorage.getItem("auto-follow-topics") || "[]");
+const setTopics = (topics) => sessionStorage.setItem("auto-follow-topics", JSON.stringify(topics));
+const postError = () => Boolean(document.querySelector(".errorlist"));
+const topicID = location.href.split("/")[5];
+
+document.querySelector("[name=AddPostForm]")?.addEventListener("click", (event) => {
+  const ids = getTopics();
+  ids.push(topicID);
+  setTopics(ids);
+  onPostCallbacks.forEach((callback) => callback());
+});
+
+const topics = getTopics();
+let success = false;
+if (topics.includes(topicID) && !postError()) {
+  success = true;
+  setTopics(topics.filter((topic) => topic !== topicID));
+}
+
+const onPostCallbacks = [];
+export const onPost = (callback) => {
+  onPostCallbacks.push(callback);
+};
+
+export const onSuccessfulPost = (callback) => {
+  if (success) callback(topicID);
+};
+export const onNonSuccessfulPost = (callback) => {
+  if (!success) callback(topicID);
+};
--- a/addons/auto-follow-topics/module.js
+++ b/addons/auto-follow-topics/module.js
@@ -12,7 +12,8 @@ document.querySelector("[name=AddPostForm]")?.addEventListener("click", (event) 
 
 const topics = getTopics();
 let success = false;
-if (topics.includes(topicID) && !postError()) {
+const hasTopic = topics.includes(topicID);
+if (hasTopic && !postError()) {
   success = true;
   setTopics(topics.filter((topic) => topic !== topicID));
 }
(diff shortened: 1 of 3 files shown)


addons/auto-follow-topics/module.js
```diff
@@ -0,0 +1,30 @@
+const getTopics = () => JSON.parse(sessionStorage.getItem("auto-follow-topics") || "[]");
```
WorldLanguages, line 1:
```suggestion
// This module should run after page load - only use this with `runAtComplete:true`
const getTopics = () => JSON.parse(sessionStorage.getItem("auto-follow-topics") || "[]");
```

addons/auto-follow-topics/module.js
```diff
@@ -0,0 +1,30 @@
...
+
+const onPostCallbacks = [];
+export const onPost = (callback) => {
+  onPostCallbacks.push(callback);
+};
+
+export const onSuccessfulPost = (callback) => {
+  if (success) callback(topicID);
+};
+export const onNonSuccessfulPost = (callback) => {
```
WorldLanguages, line 28: What does a "non successful post" actually mean? Why are you clearing storage (in forum-post-countdown) in this case? 🤔

WorldLanguages: The variable `success` being false does not mean there was a post error. It only means there was not a successful post. But visiting a different topic shouldn't clear the countdown, right?

mybearworld: Thanks, you're right. I fixed that in https://github.com/ScratchAddons/ScratchAddons/pull/7273/commits/4d3a08b6574be21497d62eddc757a6c189f215ba.

addons/auto-follow-topics/module.js
```diff
@@ -0,0 +1,32 @@
...
+const onPostCallbacks = [];
+export const onPost = (callback) => {
+  onPostCallbacks.push(callback);
+};
+
+export const onSuccessfulPost = (callback) => {
+  if (hasTopic && success) callback(topicID);
+};
+export const onNonSuccessfulPost = (callback) => {
+  if (hasTopic && !success) callback(topicID);
```
WorldLanguages, line 31: This still isn't convincing to me. Shouldn't this check for `postError()`?

mybearworld: `postError()` is called when setting the `success` variable.

WorldLanguages: `success` being true does guarantee there wasn't any post errors. However, `success` being false tells us nothing about whether there was a post error or not. Again, how do we want to define a "non successful post"?

WorldLanguages: The variable `success` is true, if and only if, `hasTopic && !postError()` is true
That is, `!success` is true if either `hasTopic` if false, `postError()` is true, or both at the same time

To see if there's a non-successful post, you evaluate `hasTopic && !success`.
But as said above, this is equivalent to `hasTopic && (!hasTopic || postError())`
Of course, `hasTopic` cannot be true and false at the same time, so the condition you end up checking is `hasTopic && postError()`

So I guess your code is equivalent to what I would expect (I would need to check again), but I found it quite confusing!

mybearworld: Should I just leave out the `success` variable in favor of calling `postError` directly?

WorldLanguages:
```suggestion
export const onNonSuccessfulPost = (callback) => {
  if (postError()) callback(topicID);
```

I guess this would work as you expect?

addons/auto-follow-topics/module.js
Samq64, line 1: Maybe it would make more sense for this to be in `/libraries/common/cs`?

Samq64 (commented):
This would also be useful for #4716.

WorldLanguages:
@Samq64 You may self-assign if you want.

addons/forum-post-countdown/userscript.js
```diff
@@ -1,10 +1,13 @@
...
-  submitButton.addEventListener("click", () => {
-    if (!localStorage.getItem("sa-forum-post-countdown")) {
-      localStorage.setItem("sa-forum-post-countdown", Date.now());
-    }
+  onPost(() => {
+    localStorage.setItem("sa-forum-post-countdown", Date.now());
+  });
+  onNonSuccessfulPost(() => {
+    localStorage.removeItem("sa-forum-post-countdown");
   });
```
mxmou, line 11:
```suggestion
  onSuccessfulPost(() => {
    localStorage.setItem("sa-forum-post-countdown", Date.now());
  });
```
This would make the most sense and is probably what the person who reported #7269 wanted. There's no reason to remove the cooldown when a post fails.

PR #7214, opened 2024-02-23
Redesign

Waakul:
## Note
Please don't judge through screenshots, the screenshots are completely color and quality corrupted, Please try the pr to see the changes instead.

Discussion #7215 

### Changes
- Recolor for both dark and light modes
- light Box-shadows instead of borders
- Expand width on focusing the search bar in the settings page
- Nicer looking tabs
- Tab's width adjust to fill availible space
- Center aligned UI instead of left alignment
- Improved shadows
- A very very few bug fixes from the previous UI (I forgot what they are)
- Things that are supposed to be in containers that look bad when not now in containers
- A puzzle icon instead of the settings icon cause it's more meaningful
- Consistent Design Language
- A ton of minor redesigns i forgot, check changes ig

### Reason for changes
Overall Modern Look Without Ruining Accessibility

### Tests
Tested In: Microsoft Edge, Firefox

PR #7279, opened 2024-03-19
Category tab margin for better accessibilty

### LINKED ISSUE #7278: Accessibility issues in category tabs
### Why this would be helpful

image
As you see here, the background of the tab is hard to distinguish from the background and the search toolbar background.

### My suggestion

Solution:
image
use something like this, margin to distinguish and border radius so it's easy to follow the border and also for consistent design lang

### Possible alternatives

i dont have any other ideas

### Additional context

image

Waakul:
Resolves #7278 

### Changes
image

### Reason for changes
Better accessibilty as mentioned in #7278 

### Tests
Microsoft Edge (Chromium)

### DIFF
--- a/webpages/settings/components/category-selector.html
+++ b/webpages/settings/components/category-selector.html
@@ -28,6 +28,8 @@
     cursor: pointer;
     display: flex;
     align-items: center;
+    margin: 10px;
+    border-radius: 4px;
   }
   .category.category-small {
     padding: 15px 20px 15px 20px;
--- a/webpages/styles/components/categories.css
+++ b/webpages/styles/components/categories.css
@@ -8,5 +8,5 @@
   overflow-y: auto;
   scrollbar-width: thin;
   scrollbar-color: var(--gray-text) transparent;
-  padding: 20px 0;
+  padding: 5px 0 20px 0;
 }


Samq64:
I would suggest adding the sidebar changes from #7214 since it's unlikely to be merged any time soon.

> Should i make the indicator on the side full-height?

Maybe, but I think it also looks fine the way it is now.

Secret-chest:
Can we just make it wider? Adding a margin reduces the click area, read Fitts' law.

BroJac5246:
> Adding a margin reduces the click area, read Fitts' law.

However, your cursor generally isn't in the leftmost part of the tab, so it doesn't have to move any further. Also, the margin only slightly decreases the clickable area and therefore shouldn't have a large impact on clickability. Good UIs have find a balance between usability and design and this change will be worth it IMO.

Secret-chest:
> However, your cursor generally isn't in the leftmost part of the tab, so it doesn't have to move any further. Also, the margin only slightly decreases the clickable area and therefore shouldn't have a large impact on clickability. Good UIs have find a balance between usability and design and this change will be worth it IMO.

Can't you just make it wider?

BroJac5246:
> > However, your cursor generally isn't in the leftmost part of the tab, so it doesn't have to move any further. Also, the margin only slightly decreases the clickable area and therefore shouldn't have a large impact on clickability. Good UIs have find a balance between usability and design and this change will be worth it IMO.
> 
> Can't you just make it wider?

You could, but it's already plenty wide.

webpages/settings/components/category-selector.html
```diff
@@ -59,17 +61,19 @@
...
+    background: var(--sel-darken);
+  }
   .category::before {
     content: "";
     transition: all 0.2s ease;
     display: block;
     width: 4px;
-    border-radius: 0 4px 4px 0;
-    height: 40px;
+    border-radius: 4px 0 0 4px;
```
Samq64, line 75: The RTL `border-radius` will also need to be reversed.

Samq64 (approved):
You're right, the full height bar does look a lot better. Could it be applied to the subcategories too?

Samq64:
You're right, the full height bar does look a lot better. Could it also be applied to the subcategories?

webpages/settings/components/category-selector.html
Waakul: oh ye

Samq64: Line 84 should be changed to `0 4px 4px 0`, I just couldn't add a review on the unchanged line.

Waakul: > Line 84 should be changed to `0 4px 4px 0`, I just couldn't add a review on the unchanged line.

resolved, sorry for a long time

Waakul:
@Samq64 @mxmou just a reminder to please review this.

Samq64 (dismissed):
Seems fine although it does take up a lot of space on smaller screens.

Waakul:
reduce margin a bit? (vote 👍 or 👎 )

webpages/settings/components/category-selector.html
```diff
@@ -28,6 +28,8 @@
     cursor: pointer;
     display: flex;
     align-items: center;
+    margin: 10px;
```
mxmou, line 31:
```suggestion
    margin: 10px;
    margin-bottom: 0;
```
This will fix the issue with small categories: https://github.com/ScratchAddons/ScratchAddons/pull/7279#issuecomment-2051613850

mxmou (commented):
I don't like the amount of space this adds - the categories already have a lot of padding. Moving the selection indicator away from the edge instead of adding a margin might be a better solution:
image

Another option is to keep the margin but reduce the padding by 5px on each side - the result looks similar:
image

Samq64:
Too much padding is more of a problem now that the project player addons are a top-level category. Feel free to discuss in #7278 if you are still interested.

PR #7285, opened 2024-03-21
Expand the clickable area of the theme button like the settings button

Waakul:
### Changes
More clickable area of the theme button in the settings like the settings button in the popup as currently you need to click exactly on the image by:

- Enclosed it in a div with padding which has the on click event

### Reason for changes
So that it's more easy to click on the theme button and we don't need to click exactly on the image.

### Tests
Microsoft Edge 122.0.2365.92 (Official build) (64-bit)

### DIFF
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -20,6 +20,13 @@ body {
   display: flex;
   height: 60px;
 }
+.header-button {
+  right: 0;
+  position: fixed;
+  cursor: pointer;
+  padding: 17.5px;
+  line-height: 0;
+}
 #title-text,
 #settings {
   font-size: 18px;
@@ -37,18 +44,10 @@ body {
   margin-inline-end: 20px;
   vertical-align: middle;
 }
-#settings {
-  padding: 0 20px;
-  cursor: pointer;
-  display: flex;
-  align-items: center;
-  justify-content: center;
-}
-#settings > img {
+#settings-icon {
   width: 24px;
   height: 24px;
 }
-
 #popups {
   background-color: var(--content-background);
   width: 100%;
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -28,7 +28,7 @@ body {
   line-height: 0;
 }
 #title-text,
-#settings {
+.header-button {
   font-size: 18px;
   font-weight: 400;
 }
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -24,7 +24,7 @@ body {
   right: 0;
   position: fixed;
   cursor: pointer;
-  padding: 17.5px;
+  padding: 18px;
   line-height: 0;
 }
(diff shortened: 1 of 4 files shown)


webpages/settings/style.css
```diff
@@ -132,9 +132,13 @@ h1 {
...
 .theme-switch {
   height: 25px;
-  right: 25px;
+}
+#theme-div {
+  right: 0;
   position: fixed;
   cursor: pointer;
+  padding: 16px;
+  transition: 0.2s ease;
```
Samq64, line 141: There isn't anything to transition to right now although a hover background would be nice.

Waakul: Ah!, i copy pasted that code from #7214 which had a hover effect

Waakul: I'll remove that for now, and add it back when i create a pr for hover background (which will first require this pr to be merged)

Samq64 (approved):

webpages/settings/style.css
```diff
@@ -132,9 +132,12 @@ h1 {
 
 .theme-switch {
   height: 25px;
-  right: 25px;
+}
+#theme-div {
```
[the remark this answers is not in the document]
Waakul: Ok

webpages/settings/style.css
```diff
@@ -132,9 +132,12 @@ h1 {
...
+  right: 0;
   position: fixed;
   cursor: pointer;
+  padding: 16px;
```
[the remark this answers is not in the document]
Waakul: For the padding one, I think it's better of looking as a square like the settings button in the popup. The line height somehow makes it need to have 17.5px as the padding so I'll do it and clear that.

webpages/popup/style.css
```diff
@@ -20,8 +20,15 @@ body {
   display: flex;
   height: 60px;
 }
+.header-button {
+  right: 0;
+  position: fixed;
```
mxmou, line 25:
```suggestion
```
This isn't necessary because the title has `flex-grow: 1`.

webpages/popup/style.css
```diff
@@ -20,8 +20,15 @@ body {
...
+  cursor: pointer;
+  padding: 17.5px;
```
mxmou, line 27:
```suggestion
  padding: 18px;
```
The settings icon in the popup is actually 1px smaller than the theme icon on the settings page.

mxmou (changes requested):

webpages/popup/style.css
Waakul: It was already there, dunno if it was there for some reason...

mxmou: It's needed on the settings page but not in the popup.

Waakul: oh ok

mxmou (approved):

PR #7290, opened 2024-03-23
Allow links to be clickable in the messages page

### LINKED ISSUE #5334: Add support for "More Links" addons for the messages page

If you quickly want to access links from a comment or response from someone in the messages page, you should be able to click on the link directly, so you don't have to click on the comment, get redirected to the project page, and THEN click the link.

image

### What this change does

Allow clicking links through messages page

### Possible alternatives

I don't know

No need

jazzathedev:
Resolves #5334

### Changes

Adds a few lines to allow links to be linkified when in the messages tab of scratch. small change

### Reason for changes

Consistency and makes sense

### Tests

works fine on brave

### DIFF
--- a/addons/more-links/userscript.js
+++ b/addons/more-links/userscript.js
@@ -52,6 +52,12 @@ export default async function ({ addon, console }) {
       linkifyTextNode(desc);
       break;
     }
+
+    case "messages":
+      while (true) {
+        const message = await addon.tab.waitForElement(".comment-text p", { markAsSeen: true });
+        linkifyTextNode(message);
+      }
   }
 
   (async () => {
--- a/addons/more-links/userscript.js
+++ b/addons/more-links/userscript.js
@@ -58,12 +58,21 @@ export default async function ({ addon, console }) {
         const message = await addon.tab.waitForElement(".comment-text p", { markAsSeen: true });
 
         if (message.textContent.length === 250) {
+          console.log(message);
           // The message is truncated (unless it happens to be ecxactly 250 characters long), and so links may break. We now fetch the rest of the text for link making
 
           const username = await addon.auth.fetchUsername();
 
           const messageInfo = message.parentElement.parentElement.parentElement.firstChild;
           const link = messageInfo.querySelector("a:not(.social-messages-profile-link)").href;
+
+          async function getContentFromResponse(response, message) {
+            const blob = await response.blob();
+            const usableResponse = JSON.parse(await blob.text());
+
+            // Yes, using innerHTML here, this is so the browser can handle things like &apos; for a comma
+            message.innerHTML = usableResponse.content;
+          }
 
           if (link.includes("projects")) {
             const regex = /\/projects\/(\d+)\/#comments-(\d+)/;
@@ -75,12 +84,18 @@ export default async function ({ addon, console }) {
             const response = await fetch(
               `https://api.scratch.mit.edu/users/${username}/projects/${projectId}/comments/${commentId}`
             );
-            const blob = await response.blob();
-            const usableResponse = JSON.parse(await blob.text());
-            console.log(usableResponse);
 
-            // Yes, using innerHTML here, this is so the browser can handle things like &apos; for a comma
-            message.innerHTML = usableResponse.content;
+            await getContentFromResponse(response, message);
+          } else if (link.includes("studios")) {
+            const regex = /\/studios\/(\d+)\/comments\/#comments-(\d+)/;
+            const match = link.match(regex);
+
+            const studioId = match[1];
+            const commentId = match[2];
+
+            const response = await fetch(`https://api.scratch.mit.edu/studios/${studioId}/comments/${commentId}`);
+
+            await getContentFromResponse(response, message);
           }
         }
         linkifyTextNode(message);
--- a/addons/more-links/userscript.js
+++ b/addons/more-links/userscript.js
@@ -56,6 +56,33 @@ export default async function ({ addon, console }) {
     case "messages":
       while (true) {
         const message = await addon.tab.waitForElement(".comment-text p", { markAsSeen: true });
+
+        if (message.textContent.length === 250) {
+          // The message is truncated (unless it happens to be ecxactly 250 characters long), and so links may break. We now fetch the rest of the text for link making
+
+          const username = await addon.auth.fetchUsername();
+
+          const messageInfo = message.parentElement.parentElement.parentElement.firstChild;
+          const link = messageInfo.querySelector("a:not(.social-messages-profile-link)").href;
+
+          if (link.includes("projects")) {
+            const regex = /\/projects\/(\d+)\/#comments-(\d+)/;
+            const match = link.match(regex);
+
+            const projectId = match[1];
+            const commentId = match[2];
+
+            const response = await fetch(
+              `https://api.scratch.mit.edu/users/${username}/projects/${projectId}/comments/${commentId}`
+            );
+            const blob = await response.blob();
+            const usableResponse = JSON.parse(await blob.text());
+            console.log(usableResponse);
+
+            // Yes, using innerHTML here, this is so the browser can handle things like &apos; for a comma
+            message.innerHTML = usableResponse.content;
+          }
+        }
         linkifyTextNode(message);
       }
   }
--- a/addons/more-links/addon.json
+++ b/addons/more-links/addon.json
@@ -9,7 +9,12 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["projects", "https://scratch.mit.edu/studios/*", "https://scratch.mit.edu/users/*"]
+      "matches": [
+        "projects",
+        "https://scratch.mit.edu/studios/*",
+        "https://scratch.mit.edu/users/*",
+        "https://scratch.mit.edu/messages/"
+      ]
     }
   ],
   "versionAdded": "1.0.0",


Jazza-231:
It fits with this addon best, but the addon description does say it makes links pointing outside of scratch clickable, yet this is obviously in scratch. big enough change to warrent a description change? a setting maybe?

mxmou:
This might create broken links if the message is truncated - see the discussion in #5334.

Jazza-231:
> This might create broken links if the message is truncated - see the discussion in #5334.

Hmm I didn't even know messages did that... Is the truncation done by the API or by the front-end? If it's not API could we detect the length of the message as mentioned in that issue and then make another request to the API for that message and check the full link? While we're doing that, we could just replace all truncated messages with a see more button or similar

mxmou:
> > This might create broken links if the message is truncated - see the discussion in #5334.
> 
> Hmm I didn't even know messages did that... Is the truncation done by the API or by the front-end? If it's not API could we detect the length of the message as mentioned in that issue and then make another request to the API for that message and check the full link? While we're doing that, we could just add to all truncated messages a see more button or similar

They're truncated on the server side but only when using the messages API. You could use the comments API to get the full comment.

Jazza-231:
> > > This might create broken links if the message is truncated - see the discussion in #5334.
> > 
> > Hmm I didn't even know messages did that... Is the truncation done by the API or by the front-end? If it's not API could we detect the length of the message as mentioned in that issue and then make another request to the API for that message and check the full link? While we're doing that, we could just add to all truncated messages a see more button or similar
> 
> They're truncated on the server side but only when using the messages API. You could use the comments API to get the full comment.

Ok sweet, I haven't looked much into the APIs but I'm assuming this is a feasible approach? Not sure how you request a certain comment, I'm guessing by using the comment id, but I'll have to look into it. Thanks for bringing this to my attention and I'll work on this tmr

Jazza-231:
as far as i know there is not an easy to fetch profile comments, so thats why im doing this only for projects for now. please tell me if there is a nice way to do it for profiles

addons/more-links/userscript.js
```diff
@@ -52,6 +52,54 @@ export default async function ({ addon, console }) {
...
+          const messageInfo = message.parentElement.parentElement.parentElement.firstChild;
+          const link = messageInfo.querySelector("a:not(.social-messages-profile-link)").href;
+
+          async function getContentFromResponse(response, message) {
+            const blob = await response.blob();
+            const usableResponse = JSON.parse(await blob.text());
+
+            // Yes, using innerHTML here, this is so the browser can handle things like &apos; for a comma
+            message.innerHTML = usableResponse.content;
+          }
```
RedGuy12, line 75: `await response.text()` is a thing

also it's not us blocking `.innerHTML`. it's the chrome web store. i imagine they won't allow this cause theoretically what is the API returns some script tag? that's not feasible at all but think of their perspective.

i think there's a function somewhere in the code that sanitizes a string before putting it into innerHTML

RedGuy12:
imo it's a better solution to just not linkify trailing links if the comment length is 20

WorldLanguages:
> imo it's a better solution to just not linkify trailing links if the comment length is 250

Correct, this is the way! See the most recent comments at issue #5334

WorldLanguages:
It's not really worth it to try to get the full comment just to linkify it. If the link got truncated we unfortunately won't be able to linkify it.

addons/more-links/userscript.js
Jazza-231: > `await response.text()` is a thing
> 
> also it's not us blocking `.innerHTML`. it's the chrome web store. i imagine they won't allow this cause theoretically what is the API returns some script tag? that's not feasible at all but think of their perspective.
> 
> i think there's a function somewhere in the code that sanitizes a string before putting it into innerHTML

About the response.text I did that and unless I'm insane, it didn't give me what I needed

I'll look into that function

Jazza-231:
> It's not really worth it to try to get the full comment just to linkify it. If the link got truncated we unfortunately won't be able to linkify it.

Hmmmm I'd really like to linkify all links, how about using the whole comment I just make sure the whole link is in the comment and not adding anything that might come after? If length is an issue perhaps the text can be replaced with an anchor tag which just has the word "link" or something

WorldLanguages:
Well, if we're going to request the whole body of the comment, we should display it, right? It feels wasteful to only use the information for linkifying.

Jazza-231:
> Also, could displaying entire comments on the messages page with a button be its own addon? ✨

It could be.. Not sure how that would fit nicely with this addon. All I want is to make sure that the links work in as many cases as possible

WorldLanguages:
Maybe you could only request the full comment bodies only for comments that have their links cut out by the 250 limit

Jazza-231:
Can't reliably detect if it's supposed to be a link without the full body

DNin01:
> Can't reliably detect if it's supposed to be a link without the full body

Well, if a word is cut off, then we know it could be some sort of URL if it starts with `http://` or `https://`, right?

Jazza-231:
Not all links start with https

Jazza-231:
> Not all links start with http or https, this addon linkifies stuff like youtube.com, no https needed

DNin01:
> Not all links start with http or https, this addon linkifies stuff like youtube.com, no https needed

Oh. That would still detect cut-off URLs as long as enough of them are displayed.

Jazza-231:
> Oh. That would still detect cut-off URLs as long as enough of them are displayed.

Correct. But not always. I still think this is the best method.

I'm thinking of putting this into it's own addon, it's main purpose would be to show the whole message. It would then have a setting to make links clickable, and then also a setting for making mentions clickable. This seems like the best idea. Anyone have any problems with this idea?

addons/more-links/userscript.js
RedGuy12: I wouldn't neccessarily expect the CWS to know/believe that though
I believe we once had a release blocked by one of the stores because there we used innerHTML with a constant string, and the string had a commented-out <script> tag.

Jazza-231: I just had a thought... Could someone not write HTML code in a comment that's longer than 250 characters? This would immediately turn that into real html. What's stopping them from writing a script tag in their comment, the API doesn't need to be compromised at all

mybearworld: Oh and also, `await response.json()`?

Jazza-231: > Oh and also, `await response.json()`?

I swear this is the only way I could do it...I might be going insane, legitimately. Let me check that after my ocd shower

mybearworld: > I just had a thought... Could someone not write HTML code in a comment that's longer than 250 characters? This would immediately turn that into real html. What's stopping them from writing a script tag in their comment, the API doesn't need to be compromised at all

Scratch itself also just inserts the HTML without doing anything to it. The API escapes every comment when it's posted.

Screenshot of a comment saying ""
Screenshot of "content": "&lt;img&gt;"

WorldLanguages:
> Can't reliably detect if it's supposed to be a link without the full body

You can cover 99% of cases by checking for `https://` at the start of the link.

DNin01:
> > Oh. That would still detect cut-off URLs as long as enough of them are displayed.
> 
> Correct. But not always. I still think this is the best method.

As an unofficial tool, we don't have to be perfect.

DNin01:
> I'm thinking of putting this into it's own addon, it's main purpose would be to show the whole message. It would then have a setting to make links clickable, and then also a setting for making mentions clickable. This seems like the best idea. Anyone have any problems with this idea?

An addon to show the full message would be fine but options for linkification wouldn't really fit in it.

BroJac5246:
> > Can't reliably detect if it's supposed to be a link without the full body
> 
> You can cover 99% of cases by checking for `https://` at the start of the link.

Not really. Who actually does that except for when it's copied from the address bar? To the end user, it's the same regardless of whether `https://` is specified because that's added automatically when the link is clicked.

Jazza-231:
> > I'm thinking of putting this into it's own addon, it's main purpose would be to show the whole message. It would then have a setting to make links clickable, and then also a setting for making mentions clickable. This seems like the best idea. Anyone have any problems with this idea?
> 
> An addon to show the full message would be fine but options for linkification wouldn't really fit in it.

Hmm well above all else, I want to linkify every single link we can find. Thats the purpose of the addon, and if we're requesting the whole comment for this, it makes sense to use it...

Maybe this: 
1. In the more links addon, keep the current logic of showing the whole comment so we can linkify it. Who's really going to notice a larger message?
2. Make the clickable mentions a setting in this addon, as this is the most logical. Maybe rename this addon if required.

> Not really. Who actually does that except for when it's copied from the address bar? To the end user, it's the same regardless of whether `https://` is specified because that's added automatically when the link is clicked.

To be fair, a lot of them are going to be copied, but still, what if the htt is showing and then the rest is cut off? No linkifying, which is the most important aspect of this all to me. Fetching the whole comment still remains the best option in my eyes. Computationally it will make almost no different, and the scratch api handles a LOT more requests than this would generate

Samq64:
Why not just linkify what's visible and have a dedicated add-on to expand the messages or add it to full areas? It should automaticlly be compatible with more links assuming it runs first.

Jazza-231:
> Why not just linkify what's visible and have a dedicated add-on to expand the messages or add it to full areas? It should automaticlly be compatible with more links assuming it runs first.

Mmmmmmm yeah ok I can live with that, anyone else have an opinion?

BroJac5246:
> > Why not just linkify what's visible and have a dedicated add-on to expand the messages or add it to full areas? It should automaticlly be compatible with more links assuming it runs first.
> 
> Mmmmmmm yeah ok I can live with that, anyone else have an opinion?

I like that idea, it's just that someone has to create that other addon first.

Jazza-231:
> > Mmmmmmm yeah ok I can live with that, anyone else have an opinion?
> 
> I like that idea, it's just that someone has to create that other addon first.

that someone would be me

PR #7299, opened 2024-03-25
April Fools 2024 easter egg: Scratch Haddocks

WorldLanguages:
Right now only the top part of the popup has any changes at all

**(THIS SCREENSHOT IS NOW OUTDATED)**
image

PR #7308, opened 2024-03-28
Disable some addons for April Fools

WorldLanguages:
See #5848 from 2023

### DIFF
--- a/addons/hide-flyout/april-fools.js
+++ b/addons/hide-flyout/april-fools.js
@@ -1,9 +1,10 @@
-export const isScratchAprilFools24 = () => {
+export const isScratchAprilFools24 = async (redux) => {
   let val = false;
   const now = Date.now() / 1000;
-  if (!__scratchAddonsRedux.state && now < 1712059200 && now > 1711886400) val = true;
+  if (!redux.state && now < 1712059200 && now > 1711886400) val = true;
   else {
-    val = Object.entries(__scratchAddonsRedux.state.session.session.flags).some(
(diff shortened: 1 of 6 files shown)


WorldLanguages:
Vanilla Scratch with `"everything_is_totally_normal": true` is also quite buggy.

## THE FOLLOWING VIDEOS ARE VANILLA SCRATCH

brave_yEHrYIBidE
brave_obvRsDABWx

Samq64:
Nothing changes when setting the timestamp to the same one you said on #7299.

WorldLanguages:
> Nothing changes when setting the timestamp to the same one you said on #7299.

Changing the system time only takes effect in this PR if Redux is unavailable. Try making content-scripts/load-redux.js empty or broken.

addons/hide-flyout/april-fools.js
```diff
@@ -0,0 +1,10 @@
+export const isScratchAprilFools24 = async (redux) => {
+  let val = false;
```
WorldLanguages, line 2: Should have added
`ìf(april_three_or_later) return false;`

addons/hide-flyout/april-fools.js
```diff
@@ -0,0 +1,10 @@
...
+  const now = Date.now() / 1000;
+  if (!redux.state && now < 1712059200 && now > 1711886400) val = true;
+  else {
+    await redux.waitForState((state) => state.session?.session?.flags);
```
WorldLanguages, line 6: This didn't work 100% of the time on runAtComplete:false

PR #7314, opened 2024-03-29
[README change] Fix border appearing inside of logo

### LINKED ISSUE #7313: h1 border goes into the logo in README
Well, this is probably the most insignificant issue SA has ever gotten, but:

The "Scratch Addons" text in the README with a line under it, which also goes into the SA logo next to it.

The line's inside of the SA logo.

mybearworld:
Resolves #7313

### Changes

Move the image inside the `<h1>`, which causes it not to be affected by the border.

### Reason for changes

It looks a bit better.

### Tests

Tested in Edge


Samq64 (approved):
This also aligns the top of the logo with title which I think looks better.

mxmou (approved):

PR #7315, opened 2024-03-29
Remove `forums` match regex

Samq64:
### Changes

Removes the `forums` match regular expression and replaces its uses with `https://scratch.mit.edu/discuss/*`

### Reason for changes

The mobile forums are gone and the regex doesn't seem to be doing anything useful.

### Tests

Tested on Chromium 123.

### DIFF
--- a/addons/true-youtube-links/addon.json
+++ b/addons/true-youtube-links/addon.json
@@ -9,7 +9,7 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["https://scratch.mit.edu/discuss/*"],
+      "matches": ["projects", "studios", "profiles", "https://scratch.mit.edu/discuss/*"],
       "runAtComplete": false
     }
   ],
--- a/addons/true-youtube-links/addon.json
+++ b/addons/true-youtube-links/addon.json
@@ -9,7 +9,7 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["projects", "studios", "profiles", "forums"],
+      "matches": ["https://scratch.mit.edu/discuss/*"],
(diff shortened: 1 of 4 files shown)


addons/true-youtube-links/addon.json
```diff
@@ -9,7 +9,7 @@
   "userscripts": [
     {
       "url": "userscript.js",
-      "matches": ["projects", "studios", "profiles", "forums"],
+      "matches": ["https://scratch.mit.edu/discuss/*"],
```
mxmou, line 12:
```suggestion
      "matches": ["projects", "studios", "profiles", "https://scratch.mit.edu/discuss/*"],
```

Samq64: Sure but the extra matches seem unnessaey since more links doesn't change any URLs.

WorldLanguages: What?

WorldLanguages:
Is it necessary to remove the match? Don't we already have some that replace simple URL patterns?

Samq64:
> Is it necessary to remove the match? Don't we already have some that replace simple URL patterns?

Yes, but they all seem to have some validation. I can add it back though.

addons/true-youtube-links/addon.json
Samq64: When would true YouTube links need to run outside the forums? Maybe video links in Scratch news but the addon doesn't run on the home page.

mxmou: Some people put links to the forum YouTube player into project and studio descriptions because Scratch makes them clickable. Example: https://scratch.mit.edu/projects/574340251/

PR #7317, opened 2024-03-30
`msg-count-badge`: Display compact numbers instead of large message counts

iqnite:
Resolves a suggestion somebody posted on Discord a long time ago (approved by W_L)

### Changes

The extension badge will now say "*n*k" if there are 1000 or more messages, and "9k+" for more than 9000 messages.
Hovering over the extension badge will display the exact message count, even if `msg-count-badge` is disabled.

***Developer notice:** You might see a lot of commits on this PR because I accidentally based it on the `master` branch before. Please only review the commits starting from [Add setting `25e390d`](https://github.com/ScratchAddons/ScratchAddons/pull/7317/commits/25e390d4cf10fff25a82ab526372d5c61f09a3e4)*

### Reason for changes

- When message counts are too large (*ahem* griffpatch), they overflow and have an ugly look in some browsers
- A lot of apps (Discord, YouTube, Chrome, and more) display "99+" (or "9+" or simply "+") instead of large counts
- Most users don't care about whether they have 2000 or 1999 messages, while it's still useful to know whether they have 10 or 50

### Tests

Tested on Chromium 126 and Firefox 125 by manually modifying the `count` and `isLoggedIn` constants in the `updateBadge` function. (I don't have many messages lately 😞)


addons/msg-count-badge/addon.json
```diff
@@ -23,6 +23,12 @@
       "id": "showOffline",
       "type": "boolean",
       "default": false
+    },
+    {
+      "name": "Show exact count",
```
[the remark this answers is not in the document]
DNin01: We might even want to consider making exact count displays a default, non-optional behavior.

Additionally, Chrome extension badges can only fit about four characters, so we would need to round it off at some point.

DNin01:
Related: #5217

Secret-chest:
> We might even want to consider making exact count displays a default, non-optional behavior.

That it is the current way.

> Additionally, Chrome extension badges can only fit about four characters, so we would need to round it off at some point.

Then why not start showing ++++ above 9999? Why add useless limits?

DNin01:
> Then why not start showing ++++ above 9999? Why add useless limits?

Well, "++++" is less useful to see than "10k+".

iqnite:
Updated, so now it will always have a maximum of 3 characters, still showing the highest number available (like 5k, 9k+, etc.)

Secret-chest:
> Updated, so now it will always have a maximum of 3 characters, still showing the highest number available (like 5k, 9k+, etc.)

If 4 chars is the maximum, show 10k+ after 9999.

iqnite:
> If 4 chars is the maximum, show 10k+ after 9999.

As there are some accessibility features that may increase the text size, it's safer to go for a maximum of 3 characters.

Waakul:
It's better of just showing 9k+ if it's over 9k no matter what, without any kind of setting to disable this as badge's can't have more than 4 characters.

Secret-chest:
> Since some accessibility features may increase the text size, it's safer to go for a maximum of 3 characters.

Proof?

DNin01:
Anyway, the more messages you have, the more it doesn't really matter to get precise. How would you like to be able to see the difference between 100,000 and 100,001, for example? That's partly because you'd also probably be getting, like, around one new message every second or two at that point.

And we can always display the exact count on the messaging popup where we have much more space.

> Since some accessibility features may increase the text size, it's safer to go for a maximum of 3 characters.

Very good point! I did not think of that.

iqnite:
DNin01 expressed my thoughts. And anyway, there are plenty of other PRs to review, it makes no sense to discuss about 1 character.

> Proof?

In my browser, 4 characters are fine, but it covers the extension icon entirely, and there might be some browsers with greater font sizes. Another aspect to consider is that if we look at how the UIs of Chrome, Firefox, etc. evolved over the past years, they generally made buttons and texts bigger to optimize them for accessibility and touchscreens. We're not setting useless limits, we're setting the limits a bit lower than the maximum, so it's less likely that it will need an update in the future.

Secret-chest:
> 100,000 and 100,001

Sure, but I think there's a difference between 100 and 999.

iqnite:
> Sure, but I think there's a difference between 100 and 999.

The addon considers that. It shows the exact number until 999, then it shows 1k, 2k, etc. up to 9k, after 9000 it just shows 9k+. The difference between 100 and 999 is visible with or without the setting.

Samq64:
> > If 4 chars is the maximum, show 10k+ after 9999.
> 
> Since some accessibility features may increase the text size, it's safer to go for a maximum of 3 characters.

Firefox only shows 3 characters by default.

addons/msg-count-badge/addon.json
```diff
@@ -23,6 +23,12 @@
       "id": "showOffline",
       "type": "boolean",
       "default": false
+    },
+    {
+      "name": "Show exact number for large message counts",
```
Samq64, line 28: Why make this a setting though if the message count is going to be cut off anyway?

iqnite: If we remove the option, I bet many people will ask the old behavior back for the sake of complaining.

WorldLanguages: I think it's fine, we can consider adding this setting only if people actually complain.

iqnite: 1. Secret-chest already complained. 
2. It would be ironic not to have a setting for this, as there is a "Show exact counts" addon. 
3. Some browsers may have higher character limits on extension badges.
4. I don't want to edit this pull request FOUR TIMES adding/removing ONE SETTING. Because I'm sure some reviewers will say "Oh that setting was so cute would you mind adding it back", and the reviewers after them will argue "10 bytes of extension space, screw that useless setting out".

iqnite:
Possible addition: Showing the exact message count In the extension icon tooltip (when hovering over the Scratch Addons logo in the extension toolbar).

DNin01:
Try this:
```js
chrome.browserAction.setTitle({
  title: `${chrome.i18n.getMessage("extensionName")} - ${count} unread message(s)`,
});
```
(Just an example.)

iqnite:
Do you know if there's something like that for Firefox? Also, "unread messages" should be localized.

iqnite:
I'm going to make the tooltip on a separate branch, as it requires more testing and is pretty independent from this one.

Samq64:
> I'm going to make the tooltip on a separate branch, as it requires more testing and is pretty independent from this one.

Could the setting be removed then?

addons/msg-count-badge/addon.json
iqnite: **Update:** I now removed the setting as part of [d4125e2](https://github.com/ScratchAddons/ScratchAddons/pull/7317/commits/d4125e22d27764e87d2fbf1248236ef07df725c8). Most of the reviewers agreed on this, and it's not a big thing.

addons/msg-count-badge/addon.json
```diff
@@ -9,6 +9,11 @@
       "type": "info",
       "text": "You can use Do Not Disturb (right click extension icon \u2192 Do not disturb) to temporarily hide the message count.",
       "id": "dndtohide"
+    },
+    {
+      "type": "info",
+      "text": "If you disable this addon, you can still hover over the extension icon to show the message count.",
```
Samq64, line 15: #7686 disables this addon by default to reduce the amount of background activity.

background/message-cache.js
```diff
@@ -30,22 +30,25 @@ export async function updateBadge(defaultStoreId) {
...
+      if (scratchAddons.localState.addonsEnabled["msg-count-badge"] && !scratchAddons.muted) {
+        // Do not show 0, unless that 0 means logged out
+        if (count || !isLoggedIn) {
+          const displayCount = count < 1000 ? String(count) : count <= 9000 ? Math.floor(count / 1000) + "k" : "9k+";
+          const text = isLoggedIn ? String(displayCount) : "?";
+          const color = isLoggedIn ? badgeSettings.color : "#dd2222";
+          // The badge will show incorrect message count in other auth contexts.
+          // Blocked on Chrome implementing store ID-based tab query
+          await promisify(chrome.action.setBadgeBackgroundColor.bind(chrome.action))({ color });
+          await promisify(chrome.action.setBadgeText.bind(chrome.action))({ text });
```
DNin01, line 49: This isn't related to this PR, but I wonder why we await these?

DNin01 (dismissed):
Looks fine to me.

background/message-cache.js
WorldLanguages: No idea

background/message-cache.js
```diff
@@ -30,22 +30,25 @@ export async function updateBadge(defaultStoreId) {
...
-        // The badge will show incorrect message count in other auth contexts.
-        // Blocked on Chrome implementing store ID-based tab query
-        await promisify(chrome.action.setBadgeBackgroundColor.bind(chrome.action))({ color });
-        await promisify(chrome.action.setBadgeText.bind(chrome.action))({ text });
-        return;
+      const tooltipCount = isLoggedIn ? (count > 0 ? ` (${String(count)})` : "") : " (?)";
+      chrome.action.setTitle({
+        title: `${chrome.i18n.getMessage("extensionName")}${tooltipCount}`,
+      });
+      if (scratchAddons.localState.addonsEnabled["msg-count-badge"] && !scratchAddons.muted) {
```
WorldLanguages, line 40: Could you remind me why this condition was moved here? The addonsEnabled check was previously in line 33.

background/message-cache.js
```diff
@@ -57,6 +60,7 @@ export async function updateBadge(defaultStoreId) {
   // Hide badge when logged out and showOffline is false,
   // or when the logged-in user has no unread messages,
   // or when the addon is disabled
+  chrome.action.setTitle({ title: "" });
```
WorldLanguages, line 63: Why is this new line required?

iqnite: @DNin01 the whole purpose of the tooltip was to show the message count when the badge is hidden...

mxmou: We don't want to send requests to Scratch to get the message count if the badge is hidden.

background/message-cache.js
DNin01: Now I understand. So @iqnite's original intent was to display the number on the tooltip even if `msg-count-badge` was disabled. But I believe our future plan is to only fetch the message count if necessary.

PR #7686 disabled `msg-count-badge` by default, but if we check for unread messages even if the addon is disabled, we're defeating the purpose of that PR, which was to make less requests to Scratch.

background/message-cache.js
DNin01: > @DNin01 the whole purpose of the tooltip was to show the message count when the badge is hidden...

Oh.

You can revert my change if you want, or (preferably) you could change the code so that the count is only shown on the tooltip if `msg-count-badge` is enabled and Scratch Addons isn't muted (in Do Not Disturb mode).

DNin01: @iqnite Did you see this comment?

WorldLanguages: I thought this PR was only about truncating the message count if it's a very big number.

iqnite: > I thought this PR was only about truncating the message count if it's a very big number.

It was, but apparently some people think that a tooltip that is updated together with a badge is spamming Scratch's servers and needs to be handled differently than the badge. 😕

iqnite: I'm very confused about @DNin01 's changes, I'll check back when I'm on PC (on mobile right now)

DNin01:
f78a8d5f2f045c8f6a3875bf80e626f4e0480467 is the other half of the changes which I should have probably made before.

@iqnite To reiterate, recently we've been working to reduce the amount of requests we send to Scratch in the background. One of the constant background processes that used to happen was obtaining the message count by default.

In the past, the `msg-count-badge` addon was enabled by default, and would, every so often, check the user's unread message count so that it could be displayed as a badge on the extension's icon. However, if the user did not happen to have Scratch Addons pinned to their browser toolbar (which I suspect would be true for a decent percentage of users), they would only see this badge perhaps a few times during their entire browser session, which means we were making wasted requests.

Since #7686, `msg-count-badge` is disabled by default and has been disabled for users who haven't pinned the extension to their toolbar. If the addon is disabled, [this if-then condition](https://github.com/ScratchAddons/ScratchAddons/blob/9bee9a79ef559fd95e1aa90e72e2768a25fa11f9/background/message-cache.js#L33-L37) evaluates to `false`, so the unread message count is not checked.

If we remove the condition `scratchAddons.localState.addonsEnabled["msg-count-badge"]` from the if-then, we'd be checking the count at all times, updating the tooltip (which is even less visible than the badge) even when the user doesn't need to know how many messages they have.

In my opinion, we shouldn't show a tooltip for the count if `msg-count-badge` is disabled.

Hopefully the changes I've made, [these two commits](https://github.com/ScratchAddons/ScratchAddons/pull/7317/files/017f3b7e5389d30233313ff1d92285a62fcf3c11..f78a8d5f2f045c8f6a3875bf80e626f4e0480467), make sense now. I'm sorry for the confusion.

iqnite:
Oh, I get it now.
Alternatively, is there a way to lazily evaluate the message count only when the icon is hovered?

mxmou:
> Oh, I get it now.
> Alternatively, is there a way to lazily evaluate the message count only when the icon is hovered?

I don't think the browser lets the extension know when it's hovered.

Samq64 (approved):
I'm still not a fan of the nested ternary but everything else looks fine.

DNin01 (approved):

PR #7327, opened 2024-04-02
Revert "April Fools 2024 easter egg: Scratch Haddocks"

WorldLanguages:
Reverts ScratchAddons/ScratchAddons#7299


WorldLanguages:
It's not necessary to revert it, but the code is useless now.

PR #7333, opened 2024-04-04
Settings-page and popup: add hover effect for header buttons

### LINKED ISSUE #7329: No hover effect on hovering header buttons
### Why this would be helpful

There is currently no hover effect on header buttons such as theme/settings and thus it's hard to see that you're hovering it especially when it comes to accessibility.

### My suggestion

My suggestion is to set the background color of the header button to the ```hover-darken``` CSS variable on hover.

### Possible alternatives

_No response_

### Additional context

image

Waakul:
Resolves #7329

### Changes
I've added a hover effect to the header button (i.e: Settings and theme buttons) by setting their background to the ```hover-darken``` CSS variable on hover

### Reason for changes
So that the user knows their hovering it because the mouse pointer is tiny.

### Tests
I've tested these changes in Chromium (Edge) 123


Samq64 (approved):

mxmou (changes requested):
`hover-darker` has a different opacity depending on the theme (light/dark), which doesn't make sense for the header because its color is always the same. There should be a separate CSS variable specifically for hovered header buttons.

Can you add a hover background to the three dots button as well?
image
It appears on the settings page in narrow windows.

Waakul:
> `hover-darker` has a different opacity depending on the theme (light/dark), which doesn't make sense for the header because its color is always the same. There should be a separate CSS variable specifically for hovered header buttons.
> 
> Can you add a hover background to the three dots button as well? image It appears on the settings page in narrow windows.

Done, although I couldn't use the header-button class for that, I pulled of a similar thing to make it work.

webpages/styles/components/smallmode.css
```diff
@@ -15,6 +15,11 @@
   cursor: pointer;
   width: 24px;
   height: 24px;
-  margin-inline-start: 13px;
-  margin-inline-end: 5px;
+  padding: 18px;
+  margin-inline-start: -14px;
```
mxmou, line 19:
```suggestion
  margin-inline-start: -5px;
```
This aligns the close icon with the category icons when the dropdown is opened.

Waakul: it looks a bit weird tho...

mxmou (approved):

PR #7338, opened 2024-04-04
Use absolute URLs in background/core instead of relative URLs

WorldLanguages:
### Changes

Changes a few relative URLs to absolute

### Reason for changes

To avoid problems in the future if we change the location of the background context

### Tests

- [x] Chromium
- [x] Firefox


WorldLanguages:
These relative URLs were intially added in PR #567

OneShot-Niko:
Works fine on both LibreWolf and Firefox:
kép

OneShot-Niko (approved):

Samq64 (approved):

PR #7341, opened 2024-04-06
Background: import JS from a single module instead of script tags

WorldLanguages:
### Changes

For the extension background environment entrypoint, we import all JS code from JavaScript itself, instead of doing it through multiple HTML script tags.

### Reason for changes

Our manifest V3 Chrome extension will only be able to use a service_worker background environment.

### Tests

- [x] Make sure idb.js is fine to run in a module environment

### DIFF
--- a/background/background.js
+++ b/background/background.js
@@ -1,0 +1,34 @@
+// This file is loaded as ESM.
+
+// Transition from Scratch Messaging Extension
+import "./transition.js";
+
+// Load idb
+// TODO before merging: can we run this as an ES module?
+import "../libraries/thirdparty/idb.js";
+
+// Declare scratchAddons object, create global and local state
+import "./declare-scratchaddons-object.js";
+
+// Load manifests into memory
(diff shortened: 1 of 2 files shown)


background/background.js
```diff
@@ -0,0 +1,34 @@
+// This file is loaded as ESM.
+
+// Transition from Scratch Messaging Extension
+import "./transition.js";
+
+// Load idb
+// TODO before merging: can we run this as an ES module?
```
apple502j, line 7:
```suggestion
// Running as ESM is possible but requires refactor; individual JS file should import
```

WorldLanguages: Hmm, I meant running in strict mode. I don't intend to use `import {} from ""` syntax.

WorldLanguages: Let me clarify. My code comment here means "can we do this (the import on the line below)"?

The idb.js file was never used with type=module, so I'm not 100% sure if it will behave properly by importing it this way, which enables strict mode.

PR #7345, opened 2024-04-08
`fix-uploaded-svgs`: new setting for additional fixes (scratchblocks SVGs)

StickmanRed:
Resolves #7084, #7321

### Changes

Added "toEditorSVG" function taking the SVG element and returning an adjusted version accounting for some text offset issues.

### Reason for changes

This could be useful to those importing SVGs from other programs.

### Tests

This was tested on Chrome with SVGs from scratchblocks, the "Save blocks to image" addon, and the Scratch editor. The images' text did appear as they were imported in both the editor and the stage. When editing the image in Scratch, the text stays as it originally appeared; however, when changing the contents of the text, it sometimes becomes displaced, which I believe is a consequence of the "text-anchor" property.

I haven't found any other bugs in the code, but if there are errors or if it can be rewritten, the program loops through the text elements, adds "\<tspan\>" tags inside each element if there aren't any, and adds "x", "y", and "dominant-baseline" changes to "transform: translate"


Samq64:
Could the addon's description be updated to mention scratchblocks?

StickmanRed:
> Could the addon's description be updated to mention scratchblocks?

would it be something like this? (edited content in addon.json)

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -1,4 +1,75 @@
...
+    // This iframe is needed to correct "dominant-baseline"
+    const iframe = document.createElement("iframe");
+    iframe.setAttribute("src", "about:blank");
+    document.body.append(iframe);
+
+    const svg = element.cloneNode(true);
+    iframe.contentDocument.body.appendChild(svg);
+
+    var transform, translate, translateIndex, x, y, difference;
+    var baselineAdjust, innerTextElement, innerTextSpan;
```
Samq64, line 13: Could these use `let` and `const` instead and be defined where they're first used (especially the constants)?

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -1,4 +1,75 @@
 export default async function ({ addon, console }) {
+  // Fixes "x", "y", and "dominant-baseline" being ignored by Scratch, "span" causing offset
+  function toEditorSVG(element) {
```
Samq64, line 3: Maybe this function could be named fixEditorSVG and include previous fix too?

Samq64 (commented):
Still needs some work
image

addons/fix-uploaded-svgs/userscript.js
StickmanRed: Sure! Also, the previous fix should still be there, although it is outside "toEditorSVG"... would you like me to incorporate it into the function?

Samq64: Sure, so the actual SVG only needs to be modified once.

addons/fix-uploaded-svgs/userscript.js
StickmanRed: The variables declared with "var" are reassigned inside the "for" loop... how should I define the variables?

StickmanRed:
> This still needs some work:
> 
> * The green flag and dropdown arrows are missing sometimes
> * The text after an input is now to high instead of too low
> * The SVG is not centered on the canvas
> 
> Screenshot
> image

I'll fix the second bug first, since that shouldn't happen... the other two weren't really my focus, being not text-related, but maybe the first bug could be patched using some code from "blocks2image"...

StickmanRed:
> * The green flag and dropdown arrows are missing sometimes
> * The text after an input is now to high instead of too low
> * The SVG is not centered on the canvas

For the first bug, a future version could replace the "\<use\>" elements with their respective references. I'm not sure how to fix the third bug you mentioned, since it has to do with the "transform: scale" property being removed from the top "\<g\>" element...

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -1,4 +1,101 @@
...
+
+    let transform = "",
+      translate = "",
+      translateIndex = 0,
+      x = "",
+      y = "",
+      difference = 0,
+      baselineAdjust = "",
+      innerTextElement = "",
+      innerTextSpan = "";
```
Samq64, line 26: I meant defining most of these during their first assignment, but for `transform` and `translate` it makes sense to define them without a value here or at the top of the loop.

Samq64 (commented):
Scratchblocks import fine except for the offset now. I'll need to do  some more testing to make sure it doesn't accidentally mess up other text SVGs

WorldLanguages:
> I'll need to do some more testing to make sure it doesn't accidentally mess up other text SVGs

How can we be sure this won't break existing SVGs that upload just fine? It will be hard to test all cases.

StickmanRed:
> > I'll need to do some more testing to make sure it doesn't accidentally mess up other text SVGs
> 
> How can we be sure this won't break existing SVGs that upload just fine? It will be hard to test all cases.

Theoretically, the code should only add ignored properties to "transform" (offsets created by "x", "y", and "dominant-baseline"), place text nodes inside "\<span\>" tags (I'm not sure why Scratch shifts other text down), and substitute for "\<use\>" tags (added specifically to accommodate scratchblocks). Besides a bad concept execution on my part, I don't know of any new errors it could cause...

Samq64 (approved):

Samq64:
Also would it be easy to fix Scratch 2 block shadows?

Samq64:
> > How can we be sure this won't break existing SVGs that upload just fine? It will be hard to test all cases.
> 
> Theoretically, the code should only add ignored properties to "transform" (offsets created by "x", "y", and "dominant-baseline"), place text nodes inside "<span>" tags (I'm not sure why Scratch shifts other text down), and substitute for "<use>" tags (added specifically to accommodate scratchblocks). Besides a bad concept execution on my part, I don't know of any new errors it could cause...

Also the addon currently enabled by default but it would be ironic if it breaks other SVGs.

StickmanRed:
> Also would it be easy to fix Scratch 2 block shadows?

I'm not exactly sure... the only solution that I can think of is to insert every element with the "filter" property into its individual SVG "\<image\>", which should keep the shadows but will render the blocks uneditable...

DNin01:
> > I'll need to do some more testing to make sure it doesn't accidentally mess up other text SVGs
> 
> 
> 
> How can we be sure this won't break existing SVGs that upload just fine? It will be hard to test all cases.

One thing we could try is programming a notification that appears when this addon makes changes to an SVG. Just an idea. We could show to users, just so they know the addon is doing something in case it breaks an SVG…?

But even if it's just a `console.log()`, it could be useful information.

StickmanRed:
> > > I'll need to do some more testing to make sure it doesn't accidentally mess up other text SVGs
> > 
> > 
> > How can we be sure this won't break existing SVGs that upload just fine? It will be hard to test all cases.
> 
> One thing we could try is programming a notification that appears when this addon makes changes to an SVG. Just an idea. We could show to users, just so they know the addon is doing something in case it breaks an SVG…?
> 
> But even if it's just a `console.log()`, it could be useful information.

I guess that's an option, but, after reviewing and using the addon a few times, I'm convinced that it shouldn't break anything that isn't already broken...

WorldLanguages:
For this addon, we can go two ways

1. Keep it NOT enabled-by-default, and keep experimenting with SVG fixes
2. We could enable it for all users, which would benefit the most people, but we would need to be more careful, and consider showing UI notifying the user that the SVG has been altered. Possibly an Undo button to restore it to the original SVG.

A third alternative is enabling the addon by default, but instead of fixing SVGs by default, we add a little tooltip, after one or more SVGs were uploaded, with an "SVG fix" button.

WorldLanguages:
@StickmanRed You're probably right on the fact this doesn't break any SVGs, but as I usually use SVGs as "just a vector format" (and do not read sources directly) I'm not familiar with all aspects of the SVG specification. Ideally we could look out for someone who's somewhat familiar for a second look?

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -1,4 +1,62 @@
...
+    const svg = element.cloneNode(true);
+
+    // This fixes the previous issue of SVGs imported from Affinity Designer
+    if (svg.height.baseVal.valueAsString === "100%" && svg.width.baseVal.valueAsString === "100%") {
+      svg.removeAttribute("height");
+      svg.removeAttribute("width");
+    }
+
+    // This iframe is needed to correct "dominant-baseline"
+    const iframe = document.createElement("iframe");
```
WorldLanguages, line 13: Just curious, have you considered using `window.DOMParser`? It's not that important, but maybe you didn't know about it.

WorldLanguages: Probably. Using an iframe and destroying it afterwards is still fine, though.

StickmanRed:
> @StickmanRed You're probably right on the fact this doesn't break any SVGs, but as I usually use SVGs as "just a vector format" (and usually do not read XML sources directly) I'm not familiar with all aspects of the SVG specification. Ideally we could look out for someone else who's somewhat familiar for a second look?

Is there anyone to ask who could evaluate the changes, or should I wait for someone to review them?

addons/fix-uploaded-svgs/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "SVG uploads fix",
-  "description": "Fixes a bug in the Scratch editor that makes uploaded SVG files created with some image editors, such as Affinity Designer, broken.",
+  "description": "Fixes a bug in the Scratch editor that breaks some uploaded SVG files, such as those created with Affinity Designer or generated by scratchblocks.",
```
BroJac5246, line 3: Do we want to be specifically mentioning third-party software? Do we know what platforms' SVGs experience this bug? I know this is somewhat unrelated, but now is a good time to maybe revise this.

DNin01: I think it's fine when put certain ways like "such as Affinity Designer", but "such as those created with Affinity Designer" way of wording it makes it feel a little more like free advertising. It's probably a matter of how specifically the product is being mentioned - "such as" is further away from the product names in the proposed description.

pufferfish101007, line 3: Should this be `some bugs` instead of `a bug`?

StickmanRed: Maybe this should be changed to "Fixes some bugs in the Scratch editor that break uploaded SVG files from some image editors, such as Affinity Designer or scratchblocks"?

Samq64, line 3:
```suggestion
  "description": "Fixes some bugs in the Scratch editor which breaks some uploaded SVG files from Affinity Designer and scratchblocks.",
```

WorldLanguages:
@Samq64 @DNin01 To avoid delaying this further, I suggest adding these new fixes as a disabled-by-default setting. The setting could be named "Fix more SVGs (beta)" or something like that.

@StickmanRed Don't worry, you don't have to do anything. But could you please check the "Allow edits and access to secrets by maintainers" checkbox? This checkbox is in this page (your pull request) in the sidebar, at the bottom. It will allow us to make changes to your branch. Our repository doesn't use the GitHub secrets feature, so the only thing the checkbox does here is allow edits by us.

WorldLanguages:
Thanks StickmanRed, please avoid any additional commits, any additional changes will be made by us.

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -1,4 +1,64 @@
...
+        const baselineAdjust = textElement.cloneNode(true);
+        textElement.after(baselineAdjust);
+        baselineAdjust.style.dominantBaseline = "auto";
+        const difference = textElement.getBoundingClientRect().top - baselineAdjust.getBoundingClientRect().top;
+        textElement.style.dominantBaseline = "auto";
+        baselineAdjust.remove();
+
+        let transform = textElement.getAttribute("transform") ?? "";
+        transform += ` translate(${x}, ${y + difference})`;
+        textElement.setAttribute("transform", transform);
```
WorldLanguages, line 41: @StickmanRed Just wondering, can a transform attribute have multiple translate() rules? Is the case where the existing transform attribute already has a translate() rule considered here?

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -1,4 +1,64 @@
...
+      svg.removeAttribute("height");
+      svg.removeAttribute("width");
+    }
+
+    if (addon.settings.get("removeUnsupported")) {
+      // "dominant-baseline" needs this iframe to shift accordingly
+      const iframe = document.createElement("iframe");
+      iframe.setAttribute("src", "about:blank");
+      document.body.append(iframe);
+      iframe.contentDocument.body.appendChild(svg);
```
WorldLanguages, line 16: @StickmanRed Could you remind me why is this iframe needed? I see that it gets created and deleted, and there are no other references to it. Without it the addon continues to work normally it seems. So why?

WorldLanguages: (We want to be careful because SVGs may cointain dangerous \<script> tags, so it's not reasonable to inject arbitrary data to an iframe without triple-checking what we're doing)

Samq64:
`scratchblocks` SVGs still appear broken on the stage until updated in the costume editor by moving it for example.

addons/fix-uploaded-svgs/userscript.js
mxmou: > can a transform attribute have multiple translate() rules?

The first example snippet in [the specification for the transform attribute](https://www.w3.org/TR/SVG11/coords.html#TransformAttribute) contains two translate() functions.

addons/fix-uploaded-svgs/userscript.js
StickmanRed: > @StickmanRed Could you remind me why is this iframe needed? I see that it gets created and deleted, and there are no other references to it. Without it the addon continues to work normally it seems. So why?

I thought dominant-baseline would need it, but if it works fine, then I guess that wasn't the problem...

WorldLanguages: If possible, check whether that works without declaring an iframe.

StickmanRed: Scratch does ignore the `dominant-baseline` attribute, but I don't know how to account for the displacement; I've looked for a way to get baseline information from the font, but to no avail...

The current code requires a window and the whole SVG to find the text offset, because it needs the text offset in context of the inherited properties/stylesheets. I guess a safer code would be to go through the DOM, keeping track of all the attribute values, and _then_ place the text element with its computed style in the iframe?

StickmanRed: By the way, would a sandboxed iframe prevent the danger of <script> tags?

WorldLanguages: > By the way, would a sandboxed iframe prevent the danger of <script> tags?

Possibly.

WorldLanguages: So in summary, the iframe is needed for the `getBoundingClientRect()` calls, right?

WorldLanguages: I think we can use `document.createDocumentFragment()` here. I will update the code shortly. @StickmanRed Please test your code still works as intended after my changes.

WorldLanguages: > So in summary, the iframe is needed for the `getBoundingClientRect()` calls, right?

@StickmanRed I'm still not sure if this is really the case... I don't think the methods available to the `svg` object vary depending on whether it's part of a document or not...

It would also be nice if you could attach an SVG that is fixed by this code so that this case can be more easily tested.

StickmanRed:
> `scratchblocks` SVGs still appear broken on the stage until updated in the costume editor by moving it for example.
> 
> Edit: it only happens sometimes but I haven't figured out what causes it yet. Also Scratch 2 blocks don't have shadows but I think it's fine to ignore that for now.

I just found that there's a way to keep the shadows, but the block paths will not be able to be edited; if that's not too much of a problem, and the only requirement is that the SVG should be imported straight to vector mode, then I could try testing it...

addons/fix-uploaded-svgs/userscript.js
StickmanRed: > > So in summary, the iframe is needed for the `getBoundingClientRect()` calls, right?
> 
> @StickmanRed I'm still not sure if this is really the case... I don't think the methods available to the `svg` object vary depending on whether it's part of a document or not...
> 
> It would also be nice if you could attach an SVG that is fixed by this code so that this case can be more easily tested.

From `blocks2image`: 
test1
Removing the dominant-baseline code and importing this into Scratch will show a slight upward shift of the text, since their dominant-baseline property is set to "middle"

`getBoundingClientRect()` is a window method, so any detached SVGs won't have access to it...

WorldLanguages: > `getBoundingClientRect()` is a window method

Is it, really?

MDN says `The Element.getBoundingClientRect() method returns a [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) object providing information about the size of an element and its position relative to the [viewport](https://developer.mozilla.org/en-US/docs/Glossary/Viewport).`

So I guess we need to emulate a viewport, I don't think we have anything better than an iframe.

WorldLanguages:
Off-topic but I don't understand why browsers show this SVG differently.

image

SVG file: https://github.com/user-attachments/assets/45ca5ddb-4a11-4c03-abb2-fc78ae683190

Weird behavior (also happens in Firefox):
The image is cut off and making the viewport smaller also affects how much is cut off 🤔 

brave_EXp2wG77Kx

addons/fix-uploaded-svgs/userscript.js
```diff
@@ -1,4 +1,60 @@
...
+    // Fixes the previous issue of SVGs imported from Affinity Designer
+    if (svg.height.baseVal.valueAsString === "100%" && svg.width.baseVal.valueAsString === "100%") {
+      svg.removeAttribute("height");
+      svg.removeAttribute("width");
+    }
+
+    if (addon.settings.get("removeUnsupported")) {
+      // "dominant-baseline" needs this simulated DOM to shift accordingly
+      const doc = document.createDocumentFragment();
+      doc.appendChild(svg);
```
StickmanRed, line 14:
```suggestion
      // "dominant-baseline" needs this iframe to shift accordingly
      const iframe = document.createElement("iframe");
      iframe.setAttribute("src", "about:blank");
      iframe.setAttribute("sandbox", "allow-same-origin");
      document.body.append(iframe);
      iframe.contentDocument.body.appendChild(svg);
```

Tested this with an SVG containing `script` tags; fixes were still applied while the script execution was blocked. I'm unsure if this is sufficient

WorldLanguages:
@StickmanRed Yes, an iframe sandbox is the way to go. I'm not entirely familiar with that attribute, though. So you'll have to wait until I read [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) and check exactly what sandbox we want considering this is an about:blank iframe.

PR #7352, opened 2024-04-11
`name-backpack-items`: Use `addon.tab.prompt` instead of `window.prompt`

mybearworld:
### Changes

Made `name-backpack-items` use `addon.tab.prompt` instead of `window.prompt`.

### Reason for changes

Makes the UX for it a lot better - it's really confusing to me when I try putting something into my backpack and nothing happens, and then I realize that it's because of this addon.

### Tests

Tested in Edge

### DIFF
--- a/addons-l10n/en/name-backpack-items.json
+++ b/addons-l10n/en/name-backpack-items.json
@@ -1,3 +1,4 @@
 {
+  "name-backpack-items/title": "Put script into backpack",
   "name-backpack-items/prompt": "Name this script:"
 }
--- a/addons-l10n/en/name-backpack-items.json
+++ b/addons-l10n/en/name-backpack-items.json
@@ -1,4 +1,4 @@
 {
-  "name-backpack-items/title": "Put script into backpack",
+  "name-backpack-items/title": "Add to Backpack",
   "name-backpack-items/prompt": "Name this script:"
 }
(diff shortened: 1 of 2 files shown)


WorldLanguages:
> it's really confusing to me when I try putting something into my backpack and nothing happens, and then I realize that it's because of this addon.

What do you mean by "nothing happens"?

mybearworld:
The oage

WorldLanguages:
I believe we never ported this to `addon.tab.prompt` before because the synchronous nature of `window.prompt` was useful here. But we never noticed we could just delay `realXhr.send()` until the user cancels the dialog or chooses a name.

addons-l10n/en/name-backpack-items.json
```diff
@@ -1,3 +1,4 @@
 {
+  "name-backpack-items/title": "Put script into backpack",
```
mxmou, line 2:
```suggestion
  "name-backpack-items/title": "Add to Backpack",
```
Scratch's own modals have short titles and use title case.

mxmou (approved):

Samq64 (approved):

WorldLanguages (approved):
Clicking "cancel" still adds the script to the backpack, but this is the same behavior as this addon always had.

Can't find any potential issues with calling an async IIFE on a polluted xhr send method.

PR #6877, opened 2023-11-17
Draft - MV3 manifest supported by Chrome & Firefox

WorldLanguages:
The extension is not functional at this state. This is just a demo.
This PR requires Chrome 121+ and Firefox 121+. Try downloading Chrome Canary and Firefox Nightly. The stable versions won't work as of November 2023.

### Changes

This extension features a `manifest.json` file which is accepted by both browsers.
In Chrome, a service worker is used as the background context. In Firefox, a background page is used instead.

Issue #1538, opened 2021-02-09
Manifest v3: move from background page to service worker


WorldLanguages:
1. ~~The service worker needs to be in the root folder, and has to be a single JS file - so we'll have two files in root now, manifest.json and sw.js~~ They can be in subfolders starting in Chrome 93.
2. ~~Service workers don't support modules yet, so it's not worth migrating until then.~~ They now do in Chrome 92. The sw.js file in root would just be a bunch of imports from files on background/*
3. I'm not entirely sure how objects like `window.scratchAddons` work in the service worker model. Hopefully these aren't cleared after the service worker is put to sleep? (EDIT: data isn't cleared, unlike a typical service worker) (edit 2: i was wrong, it is cleared after 5 minutes max)

DNin01:
What will we do about number 3?

Luckily, moving all the background scripts into a single file that will eventually become the service worker shouldn't be too hard. The only thing present in the background page are scripts, so we would just create a new script that imports those, and to maintain MV2 compatibility and support older Firefox versions, load that mega-script into the background page (we could make a PR for that right now!). The most difficult part will be working out all the other differences between background pages and service workers.

WorldLanguages:
> What will we do about number 3?

Well, Chrome kills service workers less frequently now. Also after some testing, the behavior doesn't seem that bad. We just have to make sure all event listeners are registered at the top level.

DNin01:
But are there going to be any issues related to variables in the service worker getting erased upon termination? The Chrome Developers documentation does say[^1]:
> Any global variables you set will be lost if the service worker shuts down.

I guess we could set alarms every 30 seconds now - that's possible since Chrome 120[^3]. I don't know for sure, but it seems like, if I understand correctly, that would theoretically keep the service worker alive indefinitely, since that's exactly how long it takes for it to be shut down, but there might still be some edge cases. Plus, we probably want to cover versions of Chrome older than that.

In summary, there might be ways to prevent the service worker from being terminated automatically, but we should expect it to still be terminated at times and keep browser support in mind. It's a good idea to intentionally let it terminate if the user hasn't been using Scratch Addons or on Scratch for a little while (say, 5 minutes). If it was terminated, we can re-declare all the global variables by regenerating or loading them from [a session storage space](https://developer.chrome.com/docs/extensions/mv3/service_workers/service-worker-lifecycle/#persist-data). It is best practice to design service workers around being disposable[^2]:
[...]

PR #7366, opened 2024-04-20
Mv3 progress background

### LINKED ISSUE #1533: Use chrome.action instead of chrome.pageAction in JS and manifest.json
Manifest v3 will only support `chrome.action` and the `action` property in manifest.json
Source: https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#action-api-unification

WorldLanguages:
Resolves #1533
Related to #1538 - Service workers do not have a global `window` object

Related: see draft PR #6877 for a preview of a manifest.json file with MV3.

### Changes

- Use `globalThis` instead of `window` for globals in the background context. It is equivalent to `self` in this case
- Consider the rename of "browser action" to "action" in multiple APIs
- Consider upcoming `host_permissions` field in extension manifest

- [ ] TODO: modify code that assumes top-level code runs once for each extension session. Instead, explicitly wrap those in a "startup" event (unsure if there's a `chrome.*` API for that or if we're supposed to use service worker lifetime events)
- [ ] TODO: make sure all event listeners are top level (see https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers#register-listeners)

### Reason for changes

These are relatively small changes that can get into an MV2 update if we want, but still get us closer to MV3.

### Tests

Tested under MV2 in Chrome and Firefox

### DIFF
--- a/background/declare-scratchaddons-object.js
+++ b/background/declare-scratchaddons-object.js
@@ -2,7 +2,7 @@ import globalStateProxy from "./imports/global-state.js";
 import localStateProxy from "./imports/local-state.js";
 import BackgroundLocalizationProvider from "./l10n.js";
 
-window.scratchAddons = {};
+globalThis.scratchAddons = {};
 
 // Event target for local background page events
 scratchAddons.localEvents = new EventTarget();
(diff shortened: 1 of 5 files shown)


WorldLanguages:
By the way, if you're curious check PR #7365 which I just merged.

PR #7368, opened 2024-04-21
Global addon api

valmontechno:
### Changes

Add an new addon that makes the Addon API accessible as a global variable `window.addon`.

### Reason for changes

Useful for prototyping an addon in the devtools console in particular by getting `vm` or `blockly` without having to create a new addon with all the files that this implies (`addons.json`,` addon.json`, `userscript.js`)

### Tests

Google Chrome: Version 123.0.6312.124 (Build officiel) (64 bits)

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -156,6 +156,7 @@
   "reorder-custom-inputs",
   "place-backpack-code-at-cursor",
   "big-save-button",
+  "global-addon-api",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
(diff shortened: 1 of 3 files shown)


mxmou:
The `addon.tab` API is already accessible in the console as `scratchAddons.eventTargets.tab[0]`.

valmontechno:
It is more convenient to have direct access to allow you to copy and paste addon code into the console without having to modify them.

WorldLanguages:
The debugging tips in the docs simply suggest to add `window.addon=addon` to whatever addon you're developing: https://scratchaddons.com/docs/develop/userscripts/debugging-userscripts/#use-the-addon-api-from-the-console

WorldLanguages:
See #7386. I added an `__addon` global that is accessible as long as 1 userscript is running on the page.

PR #7375, opened 2024-04-23
Do not open "permissions" page on every startup

WorldLanguages:
### Changes

The `/webpages/settings/permissions.html` page now opens if:
1. If the user has just revoked a host permission moments ago.
2. If the user has opened the popup and a permission is missing
3. If the user has opened the settings page and a permission is missing
4. NEW: if the user has just installed the extension, and the browser didn't grant host permissions by default.

It no longer opens that page on every extension startup. That means that if a permission is missing and the extension looks broken, the user is expected to interact with the extension in some way to know why it isn't working.

### Reason for changes

In manifest V3, it's not that straight forward to open the page once per extension session.
Also, we usually try to avoid opening new tabs unless there's user interaction. For example, we only show the "unsupported browser" page if the user opens the popup or the settings page.

### Tests

Tested in Chromium. It's just a new boolean parameter, though.

### DIFF
--- a/background/handle-permissions.js
+++ b/background/handle-permissions.js
@@ -51,3 +51,12 @@ chrome.permissions.onRemoved?.addListener(() => {
 
 checkSitePermissions(() => {}, { isStartup: true });
 checkOptionalPermissions();
+
+chrome.runtime.onInstalled.addListener((details) => {
+  if (details.reason === "install") {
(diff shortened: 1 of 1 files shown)


Samq64 (approved):
Tested on Chromium 124 and Firefox 125, both MV2 and 3.

PR #7386, opened 2024-04-28
Fix `addon.tab.direction` bugs and implement `__addon` global

### LINKED ISSUE #7374: `addon.tab.direction` sometimes fails
The implementation of the direction getter should use `addon.auth.scratchLang` instead of relying on `scratchAddons.globalState.auth.scratchLang`

Related to #6802

WorldLanguages:
### Changes

- Resolves #7374. Tested by deleting the `scratchlanguage` cookie and making sure `addon.tab.direction === "ltr"` unless your browser language is RTL (for example Arabic)
- If at least one userscript is available, `window.__addon` can be used from the console, which may be the addon object from any addon, but should be useful for experimenting. Idea from #7368


Samq64 (approved):
> `addon.self.direction`

*`addon.tab.direction`

Issue #6355, opened 2023-07-04
Unload unneeded addon.json data


DNin01:
### Why this would be helpful

According to #5190:

> ### Memory usage (background page)
> It's important to consider the memory usage of our background page. It's true that Scratch itself is pretty memory intensive, but SA is running all the time, even if the user isn't currently using Scratch! Since v1.0.0, all manifests are stored inside a variable in the background page. This includes all this data from all addons, no matter if they're enabled or not:
> 
> * Addon ID
> * The entire addon.json file as an Object
> * If the browser language is not set to English, and it's a language we support (such as Spanish), the English name and description are stored in memory too (see https://github.com/ScratchAddons/ScratchAddons/blob/master/background/load-addon-manifests.js#L58)
> 
> This `scratchAddons.manifests` array gets larger with every release. I'm not sure how to properly check the actual memory usage of this "big" object, but according to Chrome Devtools, it appears to use about 0.25MB.
> 
> We could also check whether `scratchAddons.l10n` stores any unnecessary messages in memory in the background page.

### My suggestion

To cut down on memory usage, we can:
- Unload details about disabled addons
- Potentially unload even more data if the user hasn't accessed Scratch for a while

Information would be loaded back into memory as needed, like if a user expands an addon to change settings or enables an addon.

The following information will be needed in these cases:
| Type of data | <span title="Needed for the addon to work">🧩</span> | <span title="Needed in order to be displayed in the popup">◻️</span> | <span title="Needed in order to be displayed in the full screen settings page">⚙️</span> |
|--------|--------|--------|--------|
| Name | | 🔷 | 🔷 |
| Description | | | 🔷 |
| Tags and category | | 🔷 | 🔷 |
| Settings and other details | | | |
| Selected settings | 🔷 | | |
| JS and CSS | 🔷 | | |

<details><summary>Key</summary>
<p>

🧩 Needed for the addon to work
◻️ Needed in order to be displayed in the popup
⚙️ Needed in order to be displayed in the full screen settings page

</p>
</details>

Everything information-related would be needed if the user expanded the addon card, and the ID would probably always stay in memory.

### Possible alternatives

Compress the contents. Like, maybe we can use numerical IDs instead of strings to identify addons and settings.

apple502j:
Too complicated for something that only takes less than a megabyte of data. We are not using Windows 95.

[...]

PR #7387, opened 2024-04-28
Manifest V3 extension

WorldLanguages:
### Changes

See commits and files changed.
When viewing diff, I recommend "hide whitespace" on.
There might also be some relevant commits in the master branch (commits from April 2024 onwards).
Remember that the `manifest.json` file will only work on Chrome 121+ & Firefox 121+

### Known issues (some are TODOs to resolve after MV3 release)
- In Chrome, if the worker is sleeping (formally, "stopped") and the user navigates to a new Scratch page, dark mode and any other addons might take a few seconds to apply.
- Messaging features should be considered broken unless proven otherwise. (Also worth looking: popup ports and their lifetime when service worker goes to sleep)
- When receiving a comment, Scratch Notifier will show the raw HTML of the comment, instead of attempting to display it properly. This is intended by commit https://github.com/ScratchAddons/ScratchAddons/commit/77c785def7699e4ebee0e6f851f9a2371af1a188 and the root cause is DOM APIs unavailable in the background context in Chrome.
- In Chrome 96-99, `chrome.i18n` APIs do not work in the background context. That means that in those browsers, users might see some strings in their preferred language (`navigator.language`) instead of their UI language. Most strings in the popup and settings page (except addon metadata) will still display in the browser UI language.
- The "Scratch Addons ping" only plays on Firefox. See issue #3877
- The `MANIFEST_VERSION` constants can be removed.

### Chrome

- We should test whether the extension works properly on Chrome 96 (our min version choice). The `manifest.json` file needs the `background.page` property to be removed to load properly.
- In Chrome, the background context is a service worker. To emulate the service worker going to sleep, the URL you're looking for is `chrome://serviceworker-internals` and the "stop" button (find the worker by extension ID in the URL)

- [ ] Additionally, we should emulate the MV2→MV3 upgrade using [Extension Update Testing Tool](https://github.com/GoogleChromeLabs/extension-update-testing-tool) when closer to release.

### Firefox

- Firefox does not grant any host permissions by default on MV3, so you need to click the "enable" button on `webpages/settings/permissions.html` after loading it. That page should open automatically.
- We should test whether the extension works properly on Firefox 109 (our min version choice). The `manifest.json` file needs the `background.service_worker` and `background.type` properties to be removed to load properly.
- We won't be testing the MV3 on Firefox for Android just yet. We can focus on that later.


WorldLanguages:
@Samq64 @mxmou @DNin01 @BroJac5246 

I'm honestly considering releasing the current state of this PR to a small % of Chrome users and see how it goes. If you have any objections please tell me.

Samq64:
Tested on both current and old Firefox and chromium releases and didn't notice any big problems/ Even Scratch messaging seems to be functional.

WorldLanguages:
Potential risks (risks, not bugs):
- Data loss seems unlikely - editor addons will behave the same, and I have no reason to think extension settings might be at risk.
- I will check with the [Extension Update Testing Tool](https://github.com/GoogleChromeLabs/extension-update-testing-tool) before uploading a new version (and after running the gen-manifest action and manually checking its output). So the extension will update normally for all Chrome users. It will not get auto-disabled due to new permissions. Fortunately, `declarativeNetRequestWithHostAccess` already exists because we sent feedback to the Chrome extensions team.
- Chromium already rate limits the background context of extensions if one accidentally DDoS by sending many requests to the same website. However, I don't see what could cause that.
- This update does not change the minimum browser versions, as we already did that one month ago.
- To prevent any further risks, I will only release this to a small % of Chrome users and slowly increase it day by day.

DNin01:
> @Samq64 @mxmou @DNin01 @BroJac5246 
> 
> 
> 
> I'm honestly considering releasing the current state of this PR to a small % of Chrome users and see how it goes. If you have any objections please tell me.

I'll think about giving this a test and tell you how I feel. We could also try sharing this with potential volunteers like some of the folks over on the Discord server (which I'm not on but you'll know what to do), and if that's successful, we could try starting a gradual rollout.

Samq64:
Shouldn't scratchfoundation.github.io be an option host permission? Right now disabling it triggers the site permissions page.

BroJac5246:
> I'm honestly considering releasing the current state of this PR to a small % of Chrome users and see how it goes. If you have any objections please tell me.

We need to get this out soon so it's better to risk breaking it now for some users than breaking it for all of them later IMO.

I can't review the code but it sounds like it works so that's good.

WorldLanguages:
> Shouldn't scratchfoundation.github.io be an option host permission?

It is a mandatory host permission in the "prerelease", but it's completely missing (not even optional!) in the actual releases that are sent to stores.
Also, as you said, the behavior remains the same as before.

DNin01:
When I was using the "Website dark mode and customizable colors" addon, the page would flash white whenever the service worker restarts.

See the video below (the screen will flash at about 4 seconds):

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/a9265d78-3361-4c42-bee5-b28b8714910b

This happens because the CSS variables are being reinitialized. This error also occurs when the service worker restarts:
```
scratch.mit.edu/:1
Error in event handler: TypeError: Cannot read properties of undefined (reading 'page')
    at getColor (chrome-extension://.../content-scripts/cs.js:277:38)
    at getColor (chrome-extension://.../content-scripts/cs.js:284:15)
    at setCssVariables (chrome-extension://.../content-scripts/cs.js:327:24)
    at chrome-extension://.../content-scripts/cs.js:525:7
```

I have the "Show message count on extension icon" addon enabled so that there's something to periodically wake up the service worker.

DNin01:
Since the service worker restarts whenever the popup is opened, that's one way to trigger the above issue on command, if the service worker is inactive when you open the popup.

DNin01:
Here are some more examples:

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/1c0f375d-b682-42f4-b788-bfbdeacc973c

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/cba1da11-05ce-4a2b-a660-45980991a131

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/729ade11-52ed-470e-b369-1ba561ec87d7

And finally, the editor:

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/dd56ac03-7cd5-4501-b18e-ea5311d8c5e5

This also may have created some lag and refreshed the values of the blocks in the block palette, if you see the x and y coordinates change to 0.

So yeah, stuff happens when the service worker restarts.

Hans5958:
Why not do something like `(chrome.i18n?.getMessage("unmute")) || "Turn off Do not disturb"`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility

WorldLanguages:
@Hans5958 I believe `chrome.i18n` may be an empty object. If you want to change it, download Chromium 96 and check if it works.

WorldLanguages:
@DNin01 Good points, I haven't considered those cases.

WorldLanguages:
> refreshed the values of the blocks in the block palette, if you see the x and y coordinates change to 0.

@DNin01 Can you send your settings file and make sure you can reproduce it two, three times in a row?

WorldLanguages:
@DNin01 Most likely, you won't be able to reproduce the issue after testing my last commit.

DNin01:
> @DNin01 Most likely, you won't be able to reproduce any of the issues after testing my last commit.

I couldn't reproduce the issues anymore. 👍

Samq64:
> I would assume `chrome.i18n.getMessage?.()` would work, but I haven't tested.

I can confirm it does.

DNin01:
Migrating settings, which is done every time the service worker restarts even if Scratch Addons has not just been updated only takes about 2-4 ms (and reading storage takes about ~10 ms) - not that bad.

The worst from my testing though is loading the addon manifests, which can be as fast as 200 ms but sometimes takes over a second! I did a test and, on average, reading all the manifests from the disk took 210 ms and processing them with `json()` took 40 ms, and this could vary a lot. So definitely not something we want to do constantly.

We could cache the manifests in Chrome session storage to avoid this problem. In general, keeping static objects in memory is better for performance than doing something regularly that causes disk load and garbage collection, particularly for things that don't take up that much memory to begin with.

manifest.json
```diff
@@ -1,14 +1,16 @@
 {
-  "manifest_version": 2,
+  "manifest_version": 3,
   "name": "__MSG_extensionName__",
   "description": "__MSG_extensionDescription__",
   "version": "1.38.0",
   "version_name": "1.38.0-prerelease",
   "default_locale": "en",
   "background": {
+    "service_worker": "background/background.js",
```
DNin01, line 9: A lot happens every time the service worker is reloaded, which for us is every minute[^1]. It would be a good idea to minimize any unnecessary computations, especially getting all the addon manifests as I said before.

[^1]: See [code search (2 files)](https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons+periodInMinutes&type=code)

addons/scratch-notifier/notifier.js
```diff
@@ -332,7 +332,12 @@ export function notifyNewMessages(messages) {
       parent_title: message.parent_title, // Remixes only
     };
     if (!anyNotified && settings.notification_sound === "addons-ping") {
-      new Audio(chrome.runtime.getURL("./addons/scratch-notifier/ping.mp3")).play();
+      // Note: no audio playback in service worker (Chrome)
+      // https://github.com/ScratchAddons/ScratchAddons/issues/3877
+      if (globalThis.Audio) {
+        // Firefox only
+        new Audio(chrome.runtime.getURL("./addons/scratch-notifier/ping.mp3")).play();
+      }
```
DNin01, line 340: (Optional) If you feel like it, we could theoretically use a [offscreen document](https://developer.chrome.com/docs/extensions/reference/api/offscreen) to play our notification sound.

DNin01 (dismissed):

manifest.json
WorldLanguages: The service worker will not reload 60 times per minute, in any reasonable desktop browser

WorldLanguages: For example, AFAIK Chrome will keep the service worker active for 5 minutes as long as there's enough memory

addons/scratch-notifier/notifier.js
WorldLanguages: Yes, we could

manifest.json
DNin01: > For example, AFAIK Chrome will keep the service worker active for 5 minutes as long as there's enough memory

You sure? I wrote some debugging code to count each time the addon manifests were loaded from disk into memory, turned Scratch Addons on, and read some articles while I waited and after about 30 minutes, the service worker had reloaded 30 times. Unless maybe I just need to do more extensive testing.

By the way, I just had to insert this at line 37 in `load-addon-manifests.js` to count the reloads.
```js
chrome.storage.local.set({ restarts: ((await chrome.storage.local.get("restarts")).restarts || 0) + 1 });
```

The timer, however long it is, probably resets every time you visit a page on Scratch, so if you want to test the service worker's lifecycle in the case where Scratch Addons isn't doing much of anything, I'd advise you browse on other websites for a little bit while measuring the extension's activity.

Even if it stays active for longer sometimes, Google says it's best not to keep a service worker alive indefinitely.

Anyway, browser extensions are just the kind of thing that we generally expect to use very little system resources, so we can probably do better than rebuilding data that we already have only to never use it most of the time (whenever the user is not on Scratch or stays on the same page for an extended period of time, like when playing a game on Scratch). [One of my issues](https://github.com/ScratchAddons/ScratchAddons/issues/6355#issuecomment-1619429386) got closed for this reason.

All of that might turn out to just be my opinion, but I recommend we try.

<sub>Windows 11, Edge 124, 16 GB RAM, commit https://github.com/ScratchAddons/ScratchAddons/commit/67d8edfacac18eeb4b3554e990aa421d89e617c2</sub>

WorldLanguages:
> Actually, Scratch Addons uses an alarm that is always active, so whether you have this addon enabled or not, the service worker will still wake up every minute.

Okay, I changed it so that the alarm only exists if the `msg-count-badge` addon is active.
The `fetchMessages` alarm runs every 5 minutes already, which is reasonable, so I'm not changing that one.
The `cleanCsInfoCache` alarm I just changed to run every 5 minutes instead of 1 minute if the background environment is a service worker. In the future it would be nicer to manage the cache differently instead, for example we could look into this when testing for Firefox for Android.
These three are the only periodic alarms we use.

manifest.json
WorldLanguages: I would say we look into this in the future

WorldLanguages: Chrome knows if we have a periodic alarm that runs every 1 minute. If it decides that it's better for the user (and system resources) to kill and restart the service worker 60 times per minute, I find that controversial.

DNin01: I'm looking into using `chrome.storage.session` to save addon manifests to memory in Chrome 102+. Extension session storage is cleared when the browser is restarted, or the extension is updated, reloaded, or disabled and reenabled. If I find a solution, I'll open a PR when I'm done.

DNin01: Looks like we'll also need to do this with the localized strings.

WorldLanguages: Also, the cache should be disabled in development mode, if the extension updates to a new version, and if the value of the "show addon names in English" setting changed.

DNin01: > Also, the cache should be disabled in development mode, if the extension updates to a new version, and if the value of the "show addon names in English" setting changed.

Extension session storage is cleared when the browser is restarted, or the extension is updated, reload, or disabled and then reenabled, according to my testing. This might not apply to all browsers or environments, though. Depending on the behavior, we may never have to clear session storage ourselves at all.

DNin01: But clearing it anyway would still be safe. We could clear on `chrome.runtime.onInstalled`, which fires every browser restart or extension install, update, reload, or enable.

DNin01:
Loading the addons l10n usually seems to take about 200 ms as well, and the addon manifest loading routine waits for it to finish before continuing.

RedGuy12:
> We could also try sharing this with potential volunteers like some of the folks over on the Discord server (which I'm not on but you'll know what to do), and if that's successful, we could try starting a gradual rollout.

hey, i'm managing the discord at the moment
i don't really read github anymore, but if you ping me here (or on discord) i can make an announcement on the server, we have a system set up for that already
i think it's already been sent to chrome, but I'll do that now anyways since i don't think everyone will be receiving the update right away

PR #7388, opened 2024-04-28
Move "is browser supported" logic from background to webpages

### LINKED ISSUE #7373: Popup and settings page should know if the browser is unsupported
Right now, our webpages do not know if the browser is unsupported, they rely on the background context determining that.

However, if we use newer JS syntax in **any JS file we run on the background context** it would break. This is the case since #7341 was merged.

The solution is then to make our webpages self aware, so that they can tell the user their browser is unsupported by themselves.

WorldLanguages:
Resolves #7373

### Changes

- The webpages now check by themselves if the browser is supported
- When opening in a new tab, we now set `{active: true}` just like `/webpages/settings/permissions.html` always did when auto-opening

### Reason for changes

However, if we use newer JS syntax in any JS file we run on the background context it would break. This is the case since https://github.com/ScratchAddons/ScratchAddons/pull/7341 was merged.

The solution is then to make our webpages self aware, so that they can tell the user their browser is unsupported by themselves.

### Tests

Tested popup and settings pages in Chromium


Samq64 (approved):
Was there a reason for using 2 files before?

WorldLanguages:
> Was there a reason for using the background page before?

I guess we might have considered opening the unsupported-browser page on startup/update at some point. But as a general rule we'll try to avoid opening new tabs automatically. Other extensions such as Tampermonkey do that and it's quite annoying.

PR #7391, opened 2024-04-29
fullscreen: correct canvas size when not in full screen

### LINKED ISSUE #7389: "Enhanced full screen" (`fullscreen`) makes stage blurry (inconsistent canvas size)
From feedback:

> Enhanced full screen causes the project to be blurry when exiting full screen.
The width and height of the <canvas> are set to "482" and "362" instead of "480" and "360". The style still correctly becomes "height: 360px; width: 480px;". This is what causes the blurry stage.
SA 1.37.1 Chrome/123

mxmou:
Resolves #7389

### Changes

Changes the code that measures the canvas size to exclude the stage border.

### Reason for changes

If the size is wrong, the canvas becomes blurry.

### Tests

Tested on Edge and Firefox.


WorldLanguages (approved):
I did not test but I approve

Samq64 (approved):

PR #7395, opened 2024-04-30
Fix spinner.css path (typo in URL)

Samq64:
### Changes

Fixes a typo in the `spinner.css` import path and removes `%addon-self-dir%`.

### Reason for changes

The typo caused `scratchr2` to fail to load load the spinner on the forums and `%addon-self-dir%` is unnecessary since CSS imports are already relative.

### Tests

Untested.

Also my typo fix commit has a funny typo in its name.

### DIFF
--- a/addons/scratchr2/forums.css
+++ b/addons/scratchr2/forums.css
@@ -1,4 +1,4 @@
-@import url("%addon-self-dir%/../../libraries/common/css/spinner.css");
+@import url("../../libraries/common/cs/spinner.css");
 
 body > #pagewrapper {
   min-height: 0;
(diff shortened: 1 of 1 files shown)


WorldLanguages:
> and `%addon-self-dir%` is unnecessary since CSS imports are already relative.

No

addons/scratchr2/forums.css
```diff
@@ -1,4 +1,4 @@
-@import url("%addon-self-dir%/../../libraries/common/css/spinner.css");
+@import url("../../libraries/common/cs/spinner.css");
```
WorldLanguages, line 1:
```suggestion
@import url("%addon-self-dir%/../../libraries/common/cs/spinner.css");
```

mxmou (approved):

WorldLanguages (approved):

WorldLanguages:
To clarify, this addon uses `"injectAsStyleElt": true`, so browsers have no way to know that the import is supposed to be relative to the chrome extension. At runtime, every instance of `%addon-self-dir%` is replaced properly before injecting the <style> element to the DOM.

PR #7443, opened 2024-05-17
New Addon: Timing Blocks and Line-By-Line Profiling for Debugger

### LINKED ISSUE #7418: New addon: Timing and RTC profiler
### Why this would be helpful

I would like a way for people to time their projects more easily. When optimizing a project timing is neccessary for implementation comparisons. I envision this timing method to allow for timing within absolute and relative terms.

Here's what I mean by relative timing: 
The time for each block to run may be different per computer but it's time relative to any other block remains constant. As using the subtraction operator as an atomic unit (a logical choice due to it's prevelance in most projects), we can determine the RTC of any block relative to the subtraction operator. The table that lists this RTC is given here: https://scratch.mit.edu/projects/725580642/

### How this addon works

I envision this working in one of two ways:

1. A START TIMING and END TIMING timing block. would be part of the profiling blocks. The block starts and ends timing a particular section of the code, and has an input for a label. This is the most flexible option, but the blocks themselves might cause some overhead and it requires two new blocks to be added. The list of labelled timing sections is visually displayed.

2. You can right click on a costum block and hook a timer to it. This list of hooks is visually displayed. Once hooked, right click lets you unhook. A function displays that it is hooked with an icon in the top left.

Then the profiling tab would have a new section called TIMING. 
This would display for each of the labelled timing blocks / timed custom blocks:
- the absolute and relative (RTC) time averaged per call
- The number of times each was called
- An option to delete the hook (if using hooks).

### Possible alternatives

_No response_

### Additional context

I discussed this 4 years ago: https://github.com/ScratchAddons/ScratchAddons/discussions/117

Joeclinton1:
Resolves #7418

### Changes

This is the biggest addon I've made to date. It adds:

- two new blocks: "start timer" and "stop timer"
  - the block-switching addon is modified to account for these blocks
- a new tab for debugger called "Timing"
- Timing tab use a logView to show a table of the stored timing data from the started timers
  - The info shown is: Total Time, Average Time, Percent Time, Call count
  - Via a setting you can replace Percent Time with Ratio time, which is the total time divided by the minimum total time
- A "tools" button is added which when clicked shows a dropdown toolbar. The toolbar has 3 buttons:
  - View Line-by-line: This triggers a custom profiler to enable itself, and begin tracking line by line times. Each line adds it's own timer
  - Show RTC: this begins tracking the Relative Time Cost (RTC) for each line. Where RTC is a precomputed time cost which is derived from the total set of block used in the line. When RTC is shown a column is added displaying elapsed RTC for each column
  - Show Heatmap: this recolours the block line associated with each timer, with a heatmap colour scale to visually represent the  percentage time of that line.
  - If heatmap is shown, then a slider appears that lets you adjust the maximum heatmap range.
- clicking the table headers sorts the column

### Reason for changes

See  #7418

### Tests

The functionality described in Changes has been tested and works. The RTC value calculation is not finished yet.


DNin01:
We could maybe move the timing data into the Performance tab, if necessary. There might not be enough space to fit four tabs in some languages…?

Joeclinton1:
> We could maybe move the timing data into the Performance tab, if necessary. There might not be enough space to fit four tabs in some languages…?

It would just be cluttered. the performance tab has these graphs. you'd need to put them above or below the timers which makes them less easy to use. To remedy that you need a second level set of tabs and things get annoying quickly. I made the window wider by the width of the word timers. Technically yes it's not going to fit. perhaps we can resize the whole window if it doesn't fit dynamically with js. Maybe a burger menu, could automatically enable when it runs out space, and the 4 tabs become a dropdown or side menu?

Joeclinton1:
- Making RTC account for o(n) time blocks
 - Make RTC account for stamp blocks dependency on bitmap/vector
 Going to do these now. 
 
 The o(n) blocks are:
 - procedure call
 - insertatlist
 - string contains
 - item # of
 - list contains

Joeclinton1:
I've completed the changes I wanted to make so it's now open for review.

addons-l10n/en/debugger.json
```diff
@@ -36,5 +40,19 @@
   "debugger/log-msg-list-append-too-long": "Failed to add to list '{list}', cannot add over 200,000 items.",
   "debugger/log-msg-list-insert-too-long": "Item 200,000 removed from list '{list}' to make room for insertion, cannot add over 200,000 items.",
   "debugger/log-cloud-data-nan": "Failed to set cloud variable '{var}'. Cloud variables can only contain numbers.",
-  "debugger/log-cloud-data-too-long": "Failed to set cloud variable '{var}'. Cloud variables can only contain a maximum of 256 digits."
+  "debugger/log-cloud-data-too-long": "Failed to set cloud variable '{var}'. Cloud variables can only contain a maximum of 256 digits.",
+  "debugger/timing-view-timers": "View Timers",
+  "debugger/timing-view-line-by-line": "View Line-By-Line",
+  "debugger/timing-show-rtc": "Show RTC",
+  "debugger/timing-hide-rtc": "Hide RTC",
```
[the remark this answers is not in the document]
Joeclinton1: It's an acronym for Relative Time Cost, it is the relative time to the cost of the subtraction operation. https://en.m.wikipedia.org/wiki/Relative_atomic_mass like this. Perhaps there could be a info symbol tool tip next to it that explains briefly what it is when a user hovers over it? Translators should leave it as it is.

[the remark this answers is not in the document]
WorldLanguages: If we can say "RAM memory", then why not "RTC cost"?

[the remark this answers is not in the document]
Joeclinton1: Ok, we can do Relative Time and then have a tool tip explaining this is "Relative Time Cost, which is the relative time  cost of the operation to the time cost of the subtraction operation."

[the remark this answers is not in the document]
Joeclinton1: Alternatively Operation Cost, could be a good name.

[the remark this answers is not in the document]
Joeclinton1: I have updated to Operation cost, added a tool tip and made a popout that renders the RTC chart that I have on scratch in it's own window so users don't have to go to my scratch page to see what it is.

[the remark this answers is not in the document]
DNin01: > If we can say "RAM memory", then why not "RTC cost"?

You didn't mean "relative time cost cost", do you?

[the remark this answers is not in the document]
Tacodiva: > If we can say "RAM memory", then why not "RTC cost"?

I don't know if it matters but I definitely know what RAM is but I didn't know what RTC meant.

Tacodiva:
I really like this idea, but I have a couple concerns.

Firstly, I'm not sure about the inclusion of the "Operation Cost" of each block in that chart you've made. I've done quite a bit of performance work with TW and one things I've learned is how inconsistent it can be between different environments (mostly, the browser but also other factors).  I'm curious how consistent these measurements are. There's no error margin included in the measurements which makes me wonder about how useful including hard numbers like this is. 

I'm also not sure about how you're measuring the performance of blocks which yield. To me at the very least, yielding time should not count in the "Percent Time" counter. Take a look at this project:

image

This will take 100 frames (about 3.33 seconds) to run. It reports this, and says "percent time" is 100%. Now, delete the script with the `forever move 1 steps` and measure again. Suddenly even though we're measuring the exact same blocks, they execute in ~1ms still with 100% "percent time". To me, this is not correct, The blocks were still only executing for ~1ms in both cases, but the percent time is including the yield which is a bit counter intuitive to me. It would be great to distinguish between these two measurements better, and would definitely be helpful to the nerds like me who would use the timing feature to begin with.

Another thing,, I know given current Scratch Addons limitations, this would be very hard, but it would be nice to make the timing block a c block which measures the stuff between it. That way you wouldn't have to keep the labels consistant between the start and stop blocks, though it does introduce some other limitations. 

That was a lot of criticisms, but I do really like the idea! I've been wanting something like this for a while and it makes a lot of sense to have it be a part of the debugging tools.

Joeclinton1:
@Tacodiva 

Operation Cost or Relative Time Cost as I call it, is does not work particularly well on turbowarp, but works relatively well on vanilla scratch where most of the time is due to the virtual machine's js code running the blocks rather than your computers underlying speed of running operations. The whole idea of operation cost s that whilst the time to run a single block varies by machine the relative times of the scratch vm does not. It is fairly accurate for the operation blocks at least, where the uncertainty is around +-0.2 RTC, the other bocks are probably within +-1 RTC (except for some of the ones with really large RTCs)

As for yield blocks they currently are listed as N.A in the chart, and 0 when used for timing. so essentially those blocks don't have any operation cost. They are however measured by the timers, because they cost time,  it would to me be unintuitive for you to time the project and those blocks to not show up as using most of the time of the project. Although perhaps this could be setting.

As for the C blocks, well that's probably possible. We just show the user a c block but in reality we use these two function blocks as normal. But it would be more limiting. If you want to start the timer in one thread and end it in another thread then you can't do this with this. Perhaps we could allow users to right click the block and convert to a c block (which then makes the default block a c block). Another option is to make the input be the same as message blocks where you create new timer names, but then the issue is this is a little bit slower to use if you have lots of timers since what used to take 1 click now takes 3.

What do you think?

Tacodiva:
@Joeclinton1 

> It is fairly accurate for the operation blocks at least, where the uncertainty is around +-0.2 RTC, the other bocks are probably within +-1 RTC (except for some of the ones with really large RTCs).

How can I measure the RTCs myself? I feel like error margins should be included for the chart if it is going to be included.

> They are however measured by the timers, because they cost time, it would to me be unintuitive for you to time the project and those blocks to not show up as using most of the time of the project. Although perhaps this could be setting.

I mostly agree, but I still feel like there should be some kind of an indication of how much time is actually spent executing the blocks, rather then the yields. It's very weird to me that it's all lumped together, especially something designed for people who are already probably advanced Scratchers. Just seems like obvious extra information to include.

Tacodiva:
When I try to run that project and press 'o' to test the operator blocks in firefox (haven't tested with chromium) it gives 0.0ms for all of them:
image

Joeclinton1:
> When I try to run that project and press 'o' to test the operator blocks it gives 0.0ms for all of them. Same behavior in firefox and chrome. Am I doing something wrong? I see the "call count" increasing but no change in the timings. 

Well that's concerning

Joeclinton1:
> Also unrelated but I like how you can click the column titles to sort them, but it doesn't quite work how I'd expect. Right now if you click it once it sorts is descending, then clicking it again will sort it ascending, but clicking it a third time does nothing. I would expect it to switch back to sorting it descending and keep toggling back and fourth.

Third time, goes back to unsorted I think. It is the behaviour I wanted.

Joeclinton1:
I plan to remove RTC from this addon, in order to simplify things and get it out faster. The only things missing from this addon are additional things to make RTC more accepted, but if I simply get rid of it it solves the problem. I'll move the rtc version into a separate branch, just for me.

Joeclinton1:
Excluding yields without polluting the step thread is going to be a challenge, but i'll see if it's possible

addons/debugger/addon.json
```diff
@@ -79,14 +83,31 @@
...
+      "id": "auto_stop_timing",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "name": "Ratio Time instead of Percentage Time",
+      "id": "show_ratio_time",
+      "type": "boolean",
+      "default": false
+    },
```
Samq64, line 98: This is a duplicate setting. Was there supposes to be something else here?
```suggestion
```

Samq64 (commented):
Settings can now have descriptions which could be helpful here.

Joeclinton1:
> Settings can now have descriptions which could be helpful here.

Has been addressed

addons/debugger/timing/TimingManager.js
```diff
@@ -0,0 +1,53 @@
+class TimingManager {
+  constructor(settings, config, profiler) {
+    this.dummyProfiler = profiler;
+    this.settings = settings;
+    this.config = config;
+
+    this.timers = {};
```
Tacodiva, line 7: Nitpick, but this should probably be a `Map`. Because it's an object, creating a timer named `__proto__` will crash the VM.

addons/debugger/timing/ui-components/TableRows.js
```diff
@@ -0,0 +1,95 @@
...
+    this.minTimerTime = null;
+    this.debug = debug;
+    this.tableHeader = tableHeader;
+
+    this.placeholderElement.textContent = msg("timing-no-timers");
+  }
+
+  getRowValues(timer) {
+    const rowValues = {
+      label: timer.label,
```
Tacodiva, line 23: Is there anything that can be done about long labels causing overlapping text? Maybe it should go to an ellipsis
image

Tacodiva (changes requested):

addons/debugger/timing/Profiler.js
```diff
@@ -0,0 +1,73 @@
...
+      get() {
+        if (profiler.profilerActive) profiler.profile();
+        return null;
+      },
+    });
+
+    vm.runtime.sequencer.stepThread = function (...args) {
+      if (!propSet) {
+        // we define the property inside stepThread because initially there isn't an activeThread for us to use.
+        Object.defineProperty(Object.getPrototypeOf(this.activeThread), "blockGlowInFrame", {
```
Tacodiva, line 37: The debugger already hooks this when single stepping. This seems to cause problems when you single step in a thread where you're also trying to do timing line by line. I was able to crash the VM with this error:
```
Uncaught TypeError: can't redefine non-configurable property "blockGlowInFrame"
    singleStepThread moz-extension://05b7cf72-d36b-4bb2-8941-bdf96e32d77f/addons/debugger/module.js:232
    singleStep moz-extension://05b7cf72-d36b-4bb2-8941-bdf96e32d77f/addons/debugger/module.js:296
    createThreadsTab moz-extension://05b7cf72-d36b-4bb2-8941-bdf96e32d77f/addons/debugger/threads.js:206
 ```
 VM crashes are very bad because to the end user they stop projects from being able to run silently. I'm not sure the best way to fix this, but the way this is implemented currently makes it quite incompatible with single stepping. IIRC this hook can also cause a slight performance loss in a very hot path, so if possible you should be removing the hook when you don't need it (see module.js `singleStepThread`).

**Review by Tacodiva — CHANGES_REQUESTED** [review]

File: addons/debugger/timing/Profiler.js (hunk 5 above)
Comment (Joeclinton1) [reply addons/debugger/timing/Profiler.js:37]: I had previously identified the possible issue and had asked Garbo for assistance on identifying if the two addons were compatible but I guess it isn't so I will address this problem. I should be able to solve this.

**Comment by Joeclinton1** [comment]
I plan to fix this quickly now. Start your timers!

File: addons/debugger/timing/Profiler.js (hunk 5 above)
Comment (Joeclinton1) [reply addons/debugger/timing/Profiler.js]: I plan to fix this quickly now. Start your timers!

Comment (Joeclinton1) [reply addons/debugger/timing/Profiler.js]: So the UX here of doing both at the same time is that it makes no sense. You would never want to be profiling your code line by line whilst also stepping one at a time. Thus the solution is actually quite simple. Rather than make the two share the same stepThread, we will simply make it so that line by line profiling is paused whilst stepping, and that you cannot enable line by line profiling whilst in step mode (gray out icon and show tooltip explaining why can't be enabled). This will also require unpolluting stepThread.

**Comment by Joeclinton1** [comment]
I speed ran a fix for the VM crash when timing profiler conflicts with single-step debugger (single hour from reviewing code to commit/push!). After testing, both features are now compatible.

  **Root cause:** Both tried to define `blockGlowInFrame` property, causing "can't redefine non-configurable property" errors.

  **Solution:**
  - Added proper cleanup in `Profiler.unpollutStepThread()` to restore original VM state
  - Made profiler's property `configurable: true` to allow redefinition
  - Auto-disable line-by-line profiling during single-stepping (they're incompatible anyway)
  - Gray out line-by-line button with tooltip when unavailable

  Features now coexist without crashing the VM.

File: addons/debugger/timing/TimingManager.js (hunk 3 above)
Comment (Joeclinton1) [reply addons/debugger/timing/TimingManager.js]: this.timers = Object.create(null); will fix easily. Well spotted!

File: addons/debugger/timing/ui-components/TableRows.js (hunk 4 above)
Comment (Joeclinton1) [reply addons/debugger/timing/ui-components/TableRows.js]: image

File: addons/debugger/timing/TimingManager.js (hunk 3 above)
Comment (Joeclinton1) [reply addons/debugger/timing/TimingManager.js]:
```js
this.timers = Object.create(null);
```
image

Joeclinton1:
Can't believe it's been almost 1.5 years. And still not merged. I have taken some rather large breaks during the development of this haha.

Joeclinton1:
https://github.com/ScratchAddons/ScratchAddons/pull/8492 closing this pr would be helpful to fix the issue of there not being enough space the in the header.

addons/debugger/timing/TimingManager.js
Tacodiva: > this.timers = Object.create(null); will fix easily. Well spotted!

Yep that solution is fine too. My code style is I would prefer to use a `Map`, but no complaints with that :)

Tacodiva (dismissed):
Don't have time to review super thoroughly but I did give this a good look last time and all the issues I spotted have been fixed. LGTM

Joeclinton1:
This pr widens the timing addon to make room for the timing tab, but it's still not wide enough given the worst case scenario where you also are in the process of stepping through the code so there's a pause icon and you're using a language with long translations. It will need to be solved in a separate pr imo.

Samq64:
When open in line-by-line mode with no timers switching back to the timers view does not hide the column names. (And they always word wrap over the border for me)

image

In the second layer could the buttons be left-aligned to avoid shifting? Also the toggles might make more sense as checkboxes.

Joeclinton1:
> When open in line-by-line mode with no timers switching back to the timers view does not hide the column names. (And they always word wrap over the border for me)

I made some small css changes to fix this. 

> In the second layer could the buttons be left-aligned to avoid shifting? Also the toggles might make more sense as checkboxes.

I aligned it the way it is because I think it looks better this way. And I prefer it to not be checkboxes in this case.

Joeclinton1:
@Samq64 

> When open in line-by-line mode with no timers switching back to the timers view does not hide the column names. (And they always word wrap over the border for me)

I made some small changes to fix these two issues

> In the second layer could the buttons be left-aligned to avoid shifting? Also the toggles might make more sense as checkboxes.

I aligned it the way it is because I think it looks better this way. And I prefer it to not be checkboxes in this case: my reasoning is  we are swapping between line by line and timer mode, not choosing whether to include additional functionality. That would normally mean a toggle switch, not a checkbox.

image

mxmou:
I think we should decide what to do with #8216 before discussing this addon's UI design in detail.

Joeclinton1:
> I think we should decide what to do with #8216 before discussing this addon's UI design in detail.

Just approve the PR and then we can deal with the separate issue later.

Joeclinton1:
Idk if it was a good choice or not but I decided to rebase the entire thing ontop of the most recent version of master, since this PR is so old now. I probably should have just done the merge in retrospect.

At least now we can pretend it never diverged.

Joeclinton1:
image

Now that there's a footer there's ample space to move these tools buttons there.

Samq64:
Is it feasible to move everything to the bottom while leaving enough space and looking good?

Joeclinton1:
> Is it feasible to move everything to the bottom while leaving enough space and looking good?

Yes, because tools button can be removed. I will be implementing the checkbox for heatmap next and i'll give both icons so they're consistent.

addons/debugger/icons/performance.svg
Samq64, line 1: I'm not sure why this is showing up in the diff.

addons/debugger/timing/ui-components/tableHeader.js
```diff
@@ -0,0 +1,43 @@
+export function createTableHeader(config, msg) {
+  const tableHeader = Object.assign(document.createElement("div"), {
+    className: "sa-timing-timer sa-timing-header",
+    style: "display:none",
+  });
+
+  tableHeader.innerHTML = `
```
Samq64: Could innerHTML be avoided here?

addons/debugger/icons/performance.svg
Joeclinton1: Looks like code was reformatted accidently.

addons/debugger/timing/createTimingTab.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  function createExportButton() {
+    const exportButton = debug.createIconButton({
+      text: msg("export"),
+      icon: addon.self.dir + "/icons/download-white.svg",
+    });
+
+    exportButton.element.addEventListener("click", () => {
+      const csvContent = tableRows.rows.map((row) => Object.values(tableRows.getRowValues(row)).join(",")).join("\n");
```
Samq64: In line-by-line mode the file name and first column are mostly block IDs, so that should either be fixed or the button disabled in that mode.

Samq64 (commented):
The timing bugs out if a timer is repeatedly stopped in a loop. This happens with and without  the stop previous timing setting.

https://github.com/user-attachments/assets/ca813326-6186-4e51-9c07-5dd803b80c47

Samq64:
Also it would be nice if all heatmapped blocks had white text regardless of Scratch's color mode or `theme3`'s settings and if the heatmapping was re-applied when those change.

Joeclinton1:
image

Mostly for UI purposes I made it so there's not two modes but simply a checkbox called "profiling" that enables the line-by-line functionality and adds it to the existing timers.

Joeclinton1:
There's some kind of bug with the sorting, I will fix. Going to sleep now, so can't do anymore today.

Samq64:
Okay, here are some more (mostly UI) bugs that I'm commenting here before I forget:
- Footer buttons can overflow
- Misaligned sprite rows when expanding
- The table doesn't always use all vertical space
- Switching sprites doesn't keep the heatmap on

Joeclinton1:
There's not much space left in the footer but I feel it would really benefit from a clear or reset button, so that you can time things not from the start but midway through running a project.

addons/debugger/timing/ui-components/tableHeader.js
Joeclinton1: If you trust msg() to not contain malicious code then this is not a security issue.

Joeclinton1:
> The timing bugs out if a timer is repeatedly stopped in a loop. This happens with and without the stop previous timing setting.
> 
>  timing_bug.mp4

Well this would mean that a block is being missed, which is a different bug that probably affects more than just timers in loops.

Joeclinton1:
> Also it would be nice if all heatmapped blocks had white text regardless of Scratch's color mode or `theme3`'s settings and if the heatmapping was re-applied when those change.

I probably can fix this. Idk if we need to monitor changes and do the reapply. Seems like perfectionism.

addons/debugger/timing/createTimingTab.js
Joeclinton1: The first column being block ids isn't an issue. But I will look into it.

Joeclinton1:
> Okay, here are some more (mostly UI) bugs that I'm commenting here before I forget:
> 
> * Footer buttons can overflow
> * Misaligned sprite rows when expanding
> * The table doesn't always use all vertical space
> * Switching sprites doesn't keep the heatmap on

These can be fixed I'm sure. The last one I'm aware of. We'd need to reapply the heatmap on sprite change.

addons/debugger/timing/ui-components/tableHeader.js
Samq64: It's mostly for store reviews (https://github.com/ScratchAddons/ScratchAddons/issues/173#issuecomment-685974738) but there is `safeMsg` if you really want to use that here.

Joeclinton1:
Since @Samq64 left alot of review changes, for my sake, and for the sake of people following along with progress. I'm going to list them as tasks:

**Non-critical**:
- [x] Avoid the use of inner html, to create the table header, instead building dynamically.
- [x] Make the line-by-line timers use the block label not block id in the export
- [x] Make the heatmap compatible with scratch's other colour mode's and theme3 settings by always using white colour and reapplying if there are changes.
- [x] Make the footer buttons move onto a second row if the window is too small to prevent overflow
- [x] Switching from one sprite to another needs to reapply the heatmap.
- [x] Add a clear button to remove all timers
- [x] Fix bug where clicking table headers to sort only works for first click
- [x] Fix bug where the debugger block previews don't match the stack block's because of "undefined" data shape
- [x] Fix bug where a stop timer placed after another stop timer causes erratic time accumulation as we don't track if the timer is active or not
- [x] Fix bug where if profiling was previously enabled, heatmap enabled then profiling disabled, disabling heatmap doesn't reset blocks
- [x] Fix bug where if you enable profiling, start project, enable heatmap, then disable profiling, code blocks will have text set to white even if background not coloured.
- [x] Make it so that the heatmap is modified in realtime if project started whilst heatmap visible.
- [x] Add slider that lets you change the maximum of the heatmap to the button (used to exist before ui redesign)
- [x] Exported table needs headers
- [ ] ~~Fix bug where loops are counted one extra time on exit~~

**Critical**:
- [x] Fix bug where blockGlowInFrame is only redefined for prototype but not for current instances, so it works on second run through, but not initial where the current thread instances still use old definition.

**Unrelated Bug**:
There is a bug where the debugger blocks are not being coloured the first time you open the project editor, due to workspace not loading in time, but this is present on master branch too.

Joeclinton1:
> * Misaligned sprite rows when expanding
> * The table doesn't always use all vertical space

@Samq64 I can't figure out what you mean by these can you please provide screenshots?

Joeclinton1:
- [ ] Fix a bug causing blocks to be skipped when stopping and starting project for unknown reason
- [ ] Fix a bug where initially not all blocks are being timed on first click of green flag, after enabling profiling

These are likely the same bug. I think there is a flaw in the method I'm using for the profiling when it comes to loops.

Joeclinton1:
There is a bug where repeats and repeat until will get counted a second time on exit, but technically that's exactly what scratch does so it's correct. Idk if it's a bug.

Other than that, I've fixed everything.

addons-l10n/en/debugger.json
```diff
@@ -36,5 +40,17 @@
...
+  "debugger/timing-hide-rtc": "Hide Operation Cost",
+  "debugger/timing-heatmap": "Heatmap",
+  "debugger/timing-label": "Label",
+  "debugger/timing-total-time": "Total Time",
+  "debugger/timing-average-time": "Average Time",
+  "debugger/timing-ratio-time": "Ratio Time",
+  "debugger/timing-percent-time": "Percent Time",
+  "debugger/timing-call-count": "Call Count",
+  "debugger/timing-no-timers": "No timers started.",
+  "debugger/timing-disabled-single-step": "Cannot profile line-by-line while single-stepping"
```
mxmou: Unused

addons/debugger/icons/flame.svg
mxmou, line 1: We usually optimize SVGs using svgo.

addons/debugger/timing/createTimingTab.js
```diff
@@ -0,0 +1,357 @@
...
+
+    // Create slider track
+    const sliderTrack = Object.assign(document.createElement("div"), {
+      className: "sa-timing-heatmap-slider-track",
+    });
+
+    // Create slider thumb
+    const sliderThumb = Object.assign(document.createElement("div"), {
+      className: "sa-timing-heatmap-slider-thumb",
+    });
```
mxmou, line 84: Could the slider code be simplified by using an input with `type=range` and styling it with CSS?

addons/debugger/timing/HeatmapManager.js
```diff
@@ -0,0 +1,170 @@
+function valueToHeatmapColor(value, min = 0, max = 1) {
+  const clampedValue = Math.max(min, Math.min(max, value));
+  const normalizedValue = (clampedValue - min) / (max - min);
```
mxmou: This will be NaN if `max - min` is zero, which causes everything to be black if the slider is set to zero:
Image

addons-l10n/en/debugger.json
```diff
@@ -36,5 +40,17 @@
   "debugger/log-msg-list-append-too-long": "Failed to add to list '{list}', cannot add over 200,000 items.",
   "debugger/log-msg-list-insert-too-long": "Item 200,000 removed from list '{list}' to make room for insertion, cannot add over 200,000 items.",
   "debugger/log-cloud-data-nan": "Failed to set cloud variable '{var}'. Cloud variables can only contain numbers.",
-  "debugger/log-cloud-data-too-long": "Failed to set cloud variable '{var}'. Cloud variables can only contain a maximum of 256 digits."
+  "debugger/log-cloud-data-too-long": "Failed to set cloud variable '{var}'. Cloud variables can only contain a maximum of 256 digits.",
+  "debugger/timing-profiling": "Profiling",
+  "debugger/timing-show-rtc": "Show Operation Cost",
+  "debugger/timing-hide-rtc": "Hide Operation Cost",
```
mxmou: These seem to be unused.

mxmou (commented):
The table looks broken after a clone has been deleted:
Image

When profiling is enabled, blocks from other sprites are sometimes shown as just the sprite name. For example, in [this project](https://scratch.mit.edu/projects/1173154181/editor) after starting profiling and waiting until the loading screen finishes:
Image

A few UX suggestions:
<ul>
<li>Checkboxes should be on the left.<br />
Image</li>
<li>I don't like how the whole footer expands to make room for the heatmap slider. The button could have a negative vertical margin to avoid that:<br />
Image</li>
<li>This would require changes to the `addon.tab.addBlock` API, but it would be nice if the input field on "start timer" and "stop timer" blocks contained some text by default ("timer1"?). Alternatively, make the blocks do something even if the input is empty.</li>
</ul>

addons/debugger/icons/flame.svg
Joeclinton1: If you know how to do it, could you just edit my pr.

Joeclinton1:
> * Checkboxes should be on the left.

I disagree, for consistency. with the icon, I have placed it on right side. Only within a vertical layout would this really matter.

addons/debugger/icons/flame.svg
mxmou: Done.

Joeclinton1: Thanks :)

mxmou:
> > * Checkboxes should be on the left.
> 
> I disagree, for consistency. with the icon, I have placed it on right side. Only within a vertical layout would this really matter.

Placing them on the left is the standard, and UIs that follow conventions like that are generally easier to understand. Another option is to make them look like toggle switches.

Joeclinton1:
Optimization of svg's has caused a bug with the icons stretching

mxmou:
> Optimization of svg's has caused a bug with the icons stretching

Fixed.

addons/debugger/timing/createTimingTab.js
Joeclinton1: Can't get css quite right (can't get the slider width to be correct without hard coding) and hard to not have this cause cross browser issues

Joeclinton1:
> When profiling is enabled, blocks from other sprites are sometimes shown as just the sprite name.

It's unrelated to the block being in a different sprite. It's the stop block not generating a block preview for the label. This is a separate bug to do with the block previews, not caused by timing tools tab. 

```js
 var jsonData;
const fakeBlock = {
  workspace: addon.tab.traps.getWorkspace(),
  jsonInit(data) {
    jsonData = data;
  },
};
const blockConstructor = ScratchBlocks.Blocks[block.opcode];
if (blockConstructor) {
  try {
    blockConstructor.init.call(fakeBlock);
  } catch (e) {
    // ignore
  }
}
if (!jsonData) {
  return null;
}
  ```
Basically it failed to get json data.

it fails because CONTROL_STOP block doesn't just return jsonInit, instead it calls a load of functions to procedurally generate the block.

---

This can easily be fixed with another case in Block Preview for specifically this block

**Comment by Joeclinton1** [comment]
> Placing them on the left is the standard, and UIs that follow conventions like that are generally easier to understand. Another option is to make them look like toggle switches.

Ok I did it.

**Comment by Joeclinton1** [comment]
@Samq64 

> Misaligned sprite rows when expanding

Still don't know what this means, but maybe it's fixed from one of the recent commits.

> The table doesn't always use all vertical space

This is a bug from the resizing PR. As we aren't resizing the dynamic log-view's number of rows to include. I have fixed it now

**Comment by Samq64** [comment]
> I disagree, for consistency. with the icon, I have placed it on right side.

Maybe the icon could be removed then, or would it not be obvious it's a checkbox?

> Still don't know what this means, but maybe it's fixed from one of the recent commits.

Seems to be from my quick testing.

**Comment by Joeclinton1** [comment]
> Maybe the icon could be removed then, or would it not be obvious it's a checkbox?

You could. Looks like:

image

vs

image

definitely cleaner in the version without the icon

---

Atm, i'm looking at it, and the choice to move to left side has made it uglier

**Comment by mxmou** [comment]
> image

I think that looks better than both versions with the icon.

**Comment by Joeclinton1** [comment]
@mxmou I removed the icons. Surely this is the last ui change needed right?

**Review by mxmou — APPROVED** [review]
> @mxmou I removed the icons. Surely this is the last ui change needed right?

Just one small thing: I suggest adding `height: 16px` to the buttons with checkboxes (except the heatmap one when expanded). That way, they would have the exact same height as the buttons with icons.

**Comment by Joeclinton1** [comment]
> Just one small thing: I suggest adding `height: 16px` to the buttons with checkboxes (except the heatmap one when expanded). That way, they would have the exact same height as the buttons with icons.

Yes good idea, have made this change. Thank you for approving!

**Comment by Joeclinton1** [comment]
@Tacodiva Please review so I don't have to think about this PR anymore.

File: addons/debugger/addon.json [hunk 12]
```diff
@@ -79,13 +83,27 @@
...
+      "default": false,
+      "description": "Replaces the percentage column with a ratio of each block’s total time to the smallest block’s total time."
     }
   ],
   "tags": ["editor", "codeEditor", "featured"],
   "versionAdded": "1.16.0",
   "latestUpdate": {
     "version": "1.44.0",
-    "temporaryNotice": "The debugger window is now resizable."
+    "temporaryNotice": "New timing tab to help you with micro-optimisation of your project and the debugger window is now resizable."
```
Samq64:
```suggestion
    "temporaryNotice": "New timing tab to help you with micro-optimization of your project and the debugger window is now resizable.",
    "newSettings": ["auto_stop_timing", "show_ratio_time"],
    "isMajor": true
```

Joeclinton1: what is this anti uk spelling slander

Joeclinton1: I'll do it, but reluctantly, haha

Joeclinton1:
@Samq64 for changes that small please just edit the PR, to save time.

addons/debugger/addon.json
Samq64: I thought using US English was in the docs somewhere but I can't find it right now.

mxmou: > I thought using US English was in the docs somewhere but I can't find it right now.

It should probably be added to the docs then. We've always used American spelling, which make sense because Scratch also uses it.

Joeclinton1: So @Samq64 I see you added Status: needs review, but you could review it yourself seeing as you already asked for changes so you must have looked at it.

Samq64: I can test it again and approve but I don't have the knowledge to properly review everything here.

Joeclinton1: @Samq64 Hmm, ok, this is why I'd like @Tacodiva to review. @GarboMuffin would have the knowledge for this too.

Tacodiva (approved):
Looks good to me :)

Joeclinton1:
> Looks good to me :)

Thanks for the review :)

Joeclinton1:
Thank you everyone for your time, and reviews. This has been a very long process and monumental effort, but now that it's approved (and I believe @Samq64 also approves based on feedback) I am going to merge this. I'm looking forward to the release to production!

Joeclinton1:
I noticed a bug in the timing tools: my css selection for textElements to force to be white (needed for when users have high contrast on, which has black text which doesn't contrast well with heatmap enabled) will exclude certain text elements meaning not all text gets changed to white.

I need to change lines 14-16 of addons\debugger\timing\HeatmapManager.js from 
```js
const textElements = Array.from(block.svgGroup_.children)
    .filter((el) => !el.classList.contains("blocklyDraggable"))
    .flatMap((el) => Array.from(el.querySelectorAll("text")));
```

to

```js
  const textElements = block.svgGroup_.querySelectorAll(
    ":scope > :not(.blocklyDraggable) text, :scope > text:not(.blocklyDraggable)"
  );
```

Issue #6721, opened 2023-09-30
Coding experience checklist


DNin01:
As a welcoming open-source project, it's important to make sure contributors have a good experience. Enhancing the process of writing code with guidance along the way is one step closer.

- [ ] Code style preferences out of the box: #6621
- [ ] Explain how to open a folder in Visual Studio Code so the settings apply
- [ ] Code snippets for quick code autocomplete: #6641
- [ ] Recommend helpful extensions: #6593
- [ ] JSDoc types to explain `addon.*` APIs: #6488
- [ ] Make directories belonging to addons easier to search instead of having to scramble to find the addon ID you're looking for

PR #7462, opened 2024-05-23
Css Nesting Build Script

TheColaber:
Creates a package.json command `yarn transpile-css`. When run, it scans all CSS and HTML files and replaces any code that uses nested CSS with plain CSS. This is for compatibility support. Developers will be able to write all CSS using nested CSS if desired and it will still work for the user's devices. Note: This command probably won't be used by developers, it will likely be used when we are sending the extension to web stores.

I made 2 test files, [webpages/settings/components/addon-group-header.html](https://github.com/ScratchAddons/ScratchAddons/compare/master...TheColaber:css-nesting?expand=1#diff-dc06b1b02ec1628e6aeba560f44f9194b295a39fdc3987ad77de4a54d3860790) and [webpages/styles/components/badges.css](https://github.com/ScratchAddons/ScratchAddons/compare/master...TheColaber:css-nesting?expand=1#diff-dfba69e2cf170012d6cf699e2173b6489fa37a5f8fe255d16b29cb18e6c3f09f). Run the package command and the files will get replaced.

I considered using an HTML parser library rather than using regex to find `<style>` elements but I decided not to since most of the time it alters the HTML especially when it comes to using self-closing tags. We should discuss this further since I'm not really sure of the best route.

Also please keep in mind that I have no idea if this is how I should have done things, but now we have somewhere to start at least :)

@WorldLanguages told me to work on this.


webpages/settings/components/addon-group-header.html
```diff
@@ -6,38 +6,38 @@
...
-    transition: background-color 0.2s ease;
-  }
+    &.margin-above {
+      margin-top: 20px;
+    }
 
-  .addon-group:hover > img {
-    background-color: var(--hover-highlight);
-  }
+    & > img {
```
WorldLanguages, line 20: Personally I find it quite confusing to know when the ampersand is needed and when it's not. Since Chrome 120+ and Firefox 117+ it is needed in fewer places, see: https://developer.chrome.com/blog/css-nesting-relaxed-syntax-update

WorldLanguages:
If this is merged with a package.json file, we should look for other issues/PRs that suggest this file should exist.

Samq64:
#6272 has its own `package.json` in the the `webpages` directory and uses npm instead of yarn.

WorldLanguages:
I mean, mxmou doesn't need to review the build script, but I'm wondering if it's worth exploring this, as this may simplify writing CSS.
I just checked the original [scratchr2.user.css](https://github.com/mxmou/customize-scratch/blob/master/scratchr2.user.css) and it does not use nesting, though. I don't know if that's intentional or if the preprocessor didn't support it.

DNin01:
> If this is merged with a package.json file, we should look for other issues/PRs that suggest this file should exist.

I have a related task in the checklist of https://github.com/ScratchAddons/ScratchAddons/issues/6721.

Hans5958:
One step closer to build step...

I prefer `npm` over `yarn`, since that's what we used around here. ([ScratchAddons/contributors](https://github.com/ScratchAddons/contributors), [ScratchAddons/website-v2-script](https://github.com/ScratchAddons/website-v2-script), [ScratchAddons/l10n-script](https://github.com/ScratchAddons/l10n-script))

mxmou:
> I just checked the original [scratchr2.user.css](https://github.com/mxmou/customize-scratch/blob/master/scratchr2.user.css) and it does not use nesting, though. I don't know if that's intentional or if the preprocessor didn't support it.

I don't remember why I didn't use nesting - probably because I was copying selectors from Scratch's styles. Nested selectors would certainly be useful.

GrahamSH-LLK:
This should be be able to be solved within vite fairly easily (eg. using built in lightningcss support)
I can add it to my PR?

Hans5958:
> This should be be able to be solved within vite fairly easily (eg. using built in lightningcss support) I can add it to my PR?

Would be nice, but then I don't think your PR will be merged in the near future.

GrahamSH-LLK:
> > This should be be able to be solved within vite fairly easily (eg. using built in lightningcss support) I can add it to my PR?
> 
> Would be nice, but then I don't think your PR will be merged in the near future.
> 
> I would say we can do this first then replace it later if needed.

well my PR works quite well and really _could_ be merged 😛 

obviously we need to get mv3 done before anything else but i really do think it's ready to merge.

Samq64:
CSS nesting will be considered widely available on MDN in June and Scratch is already using it with no polyfill in a few places.

DNin01:
> CSS nesting will be considered widely available on MDN in June (...)

Really!

Scratch Addons already requires developers to use Chrome 121+ or Firefox 121+, both of which support CSS nesting, so for addon developers we may not have to offer a development build script at all—it would only be necessary for production builds.

WorldLanguages:
> Scratch Addons already requires developers to use Chrome 121+ or Firefox 121+, both of which support CSS nesting, so for addon developers we may not have to offer a development build script at all—it would only be necessary for production builds.

Yes, I believe that was always the idea we had.

However if Scratch is already shipping modern CSS to clients, then there's no need for us to polyfill it.

Samq64:
In Chromium 96 the only obvious unsupported CSS is in the debugging tips dialog, everything else looks okay at a glance.

WorldLanguages:
Hmm then forget what I just said

PR #7464, opened 2024-05-24
Fix MV3 on Firefox for Android

### LINKED ISSUE #6976: Popup very zommed out on Firefox Android
### Description

On Firefox for Android the popup is scaled incorrectly so everything is too small.

### Steps to reproduce

1. Install on Firefox for Android
2. Switch to portrait mode
3. Open Add-ons > Scratch Addons

Screenshot_20231218-083418_Firefox

### Environment

Browser: Firefox for Android 120
Operating system: Android 11
Screen: 8.0" (800x1260)

### LINKED ISSUE #7461: Doesn't work on Android Firefox since the recent update
### Description

Since the recent update to the extension, it is not working for me on my Android Mobile running Firefox 126.

Earlier, it used to work flawlessly but now the settings page doesn't work and I can't enable any addons.

### Steps to reproduce

1. Install the extension on Firefox for Android in Android devices
2. Try to open the extension's settings and search for any addon and try to enable it

_No response_

### Environment

Browser: Firefox for Android - Version: 126
Scratch Addons version: 1.38.0
Operating system: Android 14

Samq64:
Resolves #6976
Resolves #7461

### Changes

Sets the viewport on the popup and checks if `chrome.contextMenus` is defined before using it since it errors on Firefox for Android now.

Also removes some weird icon handling I added back in #4810. Now it always uses the icons set in the extension's manifest.

### Reason for changes

The popup wasn't scaling correctly and after MV3 the extension completely broke. It's also nice to be able distinguish Git versions even if they're stable.

### Tests

Tested on Andorid Firefox Beta, desktop Firefox and Chromium. The viewport doesn't seem to change the popup on Desktop and Firefox works on Android again including Scratch notifier.

### DIFF
--- a/background/handle-notifications.js
+++ b/background/handle-notifications.js
@@ -32,26 +32,24 @@ chrome.storage.local.get("muted", (obj) => {
   scratchAddons.muted = obj.muted;
 });
 
+chrome.contextMenus?.removeAll();
 let currentMenuItem = null;
 
-// chrome.contextMenus is broken on Android Firefox with MV3
-if (chrome.contextMenus !== undefined) {
-  chrome.contextMenus.removeAll();
+// NOTE: chrome.contextMenus equals `undefined` on Firefox for Android!
 
-  chrome.contextMenus.onClicked.addListener(({ parentMenuItemId, menuItemId }) => {
-    if (parentMenuItemId === "mute") {
-      const mins = Number(menuItemId.split("_")[1]);
-      contextMenuMuted();
-      muteForMins(mins);
-    } else if (menuItemId === "unmute") {
-      contextMenuUnmuted();
-      unmute();
-    }
-  });
-}
+chrome.contextMenus?.onClicked.addListener(({ parentMenuItemId, menuItemId }) => {
+  if (parentMenuItemId === "mute") {
+    const mins = Number(menuItemId.split("_")[1]);
+    contextMenuMuted();
+    muteForMins(mins);
+  } else if (menuItemId === "unmute") {
+    contextMenuUnmuted();
+    unmute();
+  }
+});
 
 function contextMenuUnmuted() {
-  if (chrome.contextMenus === undefined) return;
+  if (chrome.contextMenus === undefined) return; // Firefox for Android
   if (currentMenuItem === "unmute") chrome.contextMenus.remove("unmute");
   currentMenuItem = "mute";
   chrome.contextMenus.create({
@@ -76,7 +74,7 @@ function contextMenuUnmuted() {
 }
 
 function contextMenuMuted() {
-  if (chrome.contextMenus === undefined) return;
+  if (chrome.contextMenus === undefined) return; // Firefox for Android
   if (currentMenuItem === "mute") chrome.contextMenus.remove("mute");
   currentMenuItem = "unmute";
   chrome.contextMenus.create({
--- a/background/handle-notifications.js
+++ b/background/handle-notifications.js
@@ -75,6 +75,8 @@ function contextMenuUnmuted() {
 
 function contextMenuMuted() {
   if (chrome.contextMenus === undefined) return; // Firefox for Android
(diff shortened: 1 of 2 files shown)


background/handle-notifications.js
```diff
@@ -2,6 +2,9 @@ import { updateBadge } from "./message-cache.js";
 
 const BROWSER_ACTION = globalThis.MANIFEST_VERSION === 2 ? "browser_action" : "action";
 
+// chrome.contextMenu is broken on Firefox mobile
+const isAndroidFirefox = navigator.userAgent.includes("Firefox") && navigator.userAgent.includes("Android");
```
WorldLanguages, line 6: I think it's better to change this to `chrome.contextMenus === undefined`

Samq64:
To test on Android:
1. Install ADB and web-ext
2. On Android install Firefox beta
3. Enable the developer options by repeatedly tapping the build number in settings
4. Enable USB debugging in the developer options
5. In Firefox beta enable remote debugging via USB
6. Completely close Firefox
7. Plug the Android device into the computer and tap allow every time it asks
8. Open a terminal in Scratch Addons
9. Run `adb devices`
11. Run `web-ext run -t firefox-android --adb-device ABC123 --firefox-apk org.mozilla.firefox_beta` with `ABC123` being the device ID from the previous command
12. The extension should install. Tap OK.
13.  If a purple message appears has to be dismissed since it steals all focus.

BroJac5246:
> On Android install Firefox beta

This also works on regular Firefox, you just have to change the package name in the web-ext command to `org.mozilla.firefox`.
They need a better way to do this than using ADB but at least it works...

Anyhoo, can confirm the extension now works (Firefox for Android Stable 126.0; Android 14 QPR2 on Google Pixel 7a). I can't test everything but the extension loads correctly and the addons I try seem to work correctly (at least as well as they can on mobile).

The initial permissions page didn't display correctly in either orientation but was usable exclusively in portrait mode. I don't know if that's a new development, however. Either way, we should get this out as soon as possible as long as it's confirmed this won't break anything on desktop browsers.

WorldLanguages:
Thanks for figuring out how to test on mobile Firefox, I wasn't that lucky as I got some weird error when trying to run the web-ext command.

webpages/popup/index.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
WorldLanguages, line 5: Could this cause some issues on desktop? 🤔 The size of the popup has frequently created problems.
Maybe to be on the safe side we should add this dynamically with JavaScript if the user agent matches mobile Chromium or mobile Firefox?

background/handle-notifications.js
```diff
@@ -32,21 +32,26 @@ chrome.storage.local.get("muted", (obj) => {
...
-    const mins = Number(menuItemId.split("_")[1]);
-    contextMenuMuted();
-    muteForMins(mins);
-  } else if (menuItemId === "unmute") {
-    contextMenuUnmuted();
-    unmute();
-  }
-});
+// chrome.contextMenus is broken on Android Firefox with MV3
+if (chrome.contextMenus !== undefined) {
```
WorldLanguages, line 38: I'll change this to use `?.` syntax instead.

webpages/popup/index.html
Samq64: I haven't noticed any problems so far and Vue 3 also adds it but it's better to be safe I guess.

mxmou: Unless it's handled differently in extension popups for some reason, the viewport tag has no effect on desktop.

WorldLanguages: Then let's assume this didn't break anything unless we've got proof otherwise

WorldLanguages (approved):

PR #7468, opened 2024-05-25
Error in settings-page if background ctx unavailable

### LINKED ISSUE #5306: Allow more control when Scratch Addons fails to load

WorldLanguages:
Resolves #5306

### Changes

image

- The error page is still not localized - so it's English only
- The popup still doesn't show any error message if the background context is unavailable. 

### Tests

Tested in Chrome and Firefox on MV3

### DIFF
--- a/background/background-sucess-status.js
+++ b/background/background-sucess-status.js
@@ -1,0 +1,6 @@
+// If this script is running, then none of the previews scripts threw any errors.
+
+chrome.storage.session?.set({ backgroundContextOk: true });
+// Ideally we would only run this once per service worker startup, but
+// there's no straight-forward way it seems. A combination of
+// onStartup and onUpdate might not cover all cases.
--- a/background/background.js
+++ b/background/background.js
@@ -30,3 +30,5 @@ import "./get-popups.js";
 // Respond to requests by the settings page and others
 import "./handle-settings-page.js";
 import "./handle-licenses.js";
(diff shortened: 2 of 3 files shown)


Samq64 (approved):

Samq64:
Could the option to export settings be uncommented? I'm going to PR some styles for the page so I could also include it there.

WorldLanguages:
I think I want to change the "Extension settings are corrupted or temporarily unavailable" wording in this case, and add the export button as Samq said.

Samq64:
> Could the option to export settings be uncommented?

Replying to myself, for unsupported browsers the buttons were added to the existing page on the website instead of using this one.

WorldLanguages:
@Samq64 It would be nice to allow users to export their settings in this screen.

PR #7474, opened 2024-05-28
`localhost` fixes

### LINKED ISSUE #7455: Addons don't work when running scratch-www locally
### Description

[The `matches` array for `web_accessible_resources`](https://github.com/ScratchAddons/ScratchAddons/blob/e6f38f6/manifest.json#L77) should include `localhost` and `scratchfoundation.github.io` if the extension is installed from source. It currently doesn't, which breaks userstyles without `"injectAsStyleElt": true` and all userscripts.

### Steps to reproduce

Set up a scratch-www development server and visit `localhost:8333`.

image

### Environment

Browser: Edge 126 dev
Scratch Addons version: e6f38f6
Operating system: Windows

### LINKED ISSUE #7456: Run on `localhost:8602`
### Why this would be helpful

scratchfoundation/scratch-gui#9536 changed the port for the development server from 8601 to 8602.

### My suggestion

Add the necessary match and host permission to run Scratch Addons on `localhost:8602`.

### Possible alternatives

_No response_

### Additional context

_No response_

mxmou:
Resolves #7455
Resolves #7456

### Changes

* Adds `localhost:8602` to `host_permissions` and updates other code that expects the GUI port to be 8601
* Replaces `localhost:8333` and `localhost:8601` content script matches with just `localhost` - Firefox doesn't like ports in match patterns for some reason
* Adds `localhost` and `scratchfoundation.github.io` to the `matches` field under `web_accessible_resources`
* **For up-to-date changes, see comment https://github.com/ScratchAddons/ScratchAddons/pull/7474#issuecomment-2197876142**

@WorldLanguages The `gen-manifest` script needs to be updated.

### Tests

Tested on Edge and Firefox with the develop branch of scratch-gui and scratch-www.

### DIFF
--- a/.github/gen-manifest.mjs
+++ b/.github/gen-manifest.mjs
@@ -2,9 +2,13 @@ const PERMISSIONS_IGNORED_IN_CHROME = ["clipboardWrite", "webRequestBlocking"];
 const PERMISSIONS_IGNORED_IN_FIREFOX = [];
 // These host permissions below should be removed during production manifest gen.
 const PERMISSIONS_ALWAYS_IGNORED = [
+  "scripting",
   "https://scratchfoundation.github.io/scratch-gui/*",
+  "https://scratchfoundation.github.io/*",
   "http://localhost:8333/*",
   "http://localhost:8601/*",
+  "http://localhost:8602/*",
+  "http://localhost/*",
 ];
 
 /**
@@ -24,8 +28,8 @@ export default (env, manifest) => {
   manifest.icons["16"] = "images/icon-16.png";
 
   const removePermission = (permToRemove) => {
-    // Affects `permissions`, `host_permissions`, `optional_permissions`, `optional_host_permissions`
-    // as well as content script matches.
+    // Affects `permissions`, `host_permissions`, `optional_permissions`, `optional_host_permissions`,
+    // `web_accessible_resources[i].matches`, and content_scripts[i].matches
     manifest.permissions = manifest.permissions.filter((perm) => perm !== permToRemove);
     manifest.host_permissions = manifest.host_permissions.filter((perm) => perm !== permToRemove);
     manifest.optional_permissions = manifest.optional_permissions.filter((perm) => perm !== permToRemove);
@@ -34,6 +38,9 @@ export default (env, manifest) => {
     }
     manifest.content_scripts.forEach((contentScript) => {
       contentScript.matches = contentScript.matches.filter((match) => match !== permToRemove);
+    });
+    manifest.web_accessible_resources.forEach((resourcesObj) => {
+      resourcesObj.matches = resourcesObj.matches.filter((match) => match !== permToRemove);
     });
   };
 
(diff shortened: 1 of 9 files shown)


mxmou:
> Hmm, looks like Firefox does not behave like Chrome does with localhost ports. Is this a Firefox MV3 bug or why didn't we catch it before?

It doesn't work in MV2 either, so it was either broken by a Firefox update or, more likely, nobody tried to test a local scratch-www build in Firefox.

Samq64 (approved):

WorldLanguages:
> Is there no better way? Our content scripts will run no matter the port.

I was thinking, maybe we should prevent our content scripts from running if the port doesn't match. We already early-return in SVG documents and cross-origin iframes, this would be a third possible reason.

The motivation is that a lot of people running the dev version of Scratch Addons also develop websites that run in any other ports, and it could be quite unexpected to be running our code in any localhost website...

WorldLanguages:
Here is a possible way to do this.

1. manifest.json should always reference the port within content_scripts, which will be enough to support Chrome. (it's fine to ask for host permission for all ports)
2. For Firefox support, the `browser.scripting.registerContentScripts` API ([see MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/registerContentScripts)) should be used with the broader `http://localhost/*` match, only if `typeof browser !== "undefined"`.
3. Additionally, early return if the port doesn't match, which will only happen on Firefox.

I can work on this, just tell me if it's ok.

mxmou:
> Here is a possible way to do this.
> 
> 1. manifest.json should always reference the port within content_scripts, which will be enough to support Chrome. (it's fine to ask for host permission for all ports)
> 2. For Firefox support, the `browser.scripting.registerContentScripts` API ([see MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/registerContentScripts)) should be used with the broader `http://localhost/*` match, only if `typeof browser !== "undefined"`.
> 3. Additionally, early return if the port doesn't match, which will only happen on Firefox.

Sounds good.

WorldLanguages:
mxmou please test again in both browsers if you have some time (just check that you can still test local Scratch builds properly after my changes - everything else is on me)

DNin01:
In addition to only running on specific ports, I think we should also check for clues that we are running on a Scratch page. For example, there may be certain elements we could check for the existence of to confirm that it's a Scratch page. This might cause an unstyled content flash the first visit, but after that, we could remember that that port is running an instance of Scratch.

WorldLanguages:
The ports used by scratch-www and scratch-gui are too rare for that to be worth it.

mxmou:
> mxmou please test again in both browsers if you have some time (just check that you can still test local Scratch builds properly after my changes - everything else is on me)

Still works on both browsers.

WorldLanguages (approved):

WorldLanguages:
Ok, I tested gen-manifest.mjs and it works as expected

rickbot-dot:
revert this its breaking the extension on firefox

rickbot-dot:
never mind there is a new pr to fix this

PR #7475, opened 2024-05-28
PoC: transpile commonly requested blocks

pufferfish101007:
This is a proof of concept which, once fleshed out a bit, could resolve various suggestions on discord, and things like #7002

### Changes

This is a proof of concept which currently adds 3 new blocks to the flyout: operator_lte, operator_gte & looks_previouscostume. These are not 'hacked blocks' that are hidden from the flyout - these are new blocks which are transpiled away to vanilla scratch when you save the project, and are lazily transpiled back to these shorter representations when you edit a target with the addon enabled. The addon also monkey-patches the VM whilst the addon is on in order to support these new blocks without transpiling to vanilla scratch and back before running a project. Some blocks have multiple ways of being represented in vanilla scratch, so I've made a half-hearted effort to support these. There is no option to view these block combinations in vanilla scratch; if you have the addon enabled, as soon as you press 'save now', and block combinations that can be shortened will be shortened.
This should be fairly easy to extend to add more new blocks.

### Reason for changes

People want it apparently. Whether it's a good idea or not is certainly questionable.

### Known issues

- At the moment, when transpiling to the scratch addons versions of the blocks, any blocks that become unused are not discarded, so you end up with a lot of extra blocks which are never seen clogging up your project.json. Should be easily fixable.
- shadow blocks are also discarded so if you remove inputs from these blocks you end up with scary holes with Nothing In Them
- there are a heck of a lot of ways of writing 'previous costume' as a `join` block so I just didn't before trying to account for all these cases

### Tests

Tested in chrome 125; it works, with some unexpected behaviour as described above.
You can try an example project out at https://scratch.mit.edu/projects/1027863622/editor/


pufferfish101007:
oh lordy lord the merge conflicts are insane

ZXMushroom63:
@pufferfish101007 if you want some implementations, I built a custom scratch editor built entirely around the concept of transpiling to vanilla. It would be really cool to see some of these as part of an addon, even though some may be a bit overcomplicated.

https://github.com/ZXMushroom63/scratch-vm/blob/develop/src/serialization/sb3/sbpp2sb3.js #transpiler code
http://github.com/ZXMushroom63/scratch-gui/releases/latest #downloadable editor demo
https://scratch.mit.edu/projects/960924608 #example transpiled project (try pressing see inside :p)

So far I've built transpilers for: griffpatch's Impulse physics engine, string starts with, string ends with, color at current position, replace a with b in x, ternary (`<> ? a : b`), clamp x between l and h, to hexadecimal, sprite fencing controls, split string, substring, minimum, maximum, and a power operator.

Not sure how this would pan out as an extension, but it is possible to make blocks invisible and delete them if they aren't used, so at least for statement blocks, you could create hidden custom block definitions for polyfilling, using monkey patching to run code on the caller, and delete them if they aren't used anymore.

More complex reporters are probably next to impossible, unless you find a way to switch between a stack-respecting vanilla transpile and raw addon code seamlessly, which I can't really see happening in a neat way.

image
image

pufferfish101007:
@ZXMushroom63 that's a cool scratch editor and an interesting idea. However, I think a lot of what you're describing would work nicely with a combination of #8674 and the backpack, and I don't think prebaking complicated transpiled blocks into the addon is a good idea: they can add hidden complexity and there's not necessarily one best way of implementing them.

I think that in order for this addon to be useful, I would want to either:
- change to allow users to define their own transpiled block sets (which could then be shared if we had per-addon exportable settings), but frankly that seems like it would lead to a very very confusing settings page, and I'm not sure anyone would really want to take the time to set that up
- or change to only provide blocks that cannot be defined as custom blocks (including custom reporters) - such as if/elif/else with expandable branches, or expandable join/+/* blocks etc (as found in blockly)

Essentially I don't think that the addon in its current state is very useful (especially compared to #8674, or a good fit for the extension, and adding loads of built-in transpileable blocks is not a good use of resources.

evasammenziemalin-boop:
please update this. i would love to be hable to use this

pufferfish101007:
> please update this. i would love to be hable to use this

what in particular would you like to see updated? as it stands, this should work fine with the scratch editor as it is at the moment.

evasammenziemalin-boop:
> > please update this. i would love to be hable to use this
> 
> what in particular would you like to see updated? as it stands, this should work fine with the scratch editor as it is at the moment.

it doesn't work for me or i don't know how to use it

pufferfish101007:
If you download the zip from https://github.com/pufferfish101007/ScratchAddons/archive/refs/heads/transpilation-poc.zip and extract it, then load it as an unpacked extension from chrome://extensions with developer mode turned on. Then make sure the current version of scratch addons from the web store is disabled, and it should just work (although there are bugs as mentioned in the pull request description).

I've merged upstream into this branch so other addons will be up-to-date; I don't currently have any plans to update this further to fix bugs or maintain compatibility with upcoming scratch changes anytime soon. #8674 might be a bit more reliable (no guarantees that it will be though) and does the same sort of thing, if you're really interested.

evasammenziemalin-boop:
> #8674 might be a bit more reliable (no guarantees that it will be though) and does the same sort of thing.

i already used this

PR #7479, opened 2024-05-29
Remove `handle-licenses.js`

### LINKED ISSUE #7473: Remove `handle-licenses.js`
### Why this would be helpful

The `handle-licenses.js` module preloads licenses into memory so they can be displayed on the `/webpages/settings/licenses.html` page. However, this page is rarely opened and is not needed immediately.

Plus, since the upgrade to Manifest V3, this data gets purged from memory every time the service worker is terminated from inactivity, defeating the purpose of loading it into memory in the first place.

### My suggestion

Instead of the background page providing the license information, it should be loaded by the licenses page as soon as it is opened.

### Possible alternatives

Build the page. This could be done during development or just when packing for production.

### Additional context

_No response_

DNin01:
Resolves #7473

### Changes

The `handle-licenses.js` background module has been removed and license text fetching is handled by the licenses page when it is opened.

### Reason for changes

Previously, license text was loaded into the background context's memory persistently, even though the licenses page was rarely opened. This doesn't make much sense anymore because of the non-persistent nature of MV3 background contexts.

### Tests

Tested with Edge 125 on Windows 11.

The page still loads very fast.


DNin01:
I won't work on #7494 until this pull request is merged.

DNin01:
> I won't work on #7494 until this pull request is merged.

Never mind, I did anyway.

webpages/settings/licenses.html
```diff
@@ -23,6 +23,7 @@ <h1 v-cloak>{{ msg("licenses") }}</h1>
     </div>
     <div class="main">
       <div class="licenses-block">
+        <h2 id="loading">Loading...</h2>
```
Samq64, line 26: I don't think this is necessary. On a device that takes at least 6 seconds to load the popup the licensees page still loads in less than one. It doesn't hurt to keep though.

WorldLanguages: Do we want to localize this?

mxmou: On my laptop, the loading text only appears for a few frames, which is ugly.

DNin01: I just changed it:
- "Loading..." is localized
- "Loading..." is not displayed for 250ms

webpages/settings/licenses.js
```diff
@@ -14,37 +15,43 @@ const vue = new Vue({
...
+await Promise.all(
+  licenses.map((name) =>
+    fetch(`/libraries/licenses/${name}.txt`)
+      .then((res) => res.text())
+      .then((text) => (licenseNameToText[name] = text))
+  )
+);
+
+// Display the license of each library
+for (const library of libraries) {
```
Samq64, line 45: Do the libraries need to be looped though twice?

DNin01: Does it matter? The main reason there are two for-of loops is because of the sequence of operations that needs to happen, and there's a step between them - getting the text of each license - that is programmed to handle data in bulk.
https://github.com/ScratchAddons/ScratchAddons/blob/1c909ba58f8714cd4ed67bf91532db9cac8b680a/webpages/settings/licenses.js#L34-L42
This approach prevents redundant fetches (each file is only fetched once) and allows the fetches to complete asynchronously. For example, if two libraries both have the same license, that license text will be fetched once and then displayed in two different places on the page.

If you're just trying to understand the code, the procedure is this:
1. Get the URL parameter corresponding to which library licenses to display on the page and store it in the `libraries` array. (Note: this could be arbitrary, so we need to validate it)
2. Fetch and load the file that defines which license each library uses and store it in the `libraryLicensesList` variable.
3. **For each library** name, store the name of the license it uses in the `licenses` array, if that license is included in `libraryLicensesList`.
4. Remove duplicate values from `licenses`.
5. Fetch and load each license text file, of the licenses that we need to display.
6. **For each library** license to be displayed, show its license text on the page.

Samq64 (approved):

background/background.js
```diff
@@ -29,4 +29,3 @@ import "./get-popups.js";
 
 // Respond to requests by the settings page and others
```
mxmou, line 30:
```suggestion
// Respond to requests by the settings page
```

mxmou (approved):

Issue #7481, opened 2024-05-31
Settings page works - but addons don't - on v1.38.1


The-Abigail608:
### Description

After receiving the MV3 update, nothing works, even if I wait, it does not load unless i reimport my user settings, which reloads the extension, i don't completely remember the update notification I got, but I remember it saying something along the lines of "You've been selected as a tester of the MV3 update, and none of the addons had loaded. I've tried waiting and it doesn't work, it sometimes keeps working between some pages with certian addons off (like scratchstats, mystuff share, and mystuff search) but if I click off for more than 30 seconds it entirely unloads

### Steps to reproduce

1. Have update happen out of blue
2. have these addons enabled:
[scratch-addons-settings (1).json](https://github.com/user-attachments/files/15515498/scratch-addons-settings.1.json)
3. Scratch Addons fails to work.

### Environment

Browser: Chrome Version 124.0.6367.225 (Official Build) (64-bit)
Scratch Addons version: v1.38.1
Operating system: Chrome OS

Samq64:
Everything worked fine on v.38.0, right? This might be caused by #7417.

Also I'm pretty sure the update has been fully rolled out now and the notification was never updated.

WorldLanguages:
I believe around 20% of Chrome users went directly from v1.37.1 to v1.38.1

WorldLanguages:
> but if I click off for more than 30 seconds it entirely unloads

@The-Abigail608 What do you mean by this?

The-Abigail608:
if scratch addons is reloaded and im on a scratch tab where it's working, i probably meant to say scratch addons stops working if i wait some time and go to a new scratch tab, it might be caused by one of the addons i have enabled throwing an error, but i am not completely sure.

WorldLanguages:
@The-Abigail608 So the addons/features do not always work, but the settings page always works fine, right?

TheColaber:
Can you go on a scratch tab, press ctrl + shift + I, and screenshot and share with us all the errors?

The-Abigail608:
I am unable, sadly. Inspect is blocked.

WorldLanguages:
@The-Abigail608 So the addons/features do not always work, but the settings page always works fine, right?

The-Abigail608:
Funny you should mention that, dark mode didnt activate when i went from the main page (with addons working) to settings. it will also happen if i go from my profile (addons loaded) to my stuff (stops working) or main page to my profile.

[...]

PR #7483, opened 2024-05-31
Temp solution for `necropost-highlighter` crash

DNin01:
Partially addresses #7481

The `necropost-highlighter` addon uses regex match patterns which can't be serialized to JSON. When loaded from cache, the regex is lost and causes an "invalid URL" error. This temporary solution catches that error.

### Tests

Tested.

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -446,6 +446,12 @@ function userscriptMatches(data, scriptOrStyle, addonId) {
 }
 
 function urlMatchesLegacyPattern(pattern, urlUrl) {
+  try {
+    new URL(pattern);
+  } catch (err) {
+    console.error(err);
+    return false;
+  }
   const patternUrl = new URL(pattern);
   // We assume both URLs start with https://scratch.mit.edu
 


Samq64:
Why not just update the `necropost-highlighter` matches to use a glob instead of regex?

DNin01:
> Why not just update the `necropost-highlighter` matches to not use regex instead?
> 
> The mobile forums aren't a thing anymore so `https://scratch.mit.edu/discuss/*/` should be fine.

To be extra specific, the matches should probably be:
```json
[
  "https://scratch.mit.edu/discuss/search/",
  "https://scratch.mit.edu/discuss/1/",
  "https://scratch.mit.edu/discuss/3/",
  "https://scratch.mit.edu/discuss/4/",
  "https://scratch.mit.edu/discuss/6/",
  "https://scratch.mit.edu/discuss/7/",
  "https://scratch.mit.edu/discuss/9/",
  "https://scratch.mit.edu/discuss/11/",
  "https://scratch.mit.edu/discuss/49/"
]
```
If we just simply replace it with `https://scratch.mit.edu/discuss/*/`, then, if I'm right, it would run on forum posts and other pages.

Samq64:
> If we just simply replace it with https://scratch.mit.edu/discuss/*/, then, if I'm right, it would run on forum posts and other pages.

> No, because there's a slash at the end it won't run on pages deeper than that.

DNin01:
> > If we just simply replace it with https://scratch.mit.edu/discuss/*/, then, if I'm right, it would run on forum posts and other pages.
> 
> No, because there's a slash at the end it won't run on pages deeper than that.

But not all forums are matched, only the specified forum IDs, because there are some forums where necroposts are not as much of a problem, so the userscript doesn't run on them. At least that's what I'm assuming?

Running on all topic listing pages would be fine as long as it doesn't error on other non-topic listing pages that `/forums/*/` match. Or are there any?

DNin01:
Let's try a more permanent solution: #7486

PR #7486, opened 2024-06-01
Bug-fix: clone manifests before caching them

### LINKED ISSUE #7481: Settings page works - but addons don't - on v1.38.1

### LINKED ISSUE #7484: Service worker restart breaks regex matches
### Description

#7417 broke regex userscipt and userstyle matching after the service worker is restarted. This doesn't effect glob matching, only regex.

### Steps to reproduce

1. Add a regex` addon.json` match (e.g. `^https://scratch.mit.edu/discuss/(?:m/|)(?:search|1|3|4|6|7|9|11|49)/`")
2. Enable the addon
3. Terminate the service worker from `chrome://serviceworker-internals/`
4. Open a Scratch tab

This is the root cause of #7481.

### Environment

Browser: Chromium 125
Scratch Addons version: v1.38.1
Operating system: Linux

DNin01:
Resolves #7481
Resolves #7484

### Changes

Performs a deep clone on each addon-manifest cache entry.

### Reason for changes

Since manifests are cached before they are processed and processing happens whether or not the manifests were loaded from disk or from the cache, it only makes sense that the cache is not modified during processing. As an example, this fixes an issue where values that couldn't be serialized to JSON were ending up in the cache and causing errors because the processing code didn't expect such data.

### Tests

Tested.

### DIFF
--- a/background/load-addon-manifests.js
+++ b/background/load-addon-manifests.js
@@ -66,7 +66,7 @@ const localizeSettings = (addonId, setting, tableId) => {
       } else {
         const file = await (await fetch(`/addons/${addonId}/addon.json`)).json();
         manifest = file;
-        newCache[addonId] = file;
+        newCache[addonId] = JSON.parse(JSON.stringify(file));
       }
     } catch (ex) {
       console.error(`Failed to load addon manifest for ${addonId}, crashing:`, ex);


WorldLanguages:
Good idea, I didn't have time today but I was thinking we had to include the regexp objects somehow. But you're right that avoiding mutations to the cache fixes the issue easily.

Samq64 (approved):

background/load-addon-manifests.js
```diff
@@ -66,7 +66,7 @@ const localizeSettings = (addonId, setting, tableId) => {
       } else {
         const file = await (await fetch(`/addons/${addonId}/addon.json`)).json();
         manifest = file;
-        newCache[addonId] = file;
+        newCache[addonId] = JSON.parse(JSON.stringify(file));
```
WorldLanguages, line 69: Note that `structuredClone()` worked equivalently but isn't supported by Chrome 96.

DNin01: 😮 That's a feature I didn't know!

`storage.session` is only supported in Chrome 102+ and Firefox 115+, so actually, we wouldn't even have to use `structuredClone()` in older versions, so technically we could replace `JSON.parse(JSON.stringify(obj))` with `structuredClone()`. We could also fall back to `JSON.parse(JSON.stringify(obj))` if `structuredClone()` is not available, or we could use [a polyfill](https://github.com/ungap/structured-clone).

It seems to be a little faster to use `structuredClone()`.
console

hueychen27: Average of about 0.58772786458 ms difference and a percent increase of about 62.658337671%. Expression used: |((0.994873046875+0.989990234375+0.8291015625))-((1.7451171875+1.5361328125+1.2958984375))|/((0.994873046875+0.989990234375+0.8291015625))

DNin01: I tried with an even larger object and the time save was really substantial - `JSON.parse(JSON.stringify())` took about a quarter of a second while `structuredClone()` only took, like, 4 ms.

It's a little more complicated than that, though - cloning a smaller amount of data, like an object that takes only a millisecond or two to clone, `JSON.parse(JSON.stringify())` pulls ahead, surprisingly, which can add up when you're cloning hundreds of small objects.

It's possible this behavior will also vary from browser to browser and platform to platform.

hueychen27: 250 ms to 4 ms??! That must mean the JSON method grows almost exponentially in time as the object gets larger.

WorldLanguages: > `storage.session` is only supported in Chrome 102+ and Firefox 115+, so actually, we wouldn't even have to use `structuredClone()` in older versions

Huh, good point, I missed that.
You may submit a PR to change it, if you want. But make sure to add a code comment or a useless if statement, so that it's clear that structuredClone should not be used in other contexts.

PR #7491, opened 2024-06-01
`hide-flyout`: Fix input and dropdown closing palette

### LINKED ISSUE #5263: hide-flyout: Opening dropdown in flyout hides flyout
### Description

The flyout will hide when opening a dropdown from a block in the flyout and hovering over that dropdown.
image
image

### Steps to reproduce

1. Enable hide-flyout
2. Open dropdown of a block
3. Hover over that dropdown

_No response_

### Environment

Browser: Chrome 106
Operating system: Windows

### LINKED ISSUE #7024: `hide-flyout` palette & category hover automatically closes blocks pane when editing an input
### Description

<sub>(Note: for some reason I was unsure if this was a bug so I used the suggestion template originally, reformatted this manualy)</sub>

`hide-flyout` is really useful, especially category hover & pallete area hover- However it will still automatically close if you are editing an input before dragging it on the code pane.

This can be annoying if you edit a variable setter and then drag (which I do) instead of dragging and then editing

### Steps to reproduce

1. Enable `hide-flyout` and either palette or category hover
2. Go in the editor
3. Edit an input

### Possible alternatives

1. Make it optional (though that'd be odd)
2. Don't implement
3. Auto-lock when editing an input

(Keeping this in for now)
### Additional context

hide-flyout-closes-on-input
(occurs after merely attempting to edit the input)

hueychen27:
Resolves #7024
Resolves #5263

### Changes

If statement expanded to check for the input or dropdown element being focused within, cancelling the palette from closing.

### Reason for changes

<s>These changes should be built upon further to identify where the input and dropdown elements originate from (palette or code area,) but the bugs introduced are minor compared to the bugs being fixed by this pull request.</s> The changes fix all bugs successfully, and they provide a greater user experience with this addon.

### Tests

`web-ext run` has been used to test in Firefox v126.0.1 (64-bit).
Results:  
<s>It works a little too much as clicking on inputs outside of the palette will keep the palette on until the input is deactivated. This is only one bug found. All else works normally.</s> Everything works normally based on many tests performed.

[showcase_for_bug_fix_of_palette_1.webm](https://github.com/ScratchAddons/ScratchAddons/assets/78285518/6f149134-0aae-4e8f-bb0e-952834438ef9)


Samq64:
> web-ext run has been used to test in Firefox v126.0.1 (64-bit).

Loading from about:debugging has always worked fine for me.

> It works a little too much as clicking the inputs in blocks outside of the palette cause the palette to become permanently in place when shown until the input is let go of. All else works normally.

Since the palette is open by default auto-hiding is broken when first opening the editor until a block dropdown is closed .

WorldLanguages:
> It works a little too much as clicking the inputs in blocks outside of the palette cause the palette to become permanently in place when shown until the input is let go of.

A video would be appreciated, unless it is fixed. We don't want the bug-fix to potentially result in worse behavior.

hueychen27:
A video will be added showing a lot of testing.

hueychen27:
> > web-ext run has been used to test in Firefox v126.0.1 (64-bit).
> 
> Loading from about:debugging has always worked fine for me.
> 
> > It works a little too much as clicking the inputs in blocks outside of the palette cause the palette to become permanently in place when shown until the input is let go of. All else works normally.
> 
> Since the palette is open by default auto-hiding is broken when first opening the editor until a block dropdown is closed .

A second commit was added fixing that issue.

TheColaber:
this could probably be done using blockly?

TheColaber:
look into Blockly.DropDownDiv.isVisible and Blockly.WidgetDiv.isVisible

hueychen27:
> look into Blockly.DropDownDiv.isVisible and Blockly.WidgetDiv.isVisible

Alright.

mxmou:
You can check if the `WidgetDiv` was opened from inside the flyout like this:
```js
const owner = Blockly.WidgetDiv.owner_;
const openedFromFlyout = owner instanceof Blockly.Field && owner.sourceBlock_?.isInFlyout;
```
and similarly for the `DropDownDiv`.

There's still a remaining bug even after adding this check: clicking outside the flyout to close the dropdown doesn't close the flyout. One way to fix this is to
* set a global variable to `true` before returning from `onmouseleave` if the palette wasn't closed because an input/dropdown is visible,
* set it to `false` in `onmouseenter`, and
* override `Blockly.WidgetDiv.hide` and `Blockly.DropDownDiv.hide` to check if the variable is set and close the palette.

hueychen27:
@mxmou What should the global variable be named?

hueychen27:
@mxmou Never mind, I have decided (request change if desired.) Everything is fixed, and I found another solution similar to yours to fix the remaining bug of clicking outside the flyout when the dropdown or input is clicked off of.

addons/hide-flyout/userscript.js
```diff
@@ -9,6 +9,7 @@ export default async function ({ addon, console, msg }) {
   let flyoutLock = false;
   let closeOnMouseUp = false;
   let scrollAnimation = true;
+  let notClosedByVisibleDropdownOrInput = true;
```
Samq64, line 12: Could this be reversed? Not-variables get confusing quickly especially if they aren't constants. Also visible doesn't need to be in the variable's name.
```suggestion
  let closedByDropdownOrInput = false;
```
Remember to also reverse the values elsewhere too.

hueychen27: I was thinking to remove the "not" from the variable name, but decided otherwise. My next commit removes the global variable. After looking closer, the variable need not to be used at all, but instead, it can be substituted for another value.

addons/hide-flyout/userscript.js
```diff
@@ -80,7 +80,19 @@ export default async function ({ addon, console, msg }) {
...
   function onmouseleave(e, speed = getSpeedValue()) {
-    if (flyoutLock) return;
+    // locked palette, inputting text, and hovering over dropdown menu do not close palette
+    const widget = Blockly.WidgetDiv.owner_;
+    const dropdown = Blockly.DropDownDiv.owner_;
+    const widgetOpenedFromFlyout = widget instanceof Blockly.Field && widget.sourceBlock_?.isInFlyout;
+    const dropdownOpenedFromFlyout = dropdown instanceof Blockly.Field && dropdown.sourceBlock_?.isInFlyout;
+    const widgetOrDropdownOpenedFromFlyout = widgetOpenedFromFlyout || dropdownOpenedFromFlyout;
+    // Don't forgot to close when the mouse leaves the flyout even when clicking off of a dropdown or input
+    if (widgetOrDropdownOpenedFromFlyout) closeOnMouseUp = true; // chef's kiss to the following conditional madness above and below
```
Samq64, line 90:
```suggestion
    if (widgetOrDropdownOpenedFromFlyout) closeOnMouseUp = true;
```

Samq64 (approved):

addons/hide-flyout/userscript.js
hueychen27: If you prefer, then I will accept.

mxmou (approved):

hueychen27:
New bug (not introduced though): When clicking on a variable or list in the palette, hovering over the reporter value message will also close the palette.

hueychen27:
I think you guys should still merge this because this fixes the more major bugs unless you want to fix all bugs.

hueychen27:
@mxmou Good choice

WorldLanguages:
Nice work everyone :)

hueychen27:
I can officially say that I have contributed to Scratch Addons with code 😀

PR #7498, opened 2024-06-04
Include copyright notices in `licenses.html`

### LINKED ISSUE #7494: Legal: Should the licenses page include the author and copyright year?
When you use a work under a copyright license that requires including the license and copyright notice generally means that you should also attribute the author or keep the author and copyright year. But I don't see the authors anywhere.

Proper credit is important. If we show the licenses, we should also show the copyright notices, and it should be part of Scratch Addons, not on a remote source such as our website.

Let me know if I'm missing something, because I looked everywhere. I know we link to the package sources from our website, but nothing's better than clear, obvious attribution.

DNin01:
Resolves #7494

### Changes

- Adds a copyright notice, including the year and full name fields, to each license that requires it, and updated `license-info.json` with that information.
- Line breaks have been inserted in licenses without line breaks for hard word wrapping for consistency.

### Reason for changes

The licenses of the libraries we use require giving attribution.

### Notes

- License text is sourced from the source code of the libraries _for the versions that we use_. License information may need to be revised if we update a dependency.
- `scratchblocks` uses an MIT license, but the copyright notice is written a little differently, so I made a copy of the MIT license file just for it.
- It looks to me like the Scratch Addons website has its own license info file, so we shouldn't need to make any changes to keep displaying library license information on our website.
- ~~Merging this may cause conflicts in #7479 (and also the other way around), but I don't really mind.~~ That's been sorted now - don't worry.

### Tests

Tested.


Hans5958:
Keep in mind that there is [the "Dependencies" section on the Credits page](https://scratchaddons.com/credits/#dependencies), which links what libraries that are used and its license identifier, which should be fine, especially for reference only.

libraries/license-info.json
```diff
@@ -1,16 +1,64 @@
...
+  },
+  "chartjs": {
+    "license": "MIT",
+    "year": "2014-2022",
+    "fullname": "Chart.js Contributors"
+  },
+  "color-picker-web-component": {
+    "license": "MIT",
+    "year": "2020",
+    "fullname": "Wouter Vroege <wouter@woutervroege.nl>"
```
GrahamSH-LLK, line 40: this library has been quite modified by me, not sure how licensing works. source is no longer available tho

WorldLanguages:
Okay, I merged the other one

DNin01:
Alright, the conflicts have been resolved, and all the correct licenses seem to be displayed.

DNin01:
Why was the https://github.com/ScratchAddons/ScratchAddons/labels/status%3A%20awaiting%20answer%2Ffollowup label added?

Hans5958 (approved):
Didn't check that much (read the code but not testing it), but should be good.

WorldLanguages:
@DNin01 It's for the reminders, I'm not waiting for any followups.

Samq64 (approved):

PR #7504, opened 2024-06-05
Add some styles to the error  page

Samq64:
### Changes

Adds some basic styles on the error page.
image

### Reason for changes

No styling looks ugly especially in the popup.

### Tests

Tested on Chromium and Firefox.

### DIFF
--- a/webpages/error/index.html
+++ b/webpages/error/index.html
@@ -10,7 +10,7 @@
   </head>
 
   <body>
-    <h2 id="problem-title">Scratch Addons encountered a problem:</h2>
+    <h2>Scratch Addons encountered a problem:</h2>
 
     <h1 id="problem-type">(unknown error)</h1>
 
--- a/webpages/error/index.html
+++ b/webpages/error/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <title>Error - Scratch Addons</title>
     <!-- <script src="../set-lang.js" type="module"></script> -->
     <script src="./index.js" type="module"></script>
     <script src="./export.js" type="module"></script>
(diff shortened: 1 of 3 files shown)


Samq64:
There's also the option of using the Vue template like the license ans permissions page but it's probably better not to rely on that on the error page.

mxmou:
This always makes the page background dark, even if the extension is in light mode. I think always using a light color scheme is a better choice.

Could the font size be slightly smaller - maybe `1rem`?

Samq64:
> I think always using a light color scheme is a better choice.

To keep using CSS variables though I'll need to import the dark stylesheet anyway since the light version isn't complete.

mxmou:
> > I think always using a light color scheme is a better choice.
> 
> To keep using CSS variables for all colours though I'll need to import the dark stylesheet anyway since the light version isn't complete.

All variables that are currently used on the page are defined in `colors-light.css`.

(I don't like the current theme system with a separate stylesheet for light mode - it only made sense before we stared using CSS variables for colors. There could be a single file with all color definitions and the theme could be controlled by changing the root element's class name: `:root.dark`, `:root.light`.)

Samq64:
> All variables that are currently used on the page are defined in colors-light.css.

`--blue-text` is just `--blue` in light mode but `--blue` only exists in the dark colour palette. It's fixed now.

Samq64:
Also would it make sense to use `prefers-color-scheme`?

WorldLanguages:
> Also would it make sense to use `prefers-color-scheme` on the error page?

No. I think we should only support light mode on the error page.

mxmou (approved):

WorldLanguages:
Personally I prefer the default font over Sora for this page. Maybe it's just me.

WorldLanguages (approved):
I prefer if you merge it without Sora, but it's up to you

PR #7505, opened 2024-06-05
Call background alarms less often (`msg-count-badge` affected)

WorldLanguages:
- Removed an unnecessary (since MV3) alarm that was previously named `cleanCsInfoCache`. See comment https://github.com/ScratchAddons/ScratchAddons/issues/7473#issuecomment-2138037284
- If the user is considered inactive, the background context of the extension should only wake up every 150 seconds (2.5 minutes). This is achieved by reducing the frequency of "the alarm that runs most often" (`updateBadge`)

.

After this PR, these are the events that may trigger the background context of the extension to wake up:
- The `updateBadge` alarm every 1 minute, **but every 2.5 minutes if user is considered inactive.**
- The `fetchMessages` alarm every 5 minutes.
- Opening the popup or settings page, changing settings, using the muting feature, etc.
- Navigating to a new Scratch page.
- Any scratch.mit.edu cookies changing for any reason (such as other browser extensions)

.

The user is considered inactive if none of the following have happened for the last 10 minutes:
- Opening the popup or settings page
- Navigating to a new Scratch page

Additionally, even if other conditions are met, if the selected browser tab is a scratch.mit.edu page after the 10 minutes have passed, the period is renewed for another 10 minutes.

### DIFF
--- a/background/imports/inactivity.js
+++ b/background/imports/inactivity.js
@@ -1,0 +1,37 @@
+import { handleBadgeAlarm } from "../message-cache.js";
+
+const INACTIVITY_DELAY_MINS = 10;
+const INACTIVITY_ALARM = "inactivityAlarm";
+
+// This function should be called each time the user interacts with scratch.mit.edu or with the extension.
+// We consider the user inactive if it's been some minutes since this function was last called.
+export function setUserAsActive() {
+  chrome.alarms.clear(INACTIVITY_ALARM);
+  chrome.storage.session?.set({ inactivity: false });
+
+  chrome.alarms.create(INACTIVITY_ALARM, {
+    delayInMinutes: INACTIVITY_DELAY_MINS,
+  });
+
+  handleBadgeAlarm();
+}
+
+chrome.storage.session?.get("inactivity", (o) => {
+  if (o.inactivity === undefined) {
+    setUserAsActive(); // Consider user active on startup
+  }
+});
+
+chrome.alarms.onAlarm.addListener(async (al) => {
+  if (al.name === INACTIVITY_ALARM) {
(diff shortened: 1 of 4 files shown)


background/imports/inactivity.js
```diff
@@ -0,0 +1,37 @@
+import { handleBadgeAlarm } from "../message-cache.js";
+
+const INACTIVITY_DELAY_MINS = 10;
+const INACTIVITY_ALARM = "inactivityAlarm";
+
+// This function should be called each time the user interacts with scratch.mit.edu or with the extension.
+// We consider the user inactive if it's been some minutes since this function was last called.
+export function setUserAsActive() {
+  chrome.alarms.clear(INACTIVITY_ALARM);
+  chrome.storage.session?.set({ inactivity: false });
```
Samq64, line 10: Is there a reason the key isn't simply called `active`?

WorldLanguages: Well, at the beginning of the session we want to consider the user active, but nothing will be in storage, so "inactivity" will be undefined which is falsey.

WorldLanguages: But if you don't want to switch all the booleans around, it could be renamed to `inactive`, sure.

WorldLanguages:
@DNin01 If possible, could you breafly read/test this? You don't need to confirm that it works as expected, just want to make sure there aren't any big noticeable mistakes.

background/imports/inactivity.js
DNin01, line 1: To support browsers without `chrome.storage.session`, we could use `chrome.alarms.get` instead. We would just check if the inactivity alarm is running, which would indicate that the user is active.

WorldLanguages: Not worth it really

PR #7506, opened 2024-06-05
`better-img-uploads`: discard EXIF metadata before upload

### LINKED ISSUE #7380: Image file metadata and the `better-img-uploads` addon
It has been reported through the feedback page that:

> The HD image uploader does not remove information from the metadata of images. This can lead to geolocation data being preserved, which is a privacy risk. 

If uploading the image normally through vanilla Scratch upload does cause the metadata to be removed for some reason, then we should consider changing our addon so that it's also removed.

WorldLanguages:
Resolves #7380

### Changes

Draws the original image into a canvas before uploading it to the project

### Tests

It should work

### DIFF
--- a/addons/better-img-uploads/userscript.js
+++ b/addons/better-img-uploads/userscript.js
@@ -116,27 +116,32 @@ export default async function ({ addon, console, msg }) {
         continue;
       }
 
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Tested in Chromium and Firefox

PR #7053, opened 2024-01-06
New addon: Extend wrapped C blocks

CST1229:
Resolves [this suggestion from the TurboWarp server
](https://discord.com/channels/837024174865776680/837024174865776683/1193279234618171444)

### Changes

Adds an addon that makes C blocks extend when they're wrapped around scripts ("chomping"):
image

### Reason for changes

Someone asked for it, and it's a nice aesthetic change. This feature was also in Scratch 2.0 and 1.4:
image image

### Tests

Tested on Firefox, with dynamic enable/disable and exiting/entering the editor. Seems to work fine. I also enabled a bunch of code area-affecting addons too and it still seems to work fine.

(I'm not sure if this will get accepted, as it's a very minor thing. Also, this is the third "like in Scratch 2.0" addon lol)

PR #7509, opened 2024-06-07
`editor-theme3`: add customizable insertion markers

shock59:
Resolves [this Discord community suggestion](https://discord.com/channels/806602307750985799/1239960555184193638)

### Changes

Modifies the "Customizable block colors" addon to have options for using the block's colour as the colour for the insertion marker and to customise the opacity of the insertion marker.

image

### Reason for changes

It looks nice and gives the user more control over how the editor looks.

### Tests

Tested in Vivaldi (Chromium) and it works fine.


addons/editor-theme3/insertion_marker.css
```diff
@@ -0,0 +1,3 @@
+.blocklyInsertionMarker .blocklyPath.blocklyBlockBackground {
```
Samq64, line 1: This selector applies to all children when it shouldn't.
```suggestion
.blocklyInsertionMarker > .blocklyPath {
```

Samq64 (commented):
I think this would fit better in #7053.

Also this PR isn't currently compatible with dark mode.

addons/editor-theme3/addon.json
```diff
@@ -364,6 +370,20 @@
...
       "default": false
+    },
+    {
+      "name": "Colored insertion markers",
+      "id": "insertion-markers",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "name": "Insertion marker opacity (0-100%)",
```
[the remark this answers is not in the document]
mxmou: I haven't tested the addon, but if I'm reading the code right, this setting works even if the other one is disabled.

addons/editor-theme3/addon.json
```diff
@@ -12,7 +12,12 @@
     {
       "name": "DNin01",
       "link": "https://github.com/DNin01"
-    }
+    },
+    {
+      "name": "Shock59",
+      "link": "https://github.com/shock59"
+    },
+    { "name": "Ashimee", "link": "https://github.com/Ashimee" }
```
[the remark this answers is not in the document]
shock59: They wrote the original method for the coloured insertion marker on Discord, which I adapted into the addon userscript.

shock59:
Thanks for pointing out the incompatibility with dark mode, I didn't realise that. I will try and resolve it and update the PR.

I feel like it would be very confusing for the user to combine it with the "Extend wrapped C blocks" addon, they don't appear to be related in any way?

shock59:
Dark mode overrides the insertion marker colour at line 334 in experimental_editor.css. I'm not how it would be possible to check if the coloured insertion markers were enabled and stop it from doing that, if anyone has any ideas that would be appreciated.

addons/editor-theme3/theme3.js
```diff
@@ -354,6 +354,11 @@ export default async function ({ addon, console, msg }) {
       colourSecondary = secondaryColor(extensionsCategory);
       colourTertiary = tertiaryColor(extensionsCategory);
     }
+
+    if (this.isInsertionMarker() && addon.settings.get("insertion-markers")) {
+      return true;
```
mxmou, line 359:
```suggestion
      this.svgPath_.style.fill = this.getColour(); // override editor-dark-mode style
      return;
```

mxmou (changes requested):
The suggestion below shows how the dark mode incompatibility can be fixed. Other bugs I found:
* Insertion markers for debugger blocks have the custom block color.
* When using the "Black" preset, all insertion markers are black.

shock59:
The bug with the debugger blocks is concerning, i'm not really sure how it could be addressed. The only method I can think of would be manually defining the styles like I did [here](https://github.com/shock59/ScratchAddons/blob/0bd05f72d506681e8b357c6dde2cfe4be2a3953c/addons/editor-theme3/insertion_marker.css) before I worked out you could achieve it with JavaScript

mxmou:
> The bug with the debugger blocks is concerning, i'm not really sure how it could be addressed. The only method I can think of would be manually defining the styles like I did [here](https://github.com/shock59/ScratchAddons/blob/0bd05f72d506681e8b357c6dde2cfe4be2a3953c/addons/editor-theme3/insertion_marker.css) before I worked out you could achieve it with JavaScript

Try overriding `Blockly.InsertionMarkerManager.prototype.createMarkerBlock_` instead of `Blockly.Block.prototype.setColour`. It takes the original block as a parameter, so you will be able to simply copy its color to the insertion marker.

addons/editor-theme3/theme3.js
```diff
@@ -370,6 +371,13 @@ export default async function ({ addon, console, msg }) {
     }
   };
 
+  const oldInsertionMarkerCreateMarkerBlock = Blockly.InsertionMarkerManager.prototype.createMarkerBlock_;
+  Blockly.InsertionMarkerManager.prototype.createMarkerBlock_ = function (originalBlock) {
```
Samq64, line 375: This doesn't check if the setting is actually enabled.

addons/editor-theme3/insertion_marker.css
```diff
@@ -0,0 +1,3 @@
+.blocklyInsertionMarker > .blocklyPath {
+  fill-opacity: calc(var(--editorTheme3-insertionMarkerOpacity) * 0.01) !important;
```
Samq64, line 2:
```suggestion
  fill-opacity: calc(var(--editorTheme3-insertionMarkerOpacity) * 0.01);
```

addons/editor-theme3/addon.json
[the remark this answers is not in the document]
Samq64: Could notes be added to these credits?

addons/editor-theme3/theme3.js
Samq64: Also ideally this should support dynamic settings and disable.

shock59: Whoops, I accidentally removed the check when I moved it to a new function. I'll fix it tomorrow

addons/editor-theme3/addon.json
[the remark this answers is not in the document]
shock59: Maybe the original userscript note would be applicable? It's been modified so heavily now though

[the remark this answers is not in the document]
mxmou: The notes could be "addon" and "original code".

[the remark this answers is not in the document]
shock59: Whoops, I just commited an update with your original notes but I can remove them. I agree that the notes are a bit long and it's probably fine to leave it without them

shock59:
> * When using the "Black" preset, all insertion markers are black.

I still haven't found a way to fix this, if anyone has any ideas that would be much appreciated. It does work for the debugger blocks now though :)

mxmou:
> When using the "Black" preset, all insertion markers are black.

This still needs to be fixed. I think all you need to do is use the tertiary color instead of the primary color if `isColoredTextMode()` returns true.

addons/editor-theme3/theme3.js
```diff
@@ -354,6 +354,7 @@ export default async function ({ addon, console, msg }) {
       colourSecondary = secondaryColor(extensionsCategory);
       colourTertiary = tertiaryColor(extensionsCategory);
     }
+
```
mxmou, line 357:
```suggestion
```

mxmou (dismissed):

addons/editor-theme3/theme3.js
```diff
@@ -370,6 +370,16 @@ export default async function ({ addon, console, msg }) {
     }
   };
 
+  const oldInsertionMarkerCreateMarkerBlock = Blockly.InsertionMarkerManager.prototype.createMarkerBlock_;
+  Blockly.InsertionMarkerManager.prototype.createMarkerBlock_ = function (originalBlock) {
+    const markerBlock = oldInsertionMarkerCreateMarkerBlock.call(this, originalBlock);
+    if (addon.settings.get("insertion-markers"))
```
Samq64, line 376:
```suggestion
    if (!addon.self.disabled && addon.settings.get("insertion-markers"))
```

Samq64 (dismissed):

Jazza-231:
Love this addon idea, got a few ideas I've discussed in the suggestion thread and with @shock59, the discussion surrounded adding an option to change the insertion marker style to be able to be a fill or a stroke colour.
Fill:
image
Stroke:
image

As well as thinking about adding colour to this insertion marker:
image

I have implemented the settings and code for the first one which I will try to push to the branch, but help would be appreciated with the latter item :D

shock59:
The new additions should work fine but to make sure I'll test them properly tomorrow. Also we couldn't decide on what the best way to handle the opacity of the optional indicator outline would be - it could feel cluttered having another extra option to customise it but it would be strange to have it as a static value when the fill opacity is now customisable. Any input on this would be appreciated

shock59:
There are a few bugs and issues with @Jazza-231's additions which I will work to solve:
- Enabling the insertion marker outline when the toggle for coloured insertion markers is turned off does nothing but you can still change it
- "Insertion Marker Style" is an unclear name, especially because it is tied to the coloured insertion marker setting (it probably shouldn't be)
- There is no way to change the opacity of the insertion marker outline
- It would look good if selecting the "Black" preset enabled the outlined insertion markers

Samq64:
@Jazza-231 How did you get write access? Also next time if a PR already has 2 approvals could you open a new one instead?

Jazza-231:
> How did you get write access? 
Sam gave it to me...?
> Also next time if a PR already has 2 approvals could you open a new one instead?
Probably a good idea but oh well

Jazza-231:
> Enabling the insertion marker outline when the toggle for coloured insertion markers is turned off does nothing but you can still change it

Fixed
> "Insertion Marker Style" is an unclear name, especially because it is tied to the coloured insertion marker setting (it probably shouldn't be)

Fixed
> There is no way to change the opacity of the insertion marker outline

Added
> It would look good if selecting the "Black" preset enabled the outlined insertion markers

Added

addons/editor-theme3/addon.json
```diff
@@ -364,6 +371,36 @@
...
+    {
+      "name": "Insertion marker style",
+      "id": "markerStyle",
+      "type": "select",
+      "default": "fill",
+      "potentialValues": [
+        { "id": "none", "name": "No color" },
+        { "id": "fill", "name": "Fill color" },
+        { "id": "stroke", "name": "Outline color" },
+        { "id": "both", "name": "Fill and outline color" }
```
mxmou, line 384:
```suggestion
        { "id": "none", "name": "Gray" },
        { "id": "fill", "name": "Colored" },
        { "id": "stroke", "name": "Gray with outline" },
        { "id": "both", "name": "Colored with outline" }
```

addons/editor-theme3/addon.json
```diff
@@ -364,6 +371,36 @@
...
+      "default": 20,
+      "if": { "settings": { "markerStyle": ["none", "fill", "both"] } }
+    },
+    {
+      "name": "Insertion marker outline opacity",
+      "id": "strokeOpacity",
+      "type": "integer",
+      "min": 0,
+      "max": 100,
+      "default": 80,
```
mxmou, line 402: Could this be lower - maybe 50? The Black preset currently makes insertion markers quite similar to regular blocks.

addons/editor-theme3/addon.json
shock59: Would the names "Gray with colored outline" and "Colored with gray outline" be clearer?

addons/editor-theme3/addon.json
shock59: That sounds like a good idea

addons/editor-theme3/addon.json
mxmou: The outline is always colored. Maybe "Gray with colored outline" and "Colored with outline"?

shock59: I forgot about that. To be honest I find this implementation a bit weird, it would be good if the outline could be grey instead of coloured as an option. That would actually make more sense for the black preset too in my opinion.

A good way to do this would be by making the option to customise the fill opacity and the outline opacity always available (setting the opacity to 0% either of them would hide it completely), and then have the "Insertion marker style" setting only affect if they are coloured or not.

mxmou: I think it would be better to separate the fill and outline settings:
image

The default should be the same as Scratch (gray fill, no outline). I like how the black preset currently makes both the fill and the outline colored, so I don't think that should be changed.

shock59: That seems like a nice way of implementing it.

addons/editor-theme3/addon.json
Jazza-231: The black preset sets the opacity to 5 though

addons/editor-theme3/addon.json
Jazza-231: We had reasoning not to separate them though yes? What was that reasoning...

shock59: I don't remember having a reason not to seperate the options, it looks more readable this way to me and gives the user more options at the same time. I think we had some miscommunications while making the feature originally

Jazza-231: Ah ok, feel free to implement these suggestions then, they seem good.

shock59:
> @Jazza-231 How did you get write access? Also next time if a PR already has 2 approvals could you open a new one instead?

I gave her access because we were discussing the ideas in Discord and she already had ideas on how to implement them, so it seemed reasonable since the PR hadn't been merged yet and since the new features might involve rewriting some of the things I already introduced in this PR

addons/editor-theme3/addon.json
shock59: I assume he was talking about the outline, not the fill

WorldLanguages:
> @WorldLanguages This isn't a new addon, it's adding additional options to Customizable block colors

Sorry! I will update the title

Samq64:
Adding 3 non-colour settings to `editor-thtme3` seems like a lot. Maybe this would be better as a new addon?

shock59:
`editor-theme-3` already has so much of the code implemented in it though, and it would let people use the same presets to customise the insertion marker. It does make the settings page for it pretty large though

addons/editor-theme3/addon.json
WorldLanguages: Commit https://github.com/ScratchAddons/ScratchAddons/pull/7509/commits/a1c730064b6b9cf586a795aea96896b895863ac7 changed the Black preset, but not the default value of the setting.

WorldLanguages:
@mxmou I suggest you review again and push changes if you want (such as commiting your own suggestions). If not then you could tell me what's missing / still in discussion so I can review.

If we need to migrate addon settings (presets) I can help with that.

addons/editor-theme3/addon.json
shock59: In the latest commit I changed the default setting to 0 so it would look the same as vanilla Scratch by default since there is now support for a grey outline.

addons/editor-theme3/theme3.js
```diff
@@ -370,6 +370,26 @@ export default async function ({ addon, console, msg }) {
...
+      const styleColour = isColoredTextMode() ? originalBlock.getColourTertiary() : originalBlock.getColour();
+      const markerColourStyle = addon.settings.get("markerStyle");
+
+      if (markerColourStyle === "both") {
+        markerBlock.svgPath_.style.fill = styleColour;
+        markerBlock.svgPath_.style.stroke = styleColour;
+      } else {
+        markerBlock.svgPath_.style[markerColourStyle] = styleColour;
+        if (markerColourStyle !== "stroke") {
+          markerBlock.svgPath_.style.stroke = "rgb(0, 0, 0)";
```
mxmou, line 386:
```suggestion
          markerBlock.svgPath_.style.stroke = "var(--editorDarkMode-workspace-insertionMarker, rgb(0, 0, 0))";
```

addons/editor-theme3/addon.json
```diff
@@ -364,6 +372,34 @@
...
+    {
+      "name": "Insertion marker opacity",
+      "id": "fillOpacity",
+      "type": "integer",
+      "min": 0,
+      "max": 100,
+      "default": 20
+    },
+    {
+      "name": "Insertion marker outline opacity",
```
mxmou, line 397: This setting doesn't work if the style is "gray" because of the `addon.settings.get("markerStyle") !== "none"` check in the userscript.

mxmou (commented):
I still think the fill and outline settings should be separated.

mxmou (approved):

mxmou:
@WorldLanguages The Black presets needs to be migrated.

background/get-addon-settings.js
```diff
@@ -654,6 +654,31 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
...
+              "custom-color": "#FF99AA",
+              "Pen-color": "#13ECAF",
+              "sa-color": "#34E4D0",
+              "comment-color": "#FEF49C",
+              "input-color": "#202020",
+              text: "colorOnBlack",
+              // "monitors": false
+            },
+            manifest.presets.find((p) => p.id === "black")
+          );
```
WorldLanguages, line 681: @mxmou If you want, you can add `//` to more lines. I added it to "monitors" so that the migration happens even if that was enabled by the user but the rest still match. Depending on how terribly blocks will look if this migration does not take place, you may want to make this less strict by commenting out more settings.

WorldLanguages: This migration works the same as the user clicking the Black preset again, so the "monitors" setting will get disabled even if it was enabled prior to the migration.

shock59: Personally I don't think that users who already have Black applied should have their settings updated to fit the new preset as it's such a drastic change and would make the insertion markers look quite different to how they did before the update. I don't really mind either way though

WorldLanguages: So are you saying that, if `settings["text"]==="colorOnBlack"` at the time of the update, then the new settings (`fillStyle`, `fillOpacity`, `strokeStyle` and `strokeOpacity`) should be set to the Black preset values instead of the setting defaults?

> update the monitor-related settings.

Did you mean "insertion markers" instead of monitor?

mxmou: > So are you saying that, if `settings["text"]==="colorOnBlack"` at the time of the update, then the new settings (`fillStyle`, `fillOpacity`, `strokeStyle` and `strokeOpacity`) should be set to the Black preset values instead of the setting defaults?

Yes, that's what I meant.

WorldLanguages: Okay, when I find some time I will update the setting migration logic and merge this PR.

WorldLanguages:
@Samq64 @mxmou I suggest we add a note that explains what an "insertion marker" is (something in the lines of "the shadow when dragging a block near another block"

addons/editor-theme3/addon.json
```diff
@@ -12,7 +12,13 @@
     {
       "name": "DNin01",
       "link": "https://github.com/DNin01"
-    }
+    },
+    {
+      "name": "Shock59",
+      "link": "https://github.com/shock59"
+    },
+    { "name": "Ashimee", "link": "https://github.com/Ashimee" },
```
WorldLanguages, line 20: They deleted their GitHub account, or maybe changed usernames.
Are we sure they want to be credited?

PR #7513, opened 2024-06-10
Messaging: count identical studio-activity messages as many messages

WorldLanguages:
### Changes

If the user hasn't read messages for 20 days and they have received the same "there was activity in X studio" message multiple times, it is shown once, like it always has.

But the number of the "studio activity" row will reflect the actual number of messages that are "eaten" by the category:
image

This was also added to forum activity messages.

### Reason for changes

I've seen users more than once get confused on why we don't display all of their messages. But yes we were displaying all of them. This should help clarify the situation.

### Tests

Tested but without identical messages.

### DIFF
--- a/popups/scratch-messaging/popup.html
+++ b/popups/scratch-messaging/popup.html
@@ -221,7 +221,7 @@
             </div>
             <span class="message-type-title-text">{{ uiMessages.forumMsg }}</span>
             <span class="float-right"
-              ><img class="small-icon" src="../../images/icons/forum.svg" /> {{ forumActivityAmt }}</span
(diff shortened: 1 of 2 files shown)


mxmou:
This is still confusing. Maybe the number of identical messages could be shown next to the message?
image

WorldLanguages:
> This is still confusing.

It's better than nothing and I try to spend my time on higher prio issues.

WorldLanguages:
According to cobalt, forum activity messages work differently, so it makes no sense to change how they work.

> forum activity notifs are sent whenever a new post is added to the thread, and won't ever send you another until you read that one

Issue #6671, opened 2023-09-20
Reconsider `Clean up Blocks+` features, make it separate?, should it be default?


WorldLanguages:
This is about the `Clean up Blocks +` feature from `editor-devtools` which predates Scratch Addons.

- Related: #5630
- Related: #6107
- Related: #6360
- Related: #4131
!!! "Clean Up Blocks +" and script-snap use different grids !!! (why???)
- Related: #3887 
`Clean up blocks+ is already much better than Scratch's clean up blocks, but it would be useful to have it do more without spamming the user with a bunch of alerts or accidentally removing stuff that the user wants to keep.`

Related: multiple feedback messages about why "clean up blocks+" is asking them to delete unused variables.
![image](https://github.com/ScratchAddons/ScratchAddons/assets/17484114/18e0a61f-3d03-4003-8b34-7901ba8f77d3)

We should consider:
1. Whether prompting to remove unused variables should be the extension default after you just installed Scratch Addons. Maybe we can move those things to an off-by-default setting.
2. Issue #3700 suggests a dedicated `Clean up Blocks +` addon that is separate from all others.

DNin01:
I'd actually like to see _everything_ from the developer tools addon be split up into separate addons, like we did with the find bar and another feature or two.

DNin01:
Related: I think I remember somewhere the Scratch Team said that they intentionally went back to Scratch 1.4 behavior (the one-dimensional format layout) instead of laying out the blocks in a grid like in 2.0 because text-based programming languages use a one-dimensional list of lines. Blocks in Scratch can be dragged around in all directions, without having to stick to a vertical plane, but supposedly the team wanted to make the formatted layout more closely resemble code written in text.

BroJac5246:
> Related: I think I remember somewhere the Scratch Team said that they intentionally went back to Scratch 1.4 behavior (the one-dimensional format layout) instead of laying out the blocks in a grid like in 2.0 because text-based programming languages use a one-dimensional list of lines. Blocks in Scratch can be dragged around in all directions, without having to stick to a vertical plane, but supposedly the team wanted to make the formatted layout more closely resemble code written in text.

It's a valid point, but it just doesn't make sense. It's not ideal for this format (block-based coding).

Secret-chest:
Maybe a max columns setting?

BroJac5246:
> Maybe a max columns setting?

Is there a specific reason for suggesting that?

DNin01:
[...]

PR #7519, opened 2024-06-11
New Addon: Separate cleanup-plus tool from dev tools

Joeclinton1:
Does not fully resolve an issue, but is a good first step towards #6671 and #3700

### Changes

I have created a new addon titled "Enhance clean up tools", which is internally titled "editor-cleanup-plus". All code from devtools, that used to do cleanup has been moved to this new addon. Any code that was shared between cleanup and the other tools included within dev tools has been moved to a utils file in devtools.

It also adds dynamic disable to clean up tools since devtools doesn't have it. (dev tools really should be updated to have this but it's not a big deal right now)

### Reason for changes

This is a necessary first step towards splitting up devtools into separate addons, and making cleanup tools easier to maintain and extend. With this change various other issues regarding cleanup can now be addressed.

### Tests

I tested each feature of dev tools one by one, and the dynamic enable/disable for the cleanup tools. Tested on Chrome. Please tell me if I missed anything.


TheColaber (commented):
i havent tested this addon, nor the updated make-space feature

addons/editor-cleanup-plus/addon.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Enhanced clean up tool",
+  "description": "Enhances the clean up tool with smart column-by-column block organisation and automatic removal of unused lists/variables.",
```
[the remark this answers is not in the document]
TheColaber: "Enhanced block clean up"?
"Improves the 'Clean Up' feature in the right click context menus in the editor with smart column organization and suggestions of removing unused variables and lists."? - W_L got this one.

addons/editor-cleanup-plus/userscript.js
```diff
@@ -0,0 +1,142 @@
...
+import { getVariableUsesById, getOrderedTopBlockColumns } from "../editor-devtools/utils.js";
+
+export default async function ({ addon, console, msg, safeMsg: m }) {
+  const blockly = await addon.tab.traps.getBlockly();
+  const getWorkspace = () => blockly.getMainWorkspace();
+
+  let originalMsg = blockly.Msg.CLEAN_UP;
+  addon.self.addEventListener("disabled", () => (blockly.Msg.CLEAN_UP = originalMsg));
+  addon.self.addEventListener("reenabled", () => (blockly.Msg.CLEAN_UP = m("clean-plus")));
+  blockly.Msg.CLEAN_UP = m("clean-plus");
```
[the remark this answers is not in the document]
Joeclinton1: What is the alternative to changing the message?

Samq64 (changes requested):
Is is possible to clean up the diffs or is GitHub just bad at displaying them?

Joeclinton1:
> Is is possible to clean up the diffs or is GitHub just bad at displaying them?

So the cause of the diffs is because it's largely rewritten, and also reformatted, that github can't identify which parts belong to which other ones. It would be challenging to reformat it to make github find the diffs.

addons/editor-devtools/DevTools.js
```diff
@@ -36,64 +34,12 @@ export default class DevTools {
...
-    };
 
-    let originalMsg = blockly.Msg.CLEAN_UP;
-    if (this.addon.settings.get("enableCleanUpPlus")) blockly.Msg.CLEAN_UP = this.m("clean-plus");
-    this.addon.settings.addEventListener("change", () => {
-      if (this.addon.settings.get("enableCleanUpPlus")) blockly.Msg.CLEAN_UP = this.m("clean-plus");
-      else blockly.Msg.CLEAN_UP = originalMsg;
-    });
-
-    this.addon.tab.createBlockContextMenu(
```
Samq64, line 58: This context menu item should stay in devtools. It's the one for pasting blocks.

Samq64 (commented):
I say we merge this first and then refactor/port to modern Blockly after.

Samq64:
> Make space functionality is rewritten to be independent of cleanup blocks

Should it also be separeated into another addon here then since all devtools features will be split off at some point anyway?

Samq64:
Everything seems to work fine including cleanup+ on modern Blockly. It's definitely possible I introduced some regressions though.

addons/editor-cleanup-plus/addon.json
```diff
@@ -0,0 +1,24 @@
...
+  "dynamicDisable": true,
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.39.0",
+  "tags": ["editor", "codeEditor", "recommended"],
+  "enabledByDefault": false
```
Samq64, line 23: Shouldn't this be the other way around? Cleanup+ enabled by default, devtools disabled?

WorldLanguages: On issue #6671 I think I explained some reasons in favor of not enabling cleanup+ by default

Samq64: I agree with unused variable deletion being an off by default setting. I don't see a reason why devtools should stay enabled by default after this though.

Samq64 (approved):

addons/editor-devtools/utils.js
mxmou, line 1: This should probably be in `libraries/common/cs`, otherwise it will need to be moved when the devtools addon is fully split.

addons/editor-devtools/DevTools.js
```diff
@@ -484,12 +249,6 @@ export default class DevTools {
     return ids;
   }
 
-  /**
-   * Initiates a drag event for all block stacks except those in the set of ids.
-   * But why? - Because we know all the ids of the existing stacks before we paste / duplicate - so we can find the
-   * new stack by excluding all the known ones.
-   * @param ids Set of previously known ids
-   */
```
mxmou, line 492: Was this comment intentionally removed? It seems useful to me. Same for a few other comments in the same file.

addons/editor-devtools/addon.json
```diff
@@ -1,18 +1,16 @@
 {
   "name": "Developer tools",
-  "description": "Adds new menu options to the editor: copy/paste blocks, better clean up, and more!",
+  "description": "Adds copy/cut/paste blocks, variable swap options, and make-space option to context menu.",
```
mxmou, line 3:
```suggestion
  "description": "Adds options to copy/cut/paste blocks, swap variables, and make space to the right click context menu.",
```

addons/editor-devtools/addon.json
```diff
@@ -28,6 +26,10 @@
     }
   ],
   "versionAdded": "1.0.0",
+  "latestUpdate": {
+    "version": "1.39.0",
+    "temporaryNotice": "The Cleanup-plus tool has been moved to a separate addon."
```
mxmou, line 31:
```suggestion
    "temporaryNotice": "The \"Clean up Blocks +\" option has been moved to a separate addon."
```

addons/editor-cleanup-plus/addon.json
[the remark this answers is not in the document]
mxmou: > "Enhanced block clean up"? "Improves the 'Clean Up' feature in the right click context menus in the editor with smart column organization and suggestions of removing unused variables and lists."? - W_L got this one.

My suggestions:
* "multiple columns" instead of "smart column organization"
* "in the code area" instead of "in the editor"

addons/editor-cleanup-plus/userscript.js
```diff
@@ -0,0 +1,142 @@
+import UndoGroup from "../editor-devtools/UndoGroup.js";
+import { getVariableUsesById, getOrderedTopBlockColumns } from "../editor-devtools/utils.js";
+
+export default async function ({ addon, console, msg, safeMsg: m }) {
+  const blockly = await addon.tab.traps.getBlockly();
+  const getWorkspace = () => blockly.getMainWorkspace();
```
mxmou, line 6:
```suggestion
  const getWorkspace = () => addon.tab.traps.getWorkspace();
```
devtools was canged to use this API in #7892.

addons/editor-cleanup-plus/userscript.js
```diff
@@ -0,0 +1,142 @@
...
+  blockly.Msg.CLEAN_UP = m("clean-plus");
+
+  const oldCleanUpFunc = blockly.WorkspaceSvg.prototype.cleanUp;
+  blockly.WorkspaceSvg.prototype.cleanUp = function () {
+    if (addon.self.disabled) return oldCleanUpFunc.call(this);
+    doCleanUp();
+  };
+
+  const doCleanUp = () => {
+    let workspace = blockly.getMainWorkspace();
```
mxmou, line 20:
```suggestion
    let workspace = getWorkspace();
```

addons/editor-devtools/DevTools.js
```diff
@@ -19,6 +17,7 @@ export default class DevTools {
     this.canShare = false;
 
     this.mouseXY = { x: 0, y: 0 };
+    this.blockly = null;
```
mxmou, line 20: This doesn't seem to be referenced anywhere.

addons/editor-devtools/DevTools.js
Samq64: @Joeclinton1

addons/editor-cleanup-plus/addon.json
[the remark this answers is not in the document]
Samq64:
```suggestion
  "description": "Improves the clean up feature in the code area's right click context menu with multiple columns and suggestions of removing unused variables and lists."
```

Joeclinton1:
> Everything seems to work fine including cleanup+ on modern Blockly after the merge. It's definitely possible I introduced some regressions though since I didn't double-check if the new cleanup userscript or the util functions need to be updated.

I have checked that the addon still does what it is meant to do. Nothing has regressed. The addon in my opinion, should merged quickly to prevent further delay and avoid stalling progress on further splitting dev tools, or enhance clean up plus in the future.

addons/editor-devtools/DevTools.js
```diff
@@ -8,6 +8,9 @@ export default class DevTools {
     this.addon = addon;
     this.msg = msg;
     this.m = m;
+    /**
+     * @type {VirtualMachine}
+     */
```
mxmou: This comment should actually be removed - the variable it refers to was removed in #4817.

mxmou (changes requested):
Undoing Make Space requires pressing Ctrl+Z multiple times - the events should be grouped so that one undo is enough. Everything else seems to work (tested on Firefox).

Joeclinton1:
> Undoing Make Space requires pressing Ctrl+Z multiple times - the events should be grouped so that one undo is enough. Everything else seems to work (tested on Firefox).

Nice catch! Will fix now.

mxmou:
> oh btw, the copy and paste function from the current state of dev tools has a bug where this isn't undone in one group either. I guess I should fix that too?

Yes, I think it would make sense to fix both if it's a simple fix.

Joeclinton1:
> > oh btw, the copy and paste function from the current state of dev tools has a bug where this isn't undone in one group either. I guess I should fix that too?
> 
> Yes, I think it would make sense to fix both if it's a simple fix.

idk how simple it is going to be, but I will try to fix it.

mxmou:
The but isn't new - I can repro on [this random commit from a few years ago](https://github.com/ScratchAddons/ScratchAddons/commit/621824028eab397634ec7a9326e57ab8e31ee418).

Joeclinton1:
> The bug isn't new - I can repro on [this random commit from a few years ago](https://github.com/ScratchAddons/ScratchAddons/commit/621824028eab397634ec7a9326e57ab8e31ee418).

There's no way undos have required two ctrl-z's for years now, and noone has noticed.

Joeclinton1:
For some reason the event for the drag and the event for the block paste are created like 500ms apart, which is very annoying.

mxmou:
> > The bug isn't new - I can repro on [this random commit from a few years ago](https://github.com/ScratchAddons/ScratchAddons/commit/621824028eab397634ec7a9326e57ab8e31ee418).
> 
> There's no way undos have required two ctrl-z's for years now, and noone has noticed.

It wouldn't be the first time nobody noticed an undo-related bug: #6691 was probably introduced when dynamic enable/disable was added to custom block shape in #4627.

mxmou (approved):

Joeclinton1:
@mxmou Any changes you made in https://github.com/ScratchAddons/ScratchAddons/pull/8073 are going to get overwritten by this pr as it is based on a version of master from before you made those changes. Can you check that this isn't a problem?

mxmou:
The devtools addon will need more changes to make it work on modern Blockly anyway, so it doesn't really matter if this PR breaks something there.

addons/editor-cleanup-plus/addon.json
```diff
@@ -0,0 +1,24 @@
...
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.39.0",
```
Samq64, line 21:
```suggestion
  "versionAdded": "1.43.0",
```

addons/editor-devtools/addon.json
```diff
@@ -28,6 +26,10 @@
     }
   ],
   "versionAdded": "1.0.0",
+  "latestUpdate": {
+    "version": "1.39.0",
```
Samq64, line 30:
```suggestion
    "version": "1.43.0",
```

Samq64 (approved):
This can be merged now. Could you link #4546?

libraries/common/cs/devtools-utils.js
```diff
@@ -41,6 +41,13 @@ export const getTopBlocks = (workspace) => {
   return topBlocks;
 };
 
+export const autoPositionComment = (comment) => {
+    comment.needsAutoPositioning_ = true;
+    comment.autoPosition_()
```
mxmou: This should check if `autoPosition_()` exists - only comments attached to blocks have it. Currently, an error is thrown if there are any standalone comments in the workspace:
image

mxmou (changes requested):

libraries/common/cs/devtools-utils.js
Joeclinton1: Thank you for catching this!

mxmou (approved):

PR #7525, opened 2024-06-13
New addon: Editor workspace dots theme (workspace-dots)

Jazza-231:
Resolves https://discord.com/channels/806602307750985799/1247520844763234356

### Changes

Adds a new addon to customise the style of the workspace dots. This lets them be the default dots, crosshairs, a solid grid, vertical or horizontal lines, or hidden.

### Reason for changes

Customisation is always a good thing, and this can also massively help with the vision impaired.

### Tests

It works with all addons as far as I know, particularly with the editor dark mode addon and the change snapping grid size addon.

Workspace dots


addons/workspace-dots/addon.json
```diff
@@ -0,0 +1,40 @@
...
+      ]
+    },
+    {
+      "name": "Use a custom spacing value",
+      "id": "useSpacing",
+      "type": "boolean",
+      "default": true
+    },
+    {
+      "name": "Spacing value",
```
Samq64, line 29: This is confusing when combined with `script-snap`'s setting.

Jazza-231: Yes, yes it is, but it is the exact feature the original suggestion wanted, and it does have its place. I believe they wanted to mimic the style of scratch 1.4? But I do agree it is confusing, perhaps a notice or better wording?

GarboMuffin:
havent looked at the code but just from the gif, can you make it so the patterns with overlapping lines don't have darker/brighter spots where they overlap? perhaps by not using transparency

Jazza-231:
> havent looked at the code but just from the gif, can you make it so the patterns with overlapping lines don't have darker/brighter spots where they overlap? perhaps by not using transparency

Transparency is defined in editor dark mode
```json
    {
      "name": "workspace-dots",
      "value": {
        "type": "textColor",
        "black": "rgba(0, 0, 0, 0.13)",
        "white": "rgba(255, 255, 255, 0.13)",
        "source": {
          "type": "settingValue",
          "settingId": "workspace"
        }
      }
    },
```

GarboMuffin:
then I think editor dark mode should be patched or find some creative blend mode use to make the overlap look consistent throughout

Jazza-231:
> then I think editor dark mode should be patched or find some creative blend mode use to make the overlapping parts not stand out

Yeah it could be patched in editor dark mode or in this addon, the problem is that the background fill is a pattern element, so even if I remove the opacity from the stroke attribute and add opacity to the pattern element, no opacity or opacity filter or anything works, and stroke opacity gets the same problem we have now

mxmou:
> > then I think editor dark mode should be patched or find some creative blend mode use to make the overlapping parts not stand out
> 
> Yeah it could be patched in editor dark mode or in this addon, the problem is that the background fill is a pattern element, so even if I remove the opacity from the stroke attribute and add opacity to the pattern element, no opacity or opacity filter or anything works, and stroke opacity gets the same problem we have now

Editor dark mode could be changed to use a solid color. I'll make a PR.

Jazza-231:
> Editor dark mode could be changed to use a solid color. I'll make a PR.

Scratch does that, I think we didn't so that it would match better with the background if they chose a colour that wasn't the black or white defaults, but either we make a new setting for the workspace dots instead of it being an automatic variable, leave it as the two colours for light and dark it is and remove the transparency, or do some math to calculate the colour it would be if there was transparency then just give that rgb (no a) to the dots variable

mxmou:
> do some math to calculate the colour it would be if there was transparency then just give that rgb (no a) to the dots variable

I'll try to do that. The `addon.json` format already supports it (using `"type": "alphaBlend"`).

addons/workspace-dots/addon.json
TheColaber: should script snapping get rid of the setting? however technically (other than design/looks) this addon is only useful when paired with script-snapping, but i might be wrong. we should discuss what should modify this scratch editor setting because it probably shouldnt be in 2 places, it should be consistent.

addons/workspace-dots/addon.json
```diff
@@ -0,0 +1,40 @@
...
+        { "id": "dots", "name": "Default dots" },
+        { "id": "crosshairs", "name": "Crosshairs" },
+        { "id": "lines", "name": "Solid lines" },
+        { "id": "vertical", "name": "Vertical Lines" },
+        { "id": "horizontal", "name": "Horizontal Lines" },
+        { "id": "none", "name": "None" }
+      ]
+    },
+    {
+      "name": "Use a custom spacing value",
```
TheColaber, line 23: whats the point of this setting?

addons/workspace-dots/addon.json
TheColaber: wait... whats the difference between `spacing_` and `scale_`?

addons/workspace-dots/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+      this.gridPattern_.setAttribute("width", safeSpacing);
+      this.gridPattern_.setAttribute("height", safeSpacing);
+
+      var half = Math.floor(spacing / 2) + 0.5;
+      var start = half - this.length_ / 2;
+      var end = half + this.length_ / 2;
+
+      half *= scale;
+      start *= scale;
+      end *= scale;
```
TheColaber, line 31: this code is (mostly) the same as on https://github.com/scratchfoundation/scratch-blocks/blob/develop/core/grid.js#L136. why not just run `oldFunction.call(this, spacing)`?

addons/workspace-dots/userscript.js
```diff
@@ -0,0 +1,81 @@
...
+
+    blockly.getMainWorkspace().grid_.update(blockly.getMainWorkspace().grid_.scale_);
+  }
+
+  updateGrid();
+
+  addon.settings.addEventListener("change", updateGrid);
+
+  addon.self.addEventListener("disabled", () => {
+    blockly.getMainWorkspace().grid_.update = oldFunction;
```
TheColaber, line 75: you shouldnt set this back to the old function because if another addon pollutes this function after this addon, after this addon gets disabled, the other addon will no longer work. instead, use an `if (!addon.self.disabled)` inside the polluted function.

addons/workspace-dots/addon.json
Samq64: The two settings technically do different things since this one is only for the appearance.

addons/workspace-dots/addon.json
Jazza-231: That disables the custom spacing setting and uses the default spacing provided by the snapping addon

addons/workspace-dots/addon.json
Jazza-231: The point of this setting is rather confusing but it's main use is to provide a grid that does NOT match up with the snapping addon, for visual purposes only, and does not affect snapping. The grid would become as useless as if you turned the grid off, which is also a visual choice

addons/workspace-dots/userscript.js
Jazza-231: It's a stupid reason, i recently got terrified of .call after so many issues with what this object to pass through it with the disable auto nesting addon, but I will change it. Let me dm you tho

addons/workspace-dots/userscript.js
Jazza-231: Good point.

addons/workspace-dots/userscript.js
Jazza-231: Just realise i literally cant because i need to do calculations inside of that function, no way around it, so it has to be modified fully and not just called with scripts before or after

addons/workspace-dots/userscript.js
Jazza-231: Done

addons/workspace-dots/userscript.js
mxmou: I think something like this would work:
```js
blockly.getMainWorkspace().grid_.update = function (scale) {
  const oldSpacing = this.spacing_;
  if (spacingSetting & !addon.self.disabled) this.spacing_ = spacingAmount;
  oldFunction.call(this, scale);

  let strokeWidthX;
  let strokeWidthY;
  // ...
  this.line1_.setAttribute("stroke-width", strokeWidthY);
  this.line2_.setAttribute("stroke-width", strokeWidthX);

  this.spacing_ = oldSpacing;
}

Comment (Jazza-231) [reply addons/workspace-dots/userscript.js]: Is there any real reason? its already a small enough function

Comment (mxmou) [reply addons/workspace-dots/userscript.js]: The original function might change in a Scratch update or be overridden by another addon in a future SA version. If you decide to keep the copied code anyway, there should be a comment indicating where it's from.

File: addons/workspace-dots/addon.json (hunk 1 above)
Comment (TheColaber) [reply addons/workspace-dots/addon.json:29]: this is weird and my brain hurts :P why would we not want to be consistent with the snapping?

File: addons/workspace-dots/userscript.js [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+
+  const oldFunction = blockly.getMainWorkspace().grid_.update;
+
+  function updateGrid() {
+    let themeSetting = addon.settings.get("theme");
+    let spacingSetting = addon.settings.get("useSpacing");
+    let spacingAmount = addon.settings.get("spacing");
+
+    // https://github.com/scratchfoundation/scratch-blocks/blob/develop/core/grid.js#L136
+    blockly.getMainWorkspace().grid_.update = function (scale) {
```
TheColaber, line 12: why isnt this on `Blockly.Workspace.Grid.prototype.update`? Wouldn't that break when leaving the editor? At the very least it would probably break if you were to enable the addon after creating a custom block?

addons/workspace-dots/userscript.js
```diff
@@ -0,0 +1,84 @@
...
+      this.setLineAttributes_(this.line1_, strokeWidthY, start, end, half, half);
+      this.setLineAttributes_(this.line2_, strokeWidthX, half, half, start, end);
+    };
+
+    blockly.getMainWorkspace().grid_.update(blockly.getMainWorkspace().grid_.scale_);
+  }
+
+  updateGrid();
+
+  addon.settings.addEventListener("change", updateGrid);
```
TheColaber, line 76: `updateGrid.call(this)`?

addons/workspace-dots/userscript.js
TheColaber: Ohhh also I noticed you pollute everytime `updateGrid` is called. you should only pollute once and then call `workspace.grid.update()` when `updateGrid` is called.

addons/workspace-dots/addon.json
Samq64: So it's possible to have a Scratch 1.x theme and still be able to snap normally. I'm not sure it's worth it though.

TheColaber: @Samq64 can you expand bc im unfamilar with how scratch 1.x behavior is.

Samq64: https://en.scratch-wiki.info/w/images/Scratch_Interface.png

The workspace has a vertical line pattern.

TheColaber: but you'd probably want the blocks to still align with a line, but maybe not just every line right? so maybe the setting could be something where it divides the snap spacing by a value?

addons/workspace-dots/userscript.js
Jazza-231: Remind me about this tmr

addons/workspace-dots/userscript.js
Jazza-231: Nah this works fine, if you pass a function without calling it then it will just be called when the settings change

addons/workspace-dots/userscript.js
Jazza-231: This is hell. Read from https://discord.com/channels/751206349614088204/755131743136383107/1252657263257059419

addons/workspace-dots/addon.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "Editor workspace dots theme",
+  "description": "Allows the customisation of the style of the editor workspace dots.",
```
mxmou, line 3:
```suggestion
  "name": "Customizable code editor grid",
  "description": "Change the style of the grid pattern in the code editor.",
```

addons/workspace-dots/addon.json
```diff
@@ -0,0 +1,34 @@
...
+  "tags": ["editor", "codeEditor", "theme"],
+  "versionAdded": "1.39.0",
+  "userscripts": [{ "url": "userscript.js", "matches": ["projects"] }],
+  "settings": [
+    {
+      "name": "Dots theme",
```
mxmou, line 9:
```suggestion
      "name": "Grid style",
```

addons/workspace-dots/addon.json
```diff
@@ -0,0 +1,34 @@
...
+        { "id": "dots", "name": "Default dots" },
+        { "id": "crosshairs", "name": "Crosshairs" },
+        { "id": "lines", "name": "Solid lines" },
+        { "id": "vertical", "name": "Vertical Lines" },
+        { "id": "horizontal", "name": "Horizontal Lines" },
+        { "id": "none", "name": "None" }
+      ]
+    },
+    {
+      "name": "Unnamed setting",
```
mxmou, line 23: This setting should be hidden if `theme` is `none`. I think "Spacing" would be a good name for it, but that would only make sense if higher values meant bigger spacing - the original spacing could be divided by 5 and then multiplied by this setting.

addons/workspace-dots/userscript.js
```diff
@@ -0,0 +1,58 @@
+export default async function ({ addon, console }) {
+  const ScratchBlocks = await addon.tab.traps.getBlockly();
+
+  // https://github.com/scratchfoundation/scratch-blocks/blob/develop/core/grid.js#L136
+  const oldUpdate = ScratchBlocks.Grid.prototype.update;
+  ScratchBlocks.Grid.prototype.update = function (scale) {
+    const spacingMultiplier = addon.settings.get("spacingMultiplier");
+    if (!addon.self.disabled) this.spacing_ /= spacingMultiplier;
+    oldUpdate.call(this, scale);
+    if (!addon.self.disabled) this.spacing_ *= spacingMultiplier;
```
mxmou, line 10: Storing the original spacing is better, especially if the calculation becomes more complex - see above.

addons/workspace-dots/addon.json
Jazza-231: > This setting should be hidden if `theme` is `none`.

Yes. I don't think "Spacing" is a good name thp

addons/workspace-dots/userscript.js
Jazza-231: I will do this

Jazza-231:
From @mxmou:
>I think "Spacing" would be a good name for it, but that would only make sense if higher values meant bigger spacing - the original spacing could be divided by 5 and then multiplied by this setting.

@WorldLanguages @TheColaber Lets get some more ideas for what this setting should be named, its functionality shouldn't be changed as having it as a divisor of the actual spacing is the best idea

TheColaber:
The current functionality is: a value of 1 would result in one dot per row/column of spacing, a value of 2 would result in 2 dots per row/column of spacing, and so on. It's tricky to find a good name for this.

CST1229:
> The current functionality is: a value of 1 would result in one dot per row/column of spacing, a value of 2 would result in 2 dots per row/column of spacing, and so on. It's tricky to find a good name for this.

"Density"??

TheColaber (approved):
Other than name for the setting, this is good to go. We need the name. Maybe something like Grid Repeat or something?

mxmou:
> Other than name for the setting, this is good to go. We need the name. Maybe something like Grid Repeat or something?

I like the "density" suggestion - maybe "grid line density" would be more clear.

addons/workspace-dots/addon.json
```diff
@@ -0,0 +1,35 @@
...
+      "id": "spacingMultiplier",
```
mxmou, line 24:
```suggestion
      "id": "spacingDivisor",
```

Jazza-231:
> > Other than name for the setting, this is good to go. We need the name. Maybe something like Grid Repeat or something?
> 
> I like the "density" suggestion - maybe "grid density", "grid line density" or "grid line frequency" would be less ambiguous.

I like grid density, committed that

addons/workspace-dots/addon.json
```diff
@@ -0,0 +1,35 @@
...
+      "name": "Grid style",
+      "id": "theme",
+      "type": "select",
+      "default": "dots",
+      "potentialValues": [
+        { "id": "dots", "name": "Default dots" },
+        { "id": "crosshairs", "name": "Crosshairs" },
+        { "id": "lines", "name": "Solid lines" },
+        { "id": "vertical", "name": "Vertical Lines" },
+        { "id": "horizontal", "name": "Horizontal Lines" },
```
mxmou, line 18:
```suggestion
        { "id": "vertical", "name": "Vertical lines" },
        { "id": "horizontal", "name": "Horizontal lines" },
```

mxmou (approved):

Jazza-231:
🔥

addons/workspace-dots/userscript.js
```diff
@@ -0,0 +1,59 @@
...
+            return DOT_LENGTH;
+        }
+      })();
+    }
+    oldSetLineAttr.call(this, line, width, x1, x2, y1, y2);
+  };
+
+  function updateGrid() {
+    const workspace = addon.tab.traps.getWorkspace();
+    const grid = workspace.getGrid();
```
WorldLanguages, line 50: This will log an error in the console if workspace is null, but it's fine.

WorldLanguages: This is equivalent to how `custom-block-text` works, though, so it's fine.

WorldLanguages:
Hey, this removed a setting. Now what? Don't you want to migrate it?

Jazza-231:
> Hey, this removed a setting. Now what? Don't you want to migrate it?

I'm not entirely sure what you are referring to?

DNin01:
> > Hey, this removed a setting. Now what? Don't you want to migrate it?
> 
> I'm not entirely sure what you are referring to?

Sometimes when we make changes to settings such as replacing a setting or two with others, we need to go through and "migrate" them when Scratch Addons updates so they continue to work without having to be reconfigured by the user.

For example:
- If we combined a set of on/off-switch settings into a single selection setting, we would need to set that new selection setting appropriately based on the values of the on/off-switch settings that are being replaced.
- If a color on Scratch changes, we would also want to update that color in a user's settings, which we did the last time Scratch's accent color was changed.

This is done [in `get-addon-settings.js`](https://github.com/ScratchAddons/ScratchAddons/blob/master/background/get-addon-settings.js#L106).

Jazza-231:
> Hey, this removed a setting. Now what? Don't you want to migrate it?

Just to remind myself, this is handled by https://github.com/ScratchAddons/ScratchAddons/pull/7573

PR #7526, opened 2024-06-13
New addon: Sprite clone count

Jazza-231:
Resolves https://discord.com/channels/806602307750985799/1239865954708492318

### Changes

Adds a new addon that displays individual sprite's clone counts.

### Reason for changes

It's just wildly useful

### Tests

Dynamic enable and disable work fine, no known incompatibilities


Jazza-231:
> Why not add this to the existing clone count addon?

That addon is scary (I'll do it later)

Jazza-231:
@Samq64 Done:tm:

addons/clones/addon.json
```diff
@@ -12,6 +12,10 @@
     {
```
TheColaber, line 12: we probably need to update the description

addons/clones/addon.json
```diff
@@ -41,11 +49,19 @@
       "name": "Show icon only",
       "default": false,
       "type": "boolean"
+    },
+    {
+      "id": "showSpriteCount",
```
TheColaber, line 54: maybe the setting should be to see one or the other or both? select setting? not really sure tbh...

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,68 @@
...
+      spriteName.querySelector(".clone-count")?.remove();
+      if ((counts[spriteName.innerText.split("\n")[0]] !== undefined) & !remove) {
+        const count = document.createElement("div");
+
+        count.classList.add("clone-count");
+        count.innerText = `(${counts[spriteName.innerText.split("\n")[0]]} clones)`;
+        spriteName.appendChild(count);
+      }
+    });
+  }
```
TheColaber, line 32: maybe there is a different method (similar to how the stage clone count already works) to not have to delete and create an element everytime.

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,68 @@
...
+      }, 100);
+    };
+
+    updateCounts(false);
+  }
+
+  function unpolute() {
+    vm.runtime.changeCloneCounter = ogCloneCounter;
+    updateCounts(true);
+  }
```
TheColaber, line 51: dont unpollute :) use the if check.

TheColaber (commented):
jazza with these fire addons 🔥

addons/clones/addon.json
Jazza-231: "description": "Adds a counter above the stage in the editor which shows the total amount of clones. There is also an option to add a counter to individual sprites.",

addons/clones/addon.json
Jazza-231: Eh having it as a select setting doesn't really make sense, its not options for the same setting its two different features

addons/clones/spriteCount.js
Jazza-231: I could make it not delete and re-add for things like just updating the count but whenever a sprite rename occurs it does need to be readded. I'll look into it.

addons/clones/spriteCount.js
Jazza-231: Yea I discovered that literally 2 days ago lmao, thanks for the reminder

Jazza-231: Doneeee

Jazza-231:
> Oh also changing sprite names works too, with a tinnnnyyy little flicker that we can not get around afaik

Heheheheheheh I fixeddd ittttttt

addons/clones/spriteCount.js
Jazza-231: Doneeeeee

addons/clones/addon.json
Jazza-231: Doneeeeeeeeee

Jazza-231:
> jazza with these fire addons 🔥

Clobr with these fire reviews cuz you just accidentally solved my biggest issue with this addon: the slight flicker when changing a sprite name

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,67 @@
...
+
+        if (existingElement) {
+          count = existingElement;
+        } else {
+          count = document.createElement("div");
+          count.classList.add("sa-clone-count");
+          spriteName.appendChild(count);
+        }
+
+        count.innerText = `(${counts[spriteName.innerText.split("\n")[0]]} clones)`;
```
TheColaber, line 38: needs to be localized

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,67 @@
...
+    vm.runtime.targets
+      .filter((target) => !target.isOriginal)
+      .map((target) => target.sprite.name)
+      .forEach((target) => {
+        if (!counts[target]) counts[target] = 1;
+        else counts[target] += 1;
+      });
+
+    const spriteWrapper = await addon.tab.waitForElement("[class*=sprite-selector_items-wrapper]");
+    const spriteNames = Array.from(spriteWrapper.querySelectorAll("[class*=sprite-selector-item_sprite-name]"));
```
TheColaber, line 18: we shouldnt* need a waitForElement here. You can probably merge these 2 lines into one and just use one querySelectorAll

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,67 @@
...
+      updateCounts();
+    });
+  });
+
+  addon.self.addEventListener("disabled", () => updateCounts());
+  addon.self.addEventListener("reenabled", () => updateCounts());
+
+  addon.settings.addEventListener("change", () => {
+    updateCounts();
+  });
```
TheColaber, line 66: this is just me nitpicking but can you turn these into `addon.self.addEventListener("reenabled", updateCounts)`

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,67 @@
...
+    const spriteWrapper = await addon.tab.waitForElement("[class*=sprite-selector_items-wrapper]");
+    const spriteNames = Array.from(spriteWrapper.querySelectorAll("[class*=sprite-selector-item_sprite-name]"));
+
+    spriteNames.forEach((spriteName) => {
+      if (counts[spriteName.innerText.split("\n")[0]] !== undefined) {
+        const existingElement = spriteName.querySelector(".sa-clone-count");
+        let count;
+
+        if (remove) {
+          existingElement?.remove();
```
TheColaber, line 26: can we hide instead of remove? its always faster to show/hide dom rather than delete/recreate.

addons/clones/spriteCount.js
Jazza-231: We do, i dont know why but we do need the wait for element

addons/clones/spriteCount.js
Jazza-231: ah yes

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,67 @@
...
+    vm.runtime.changeCloneCounter = (e) => {
+      ogCloneCounter.call(vm.runtime, e);
+
+      queueMicrotask(() => {
+        updateCounts();
+      });
+    };
+  }
+
+  pollute();
```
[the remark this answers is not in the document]
Jazza-231: eh its cleaner

addons/clones/spriteCount.js
Jazza-231: i could have sworn i did this

addons/clones/spriteCount.js
Jazza-231: remove is only used for when the addon is disabled, but yeah ill look into it

addons/clones/spriteCount.js
Jazza-231: Or we dont? wtf

addons/clones/spriteCount.js
```diff
@@ -0,0 +1,56 @@
...
+      .filter((target) => !target.isOriginal)
+      .forEach((target) => {
+        const name = target.sprite.name;
+        counts[name] = (counts[name] || 0) + 1;
+      });
+
+    const spriteNames = Array.from(
+      document
+        .querySelector("[class*=sprite-selector_items-wrapper]")
+        ?.querySelectorAll("[class*=sprite-selector-item_sprite-name]")
```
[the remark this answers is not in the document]
TheColaber: also, we should look into not having to call querySelector every frame of a project as its probably slow in big projects. not sure tho.

TheColaber (approved):
🍡

addons/clones/spriteCount.js
Samq64, line 1: A single userscript per addon is usually preferred.

WorldLanguages: The two features are far from identical, I'm not sure if any code is shared at all

Samq64 (approved):

WorldLanguages:
What's the usecase?

Samq64:
I'm not sure, since most clone leaks will happen in the current sprite being worked on.

WorldLanguages (changes requested):
> So I have this problem with my game cause it is making way to many clones, but I don't know which sprites get cloned so I can't fix it, that's why a way to know which sprites are cloned is my idea

Only one person had this problem, does it make sense to create a new addon just for them?

jazzathedev:
> > So I have this problem with my game cause it is making way to many clones, but I don't know which sprites get cloned so I can't fix it, that's why a way to know which sprites are cloned is my idea
> 
> Only one person had this problem, does it make sense to create a new addon just for them?

Found some recent motivation, Hi all.

I have faced the issue of not knowing where clones came from on multiple occasions. It's rather useful, especially for people newer to scratch, to have this data. Also i think all of the requested changes were addressed, so feel free to take another look if i need to do anything else

PR #7529, opened 2024-06-13
[module.js] Modify user-agent everywhere on forums

### LINKED ISSUE #6065: Forum user-agent is not always modified ("version"→"versions")
> (also I'm not sure if this is just me, but it shows as "version" for me and not "versions" so I had no idea, is that meant to happen)
> Nevermind, it only happens when you create a new topic in Bugs and Glitches, not when you create a post. Though that could be an issue because I've seen people who remove the user agent part from their topic, and then post it later.

_Originally posted by @medlans in https://github.com/ScratchAddons/ScratchAddons/issues/6062#issuecomment-1550939949_

OneShot-Niko:
Resolves #6065

### Changes

~~Changes the if check condition to see if the URL includes `topic` (which means both topic creation and topic pages are now included) or `/discuss/settings/` (even though it's a bit broken on Scratch's side irregardless of SA, better everywhere than nowhere) instead of only checking if the URL is the new topic creation page on BaG.~~

~~AKA replaces the pathname comparison from a complete match to just if the pathname `includes` either one of the ones mentioned above~~
Uses RegExp to just check if the URL contains `topic` or `discuss/settings`, which makes it change UA everywhere on the forums.

### Reason for changes

if the question ends up in the wrong forum or the OP doesn't leave their user agent in the original post, then that ends up needing an extra question that might even be forgotten.

### Tests

Tested on Firefox 126.0.1 and Chromium (Vivaldi 6.7.3329.41 (Stable channel) (32 bit))

https://github.com/ScratchAddons/ScratchAddons/assets/150537842/fe113224-4bc2-4162-a041-80a8eeb61f1c

ironically encountering a bug where version is double-pluralized (`versionss` instead of `versions`), console had no err or warn though so not sure what happened there

### DIFF
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -302,7 +302,7 @@ if (isScratchGui || isProject) {
   }
 }
 
-if (window.location.pathname.includes("topic") || window.location.pathname.includes("discuss/settings/")) {
+if (location.pathname.includes("topic") || location.pathname.includes("discuss/settings/")) {
   const checkUA = () => {
     if (!window.mySettings) return false;
     const ua = window.mySettings.markupSet.find((x) => x.className);
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -302,7 +302,8 @@ if (isScratchGui || isProject) {
   }
 }
 
-if (location.pathname.match(/^(?:https?:\/\/scratch\.mit\.edu\/)?(topic|\/discuss\/settings)?/)) {
+// Same as editingScreens match
+if (location.pathname.match(/^\/discuss\/(?:topic\/\d+|\d+\/topic\/add|post\/\d+\/edit|settings\/[\w-]+)\/?$/) {
   const checkUA = () => {
     if (!window.mySettings) return false;
     const ua = window.mySettings.markupSet.find((x) => x.className);
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -302,7 +302,7 @@ if (isScratchGui || isProject) {
   }
 }
 
-if (window.location.pathname.includes("topic")) {
+if (window.location.pathname.includes("topic") || window.location.pathname.includes("discuss/settings/")) {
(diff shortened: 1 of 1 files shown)


content-scripts/inject/module.js
```diff
@@ -302,7 +302,7 @@ if (isScratchGui || isProject) {
   }
 }
 
-if (location.pathname === "/discuss/3/topic/add/") {
+if (location.pathname.includes("topic") || location.pathname.includes("discuss/settings/")) {
```
Samq64, line 305: It would make sense to use Regex here.

OneShot-Niko: Will do in a bit, i'm just really bad at regex so it might take awhile

DNin01: You might like https://github.com/ziishaned/learn-regex

OneShot-Niko: alright, done
> You might like [ziishaned/learn-regex](https://github.com/ziishaned/learn-regex)

thank you for that!

content-scripts/inject/module.js
```diff
@@ -302,7 +302,7 @@ if (isScratchGui || isProject) {
   }
 }
 
-if (location.pathname === "/discuss/3/topic/add/") {
+if (location.pathname.match(/^(?:https?:\/\/scratch\.mit\.edu\/)?(topic|\/discuss\/settings)?/)) {
```
Samq64, line 305: The `editingScreens` match uses this:
```suggestion
// Same as editingScreens match
if (location.pathname.match(/^\/discuss\/(?:topic\/\d+|\d+\/topic\/add|post\/\d+\/edit|settings\/[\w-]+)\/?$/) {
```
But now that I think about it it's fine if the match isn't perfect.

Samq64:
When editing a post the existing contents get replaced by the user-agent.

If you don't mind I'm thinking of just addonifing the whole thing in a separate PR.

WorldLanguages:
Sure go ahead

PR #7532, opened 2024-06-14
New addon: Disable auto nesting

Jazza-231:
Resolves https://discord.com/channels/806602307750985799/1247667402079207514

### Changes

Adds a new addon that lets you hold a modifier key to stop a c block expanding around the blocks below it

### Reason for changes

Convenience, its so freaking useful, and griffpatch suggested it (even tho I suggested it two years ago)

### Tests

Works when you hold ctrl and drag, doesn't update when you press or unpress ctrl without dragging enough for a new connection spot to be made, garbo might help with that

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -156,6 +156,7 @@
   "reorder-custom-inputs",
   "place-backpack-code-at-cursor",
   "big-save-button",
+  "disable-auto-nesting",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/disable-auto-nesting/addon.json
+++ b/addons/disable-auto-nesting/addon.json
@@ -1,0 +1,41 @@
+{
+  "name": "Disable auto nesting with C-Blocks",
+  "description": "When holding the ctrl key (customisable) while dragging a C-Block, they will not expand around the block stack, and will instead place like a normal block.",
+  "credits": [
+    {
+      "name": "Jazza",
+      "link": "https://scratch.mit.edu/users/greeny--231/"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "tags": ["editor", "featured"],
+  "versionAdded": "1.39.0",
+
+  "info": [
+    {
+      "type": "notice",
+      "text": "Examples of a C-Block are if statements or forever loops.",
+      "id": "cBlockExample"
+    }
+  ],
+  "settings": [
+    {
+      "name": "Modifier key",
+      "type": "select",
+      "id": "key",
+      "default": "ctrl",
+      "potentialValues": [
+        { "id": "ctrl", "name": "Control" },
+        { "id": "alt", "name": "Alt" },
+        { "id": "shift", "name": "Shift" }
+      ]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true
+}
--- a/addons/disable-auto-nesting/userscript.js
+++ b/addons/disable-auto-nesting/userscript.js
@@ -1,0 +1,69 @@
+export default async function ({ addon, console }) {
+  const ScratchBlocks = await addon.tab.traps.getBlockly();
+  const oldFunction = ScratchBlocks.Block.prototype.getFirstStatementConnection;
+  const settingsToKey = {
+    ctrl: "Control",
+    alt: "Alt Meta",
+    shift: "Shift",
+  };
+
+  let modifierHeld = false;
+
+  function keyDown(e, key) {
+    if (settingsToKey[key].includes(e.key)) {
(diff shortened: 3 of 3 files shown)


OneShot-Niko:
> hold ctrl and drag

Wouldn't that conflict with `block-cherry-picking`?

Jazza-231:
Thank you @GarboMuffin for your hard work, I have tested the following
- Dynamic enable and disable
- Every setting for the modifier key working as expected
- Going in and out of the editor
- Terminating c blocks like the forever loop
- Non-C blocks behave the same
- More complex c blocks like the if else

And they all work. Reviews and approvals would be greatly appreciated!

GarboMuffin:
it should be documented that on macOS the ctrl option is actually command and alt is actually option

WorldLanguages:
@Jazza-231 Can we decide on a single key instead of making it customizable?
I wonder if we have any modifier key which isn't taken by any other addons.

Jazza-231:
> Can we decide on a single key instead of making it customizable?

Why?

> I wonder if we have any modifier key which isn't taken by any other addons.

Yeah the shift key isn't used in the editor, which makes it a worse choice. I personally would only use ctrl or alt, but customisation is always a good thing.

WorldLanguages:
In Discord, Griffpatch suggested Ctrl as the preferred key for this feature.

Also on Discord, I told Jazza that we could consider alternative ways to make the key customizable.
As a first step towards a "hotkey manager" discussed in various issues, we could offer users a beta/experimental addon that if enabled, allows you to customize the keys that any supported (and enabled) addons listen to. That would include this new addon.

Reading other addon's settings can be achieved by sharing a JS module, and it's trivial if both sides cooperate. An alternative to this is extending the `addon.json` schema to allow an addon setting to be displayed within a different addon's row in the settings page. In paper, each addon would own its own setting, but when displaying the settings page UI, all these settings would be shown as if they were part of this new experimental addon.

Combining all key customization settings into a single addon is a good first step towards future efforts. It also has the benefit of adding a beta/experimental tag, and adding clear notices on what happens if two addons listen to the same key. Lastly, for users that really want customizations, they can use this new feature even though we haven't made any progress into a dedicated hotkey manager UI in our settings page yet.

lisa-wolfgang:
> In Discord, Griffpatch suggested Ctrl as the preferred key for this feature.

I'd agree that Ctrl is a sensible default. The only other editor addon that uses Ctrl (block cherry-picking, known as "Grab single block with Ctrl key") is a feature used when dragging blocks out of a stack, while this feature is used when dragging blocks into a stack.

jazzathedev:
I believe this is ready to go

mxmou:
This will need changes to be compatible with new Blockly (can be done after merging).

Regarding the modifier key setting - I think it's better not to have that until there's a system to prevent conflicts between addons. Without such a system, a setting like this essentially prevents us from adding another addon that uses any of these keys in the same context (dragging blocks) unless we allow conflicts and let the user deal with them (which is likely worse for the user than not having customization).

jazzathedev:
What is this new Blockly? Is this something on scratch's end or ours?

And yes, I agree that for now we shouldn't have it as a setting, if we need to remove the ability to change it for whatever future reason before conflict prevention, im sure users would be pissed

Samq64:
> What is this new Blockly? Is this something on scratch's end or ours?

Scratch is migrating `scratch-blocks` to use Blockly as a libary instead of the current fork. Our tacking issue for it is #7768.

jazzathedev:
Ok as mxmou said, we can merge this and then I'll take a lok at the new blockly, if anyone has a quick link to a PR where we move to the new blockly just so i can see the process that would be great, i will start learning that later today

mxmou:
[There are lots of PRs updating addons to support new Blockly,](https://github.com/ScratchAddons/ScratchAddons/pulls?q=is%3Apr+%22new+blockly%22) and the instructions for setting it up for testing are in #7768. You can use `if (ScratchBlocks.registry)` in the addon's code to detect the update. Note that, in the new version, the `ScratchBlocks` object contains both [things exported by google/blockly](https://github.com/google/blockly/blob/develop/core/blockly.ts) and [those added by Scratch](https://github.com/scratchfoundation/scratch-blocks/blob/spork/src/index.ts).

CST1229:
> [There are lots of PRs updating addons to support new Blockly,](https://github.com/ScratchAddons/ScratchAddons/pulls?q=is%3Apr+%22new+blockly%22) and the instructions for setting it up for testing are in #7768. You can use `if (ScratchBlocks.registry)` in the addon's code to detect the update. Note that, in the new version, the `ScratchBlocks` object contains both [things exported by google/blockly](https://github.com/google/blockly/blob/develop/core/blockly.ts) and [those added by Scratch](https://github.com/scratchfoundation/scratch-blocks/blob/spork/src/index.ts).

by the way there's also https://scratchfoundation.github.io/scratch-editor/spork/scratch-gui/ for an easier way to test new blockly

PR #7534, opened 2024-06-14
Add `user-select: none` and `draggable=false` to some elements in the extension UI

mxmou:
### Changes

Popup after pressing Ctrl+A on v1.38.2:
image

After this PR:
image

Longer blocks of text, such as addon descriptions on the settings page and comments in the Messaging popup, can still be selected. `draggable="false"` was added to all images except the screenshot on the Permissions page.

### Tests

Tested on Edge and Firefox.


Samq64:
I think addon names and maybe also setting names should stay selectable.

Samq64:
Is there a way to make all SVGs non-draggable by default?

mxmou:
> Is there a way to make all SVGs non-draggable by default?

I don't think so - Chrome supports the non-standard CSS property `-webkit-user-drag` but there's no equivalent in Firefox. Inline SVGs are non-draggable by default and it's also easier to change their color in light mode, but switching to those would be more complex than adding the draggable attribute to every icon.

WorldLanguages (approved):
Untested

DNin01:
> > Is there a way to make all SVGs non-draggable by default?
> 
> I don't think so - Chrome supports the non-standard CSS property `-webkit-user-drag` but there's no equivalent in Firefox. Inline SVGs are non-draggable by default and it's also easier to change their color in light mode, but switching to those would be more complex than adding the draggable attribute to every icon.

Would using inline SVGs be viable with a build step?

hueychen27:
How about the version and the "Scratch Addons" name be selectable too?

mxmou:
@WorldLanguages Do you think this could be merged for v1.40 or v1.41? Also #4656, which has been open for over two years and there haven't been any objections.

Samq64:
Could the version number in the popup header and game titles stay selectable?

mxmou:
> Could the version number in the popup header and game titles stay selectable?

Being able to copy the version number can be useful, but I'm not sure about cloud game titles. Project titles in the Messaging tab have always been unselectable.

Samq64 (approved):

PR #7546, opened 2024-06-18
gamepad: dark mode support

mxmou:
### Changes

image

The light mode colors were also changed to be more consistent with Scratch's color palette.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -458,6 +458,17 @@
       }
     },
     {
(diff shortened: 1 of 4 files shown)


CubesterYT:
Finally, I'm surprised it took this long.

WorldLanguages (approved):

Samq64 (approved):
Using editor dark mode on the project page can be a little confusing but I can't think of a better solution and it's much better than before.

Issue #6233, opened 2023-06-19
Addonify Everything


Samq64:
@CST1229 suggested this a little while ago on the dev Discord server and it's a good idea.

![image](https://github.com/ScratchAddons/ScratchAddons/assets/81489795/41b04949-f056-43a0-938e-72f04d5fd91e)

DNin01:
Just to make it clear, this will NOT turn these features into toggleable settings. These will still be behind the scenes but will work like addons.

iqnite:
Like, building core features on the Scratch Addons API? That would make code a lot more readable, so it's definitely worth to consider!

WorldLanguages:
We can just add a new "hidden" tag or something like that

WorldLanguages:
(I mean, don't worry about it at this stage, just create the addons)

DNin01:
> We can just add a new "hidden" tag or something like that

Or section them off into a group of minor addons (#2617)

iqnite:
Maybe add a "Show core addons" toggle in "More settings", which will show all the addons with the "Core" tag, with a note on the addons telling the user that they can't be changed. Those addons would also show in the "Active in this tab" section in the extension popup, making it easier to debug certain problems.

Secret-chest:
The version*s* should probably be optional.

mxmou:
> The version_s_ should probably be optional. Privacy.

Maybe, but it's already possible to remove the S before submitting the post.

DNin01:
> The version<i>s</i> should probably be optional. Privacy.

That could be a good thing to let users have control over. It would probably be fine if changing "version" to "versions" was done by a "core" addon that could be disabled. Additionally, it could be made clear that this is intended to be helpful and help us find problems that could be caused by Scratch Addons.

Related: #5274

PR #7547, opened 2024-06-18
Addonify forum user agent modifier

### LINKED ISSUE #5274: Be more transparent about "no flash versionS detected"
See #1872. We automatically change the user agent when SA users post in the forums and click the globe icon. Scratch adds "no flash version detected" and we change it to "no flash versions detected".
Users might not expect this behavior, and it's not documented anywhere. We don't intentionally hide this behavior, but users might not realize the extra "s" when posting.
We could: (choose one)
1. Tell the user right in the forum post screen, after the globe icon is pressed (how would we explain it? will be confusing to some users)
2. Mention it somewhere in scratchaddons.com/docs, such as inside the extension privacy policy
3. Add a toggle for this somewhere in the settings page, and default it to enabled

### LINKED ISSUE #6065: Forum user-agent is not always modified ("version"→"versions")

Samq64:
Resolves #5274
Resolves #6065
Closes #7529
Progress on #6233

Requires a manifest schema change.

### Changes

Adds a hidden addon tag and moves the forum user agent modifier into a new addon.

### Reason for changes

To move things away from the core extension logic that don't need to be there. It also makes only running on certain pages easier.

### Tests

Tested on Chromium 126. If an addon does not have a valid tag it is now hidden instead of categorized as an editor addon.


DNin01:
Related: #2617

addons/forum-user-agent/userscript.js
```diff
@@ -0,0 +1,9 @@
+export default async function ({ addon, console }) {
+  if (!window.mySettings) return;
+  const ua = window.mySettings.markupSet.find((x) => x.className);
+  ua.openWith = window._simple_http_agent = ua.openWith.replace("version", "versions");
+  if (location.pathname === "/discuss/3/topic/add/") {
```
OneShot-Niko, line 5: #6065 isn't resolved because this still only checks for adding a new topic in BaG, from what I've tested, it doesn't work anywhere else:
kép
(QaS)
```suggestion
  if (location.pathname.match(/^(?:https?:\/\/scratch\.mit\.edu\/)?(topic|\/discuss\/settings)?/))
```

OneShot-Niko (changes requested):

addons/forum-user-agent/userscript.js
Samq64: That's supposed to only replace the user-agent on posts where it's already pre-filled so it doesn't mess with post editing but clicking the button should still work.

OneShot-Niko: > That's supposed to only replace the user-agent on posts where it's already pre-filled so it doesn't mess with post editing but clicking the button should still work.

Odd, for me it didn't add the S... Gonna test again

OneShot-Niko: kép
not sure what happened then

Samq64:
I think it's better to wait for #2617.

WorldLanguages:
> I think it's better to wait for #2617.

Hmm, that may take a while.

How visible do we want this addon to be? I don't think it's a big deal if it shows along other useful addons.

mxmou:
> How visible do we want this addon to be? I don't think it's a big deal if it shows along other useful addons.

This PR currently hides the addon from the settings page completely.

WorldLanguages:
Original issue #5274 considered that we add a toggle:

> Add a toggle for this somewhere in the settings page, and default it to enabled

So why hide this? Let this be the 17th enabled-by-default addon. (I understand that addon-ifying other code requires hidden addons, but not this specific feature!)

WorldLanguages:
Like [DNin01 said](https://github.com/ScratchAddons/ScratchAddons/issues/6233#issuecomment-2179225268), we just need to add an "info" that explains the user why this is useful to us.

Samq64:
Is the user-agent change even that useful now that ScratchDB is down?

WorldLanguages:
> Is the user-agent change even that useful now that ScratchDB is down?

We can't search forum posts by keyword anymore, but it still helps to decide whether a bug is caused by us or not. If many reports lack "versions" then the problem is probably not caused by the extension.

addons/forum-user-agent/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Forum user-agent Modifier",
+  "description": "Changes the text \"Flash version\" to \"Flash versions\" when clicking the user-agent button on the forums. Previously part of the extension itself.",
```
WorldLanguages, line 3:
```suggestion
  "description": "Changes the text \"Flash version\" to \"Flash versions\" when clicking the user-agent button (globe icon) on the forums.",
```

addons/forum-user-agent/userscript.js
```diff
@@ -0,0 +1,9 @@
...
+    const textarea = document.getElementById("id_body");
+    textarea.value = ua.openWith;
```
WorldLanguages, line 7: Add `dynamicEnable:false` to the manifest explicitly so that we avoid data loss here (we don't want a future contributor to think it's a good idea to add dynamicEnable to this addon)

addons/forum-user-agent/addon.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Forum user-agent Modifier",
```
WorldLanguages, line 2:
```suggestion
  "name": "Modify user-agent slightly in forums",
```

addons/forum-user-agent/userscript.js
WorldLanguages: This also may be racy (possibly! maybe some image takes too long to work). Maybe add `runAtComplete:false` and change this to a waitForElement

addons/forum-user-agent/addon.json
WorldLanguages: It's hard to come up with a good name, but I think it makes sense to mention the forums, and it is also better if we emphasize the fact the addon only does a very "subtle" or "small" change.

DNin01:
> Is the user-agent change even that useful now that ScratchDB is down?

Since the ScratchDB shutdown, ocular still works, it just uses a much simpler algorithm, so searching is still at least possible. Plus, new projects will hopefully take ScratchDB's place in the near future, and Googling your search terms with `site:scratch.mit.edu` (which means to only search for content on Scratch) has always been an option too.

mxmou:
> We can't search forum posts by keyword anymore

[We can.](https://ocular.jeffalo.net/search?q=%22no%20flash%20versions%20detected%22&sort=newest) See [jeffalo's announcement](https://scratch.mit.edu/discuss/post/8026911/) about the new search.

addons/forum-user-agent/addon.json
mxmou: I think something like "Modify browser version [on forums](https://scratchaddons.com/docs/reference/addon-manifest/#description:~:text=on%20(not%20in)%20the%20forums)" or "Modify version information on forums" would be better because we don't actually modify the browser's user agent string. The description could say "globe icon" instead of "user agent button".

addons/forum-user-agent/userscript.js
Samq64: Have there been any issues other forum addon because of this?

WorldLanguages: No, but this is enabled by default, so we can be more careful

addons/forum-user-agent/addon.json
WorldLanguages: > we don't actually modify the browser's user agent string

I personally don't think this matters. Most people don't know what the user-agent string is anyway.
Those who know, after reading the description, will understand that the addon only affects the output of the globe icon.

Still, `Modify version information on forums` is okay, the addon name is boring and it's quite clear that it does nothing interesting.

WorldLanguages:
Can we all agree on `Modify version information on forums` as the addon name?

Samq64:
Currently the description doesn't mention user-agent anywhere, should I add it back?

mxmou:
Putting "globe icon" in parentheses is weird - [everyone calls it the globe icon](https://ocular.jeffalo.net/search?q=%22globe%20icon%22&sort=relevance) (or [globe button](https://ocular.jeffalo.net/search?q=%22globe%20button%22&sort=relevance)).

WorldLanguages (approved):
Looks good to me

If we want to support dynamicEnable or prevent potential textarea data loss (for example, if addons are injected quite late into https://scratch.mit.edu/discuss/3/topic/add/ then everything the user wrote into the body of the post would be replaced) we could add a condition such as `document.timeline.currentTime < 8000` or `textarea.value.length < X` to make sure we aren't overriding what the user wrote.

addons/forum-user-agent/addon.json
```diff
@@ -0,0 +1,20 @@
...
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["editingScreens"]
+    }
+  ],
+  "tags": ["community", "forums"],
+  "versionAdded": "1.39.0",
```
mxmou, line 17: This will make the addon appear under "New addons". Should there be a note explaining that the new feature is the ability to disable it?

mxmou (approved):

addons/forum-user-agent/addon.json
```diff
@@ -0,0 +1,24 @@
...
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["editingScreens"]
+    }
+  ],
+  "tags": ["community", "forums"],
+  "versionAdded": "1.40.0",
```
WorldLanguages, line 17:
```suggestion
  "versionAdded": "1.39.99",
```

The settings page doesn't consider addons where versionAdded == latestUpdate.version, so let's lie to it...

We can restore this to `1.40.1` before releasing 1.41.0

WorldLanguages:
```suggestion
  "versionAdded": "1.0.0",
```

This may break less things

WorldLanguages: If you want to check what was the version when we added this functionality to cs.js, go ahead...

mxmou: > If you want to check what was the version when we added this functionality to cs.js, go ahead...

It's 1.13.0.

PR #7550, opened 2024-06-19
Fix positive integer setting type to yse the min and max values

Jazza-231:
### Changes

In addon-settings.html it changes
```html
      <template v-if="setting.type === 'positive_integer'">
        <input
          type="number"
          class="setting-input number"
          v-model="addonSettings[setting.id]"
          @change="checkValidity() || updateSettings()"
          :disabled="!addon._enabled"
          min="0"
          number
        />
      </template>
``` 
to
```html
      <template v-if="setting.type === 'positive_integer'">
        <input
          type="number"
          class="setting-input number"
          v-model="addonSettings[setting.id]"
          @change="checkValidity() || updateSettings()"
          :disabled="!addon._enabled"
          :min="(setting.min && setting.min > 0) ? setting.min : 0"
          :max="setting.max"
          number
        />
      </template>
```

### Reason for changes

The positive integer setting type should respect min and max.
This actually causes the paint-snap addon which uses positive_integer (the only one) to have its min and max values not respected.

### Tests

Works both in the main settings page and the poup settings on brave (chromium)

### DIFF
--- a/webpages/settings/components/addon-setting.html
+++ b/webpages/settings/components/addon-setting.html
@@ -79,7 +79,8 @@
           v-model="addonSettings[setting.id]"
           @change="checkValidity() || updateSettings()"
           :disabled="!addon._enabled"
-          min="0"
+          :min="(setting.min && setting.min > 0) ? setting.min : 0"
+          :max="setting.max"
           number
         />
       </template>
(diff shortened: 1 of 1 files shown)


DNin01 (commented):
This setting type isn't really used anymore and has been replaced by `integer`, which is functionally the same other than it can be below 0 and `integer`s can have a set minimum and maximum. Consider `positive_integer` deprecated.

At least that's what I know.

At that point, it also doesn't really make sense to set minimums for `positive_integer`s because you can already do that with `integer`s, and if you don't want it to be below 0, just explicitly set a minimum of 0.

TheColaber:
so then we need to delete "positive_integer" alltogether and make current settings that use this switch to integer.

Jazza-231:
Oh ok ill do that instead

Samq64:
Related PR: #6791

WorldLanguages:
Jazza said that `paint-snap` is using a positive_integer with a min:4, so please do send a PR to fix that.

Jazza-231:
@WorldLanguages Pr made

PR #7552, opened 2024-06-19
Add new method of loading the extension in README.md

hueychen27:
`web-ext` method with `npm`

### Changes

This adds some more text showing a different method of loading the extension to Firefox temporarily like the other method.

### Reason for changes

This is a good notice for developers as this method automatically updates the extension in the background while editing code.

### Tests
The Markdown looks fine.


README.md
```diff
@@ -77,6 +77,8 @@ To load the extension into most Chromium-based browsers, go to `chrome://extensi
 
 Go to `about:debugging`, select "This Firefox", click "Load Temporary Add-on...", and select the `manifest.json` file in the `ScratchAddons` folder.
 
+Alternatively, install `web-ext` with `sudo npm install --global web-ext` in your terminal and in the `ScratchAddons` folder, run `web-ext run` to automatically load the temporary addon in a new, isolated Firefox window. This method is useful in development as it automatically updates the addon in the background. You still need to refresh the web page for Scratch Addon have your latest changes.
```
Samq64, line 80:
```suggestion
Alternatively, install `web-ext` with `npm install --global web-ext` as administrator from the terminal and in the `ScratchAddons` folder, run `web-ext run` to automatically load the temporary extension in a new, isolated Firefox window. This method is useful in development as it automatically updates the extension in the background. The tab will still need to be refreshed for addon changes to take effect.
```

WorldLanguages:
> This method is useful in development as it automatically updates the addon in the background.

Saying this is quite dangerous...
Keep in mind that for most contributors, reloading the page is often enough, no matter how the extension was loaded.

Samq64:
> Saying this is quite dangerous... Keep in mind that for most contributors, reloading the page is often enough, no matter how the extension was loaded.

It does seem to fix Firefox's caching issues on though. The useful for development part can be removed though.

README.md
```diff
@@ -77,6 +77,8 @@ To load the extension into most Chromium-based browsers, go to `chrome://extensi
 
 Go to `about:debugging`, select "This Firefox", click "Load Temporary Add-on...", and select the `manifest.json` file in the `ScratchAddons` folder.
 
+Alternatively, install `web-ext` with `npm install --global web-ext` as administrator from the terminal and in the `ScratchAddons` folder, run `web-ext run` to automatically load the temporary extension in a new, isolated Firefox window. This method is useful in development as it automatically updates the extension in the background. Because the browser is isolated, no other extensions will appear and Scratch Addons will be in its default state. The tab may still need to be refreshed for extension changes to take effect.
```
Samq64, line 80:
```suggestion
Alternatively, install `web-ext` with `npm install --global web-ext` as administrator from the terminal and in the `ScratchAddons` folder, run `web-ext run` to automatically load the temporary extension in a new, isolated Firefox window. This method automatically updates the extension in the background. Because the browser is isolated, no other extensions will appear and Scratch Addons will be in its default state. The tab may still need to be refreshed for extension changes to take effect.
```

Samq64 (approved):

README.md
```diff
@@ -77,6 +77,8 @@ To load the extension into most Chromium-based browsers, go to `chrome://extensi
 
 Go to `about:debugging`, select "This Firefox", click "Load Temporary Add-on...", and select the `manifest.json` file in the `ScratchAddons` folder.
 
+Alternatively, install `web-ext` with `npm install --global web-ext` as administrator from the terminal and in the `ScratchAddons` folder, run `web-ext run` to automatically load the temporary extension in a new, isolated Firefox window. This method automatically updates the extension in the background. Because the browser is isolated, no other extensions will appear and Scratch Addons will be in its default state. The tab may still need to be refreshed for extension changes to take effect.
```
DNin01, line 80: What does "The tab may still need to be refreshed for extension changes to take effect" mean? Is that about extension pages?

Samq64: The Scratch tabs where addons are running.

Samq64:
Actually the second half of the paragraph seems unnecessary to have in the readme and could just be included in the website's installation docs.

README.md
hueychen27: Let me clarify with my new commit.

hueychen27:
I discovered that `web-ext` supports Google Chrome with `web-ext run --chromium-binary /path/to/chromium`

Samq64:
> I discovered that `web-ext` supports Google Chrome with `web-ext run --chromium-binary /path/to/chromium`

Does it matter though? On Firefox there's another advantage since the about:debugging method overrides the store version but that's not the case Chromium. Also as WorldLanguages said changing addon userscripts/styles don't need an extenson reload.

hueychen27:
> > I discovered that `web-ext` supports Google Chrome with `web-ext run --chromium-binary /path/to/chromium`
> 
> Does it matter though? On Firefox there's another advantage since the about:debugging method overrides the store version but that's not the case on Chromium. Also as WorldLanguages said changing addon userscripts/styles don't need an extenson reload.

I was questioning its usefulness but decided to add it because it is information.

WorldLanguages (changes requested):
I still don't get the motivation.

- In general, due to the way the extension works, autoreload only makes things more confusing, and you still need to reload about the same number of times.
- The Firefox caching issues rarely happen, and I've never experienced them when loading unpacked.
- If we want to "keep the extension settings separate from the store release", we can already do that by editing the manifest.json file in this repository, a new way to load the extension isn't needed.

If you really want to go ahead with recommending this method of loading the extension, I recommend it gets added at the bottom of https://scratchaddons.com/docs/getting-started/installing/ instead of the README.

hueychen27:
> I still don't get the motivation.
> 
>     * In general, due to the way the extension works, autoreload only makes things more confusing, and you still need to reload about the same number of times.
> 
>     * The Firefox caching issues rarely happen, and I've never experienced them when loading unpacked.
> 
>     * If we want to "keep the extension settings separate from the store release", we can already do that by editing the manifest.json file in this repository, a new way to load the extension isn't needed.
> 
> 
> If you really want to go ahead with recommending this method of loading the extension, I recommend it gets added at the bottom of https://scratchaddons.com/docs/getting-started/installing/ instead of the README.

I find it easier to test the extension with one command instead of using `about:debugging`. It is much more convenient in my personal experience even with auto-reload. `web-ext` is also an automated script for `about:debugging`, but with extra features.

PR #7553, opened 2024-06-19
Update backpack.css

hackpro4005:
Resolves #

no issues

### Changes

backpack rows

 
edited backpack.css

### Reason for changes

idk

for a better look

### Tests
no issues


TheColaber:
@hackpro4005 what's your motivation for these changes? I don't think we are planning on changing this number based on the comment above. I can reopen this PR with more context.

DNin01:
This would probably need to be part of a different addon to give users the choice to change the look of the backpack drawer.

You can [create an issue](https://github.com/ScratchAddons/ScratchAddons/issues/new/choose) and start a discussion there, if you would like.

PR #7559, opened 2024-06-20
Rework display none while disabled

TheColaber:
Extends #4924

Changes Tab.displayNoneWhileDisabled to create a style element on the page that has a display: none !important when the addon is disabled.

Hopefully this can get merged before any new addons uses the api...


addons/one-click-studio-invites/userscript.js
```diff
@@ -41,7 +41,7 @@ export default async function (
       button.disabled = true;
     }
 
-    addon.tab.displayNoneWhileDisabled(button, { display: "block" });
+    addon.tab.displayNoneWhileDisabled(button);
```
mxmou, line 44: Same as above, but with `display: block`.

addons/vol-slider/userscript.js
```diff
@@ -55,9 +55,7 @@ export default async function ({ addon, console }) {
   innerContainer.appendChild(icon);
   innerContainer.appendChild(slider);
   container.appendChild(innerContainer);
-  addon.tab.displayNoneWhileDisabled(container, {
-    display: "flex",
-  });
+  addon.tab.displayNoneWhileDisabled(container);
```
mxmou, line 58: Same

mxmou (changes requested):

addons/emoji-picker/userscript.js
```diff
@@ -179,7 +179,7 @@ export default async function ({ addon, console, msg }) {
     emojiButtonText.textContent = "🙂︎";
     emojiButtonText.classList.add("sa-emoji-button");
     emojiButton.appendChild(emojiButtonText);
-    addon.tab.displayNoneWhileDisabled(emojiButton, { display: "inline-block" });
+    addon.tab.displayNoneWhileDisabled(emojiButton);
```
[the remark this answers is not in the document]
TheColaber: i tested this addon, and all the ones you mentioned below, and they seemed to work without adding any extra styles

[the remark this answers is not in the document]
mxmou: They work but the layout is slightly broken - for example, the volume slider isn't centered. Compare them with the stable version.

addons/emoji-picker/userstyle.css
```diff
@@ -1,4 +1,5 @@
 .sa-emoji-picker {
+  display: inline-block;
```
mxmou, line 2: `.sa-emoji-button-container` also needs this style. It currently looks like this in this PR:
image

On master:
image

mxmou (approved):

Samq64 (approved):

TheColaber:
Updated new addon. Good to go.

PR #7561, opened 2024-06-20
Set paint-snap to use integer

Jazza-231:
Resolves https://github.com/ScratchAddons/ScratchAddons/pull/7550#issuecomment-2181066266

### Changes

Makes `paint-snap` threshold setting use integer

### Reason for changes

See resolves

### DIFF
--- a/addons/paint-snap/addon.json
+++ b/addons/paint-snap/addon.json
@@ -97,7 +97,7 @@
       "default": false
     },
     {
-      "type": "positive_integer",
+      "type": "integer",
       "id": "threshold",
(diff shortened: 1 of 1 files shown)


WorldLanguages (approved):
It is okay to change the type of a setting from positive_integer to integer, will not cause problems

Of course it's possible that the previous min/max values were bypassed but that has always been the case

Samq64 (approved):

DNin01 (approved):
This also makes the minimum and maximum actually apply, whereas before you could set the value to any positive integer.

PR #7566, opened 2024-06-21
Fix workspace update bugs

### LINKED ISSUE #2346: Certain addons make "Save now" appear even if no changes were made
**Describe the bug**

If the `data-category-tweaks-v2` addon is enabled, the "Save now" button always shows in the menu bar, even if no changes were made to the project.

**To Reproduce**

1. Enable data category tweaks
2. Open the editor

**Expected behavior**

The "Save now" button would only appear when making changes to the project.

- Browser: Edge
- Version: 90

mxmou:
Resolves #2346

### Changes

* Addons that used `emitWorkspaceUpdate()` to redraw blocks (cat-blocks, custom-block-shape, custom-block-text, editor-square-inputs, editor-theme3, zebra-striping) now use the `updateAllBlocks` module, which has been changed to call `Blockly.Xml.clearWorkspaceAndLoadFromXml()` directly
* Addons that used it to a add custom extension category (data-category-tweaks-v2, debugger) now use a new shared module which dispatches the `UPDATE_TOOLBOX` Redux action

This fixes the following bugs caused by workspace updates:
* "Save now" appearing when enabling/disabling an addon, changing settings, or entering the editor from the project page
* The undo stack being cleared when enabling/disabling an addon or changing settings

folders and gamepad still use `emitWorkspaceUpdate()`, but they do so when the project actually changes, so showing the "Save now" option is appropriate.

### Tests

Tested on Edge and Firefox.


addons/custom-block-shape/update-all-blocks.js
```diff
@@ -1,24 +1,20 @@
...
       workspace.getToolbox().refreshSelection();
-      workspace.toolboxRefreshEnabled_ = true;
     }
+    workspace.toolboxRefreshEnabled_ = true;
   }
 
-  // There's no particular reason for checking whether events were originally enabled.
-  // Unconditionally enabling events at this point could, in theory, cause bugs in the future.
-  if (eventsOriginallyEnabled) blockly.Events.enable(); // Re-enable events
+  blockly.Events.enable();
```
mxmou, line 19: Calling `enable()` unconditionally is correct: if `disable()` was called while events were disabled, `enable()` needs to be called twice to actually reenable them. See https://github.com/scratchfoundation/scratch-blocks/blob/1421093/core/events.js#L282-L303

addons/data-category-tweaks-v2/userscript.js
```diff
@@ -184,19 +186,11 @@ export default async function ({ addon, console, msg, safeMsg }) {
...
-  // If editingTarget is set, the editor has already rendered and we have to tell it to rerender.
-  if (vm.editingTarget) {
-    vm.emitWorkspaceUpdate();
-  }
-
   addon.settings.addEventListener("change", (e) => {
-    // When the separate list category option changes, we need to do a workspace update.
+    // When the separate list category option changes, we need to update the toolbox XML.
     // For all other options, just refresh the toolbox.
-    // Always doing both of these in response to a settings change causes many issues.
```
mxmou, line 195: I removed this comment because I don't know what the "many issues" could be other than the ones this PR fixes.

TheColaber (commented):
What is the motivation for having all of the addons import from custom-block-shape? Why not have a Tab API or maybe just have each addon update whatever it needs to update in it's code?

mxmou:
> What is the motivation for having all of the addons import from custom-block-shape? Why not have a Tab API or maybe just have each addon update whatever it needs to update in it's code?

Two addons already imported custom-block-shape's module: custom-block-text and editor-square-inputs. I agree that it should probably be moved somewhere else - maybe to the addon API, but it could also be in `libraries`.

I don't think each addon should have its own code for this. Having many slightly different versions of the same code can easily cause bugs like the one fixed by #5542 to happen.

addons/custom-block-shape/update-all-blocks.js
WorldLanguages: Thanks. I think I wrote that code comment. As I am not fully familiar with the scratch-blocks codebase, I didn't want to risk it.

libraries/common/cs/update-toolbox-xml.js
```diff
@@ -0,0 +1,14 @@
+export default function updateToolboxXML(tab) {
+  tab.waitForElement("[class*='gui_blocks-wrapper_']").then((blocksWrapper) => {
```
WorldLanguages, line 2: We can add `!state.scratchGui.mode.isPlayerOnly` as a small optimization here

WorldLanguages:
> Two addons already imported custom-block-shape's module: custom-block-text and editor-square-inputs. I agree that it should probably be moved somewhere else - maybe to the addon API, but it could also be in `libraries`.

You may move it to libraries if you want. I don't think it really fits into the addon API.

libraries/common/cs/update-toolbox-xml.js
mxmou: Actually, this could be a simple `querySelector` - if the toolbox doesn't exist, there's no reason to update it.

mxmou:
> You may move it to libraries if you want. I don't think it really fits into the addon API.

I already moved it to libraries - see the latest commit.

libraries/common/cs/update-all-blocks.js
```diff
@@ -0,0 +1,34 @@
+export function updateAllBlocks(
+  blockly,
+  workspace,
```
TheColaber, line 3: We should make this api get the workspace by itself instead of having it passed in because it should always return the workspace from `Tab.Traps.getWorkspace()` since otherwise, problems such as https://github.com/ScratchAddons/ScratchAddons/pull/7587 may occur. The same might as well be done for blockly as well?

mxmou: I'll change it to only take `tab` as an argument.

WorldLanguages:
> I'll change it to only take `tab` as an argument.

Good idea. We kept adding arguments and the calls were getting too long.

WorldLanguages:
@mxmou I suggest you take a quick second look at all the code changes in this PR (it's been some time, so it's effectively a second review), if you tell me that you've done this I will merge this as admin

mxmou:
> @mxmou I suggest you take a quick second look at all the code changes in this PR (it's been some time, so it's effectively a second review), if you tell me that you've done this I will merge this as admin

Done. I also tested the affected addons.

PR #7567, opened 2024-06-21
cat-blocks: fix "watch mouse cursor" setting

### LINKED ISSUE #2391: Cat Blocks: Watch mouse pointer setting does not work for newly loaded extensions
**Describe the bug**

When you load an extension into scratch, the hat blocks on the flyout do not watch you. Yet when you drag them out, the blocks in the workspace do.

**To Reproduce**

1. Go to Editor
2. Click on image
3. Load Makey makey for example
4. See how it's hat blocks in the flyout aren't watching you

**Expected behavior**

**Screenshots**

I could repro on both chrome and firefox, on both scratch and tw.

mxmou:
Resolves #2391

### Changes

Makes the "watch mouse cursor" setting in the cat blocks addon work more consistently.

### Tests

Tested on Edge.

### DIFF
--- a/addons/cat-blocks/userscript.js
+++ b/addons/cat-blocks/userscript.js
@@ -182,7 +182,7 @@ export default async function ({ addon, console }) {
       // Set to the correct initial position
       this.svgFace_.style.transform = "translate(-87px, 0px)";
     }
-    if (this.shouldWatchMouse()) {
+    if (shouldWatchMouseCursor) {
       this.windowListener = function (event) {
         var time = Date.now();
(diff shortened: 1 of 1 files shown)


TheColaber (approved):
Currently turning on and off the setting does nothing, but I think that is old behavior. We [who?] are planning on merging this PR today.

mxmou:
> Currently turning on and off the setting does nothing, but I think that is old behavior.

Changing the setting requires a reload. This PR doesn't change that.

WorldLanguages (approved):

PR #7569, opened 2024-06-21
find-bar: fix Ctrl+Z

### LINKED ISSUE #3534: find-bar: pressing Ctrl+Z when input is not focused anymore, refocuses it
Open a project, type something in the devtools box, put some blocks in the workspace, and press ctrl+z. Scratch undoes, and devtools input is also "undone" by the browser. Apparently only happens in Chrome, which sometimes lets ctrl+z undo after the input was blurred.

A terrible bug fix is:
```js
    document.addEventListener("keydown", (e) => {
      if ((e.metaKey || e.ctrlKey) && (e.key === "z" || e.key === "y")) {
        if (document.activeElement === document.body) e.preventDefault();
      }
    });
```

mxmou:
Resolves #3534

### Changes

Adds something very similar to the "terrible bug fix" suggested by W_L in the linked issue.

### Tests

Tested on Edge.

### DIFF
--- a/addons/find-bar/userscript.js
+++ b/addons/find-bar/userscript.js
@@ -176,6 +176,21 @@ export default async function ({ addon, msg, console }) {
           return true;
         }
       }
+
+      // In Chrome, Ctrl+Z will undo edits to the find bar input even if it doesn't have focus.
+      // Call preventDefault() to make sure that the event only goes to scratch-blocks or scratch-paint.
+      // Blockly.onKeyDown_:
+      // https://github.com/scratchfoundation/scratch-blocks/blob/1421093/core/blockly.js#L185
+      // KeyboardShortcutsHOC.handleKeyPress:
+      // https://github.com/scratchfoundation/scratch-paint/blob/8119055/src/hocs/keyboard-shortcuts-hoc.jsx#L29
+      if (!Blockly.utils.isTargetInput(e) && addon.tab.redux.state?.scratchPaint.textEditTarget === null) {
+        if (
+          (ctrlKey || e.altKey) &&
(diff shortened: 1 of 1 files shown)


TheColaber (commented):
This seems to fix all inputs in the scratch editor. In that case, this should be a new addon, and find-bar should just make sure that this bug doesn't happen to itself. Also, find bar already puts an event listener on keydown, so it might be better to reuse that one.

Samq64:
> This seems to fix all inputs in the scratch editor. In that case, this should be a new addon, and find-bar should just make sure that this bug doesn't happen to itself.

Why? The find bar is already enabled by default.

mxmou:
> This seems to fix all inputs in the scratch editor. In that case, this should be a new addon, and find-bar should just make sure that this bug doesn't happen to itself. Also, find bar already puts an event listener on keydown, so it might be better to reuse that one.

I don't think there's a way to only fix this for specific inputs. I'll move the code to the existing event listener though.

WorldLanguages:
> Adds something very similar to the "terrible bug fix" suggested by W_L in the linked issue.

Well, it looks better than mine.

> This seems to fix all inputs in the scratch editor. In that case, this should be a new addon

I don't think we should be so strict. I don't remember anyone suggesting we fix this for inputs other than the find bar.

WorldLanguages (approved):
Briefly tested it in Chromium. I don't think this is terrible, so it can be merged.

Maybe test it in Firefox before merging, just to be sure.

Samq64 (approved):
On Chromium the search box's undo history is kept when re-focusing it but not on Firefox.

PR #7572, opened 2024-06-21
Rewrite zebra striping addon

TheColaber:
Changes this addon to use `Blockly.BlockSvg.setColour` instead of CSS. This allows addons such as colored context menus and save block image to be compatible with the addon.


TheColaber:
@mxmou im using my own lighten function, and since editor-theme3 sets an "a" value for the block colors, my lighten function tries to lighten the "a" value. That causes scratch to error. What do you think is the best way to approach this?

mxmou:
> @mxmou im using my own lighten function, and since editor-theme3 sets an "a" value for the block colors, my lighten function tries to lighten the "a" value. That causes scratch to error. What do you think is the best way to approach this?

If it's possible to use the `text-color` module or tinycolor, that's probably the best solution.

TheColaber:
I was wondering if I should use `ScratchBlocks.Colours` to get the original color of the block if it needs to get un-zebra-striped. Currently, I'm storing the old value of the color in `ScratchBlocks.Block`. I am not already using `ScratchBlocks.Colours` because some category names such as `events` don't align with the category stored in the block (events/event). Even though this addon should never zebra stripe an event block, I just wasn't sure what to do.

mxmou:
> I was wondering if I should use `ScratchBlocks.Colours` to get the original color of the block if it needs to get un-zebra-striped. Currently, I'm storing the old value of the color in `ScratchBlocks.Block`. I am not already using `ScratchBlocks.Colours` because some category names such as `events` don't align with the category stored in the block (events/event). Even though this addon should never zebra stripe an event block, I just wasn't sure what to do.

Debugger blocks and TurboWarp custom extensions don't use colors from `ScratchBlocks.Colours`, so I think storing the original color is better.

addons/zebra-striping/userscript.js
```diff
@@ -1,70 +1,59 @@
...
-          }
-          for (const field of input.fieldRow) {
-            if (field.fieldGroup_) {
-              elements.push(field.fieldGroup_);
-            }
-          }
-        }
-        for (const el of elements) {
-          el.classList.toggle("sa-zebra-stripe", isStriped);
+        block.striped =
```
mxmou, line 19: Could this property and `originalColour_` have a `sa` prefix?

mxmou (changes requested):
Bugs:
* Stack blocks inside C-blocks (for example "wait" inside "forever") aren't striped. This might be an intentional change but I don't think it makes sense - they're striped in both Snap<em>!</em> and the current version of this addon.
* theme3's Black preset is broken. The secondary, tertiary and quaternary colors should be based on the original ones instead of letting `setColour` generate them from the primary color.

addons/zebra-striping/userscript.js
```diff
@@ -1,70 +1,70 @@
...
+          const secondary = block.getColourSecondary();
+          const tertiary = block.getColourTertiary();
+          const quaternary = block.getColourQuaternary();
+          block.setColour(
+            "#" + tinycolor(color).lighten(amount).toHex(),
+            "#" + tinycolor(secondary).lighten(amount).toHex(),
+            "#" + tinycolor(tertiary).lighten(amount).toHex(),
+            "#" + tinycolor(quaternary).lighten(amount).toHex()
+          );
+          block.setColour("#" + tinycolor(color).lighten(amount).toHex());
```
Samq64, line 51:
```suggestion
```

Samq64 (commented):
With the default settings the colours are slightly different and dropdowns are effcected.
image
image

TheColaber:
i have fixed all issues. @Samq64 Please rereview

TheColaber:
note that is was possible to only create one new property on Blockly.Block instead of two, but I used two for code understandability and readability. its also possible to have two/one "global" variables to achieve the same, like it did before these changes, but this is up to preference.

addons/zebra-striping/userscript.js
```diff
@@ -1,69 +1,85 @@
...
+    ) {
+      const shade = addon.settings.get("shade");
+      const intensity = addon.settings.get("intensity");
+      const amount = ((shade === "lighter" ? 1 : -1) * intensity) / 2;
       // Conveniently getDescendants() returns blocks in an order such that each block's
       // parent will always come before that block (except the first block which has no
       // parent).
       for (const block of this.getDescendants()) {
         const parent = block.getSurroundParent();
+        let wasStripped = !!block.sa_striped;
```
mxmou, line 27: Typo
```suggestion
        let wasStriped = !!block.sa_striped;
```

addons/zebra-striping/userscript.js
```diff
@@ -1,69 +1,85 @@
...
+        // Ensure the shadow block has the correct stroke by updating it's color to match parent colors.
+        if (block.isShadow() && !parent.sa_striped) {
+          block.updateColour();
+          continue;
+        }
```
mxmou, line 32: If the shadow is a dropdown, not a text/number input, its primary, secondary and quaternary color also need to be updated to match the parent. I think the simplest way to implement that is to remove this `continue` and treat shadow blocks the same way as normal block, except that they're striped if they have the same color as the parent and the parent is also striped (normal blocks are striped if the parent is not striped).

mxmou:
> With the default settings the colours are slightly different

I think the new colors are better - they're more similar to the original category colors, but the difference in brightness is still noticeable enough.

addons/zebra-striping/userscript.js
```diff
@@ -1,69 +1,82 @@
...
+        // field strokes don't automatically update when setColour is called.
+        // so manually update field dropdowns to match parent color
         for (const input of block.inputList) {
-          if (input.outlinePath) {
-            elements.push(input.outlinePath);
-          }
           for (const field of input.fieldRow) {
-            if (field.fieldGroup_) {
-              elements.push(field.fieldGroup_);
+            if (field instanceof ScratchBlocks.FieldDropdown) {
```
mxmou, line 66:
```suggestion
            if (field instanceof ScratchBlocks.FieldDropdown && field.box_) {
```

addons/zebra-striping/userscript.js
```diff
@@ -1,69 +1,82 @@
...
+          // not a insertion marker
+          !block.isInsertionMarker() &&
+          // has a parent
+          parent &&
+          // shadow blocks should be striped if it's parent is striped. otherwise, it should be opposite of parent.
+          (block.isShadow() ? parent.saStriped : !parent.saStriped) &&
+          // parent and child are same color (even shadows match block color)
+          // we dont check category because other addons can make blocks the same color.
+          // if the block was stripped we need to look at it's original color in order to get it's real color.
+          parent.getColour() === (block.saOrginalColour ? block.saOrginalColour[0] : block.getColour());
```
mxmou, line 43: This still doesn't work for shadow blocks - I thought shadow dropdowns would have the same color as the parent block, but that doesn't seem to be the case. Comparing the secondary color if `block` is a shadow works though, but the parent's `saOriginalColor` needs to be checked too. The correct result looks like this:
image

The current code makes it look like this instead:
image

addons/zebra-striping/userscript.js
```diff
@@ -1,69 +1,82 @@
...
+
+        if (!shouldStripe && block.saOrginalColour) {
+          block.setColour(...block.saOrginalColour);
+          block.saOrginalColour = null;
+        } else if (shouldStripe && !block.saOrginalColour) {
+          const color = block.getColour();
+          const secondary = block.getColourSecondary();
+          const tertiary = block.getColourTertiary();
+          const quaternary = block.getColourQuaternary();
+          block.saOrginalColour = [color, secondary, tertiary, quaternary];
```
mxmou, line 56:
```suggestion
          block.saOriginalColour = [color, secondary, tertiary, quaternary];
```

mxmou (changes requested):

addons/zebra-striping/userscript.js
TheColaber: why check .box_ (so i can put it in the comments)

addons/zebra-striping/userscript.js
TheColaber: @mxmou maybe i can have shadow blocks check if there parent is the same, but all other regular blocks would check primary color. that might make the most sense.

addons/zebra-striping/userscript.js
mxmou: There's no box if the dropdown is in a shadow: https://github.com/scratchfoundation/scratch-blocks/blob/develop/core%2Ffield_dropdown.js#L139

addons/zebra-striping/userscript.js
mxmou: > maybe i can have shadow blocks check if there parent is the same

If you meant "is the same category", that wouldn't work because the category of many shadow blocks is unfortunately null. Maybe you could check if the shadow block contains a dropdown field?

mxmou (approved):

WorldLanguages:
I believe Google/Scratch are working on a new `scratch-blocks` that would use Blockly as a dependency, instead of being a direct fork. See: https://github.com/gonfunko/scratch-blocks/issues

It's hard to know when that will come into play, but it makes me wonder if it's worth it to use blockly APIs in addons that previously worked fine with CSS.

TheColaber:
it is very likely many css addon will have to change as well anyway. either way, for compatibility reasons, it is better for addons to modify blockly.

GrahamSH-LLK:
@TheColaber There's very little point in working on this. Blockly is killing setColour (https://github.com/google/blockly/issues/8280) and recommending using CSS instead.

PR #7579, opened 2024-06-23
Change `delete-others` addon name + description

TheColaber:
Based off:

https://github.com/ScratchAddons/ScratchAddons/blob/d0daf6be4c3018695036655fbf6fab11b14d362c/addons/move-to-top-bottom/addon.json#L2-L3

and the feedback left from a transifex reviewer: https://github.com/ScratchAddons/ScratchAddons/pull/7030/files#r1639627849

### DIFF
--- a/addons/delete-others/addon.json
+++ b/addons/delete-others/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Delete other assets",
-  "description": "Adds an option to the right click menu to easily delete all other costumes, backdrops, or sounds.",
+  "description": "Adds an option to the right click context menu for costumes and sounds to delete all items expect for the selected one.",
   "tags": ["editor", "costumeEditor", "recommended"],
   "versionAdded": "1.39.0",
   "userscripts": [
--- a/addons/delete-others/addon.json
+++ b/addons/delete-others/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Delete other assets",
+  "name": "Delete other costumes or sounds",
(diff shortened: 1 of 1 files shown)


addons/delete-others/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Delete other assets",
```
WorldLanguages, line 2: We could also rename it to `Delete other costumes`, as that's what most users will use it for.

OneShot-Niko: It can still be used for costumes though, so it wouldn't really make sense.

mxmou: "Delete other costumes or sounds" would be better for searchability. The addon appears when searching for "delete other costumes" or "delete other sounds" on master, but not in this PR because it removes the word "other" from the description.

OneShot-Niko (changes requested):

mxmou (approved):

TheColaber:
@mxmou I changed the name to your suggestion

WorldLanguages:
> It can still be used for sounds though, so it wouldn't really make sense.

@OneShot-Niko Well, `Sprite folders` is not named `Sprite, backdrop, costume and sounds folders`.

WorldLanguages (approved):

OneShot-Niko:
> @OneShot-Niko Well, `Sprite folders` is not named `Sprite, backdrop, costume and sounds folders`.

Fair

PR #7582, opened 2024-06-24
`editor-dark-mode`: Scratch Lab theme

### LINKED ISSUE #7581: Scratch Lab theme for "Editor dark mode and customizable colors"
### Why this would be helpful

Some people would want their Scratch experience to look like Scratch Labs'.

### My suggestion

Add a Scratch Labs theme preset.

### Possible alternatives

_No response_

### Additional context

_No response_

nolongerusinggithub:
Resolves #7581

### Changes

Adds a singular new theme.
Not sure if this is needed since a user can always just change to the blue preset and then change header to #0fbd8c, but convenience I guess

### Reason for changes

Some people want a scratchlab preset (according to #7581)

### Tests

Tested on Firefox/Gecko 127.0.1 (32 bit) - haven't tested on chrome/chromium
ScratchLab only seems to have a green navbar (with blue colors) and nothing else changes color-wise besides that. Please correct me if I'm wrong though, from what I can tell thats literally all that's different on scratchlab.

### DIFF
--- a/addons/editor-dark-mode/addon.json
+++ b/addons/editor-dark-mode/addon.json
@@ -1373,6 +1373,31 @@
         "stageHeader": "#e8edf1",
         "border": "#00000026"
       }
+    },
+          {
+      "name": "ScratchLab",
+      "id": "scratch-lab",
+      "description": "Colors used by the ScratchLab editor.",
+      "values": {
+        "page": "#e5f0ff",
+        "primary": "#4d97ff",
+        "highlightText": "#4d97ff",
+        "menuBar": "#0fbd8c",
+        "popup": "#4d97ffe6",
+        "activeTab": "#ffffff",
+        "tab": "#d9e3f2",
+        "selector": "#e9f1fc",
+        "selector2": "#d9e3f2",
+        "selectorSelection": "#ffffff",
+        "accent": "#ffffff",
+        "input": "#ffffff",
+        "workspace": "#f9f9f9",
+        "categoryMenu": "#ffffff",
+        "palette": "#f9f9f9cc",
+        "fullscreen": "#ffffff",
+        "stageHeader": "#e8edf1",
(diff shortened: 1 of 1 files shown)


addons/editor-dark-mode/addon.json
```diff
@@ -1373,6 +1373,31 @@
         "stageHeader": "#e8edf1",
         "border": "#00000026"
       }
+    },
+          {
+      "name": "ScratchLab",
+      "id": "scratch-lab",
+      "description": "The colors used by the ScratchLab's project editor.",
```
DNin01, line 1380: It's officially called "Scratch Lab" with a space:
```suggestion
      "name": "Scratch Lab",
      "id": "scratch-lab",
      "description": "The colors used by the Scratch Lab's project editor.",
```

OneShot-Niko: Don't know how I keep on forgetting the space...

medlans: You could be thinking about ScratchX which didn't have a space (I used to mess up Scratch Labs' name before realizing because of that)

Samq64:
Editor dark mode already has 10 presets. Do we need another?

OneShot-Niko:
> Editor dark mode already has 10 presets. Do we need another?

I think that 11 isn't too much, but that might vary from person to person.
I should've probably discussed that before making this PR fdb99343d31dafece2dc4698a31fe555

GrahamSH-LLK (dismissed):
format with prettier

OneShot-Niko:
> format with prettier

its a copypaste of `scratch-blue` with the navbar changed to scratchlab's green (since that is all that seems to change in scratchlab) so it's formatted the same irregardless of prettier

DNin01:
Fixed. One of the brackets was over-indented.

PR #7584, opened 2024-06-24
scratch-messaging: show fragment if fetching comment failed

### LINKED ISSUE #2466: Messaging tab - comments on extremely old threads don't load correctly
**Describe the bug**

When a comment referenced in a user's messages is from a long time ago, the comment fails to appear in the Messaging tab.

**To Reproduce**

1. Using an alt account or other willing test subject, navigate to an extremely old comment thread (a year should suffice, or perhaps less time on active/popular projects, studios, or profiles) and reply to one of the messages sent by your main account
2. Go to Scratch Addon's Messaging tab and confirm that the message doesn't show (an empty header should be there instead)
3. Navigate to your main account's Messages page (on scratch.mit.edu) and confirm that the message exists
4. Open the provided link to the comment in the Messages page and confirm that the message is showing up

**Expected behavior**

If it's a timeout issue, it would probably take too long to actually load the message, especially for really ancient threads, so perhaps a notice like [message failed to load] or [message not found] should appear.

**Screenshots**

This is what it looked like when the messages were unread:
image

After messages were marked as read and the "Load more messages" button was used: (note the "3" count, then count how many comments there actually are)
image
The comment in question (https://scratch.mit.edu/projects/275999360/#comments-206329971) is about a year old, and there has been a significant amount of comments posted since then.

- OS: Windows 10
- Browser: Chromium 90.0.4430.93
- Version: 1.14.2

### LINKED ISSUE #3916: Scratch messaging only shows the project title on unshared projects
### Description

If someone comments on a project and then I unshare it, Scratch messaging just shows the project title without the comment.

### Reproduction steps

1.  Comment on a project made by your alt. (if you have a second account and a project you don't mind unsharing)
2. Log into your alt and unshare the project.
3. Open Scratch messaging.

### Expected behaviour

The comment should either appear in the Scratch messaging tab or not appear at all.

### Actual behaviour

Only the project title appears.

### Screenshots

image

### Environment

- Browser: Firefox 95
- Scratch Addons version: 1,22,1
- Operating system: Windows 11

mxmou:
Resolves #2466
Resolves #3916

### Changes

If a comment can't be loaded, shows the fragment returned by the messages API instead, without context (note that the reply appears as a top-level comment in the screenshot below).

image

### Tests

Tested on Edge and Firefox. I only tested comments on an unshared project, but this should fix old threads too because the root cause of the bug is the same.

Trying to reply to a comment on an unshared project shows an error message.

### DIFF
--- a/popups/scratch-messaging/api.js
+++ b/popups/scratch-messaging/api.js
@@ -131,10 +131,30 @@ export async function sendLegacyComment(addon, { resourceType, resourceId, conte
     });
 }
 
-export async function fetchComments(addon, { resourceType, resourceId, commentIds, page = 1, commentsObj = {} }) {
-  if (resourceType === "user")
-    return fetchLegacyComments(addon, { resourceType, resourceId, commentIds, page, commentsObj });
-  return fetchMigratedComments(addon, { resourceType, resourceId, commentIds, page, commentsObj });
+export async function fetchComments(addon, { resourceType, resourceId, commentMessages, page = 1, commentsObj = {} }) {
+  const commentIds = commentMessages.map((message) => message.comment_id);
+  let comments;
+  if (resourceType === "user") {
+    comments = await fetchLegacyComments(addon, { resourceType, resourceId, commentIds, page, commentsObj });
+  } else {
+    comments = await fetchMigratedComments(addon, { resourceType, resourceId, commentIds, page, commentsObj });
+  }
+  for (let message of commentMessages) {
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Thanks mxmou

Sorry if the code is messy. It is a few years old by now and I wrote most of it. (To be fair, old code in the original extension by griffpatch wasn't much better, but anyway)

PR #7586, opened 2024-06-24
`mediarecorder` options UI changes

### LINKED ISSUE #7460: Revamp `mediarecorder`'s "Start Recording" modal
### Why this would be helpful

The current UI is just a _little_ unnatural and unintuitive, and there is room to improve the user experience by enhancing the way the options are presented.

### My suggestion

Here's what I came up with:
Options for starting a recording with multiple-choice settings to choose how long until the recording is started or stopped and checkboxes to record project or microphone audio

### Possible alternatives

_No response_

### Additional context

_No response_

DNin01:
Resolves #7460

### Changes

Makes numerous small changes to the recording options modal of the `mediarecorder` addon:
- Reorganized the options modal and edited various strings.
- Made the modal 100px thinner.
- `recordOptionStopInput` is now automatically unchecked when `recordOptionFlagInput` is unchecked.
- Fixed a bug where `recordOptionStopLabel`'s `title` attribute (which says 'You need to enable "Wait until green flag is clicked" to use this option.') wasn't reset when the dependent checkbox was checked.
- Fixed a bug where `recordOptionStopInput` wasn't disabled if `recordOptionFlagInput` was initially disabled (not before possible until #7173).
- Made the checkboxes purple (or the highlight color you have set).

The diff looks convoluted, but I made two separate commits, one of which moves blocks of code, so diff them one after another and it should be easier to tell what changed.

### Reason for changes

To improve the UX.

### Tests

Tested in Edge 126.


DNin01:
Before vs. after screenshots:
Modal with options for starting a recording, before VS after

addons-l10n/en/mediarecorder.json
```diff
@@ -2,19 +2,20 @@
...
-  "mediarecorder/record-description": "Record the stage as a {extension} file. You can save it to your computer after the recording is finished.\nNote: variable and list monitors will not be visible.",
+  "mediarecorder/record-after-flag": "Wait until green flag is clicked",
+  "mediarecorder/record-until-stop": "End recording when project is stopped",
+  "mediarecorder/record-description": "Record a video of the stage and save it to your computer as a {extension} file after the recording is finished.\nNote: variable and list monitors will not be visible in the recording.",
   "mediarecorder/start": "Start",
   "mediarecorder/cancel": "Cancel",
   "mediarecorder/added-by": "Added by Scratch Addons browser extension",
   "mediarecorder/click-flag": "Waiting...",
-  "mediarecorder/click-flag-description": "Click the green flag to start recording. Click this button to stop.",
+  "mediarecorder/click-flag-description": "Click the green flag to start recording. Click here to abort the recording.",
```
mxmou, line 19: You changed "stop" to "end" in the popup and "abort" in this tooltip. The same word should be used everywhere - I prefer "stop", but "end" is also an option.

addons/mediarecorder/style.css
```diff
@@ -34,11 +38,19 @@ p.mediaRecorderPopupOption {
 
 .mediaRecorderPopupOption input[type="checkbox"] {
   height: 1.5rem;
+  width: 15px;
```
mxmou, line 41: I think it's better to keep the browser's default checkbox size for consistency with other popups that contain checkboxes (Make a Block, gamepad settings).

addons-l10n/en/mediarecorder.json
DNin01: Correct me if I'm wrong, but I think this tooltip appears when you hover over the "Waiting..." button when the recorder is waiting for the green flag to be clicked before recording. If you click "Waiting...", it cancels the recording. That's why I used the term "abort".

DNin01: About the change of "stop" to "end", would you like if I changed "End recording after:" to "Stop recording after:" and/or do the same with "End recording when project is stopped"?

I used the term "end" in that second one just to not say "stop" twice, but it's fine either way.

mxmou (approved):

WorldLanguages:
I will test this shortly.

addons-l10n/en/mediarecorder.json
```diff
@@ -2,19 +2,20 @@
...
-  "mediarecorder/record-mic": "Include sounds from microphone",
+  "mediarecorder/record-audio": "Record project audio",
+  "mediarecorder/record-mic": "Record microphone audio",
   "mediarecorder/record-audio-description": "This does not include Text-to-Speech.",
-  "mediarecorder/record-after-flag": "Do not start recording until the Green Flag is clicked",
-  "mediarecorder/record-until-stop": "Stop recording after the project has stopped",
-  "mediarecorder/record-description": "Record the stage as a {extension} file. You can save it to your computer after the recording is finished.\nNote: variable and list monitors will not be visible.",
+  "mediarecorder/record-after-flag": "Wait until green flag is clicked",
+  "mediarecorder/record-until-stop": "End recording when project is stopped",
+  "mediarecorder/record-description": "Record a video of the stage and save it to your computer as a {extension} file after the recording is finished.\nNote: variable and list monitors will not be visible in the recording.",
```
WorldLanguages, line 14: Maybe it's a good opportunity to 1. split this into two strings, and 2. make it easier to read?

WorldLanguages: A possible alternative to the "monitors are not visible" note, is instead adding it as a checkbox that can't be enabled, and that clearly signals that the feature is not supported yet (for now... see other issues)

DNin01: I'll experiment with putting the note in a separate, highlighted sentence, kind of like this:

> Record a video of the stage and save it to your computer as a _{extension}_ file after the recording is finished.

> [!NOTE]
> Variable, list, and reporter monitors, as well as ask prompts will not be visible in the recording.

The note could also be put more concisely:

> [!NOTE]
> Some elements are not visible in recordings. <a href='#'>Learn more</a>

Which one do you prefer?

mxmou: > > [!NOTE]
> > Variable, list, and reporter monitors, as well as ask prompts will not be visible in the recording.

I like this version.

WorldLanguages: Yes, the first one.

DNin01:
Here's how it looks now:
Modal with options for starting a recording

mxmou (approved):

WorldLanguages:
I just have something I forgot to bring up.

image

If the project runs for 40 seconds and then stops, what will the user expect was recorded? 30 seconds, or 40 seconds?

Here are some ways to make it more clear I can think of.
- The default should be increased to something more realistic, such as 120 seconds.
- Instead of rewording "Record Duration" to "End recording after X seconds" (proposed), we can think of another way to word it. Such as "Record up to X seconds" or similar.
- If you disagree with me and don't think this could be confused, we can still increase the default to something bigger than 30 seconds,

DNin01:
Labeling it "Record for:" might be helpful. Additionally, we could change it so that the recording can either end after a set amount of time or after the project stops running (or after the 10-minute limit), but not both.

https://github.com/ScratchAddons/ScratchAddons/assets/106490990/06aa48ec-23da-494b-80eb-a29e006bf375

mxmou:
> Additionally, we could change it so that the recording can either end after a set amount of time or after the project stops running (or after the 10-minute limit), but not both.

Radio buttons would make more sense for that:
image

WorldLanguages (approved):
Ok, we can merge this then. We can use radio buttons with the existing strings, so we won't unnecessarily change them once again.

DNin01:
> Ok, we can merge this then. We can use radio buttons with the existing strings, so we won't unnecessarily change them once again.

You mean merge this PR's [current changes](https://github.com/ScratchAddons/ScratchAddons/pull/7586#issuecomment-2198277720) and then open another PR to add something like radio buttons? (I haven't committed any changes to the recording length settings yet.)

DNin01:
@WorldLanguages?

WorldLanguages:
> You mean merge this PR's https://github.com/ScratchAddons/ScratchAddons/pull/7586#issuecomment-2198277720 and then open another PR to add something like radio buttons?

Yes I think that's fine.
You may keep this PR open if future plans require changing the same strings once again (we don't want to waste translators' time). I don't think that's the case, so I would merge this if you're fine with that.

DNin01:
I just noticed something, @WorldLanguages.

In this PR, "start (or) end recording after:" and "seconds" are two separate strings, and we just insert the input field between them by concatenating. Could this be a problem for other languages where there are no words either before or after the input? What would a translator do in that case? Would they know what was going on?

So instead of using two strings, should we use a dollar sign, which would act as the input field's point of insertion?

WorldLanguages:
I don't think it's worth the effort, the UI is easy to understand anyway.

PR #7616, opened 2024-07-03
remove exact count

GrahamSH-LLK:
### Changes

Removes exact count addon
### Reason for changes

It causes undue load on Scratch, and has been asked to be removed.


GrahamSH-LLK:
keeping user as it doesn't cause undue load

Issue #6729, opened 2023-10-01
Throttle fetch() calls to the Scratch API to mitigate bugs


WorldLanguages:
### Why this would be helpful

In the past we've accidentally sent more requests to the Scratch API than necessary (for example see #5991, which was only fine because of Cache-Control and disk cache) and edge cases such as #6614 where an infinite loop could potentially arise.

### My suggestion

1. Add an additional check to all uses of `fetch()` with these considerations:
- It's very unlikely an addon will need to send more than a few requests per second, throw in those cases
- Userscripts don't really have a reason to call the Scratch API if the tab is in the background, unless the page is loading or has recently finished loading.
- Add some checks to background code. Chrome already has some DDoS protection, but I don't think Firefox does.

2. Revisit addons such as `exact-count` that need lots of requests to work.

3. Check other types of requests we might be doing, such as XHR, loading images, etc.

### Additional context

Related: Chromium sometimes throttles in extension background pages to make sure they don't accidentally DDoS: https://www.chromium.org/throttling/

WorldLanguages:
It would also mitigate infinite-scrolling bugs, for example see #2980

WorldLanguages:
Related: PR #7505 intends to make our background context (service worker / event page) sleep for longer if not using Scratch, which should mean we send less requests to Scratch servers overall, while also saving users CPU, memory and battery.

WorldLanguages:
Okay, #7505 was merged, so when v1.39.0 is released hopefully we'll be sending less requests to the Scratch servers. It's hard to estimate how many requests we're sending if we add all of our users, but it could be adding up to a lot.

WorldLanguages:
> 2\. Revisit addons such as `exact-count` that need lots of requests to work.

See merged PR #7616

DNin01:
We could also maybe add a rate limit to Infinite scrolling. If you mash the End key, you can load very fast, much faster than the servers would appreciate. Would waiting a second between page loads, even silently, help a little bit?

WorldLanguages:
> We could also maybe add a rate limit to Infinite scrolling. If you mash the End key, you can load very fast, much faster than the servers would appreciate. Would waiting a second between page loads, even silently, help a little bit?

Well, we already only send 1 request at a time (unless there's a bug). But yes, we could add some small delay, by simulating the Scratch server taking longer to respond, but what we would actually be doing is sending the request a few ms later.

WorldLanguages:
[...]

PR #7617, opened 2024-07-03
remove harmful and dead exact-count features

GrahamSH-LLK:
### Changes

Removes studio and forum exact count features
### Reason for changes

Studio causes large amounts of load on Scratch. Forum is dead, as scratchdb is dead.


apple502j (approved):

apple502j:
`TYPE:SCREAMING REQUEST FROM DEVELOPERS`

WorldLanguages:
Resolves part of #7617

WorldLanguages:
> Studio causes large amounts of load on Scratch.

Fetching an additional scratchr2 page on each profile sounds more intensive. But I just checked and we do not send cookies with that request, so we allow Scratch to return a cached version of the page.

GrahamSH-LLK:
> Resolves part of #7617

uhh that's this PR lol

mxmou:
> Resolves part of #7617

#6729?

DNin01:
Might also want to say something about these features being removed. If we don't, users could think it's a bug.

mxmou:
> Wait, why is exact count for studios being removed?

From the PR description:

> Studio causes large amounts of load on Scratch.

DNin01:
> > Wait, why is exact count for studios being removed?
> 
> 
> 
> From the PR description:
> 
> 
> 
> > Studio causes large amounts of load on Scratch.

Oh, it was slowing the servers down?

If W_L agrees on that, that makes sense.

Perhaps there's a more efficient way to implement it…

mxmou:
> Maybe if there's a more efficient way to implement it we could re-add it in the future.

The current version of the [exactcount website](https://mxmou.github.io/exactcount/) makes fewer requests, but I think the main issue is sending additional requests on each page load. Using the more efficient algorithm and only shown the exact count when the user clicks the "100+" would be much better than what the current version of the addon did.

Samq64:
Studios should probably have been a separate PR since it wasn't broken.

mxmou:
> Fetching an additional scratchr2 page on each profile sounds more intensive.

It's actually six additional scratchr2 pages: shared projects, favorited projects, curated studios, followed studios, followed users, followers.

WorldLanguages:
> Might also want to say something about these features being removed. If we don't, users could think it's a bug.

Sorry, I was going to say that we would readd this to require a click, but I missed it due to exams. We will merge #7621 before releasing a new version.

WorldLanguages:
About profiles, I created #7623 and let's continue the discussion there.

PR #7621, opened 2024-07-06
exact-count: better exact count for studios

mxmou:
### Changes

Adds the "exact count for studios" option (recently removed in #7617) back, but instead of loading the count (which sends a large number of requests to Scratch) every time a studio page is opened, the user has to click the "100+" first:
image
image
image

Also replaces the `countProjects` function with the updated one from the exactcount website, which is much more readable - no callbacks, only loops with `await`. It doesn't actually reduce the average number of requests per studio though - I tested it on various studios.

Comparison with the previous version of the addon: https://github.com/ScratchAddons/ScratchAddons/compare/27ed288c2d823284616aabddc53e6ea380f88b7e...mxmou:ScratchAddons:studio-exact-count

### Reason for changes

This feature was quite useful because vanilla Scratch doesn't provide a way to see the exact number of projects in a studio. The main issue with the previous implementation is solved in this PR.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/exact-count/addon.json
+++ b/addons/exact-count/addon.json
@@ -38,7 +38,7 @@
     },
     {
       "url": "studio.js",
-      "matches": ["https://scratch.mit.edu/studios/*"],
+      "matches": ["studios"],
       "if": {
         "settings": { "studio": true }
       }
@@ -47,7 +47,7 @@
   "userstyles": [
     {
       "url": "studio.css",
-      "matches": ["https://scratch.mit.edu/studios/*"],
+      "matches": ["studios"],
       "if": {
         "settings": { "studio": true }
       }
--- a/addons/exact-count/addon.json
+++ b/addons/exact-count/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Show exact count",
-  "description": "Shows the exact count for user information.",
+  "description": "Shows exact numbers on profiles and the exact project count in studios.",
   "credits": [
     {
       "name": "TheColaber",
@@ -20,6 +20,12 @@
       "id": "user",
       "type": "boolean",
       "default": true
+    },
+    {
+      "name": "Exact count for studios",
+      "id": "studio",
+      "type": "boolean",
+      "default": true
     }
   ],
   "userscripts": [
@@ -28,6 +34,22 @@
       "matches": ["https://scratch.mit.edu/users/*/"],
       "if": {
         "settings": { "user": true }
+      }
+    },
+    {
+      "url": "studio.js",
+      "matches": ["https://scratch.mit.edu/studios/*"],
+      "if": {
+        "settings": { "studio": true }
+      }
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "studio.css",
+      "matches": ["https://scratch.mit.edu/studios/*"],
+      "if": {
+        "settings": { "studio": true }
       }
     }
   ],
--- a/addons-l10n/en/exact-count.json
+++ b/addons-l10n/en/exact-count.json
@@ -1,0 +1,3 @@
+{
+  "exact-count/studio-tooltip": "Click to show exact count"
+}
--- a/addons/exact-count/studio.css
+++ b/addons/exact-count/studio.css
@@ -1,0 +1,43 @@
+@import url("../../libraries/common/cs/spinner.css");
+
+.studio-tab-nav li .tab-count {
+  position: relative;
+}
+
+.studio-tab-nav li .tab-count .sa-exact-count-tooltip {
+  display: none;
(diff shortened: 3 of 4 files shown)


DNin01 (approved):

DNin01:
I like the loading on click. Maybe we should do that with the profile counts too (in a different PR)?

mxmou:
> I like the loading on click. Maybe we should do that with the profile counts too (in a different PR)?

The profile counts wouldn't be very useful if you had to click to load them - it's already possible to click "View all" to see the count in vanilla Scratch.

addons/exact-count/addon.json
```diff
@@ -20,6 +20,12 @@
       "id": "user",
       "type": "boolean",
       "default": true
+    },
+    {
+      "name": "Exact count for studios",
+      "id": "studio",
```
WorldLanguages, line 26: Yes, this is the same setting ID that was removed in #7617, so it's like we never removed the setting 👍

WorldLanguages: The last value of the "forumuser" setting will not be stored anywhere, so if we ever readd that feature somehow, users will need to enable it again.

addons/exact-count/addon.json
```diff
@@ -29,6 +35,22 @@
       "if": {
         "settings": { "user": true }
       }
+    },
+    {
+      "url": "studio.js",
+      "matches": ["https://scratch.mit.edu/studios/*"],
```
WorldLanguages, line 41: Just to be consistent with other addons:

```suggestion
      "matches": ["studios"],
```

addons/exact-count/addon.json
```diff
@@ -29,6 +35,22 @@
...
+      "if": {
+        "settings": { "studio": true }
+      }
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "studio.css",
+      "matches": ["https://scratch.mit.edu/studios/*"],
```
WorldLanguages, line 50:
```suggestion
      "matches": ["studios"],
```

WorldLanguages:
How should it behave if the user intends to switch to the Projects tab while in another tab, and coincidentally hovers the 100+ count? Should we switch the tab and also calculate the exact count? Or only switch the tab and expect the user to click once again to see the exact count? The current behavior is to load the exact count and not switch the tab.

image

WorldLanguages:
> Comparison with the previous version of the addon: [27ed288...mxmou:ScratchAddons:studio-exact-count](https://github.com/ScratchAddons/ScratchAddons/compare/27ed288c2d823284616aabddc53e6ea380f88b7e...mxmou:ScratchAddons:studio-exact-count)

Thanks, very useful.

> It doesn't actually reduce the average number of requests per studio though - I tested it on various studios.

We can't do better than a binary search, of course.

Here are 2 suggestions, though:

- Latency to the Scratch servers may vary user by user, so it would be nice to limit the amount of requests per second. For example, we could add a delay of 200ms between the response and the next request.
- Ideally we should test all edge cases, including project counts which are multiples of 40 +/- 1. It's also possible that some offsets are cached and some aren't, so it's also possible that the Scratch API returns seemingly nonsensical data, and we don't want to cause an infinite loop of requests. Maybe instead of considering all these cases we can instead limit the amount of requests to a reasonable number, for example no more than 30 requests.

mxmou:
> How should it behave if the user intends to switch to the Projects tab while in another tab, and coincidentally hovers the 100+ count? Should we switch the tab and also calculate the exact count? Or only switch the tab and expect the user to click once again to see the exact count? The current behavior is to load the exact count and not switch the tab.

I'll change it to do both. If the user wanted to switch the tab, not switching would be confusing. If they just wanted to see the count, they can switch back to the previous tab easily.

WorldLanguages (approved):

WorldLanguages:
This can happen if the viewport is less than 940px wide. Low-prio.

image

DNin01:
> This can happen if the viewport is less than 940px wide. Low-prio.

What?!

Scratch probably changes how the description's validation error message shows up on smaller screens. [`scratchfoundation/scratch-www/src/components/forms/validation-message.scss#L61-L73`](https://github.com/scratchfoundation/scratch-www/blob/c1281cedfd8b8cfc5394bbad2fa9993453f24817/src/components/forms/validation-message.scss#L61-L73)

mxmou:
> Scratch probably changes how the description's validation error message shows up on smaller screens.

Indeed - they're on the right side of the input on large screens and below the input on narrow ones.
image
image

mxmou:
> This can happen if the viewport is less than 940px wide. Low-prio.

Fixed.

PR #7630, opened 2024-07-08
Update clipboardWrite permission logic for Firefox 127+

### LINKED ISSUE #7625: Firefox 127+ removes the need for `addon.tab.copyImage` and `"clipboardWrite"` permission
This is the current implementation of `addon.tab.copyImage()`, which is used by `bitmap-copy` and a `blocks2image` PR:

```js
  copyImage(dataURL) {
    if (!dataURL.startsWith(DATA_PNG)) return Promise.reject(new TypeError("Expected PNG data URL"));
    if (typeof Clipboard.prototype.write === "function") {
      // Chrome /* Edit: AND NOW FIREFOX 127+ */
      const blob = dataURLToBlob(dataURL);
      const items = [
        new ClipboardItem({
          "image/png": blob,
        }),
      ];
      return navigator.clipboard.write(items);
    } else {
      // Firefox needs Content Script
      // This message goes to the background event page where the `browser.clipboard.setImageData` API is used.
      return scratchAddons.methods.copyImage(dataURL).catch((err) => {
        return Promise.reject(new Error(`Error inside clipboard handler: ${err}`));
      });
    }
  }
```

See MDN: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#browser_compatibility

image

WorldLanguages:
Resolves #7625

This affects addons that request the "clipboardWrite" addon permission in order to copy PNG images into the clipboard. The extension then requests the "clipboardWrite" **extension permission** if necessary, after the user tries to enable an addon that requests it. This extension permission is deprecated in Chrome, but not in Firefox.

The only addon that uses this permission is currently `bitmap-copy`, and soon `blocks2image`.

### Changes

Firefox 127+ will behave exactly like Chrome always had, in respect to this permission, in order to avoid unnecessary permission prompts.
- A permission prompt will not appear when enabling an addon that requires this permission.
- Existing addons that get this permission added with an update, won't get disabled automatically if the permission is missing.

Firefox versions 126 and below (less than 15% of users) are not affected by these changes.

### Reason for changes

`Clipboard.prototype.write` was implemented in Firefox 127.
See MDN: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#browser_compatibility

### Tests

Both Chrome and Firefox only allow the clipboard to be modified if there's a "transient activation", such as a gesture, key press, or click. The behavior may not be identical, but we always copy in response to a user gesture, so this is not a problem.

### DIFF
--- a/background/imports/util.js
+++ b/background/imports/util.js
@@ -2,7 +2,7 @@
 const browserLevelPermissions = ["notifications"];
 if (typeof browser !== "undefined") {
   // Firefox only
-  if (typeof Clipboard.prototype.write === "function") {
+  if (typeof Clipboard.prototype.write !== "function") {
     // Firefox 109-126 only
     browserLevelPermissions.push("clipboardWrite");
   }
--- a/background/imports/util.js
+++ b/background/imports/util.js
@@ -1,5 +1,12 @@
+// REMINDER: update similar code at /webpages/settings/index.js
 const browserLevelPermissions = ["notifications"];
-if (typeof browser !== "undefined") browserLevelPermissions.push("clipboardWrite");
+if (typeof browser !== "undefined") {
+  // Firefox only
+  if (typeof Clipboard.prototype.write === "function") {
+    // Firefox 109-126 only
+    browserLevelPermissions.push("clipboardWrite");
+  }
+}
 
 export const getMissingOptionalPermissions = () => {
   return new Promise((resolve) => {
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -56,8 +56,15 @@ let fuse;
     },
   });
 
+  // REMINDER: update similar code at /background/imports/util.js
   const browserLevelPermissions = ["notifications"];
-  if (typeof browser !== "undefined") browserLevelPermissions.push("clipboardWrite");
+  if (typeof browser !== "undefined") {
+    // Firefox only
+    if (typeof Clipboard.prototype.write === "function") {
+      // Firefox 109-126 only
+      browserLevelPermissions.push("clipboardWrite");
+    }
+  }
   let grantedOptionalPermissions = [];
   const updateGrantedPermissions = () =>
     chrome.permissions.getAll(({ permissions }) => {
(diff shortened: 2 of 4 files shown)


background/imports/util.js
```diff
@@ -1,5 +1,12 @@
+// REMINDER: update similar code at /webpages/settings/index.js
```
Samq64, line 1: Is there a reason for the duplicate code?

WorldLanguages: Probably not, but I'm not going to fix it yet. I'm just adding a comment for now.

background/imports/util.js
```diff
@@ -1,5 +1,12 @@
...
 const browserLevelPermissions = ["notifications"];
-if (typeof browser !== "undefined") browserLevelPermissions.push("clipboardWrite");
+if (typeof browser !== "undefined") {
+  // Firefox only
+  if (typeof Clipboard.prototype.write === "function") {
```
Samq64, line 5:
```suggestion
  if (typeof Clipboard.prototype.write !== "function") {
```

webpages/settings/index.js
```diff
@@ -56,8 +56,15 @@ let fuse;
     },
   });
 
+  // REMINDER: update similar code at /background/imports/util.js
   const browserLevelPermissions = ["notifications"];
-  if (typeof browser !== "undefined") browserLevelPermissions.push("clipboardWrite");
+  if (typeof browser !== "undefined") {
+    // Firefox only
+    if (typeof Clipboard.prototype.write === "function") {
```
Samq64, line 63:
```suggestion
    if (typeof Clipboard.prototype.write !== "function") {
```

Samq64:
Also is the Firefox check necessary? Checking if the function exists everywhere shouldn't hurt.

Samq64:
May I push to this branch?

WorldLanguages:
Sure

Samq64:
I tried to remove the browser condition but `Clipboard` was undefined on Chrome.

Samq64 (approved):
Importing settings and enabling work as expected on Firefox 109, 128 and Chromium. I won't test the addon right now since Scratch is down but I did earlier today after applying the patch and that was fine too.

WorldLanguages:
> I tried to remove the browser condition but `Clipboard` was undefined on Chrome.

What? How is this possible?

WorldLanguages:
Oh, I see, `globalThis.Clipboard` is not exposed to service workers.

PR #7631, opened 2024-07-08
`find-bar`: Don't prevent Ctrl+F outside editor

### LINKED ISSUE #7627: `find-bar`: do not capture Ctrl+F when outside editor
### Why this would be helpful

On some occasions, I want to be able to use my browser's find tool instead of Scratch Addon's find bar because I want to find a piece of text. Also, outside of the editor, I cannot use the browser's find tool at all.

### My suggestion

Outside of the editor, do not capture Ctrl+F.

### Possible alternatives

Click the address bar and then press Ctrl+F.

### Additional context

_No response_

DNin01:
Resolves #7627

### Changes

Do not prevent the default behavior of pressing <kbd>Ctrl</kbd>+<kbd>F</kbd> when not in the editor.

### Reason for changes

The find bar can only be used in the editor, and <kbd>Ctrl</kbd>+<kbd>F</kbd> usually opens the browser's find bar.

### Tests

Tested in Edge 126 on Windows.

### DIFF
--- a/addons/find-bar/userscript.js
+++ b/addons/find-bar/userscript.js
@@ -136,7 +136,7 @@ export default async function ({ addon, msg, console }) {
     }
 
     eventKeyDown(e) {
-      if (addon.self.disabled || !this.findBarOuter) return;
+      if (addon.self.disabled || !this.findBarOuter || addon.tab.editorMode !== "editor") return;
 
       let ctrlKey = e.ctrlKey || e.metaKey;
 


hueychen27:
What about using browser's find tool in the editor?

mxmou (approved):

mxmou:
> What about using browser's find tool in the editor?

You can open a separate issue for that.

hueychen27:
> > What about using browser's find tool in the editor?
> 
> You can open a separate issue for that.

Alright, here is issue #7635

PR #6614, opened 2023-09-01
chrome.cookies.onChanged: avoid infinite loops

WorldLanguages:
Resolves (mitigates?) #6460

### Changes

After 5 received events "in a row" (during 1500ms), `chrome.cookies.onChanged` events are added to a queue and only one is processed every 1500ms.

### Tests

To emulate #6460 (requesting `/session/` itself setting cookies unexpectedly), add this to the top of handle-fetch.js:

```js
const extraInfoSpec = ["blocking", "responseHeaders"];
if (Object.prototype.hasOwnProperty.call(chrome.webRequest.OnBeforeSendHeadersOptions, "EXTRA_HEADERS"))
  extraInfoSpec.push("extraHeaders");

chrome.webRequest.onHeadersReceived.addListener(
  function (details) {
    details.responseHeaders.push({
      name: "Set-Cookie",
      value: "scratchcsrftoken=abc; Domain=.scratch.mit.edu; expires=Fri, 30-Aug-2024 12:00:00 GMT; Max-Age=31449600; Path=/; secure",
    });
    return {
      responseHeaders: details.responseHeaders,
    };
  },
  {
    urls: ["https://scratch.mit.edu/session/"],
    types: ["xmlhttprequest"],
  },
  extraInfoSpec
);
```

PR #7646, opened 2024-07-15
Add Scratch Addons identification to /csrf_token route

jeffalo:
At Scratch we're seeing some extremely high request rates to these endpoints (orders of magnitude higher than other endpoints), we would like to identify which requests are coming from Scratch Addons.

This change adds a `?sa=true` parameters to requests to these endpoints. It has no functional impact on Scratch Addons' use, it will help us diagnose the issue.

If possible, if we could get these changes in as soon as possible, that would be great!!

### DIFF
--- a/background/handle-auth.js
+++ b/background/handle-auth.js
@@ -9,7 +9,7 @@ const promisify =
 
 function getDefaultStoreId() {
   // Request Scratch to set the CSRF token.
-  return fetch("https://scratch.mit.edu/csrf_token/", {
+  return fetch("https://scratch.mit.edu/csrf_token/?sa=true", {
     credentials: "include",
   })
     .catch(() => {})
--- a/webpages/popup-loader.js
+++ b/webpages/popup-loader.js
@@ -53,7 +53,7 @@ async function getActualCookieStore() {
 async function refetchCookies(needsRequest = true) {
   if (needsRequest) {
     try {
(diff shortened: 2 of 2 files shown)


WorldLanguages:
The requests sent by the background page of the extension to scratch.mit.edu endpoints (/session, /csrf_token, etc) can be identified with some degree of confidence by checking the value of the `Sec-Fetch-Site` header. It is set to `none` (instead of `same-origin`) by the browser, even though host permissions were granted by the user.

It is unlikely there is other software used by hundreds that requests /csrf_token so I think we should assume it is Scratch Addons.

WorldLanguages:
Related: #6729

WorldLanguages:
To clarify, due to #6614 we can affirm that Scratch Addons is sending no more than `1 request for each 1.5 seconds / user`. But apparently, when all users are online, it can really add up.

WorldLanguages:
Superseded by https://github.com/ScratchAddons/ScratchAddons/pull/7647.

These requests are no longer sent, so there's nothing we can add these URL params to.
Also, checking the coexistence of the `Cookie` header and `Sec-Fetch-Site:none` should be enough to identify that a browser extension is sending the requests from its background process. Further work to identify and throttle requests sent by the extension in all contexts, including the popup and Scratch tabs, can be discussed in issue #6729.

PR #7647, opened 2024-07-15
[v1.38.x] Stop fetching scratch.mit.edu/csrf_token and more optimizations

WorldLanguages:
Resolves #7646

### Changes

- [x] Uses an appropriate API to get the ID of the default cookie store. (See PR #2976 which added support for multiple cookie stores)
- [x] Check whether the popup needs to fetch /csrf_token at all
- [x] Possibly enforce stronger rate limits. Currently 1500ms on handle-auth.js
- [x] Would like to cherrypick #7505 into this branch as well.
- [x] Cherrypick parts of #7621

### Reason for changes

See PR description: #7646

### DIFF
--- a/background/handle-auth.js
+++ b/background/handle-auth.js
@@ -7,21 +7,20 @@ const promisify =
   (...args) =>
     new Promise((resolve) => callbackFn(...args, resolve));
 
-function getDefaultStoreId() {
-  // Request Scratch to set the CSRF token.
-  return fetch("https://scratch.mit.edu/csrf_token/", {
-    credentials: "include",
-  })
-    .catch(() => {})
-    .then(() =>
-      promisify(chrome.cookies.get)({
-        url: "https://scratch.mit.edu/",
-        name: "scratchcsrftoken",
-      })
-    )
-    .then((cookie) => {
-      return (scratchAddons.cookieStoreId = cookie.storeId);
-    });
+async function getDefaultStoreId() {
+  const CHROME_DEFAULT = "0";
+  const FIFEFOX_DEFAULT = "firefox-default";
+  const cookieStores = await chrome.cookies.getAllCookieStores();
+  if (cookieStores.length === 0) throw "";
+  if (cookieStores.some((store) => store.id === CHROME_DEFAULT)) {
+    // Chrome
+    return CHROME_DEFAULT;
+  }
+  if (cookieStores.some((store) => store.id === FIFEFOX_DEFAULT)) {
+    // Firefox
+    return FIFEFOX_DEFAULT;
+  }
+  return cookieStores[0].id;
 }
 
 (async function () {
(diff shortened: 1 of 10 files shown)


WorldLanguages:
Relevant commits:

### /csrf_token/ fetched by the background page:
Commit https://github.com/apple502j/ScratchAddons/commit/a35c0a860615a1aec353bd8831cc2c414c8eeb4c

>   // Extensions are executed on the default store.
  // To get the default store ID reliably, we need to set cookies ourselves.
  // We can't just grab it from existing cookies because that'll break in case
  // someone deletes cookies.
  // The temporary cookie expires after 10 seconds but is removed anyway.

@apple502j Was `browser.cookies.getAllCookieStores[0]` not a thing back then? It appears to get the default store ID reliably.

### /csrf_token/ fetched by the popup:
Commit https://github.com/apple502j/ScratchAddons/commit/ee2aa38b26e5fe43559f9adec7d2b0912597f93f

I believe the PR intended to add incognito support to Scratch Messaging. In Firefox, you can check the messages of the incognito logged-in account through the popup.

WorldLanguages:
Relevant commits:

### /csrf_token/ fetched by the background page:
Commit https://github.com/apple502j/ScratchAddons/commit/a35c0a860615a1aec353bd8831cc2c414c8eeb4c

>   // Extensions are executed on the default store.
  // To get the default store ID reliably, we need to set cookies ourselves.
  // We can't just grab it from existing cookies because that'll break in case
  // someone deletes cookies.
  // The temporary cookie expires after 10 seconds but is removed anyway.

@apple502j Was `browser.cookies.getAllCookieStores[0]` not a thing back then? It appears to get the default store ID reliably.

### /csrf_token/ fetched by the popup:
Commit https://github.com/apple502j/ScratchAddons/commit/ee2aa38b26e5fe43559f9adec7d2b0912597f93f

I believe the PR intended to add incognito support to Scratch Messaging. In Firefox, you can check the messages of the incognito logged-in account through the popup.

background/handle-auth.js
```diff
@@ -7,21 +7,20 @@ const promisify =
...
-      return (scratchAddons.cookieStoreId = cookie.storeId);
-    });
+async function getDefaultStoreId() {
+  const CHROME_DEFAULT = "0";
+  const FIFEFOX_DEFAULT = "firefox-default";
+  const cookieStores = await chrome.cookies.getAllCookieStores();
+  if (cookieStores.length === 0) throw "";
+  if (cookieStores.some((store) => store.id === CHROME_DEFAULT)) {
+    // Chrome
+    return CHROME_DEFAULT;
```
WorldLanguages, line 17: Whoops, I forgot to set `scratchAddons.cookieStoreId`.

WorldLanguages: This function is very badly named :P

PR #7650, opened 2024-07-15
Commits from `v1.38.x` branch (multiple opts to send less requests)

WorldLanguages:
\-


WorldLanguages:
Need to add 2 more commits before merging!

WorldLanguages:
Should also add #7666 to this, when merged.

WorldLanguages:
@Samq64 @DNin01 If anyone can double-check I cherry-picked all the correct commits (and no more) from the v1.38.x branch I would be thankful. If not I can take a look in a few days, don't worry.

Issue #7648, opened 2024-07-15
About releases ~v1.38.3


WorldLanguages:
4 hours ago, jeffalo said: (source: #7646)

> At Scratch we're seeing some extremely high request rates to these endpoints [scratch.mit.edu/csrf_token/] (orders of magnitude higher than other endpoints), we would like to identify which requests are coming from Scratch Addons.

Because of PR #6614 which added rate-limiting to this part of Scratch Addons, we know that each online user is not sending more than 0.7 requests of this type per second (this is a way more strict limit than the Chrome built-in anti-DDoS protection). However, and let me speculate, these numbers can add up. If 100,000 Scratch Addons users are online, overall 4M requests are possibly being sent each minute.

I merged PR #7647 to mitigate the issue, which makes the following changes:
- The affected endpoint (/csrf_token) is not accessed anymore as I couldn't quickly figure out its purpose.
- The rate-limiting has been changed to be more strict. It should be about 0.4 requests per second now.
- Other merged PRs that also intend to reduce the amount of requests were cherry-picked.

Version v1.38.3 was already sent to the Chrome Web Store for review and should reach to users shortly.

An additional note: it might be a Mandela effect, but as far as we can remember, Scratch used to ping this endpoint on each page load. This may have changed in the last year or so without us noticing.

WorldLanguages:
Related: https://github.com/ScratchAddons/ScratchAddons/issues/6729

WorldLanguages:
After v1.38.3 is reviewed, I will release v1.38.4 to fix a small code change mistake I missed: https://github.com/ScratchAddons/ScratchAddons/pull/7647/files#r1678271650

Users experiencing issues because of this mistake (specially after logging in or out) can disable and reenable the extension to reload it. Or they can also wait a few minutes, as Manifest V3 will periodically reload the background process.

hueychen27:
So I was not the only one noticing that log out issue.

WorldLanguages:
I'm working on more optimizations here: PR #7666

WorldLanguages:
See #7651 for discussion.

DNin01:
I'm not sure if we do this already, but we don't have to check for messages when "do not disturb" is on.

WorldLanguages:
> I'm not sure if we do this already, but we don't have to check for messages when "do not disturb" is on.

True, but it probably doesn't make a difference in the grand scheme of things.

WorldLanguages:
[...]

Issue #7623, opened 2024-07-06
`exact-count`: rethink exact counts for profiles


WorldLanguages:
Related: PR #7617
Related: issue #6729

For each visited profile, we fetch the following pages (unauthenticated): shared projects, favorited projects, curated studios, followed studios, followed users, followers.

I bet most users are not really interested on most of those counts. Specially favorites and studios, that's 3/6 requests.
Maybe we can change those to be requested on hover, or even remove them. Users are already one click away from the View All button.

Other things we can consider:
- If the user is visiting a profile comment (URL has #comment-{id} on it) it's likely they won't even look at any of the rows.
- We can delay the requests for the bottom rows (following/followers) and only request those after the user scrolls down.
- When viewing our own profile, we might still be interested in all 6 numbers.
- When viewing our own profile, we might prefer to bypass caching by sending the request with authentication cookies, to get up to date counts, specially for following/followers.

WorldLanguages:
There are 4 thumbs up reactions, so that's some discussion happening I guess.

Also: we should never fetch /users/x/projects if the amount of shared projects is less than 100 (in that case, Scratch already shows the exact number). This is an obvious optimization we missed.

WorldLanguages:
> Also: we should never fetch /users/x/projects if the amount of shared projects is less than 100 (in that case, Scratch already shows the exact number). This is an obvious optimization we missed.

See PR #7666

The optimization above will be part of v1.38.5

I will also remove exact counts for favorited projects, curated studios and followed studios. We can readd them later if there's user feedback about it, or if someone is interested.

WorldLanguages:
v1.38.5 ships with the following changes:

- No more exact counts for favorited projects, curated studios, followed studios
- No exact count for shared projects if the displayed amount is less than 100 (count is already exact as-is)

If we receive anything related to this in the feedback form (specially the first part) , I'll make sure to post it here.

WorldLanguages:
No feedback for now

WorldLanguages:
Further steps for reducing requests, specially in the background context of the extension, is being discussed here: #7651

WorldLanguages:
There was no feedback at all after the v1.38.5 change

WorldLanguages:
Nevermind, there is. I'm having some trouble keeping up with user feedback submissions, sorry.

If anyone wants to help reading through feedback please join Discord or tell me.

WorldLanguages:
Feedback:

[...]

PR #7666, opened 2024-07-20
[v1.38.x] Send less requests to `scratch.mit.edu/session` etc.

WorldLanguages:
Resolves comment https://github.com/ScratchAddons/ScratchAddons/pull/7651#issuecomment-2239953482
Related: #7648

When viewing the diff, I recommend you enable "hide whitespace".

### Changes

This PR intends to make very small changes/optimizations to reduce the load on Scratch servers. It should also slightly help with battery and data usage.

These are small code changes. Big restructurings are intentionally not made.
The changes are made in such a way that anyone can clearly see that, after the changes, it's impossible that the extension sends more than requests than before.

#### Background process (service worker / event page)

The amount of requests to `/session` will decrease significantly. This is strongly related to the Manifest V3 lifecycle and using `chrome.storage.session` to cache data.

- Previously, `/session` would be fetched each time the service worker wakes up.
- Now, the response to `/session` is cached through `chrome.storage.session` and the amount of background requests to that endpoint should be similar to our old Manifest V2 versions.
- The function that previously caused an infinite loop (`addToQueue`) remains intentionally unchanged. It may still request `/session` if Scratch cookies changed.

#### Non-addon code running in Scratch tabs (content scripts in both worlds)

The amount of requests to `/session` will decrease significantly.

- In `scratch-www` pages, such as the front page, projects, and studios, we don't need to fetch `/session` ourselves on page load, as Scratch will do this automatically. We can obtain the response data either by polluting XHR or by looking at Redux. (In this PR I fetch XHR as it's the easiest and more reliable option).
- If no addon calls `addon.auth.fetchX` (X can be "Username", "IsLoggedIn", "UserId" or "XToken") there is no need to fetch `/session`. In fact, in non-www pages (that is, "scratchr2" pages) no requests to `/session` will be sent at all, until one of the running userscripts calls one of these methods.
- In non-www pages (that is, "scratchr2" pages) there is no need to fetch `/session` more than once. Only in some www pages the user can log in or out without a reload. This may mean data gets out of sync if the user logs in to a different account, but that's what already happens on vanilla Scratch, and what tricks like "follow yourself" rely on.

#### Addons

- The `forum-post-countdown` addon reuses the session data fetched by the `isLoggedIn()` call, so overall 1 request is sent instead of two.
- `exact-count`: Do not fetch exact project count if it does not exceed 100. Do not fetch exact favorites and studio counts. (May revert this last decision if there's user feedback about it, see issue #7623)

#### Extension popup

No changes

### Tests

Have to test

### DIFF
--- a/background/handle-auth.js
+++ b/background/handle-auth.js
@@ -126,7 +126,6 @@ async function checkSession(firstTime = false) {
   isChecking = true;
   const { scratchSession } = (await chrome.storage.session?.get("scratchSession")) ?? {};
   if (firstTime && scratchSession) {
-    // We didn't wake up due to a cookie change
     console.log("Used cached /session info.");
     json = scratchSession;
   } else {
--- a/background/handle-auth.js
+++ b/background/handle-auth.js
@@ -26,7 +26,7 @@ async function getDefaultStoreId() {
 (async function () {
   const defaultStoreId = await getDefaultStoreId();
   console.log("Default cookie store ID: ", defaultStoreId);
-  await checkSession();
+  await checkSession(true);
   startCache(defaultStoreId);
 })();
 
@@ -123,13 +123,13 @@ async function setLanguage() {
 
 let isChecking = false;
 
-async function checkSession() {
+async function checkSession(firstTime = false) {
   let res;
   let json;
   if (isChecking) return;
   isChecking = true;
   const { scratchSession } = (await chrome.storage.session?.get("scratchSession")) ?? {};
-  if (scratchSession && canUseCachedSession) {
+  if (firstTime && scratchSession && canUseCachedSession) {
     // We didn't wake up due to a cookie change
     console.log("Used cached /session info.");
     json = scratchSession;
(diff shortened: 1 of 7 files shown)


background/handle-auth.js
```diff
@@ -96,7 +96,11 @@ const addToQueue = (item) => {
     queue.shift();
   }
 };
-chrome.cookies.onChanged.addListener((e) => addToQueue(e));
+let canUseCachedSession = true;
+chrome.cookies.onChanged.addListener((e) => {
+  canUseCachedSession = false;
+  addToQueue(e);
+});
```
WorldLanguages, line 103: I still need to think if this is a good idea.

The service worker waking up due to this event is probably quite rare. The code (written for Manifest V2) may not contemplate this.

WorldLanguages: I will *probably* remove this `canUseCachedSession` variable and only check for `firstTime`. The goal of this PR is only to add optimizations to existing functions, not to change how things work.

Samq64 (approved):
I don't have a new Scratcher to test the 120 second rule.

WorldLanguages:
Regression: #7744

PR #7667, opened 2024-07-20
Changes to `editor-theme3/addon.json`

WorldLanguages:
### Changes

- Remove credits to unreachable developer.
- Add info note clarifying what an "insertion marker" is.

### DIFF
--- a/addons/editor-theme3/addon.json
+++ b/addons/editor-theme3/addon.json
@@ -29,6 +29,11 @@
       "id": "aboutHighContrast",
       "type": "info",
       "text": "The new High Contrast block colors are available as a preset below."
+    },
+    {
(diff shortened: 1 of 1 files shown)


WorldLanguages:
cc @sfederici

Samq64:
I wonder if anyone can think of an easier to understand term so the note isn't necessary?

WorldLanguages:
It can be moved to a "setting description" later.

Samq64 (approved):

TheColaber (approved):

mxmou:
> I wonder if anyone can think of an easier to understand term so the note isn't necessary?

I don't think there's a commonly used name for this feature (other than "insertion marker", which is used by the developers).

PR #7686, opened 2024-07-28
`msg-count-badge`: disable by default & some existing users

WorldLanguages:
Resolves comment https://github.com/ScratchAddons/ScratchAddons/pull/7651#issuecomment-2254267589

### Changes

- The addon `msg-count-badge` is no longer enabled by default.
- When updating to v1.39.0, existing users who do not have Scratch Addons pinned to their extension toolbar will have this addon automatically disabled.
- Users can easily reenable the addon as it will appear at the top of the settings page if disabled.

### Reason for changes

- Removes a historical default due to our Scratch Messaging roots that makes little sense now.
- Less data and battery usage wasted updating the extension badge when it is not visible to the user.
- Sending less requests to Scratch servers.

### Tests

Tested in Chromium

### DIFF
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -174,7 +174,7 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
 
     if (addonSettings["editor-theme3333"]?._version === 3) {
       // Detected update to v1.39 (code below will set the version to 4)
-      if (addonsEnabled["msg-count-badge"]) {
+      if (addonsEnabled["msg-count-badge"] && chrome.action.getUserSettings) {
         chrome.action
           .getUserSettings()
           .then((userSettings) => {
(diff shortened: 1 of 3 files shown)


background/get-addon-settings.js
```diff
@@ -170,6 +172,21 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
       addonSettings["workspace-dots"]["theme"] = "none";
     }
 
+    if (addonSettings["editor-theme3333"]?._version === 3) {
```
WorldLanguages, line 175:
```suggestion
    if (addonSettings["editor-theme3"]?._version === 3) {
```

mxmou: Does this really need to interact with theme3? I think it would make more sense to add a property to msg-count-badge settings indicating that the migration has been applied.

WorldLanguages: Yes, that would also work.

WorldLanguages: If we are discussing this I guess it means we all somewhat agree that this change (not how it's implemented) isn't too bad

mxmou: I agree with the change.

WorldLanguages: > Does this really need to interact with theme3?

It only reads from editor-theme3, I don't think it's too bad.
The only way the version doesn't get increased to 4 is a crash, so this can't run more than twice.

PR #7690, opened 2024-07-30
Remove scroll limits

Jazza-231:
Resolves: https://discord.com/channels/806602307750985799/1267516480614957218

### Changes

Added an addon that removes the bounds limit on the code area. This allows users to scroll as far as they want from their code. They can not get lost, as there is an existing button to center them on the code.

### Reason for changes

I often find myself wanting to scroll further away from code than scratch lets me. This situation gets worse when the bug with the collapsible block palette shows its head (which seems to be back)

### Tests

Works on brave. Dynamic enable and disable work. Allows scrolling as far as you want.

### More
Should handling of the scrollbars be different? Currently I don't touch them, and they eventually disappear. Maybe a better option is to not remove the fencing, but make the code area larger (through code? through shadow blocks?)
Zooming in (mouse scroll, mouse pinch, buttons) resets the view to the closest "normally allowed" point

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -158,6 +158,7 @@
   "workspace-dots",
   "asset-conflict-dialog",
   "remaining-replies",
+  "remove-editor-limits",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/remove-editor-limits/addon.json
+++ b/addons/remove-editor-limits/addon.json
@@ -1,0 +1,10 @@
+{
+  "name": "Remove code area limits",
+  "description": "Allows you to scroll as far away from existing code as you want.",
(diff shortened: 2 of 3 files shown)


Jazza-231:
Video demo:

https://cdn.discordapp.com/attachments/1267516480614957218/1267708701033238619/2024-07-30_13-01-00.mp4?ex=66a9c53b&is=66a873bb&hm=4aae90b2a92414fa6b5902cd8981d729b49c0d0d262a0d96c6f64c90c111a6ee&

WorldLanguages:
> They can not get lost, as there is an existing button to center them on the code.

What button do you mean here? The reset zoom button?

DNin01:
> > They can not get lost, as there is an existing button to center them on the code.
> 
> What button do you mean here? The reset zoom button?

I think so.

iqnite:
Microsoft Excel adapts the scrollbars when you scroll past the area you've written into.

PR #7697, opened 2024-08-02
Re-add forum timezone addon

### LINKED ISSUE #7672: Re-add forum timezone addon
I removed the "Local timezone on forums" addon in #7661 since it has been broken for multiple releases but it could be brought back if someone wants to re-implement it without an external dependency.

Ideally a new version would work everywhere and support 12-hour time.

ajskateboarder:
Resolves #7672

### Changes

I brought back the forum time zones addon by using a reliable alternative to the now-dead mobile forums. By posting to https://scratch.mit.edu/i18n/setlang with your language code and credentials omitted, you can use an English version of the page that doesn't modify user settings. Then you can parse the forum dates into local time - [thanks @NotFenixio for that](https://github.com/ScratchAddons/ScratchAddons/issues/7672#issuecomment-2247559085) - and use the same formatter from before

This addon will also apply your local timezone to "Last edited by user (time)" messages

### Reason for changes

It brings back an addon

### Tests

Works on Chrome 127 and Firefox 128

### DIFF
--- a/addons/forum-time-zones/userscript.js
+++ b/addons/forum-time-zones/userscript.js
@@ -109,7 +109,13 @@ export default async function ({ addon, console }) {
     const params = { method: "POST", body: form, credentials: "omit" };
 
     let html = await fetch("https://scratch.mit.edu/i18n/setlang/", params);
+    let count = 0;
+
     while (!html.ok) {
+      count++;
+      if (count > 3) {
+        throw new Error("Max retries exceeded while fetching https://scratch.mit.edu/i18n/setlang/");
+      }
       html = await fetch("https://scratch.mit.edu/i18n/setlang/", params);
     }
 
--- a/addons/forum-time-zones/userscript.js
+++ b/addons/forum-time-zones/userscript.js
@@ -1,0 +1,155 @@
+export default async function ({ addon, console }) {
+  const MONTH_TO_INT = {
+    "Jan.": 0,
+    "Feb.": 1,
+    March: 2,
+    April: 3,
+    May: 4,
+    June: 5,
+    July: 6,
+    "Aug.": 7,
+    "Sept.": 8,
+    "Oct.": 9,
+    "Nov.": 10,
+    "Dec.": 11,
+  };
+
+  /** @type {{edit: {el: HTMLElement; fetched: string; og: string} | undefined; post: {el: HTMLElement; fetched: string; og: string}}[]} */
+  const cache = [];
+  const locale = addon.auth.scratchLang;
+
+  const relativeFormatter = new Intl.RelativeTimeFormat(locale, {
+    localeMatcher: "best fit",
+    numeric: "auto",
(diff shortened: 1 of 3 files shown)


WorldLanguages:
This will probably not get merged in time for v1.39.0, so the addon ID will need to change to `forum-time-zones-v2`. Don't worry though, that can be changed just before merging, it's not necessary to change it now. I'm writing this so that we don't forget.

addons/forum-time-zones/userscript.js
```diff
@@ -0,0 +1,155 @@
...
+  let fetched = false;
+
+  if (locale !== "en-US") {
+    const form = new FormData();
+    form.append("language", "en");
+    // ommiting credentials prevents setting from becoming permanent
+    const params = { method: "POST", body: form, credentials: "omit" };
+
+    let html = await fetch("https://scratch.mit.edu/i18n/setlang/", params);
+    while (!html.ok) {
```
pufferfish101007, line 112: Might this cause an infinite request loop if something goes wrong? Why might this fail in the first place?

pufferfish101007 (changes requested):

addons/forum-time-zones/userscript.js
ajskateboarder: I noticed that /i18n/setlang is prone to "random" server errors, so I added in a while loop. You are right that it may cause an infinite request loop though.. would a max retry count be good?

WorldLanguages: @pufferfish101007 Nice catch

Yeah, we don't allow infinite loops that send requests even if they are theoretically impossible under normal conditions. You could try up to 3 times

ajskateboarder: Ok, will do

Samq64:
> This will probably not get merged in time for v1.39.0, so the addon ID will need to change to forum-time-zones-v2

Actually, is it necessary? The original addon had no settings and users who previously had it enabled wouldn't mind having this version automatically enabled.

Samq64 (changes requested):
EDT with me logged out:

screenshot-2024-08-27-18-42-07
screenshot-2024-08-27-18-42-39

Neither en-US or en-GB work. I don't see any errors in the console.

ajskateboarder:
> When I'm logged out (EDT):
> 
> screenshot-2024-08-27-18-42-07 screenshot-2024-08-27-18-42-39
> 
> One browser is en-US and the other is en-GB and neither display the correct time. I don't see any errors in the console.

@Samq64 if I'm remembering correctly, forum pages show local dates when signed out, so I'll add in a guard clause to check if the user is signed in

ajskateboarder:
sorry people I'm gonna have to figure out this one later

WorldLanguages:
I never figured out how the forums handle timezones. It's so weird.

PR #7698, opened 2024-08-02
`download-button`: include proj. ID in filename if Shift

WorldLanguages:
Resolves #7695

### Changes

Example project: https://scratch.mit.edu/projects/504994642/

When clicking the "download" button normally, the default name for the file is: `Happy Dango day (April 1)`

New: when shift+clicking the "download" button, the default name of the file is: `504994642 Happy Dango day (April 1)`

### Reason for changes

Small change that helps sfederici.
We shouldn't document this, if we ever want to change Shift+click to do something else we can.

### Tests

Tested in Chromium

### DIFF
--- a/addons/download-button/userscript.js
+++ b/addons/download-button/userscript.js
@@ -12,13 +12,15 @@ export default async function ({ addon, console, msg }) {
   const isOwn = username === projectAuthor;
   const shared = addon.tab.redux.state.preview.projectInfo.is_published;
 
-  async function download() {
+  async function download(beginFilenameWithId) {
     const downloadButton = document.querySelector(".sa-download-button");
     downloadButton.classList.add("loading");
     try {
       const project = await vm.saveProjectSb3();
       const title = isOwn ? document.querySelector(".project-title input") : document.querySelector(".project-title");
-      downloadBlob(`${isOwn ? title.value : title.innerText}.sb3`, project);
+      const titleStr = isOwn ? title.value : title.innerText;
+      const projectId = window.location.pathname.split("/")[2];
+      downloadBlob((beginFilenameWithId ? `${projectId} ` : "") + titleStr + ".sb3", project);
     } finally {
       downloadButton.classList.remove("loading");
     }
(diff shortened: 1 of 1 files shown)


WorldLanguages:
@sfederici Please read the "changes" and tell me if it's okay for your usecase.

sfederici:
This change would solve my problem. As I need it always, a future change in the addon settings could be even more useful.

DNin01 (commented):
Generally, I don't like features that are hidden behind key combinations - they're hard to discover (even for developers, without tips) and aren't available to touch screen users.

An alternative could be to offer an addon setting to set whether the ID should be included in the file name or not. That way, if you always want to know the ID in your downloads, you wouldn't have to remember to hold Shift.

Either way, this behavior differs from the File > Save to your computer option. To make that download button add the project's ID to the file name, it'd have to be done by a separate addon.

Additionally, when uploading projects into the editor, the file name is used as the project name, so if, for example, you downloaded one of your own projects with the ID in the file name and then loaded it back up, the ID would be part of the uploaded project's name and you'd have to edit it out manually. We could try to do that automatically, though.

WorldLanguages:
It's not worth it to add a setting into the extension (and translation process) for this in my opinion. But it's still a reasonable request so I implemented it

Samq64:
Could the changes be pushed?

Samq64 (approved):

Samq64:
This can be made a setting if it's requested again.

sfederici:
I wasn't aware this had been already integrated in ScratchAddons. Thanks! I would like to suggest just a small change, if I can: moving the ID at the end of the standard file name

Samq64:
Would you prefer "Project 1234", "Project (1234)" or "Project [1234]"?

sfederici:
> Would you prefer "Project 1234", "Project (1234)" or "Project [1234]"?

My favorite form would be "Project-1234" if possible. Thanks in advance.

WorldLanguages:
@Samq64 Did you implement the change? (sorry I've got too many notifications and can't read them all at once to know)

Samq64:
> @Samq64 Did you implement the change? (sorry I've got too many notifications and can't read them all at once to know)

Thanks for reminding me! I have just opened a PR.

PR #7702, opened 2024-08-03
Change the way we process cookie changes

TheColaber:
- Remove old rate limit logic
- Rather add a 500ms timeout that pushes back from each cookie change
- Once that timeout is able to finish, process cookie code for most recent cookie changes
- While process code is running, no cookie changes can be added to the queue. (this is to avoid any infinite loops)
- Fix the scratchlanguage cookie not triggering cookies.onChange and resulting in that process logic never running
- Add some comments

### DIFF
--- a/background/handle-auth.js
+++ b/background/handle-auth.js
@@ -44,7 +44,7 @@ const addToQueue = (item) => {
   };
 
   const { cookie } = item;
-  if (cookie.name !== "scratchsessionsid" && cookie.name !== "scratchlanguage" && cookie.name !== "scratchcsrftoken") {
+  if (cookie.name !== "scratchsessionsid" && cookie.name !== "scratchcsrftoken") {
     // Ignore this event
     return;
   }
@@ -57,7 +57,7 @@ const addToQueue = (item) => {
 };
 
 const processCookieChanges = () => {
-  // Keys could only be "scratchsessionsid", "scratchlanguage", or "scratchcsrftoken"
+  // Keys could only be "scratchsessionsid" or "scratchcsrftoken"
   const mostRecentCookies = {};
   for (const change of cookieQueue.reverse()) {
     if (!mostRecentCookies[change.name]) {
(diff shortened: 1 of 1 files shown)


background/handle-auth.js
```diff
@@ -7,178 +7,216 @@ const promisify =
   (...args) =>
     new Promise((resolve) => callbackFn(...args, resolve));
 
+const CHROME_DEFAULT = "0";
+const FIFEFOX_DEFAULT = "firefox-default";
 async function getDefaultStoreId() {
```
WorldLanguages, line 12: There's nothing wrong with defining constants inside the function they're used in.

WorldLanguages: `Fifefox`? Do you mean Firefox?

WorldLanguages: Whoops, it was my tipo after all.

background/handle-auth.js
```diff
@@ -9,176 +9,214 @@ const promisify =
...
+
+// It's good to know that chrome.cookies works without internet connection.
+const getCookieValue = async (name) => {
+  const cookie = await chrome.cookies.get({
+    url: "https://scratch.mit.edu/",
+    name,
   });
-}
+  return (cookie && cookie.value) || null;
+};
```
WorldLanguages, line 150: It's not a good opportunity to rewrite functions. You slow down reviews!

libraries/common/message-cache.js
```diff
@@ -177,7 +177,7 @@ export async function openMessageCache(cookieStoreId, forceClear) {
  */
 export async function updateMessages(cookieStoreId, forceClear, username, xToken) {
   await openMessageCache(cookieStoreId, forceClear);
-  if (username === null) return [];
+  if (typeof username === "undefined" || username === null) return [];
```
WorldLanguages, line 180: Why do we need to change this file? Is it actually related to the other changes?

TheColaber: yeah, now it will pass in undefined vs passing in null. it matters

PR #7711, opened 2024-08-07
Add early return to 3.0 scratchblocks addon

jeffalo:
If Scratch updates to a newer scratchblocks plugin, it would use the filename scratchblocks.min.js (I know this because I implemented it), disable 3.0 scratchblocks addon with an early return as it would be redundant.

### DIFF
--- a/addons/scratchblocks/userscript.js
+++ b/addons/scratchblocks/userscript.js
@@ -46,6 +46,9 @@ async function getLocales(addon) {
   return lang;
 }
 export default async function ({ addon, msg }) {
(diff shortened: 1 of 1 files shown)


GrahamSH-LLK (approved):
tested w/ response override, early return works

GrahamSH-LLK:
We actually don't need the early return, the waitForElement below should handle it

jeffalo:
whoops!

WorldLanguages:
> We actually don't need the early return, the waitForElement below should handle it

Nice catch!

Issue #5513, opened 2022-12-29
MIDI device support


DNin01:
### Why this would be helpful

This addon could allow projects to be controlled by musical instruments. This could let us make projects designed around instrument control, like projects that teach you how to play an instrument or let you record songs.

### How this addon works

Basically, it's the same thing as `gamepad`, but for MIDI devices. You'd map notes on the instrument to keys on your keyboard.

WorldLanguages:
oh boy this would be so interesting...

WorldLanguages:
I guess this is already workaroundable, tho. Wouldn't any software that lets you map MIDI device keys to normal keys work fine for Scratch? Unlike gamepads with joysticks, it sounds like a global solution is good enough for Scratch usage.

mxmou:
> I guess this is already workaroundable, tho. Wouldn't any software that lets you map MIDI device keys to normal keys work fine for Scratch? Unlike gamepads with joysticks, it sounds like a global solution is good enough for Scratch usage.

That might actually be better because projects designed to be used with a MIDI device could mention the software in the description without breaking Scratch's rules.

WorldLanguages:
This can be reopened if Scratch-specific functionality would be required in some way.

DNin01:
> > I guess this is already workaroundable, tho. Wouldn't any software that lets you map MIDI device keys to normal keys work fine for Scratch? Unlike gamepads with joysticks, it sounds like a global solution is good enough for Scratch usage.
> 
> 
> 
> That might actually be better because projects designed to be used with a MIDI device could mention the software in the description without breaking Scratch's rules.

As long as the software wasn't a browser extension. Also, they'd need to make sure what they're referencing is safe, even if it isn't a browser extension.

Maybe the projects could not say that you could use a MIDI device but instead users could find their own way to set up keyboard mappings and software themselves to enable the usage of MIDI devices. I don't talk about how to connect game controllers or enable 60 FPS modes in my projects, but I do make it possible sometimes, for anyone interested and with that knowledge. I could do that with a musical-instrument project normally controlled with the keyboard but perfectly capable of being controlled by a MIDI device, with the only limitation being the number of keys that are accessible.

And yeah, I agree that general-purpose MIDI control software would cover this.

WorldLanguages:
If #7714 gets merged, this could be part of the same addon (MIDI output **and input**)

DNin01:
[...]

PR #7714, opened 2024-08-09
New addon: MIDI output

mxmou:
This is a feature suggestion, but I'm opening a draft PR instead of an issue because I already have a working implementation. It's the opposite of #5513: it allows projects to send (but not receive) MIDI messages using the Music extension.


mxmou:
I used this project with multiple instruments for testing: https://scratch.mit.edu/projects/992168376

addons/midi/addon.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "MIDI output",
+  "description": "Connect Music blocks to a MIDI device.",
```
WorldLanguages, line 3: Unlike the "gamepad support" addon, many of our users (myself included) are probably not familiar with MIDI devices. Maybe we can mention an example of what can be done with the addon here, and how to know if a device is compatible.

PR #7716, opened 2024-08-09
Remove unnessary uses of safeMsg in data-category-tweaks and devtools

TheColaber:


mxmou (approved):

Samq64 (approved):

addons/data-category-tweaks-v2/userscript.js
```diff
@@ -152,6 +152,7 @@ export default async function ({ addon, console, msg, safeMsg }) {
   // https://github.com/scratchfoundation/scratch-vm/blob/a0c11d6d8664a4f2d55632e70630d09ec6e9ae28/src/engine/runtime.js#L1381
   const originalGetBlocksXML = vm.runtime.getBlocksXML;
   vm.runtime.getBlocksXML = function (target) {
+    ScratchBlocks.Msg["CATEGORY_LISTS"] = msg("list-category");
```
WorldLanguages, line 155: Is this name arbitrary? If only we are going to use it, why not choose a less general name?

WorldLanguages (changes requested):
I need to second check every change for security reasons.

addons/data-category-tweaks-v2/userscript.js
TheColaber: it follow the naming consistancy of scratch but is could be sa-* if wanted.

WorldLanguages: @TheColaber Change it to `SA_CATEGORY_LISTS` then

PR #7720, opened 2024-08-12
Full area updates

### LINKED ISSUE #3858: `full-areas` breaks some forum signatures that appear normal without a scrollbar
### Description

the `full-areas` addon causes some signatures that are made with images to appear with a linebreak
### Repoduction steps

1. Turn on full-areas and its setting "Show scroll for forum signatures"
2. go to [one of my forum posts](https://scratch.mit.edu/discuss/post/5821248/), which contains an example of a broken signature
3. See that the signature cuts off at the 4th image

### Expected behaviour

It shouldn't cause it to cut off (maybe use a thin scrollbar?)
Example of what should happen:
image
A thin scrolbar, works much better & doesn't linebreak
### Actual behaviour

The last image cuts off.

### Screenshots
image

### Environment

- Browser: Firefox 94
- Scratch Addons version: 1.22.1
- Operating system: Windows 10

### Additional context

This might work differently on chrome or other chromium based browser? let me check

Samq64:
Mostly Resolves #3858
Closes #5609

### Changes

- Scrolls to the top-left when dynamically disabled
- Uses thin scrollbars on signatures
- Adds the option to scroll project titles
- Shortens the setting names

### Reason for changes

Being able to scroll other's project titles is nice and leaving half-scrolled areas when disabling isn't ideal.

### Tests

Tested on Chromium

### DIFF
--- a/addons/full-signature/addon.json
+++ b/addons/full-signature/addon.json
@@ -19,26 +19,32 @@
   ],
   "settings": [
     {
-      "name": "Show scroll for \"What I've been doing\"",
+      "name": "What I've been doing",
       "id": "whatworkingon",
       "type": "boolean",
       "default": true
     },
     {
-      "name": "Show scroll for \"What's Happening?\"",
+      "name": "What's Happening?",
       "id": "whathappen",
       "type": "boolean",
       "default": true
     },
     {
-      "name": "Show scroll for forum signatures",
+      "name": "Forum signatures",
       "id": "signature",
       "type": "boolean",
       "default": true
     },
     {
-      "name": "Show scroll for forum Scratchblocks",
+      "name": "Forum Scratchblocks",
       "id": "blocks",
+      "type": "boolean",
+      "default": true
+    },
+    {
+      "name": "Project titles",
+      "id": "titles",
       "type": "boolean",
       "default": true
     }
@@ -55,6 +61,10 @@
       "url": "happen.js",
       "matches": ["https://scratch.mit.edu/"],
       "runAtComplete": false
+    },
+    {
+      "url": "disable.js",
+      "matches": ["projects", "profiles", "topics"]
     }
   ],
   "userstyles": [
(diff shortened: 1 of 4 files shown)


addons/full-signature/disable.js
```diff
@@ -0,0 +1,12 @@
+export default async function ({ addon, console }) {
+  addon.self.addEventListener("disabled", () => {
+    document
+      .querySelectorAll(
+        ".activity-stream, .activity-ul, .postsignature, .project-title.no-edit, .scratchblocks, .blocks3"
+      )
+      .forEach((el) => {
+        el.scrollTop = 0;
+        el.scrollLeft = 0;
```
Samq64, line 9: Will this need RTL support?

mxmou: Scratch only supports RTL layout in the editor.

mxmou: I changed a project's title to Arabic text using Inspect and only the *end* of the title was visible, even if I set the language to Arabic.

Samq64:
https://github.com/ScratchAddons/ScratchAddons/issues/3858#issuecomment-981360050
> Thin scrollbars could still cause issues with slightly wider signatures. Could the width of the signature area be increased instead?

When I tried the scrollbar was clipped.

addons/full-signature/addon.json
```diff
@@ -19,28 +19,34 @@
...
       "id": "blocks",
       "type": "boolean",
       "default": true
+    },
+    {
+      "name": "Project titles",
+      "id": "titles",
+      "type": "boolean",
+      "default": true
     }
```
mxmou: This should probably be above the forum settings.

addons/full-signature/titles.css
```diff
@@ -0,0 +1,4 @@
+.preview .project-title.no-edit {
+  overflow-x: auto;
+  text-overflow: clip;
```
mxmou, line 3: I'd add `scrollbar-width: thin` here too - the default scrollbar uses a significant amount of vertical space on some browsers.

addons/full-signature/disable.js
```diff
@@ -0,0 +1,12 @@
+export default async function ({ addon, console }) {
+  addon.self.addEventListener("disabled", () => {
```
mxmou: Could something similar be done when a setting is dynamically disabled?

mxmou (approved):

Samq64:
Oops, I forgot to bump latestUpdate.

Edit: fixed by #8491

PR #7737, opened 2024-08-19
Move `chrome.*` apis in settings page to seperate file

TheColaber:
I did update some older code but this is mostly just code moving to a central file.

This will probably get changed around a little bit to probably move even more code into this new file. Imagine this settings page implementation in TruboWarp... the goal is to having only to change this file to allow everything to work normally.

This needs some good amount of testing.


WorldLanguages:
After this is merged I should create an issue explaining what we want the following steps to be in regards to further separating the settings page from the extension.

webpages/settings/index.js
```diff
@@ -665,5 +643,7 @@ let fuse;
     }
   });
 
-  chrome.runtime.sendMessage("checkPermissions");
+  sendPermissionCheck();
 })();
+
+checkAndOpenUnsupportedPage();
```
WorldLanguages, line 649: Setting pages shouldn't need to call this. Instead, the extension itself should call it.

WorldLanguages: That also means, there's no reason to expose this function to settings pages.

webpages/settings/index.js
```diff
@@ -639,8 +617,8 @@ let fuse;
   window.onresize = resize;
   resize();
 
-  chrome.management.getSelf((info) => {
-    if (info.installType === "development") vue.devMode = true;
+  getInstallType().then((installType) => {
```
WorldLanguages, line 620: It's way more handy if we can expose an `installType` variable. Would also remove the need for all of this code to be asynchronous.

webpages/settings/index.js
```diff
@@ -624,7 +602,7 @@ let fuse;
     }
   });
 
-  document.title = chrome.i18n.getMessage("settingsTitle");
+  document.title = getMessage("settingsTitle");
```
WorldLanguages, line 605: Again, the responsibility of setting the tab title should be separate and would not belong here.

webpages/settings/index.js
TheColaber: it'd be pretty hard to guarantee the call to setup this value would finish before it is requested

webpages/settings/index.js
```diff
@@ -189,19 +194,13 @@ let fuse;
...
       },
       openReview() {
-        if (typeof browser !== "undefined") {
-          window.open(`https://addons.mozilla.org/en-US/firefox/addon/scratch-messaging-extension/reviews/`);
-        } else {
-          window.open(
-            `https://chrome.google.com/webstore/detail/scratch-addons/fbeffbjdlemaoicjdapfpikkikjoneco/reviews`
-          );
-        }
+        window.open(getReviewURL());
```
Samq64, line 203: `openReview()` hasn't been used in ages, it can just be deleted.

webpages/settings/settings-utils.js
Samq64, line 1: Should this file be kept separate or be merged with `settings-page-apis.js`?

Samq64:
@TheColaber I've updated your branch and made a couple minor changes. I hope you don't mind.

Issue #6662, opened 2023-09-17
Right click option on stack in editor: convert to scratchblocks


ksdavidc:
### Why this would be helpful

I use scratchblocks (sb) pseudocode in teaching, and sometimes convert it to images for printing worksheets, etc.

Right now I hand type the sb code, but it would be **AWESOME** if i could get the code directly from the editor for pasting into my text editor. 

### My suggestion

When you right click a stack (Duplicate, add comment, etc.) it would be nice if there were an option to copy the sb pseudocode for that stack (ultimately, for all the code in a sprite, project??? ;-) ) say below where copy block, etc are.

I imagine this might be possible to implement, since SA already has some sb related addons, so you must have a way of working with sb, but I don't have a clue of the details of how that would be done.

### Possible alternatives

Right now I hand type the sb code, but it would be **AWESOME** if I could get the code directly from the editor for pasting into my text editor. 

Sometimes I use a parse-sb3-blocks, but this has some limitations.

### Additional context

Apologies if this is a duplicate. I did a search and nothing similar seemed to come up.

ksdavidc:
And another thing: convert sb to actual code blocks??? hahahahaha well, one can dream....

Samq64:
There's already an addon to save script images: https://scratch.mit.edu/scratch-addons-extension/settings#addon-blocks2image

As for copying the scratchblocks BBCode, that's a duplicate of #2336.

> And another thing: convert sb to actual code blocks??? hahahahaha well, one can dream....

It was actually possible to backpack scratchblocks at one point and there's a suggestion for that too: #3225.

ksdavidc:
Thank you for looking at this. However, I think you are referring to different issues.

> There's already an addon to save script images: https://scratch.mit.edu/scratch-addons-extension/settings#addon-blocks2image
> 
Not what I want. I want text, not images. _Sometimes_ I use images, but what I really need is the text so I can edit the text directly. 

> As for copying the scratchblocks BBCode, that's a duplicate of #2336.

I understand, but that is not what I am asking. I am specifically talking about the editor and scratchblocks and not bbcode. 

a) I am not sure, but I am under the impression scratchblocks is a different repo and a thing unto itself.. I don't know much about Bbcode, but I think it recognizes sb, but is an overset of it created for a different purpose. I could be wrong...

[...]

PR #7748, opened 2024-08-23
New addon: Copy scratchblocks text from editor blocks

### LINKED ISSUE #2336: New editor addon: copy block/stack to clipboard (scratchblocks/BBCode)
so basically if you right click a script in the editor you can copy the bb code from the script into your clip board and paste it in the forums.

pumpkinhasapatch:
Resolves #2336 #6662

### Changes

I made a new addon that allows real blocks in the Scratch Project Editor to be generated as scratchblocks text tags. It adds a new right-click option called "Copy scratchblocks text" to blocks in the project editor. When this is clicked, it finds the JSON code of the selected blocks and uses [parse-sb3-blocks](https://github.com/apple502j/parse-sb3-blocks) by @apple502j to convert them into scratchblocks text, then copies it to the user's clipboard for easy use on other websites or text editors.

[Screencast from 2024-08-23 02-32-05.webm](https://github.com/user-attachments/assets/6d7d07f2-306d-4824-98fe-f10af98aa390)

### Reason for changes

It would be convenient for users to be able to copy and paste blocks from the Project Editor in places where only scratchblocks text is supported, like the Scratch Forums using BBCode tags or Scratch Wiki pages using the [Block Plugin](https://en.scratch-wiki.info/wiki/Block_Plugin).

### Tests

Tested on Chromium with Scratch Addons v1.39.0-pre.

There is currently a bug that causes some types of blocks containing inputs to not convert correctly into scratchblocks and throw errors in the console. I think this is caused by block serialization, a process where the JSON formatting for blocks stored in an .sb3/project.json is changed slightly by [this code in scratch-vm](https://github.com/scratchfoundation/scratch-vm/blob/1eec35a3b7b2f1afb0a633ef931ce553a2c9791b/src/serialization/sb3.js) when it is loaded into the editor, but can't figure out how to fix it. :/

<details>
<summary>Click to show affected blocks</summary>
Screenshot from 2024-08-03 22-31-54
</details>

The current code contains commented-out example variable values and JavaScript mappings for parse-sb3-blocks. These are not necessary for the addon to function, but make it easier to debug any issues in the Browser Console and may be removed before merging.


WorldLanguages:
@TheColaber Do you know how to fix the bugs mentioned in this PR?

TheColaber:
i am not familiar with this library, maybe @apple502j can help since they made it. otherwise, the best solution to make a library that converts ScratchBlocks to bbcode.

mxmou:
I think the right way to implement this is to [serialize the sprite to JSON](https://github.com/scratchfoundation/scratch-vm/blob/32a2054/src/virtual-machine.js#L459-L467), then pass the result to the parse-sb3-blocks library.

addons/editor-copy-scratchblocks/userscript.js
```diff
@@ -0,0 +1,58 @@
...
+    { blocks: true }
+  );
+
+  async function convertToScratchblocks(block) {
+    // const blockId = block.getRootBlock().id; // Start from the topmost block, parsing the entire stack
+    let blockId = block.id;
+
+    // Get blocks in current sprite as project.json format
+    const vm = addon.tab.traps.vm;
+    let blocksJSON = vm.editingTarget.blocks._blocks;
```
mxmou, line 37:
```suggestion
    let blocksJSON = JSON.parse(vm.toJSON(vm.editingTarget.id)).blocks;
```

pumpkinhasapatch: Thanks, that code works beautifully. I did some testing and it seems like this addon now works with any combination of blocks in vanilla Scratch. (parse-sb3-blocks would need code changes to support modded blocks/extensions if this addon gets put into TurboWarp or similar.)

mxmou: > (parse-sb3-blocks would need code changes to support modded blocks/extensions if this addon gets put into TurboWarp or similar.)

TurboWarp also serializes some vanilla blocks differently - see https://github.com/apple502j/parse-sb3-blocks/issues/9

Samq64:
The library should be minified and stored in `/libraries/thirdparty/cs` but even then it's 1.7MB which is quite large considering the whole extension is 20MB uncompressed.

pumpkinhasapatch:
> The library should be minified and stored in `/libraries/thirdparty/cs` but even then it's 1.7MB which is quite large considering the whole extension is 20MB uncompressed.

The parse-sb3-blocks library was minified already. When I uploaded it here the code bot tried to format the entire minified file and probably increased the file size.

The library is also very big because it contains the text for all blocks in every language Scratch supports. Blocks in LOTE do not render correctly on the forums, but they might be useful for some people on Scratch Wiki or other places scratchblocks is supported.

pumpkinhasapatch:
Any more feedback?

pumpkinhasapatch:
> I'm not sure if these are bugs with the library but the stop block includes the text `{STOP_OPTION}` for some reason and copying the translate blocks throws the error `Cannot read properties of undefined (reading 'toScratchblocks')` in the library.

@Samq64 These are issues with the parse-sb3-blocks library. All Scratch blocks are defined in this [block mapping](https://github.com/apple502j/parse-sb3-blocks/blob/main/src/block-mapping/all-blocks.js): `{STOP_OPTION}` is a placeholder for the stop block menu, for some reason the menu text is added after `{STOP_OPTION}` instead of replacing it, giving the text `stop {STOP_OPTION} [all v]` on this one block. We should create an issue on Apple's repo for this.

The addon does not work directly on reporter or boolean shaped blocks. You have to put them inside of a normal square shaped block and right click that to get the block text. This is a limitation of the library but I might be able to make a workaround that automatically does this in the addon's code. 

My addon just finds selected blocks in the Scratch editor and gives it to the library, the addon does not do the actual block-to-text conversion.

pumpkinhasapatch:
> I'm not sure if these are bugs with the library but the stop block includes the text `{STOP_OPTION}` for some reason and copying the translate blocks throws the error `Cannot read properties of undefined (reading 'toScratchblocks')` in the library.

@Samq64 These are issues with the parse-sb3-blocks library. All Scratch blocks are defined in this [block mapping](https://github.com/apple502j/parse-sb3-blocks/blob/main/src/block-mapping/all-blocks.js): `{STOP_OPTION}` is a placeholder for the stop block menu, for some reason the menu text is added after `{STOP_OPTION}` instead of replacing it, giving the text `stop {STOP_OPTION} [all v]` on this one block. We should create an issue on Apple's repo for this.

The addon does not work directly on reporter or boolean shaped blocks. You have to put them inside of a normal square shaped block and right click that to get the block text. This is a limitation of the library but I might be able to make a workaround that automatically does this in the addon's code. 

My addon just finds selected blocks in the Scratch editor and gives it to the library, the addon does not do the actual block-to-text conversion.

NeuronPulse:
May I ask if this process can be reversed, i.e., from Scratchblocks text to blocks? For example, the graphical kittenN editor similar to scratch, they can convert blocks and text, which is a very useful function.

Samq64:
No, but there's an issue for that: #8573

Samq64:
I am closing this because none of the discussion starting at https://github.com/ScratchAddons/ScratchAddons/issues/2336#issuecomment-3312454476 favours this implementation specifically over https://github.com/ScratchAddons/ScratchAddons/pull/7988 or https://github.com/ScratchAddons/ScratchAddons/pull/8572. The current consensus seems to be to generate mappings from Scratch's translations which none of the implementations currently do but the other PRs already parse in the userscript making them easier to adapt.

PR #7750, opened 2024-08-24
Fix preview tabs compat

### LINKED ISSUE #6698: `preview-project-description` incompatible with `project-notes-tabs`
### Description

Currently, there are multiple issues relating to compatibility with `project-notes-tabs` in `preview-project-description` that weren't fixed in the PR that added the addon that we should eventually fix. The following issues only happen when combined with `project-notes-tabs`.

- [ ] Enabling and disabling Preview removes the tabs
brave_3pUitfoQL9

---

- [ ] The Preview toggle is placed above the remix attribution card
Elements in order: preview toggle, remix attribution, instructions tab, notes and credits tab, description field

---

- [ ] Entering and exiting the editor, things get even weirder

### Steps to reproduce

Open a project page with both of these addons enabled. Steps to reproduce individual issues outlined above.

_No response_

### Environment

Browser: Edge 117
Scratch Addons version: https://github.com/ScratchAddons/ScratchAddons/pull/6697/commits/478a606de0948ac95a1099066e597fa5c4da5dd9
Operating system: Windows 11

Samq64:
Resolves #6698

### Changes

Adds dynamic enable/disable to `project-notes-tabs` and fixes compatibility with `preview-project-description`.

### Tests

Tested on Chromium. I still need to fix some console errors when entering the editor and test on remixed projects. Dynamic login is still buggy.

### DIFF
--- a/addons/delete-others/addon.json
+++ b/addons/delete-others/addon.json
@@ -1,7 +1,7 @@
 {
   "name": "Delete other costumes or sounds",
   "description": "Adds an option to the right click context menu for costumes and sounds to delete all items except for the selected one.",
-  "tags": ["editor", "costumeEditor", "recommended"],
+  "tags": ["editor", "costumeEditor", "featured"],
   "versionAdded": "1.39.0",
   "userscripts": [
     {
--- a/addons/preview-project-description/userscript.js
+++ b/addons/preview-project-description/userscript.js
@@ -1,4 +1,4 @@
-import { disableTabs, enableTabs } from "../project-notes-tabs/module.js";
+import { disableTabs, enableTabs, eventTarget } from "../project-notes-tabs/module.js";
 
 export default async function ({ addon, console, msg }) {
   const divElement = Object.assign(document.createElement("div"), {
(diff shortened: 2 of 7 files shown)


WorldLanguages:
Does this also fix the project-notes-tabs bugs mentioned in #6698? (those are unrelated to addon compatibility, but it's a good opportunity to fix them)

Samq64:
Yes it does now.

mxmou:
When enabling or disabling the preview with project-notes-tabs enabled, the switch moves slightly, which is annoying. This could be fixed by changing the CSS to always position the switch the same way, but I think it would be even better to keep the tabs while previewing if that's possible.

Samq64:
> I think it would be even better to keep the tabs while previewing if that's possible.

The preview could also be used to check how it looks of non-Scratch Addons users, but it doesn't turn off other addons.

mxmou:
@WorldLanguages I think you wrote the code that disables tabs while previewing - do you remember if that was a design choice or a workaround for a bug? If the preview is meant to show how users without SA will see the description, other addons should also be disabled, especially more links.

WorldLanguages:
It was not an intentional design choice. You're right about the current behavior being inconsistent. I think we should choose a side solely on whether "More Links" is easy to temporarily turn off just for this area of the page.

mxmou:
> It was not an intentional design choice. You're right about the current behavior being inconsistent. I think we should choose a side solely on whether "More Links" is easy to temporarily turn off just for this area of the page.

Another thing to consider: if we choose not to disable addons, there will be less work when new addons are added in the future.

WorldLanguages:
> I think it would be even better to keep the tabs while previewing if that's possible.

Yeah, we should try to keep the tabs then.

Issue #7761, opened 2024-08-28
Scratch Addons v1.39.0 doesn't work in Firefox


JakubMzTrencina:
### Description

I've opened Scratch today, and the first thing I've noticed was the navigation bar being purple (I use an addon to make it blue, as purple isn't my color) Then I checked the forums, and the field where one writes a post was lacking some features I've enabled on SA. Opening the SA configuration, it seems it's loading forever, and doesn't do anything. I tried disabling them, and even reinstalling, but it didn't fix the issue. I suspect it happened when SA updated.

Edit: The affected version was removed by World-Languages from the Firefox store. If you've encountered the same bug, you probably have the 1.39.0 version. You can downgrade by following [this](https://github.com/ScratchAddons/ScratchAddons/issues/7761#issuecomment-2315312182).

Any existing settings from before the update will persist.

### Steps to reproduce

I don't really think there's a way to reproduce this issue, just try loading the Scratch page or openning SA and you may or may not see the issue.

### Environment

Browser: Firefox 129.0.2
Scratch Addons version: v1.39.0
Operating system: Windows 10

JakubMzTrencina:
i've just realised i should've used the "Send Feedback" page on the SA webpage
ah well, i guess feel free to close this, if you find it useless

CST1229:
I'm also encountering this issue by the way. I posted this in the SA Discord too:

> The settings page errors when trying to get the extension's settings, apparently. Maybe it's some error in the background page (or whatever its MV3 equivalent is).
> ![image](https://github.com/user-attachments/assets/ae6b8c65-3905-44ab-a43e-b6397732fe71)
> And the popup has the error TypeError: can't access property "manifests", res is undefined.
> ![image-1](https://github.com/user-attachments/assets/fb8c056b-d9e3-4cd3-88c8-5c1e0e3750f7)
> I tried restarting the browser and the extension several times but it didn't fix the issue.
> Inspecting the extension's background(???) page using about:debugging then reloading it, it seems that it complains about chrome.scripting not being available?
> ![image](https://github.com/user-attachments/assets/dd5e5314-ceac-4f64-b11f-dff783ec984b)
> 
> Firefox version: Firefox Developer 130.0b9 (64-bit)

> now the extension runs into this error (showing the error page)
> ![image](https://github.com/user-attachments/assets/57d0ceb5-259b-45aa-863a-c443426462da)
> and the extension.... works perfectly when loading it from file?

[...]

PR #7762, opened 2024-08-28
Check if chrome.scripting exists before using it

CST1229:
Tries to address #7761, not sure if this will fix it or not.

### Changes

Checks for the `chrome.scripting` API being available before using it in firefox-local-support.js, as it does not exist on Firefox for release builds (which do not have the `scripting` permission).

### Reason for changes

Without this, the extension completely breaks on Firefox.

### Tests

I think this is a Firefox-specific bug, so I only tested there. I tested by manually removing the `scripting` permission from manifest.json, and it seems to make the extension work.

### DIFF
--- a/background/firefox-localhost-support.js
+++ b/background/firefox-localhost-support.js
@@ -1,4 +1,4 @@
-if (typeof browser === "object") {
+if (typeof browser === "object" && chrome.scripting) {
   const manifest = chrome.runtime.getManifest();
   const manifestScripts = manifest.content_scripts.filter((script) =>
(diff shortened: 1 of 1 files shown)


rickbot-dot:
Single line of code changed, LGTM

WorldLanguages (approved):
My bad

Samq64 (approved):
Localhost still works fine with the permission.

WorldLanguages:
I will merge this today, don't panic

Samq64:
The `chrome.scripting = "sa_no_scripting"` line at the bottom of the file could be removed.

Issue #7577, opened 2024-06-23
Remove ScratchDB addons


Samq64:
ScratchDB is down for the foreseeable future (https://scratchdb.lefty.one/) so we should remove addons that don't work without it.

Effected addons: 
- `scratchstats`
- `forum-search`
- `exact-count` (forum post count)
- `my-ocular` (might be fine)

mxmou:
> `my-ocular` (might be fine)

`my-ocular` still works.

DNin01:
We could also keep these addons for now but tuck them away somewhere and add a note to each of them explaining why they aren't working.

Related: #6987

Secret-chest:
Could Lefty publish the software so we can host it?

DNin01:
> Could Lefty publish the software so we can host it?

You could try asking him. Note that it is likely much more complicated than just typing a command, and it could take weeks, sometimes months to finish building an index.

There may also be some people who are building new projects to take ScratchDB's place.

sfederici:
> ScratchDB is down for the foreseeable future (https://scratchdb.lefty.one/) so we should remove addons that don't work without it.
> 
> Effected addons:
>     * `exact-count` (forum post count)

I have the exact-count addon enabled for counting projects in studios, but I don't get the exact number anymore

sfederici:
> You need to click the count to get the exact number.

I didn't know that. Thanks a lot for letting me know. Maybe this could be added to the information box of the addon?

mxmou:
> > You need to click the count to get the exact number.
> 
> I didn't know that. Thanks a lot for letting me know. Maybe this could be added to the information box of the addon?

I'll make a PR.

mxmou:
I think this is implemented:
* `scratchstats` and `exact-count`'s forum setting have been removed
* `forum-search` now uses the new ScratchDB API
* `my-ocular`'s "View on ocular" link works again after https://github.com/jeffalo/ocular/commit/73b0791718b5a64b57c0b964198865a931050923

PR #7765, opened 2024-08-29
exact-count: explain that a click is required in studios

mxmou:
### Changes

Adds the following info to the exact count addon:
image

### Reason for changes

https://github.com/ScratchAddons/ScratchAddons/issues/7577#issuecomment-2317063674

### Tests

Tested on Edge.

### DIFF
--- a/addons/exact-count/addon.json
+++ b/addons/exact-count/addon.json
@@ -16,6 +16,10 @@
   ],
   "info": [
     {
+      "id": "studioClick",
+      "text": "To see the exact number of projects in a studio, click the \"100+\" on the Projects tab. The \"exact count for studios\" setting below needs to be enabled."
+    },
+    {
       "id": "setting-removal",
       "text": "The forum setting has been removed since the ScratchDB database used by it is offline."
     }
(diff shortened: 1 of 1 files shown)


Samq64 (approved):

addons/exact-count/addon.json
```diff
@@ -15,6 +15,10 @@
     }
   ],
   "info": [
+    {
+      "id": "studioClick",
+      "text": "To see the exact number of projects in a studio, click the \"100+\" on the Projects tab. The \"exact count for studios\" setting below needs to be enabled."
```
DNin01, line 20:
```suggestion
      "text": "To see the exact number of projects in a studio, click the \"100+\" label on the Projects tab. The \"exact count for studios\" setting below also needs to be enabled."
```

mxmou: To me, "also needs to be enabled" sounds like this setting isn't the only thing that needs to be enabled. I think the following might be better:
```suggestion
      "text": "To see the exact number of projects in a studio, enable the \"exact count for studios\" setting below and click the \"100+\" label on the studio's Projects tab."
```

DNin01 (approved):

PR #7767, opened 2024-08-30
Mute for 4 hours and until tomorrow options

DNin01:
### Changes

Changes the "Do not disturb" context submenu:
- Removes the "24 hours" option and replaces it with "until tomorrow", which lasts until midnight the next day.
- Also adds a "4 hours" option.

Scratch Addons extension action dropdown with "Do not disturb" selection submenu with options "For 15 minutes", "For 1 hour", "For 4 hours", "For 8 hours", "Until tomorrow", and "Until turned off"

### Reason for changes

24 hours would most often be used to prevent distractions for an entire day, but that could have the potential to extend several hours into the following day.

### Tests

Tested in Edge 128.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -36,7 +36,7 @@
     "message": "Until tomorrow"
   },
   "untilEnabled": {
-    "message": "Until I turn it off"
+    "message": "Until turned off"
   },
   "muteFor": {
     "message": "Do not disturb"
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -26,8 +26,11 @@
   "1hour": {
     "message": "For 1 hour"
   },
-  "nhours": {
-    "message": "For $1 hours"
+  "4hours": {
+    "message": "For 4 hours"
+  },
+  "8hours": {
+    "message": "For 8 hours"
   },
   "untilTomorrow": {
     "message": "Until tomorrow"
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -26,14 +26,14 @@
   "1hour": {
     "message": "For 1 hour"
   },
-  "8hours": {
-    "message": "For 8 hours"
+  "nhours": {
+    "message": "For $1 hours"
   },
-  "24hours": {
-    "message": "For 24 hours"
+  "untilTomorrow": {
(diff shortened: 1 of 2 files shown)


TheColaber:
this is an interesting change since we copied the values from discord and other bug companies but the changes make sense 🤷‍♂️

Samq64:
Could `1hour` be merged into `nhours`?

DNin01:
> Could `1hour` be merged into `nhours`?

It could, but the string would need to be singular or plural, otherwise it would say "1 hours". I haven't yet checked if that works when using chrome.i18n.

mxmou:
In Slovenian, the word "hours" has different forms in "4 hours" and "8 hours". If the chrome.i18n API supports plurals, use them, if not, it's probably best to make two separate strings.

WorldLanguages:
Separate strings in this case, because of languages with multiple plurals

_locales/en/messages.json
```diff
@@ -26,14 +26,17 @@
...
     "message": "For 8 hours"
   },
-  "24hours": {
-    "message": "For 24 hours"
+  "untilTomorrow": {
+    "message": "Until tomorrow"
   },
   "untilEnabled": {
-    "message": "Until I turn it back on"
+    "message": "Until I turn it off"
```
WorldLanguages, line 39: Discord uses `Until I turn it back on`, do we actually know better than them?

WorldLanguages (approved):

_locales/en/messages.json
WorldLanguages: image

mxmou: In Discord it makes sense because "it" = "server". The context in our UI is different, so I think something shorter like "Until unmuted" would be better.

DNin01: > In Discord it makes sense because "it" = "server". The context in our UI is different, so I think something shorter like "Until unmuted" would be better.

We've replaced the term "mute" with "do not disturb", but we could say "until turned off".

DNin01:
```suggestion
    "message": "Until turned off"
```

mxmou (approved):
Not tested

PR #7777, opened 2024-09-03
addon.tab.editorMode: Use Redux on /projects/editor/

### LINKED ISSUE #6452: `addon.tab.editorMode` should not rely on the URL
### Description

`addon.tab.editorMode` always returns `editor` if the URL does not change while on fullscreen.

**This is not the case when running Scratch Addons (in dev mode) on scratchfoundation.github.io as there's a special if statement for that that uses Redux to check for fullscreen.**

### Steps to reproduce

1. Go to https://scratch.mit.edu/projects/editor/ while logged out
2. Go full screen
3.  `addon.tab.editorMode` still returns `editor`

The easiest fix is probably to use Redux

### Environment

Any browser

Samq64:
Resolves #6452

Tested on Chromium on the website and localhost.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -216,14 +216,18 @@ export default class Tab extends Listenable {
    * @type {?string}
    */
   get editorMode() {
-    // Redux isn't really necessary unless entering fullscreen from https://scratch.mit.edu/projects/editor/
-    const mode = this.redux.state?.scratchGui?.mode;
-    const projectNotAvailable = this.redux.state?.preview?.projectNotAvailable;
-    if (projectNotAvailable || !mode) return null;
-    if (mode?.showBranding) return "embed";
-    if (mode.isFullScreen) return "fullscreen";
-    if (mode.isPlayerOnly) return "projectpage";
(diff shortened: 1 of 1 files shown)


WorldLanguages:
We should remember that Redux isn't strictly necessary here. It's only needed to support this URL: https://scratch.mit.edu/projects/editor/

If we have trouble using Redux someday, we will come back to this

WorldLanguages:
Can we change this to only use Redux when it's strictly necessary? I don't really like the idea of engaging with Redux for the most basic APIs when it's not really necessary.

These are the only situations where we need to use Redux, I believe:

1. When running scratch-gui locally (`isScratchGui`)
2. If `location.pathname === "/projects/editor" || location.pathname === "/projects/editor/"`

Samq64:
This PR also fixes 404 project pages not returning null. Is it worth keeping that?

mxmou (approved):

WorldLanguages (approved):

WorldLanguages:
Note that this did NOT change whether the `.sa-body-editor` class is added to the body or not. I'm going to reopen the issue.

PR #7589, opened 2024-06-25
Add dynamic enable, disable, and settings changes to cat-blocks

TheColaber:
I tried fixing the "Watch Mouse Pointer" setting since it's pretty buggy in RTL and when using zoom. I couldn't figure it out so hopefully someone smarter than me (*cough* @GarboMuffin) can fix it or maybe in another PR in the future.

Also, should this addon still say it uses the "scratch-blocks" library in the manifest?

PR #5879, opened 2023-04-04
Reporter wall thickness / Flat Hats

TheYande:
### Resolves
prevents super chonky reporters taking up space (below is probably overexaggerating but i encounter this problem alot) 
image

### Changes

add the setting to custom block shape.

image
Flat Hats! (from discord suggestion i made [https://discord.com/channels/806602307750985799/1043655310792982619](url) )

### Reason for changes

big blocks take up space and could possibly cover up other scripts

### Tests

Tested on OperaGX

PR #7781, opened 2024-09-04
Add option to `custom-block-shape` for flat hat blocks

### LINKED ISSUE #5886: Flat Hats
### Why this would be helpful

#5879 makes the bump on the hat blocks disappear!
more of a style change than fixing or resolving a problem. 

### My suggestion

Flat hat makes the hat like an upside down version of cap blocks for sandwich looking scripts

### Possible alternatives

_No response_

### Additional context

this is worded very very badly, probably because this is my 2nd addon and issues is brand new to me

TheColaber:
Resolves #5886. Exceeds #5879

This commit shows the changes to custom block shape: https://github.com/ScratchAddons/ScratchAddons/commit/0597c1255ec9872f99691ad87677ca470e248d98

This PR depends on #7589. That PR should get merged first.

This also changes the flat presets to use this new flat setting. May need transition code to enable this setting for old users?

image


Samq64:
Shouldn't this be separated from the cat block changes (#7589)?

TheColaber:
they indirectly affect each other since they both effect block constants. try enabling cat blocks and then dynamically enabling and disabling this addon. you'll see floating ears. when the pr mentioned is merged, no floating ears :)

Samq64:
#8556 rewrote the userscript for modern Blockly only keeping this one for backwards compatibility.

PR #7644, opened 2024-07-12
blocks2image: image rendering changes by Jazza

WorldLanguages:
Previously part of PR #6830

Related PR: #5604

Relevant comments:
https://github.com/ScratchAddons/ScratchAddons/pull/6830#issuecomment-2182024814
https://github.com/ScratchAddons/ScratchAddons/pull/6830#issuecomment-2196034417

### Tests

Jazza said:
> I've tested it very extensively and it all works.

PR #5604, opened 2023-01-11
Fix blocks2image height crop issue

WorldLanguages:
I'm making this PR on behalf of @dos-pasos

Resolves #5165

### Changes

See https://github.com/ScratchAddons/ScratchAddons/discussions/5601

PR #7783, opened 2024-09-05
`blocks2image`: More reliable offset calculations

Samq64:
Closes #7644

### Changes

Use `getBBox()` to determine `translateY` when exporting blocks. Also adds the tiny bit of padding from #7644 so the block borders are never clipped.

### Reason for changes

Better addon compatibility

### Tests

Tested on Chromium with custom block colours, custom block shape and cat blocks. 

~~Export all is the same as before and still needs work.~~ Done!

Related: https://github.com/ScratchAddons/ScratchAddons/pull/5604#discussion_r1742998889


WorldLanguages (approved):
I did not test this

Samq64:
The console complains about the iframe not having an src but that was not introduced here.

Samq64:
@TheColaber Could you please review this since you have a couple open PRs waiting for it?

PR #7790, opened 2024-09-08
New addon: `svg-path-editor`

Masaabu:
### Changes
Created an addon to add a button to change the path of a shape with text in the Costume Editor.

https://github.com/user-attachments/assets/6ab6a290-9a24-43c0-b34b-7f83e1ab1486

### Reason for changes
I sometimes edit SVG path and have always had to export the costume, edit it, and upload it again, but I created this addon to make that easier.

### Tests
Tested with Chrome for Windows

### DIFF
--- a/addons-l10n/en/svg-path-editor.json
+++ b/addons-l10n/en/svg-path-editor.json
@@ -1,0 +1,4 @@
+{
+    "svg-path-editor/save": "Save",
+    "svg-path-editor/label": "Edit Path"
+}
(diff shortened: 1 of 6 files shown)


TheColaber:
I think having an issue to discuss this idea first would have been better. I am not sure how useful this is to the average user.

DNin01:
Ideally SVG paths would be adjusted with a graphical user interface. This addon seems like it would be used for very specific use cases.

But I think you're onto something. For programmers, it might be better if an addon let you edit the XML of the _entire costume SVG_ with this addon, not just path data.

mxmou:
Scratch already has a UI for editing paths: the Reshape tool. An addon that adds more capabilities to that might be more useful.

PR #7796, opened 2024-09-10
Better comment capturing

DNin01:
### Changes

Improves the regular expressions that catch comments that include "Scratch Addons". It now catches a few different misspellings of "Scratch" and matches "add-on" without an "s" and/or with a hyphen or whitespace in the middle. There's also now a regex test for "web store" or "webstore" as in the Chrome Web Store.

### Tests

I did not test the code on Scratch (I don't want to post something by accident), but I still made sure the updated code works.

Is this good, @WorldLanguages?

### DIFF
--- a/content-scripts/cs.js
+++ b/content-scripts/cs.js
@@ -783,17 +783,16 @@ const isProject = pathArr[0] === "projects";
 const isForums = pathArr[0] === "discuss";
 
 if (isProfile || isStudio || isProject || isForums) {
-  const removeReiteratedChars = (string) =>
-    string
-      .split("")
-      .filter((char, i, charArr) => (i === 0 ? true : charArr[i - 1] !== char))
-      .join("");
-
   const shouldCaptureComment = (value) => {
-    const trimmedValue = value.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, ""); // Trim like scratchr2
-    const limitedValue = removeReiteratedChars(trimmedValue.toLowerCase().replace(/[^a-z]+/g, ""));
-    const regex = /scratchadons/;
-    return regex.test(limitedValue);
+    // Catch references to Scratch Addons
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Did you test it with a tool such as https://regex101.com/ or similar?

DNin01:
> Did you test it with a tool such as https://regex101.com/ or similar?

I tested it with my browser's console.

DNin01:
Works like a charm, and catches a few things that weren't caught before. (Anything highlighted orange in the picture below will trip it.)

Regex matches

On second thought, I don't know, though, if you might want me to put back the function that removes reiterated characters? e.g. `Scraatch aaaddons` will _not_ match anymore.

WorldLanguages:
The purpose of our message is to provide useful context on why the comment is about to get blocked.

Unfortunately, there's not a good way to know which comments would get blocked and which ones would not, and it can also change from time to time.

WorldLanguages:
Do you actually get muted for saying "web store" or providing a link to it?

DNin01:
> The purpose of our message is to provide useful context on why the comment is about to get blocked.

I know.

> Do you actually get muted for saying "web store" or providing a link to it? 

I don't know. Surely it can't be allowed, though.

WorldLanguages:
image

It's not blocked, so I don't think we should add it.

WorldLanguages:
> By the way, I removed the function that removes reiterated characters in this PR. I don't know if maybe I should put that back? e.g. `Scraatch aaaddons` will _not_ match anymore because I have removed that function. But I'm not sure how often comments with accidental repeated characters happen.

I think the Scratch filter also does an analysis with all repeating characters removed, so I would try to keep that if possible

DNin01:
I'm sorry, the original function does actually capture `add-ons` with a hyphen in the middle.

I think I'll open a new PR.

DNin01:
New PR: #7907

PR #7815, opened 2024-09-16
debugger: better chart colors and dark mode support

### LINKED ISSUE #7814: Debugger preformace tab doesn't support dark mode
### Description

The debugger addon's performance tab does not properly support dark mode.

### Steps to reproduce

1. Enable `debugger` and `editor-dark-mode`
2. Open the editor
3. Open the debugger and select the performance tab

screenshot-2024-09-16-19-26-05

### Environment

Browser: Chromium
Operating system: Linux

mxmou:
Resolves #7814

### Changes

* Changes the default chart colors: the line color is now the same as the debugger block color
* Makes the color of chart titles, ticks and grid lines affected by dark mode (titles already had the right color if website dark mode was also enabled)

image

### Tests

Tested on Edge and Firefox.


Samq64 (approved):

addons/editor-dark-mode/debugger.js
```diff
@@ -0,0 +1,28 @@
...
+      const options = chart.options;
+      options.scales.x.ticks.color = options.scales.y.ticks.color = color;
+      options.scales.x.grid.tickColor = options.scales.y.grid.tickColor = color;
+      options.scales.x.grid.borderColor = options.scales.y.grid.borderColor = color;
+      options.scales.x.grid.color = options.scales.y.grid.color = gridColor;
+      chart.update();
+    }
+  };
+
+  await addon.tab.waitForElement(".sa-performance-tab-content", { markAsSeen: true });
```
WorldLanguages, line 23: Why markAsSeen:true if you're only calling this once per session?

mxmou: This code is based on `dark-www/scratchstats.js`, which was deleted when the scratchstars addon was removed. It also used waitForElement in this way and I thought there was a reason for that.

WorldLanguages (approved):
Tested in Chromium

addons/editor-dark-mode/debugger.js
```diff
@@ -0,0 +1,28 @@
...
+      const options = chart.options;
+      options.scales.x.ticks.color = options.scales.y.ticks.color = color;
+      options.scales.x.grid.tickColor = options.scales.y.grid.tickColor = color;
+      options.scales.x.grid.borderColor = options.scales.y.grid.borderColor = color;
+      options.scales.x.grid.color = options.scales.y.grid.color = gridColor;
+      chart.update();
+    }
+  };
+
+  await addon.tab.waitForElement(".sa-performance-tab-content");
```
WorldLanguages, line 23: I don't know what to think about this pattern... It's not ideal for performance. If the debugger addon is disabled, it will never be found, and this selector will be searched on any DOM change.

WorldLanguages: A different solution, even if ugly, would be better in my opinion.

For example, the debugger addon could run the following after inserting this element to the document:

```js
window.dispatchEvent(new CustomEvent("ScratchAddonsDebuggerAppendedPerformanceTabContentToDOM"))
```

With this change, the waitForElement can be changed to a single querySelector() followed by an event listener

addons/editor-dark-mode/debugger.js
```diff
@@ -0,0 +1,27 @@
...
+      const options = chart.options;
+      options.scales.x.ticks.color = options.scales.y.ticks.color = color;
+      options.scales.x.grid.tickColor = options.scales.y.grid.tickColor = color;
+      options.scales.x.grid.borderColor = options.scales.y.grid.borderColor = color;
+      options.scales.x.grid.color = options.scales.y.grid.color = gridColor;
+      chart.update();
+    }
+  };
+
+  window.addEventListener("saDebuggerPerformanceTabShown", updateCharts);
```
WorldLanguages, line 23: @mxmou I think you forgot the following case here:

```js
if (document.querySelector(".sa-performance-tab-content")) updateCharts();
```

mxmou: PR with fix: #7865

PR #7822, opened 2024-09-17
search-sprites: move search box to the bottom

### LINKED ISSUE #4710: Move the sprite search box to the bottom
My suggestion is to move the sprite search box added by the `search-sprites` addon to the bottom edge. This would put it under the add sprite button, where you get the bonus of not having that button obscure sprites. Or alternatively, there could just be a separate addon to help with that.

Old example:

Search field with text "Search sprites" next to add sprite button

New example:

Scratch's sprite library search bar but next to the add sprite button

(Not to the bottom where you have to scroll down. Just have it float next to the add sprite button.)

mxmou:
Resolves #4710

### Changes

The current sprite search bar uses some vertical space and doesn't look very good. This PR replaces it with a floating search icon next to the add sprite button:
image

It also adds padding to the sprite list so that the icon doesn't block the bottom row of sprites when you scroll to the bottom:
image

The icon expands into a search bar when hovered:
image

If the search bar contains text, it won't collapse, even if it loses focus:
image

### Tests

Tested on Edge and Firefox.


WorldLanguages:
> It also adds padding to the sprite list so that the icon doesn't block the bottom row of sprites when you scroll to the bottom:

Very clever!

WorldLanguages:
Maybe we could add this the Recommended tag

Also, want to add `latestUpdate` to the manifest?

Samq64 (approved):

addons/search-sprites/search-bar.css
DNin01, line 1: Works fine, except that in small stage mode, there's not enough space for the search bar to be useable.

Crammed search bar in the sprite list

DNin01 (commented):
My update notice suggestion:
```json
"latestUpdate": {
  "version": "1.40.0",
  "temporaryNotice": "The search bar is now below the sprite list, next to the add button."
},
```

Also, you can add yourself to the credits if you want.

addons/search-sprites/search-bar.css
WorldLanguages: Nice catch, I forgot about small stage mode

mxmou:
Small stage mode now looks like this:
image

addons/editor-compact/userstyle.css
```diff
@@ -208,6 +210,35 @@ button[class*="action-menu_more-button"] {
...
+[dir="ltr"] .sa-search-sprites-container {
+  right: 4rem;
+}
+[dir="rtl"] .sa-search-sprites-container {
+  left: 4rem;
+}
+.sa-search-sprites-container:hover,
+.sa-search-sprites-container:focus-within,
+.sa-search-sprites-container:not(.sa-search-sprites-empty) {
+  width: calc(100% - 5rem);
```
DNin01, line 228: What about small stage + compact editor addon?

Crammed compact search bar

Try `max(12rem, calc(100% - 5rem))`.

mxmou: #7846

Issue #6956, opened 2023-12-14
Remix tree button dissapears when the window gets too small, and does not get added back


Jazza-231:
### Description

What should happen: The remix button should appear once the window has enough room, like the vanilla buttons.
What does happen: It doesn't. This also affects the download projecrt button (not merged)project

### Steps to reproduce

1. Open a project with the remix tree addon enabled
2. Resize the window, either by opening dev tools on the right, or actually making your browser window smaller
3. Notice that some buttons have gone, now make the window bigger, notice how vanilla buttons are all back, but SA added buttons dont reappear.

### Demonstration(s) and additional context

https://github.com/ScratchAddons/ScratchAddons/assets/81956724/d9938858-3bb3-4c34-92cd-9d189db8af92

### Environment

Browser: Brave.
Scratch Addons version: Current
Operating system: Windows 11

DNin01:
I wonder if Scratch regenerates this row of buttons every time the layout changes, removing the buttons we added. In that case, we would just have to re-add our buttons as needed.

DNin01:
Scratch also hides some of the buttons when there'd not enough space. We should consider whether we should automatically hide some of our buttons or try to cram them in.

DNin01:
This happens when the window is resized from 768px wide or more to less than 768px or the other way around.

Looks like this is handled at [`src/views/preview/presentation.jsx#L381`](https://github.com/scratchfoundation/scratch-www/blob/7853388e59142221cec2ee404e71b36aca168808/src/views/preview/presentation.jsx#L381) in the source code for the Scratch website.

This issue isn't limited to addons that add buttons. `last-edit-tooltip` is also affected because it works with the share date label, another element that is modified when the screen size dips above or below 768px.

@mxmou, do you happen to know any potential fixes for this?

mxmou:
Removing `reduxEvents` from the `waitForElement` call is enough to fix this in last-edit-tooltip. In remix-tree-button, the `urlChange` listener will need to be replaced with a `waitForElement` loop.

WorldLanguages:
> Removing `reduxEvents` from the `waitForElement` call is enough to fix this in last-edit-tooltip.

Hmm, removing all the redux events optimizations wouldn't be optimal.
We could 1) add a "resize" event that can be added into the `reduxEvents` list or through other parts of the API, or 2) attempt to modify the behavior of waitForElement so that it always queries all selectors again if the window was resized.

PR #7831, opened 2024-09-21
editor-dark-mode: improvements and bug fixes

### LINKED ISSUE #6392: Some elements that use the color purple can't be changed with Editor dark mode and customizable colors
### Description

When you edit the direction with the sprite editor, (This area: 
image , not the block) the direction dial and dragging a sprite border is based off of the new purple look even with the setting to replace those colors with blue.

### Steps to reproduce
The direction dial
1. Open the editor
2. Click on the direction thing near the sprites name
3. Its purple!

Dragging a sprite border
1. Open the editor
2. Drag a sprite/costume
3. The border is purple!
The direction dial:
image
Dragging a sprite border:
image
and 
image

### Environment

Browser:  Opera GX
Scratch Addons version:  1.10.0
Operating system:  Windows

### LINKED ISSUE #6936: Editor menu items don't stay highlighed with dark mode
> The color mode menu being highlighted isn't a bug, but editor dark mode does cause the highlight to disappear when when not hovering over any menu items.
bug.webm

_Originally posted by @Samq64 in https://github.com/ScratchAddons/ScratchAddons/issues/6935#issuecomment-1843594879_

### LINKED ISSUE #6975: editor-dark-mode: costume-editor action icons are incorrectly colored after changing zoom
### Description

In the costume editor, using editor-dark-mode you can set a colour for the buttons like move to front layer etc. When you zoom in enough, scratch moves these to a dropdown "more" menu, and they lose their custom colour both in that menu and when you zoom back out again.

### Steps to reproduce

1. Open costume editor
2. Zoom until the go to front/back layer buttons hide
3. Unzoom and see they are purple

(Make sure the icon colour is not set to purple, i chose TW red)

https://github.com/ScratchAddons/ScratchAddons/assets/81956724/69fb3a8f-2ad0-4155-86f3-1381ceeee303

I seems we dont handle dynamic elements very well in quite a few scenarios

also sorry for switching back to obs, i was a bit confused

### Environment

Browser: 
Scratch Addons version: 
Operating system:

mxmou:
Resolves #6392
Resolves #6936
Resolves #6975

### Addon changes

See commits for details. The most noticeable change is that the code area zoom icons will now finally look good in dark mode:
image

### Addon API changes

New `resizeEvent` option in `waitForElement` for #6975 (#6956 could be fixed in a similar way in the future). It should be set to true if the element may appear or disappear on window resize.

### Tests

Tested on Edge and Firefox.


WorldLanguages:
> New `resizeEvent` option in `waitForElement` for #6975 (#6956 could be fixed in a similar way in the future). It should be set to true if the element may appear or disappear on window resize.

Looks OK to me

Let me know if I understand what you did,
Setting `resizeEvent` to true means it will check whether the element exists (through `document.querySelector`) when the "resize" event is dispatched by the browser. It is equivalent to an imaginary Redux event that triggers on window resize.
But it will still respect the `markAsSeen:true` choice. So this only makes sense for **new elements** that are added to the DOM after a resize. Is this why you say #6956 is related, but not directly fixable by this new addition?

mxmou:
> Let me know if I understand what you did,
> Setting `resizeEvent` to true means it will check whether the element exists (through `document.querySelector`) when the "resize" event is dispatched by the browser. It is equivalent to an imaginary Redux event that triggers on window resize.
> But it will still respect the `markAsSeen:true` choice. So this only makes sense for **new elements** that are added to the DOM after a resize.

Correct. It works the same way as an additional Redux event would, but I made it a separate option because it's unrelated to Redux.

> Is this why you say #6956 is related, but not directly fixable by this new addition? (If you're not sure, it's fine)

Fixing #6956 is slightly more complex because the remix tree button addon doesn't currently use waitForElement - if it did, adding `resizeEvent: true` would be enough.

WorldLanguages (approved):

Samq64 (approved):
I didn't test everything but the 3 issues linked and the dial are fixed.

PR #7834, opened 2024-09-24
Adds text input to opacity slider

qloakongithub:
Adds a text input into the opacity slider.

This builds upon a suggestion made on the Discord server suggesting that a text input be added to the opacity slider. I've added this into the addon.

### Changes

Added a text input to the opacity slider.

### Reason for changes

This helps provide consistency in projects and can help with more exact measures. Plus it's just generally a quality of life thing to have.

### Tests

Tested in Chrome on ChromeOS Version 128.0.6613.133. Only remote issue is that the opacity text input is not aligned like the other text sliders are, but this is miniscule and generally doesn't cause much harm. Another thing that could be seen as a problem is that this is activated alongside the opacity slider and not the text inputs addon, but I find this to be better as it lowers the amount of addons you need on at a time, which could help some people. Otherwise there are no issues that I can spot.

### DIFF
--- a/addons/opacity-slider/userscript.js
+++ b/addons/opacity-slider/userscript.js
@@ -6,10 +6,10 @@ export default async function ({ addon, console, msg }) {
   let prevEventHandler;
   let handleClickOffset;
   let element;
-  let labelReadout;
   let saOpacityHandle;
   let saOpacitySlider;
   let saOpacitySliderBg;
+  let inputOpacity;
 
   const getColor = () => {
     let fillOrStroke;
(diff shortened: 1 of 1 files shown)


iqnite:
I don't think this is an optimal solution. The problem this PR solves is not the opacity slider not having a number input, it's the `color-inputs` addon not adding a number input to the opacity slider. This PR creates a lot of duplicate code and ignores whether the `color-inputs` addon is enabled. Maybe you should try making the `color-inputs` addon work on the opacity slider instead?

Samq64:
In its current state this PR does not change anything, so I am going to close it. Feel free do open a new PR with @iqnite's suggestions.

PR #7843, opened 2024-10-05
`player-thumb`: Fullscreen loading support and bug fixes

Samq64:
### Changes

- Adds fullscreen mode loading support and fixes the border size
- Disables green flag pointer events until the project is loaded
- Fixs `fullscreen`'s override styles from leaking outside fullscreen mode
- ~~Mostly fixed thumbnail being removed when exiting the editor before the project is started (but sometimes the Redux event fires too early)~~
- Only adds transparency to the loader background once the thumbnail has been injected
- Removed embed support because it's buggy

### Tests

Tested on Chromium, there is a race condition causing the editor menubar to be visible on the loading screen before `.sa-body-editor` is added.


WorldLanguages:
If there are no rejections, I will merge this before release, unless I forget.

Samq64:
I can keep just the `addon.json` and `fullscreen` selector changes for this release if there isn't enough time.

WorldLanguages:
@Samq64 I will merge this. Is that fine?

Samq64:
> @Samq64 I will merge this. Is that fine?

@WorldLanguages Sure

PR #7847, opened 2024-10-07
Change the order and appearance of popup tabs

### LINKED ISSUE #7842: Make "Addons" the first (far left) popup tab
### Why this would be helpful

Scratch Addons' extension action's popup has three tabs, "Messaging", "Games", and "Addons". Back in our beginning, the first and default selected tab was "Messaging". This is understandable when you consider our Scratch Messaging roots.

But overtime, we've shifted focus away from the messaging aspect, and now the Messaging tab is no longer always selected automatically (#4885) nor is it the default (#6781). This makes sense for our current mission — to provide useful and customizable addons for the Scratch _website_.

### My suggestion

One more thing we could do is move the Addons tab to the far left of the bar, like this:

Tabination with Addons, Messaging, and Games

Being the default tab now, it's more sensible and expected to be the first one. Plus, if you toggle the addons that add Messaging and Games tabs, the Addons tab will remain in the same position, making it more consistent.

### Possible alternatives

We could also rethink the tab bar entirely. (...Are we ever going to add more tabs?)

### Additional context

_No response_

mxmou:
Resolves #7842

### Changes

image
image

### Reason for changes

Addons is the default tab, so it makes sense to make it the first one. The visual changes make the tabs more similar to the settings page categories. Another reason to change the design is that users will be less likely not to notice that the order is different.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -68,17 +68,16 @@ const vue = new Vue({
 
 let manifests = null;
 // If order unspecified, addon goes first. All new popups should be added here.
-const TAB_ORDER = ["scratch-messaging", "cloud-games", "__settings__"];
+const TAB_ORDER = ["__settings__", "scratch-messaging", "cloud-games"];
 
 chrome.runtime.sendMessage("getSettingsInfo", (res) => {
   manifests = res.manifests;
-  const popupObjects = Object.keys(res.addonsEnabled)
+  let popupObjects = Object.keys(res.addonsEnabled)
     .filter((addonId) => res.addonsEnabled[addonId] === true)
     .map((addonId) => manifests.find((addon) => addon.addonId === addonId))
     // Note an enabled addon might not exist anymore!
     .filter((findManifest) => findManifest !== undefined)
     .filter(({ manifest }) => manifest.popup)
-    .sort(({ addonId: addonIdB }, { addonId: addonIdA }) => TAB_ORDER.indexOf(addonIdB) - TAB_ORDER.indexOf(addonIdA))
     .map(
       ({ addonId, manifest }) =>
         (manifest.popup._addonId = addonId) &&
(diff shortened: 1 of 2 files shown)


webpages/popup/style.css
```diff
@@ -92,18 +90,31 @@ body {
...
+}
+.popup-name::before {
+  content: "";
+  display: block;
+  position: absolute;
+  left: 10px;
+  right: 10px;
+  bottom: 0;
+  height: 4px;
+  background-color: var(--orange);
```
Samq64, line 104: I think this should be `--brand-orange` along with the category selector to keep consistency.
```suggestion
  background-color: var(--brand-orange);
```

Samq64:
This is actually quite similar to #7214 (The tabs do also expand with this PR).

screenshot-2024-10-07-14-00-52

webpages/popup/index.js
```diff
@@ -68,17 +68,16 @@ const vue = new Vue({
 
 let manifests = null;
 // If order unspecified, addon goes first. All new popups should be added here.
```
DNin01, line 70: This comment can be changed because the Addons tab is always the first now.
```suggestion
// All new popups should be added here.
```

DNin01 (approved):
I like it.

Samq64 (approved):

webpages/popup/index.js
mxmou: I think the comment is still correct: if an addon isn't included in the TAB_ORDER array, it will become the first tab.

webpages/popup/style.css
mxmou: I don't like the current approach of arbitrarily changing some elements to blue in prereleases but not others, but I don't think changing it belongs in this PR.

webpages/popup/index.js
mxmou: This happens if `scratch-messaging` is removed from the array:
image

DNin01: Oh, I see.

PR #7870, opened 2024-10-17
editor-theme3: Apply block colors to icons in "Make a block" popup

### LINKED ISSUE #4036: editor-theme3 does not affect colors for argument/label buttons
### Description

image
The color does not display correctly, the color should be 5F49D8 for all the blocks, not only the edges of the central block, but the others blocks displays with FF6680, the default color.

### Repoduction steps

1. Change the color with the Addon that  does that.
2. Go to Scratch.
3. Click on "Make a block"

### Screenshots

image

### Environment

- Browser: Chrome: 96.0.4664.110 (Build oficial) (64 bits) (cohort: Stable)
- Scratch Addons version: 1.23.2 
- Operating system: Windows

Thanks for reading and sorry for my English!

pumpkinhasapatch:
Resolves #4036
Resolves [Scratch Addons Discord -> #bug-reports -> Custom block color not applied at make a block prompt buttons](https://discord.com/channels/806602307750985799/1213678804203995206)

### Changes

This is a small change to the Customizable block colors (editor-theme3) addon. It applies custom block colors to the "Add a input/label" button block-like icons in the "Make a block" popup modal by editing the icon's vector svg data.

### Reason for changes

In the latest version of Scratch Addons as of writing, custom block colors do not affect these icons in the "Make a block" modal and they appear in Scratch's default pink custom block color which is inconsistent. This pull request changes the colors of the icons to make them match the actual rendered blocks.

| Make a block popup currently | Make a block popup with this fix |
| - | - |
|  | Screenshot from 2024-10-19 06-08-46 |

### Tests

Tested on Chromium 129 with Scratch Addons v1.40.0-pre.

Try it yourself: [Download addon fork](https://github.com/pumpkinhasapatch/ScratchAddons/archive/refs/heads/make-a-block-icon-colors.zip) -> [Installation instructions](https://github.com/ScratchAddons/ScratchAddons#installation)

This has been tested with different block colors and the "Text Color" addon options like "Black" or "Colored on white/black background", like the color presets below.

| High Contrast blocks preset | Editor dark mode + Black preset |
| - | - |
Screenshot from 2024-10-19 06-06-14 | Screenshot from 2024-10-19 06-07-45 |

### DIFF
--- a/addons/editor-theme3/theme3.js
+++ b/addons/editor-theme3/theme3.js
@@ -622,7 +622,7 @@ export default async function ({ addon, console, msg }) {
     while (true) {
       // Wait until "Make a block" popup is opened and icon elements are created
       const iconElement = await addon.tab.waitForElement("[class^=custom-procedures_option-icon_]", {
-          markAsSeen: true,
+        markAsSeen: true,
       });
       // Get img.src, remove data:image... header, then atob() decodes base64 to get the actual <svg> tags.
       let svg = atob(iconElement.src.replace(uriHeader, ""));
--- a/addons/editor-theme3/theme3.js
+++ b/addons/editor-theme3/theme3.js
@@ -629,14 +629,15 @@ export default async function ({ addon, console, msg }) {
 
       // Find and replace the default color codes in the svg with our custom ones
       // Placeholder values are used to prevent hex codes replacing each other (see PR #7545 changes)
-      svg = svg.replace("#ff6680", "%primary%") // Primary block color
-      .replace("#ff4d6a", "%inner%") // Inside empty boolean/reporter input slots
-      .replace("#f35", "%outline%") // Border around edges of block
-      .replace("#fff", "%labeltext%") // Text color for "Add a label" icon
-      .replace("%primary%", primaryColor(myBlocksCategory))
-      .replace("%inner%", isColoredTextMode() ? fieldBackground(myBlocksCategory) : tertiaryColor(myBlocksCategory))
-      .replace("%outline%", tertiaryColor(myBlocksCategory))
-      .replace("%labeltext%", isColoredTextMode() ? tertiaryColor(myBlocksCategory) : uncoloredTextColor());
+      svg = svg
+        .replace("#ff6680", "%primary%") // Primary block color
+        .replace("#ff4d6a", "%inner%") // Inside empty boolean/reporter input slots
+        .replace("#f35", "%outline%") // Border around edges of block
+        .replace("#fff", "%labeltext%") // Text color for "Add a label" icon
+        .replace("%primary%", primaryColor(myBlocksCategory))
+        .replace("%inner%", isColoredTextMode() ? fieldBackground(myBlocksCategory) : tertiaryColor(myBlocksCategory))
+        .replace("%outline%", tertiaryColor(myBlocksCategory))
+        .replace("%labeltext%", isColoredTextMode() ? tertiaryColor(myBlocksCategory) : uncoloredTextColor());
 
       //console.log(svg);
       iconElement.src = uriHeader + btoa(svg); // Re-encode image to base64 and replace img.src
--- a/addons/editor-theme3/theme3.js
+++ b/addons/editor-theme3/theme3.js
@@ -615,6 +615,29 @@ export default async function ({ addon, console, msg }) {
     Object.assign(originalColors, newColors);
   };
 
+  (async () => {
+    // Custom colors for "Add an input/label" block icons in the "Make a block" popup menu, by pumpkinhasapatch
+    const uriHeader = "data:image/svg+xml;base64,";
+    const myBlocksCategory = categories[9];
+    while (true) {
+      // Wait until "Make a block" popup is opened and icon elements are created
+      const iconElement = await addon.tab.waitForElement("[class^=custom-procedures_option-icon_]", {
+          markAsSeen: true,
+      });
+      // Get img.src, remove data:image... header, then atob() decodes base64 to get the actual <svg> tags.
+      let svg = atob(iconElement.src.replace(uriHeader, ""));
+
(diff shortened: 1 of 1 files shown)


addons/editor-theme3/theme3.js
```diff
@@ -615,6 +615,29 @@ export default async function ({ addon, console, msg }) {
...
+        markAsSeen: true,
+      });
+      // Get img.src, remove data:image... header, then atob() decodes base64 to get the actual <svg> tags.
+      let svg = atob(iconElement.src.replace(uriHeader, ""));
+
+      // Find and replace the default color codes in the svg with our custom ones
+      svg = svg.replace("#ff6680", primaryColor(myBlocksCategory));
+      svg = svg.replace("#ff4d6a", secondaryColor(myBlocksCategory));
+      svg = svg.replace("#f35", tertiaryColor(myBlocksCategory));
+      svg = svg.replace("#fff", uncoloredTextColor()); // Text color for "Add a label" icon
```
mxmou, line 634: This might break with specific combinations of colors - see #7545.

pumpkinhasapatch: Oh I see, the color codes might replace each other if they contain similar values, like if one of the custom block colors starts with `#fff` then half of it would get replaced by the text color. It could use placeholder names like your PR so that doesn't happen.

The proper way to edit svg colors is by using class and style tags, but that's more work that doesn't need to be done unless something else changes these block icons for some reason.

mxmou: Using placeholders is probably a good enough solution.

addons/editor-theme3/theme3.js
```diff
@@ -615,6 +615,35 @@ export default async function ({ addon, console, msg }) {
     Object.assign(originalColors, newColors);
   };
 
+  (async () => {
+    // Custom colors for "Add an input/label" block icons in the "Make a block" popup menu, by pumpkinhasapatch
+    const uriHeader = "data:image/svg+xml;base64,";
+    const myBlocksCategory = categories[9];
```
mxmou, line 621: This will break if any categories above My Blocks are added or removed in the future. `myBlocksCategory` should be defined as a constant at the top of the file and then referenced in the `categories` array, like the existing `extensionsCategory` and `saCategory`.

addons/editor-theme3/theme3.js
```diff
@@ -615,6 +615,35 @@ export default async function ({ addon, console, msg }) {
...
+    while (true) {
+      // Wait until "Make a block" popup is opened and icon elements are created
+      const iconElement = await addon.tab.waitForElement("[class^=custom-procedures_option-icon_]", {
+        markAsSeen: true,
```
mxmou, line 625:
```suggestion
        markAsSeen: true,
        reduxEvents: ["scratch-gui/custom-procedures/ACTIVATE_CUSTOM_PROCEDURES"],
        reduxCondition: (state) => state.scratchGui.editorTab.activeTabIndex === 0 && !state.scratchGui.mode.isPlayerOnly,
```
These options allow `waitForElement` to apply optimizations.

addons/editor-theme3/theme3.js
```diff
@@ -615,6 +615,35 @@ export default async function ({ addon, console, msg }) {
...
+        .replace("#ff6680", "%primary%") // Primary block color
+        .replace("#ff4d6a", "%inner%") // Inside empty boolean/reporter input slots
+        .replace("#f35", "%outline%") // Border around edges of block
+        .replace("#fff", "%labeltext%") // Text color for "Add a label" icon
+        .replace("%primary%", primaryColor(myBlocksCategory))
+        .replace("%inner%", isColoredTextMode() ? fieldBackground(myBlocksCategory) : tertiaryColor(myBlocksCategory))
+        .replace("%outline%", tertiaryColor(myBlocksCategory))
+        .replace("%labeltext%", isColoredTextMode() ? tertiaryColor(myBlocksCategory) : uncoloredTextColor());
+
+      //console.log(svg);
```
mxmou, line 642:
```suggestion
```

addons/editor-theme3/theme3.js
pumpkinhasapatch: Thanks for the feedback! I have not learned about redux, but I guess this sort of unloads the addon and stops checking for new icon elements when on the project page or fullscreen mode where it is not possible to open the "Make a block" modal.

pumpkinhasapatch: Can we further optimize this code, so it only checks if there are icons when a modal popup or the "Make a blocks" modal is opened instead of running a waitForElement all the time in the project editor?

addons/editor-theme3/theme3.js
pumpkinhasapatch: One line of unused code I left there for developers to check the svg file for errors before it gets converted to base64 and put back into Scratch. I can delete this if it's not needed.

addons/editor-theme3/theme3.js
mxmou: > Can we further optimize this code, so it only checks if there are icons when a modal popup or the "Make a blocks" modal is created? It would be better than running a waitForElement all the time in the project editor.

The `reduxEvents` option does exactly that! And `reduxCondition` does what you guessed - it tells the `waitForElement` API to only check if the icons exist when inside the code editor.

addons/editor-theme3/theme3.js
mxmou: You can leave it if you think it's useful for development.

addons/editor-theme3/theme3.js
mxmou: Unfortunately the only way to find what these options should be set to look at the Scratch source code (or copy them from other addons that change the same part of the editor).

WorldLanguages (approved):
Nice job!

Issue #7839, opened 2024-09-29
Scratch Notifier: Browser notifications are not shown are heard


hueychen27:
### Description

The browser never sends notifications even when a new message actively pops up, and I think this happens on Firefox only. This addon used to work a long time ago.

### Steps to reproduce

1. Use Firefox
2. Enable "Scratch notifier" addon
3. Enable notifications permission

### Environment

Browser: Firefox 130.0.1 (64-bit)
Scratch Addons version: v1.39.1
Operating system: Debian Sid

Samq64:
If I understand correctly you are able to hear the notifications but not see them? Are you using the Scratch Addons ping sound and double-checked that the system-wide do not disturb is off?

WorldLanguages:
System-wide DND is painful, I wonder if we could add a way to send a test notification when "Scratch Notifier" is enabled for the first time.

hueychen27:
Do not disturb is off as I see chess.com notifications even when inactive, but no notifications or sound are present from the Scratch Notifier. I worded it wrong

hueychen27:
> If I understand correctly you are able to hear the notifications but not see them? Are you using the Scratch Addons ping sound and double-checked that the system-wide do not disturb is off?

Oh no, I worded it wrong. I can't hear nor see notifications even though do not disturb is all off.

hueychen27:
Also, do not disturb is disabled on Scratch Addons itself.

tutacat:
Did you enable notifications? Go to Manage Extensions, then Scratch Addons, then select the Permissions tab, then enable the "Display notifications..." switch.

Samq64:
> Did you allow notifications?

The extension already prompts to allow the permission when enabling the addon or loading a settings file with it enabled.

hueychen27:
It does prompt for the permission, but I already have the permission enabled so it doesn't request for it anymore.

hueychen27:
> Did you allow notifications? This seems to be a newer default change to firefox addons permissions. Go to Manage Extensions, then Scratch Addons, then select the Permissions tab, then enable the "Display notifications..." switch.
> 
> Does the "message count on extension icon" (msg-count-badge) work too?

The message badge does work, and notifications permission was enabled the whole time since I started using the addon (I checked again, and it was still on)

WorldLanguages:
I'm not sure if Scratch Notifier was ever tested after migrating to Manifest V3.

hueychen27:
Uh oh

[...]

PR #7873, opened 2024-10-19
Add beta tag to Scratch Notifier

Samq64:
### Changes

Replaces Scratch Notifier's recommended tag with the beta tag.

### Reason for changes

The addon has been unreliable  since the MV3 transition and it doesn't seem like it's going to be fixed soon.

- #3877
- #7839

### Tests

Tested on Chromium.

### DIFF
--- a/addons/scratch-notifier/addon.json
+++ b/addons/scratch-notifier/addon.json
@@ -117,7 +117,7 @@
     }
   ],
   "versionAdded": "1.0.0",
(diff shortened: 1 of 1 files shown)


Hans5958 (approved):
I agree. Poor Mr. Notifier. I would like to tone down the expectations and the tag seems appropriate.

WorldLanguages (approved):

PR #7877, opened 2024-10-22
Cloud variable kill switch

### LINKED ISSUE #5111: Cloud variables kill switch

There are sometimes reasons you might not want to use Cloud variables, like if you don't care for online multiplayer or you want to opt out of usage data collection present in some projects.

### What this addon does

This addon would add a switch to turn off the Cloud functionality of variables to prevent the project from receiving or sending data to CVs. They would continue to work as regular variables.

(Once you turned CVs off, you would need to refresh the page to turn them back on. This is because you can't normally turn off Cloud variables and turn them on again in the middle of running a project, so this limitation would prevent those projects from doing weird, unexpected things from users exploiting the kill switch.)

### Possible alternatives

Sign out of Scratch or open the editor.

Cloud Variables extension indicator with switch set to off

DNin01:
Resolves #5111

### Changes

Added a new addon called "Cloud variable kill switch". This addon adds an on/off switch next to the Cloud Variables extension chip.

A toggle button on the Cloud Variables extension chip

Clicking it to turn it off disconnects Cloud variables from the Cloud Data server, making them act like normal variables.

To turn them on again, the page must be reloaded. This is because you could exploit the ability to re-enable Cloud variables dynamically to do things you wouldn't normally be able to do like claim occupied player slots in multiplayer games.

### Reason for changes

There are a few reasons a user may want to opt out of Cloud variables on a specific project:
- **Debugging:** Sometimes you want to check for behavior when not connected to Cloud Data.
- **Single player:** Sometimes you might not want to join multiplayer games.
- **Privacy:** Some people don't like Cloud variables that purely record your activity.

Most users also want to still be able to use Cloud variables in general, so the switch is on by default.

### Tests

I have tested this on scratch.mit.edu with Edge 130 / Windows 11.

#### Things to check

- The toggle appears after the Cloud provider has connected
- The toggle does not appear if Cloud Data is unavailable (e.g., you're signed out)
- The toggle is OFF after entering the editor
  - (I think in its current state it just doesn't appear after leaving the editor, close enough)
- Cloud variables work when the switch is ON and don't work when the switch is OFF

### DIFF
--- a/addons/cloud-off/userscript.js
+++ b/addons/cloud-off/userscript.js
@@ -1,0 +1,85 @@
+export default async function ({ addon, console, msg }) {
+  const cloud = addon.tab.traps.vm.runtime.ioDevices.cloud;
+  /** Disconnects the Cloud variables, in response to the user turning off the switch. */
+  function disableCloud() {
+    if (cloud.provider.connection.readyState === WebSocket.OPEN) {
+      // First override the events that trigger a connection retry, then close the connection
+      cloud.provider.connection.onclose = () => {};
+      cloud.provider.connection.onerror = () => {};
+      cloud.provider.connection.close();
+      connected = false;
+    } else {
+      // WebSocket isn't open, we're not going to try to stop it now
+      console.warn("Not disconnecting because WebSocket is closed");
+      setCheckbox(true);
+    }
+  }
+  /** Opens a confirmation dialog to reload the page and reconnect, in response to the user turning on the switch. */
+  function reloadPrompt() {
+    setCheckbox(false);
+    addon.tab.scratchClassReady().then(() => {
+      addon.tab
+        .confirm(msg("reenable"), msg("reloadRequired"), {
+          okButtonLabel: msg("confirm"),
+          cancelButtonLabel: msg("cancel"),
+          useEditorClasses: false, // The toggle switch that opens this dialog is outside the editor
+        })
+        .then((confirmed) => {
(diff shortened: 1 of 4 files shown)


addons/cloud-off/userscript.js
```diff
@@ -0,0 +1,85 @@
...
+    document.querySelector(".sa-toggle-cloud-off input").checked = checked;
+  }
+
+  let connected = true;
+
+  await addon.tab.redux.waitForState((state) => state.scratchGui.projectState.loadingState.startsWith("SHOWING"));
+
+  if (addon.tab.traps.vm.runtime.hasCloudData()) {
+    let checkCnt = 0;
+    while (cloud.provider?.connection.readyState !== WebSocket.OPEN) {
```
DNin01, line 70: Possible optimization: do not enter while loop when signed out and for New Scratchers

iqnite:
Maybe the fact that cloud variables can only be turned back on by reloading the page should be communicated to the user before they turn them off. I'm thinking about things like people losing their progress because they accidentally clicked the switch.

mxmou:
> Maybe the fact that cloud variables can only be turned back on by reloading the page should be communicated to the user before they turn them off. I'm thinking about things like people losing their progress because they accidentally clicked the switch.

I suggest replacing the switch with a button labeled "Turn off". Unlike a switch, this wouldn't imply that it can be turned back on.

PR #7878, opened 2024-10-22
Rephrase some confirmation messages

DNin01:
### Changes and reasons for changes

```diff
- Are you sure you want to dismiss this message from the Scratch Team? It cannot be recovered.
+ Delete this message permanently?
```
This isn't as destructive as other actions.
```diff
- Are you sure you want to share?
+ Are you ready to share your project?
```
This is more encouraging.

### DIFF
--- a/addons-l10n/en/scratch-messaging.json
+++ b/addons-l10n/en/scratch-messaging.json
@@ -15,7 +15,7 @@
   "scratch-messaging/chars-left": "{num, plural, one {1 character left} other {# characters left}}",
   "scratch-messaging/follows": "Follows",
   "scratch-messaging/stMessages": "Messages from the Scratch Team",
-  "scratch-messaging/stMessagesConfirm": "Dismiss this message?",
+  "scratch-messaging/stMessagesConfirm": "Delete this message permanently?",
   "scratch-messaging/studio-invites": "Studio invites",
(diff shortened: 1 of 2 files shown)


Samq64 (approved):

addons-l10n/en/scratch-messaging.json
```diff
@@ -15,7 +15,7 @@
   "scratch-messaging/chars-left": "{num, plural, one {1 character left} other {# characters left}}",
   "scratch-messaging/follows": "Follows",
   "scratch-messaging/stMessages": "Messages from the Scratch Team",
-  "scratch-messaging/stMessagesConfirm": "Are you sure you want to dismiss this message from the Scratch Team? It cannot be recovered.",
+  "scratch-messaging/stMessagesConfirm": "Dismiss this message?",
```
mxmou, line 18: I like the shorter question, but I think the second sentence should stay - it makes it easier to understand why there's a confirmation.

DNin01: What if it said "delete this message"?

mxmou: Maybe "Delete this message permanently?"

mxmou (approved):

PR #7890, opened 2024-10-28
New addon: "Save now" button when updating variables

### LINKED ISSUE #7735: Trigger project to save when showing/hiding variables or lists
### Why this would be helpful

Sometimes, you want to show or hide a variable or list, but because the "Save now" button does not appear, you forget to save it and close the tab as you work on something else.

### How this addon works

This addon triggers the "Save now" button to appear when showing/hiding variables or lists.

### Possible alternatives

No

### Additional context

no

mybearworld:
Resolves #7735

### Changes

Adds a new addon that makes monitor updates trigger a project change.

### Reason for changes

Making sure people don't lose changes.

### Tests

Tested with Edge. It works with:

- Dragging monitors
- Showing/hiding monitors
- Changing a variable's display type
- Changing a variable's value via the slider
- Changing a slider's range
- Adding items to lists
- Removing items from lists
- Editing list items
- Resizing lists
- Importing to a list

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -157,6 +157,7 @@
   "asset-conflict-dialog",
   "remaining-replies",
   "forum-user-agent",
+  "save-updating-variables",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/save-updating-variables/addon.json
+++ b/addons/save-updating-variables/addon.json
@@ -1,0 +1,20 @@
+{
+  "name": "\"Save now\" button when updating variables",
+  "description": "Shows the \"Save now\" button when a variable or list is being edited on the stage.",
+  "tags": ["editor"],
+  "versionAdded": "1.40.0",
+  "userscripts": [
+    {
+      "matches": ["projects"],
+      "url": "userscript.js"
+    }
+  ],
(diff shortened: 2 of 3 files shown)


mybearworld:
Should this be enabled by default? It's not very obtrusive and makes it less likely to lose work.

Samq64:
> Should this be enabled by default? It's not very obtrusive and makes it less likely to lose work.

I personally wouldn't use this addon since I sometimes accidentally trigger a change even without it and don't want to bump old projects to the top of my stuff. Maybe that's just me though.

DNin01:
> Should this be enabled by default? It's not very obtrusive and makes it less likely to lose work.

I see why it could be. One would expect that's how Scratch should work. I have forgotten to save after updating variables and lists before.

Just make sure the save button only appears after a manual change. Variables and lists can also be updated by code, but running code doesn't usually trigger autosaves.

mybearworld:
> Just make sure the save button only appears after a manual change. Variables and lists can also be updated by code, but running code doesn't usually trigger autosaves.

I didn't realize that, feels kind of counter-intuitive to me. Changing that could potentially be part of this addon as well...

WorldLanguages:
> Should this be enabled by default?

I don't think so

Issue #7768, opened 2024-08-30
(released) Tracking modern-blockly scratch-blocks rewrite


GrahamSH-LLK:
A few Google engineers are rewriting scratch-blocks to use modern Blockly. This will likely break a significant number of editor addons, as -blocks is a very old, and quite modified, fork. Below are instructions for building this rewrite locally for testing addons/extension. 

### Modern blockly scratch-blocks setup
```sh
git clone https://github.com/scratchfoundation/scratch-editor.git --branch spork --depth=1 # Git history is large
cd scratch-editor/
npm ci
npm run build
```
Once setup run `npm run start` in `scratch-editor/packages/scratch-gui` to start it on `localhost:8601`.

## Compatibility checklist

> [!NOTE]  
> This list is outdated because new bugs have been discovered after the updated editor went live on https://scratch.mit.edu.

### Enabled by default

* [x] `editor-devtools`
* [x] `find-bar`
* [x] `middle-click-popup`
* [x] `jump-to-def`
* [x] `place-backpack-code-at-cursor`
* [x] `editor-cleanup-plus`

### Recommended

* [x] `editor-number-arrow-keys`
* [x] `hide-flyout`
* [x] `fix-editor-comments`
* [x] `editor-searchable-dropdowns`
* [x] `block-switching`
* [x] `data-category-tweaks-v2`
* [x] `fix-pasted-scripts` (Does nothing, editor doesn't stick pasted scripts to mouse)
* [x] `reorder-custom-inputs`
* [x] `color-picker`
* [x] `folders`

### Featured

* [x] `zebra-striping`
* [x] `load-extensions`
* [x] `custom-zoom`
* [x] `debugger`
* [x] `block-duplicate`
* [x] `editor-comment-previews`
* [x] `block-cherry-picking` (Included in vanilla)
* [x] `hide-new-variables`
* [x] `name-backpack-items`
* [x] `rename-broadcasts`
* [x] `editor-stepping`
* [x] `blocks2image`
* [x] `script-snap`
* [x] `swap-local-global`
* [x] `variable-manager`

### Others

* [x] `number-pad`
* [x] `transparent-orphans`
* [x] `copy-reporter`
* [x] `ctrl-click-run-scripts`
* [x] `no-script-bumping`
* [x] `editor-sounds`
* [x] `editor-extra-keys`
* [x] `editor-random-direction-block`
* [x] `hide-backpack`

### Themes

* [x] `block-palette-icons`
[...]

PR #7892, opened 2024-10-28
Port away from `window.Blockly.getMainWorkspace()`

Samq64:
### Changes

- Ports most addons to use `addon.tab.traps.getWorkspace()` instead of `Blockly.getMainWorkspace()`.
- Ports `reorder-custom-inputs` to use the `getMainWorkspace()` from `addon.tab.traps.getBlockly()`.
- The `Highlighter` class  used by the debugger and `editor-stepping` now takes an `addon` object but `setGlowingThreads()` could take the workspace directly if that's better.

### Reason for changes

After #7768 `window.Blockly` will no longer have `getMainWorkspace()` and it's a change that can be done now.

Some addons re-declare `Blockly` at their scope but I ported those too since `addon.tab.traps.getWorkspace()` is less prone to crashing the editor.

### Tests

Tested the basic functionality of each addon on Chromium on the main website. Most of these addons are still pretty broken on modern Blockly so I didn't test much there.


addons/find-bar/blockly/Utils.js
```diff
@@ -18,11 +18,6 @@ export default class Utils {
     // this._myFlash = { block: null, timerID: null, colour: null };
     this.offsetX = 32;
     this.offsetY = 32;
-    this.navigationHistory = new NavigationHistory();
```
mxmou, line 21: I think it makes sense to keep `NavigationHistory` a separate class - it acts as a namespace for the methods related to navigation. You can add a constructor to it that takes `addon` as a parameter.

mxmou (approved):
Tested on Firefox.

Samq64:
I can undo the ones that use `getBlockly()`'s version if you want to keep the changes to a minimum.

WorldLanguages:
@Samq64 That might be a good idea, actually

In some edge cases, `blocklyObject.getMainWorkspace()` returns a different workspace (there's a few GitHub issues about that), and I'm not sure if it's a good idea to change this behavior unnecessarily. We might be fixing some bugs, creating new ones, or both, it's hard to know.

WorldLanguages:
Related comment: https://github.com/ScratchAddons/ScratchAddons/pull/6227#issuecomment-1599952073

WorldLanguages:
Will not merge this yet

WorldLanguages:
I think this can be merged now.

PR #7896, opened 2024-10-30
editor-dark-mode: new Blockly

mxmou:
### Changes

Makes the necessary changes to editor-dark-mode to support the new Blockly version. This PR doesn't update the code for zoom icons yet - scratch-blocks currently uses Blockly's default icons, which will probably be replaced with Scratch's ones before the new version is released.

### Tests

Tested on Edge and Firefox using both the current version of scratch-gui and the spork branch.

### DIFF
--- a/addons/editor-dark-mode/experimental_editor.css
+++ b/addons/editor-dark-mode/experimental_editor.css
@@ -166,6 +166,7 @@
 [class*="stage-selector_header_"],
 [class*="drag-layer_image_"],
 .blocklyDropDownDiv .goog-slider-horizontal .goog-slider-thumb,
+.scratchColourSlider::-webkit-slider-thumb,
 [class*="asset-panel_wrapper_"],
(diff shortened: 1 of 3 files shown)


Samq64:
The button text in the flyout stays black when high contrast mode is enabled.
screenshot-2024-10-30-12-08-10

Samq64 (approved):

WorldLanguages (approved):

Samq64:
I just noticed that Scratch will have native CSS variables for a fair amount of Blockly stuff:
screenshot-2024-11-01-20-15-13

mxmou:
> Scratch will have native CSS variables for a fair amount of Blockly stuff which could be useful to simply some dark mode styles.

I didn't use most of those yet because the addon still needs to support the current version of Scratch. Some things can indeed be simplified after the update is released.

Issue #7910, opened 2024-11-06
The Forums now use Scratch 3 Blocks


Samq64:
https://scratch.mit.edu/discuss/topic/741323/

![screenshot-2024-11-06-16-24-17](https://github.com/user-attachments/assets/006ff369-3fb6-439f-bee5-cf9ef5bbc438)

Samq64:
Better quoter has a bunch of extra newlines and is missing the flag so the code in `cs.js` is probably not working properly.

```[scratchblocks]
when
clicked
say
Test out Scratchblocks!!!
[/scratchblocks]
```

medlans:
Soo could there be a 2.0 scratchblocks (instead of 3.0 scratchblocks addon) instead then?

mybearworld:
`editor-theme3`'s styles get overriden (except for comments, where only the text color isn't right):

![image](https://github.com/user-attachments/assets/fe2750be-ea62-47f2-b4c6-d2f0ec258250)

PR #7912, opened 2024-11-06
forum-search: Use new blocks

Samq64:
See #7910

### Changes

Changes `forum-search` to use Scratch 3.0 blocks.

### Tests

Tested on Chromium.


CST1229:
real forum scratchblocks also uses your scratch blocks theme (the high contrast theme if it's enabled):
image

mxmou (approved):

PR #7921, opened 2024-11-07
editor-theme3: support new forum blocks

mxmou:
### Changes

Updates editor-theme3's forum CSS.

### Reason for changes

https://github.com/ScratchAddons/ScratchAddons/issues/7910#issuecomment-2461383225

### Tests

Tested on Edge.


Samq64:
The black setting doesn't effect invalid red blocks, but they have a high contrast version in vanilla Scratch. The red colour doesn't need to be customizable but it the addon's high contrast mode should be consistent with Scratch's.

Samq64:
The custom block colours are not applied if the editor was set to high contrast before the addon was enabled.

Also the text colour is not being applied to cat blocks and the face is always dark grey.

mxmou:
It now works if Scratch is set to high contrast mode. The way cat blocks are structured in the DOM makes it harder to apply different text colors to them based on the category - I might do that in a later PR.

Samq64 (approved):

WorldLanguages (approved):

PR #7936, opened 2024-11-13
debugger, editor-stepping, find-bar, jump-to-def: new Blockly

mxmou:
### Changes

Updates "debugger" (including the `addon.tab.addBlock()` API), "running block border", "find bar", and "jump to custom block definition" to work with the new version of Blockly. Includes some changes from #7892.

### Tests


Samq64 (approved):
On modern Blockly running Paper Minecraft nothing seems to update on the stage after switching to any other sprite but it also happens without Scratch Addons.

WorldLanguages (approved):

WorldLanguages:
I suspect this has caused #8026

PR #7941, opened 2024-11-16
Popups: Always load ./popup.html and ./popup.js

DNin01:
Changes the addon manifest structure. Remember to update documentation and schema.

### Changes

Removes the addon manifest keys `popup.html` and `popup.script` and defaults their values to `./popup.html` and `./popup.js` respectively.

### Reason for changes

Popups can only have one HTML and one JS file, and using the filenames `./popup.html` and `./popup.js` is the standard way.

### Tests

Tested in Edge 131 / Windows 11.

### DIFF
--- a/addons/cloud-games/addon.json
+++ b/addons/cloud-games/addon.json
@@ -18,9 +18,7 @@
   "popup": {
     "icon": "../../images/icons/cloud.svg",
     "name": "Games",
-    "fullscreen": true,
-    "script": "popup.js",
-    "html": "popup.html"
+    "fullscreen": true
   },
   "settings": [
     {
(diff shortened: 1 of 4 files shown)


Samq64 (approved):
This seems like an unnecessary change but it works as expected.

mxmou (approved):

DNin01:
These keys should also be removed from the manifest schema.

Samq64:
Should these schema changes be removed from https://github.com/ScratchAddons/manifest-schema/pull/74 or should both be merged as-is?

WorldLanguages:
@Samq64 Merged!

PR #7949, opened 2024-11-21
middle-click-popup: new Blockly

mxmou:
### Changes

Makes "insert blocks by name" compatible with modern Blockly.

### Tests

Tested both Scratch versions on Edge and Firefox. The addon still works correctly and is still compatible with theme3.

### DIFF
--- a/addons/middle-click-popup/BlockTypeInfo.js
+++ b/addons/middle-click-popup/BlockTypeInfo.js
@@ -351,10 +351,11 @@ export class BlockTypeInfo {
     else if (block.type === "sensing_of") name = "sensing";
     else if (block.type === "event_whenbackdropswitchesto") name = "events";
     else name = block.styleName_ ?? block.category_;
-    name = {
-      event: "events",
-      data_lists: "data-lists",
-    }[name] || name;
+    name =
+      {
+        event: "events",
+        data_lists: "data-lists",
+      }[name] || name;
 
     return {
       name,
@@ -440,7 +441,7 @@ export class BlockTypeInfo {
         }
         if (
           field instanceof Blockly.FieldLabel ||
-          !Blockly.registry && field instanceof Blockly.FieldVariableGetter
+          (!Blockly.registry && field instanceof Blockly.FieldVariableGetter)
         ) {
           if (field.getText().trim().length !== 0) parts.push(field.getText());
         } else if (field instanceof FieldColour) {
--- a/addons/middle-click-popup/BlockTypeInfo.js
+++ b/addons/middle-click-popup/BlockTypeInfo.js
@@ -304,7 +304,7 @@ export class BlockShape {
       return BlockShape.Round;
     } else if (workspaceBlock.getOutputShape() === 1) {
       return BlockShape.Boolean;
-    } else if (workspaceBlock.hat ?? workspaceBlock.startHat_) {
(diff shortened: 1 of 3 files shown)


addons/middle-click-popup/userscript.js
```diff
@@ -375,8 +423,30 @@ export default async function ({ addon, msg, console }) {
...
-    if (workspace.getGesture(fakeEvent)) {
-      workspace.startDragWithFakeEvent(fakeEvent, newBlock);
+    if (Blockly.registry) {
+      // new Blockly expects a pointerdown event
+      fakeEvent.type = "pointerdown";
+      fakeEvent.pointerType = "mouse";
+      // If the block is being dragged using the mouse or touch, the correct pointerId
+      // needs to be set so that Blockly recognizes the associated pointerup event.
+      // If the block is selected using the keyboard, pointerId will be undefined and a
+      // click will be necessary to end the "drag".
```
Samq64, line 433: This never happens and I don't see any errors in the console.

mxmou: The comment could probably be phrased better - it's referring to using arrow keys and enter to select a block. In that case, e.pointerId will be undefined (because e is a keyboard event), which doesn't cause any errors but I wanted people reading the code to know it might happen.

Samq64:
On modern blackly some hats appear as normal blocks instead of having flat tops but not all of them.

mxmou:
> On modern blackly some hats appear as normal blocks instead of having flat tops but not all of them.

For me they're all hats if theme3 is disabled and normal blocks if it's enabled. Apparently changing the block styles does weird things to hats.

addons/middle-click-popup/BlockTypeInfo.js
```diff
@@ -424,11 +429,24 @@ export class BlockTypeInfo {
...
+          field instanceof Blockly.FieldLabel ||
+          (!Blockly.registry && field instanceof Blockly.FieldVariableGetter)
+        ) {
           if (field.getText().trim().length !== 0) parts.push(field.getText());
-        } else if (field.argType_[0] === "colour") {
+        } else if (field instanceof FieldColour) {
           addInput(new BlockInputColour(inputIdx, fieldIdx));
-        } else if (field.argType_[1] === "number") {
+        } else if (field instanceof FieldNumber) {
           addInput(new BlockInputNumber(inputIdx, fieldIdx, field.text_));
```
Samq64, line 450:
```suggestion
          addInput(new BlockInputNumber(inputIdx, fieldIdx, field.text_ ?? field.getText()));
```

addons/middle-click-popup/BlockTypeInfo.js
```diff
@@ -424,11 +429,24 @@ export class BlockTypeInfo {
...
         } else {
           addInput(new BlockInputString(inputIdx, fieldIdx, field.text_));
```
Samq64, line 452:
```suggestion
          addInput(new BlockInputString(inputIdx, fieldIdx, field.text_ ?? field.getText()));
```

Samq64 (commented):
Number and string inputs are missing their values on modern Blockly.

Samq64 (approved):

WorldLanguages:
@mxmou You can now merge this by bypassing branch protections

PR #7970, opened 2024-12-01
Markdown in forums addon

### LINKED ISSUE #6557: Markdown in forum posts

mybearworld:
Resolves #6557

### Changes

Create a new addon "Markdown in forums" that allows you to use Markdown in the forums by converting the Markdown to BBCode on the fly. This also adds a new library, [marked](https://marked.js.org/).

### Reason for changes

Most platforms use Markdown for their comments/forum posts. Remembering to use BBCode for Scratch can be annoying. Also, Markdown is generally more readable more BBCode and can just be more fun to write.

### Tests

Tested in Edge.

## To Do

This PR is a draft. There are still some important things to be done for this addon to be usable.

- How to handle features that Markdown has but the forums don't?
  - [ ] Tables - currently it just stops a user from posting anything with tables
  - [X] Inline code - this is implemented by just outputting what the user entered - ``the `print` function`` -> ``the `print` function``
  - [X] Task lists - this is implemented by 
- How to handle tags that the forums have but Markdown doesn't?
  - [ ] [u]
  - [ ] [big], [small]
  - [X] [scratchblocks] - this is implemented by a special `scratchblocks` language on code blocks
  - [ ] [color]
  - [ ] [center]
  - [ ] [wp], [wiki], [google], [dictionary]
- Compatibility with quotes
  - [X] A way for quotes to use BBCode - this is implemented by a special `raw-bbcode` language on code blocks
  - [x] Outputting in Markdown when clicking on the Quote button
- [ ] Editing posts where just using `raw-bbcode` would not be ideal
- Compatibility with toolbar
  - [X] Handle toolbar buttons that Markdown supports that are there
  - [ ] Handle toolbar buttons that BBCode has that Markdown doesn't have (notably list item)
  - [X] Handle toolbar buttons that don't exist but Markdown has (e.g. headings)
  - [ ] Keyboard shortcuts
  - [ ] Scratchblocks
- [ ] Compatibility with preview panel
- Compatibility with other addons
  - [ ] `better-quoter`
  - [ ] `forum-toolbar`
  - [x] `forum-id`
  - [x] `dark-www`
  - [x] `material-forum-editor-buttons`
- [ ] Explanation of Markdown syntax?
- [ ] BBCode is provided for editing when 60 second rule applies


libraries/license-info.json
```diff
@@ -59,5 +59,10 @@
   },
   "Roboto": {
     "license": "Apache-2.0"
+  },
+  "marked": {
+    "license": "MIT",
+    "year": "2011-2018, 2018+",
+    "fullname": "MarkedJS (https://github.com/markedjs/), Christopher Jeffrey (https://github.com/chjj/)"
```
mybearworld, line 66: https://github.com/markedjs/marked/blob/a1113e02cac2d295711cb71bb56f4b5bd6e1a70c/LICENSE.md?plain=1#L11-L12

Marked has two copyright declarations – just to make sure, does this work to represent that?

Samq64: It's not ideal since it's displayed like this:
```
Copyright (c) 2011-2018, 2018+ MarkedJS (https://github.com/markedjs/), Christopher Jeffrey (https://github.com/chjj/)
```

mybearworld:
For colors, what I can come up with at the moment is:

- `<color color="#ff0000">red</color>`
  - It's not how it's actually done in Markdown
  - It'd work differently to all other `<html>`-like tags, which might be unintuitive
- `<span style="color: #ff0000">red</span>`
  - It's the way this is properly done in Markdown
  - It'd be a special case of actual HTML in the Markdown code and probably unintuitive to use

Neither of these is great

libraries/license-info.json
DNin01: I think you might have switched them around also.

If the license info structure isn't flexible enough to display the copyright notice neatly, I would suggest creating a copy of the license file with the copyright notice inserted directly so you don't have to work with the `year` and `fullname` properties.

CST1229:
`<font color="#FF0000">red</font>`? though that's a deprecated tag and would also suggest that the `face` and `size` properties would also work (and also not make it clear that this is the color tag)

Issue #7974, opened 2024-12-02
Additional Studio Stats


AeroKoder:
### Why this would be helpful

This addon would be helpful when wanting to know the creation date, and other additional stats, of a studio.

### How this addon works

This addons displays additional stats on the project and studio pages.

### Additional context

Studio Screenshot:
![Screenshot 2024-12-01 170132](https://github.com/user-attachments/assets/03aed6e1-8b3b-4b70-94ee-b3b14af05683)

Samq64:
In the future could you open an issue ahead of time, so the idea can be discussed before you submit a pull request?

AeroKoder:
> In the future could you open an issue ahead of time, so the idea can be discussed before you submit a pull request?

Yes, I just got a GitHub account and already had the code created.

PR #7975, opened 2024-12-02
Created "Additional Stats" addon

AeroKoder:
Resolves [#7974](https://github.com/ScratchAddons/ScratchAddons/issues/7974)

### Changes

Added an addon that will display additional stats on the project and studio pages.
Screenshot 2024-12-01 165935

### Reason for changes

This addon would be helpful when wanting to know the creation date, and other additional stats, of a studio or project.

### Tests

Tested on Edge.

### DIFF
--- a/addons-l10n/en/additonal-stats.json
+++ b/addons-l10n/en/additonal-stats.json
@@ -1,0 +1,5 @@
+{
+    "additonal-stats/date-modifed": "Last modifed on {date}",
+    "additonal-stats/date-created": "Created on {date}",
+    "additonal-stats/followers-per-day": "{followers} followers per day"
+}
--- a/addons/additonal-stats/addon.json
+++ b/addons/additonal-stats/addon.json
@@ -1,0 +1,49 @@
+{
+    "name": "Additonal Stats",
+    "description": "Displays additonal stats on the project and studio pages.",
+    "credits": [
+      {
+        "name": "AeroKoder",
+        "link": "https://scratch.mit.edu/users/AeroKoder/"
+      }
+    ],
+    "settings": [
+      {
+        "name": "Adds the date of when the project was created and last modifed on the project page.",
+        "type": "boolean",
+        "default": true,
+        "id": "projectDates"
+      },
+      {
+        "name": "Adds the date of when the studio was created in the studio page.",
+        "type": "boolean",
+        "default": true,
+        "id": "studioDates"
+      },
+      {
+        "name": "Tells the avarage amount of followers per day in the studio page.",
+        "type": "boolean",
+        "default": false,
+        "id": "studioFollowers"
+      }
+    ],
+    "userscripts": [
+      {
+        "url": "project.js",
+        "matches": ["projects"],
+        "if": { "settings": { "projectDates": true } }
+      },
+      {
+        "url": "studioDates.js",
+        "matches": ["https://scratch.mit.edu/studios/*"],
+        "if": { "settings": { "studioDates": true } }
+      },
+      {
+        "url": "studioFollowers.js",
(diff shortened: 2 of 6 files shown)


DNin01:
By the way, we already have an addon that displays the last edit date (`last-edit-tooltip` or "Shared/edited dates tooltip").

AeroKoder:
> By the way, we already have an addon that displays the last edit date (`last-edit-tooltip` or "Shared/edited dates tooltip").

Oh, I didn't know that existed. So should I remove the last modified date from mine?

Samq64:
Maybe the created date could also be moved to the tooltip and this new addon could be just for studios? The problem with that is project dates won't be accessible to mobile users.

AeroKoder:
> Maybe the created date could also be moved to the tooltip and this new addon could be just for studios? The problem with that is project dates won't be accessible to mobile users.

Ok so I’ll remove the project stats from my addon and add the project creation date to the last-edit-tooltip addon. And should I make the change to the last-edit-tooltip addon in a separate pull request?

AeroKoder:
> I feel followers per day is too obscure of a stat, on most studios it would be close to zero anyway... thoughts?

Yes, that's why I have it turned off by default, but it is still useful.

addons/additional-stats/studioFollowers.js
```diff
@@ -0,0 +1,14 @@
+export default async function ({ addon, console, msg }) {
+    let studio = await (await fetch(`https://api.scratch.mit.edu/studios/${addon.tab.redux.state.studio.id}`)).json();
+
+    const timeSinceCreated = (Date.now() - Date.parse(studio.history.created)) / 86400000;
+    const followersPerDay = studio.stats.followers / timeSinceCreated;
```
Samq64, line 5: This can be larger than the total number of followers if the studio was created on the same day.

Samq64 (commented):
Addon userscripts generally avoid `innerHTML`. Since you want to keep the icons you can use `cloneNode()` and and then change the text with `textContent`.

AeroKoder:
> Addon userscripts generally avoid `innerHTML`. Since you want to keep the icons you can use `cloneNode()` and and then change the text with `textContent`.

Ok, I'll change it to use the `cloneNode()`.

> This can be larger than the total number of followers if the studio was created on the same day.

And I'll fix that problem.

Samq64 (commented):
Could you add `dynamicEnable` and `dynamicDisable` to the addon.json and run `displayNoneWhileDisabled` on your elements so your addon can be turned on and off without a page reload?

AeroKoder:
> Could you add `dynamicEnable`, `dynamicDisable` to the addon.json and pass your elements to `displayNoneWhileDisabled` so your addon can be turned on and off without a page reload?

Added.

Samq64 (approved):
Changing the settings still needs a page reload and generally addons stick to one userscript per page but otherwise this looks good.

addons/additional-stats/addon.json
```diff
@@ -0,0 +1,40 @@
...
+    },
+    {
+      "url": "studioFollowers.js",
+      "matches": ["https://scratch.mit.edu/studios/*"],
+      "if": { "settings": { "studioFollowers": true } }
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "tags": ["community", "studios", "projectPage"],
```
Samq64, line 38:
```suggestion
  "tags": ["community", "studios"],
```

Samq64:
>  And should I make the change to the last-edit-tooltip addon in a separate pull request?

Yes, you should make the changes on a second branch so it's not dependent on this one.

PR #7980, opened 2024-12-06
New addon: `no-sprite-confirm`

### LINKED ISSUE #7966: Disable annoying deletion confirmation
### Why this would be helpful

The default sprite deletion confirmation dialog cannot be disabled.

### My suggestion

Disable the confirmation if scratchaddons is installed, since there is an option to do that anyways.

### Possible alternatives

_No response_

### Additional context

_No response_

Samq64:
Resolves #7966

### Changes

Adds an addon to remove Scratch's sprite deletion confirmation. It has the danger tag but whether it's necessary is debatable since edit  > restore can still be used.

### Reason for changes

Some users really don't like the confirmation.

### Tests

Tested on Chromium and Firefox.

### DIFF
--- a/addons/no-sprite-confirm/addon.json
+++ b/addons/no-sprite-confirm/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Remove sprite deletion confirmation",
+  "name": "Delete sprites without confirmation",
   "description": "Removes the confirmation prompt that asks if you are sure you want to delete a project sprite.",
   "info": [
     {
--- a/addons/no-sprite-confirm/addon.json
+++ b/addons/no-sprite-confirm/addon.json
@@ -3,7 +3,7 @@
   "description": "Removes the confirmation prompt that asks if you are sure you want to delete a project sprite.",
   "info": [
     {
-      "text": "Tip: If you accidentally delete a sprite, you can undo the deletion by clicking Edit in the menu bar then Restore.",
+      "text": "Tip: If you accidentally delete a sprite, costume, or sound, you can undo the deletion by clicking Edit in the menu bar then Restore.",
       "id": "restoretip"
     }
   ],
--- a/addons/no-sprite-confirm/addon.json
+++ b/addons/no-sprite-confirm/addon.json
@@ -1,0 +1,20 @@
+{
+  "name": "Remove sprite deletion confirmation",
+  "description": "Removes the confirmation prompt that asks if you are sure you want to delete a project sprite.",
+  "info": [
+    {
+      "text": "Tip: If you accidentally delete a sprite, you can undo the deletion by clicking Edit in the menu bar then Restore.",
+      "id": "restoretip"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.40.0",
+  "tags": ["editor", "danger"]
+}
--- a/addons/no-sprite-confirm/addon.json
+++ b/addons/no-sprite-confirm/addon.json
@@ -13,6 +13,12 @@
       "matches": ["projects"]
     }
   ],
+  "userstyles": [
+    {
+      "url": "hide.css",
+      "matches": ["projects"]
+    }
+  ],
   "dynamicEnable": true,
   "dynamicDisable": true,
   "versionAdded": "1.40.0",
--- a/addons/no-sprite-confirm/addon.json
+++ b/addons/no-sprite-confirm/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Delete sprites without confirmation",
-  "description": "Removes the confirmation prompt that asks if you are sure you want to delete a project sprite.",
+  "description": "Removes the confirmation prompt that asks if you are sure you want to delete a sprite.",
   "info": [
     {
       "text": "Tip: If you accidentally delete a sprite, costume, or sound, you can undo the deletion by clicking Edit in the menu bar then Restore.",
--- a/addons/no-sprite-confirm/addon.json
+++ b/addons/no-sprite-confirm/addon.json
@@ -22,5 +22,5 @@
   "dynamicEnable": true,
   "dynamicDisable": true,
   "versionAdded": "1.40.0",
-  "tags": ["editor", "danger"]
+  "tags": ["editor", "danger", "featured"]
 }
--- a/addons/no-sprite-confirm/userscript.js
+++ b/addons/no-sprite-confirm/userscript.js
@@ -1,8 +1,8 @@
 export default async ({ addon, console }) => {
-  while (true) {
-    const confirmButton = await addon.tab.waitForElement("[class^='delete-confirmation-prompt_ok-button_']", {
-      markAsSeen: true,
-    });
-    if (!addon.self.disabled) confirmButton.click();
-  }
+  document.body.addEventListener("click", () => {
+    setTimeout(() => {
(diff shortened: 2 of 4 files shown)


addons/no-sprite-confirm/userscript.js
```diff
@@ -0,0 +1,8 @@
+export default async ({ addon, console }) => {
+  while (true) {
+    const confirmButton = await addon.tab.waitForElement("[class^='delete-confirmation-prompt_ok-button_']", {
+      markAsSeen: true,
+    });
```
WorldLanguages, line 5: I think a global "click" listener ends up being more performant

Samq64: So a listener on `body` and then checking the element in it?

WorldLanguages: Sure, it sounds reasonable to make 1 document.querySelector() call for each click made by the user.

addons/no-sprite-confirm/addon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Remove sprite deletion confirmation",
+  "description": "Removes the confirmation prompt that asks if you are sure you want to delete a project sprite.",
+  "info": [
+    {
+      "text": "Tip: If you accidentally delete a sprite, you can undo the deletion by clicking Edit in the menu bar then Restore.",
```
WorldLanguages, line 6: I think it's better if we reuse the `hide-delete-button` tip copied word-by-word

addons/no-sprite-confirm/addon.json
DNin01, line 1: May I suggest adding a warning to the addon? Here's an example:

> [!CAUTION]
> This may result in accidents that cost you your hard work! Pay attention to which sprites you delete, because it's one click and the sprite is gone.

DNin01 (changes requested):
I think this addon should have a warning (`info: [{ type: warning }]`).

addons/no-sprite-confirm/addon.json
Samq64: I don't think it's nessessary since edit > restore still works for the last sprite, but I can add add a warning if you want.

Chameleon-Coder:
Also maybe add the ability to use the old warning?

DNin01:
> Also maybe add the ability to use the old warning?

What's "the old warning"?

addons/no-sprite-confirm/addon.json
DNin01: I mean, shouldn't all dangerous addons have a warning message?

(I don't think we've made any guidelines that dangerous addons should have a warning message because this is only our second dangerous addon, AFAIK, but if we had more dangerous addons, they would probably all have a warning or note of some sort.)

mxmou:
> > Also maybe add the ability to use the old warning?
> 
> What's "the old warning"? 

Probably the confirmation added by the remove-sprite-confirm addon.

addons/no-sprite-confirm/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Remove sprite deletion confirmation",
```
mxmou, line 2:
```suggestion
  "name": "Delete sprites without confirmation",
```

mxmou (approved):

WorldLanguages (approved):

addons/no-sprite-confirm/addon.json
Samq64: I only added the tag to add some kind of confirmation before enabling it. In the future the prompt could be separated from the tag.

addons/no-sprite-confirm/addon.json
```diff
@@ -0,0 +1,26 @@
...
+  "userstyles": [
+    {
+      "url": "hide.css",
+      "matches": ["projects"]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.40.0",
+  "tags": ["editor", "danger"]
```
WorldLanguages, line 25: Should be featured at least for 1.40.0

```suggestion
  "tags": ["editor", "danger", "featured"]
```

Samq64: With the danger tag it will also appear under beta. Is that a problem?

addons/no-sprite-confirm/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Delete sprites without confirmation",
+  "description": "Removes the confirmation prompt that asks if you are sure you want to delete a project sprite.",
```
mxmou, line 3:
```suggestion
  "description": "Removes the confirmation prompt that asks if you are sure you want to delete a sprite.",
```

DNin01:
```suggestion
  "description": "Removes the confirmation prompt that asks if you are sure you want to delete a sprite in a project.",
```

mxmou: It could also say "delete a sprite from a project". But I don't think there's a reason to mention projects - "sprite" can't mean anything that isn't inside a project.

Samq64:
#7983 has a setting to disable the confirmation completly. Would that be prefered?

LIZARD-OFFICIAL-77:
Do you know _what_ everybody has to do? Remove the confirmation when scratchaddons is installed. Regardless of _anything_.

medlans:
> Do you know _what_ everybody has to do? Remove the confirmation when scratchaddons is installed. Regardless of _anything_.

What are you talking about

CST1229:
> Do you know _what_ everybody has to do? Remove the confirmation when scratchaddons is installed. Regardless of _anything_.

once again, changing vanilla scratch behavior without notifying users, ESPECIALLY FOR SOMETHING THAT CAN CAUSE DATA LOSS, is not a good idea. THIS IS WHY DISABLE AUTOSAVE IS MARKED AS DANGEROUS.

mxmou:
> > This is a great addon. I literally can't Scratch without it. But a bug report:- The new 'Delete sprites without confirmation' addon needs a bit more reworking. The sprite pane expands but then it collapses (i.e. it jiggles once). I think you disable the addon by automatically 'clicking' the Yes button. I think you should stop it from showing in the first place.
> 
> (personally I don't know why this would happen, and trying to understand it means we need to research a bit more about when browsers decide to draw another frame and how the event loop works) Even with this issue, I believe waitForElement should be discouraged in the project editor due to performance (and I'm not sure if waitForElement would actually fix the flashing)

I don't think this is what the feedback message is describing, but one situation in which the sprite pane does consistently "jiggle" when deleting a sprite is when the window is small enough that it has a vertical scrollbar.

CST1229:
> (I'm not sure what they mean by the sprite pane jiggling, does anyone know? I thought it meant the confirmation modal was visible for a split second, but that's not really possible because of the CSS we added)

Might be relating to the "Collapsing sprite pane" addon?

DNin01:
> (I'm not sure what they mean by the sprite pane jiggling, does anyone know? I thought it meant the confirmation modal was visible for a split second, but that's not really possible because of the CSS we added)

I might know what they're talking about.

If the window is small enough, scroll bars will appear beside the page. Those scroll bars disappear while the deletion confirmation is popped up.

Our addon automatically clicks the "yes" button when the confirmation message appears, but that's not instant, so we also hide the message so users don't see it. However, the element isn't removed entirely. This means the scroll bars will still disappear briefly while we're skipping the confirmation, causing the page layout to jitter.

https://github.com/user-attachments/assets/8b4b0a2d-39a5-4567-bb05-96264e8bdc2e

In my opinion, this isn't a huge deal, but a fix we could try is preventing Scratch from hiding the scroll bars if they exist. Scratch uses CSS to set `overflow` to `hidden` on `<body>` while a modal is open, preventing the page from scrolling while one is open. And the deletion confirmation is technically a modal.

PR #7983, opened 2024-12-08
New addon: Hold to delete sprites

### LINKED ISSUE #7966: Disable annoying deletion confirmation

DNin01:
Sorta resolves #7966

### Changes

This new addon makes it so you have to press and hold the delete icon next to a sprite to delete it. The addon also has an option to skip the confirmation that would appear after pressing and holding the delete button.

As you hold the button, it will gradually fill up with the color orange until it's completely orange then the delete function is activated.

https://github.com/user-attachments/assets/6a7d783e-15ee-425f-b69f-750ee81f113c

### Reason for changes

Mainly to allow people to delete their sprites without having to confirm while still making it harder to delete by accident.

### Tests and Notes

Tested in Edge 131.
Tested alongside `editor-dark-mode` and `compact-editor`.
Should "skip confirmation" disable the deletion confirmation on right click > delete? It doesn't currently.

### DIFF
--- a/addons/hold-delete/addon.json
+++ b/addons/hold-delete/addon.json
@@ -1,0 +1,42 @@
+{
+  "name": "Hold to delete sprites",
+  "description": "Require the delete button to be held down in order to delete a sprite.",
+  "tags": ["editor"],
+  "credits": [
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
+    },
+    {
+      "name": "Samq64",
+      "link": "https://github.com/Samq64"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "hide-confirmation.css",
+      "matches": ["projects"],
+      "if": {
+        "settings": { "noConfirm": true }
+      }
+    }
+  ],
+  "settings": [
+    {
+      "name": "Skip confirmation",
+      "id": "noConfirm",
+      "type": "boolean",
+      "default": false
+    }
+  ],
+  "versionAdded": "1.40.0",
+  "dynamicDisable": true,
+  "dynamicEnable": true,
+  "updateUserstylesOnSettingsChange": true
+}
--- a/addons/hold-delete/userscript.js
+++ b/addons/hold-delete/userscript.js
@@ -1,0 +1,71 @@
+export default async function ({ addon, console }) {
+  const DELETE_BUTTON_SELECTOR =
+    "[class^='sprite-selector_items-wrapper_'] [class*='sprite-selector-item_delete-button']";
+  let deleteButton;
+  let holding = false;
+
+  document.body.addEventListener("mousedown", async (e) => {
+    if (addon.self.disabled) return;
+
+    const potentialDeleteButton = e.target.closest(DELETE_BUTTON_SELECTOR);
+    // If we haven't already, then...
+    if (e.target !== deleteButton && potentialDeleteButton) {
+      e.stopImmediatePropagation();
+      deleteButton = potentialDeleteButton;
+      setUpPressAndHold(); // ...set up this delete button.
+    }
+  });
+
+  /** Do a one-time setup for the press-and-hold logic per delete button. */
+  const setUpPressAndHold = () => {
+    startHold(deleteButton);
+
+    // Prevent the user from activating the delete button
+    deleteButton.addEventListener("click", (e) => {
+      if (addon.self.disabled) return;
+
+      if (e.isTrusted) {
+        e.stopImmediatePropagation();
+      } else if (addon.settings.get("noConfirm")) {
+        setTimeout(() => {
+          const confirmButton = document.querySelector("[class^='delete-confirmation-prompt_ok-button_']");
+          if (confirmButton) confirmButton.click();
+        }, 0);
+      }
+    });
+
+    deleteButton.addEventListener("mousedown", async (e) => {
+      if (addon.self.disabled) return;
+
+      e.stopImmediatePropagation();
+      startHold(deleteButton);
+    });
+
+    async function startHold(deleteButton) {
+      holding = true;
+
+      // Wait out the hold delay
+      for (let i = 0; i < 100; i += 2) {
+        if (!holding) break;
+        deleteButton.firstElementChild.style.background = `linear-gradient(0deg, #ff8c1a, #ff8c1a ${i}%, var(--editorDarkMode-primary, #855cd6) ${i}%)`;
+        await new Promise((resolve) => setTimeout(resolve, 17));
+      }
+
+      deleteButton.firstElementChild.style.transition = "none";
+      deleteButton.firstElementChild.style.removeProperty("background");
+      setTimeout(() => deleteButton.firstElementChild.style.removeProperty("transition"), 10);
+
+      // Then delete the sprite
+      if (holding) deleteButton.click();
+    }
+  };
+
+  // Cancel if released
+  document.body.addEventListener("mouseup", async () => {
+    holding = false;
+  });
+  // Cancel if image dragged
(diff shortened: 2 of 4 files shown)


mxmou:
I really like the animation, but I don't think this is necessary - until recently, no confirmation was the default behavior.

WorldLanguages:
It looks cool, though 😄

DNin01:
@LIZARD-OFFICIAL-77 Does this look like something you'd like? It lets you remove the confirmation.

Samq64:
If we go with this would a delay length setting be better than a toggle?

addons/hold-delete/addon.json
```diff
@@ -0,0 +1,42 @@
...
+  "name": "Hold to delete sprites",
+  "description": "Require the delete button to be held down in order to delete a sprite.",
+  "tags": ["editor"],
+  "credits": [
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
+    },
+    {
+      "name": "Samq64",
```
Samq64, line 11: There's no need to credit me here. Everything after the first commit from my PR was suggested by @WorldLanguages.

addons/hold-delete/addon.json
```diff
@@ -24,6 +24,26 @@
...
+        },
+        {
+          "id": "long",
+          "name": "Long"
+        }
+      ],
+      "default": "default"
+    },
     {
       "name": "Skip confirmation",
```
Samq64, line 48: Could this be a None option in the duration setting or was it left here so it's easier to find?

DNin01: What?

FYI, the hold duration setting controls how long you have to hold the button down to delete the sprite. There can't be a "none" option - you have to be able to hold it for some amount of time.

Were you talking about a different kind of delay?

Samq64: I meant for no holding, so you're right, keeping it this way is more intuitive.

mxmou: I just realized what this option is for and I don't think it makes sense - I thought the purpose of the addon was to replace the confirmation popup with the press and hold feature, not to effectively add a second confirmation. If we keep it, it should at least be on by default. And the delay setting should definitely have a "None" option to allow people to revert to Scratch's previous behavior.

DNin01: A "none" duration option would be equivalent to disabling the addon but just without the "skip confirmation" toggle, right?

If the addon says it adds a hold duration to the delete button, then why should it have an option for no hold duration? If the purpose of that option is to remove the confirmation, then we should rename the addon to indicate that it replaces the confirmation with something else so it makes more sense.

Samq64: I also thought this setting was for disabling the delay, not the confirmation prompt.

DNin01: I do agree we should reconsider the "skip confirmation" setting. Maybe we could instead update the addon description to say that the addon removes the confirmation and replaces it with a press and hold gesture, even though that feels like some people may miss that detail and then _oh no, the confirmation is gone!_

DNin01: And yes, the main reason I made this addon was so you could replace the confirmation with something else. Initially, though, I went with including a setting for whether or not to skip the confirmation after holding the delete button (off by default so users aren't surprised), so that's the origin of that setting.

addons/hold-delete/addon.json
```diff
@@ -0,0 +1,58 @@
...
+          "name": "Long"
+        }
+      ],
+      "default": "default"
+    },
+    {
+      "name": "Skip confirmation",
+      "id": "noConfirm",
+      "type": "boolean",
+      "default": false
```
DNin01, line 51:
```suggestion
      "default": true
```
I think the addon is much more useful if this setting is on by default, so I've come up with an idea for a notice to inform users that the press and hold gesture will skip the confirmation message:
```json
"info": [
  {
    "type": "notice",
    "id": "removesConfirmation",
    "text": "If the \"skip confirmation\" setting below is enabled, the hold gesture replaces the confirmation prompt that would appear after using the delete button."
  }
]
```

<details><summary>An alternative to that information...</summary>
<p>

...would be to rename the addon (but I don't know how well that will go):
```json
{
  "name": "Replace deletion confirmation with press and hold",
  "description": "Remove the confirmation prompt when deleting a sprite and instead require the delete button to be held down in order to delete the sprite."
}
```

</p></details>

addons/hold-delete/userscript.js
```diff
@@ -0,0 +1,77 @@
...
+  const DELETE_BUTTON_SELECTOR =
+    "[class^='sprite-selector_items-wrapper_'] [class*='sprite-selector-item_delete-button']";
+  const HOLD_SPEEDS = {
+    short: 12,
+    default: 17,
+    long: 30,
+  };
+  let deleteButton;
+  let holding = false;
+
```
AeroKoder, line 11:
```suggestion
let delete = false;
```

addons/hold-delete/userscript.js
```diff
@@ -0,0 +1,77 @@
+export default async function ({ addon, console }) {
+  const DELETE_BUTTON_SELECTOR =
+    "[class^='sprite-selector_items-wrapper_'] [class*='sprite-selector-item_delete-button']";
+  const HOLD_SPEEDS = {
+    short: 12,
+    default: 17,
+    long: 30,
+  };
+  let deleteButton;
+  let holding = false;
```
AeroKoder, line 10:
```suggestion
  let holding = false;
  let delete = false;
```

addons/hold-delete/userscript.js
```diff
@@ -0,0 +1,77 @@
...
+        deleteButton.firstElementChild.style.background = `linear-gradient(0deg, #ff8c1a, #ff8c1a ${i}%, var(--editorDarkMode-primary, #855cd6) ${i}%)`;
+        await new Promise((resolve) => setTimeout(resolve, timeout));
+      }
+
+      deleteButton.firstElementChild.style.transition = "none";
+      deleteButton.firstElementChild.style.removeProperty("background");
+      setTimeout(() => deleteButton.firstElementChild.style.removeProperty("transition"), 10);
+
+      // Then delete the sprite
+      if (holding) deleteButton.click();
```
AeroKoder, line 65:
```suggestion
      if (holding) delete = true;
```

addons/hold-delete/userscript.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  };
+
+  // Cancel if released
+  document.body.addEventListener("mouseup", async () => {
+    holding = false;
```
AeroKoder, line 71:
```suggestion
    holding = false;
    if (delete) deleteButton.click();
    delete = false
```

addons/hold-delete/userscript.js
```diff
@@ -0,0 +1,77 @@
...
+  });
+  // Cancel if image dragged
+  document.body.addEventListener("dragstart", async () => {
+    holding = false;
```
AeroKoder, line 75:
```suggestion
    holding = false;
    deleteItem = false;
```

AeroKoder (changes requested):
Making the sprite delete after the delay and after releasing the mouse button makes more sense than just immediately deleting the sprite after the delay.

addons/hold-delete/addon.json
WorldLanguages: > I thought the purpose of the addon was to replace the confirmation popup with the press and hold feature

Is that not the purpose?

mxmou: > > I thought the purpose of the addon was to replace the confirmation popup with the press and hold feature
> 
> Is that not the purpose?

In the initial version of the PR, it didn't remove the vanilla confirmation by default.

Issue #2764, opened 2021-06-13
Change how the addon names are named


Hans5958:
This is a group of suggestion related to the addon names.

- [1. Use consistent grammar](#1-use-consistent-grammar)
- [2. Shorten the names](#2-shorten-the-names)
- [3. Use title case](#3-use title case)

## 1. Use consistent grammar

The addon names are quite inconsistent in terms of the grammar. There are some that are phrases that starts with a verb (view, disable, etc.) and there are some that are just nouns.

Here's the list of a quantity of addons that I tagged based on if the names are phrases (with verb) or nouns.

![firefox_2021-06-13_19-39-57](https://user-images.githubusercontent.com/11584103/121808170-81797400-cc81-11eb-8206-edd0b9f8d15f.png)
![firefox_2021-06-13_19-42-22](https://user-images.githubusercontent.com/11584103/121808174-83dbce00-cc81-11eb-93fc-e7be54a60326.png)
![firefox_2021-06-13_19-44-24](https://user-images.githubusercontent.com/11584103/121808177-850cfb00-cc81-11eb-9aaa-6d7001b31498.png)

Seems that we have more nouns than phrases, we should change them to nouns. (Yes, I know there might some that can't be made as a noun, but just keep in minimum)

## 2. Shorten the names

Some names are just ridiculously long for just a name. Here are some examples.

- Filter messages on Scratch's messages page 
- Change new sprite default position 
- Customized quotes & code blocks on forums
- Show ocular.jeffalo.net statuses and post reactions

Again, this is quite ridiculous! My suggestion: Make them short, and put additional context inside the description. (Again, I know there might some that can, but at least we should try!)

## 3. Use title case

While we're at it, we should use title case (Feature Name) instead of sentence case (Feature name). This may be a mistake since I suggest to use sentence case, since last time there were more names with phrases than nouns, but looking that we have more names that just outright as a feature name (and a noun), we should just make them title case.

I mean, just look how other people do it. It's way better to use title case!

![firefox_2021-06-13_19-46-10](https://user-images.githubusercontent.com/11584103/121808524-1e88dc80-cc83-11eb-9054-c02e0e18e76a.png)
![firefox_2021-06-13_19-48-32](https://user-images.githubusercontent.com/11584103/121808526-1fba0980-cc83-11eb-92bd-5915b47955af.png)
[...]

PR #7986, opened 2024-12-09
Added date created to "last-edit-tooltip" addon

AeroKoder:
Resolves [#7974](https://github.com/ScratchAddons/ScratchAddons/issues/7974)

### Changes

Added the creation date of the project to the tool tip.

### Reason for changes

This addon would be helpful when wanting to know the creation date, of a project.

### Tests

Tested on Edge.

### DIFF
--- a/addons/last-edit-tooltip/addon.json
+++ b/addons/last-edit-tooltip/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Shared/edited dates tooltip",
-  "description": "Hover over a project's share date for information on when exactly it was shared and last edited.",
+  "description": "Hover over a project's share date for information on when exactly it was shared, last edited and created.",
   "credits": [
     {
       "name": "TheColaber",
@@ -9,6 +9,10 @@
     {
       "name": "Weredime (9gr)",
       "link": "https://scratch.mit.edu/users/9gr/"
+    },
+    {
+      "name": "AeroKoder",
+      "link": "https://scratch.mit.edu/users/AeroKoder/"
     }
   ],
   "dynamicEnable": true,
--- a/addons/last-edit-tooltip/addon.json
+++ b/addons/last-edit-tooltip/addon.json
@@ -24,6 +24,10 @@
     }
   ],
   "versionAdded": "1.2.0",
+  "latestUpdate": {
+    "version": "1.40.0",
+    "temporaryNotice": "Added the project creation date to the tooltip."
+  },
   "tags": ["community", "projectPage"],
   "enabledByDefault": false
 }
--- a/addons/last-edit-tooltip/addon.json
+++ b/addons/last-edit-tooltip/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Shared/edited dates tooltip",
+  "name": "Additional Project Info",
   "description": "Hover over a project's share date for information on when exactly it was created, shared and last edited.",
   "credits": [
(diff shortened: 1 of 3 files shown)


addons/last-edit-tooltip/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Shared/edited dates tooltip",
-  "description": "Hover over a project's share date for information on when exactly it was shared and last edited.",
+  "description": "Hover over a project's share date for information on when exactly it was shared, last edited and created.",
```
DNin01, line 3: Grammatical suggestion:
```suggestion
  "description": "Hover over a project's share date for information on when exactly it was created, shared, and last edited.",
```
Putting "created" first might make more sense as it's the first thing you do to start a Scratch project. Also, if "last edited", along with the word "last", is the final thing said, that makes it clear that the only thing that the word "last" applies to is "edited".

Samq64: Could this order be applied to the tooltip as well?

AeroKoder: Yes, I'm just about to commit it.

Samq64 (approved):
Could you add the [latestUpdate](https://scratchaddons.com/docs/reference/addon-manifest/#latestupdate) to the addon.json?

addons/last-edit-tooltip/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Shared/edited dates tooltip",
```
Samq64, line 2: Does someone have an idea for a better name?

addons/last-edit-tooltip/addon.json
```diff
@@ -20,6 +24,10 @@
     }
   ],
   "versionAdded": "1.2.0",
+  "latestUpdate": {
+    "version": "1.40.0",
```
Samq64, line 28:
```suggestion
    "version": "1.42.0",
```

addons/last-edit-tooltip/addon.json
AeroKoder: Maybe something like "Additional Project Info" or "Additional Project Dates"?

Samq64: Could "Additional project dates" be used with the lowercase letters to keep it consistent with other addons?

The problem with additional project info is there are already addons for script/sprite and block counts with one of those already named `project-info` internally.

AeroKoder: Ok, I will change it to “Additional project dates”

And also, I think we should consider going through all of the addons and updating their English titles to be in titlecase?

mxmou: > And also, I think we should consider going through all of the addons and updating their English titles to be in titlecase?

This has been discussed before: #2764

mxmou (approved):

addons/last-edit-tooltip/addon.json
```diff
@@ -20,6 +24,10 @@
     }
   ],
   "versionAdded": "1.2.0",
+  "latestUpdate": {
+    "version": "1.42.0",
```
WorldLanguages, line 28: Should have been 1.43.0

WorldLanguages: Ok, fixed by #8291

PR #7991, opened 2024-12-11
Move settings page search bar into header

### LINKED ISSUE #6997: Move settings search into header
From the Development Discord:
> (@TheColaber) hmm
> 

> (@rgantzos ) 

This makes a lot of sense since there's nothing else where to search bar is now.

Samq64:
Resolves #6997

### Changes

Moves the settings page search bar from below the header into it.

image

The search box is always light but I can change it to a variation of brand-orange if that looks better.

### Reason for changes

To save some vertical space.

### Tests

Tested on Chromium. The navigation bar starts to push the theme button off-screen at around 500px in English but the layout is already kind of broken elsewhere at that width anyway.


mxmou:
Moving it into the header is a good idea, but I don't like how it's placed in the corner. The search bar is one of the most important elements on the page, so I think it deserves a more prominent position - perhaps it could be aligned with the list of addons:

image

In "small mode" it can stay on the right side.

Now that there's no separate search toolbar, the `.addons-container` could have a bit more padding at the top.

> The search box is always light but I can change it to a variation of brand-orange if that looks better.

The white background makes it stand out, which I think is a good thing.

WorldLanguages:
I like mxmou's screenshot!

mxmou (approved):

WorldLanguages (changes requested):
This change breaks search autofocus in the settings popup

webpages/settings/index.html
```diff
@@ -120,17 +120,24 @@
   </head>
   <body v-cloak :class="{light: theme}">
     <div class="navbar">
```
mxmou, line 122:
```suggestion
    <div class="navbar" v-if="!isIframe">
```
This will fix autofocus. The code in `style.css` that hides the navbar in the popup (lines 204-206) can then be removed.

WorldLanguages (approved):

WorldLanguages:
This will be reverted by the "related addons" PR  :P

Issue #3414, opened 2021-09-01
Hotkey to change text in paint editor.


Endermaster3690:
### Is your request related to a problem? Please describe.

i just want to do this faster without having to move my mouse everywhere.

### Describe the solution you'd like

addon to do this

### Describe alternatives you've considered

haven't

### Additional context

no.

mxmou:
To switch to the text tool when when text is selected? I think using Enter for that would make sense.

An addon that adds keyboard shortcuts to all paint editor tools would be even more useful.

PR #8000, opened 2024-12-15
Adding some Costume Editor Keyboard Shortcuts

misaunde:
Adding a "costume-editor-shortcuts" add-on which adds keyboard shortcuts to the main tools in the paint editor.

Hi, first timer here 👋
My son does a lot of scratch, and he thought it'd be useful to have keyboard shortcuts in the costume editor. 
Anyways, I thought I'd attempt it as a Christmas present 🎄

Resolves https://github.com/ScratchAddons/ScratchAddons/issues/3414

### Changes

shortcuts_light_2
shortcuts_dark
image

https://github.com/user-attachments/assets/d668c463-4e7b-419e-95a3-0b81a92abb87

### Reason for changes

Keyboard shortcuts can help make switching between paint tools easier and is pretty standard in most modern painting programs.

### Tests

Tested in both Chrome and Firefox.


addons/costume-editor-shortcuts/addon.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "Costume Editor Shortcuts",
+  "description": "Enables some keyboard shortcuts in the Costume Editor.",
+  "credits": [
+    {
+      "name": "Hollow_Song's Dad",
+      "link": "https://scratch.mit.edu/users/Hollow_Song/"
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
```
Samq64, line 11: Could this be removed unless the feature can properly handle being turned off and back on again without a page reload?

addons/costume-editor-shortcuts/addon.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Costume Editor Shortcuts",
+  "description": "Enables some keyboard shortcuts in the Costume Editor.",
```
Samq64, line 3:
```suggestion
  "description": "Adds tool selection keyboard shortcuts in the Costume Editor.",
```

Samq64 (commented):
The shortcut indicators need to be re-added when switching between the vector and bitmap modes but everything else looks good.

Also, congratulations on getting #8000!

addons/costume-editor-shortcuts/userscript.js
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * Setup keydown listeners for shortcuts and and update UI to display available shortcuts.
+   */
+  async function initialize() {
+    if (isInitialized) return; // Prevent double initialization
+    isInitialized = true;
+    document.addEventListener("keydown", handleKeyDown);
+    document.addEventListener("focusin", userStartedTyping);
+    document.addEventListener("focusout", userStoppedTyping);
```
Samq64, line 56: I haven't tested my suggestion but I think if you add the listeners to an element in the costume editor you won't have to clean them up.

addons/costume-editor-shortcuts/addon.json
misaunde: Removed

addons/costume-editor-shortcuts/addon.json
misaunde: Updated

addons/costume-editor-shortcuts/userscript.js
misaunde: I tried this initially, but had issues getting anything to fire when the handlers were attached to any of the costume editor divs (no matter where I clicked). Probably user error, but `document` just worked.

Here's an example of what I tried:
```
  // var gui = await addon.tab.waitForElement("[class*='gui_tab-panel']");
  // var gui = await addon.tab.waitForElement("[class*='paint-editor_editor-container']");
  var gui = await addon.tab.waitForElement("#react-tabs-3");
  gui.addEventListener("keydown", handleKeyDown);
  gui.addEventListener("focusin", userStartedTyping);
  gui.addEventListener("focusout", userStoppedTyping);
```

I also wondered if there'd be an advantage to using document to not interrupt the workflow (when you're toggling between your game and costume editing). For example, if you eye droppered something on the game view or stage, your costume shortcuts would still work.

Samq64:
[TurboWarp](https://turbowarp.org/editor) has this with the only differences being the key indicators are in the tooltips and the line tool is selected with L instead of D.

WorldLanguages:
> [TurboWarp](https://turbowarp.org/editor) has this with the only differences being the key indicators are in the tooltips and the line tool is selected with L instead of D.

Ideally we should match TurboWarp I think

misaunde:
Updated:
image

Tested in Chrome and Firefox.

Since scratch is for kids, I liked the idea of making the shortcuts obvious, but, agreed, it makes the most sense to match TurboWarp. Thanks for pointing our their implementation (now I'm a little confused why Scratch doesn't already have this feature 🤔)

addons/addons.json
```diff
@@ -1,4 +1,5 @@
 [
+  "costume-editor-shortcuts",
```
DNin01, line 2: New addons are usually inserted above `"// NEW ADDONS ABOVE THIS ↑↑",` near the end of this file.

addons/costume-editor-shortcuts/addon.json
DNin01, line 1: Since the shortcut keys are displayed on the hover tooltips, it would be nice to mention to users that it is possible to see each tool's hotkey by hovering over its respective button.

DNin01 (commented):
Nice work!

addons/costume-editor-shortcuts/addon.json
misaunde: Good call, updated!

Samq64 (approved):

addons/costume-editor-shortcuts/userscript.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /**
+   * Iterate over the costume editor tool buttons and update their tooltips to include shortcuts.
+   */
+  async function addShortcutsToTitles() {
+    try {
+      const container = await addon.tab.waitForElement("[class^='paint-editor_mode-selector']");
+      container.querySelectorAll("span").forEach((span) => {
+        updateTitle(span, span.getAttribute("title"));
```
WorldLanguages, line 110: This will only work when the project editor is set to English.

@misaunde I'm not sure of any good solutions to identify which button is which, though.

WorldLanguages: I almost forget we have a dedicated method for this.

Try comparing against `addon.tab.scratchMessage("paint.brushMode.brush")` instead of hardcoding the English string `"Brush"`.

To find the message ID for a given string, type `__addon.tab.redux.state.locales.messages` into the console and it will display all key/value pairs.

WorldLanguages: The index varies on bitmap vs vector I think, so it's probably easier to compare the title attribute.

misaunde: Ohh, nice catch! Updated.

addons/costume-editor-shortcuts/addon.json
```diff
@@ -0,0 +1,19 @@
...
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "versionAdded": "1.41.0",
+  "tags": ["editor", "costumeEditor"],
```
Samq64, line 17:
```suggestion
  "tags": ["editor", "costumeEditor", "recommended"],
```

addons/costume-editor-shortcuts/addon.json
```diff
@@ -0,0 +1,19 @@
...
+  "enabledByDefault": false
```
Samq64, line 18:
```suggestion
  "enabledByDefault": true
```
This one will need some discussion.

Samq64, line 18:
```suggestion
  "enabledByDefault": true
```
This one will need some discussion.

Samq64:
Another thing: If you still want the shortcut labels on the buttons it could be made a setting if you think it's worth it.

addons/costume-editor-shortcuts/addon.json
WorldLanguages: Hmm... This is a tough one

misaunde: Since it's a new addon, it seems safer to me to not enable it by default until it's proven itself stable, and then maybe enable by default in a future release?

addons/costume-editor-shortcuts/userscript.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  /**
+   * For selecting costume editor tools by name.
+   */
+  async function switchTool(toolName) {
+    if (!toolName || addon.tab.redux.state.scratchGui.editorTab.activeTabIndex !== COSTUME_EDITOR_TAB_INDEX) return;
+
+    try {
+      const modeSelector = await addon.tab.waitForElement("[class^='paint-editor_mode-selector']");
```
WorldLanguages, line 104: I think `document.querySelector()` would be equivalent here. Am I missing anything?

WorldLanguages: Same question for line https://github.com/ScratchAddons/ScratchAddons/pull/8000/files#diff-002e6f939c0bba9955a36b01fdccd88c55b39a00c79974619be440174b164191R118

misaunde:
> Another thing: If you still want the shortcut labels on the buttons it could be made a setting if you think it's worth it.

Yea, I wondered about that. I showed my son the PR at Christmas (he was also surprised to learn TurboWarp had this feature already) and he said: "I've been painting in Scratch for 3 years without shortcuts. I'm not going to hover over a button for 3 seconds to see what the new shortcut is." 😄 

He's only a sample size of 1, but I kinda want to see how he adopts it before making further changes.

addons/costume-editor-shortcuts/userscript.js
misaunde: Hmm, I think I started with `document.querySelector()`, but got the impression somewhere along the way that `addon.tab.waitForElement` was a more appropriate way to do this in this repo. So is `document.querySelector()` preferred?

WorldLanguages: Yep https://scratchaddons.com/docs/develop/userscripts/best-practices/#only-use-waitforelement-when-necessary

misaunde: Since I check that we're on the custom tab before altering, I would assume the element is guaranteed to be there and using the aysnc addon method wouldn't buy us anything.

WorldLanguages: I'll change it to querySelector, and most async/await keywords should disappear as well as a side-effect

WorldLanguages (approved):
It's good enough now

Potential things to do next
- Adding a setting to always show the keyboard shortcuts on the screen would be interesting
- dynamicEnable & dynamicDisable (includes adding support for the case where the addon is toggled on while the user is already at the costume editor)
- It's likely other addons also have their own checks for `isUserTyping`, and we're not reusing that code. It's also possible we're missing some edge cases as there's multiple places the user can be typing (e.g. Sign Up flow while costume editor open on the background)

WorldLanguages:
Yes, bitmap mode is considered

WorldLanguages:
Suggestion: #8335

PR #8004, opened 2024-12-17
dark-www: remove styles for old Ideas page

mxmou:
### Changes

Removes styles that have become unnecessary after the Ideas page update. Also makes the new page more similar to vanilla when the "Scratch's default colors" preset is used.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/dark-www/addon.json
+++ b/addons/dark-www/addon.json
@@ -358,6 +358,17 @@
       }
     },
     {
+      "name": "box-blackText",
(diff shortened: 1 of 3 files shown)


Samq64:
The changes in this PR look good but `.video-title` uses a darker text colour without the addon compared to Scratch's default colour preset. Also the video metadata has low contrast but that's also how it is without the addon.

PR #8010, opened 2024-12-18
Include hidden files in packer action

Samq64:
Resolves https://github.com/ScratchAddons/ScratchAddons/pull/7134#issuecomment-2552137326

Tells upload-artifact to explicitly include dotfiles since it doesn't by default anymore. See https://github.com/actions/upload-artifact/issues/602.

### Tests

I successfully ran the action on my fork and loaded the extension into Chromium.

### DIFF
--- a/.github/workflows/packer.yml
+++ b/.github/workflows/packer.yml
@@ -23,3 +23,4 @@ jobs:
         with:
           name: Packed Extensions (zip)
           path: .dist/*.zip
+          include-hidden-files: true


DNin01 (approved):
This will work fine, but I think a better approach would be to not have hidden files in the first place.

Hans5958:
Would you guys prefer this or the other solution by @DNin01?

Samq64:
I haven't tested it yet but I like @DNin01's solution better.

DNin01:
I think #8015 will fix the issue that #8012 left behind.

DNin01:
Superseded.

PR #8014, opened 2024-12-19
Make 'More Settings' linkable

### LINKED ISSUE #5422: Open "more settings" automatically when visiting webpages/index.html#moresettings

It would be useful to send a link such as https://scratchaddons.com/settings#moresettings when necessary

### What this change does

We already do this with addon IDs, we'd be extending the approach to the "new settings" window which some users might not notice is there

### Possible alternatives

_No response_

_No response_

Samq64:
Resolves #5422

### Changes

Makes more settings linkable and replaces the more settings button with a link. Right now more settings stays open after a page reload but I can change that if it's too annoying.

### Reason for changes

So more settings can be linked to directly.

### Tests

Tested on Chromium. Addon linking still works.

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -84,7 +84,7 @@ let fuse;
         forceEnglishSetting: null,
         forceEnglishSettingInitial: null,
         switchPath: "../../images/icons/switch.svg",
-        moreSettingsOpen: false,
+        moreSettingsOpen: location.hash.toLocaleLowerCase() === "#moresettings",
         categoryOpen: true,
         loaded: false,
         searchLoaded: false,
@@ -344,6 +344,9 @@ let fuse;
       forceEnglishSetting(newValue, oldValue) {
         if (oldValue !== null) chrome.storage.local.set({ forceEnglish: this.forceEnglishSetting });
       },
+      moreSettingsOpen(newValue) {
+        if (!newValue) location.hash = "#";
+      }
     },
     ready() {
       // Autofocus search bar in iframe mode for both browsers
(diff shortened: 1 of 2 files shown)


DNin01 (commented):
Making it possible for More Settings to be opened directly from a link is a good idea, but it's a dialog, does it really make sense to stay in More Settings after a page reload? Also then we'd have to decide what to do about the browser back button.

It's just the convention for dialogs not to act like another location but rather be part of the same page.

But this would all make sense if More Settings was a full screen modal or a separate page and not a dialog.

webpages/settings/index.js
```diff
@@ -380,15 +381,19 @@ let fuse;
...
-          const addonId = location.hash.replace(/^#addon-/, "");
-          const groupWithAddon = this.addonGroups.find((group) => group.addonIds.includes(addonId));
-          if (!groupWithAddon) return; //Don't run if hash is invalid
-          const addon = this.manifestsById[addonId];
-
-          groupWithAddon.expanded = true;
-          this.selectedCategory = addon?.tags.includes("easterEgg") ? "easterEgg" : "all";
-          this.clearSearch();
-          setTimeout(() => document.getElementById("addon-" + addonId)?.scrollIntoView(), 0);
+          if (location.hash.toLocaleLowerCase() === "#moresettings") {
```
mxmou, line 384:
```suggestion
          if (location.hash === "#moresettings") {
```
`#addon-*` links are case-sensitive, so I think this one should be too.

webpages/settings/index.js
```diff
@@ -592,7 +597,9 @@ let fuse;
     vue.loaded = true;
     setTimeout(() => {
       const hash = window.location.hash;
-      if (hash.startsWith("#addon-")) {
+      if (location.hash.toLocaleLowerCase() === "#moresettings") {
```
mxmou, line 600:
```suggestion
      if (location.hash === "#moresettings") {
```

mxmou (approved):

DNin01 (approved):

PR #8012, opened 2024-12-19
Packer: Change folder path `.dist` → `dist`

DNin01:
The other half of ScratchAddons/packer-script#2

### Changes

Updates the "pack extensions" workflow to upload artifacts from:
```diff
- .dist/*.zip
+ dist/*.zip
```

### Reason for changes

upload-artifact no longer uploads hidden files because of the risk of uploading confidential information.

PR #8015, opened 2024-12-19
Don't include dist/ in archives

DNin01:
Fixes https://github.com/ScratchAddons/ScratchAddons/pull/8012#issuecomment-2555750817

> That didn't quite work, the "dist" folder is being included inside the chrome zip now.

This should fix the problem. Previously we were `git archive`-ing every non-dot folder including `dist`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,6 +21,7 @@
 *.md export-ignore
 LICENSE export-ignore
 .* export-ignore
+dist export-ignore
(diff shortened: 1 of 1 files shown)


Samq64:
`eslint.config.mjs` should also be listed since it's not a dot file with the new name.

Samq64 (approved):

PR #8019, opened 2024-12-21
editor-theme3: fix comment text in the forums

### LINKED ISSUE #8018: White comment text with editor-theme3 on forums
> I have no idea if this is actually a bug or not, but I think this was caused by Scratch Addons, as I don't remember this happening before turning on the "Change colors of the blocks in the forums" option in "Customizable block colors" (which I enabled due to the update to 3.0 scratchblocks). Anyway, for the bug, the block comments are almost impossible to read because the text is white, and I use a modified version of the 2.0 preset with the old extension color. This doesn't happen in the editor, only in the forums.

mxmou:
Resolves #8018

### Changes

Uses a more specific selector for comment text.

### Reason for changes

This fixes the text being unreadable in the default and Scratch 2.0 presets.

### Tests

Tested on Edge and Firefox with different combinations of the "comments" and "text color" settings. Example post that contains a scratchblocks comment: https://scratch.mit.edu/discuss/post/7975899/

### DIFF
--- a/addons/editor-theme3/forums/base.css
+++ b/addons/editor-theme3/forums/base.css
@@ -129,8 +129,7 @@ svg[class*="scratchblocks-style-scratch3"] .sb3-label,
   fill: var(--editorTheme3-commentColor);
   stroke: var(--editorTheme3-commentBorder);
 }
-.sb3-comment-label,
-.sb3-label.sb3-comment-label {
+svg[class*="scratchblocks-style-scratch3"] .sb3-label.sb3-comment-label {
   fill: var(--editorTheme3-commentText);
 }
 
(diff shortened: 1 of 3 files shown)


addons/editor-theme3/forums/base.css
```diff
@@ -129,8 +129,7 @@ svg[class*="scratchblocks-style-scratch3"] .sb3-label,
   fill: var(--editorTheme3-commentColor);
   stroke: var(--editorTheme3-commentBorder);
 }
-.sb3-comment-label,
-.sb3-label.sb3-comment-label {
+svg[class*="scratchblocks-style-scratch3"] .sb3-label.sb3-comment-label {
   fill: var(--editorTheme3-commentText);
```
Samq64, line 133: The forum comment's text is pure black by default even without high contrast mode.

mxmou: scratchblocks still uses the 2.0 colors for comments, so I don't think the addon needs to match its default output exactly.

Samq64 (approved):

PR #8020, opened 2024-12-21
Fix custom-menu-bar icons only mode

Samq64:
Hides the debug label and removes the right margin from both help buttons in the icon only mode of the customizable editor menu bar.

Someone sent feedback about the label on 1.39.2 but it also applies to the latest version.


mxmou (approved):

WorldLanguages:
Whoops, I was about to send a comment about that feedback post as well.

WorldLanguages (approved):
Tested in Chromium

PR #8023, opened 2024-12-22
`mediarecorder` default setting fix

DNin01:
### Bug fix

Changes the function that loads settings or default settings to use `Object.assign` in order to load saved settings on top of default settings.

#### Before

If **no** settings are saved, then the setting values fall back to defaults. If **any** settings are saved, then every saved setting is restored instead of using defaults. Basically, either all the defaults or all the saved settings are filled in.

This meant **if we were ever to add new options in an update**, users who had previously used the addon would find that those settings start out as an undefined value.

#### After

For each setting, if there is a saved setting value, that is used, otherwise the default value is used.

### Other changes

Renamed `LENGTH_LIMIT` to `MAX_RECORD_TIME`.

### Tests

Tested.

### DIFF
--- a/addons/mediarecorder/userscript.js
+++ b/addons/mediarecorder/userscript.js
@@ -25,7 +25,10 @@ export default async ({ addon, console, msg }) => {
 
   const getStoredOptions = () => {
     try {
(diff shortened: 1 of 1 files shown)


CST1229:
i guess i didn't know about this pr when i made #8031

CST1229 (approved):

Samq64 (approved):

PR #8031, opened 2024-12-24
`mediarecorder`: Add "Video scale" option

CST1229:
Resolves [this suggestion posted in the TurboWarp Discord server](https://discord.com/channels/837024174865776680/1321234454710321174)

### Changes

Adds a "Video scale" option in the project video recorder addon's dialog. This option forces the stage canvas' (and thus the video's) scale to be whatever the option is set to while recording; for example, if you have a 480x360 stage and set the video scale to 2, the canvas will be forced to 960x720; therefore, the outputted video file will be 960x720.
image

This option has a range from 0.01-15x, which is probably a bit too much, but should be more than enough for anyone. (Maybe if someone is, say, making a game with an 8x8 stage or something?)

### Reason for changes

Allows making higher (or lower, I guess)-resolution recordings. It also prevents the canvas from resizing during recording, which at least prevents Firefox from outputting corrupted 0-byte WEBMs.

### Tests

Only tested on Firefox.

### DIFF
--- a/addons/mediarecorder/addon.json
+++ b/addons/mediarecorder/addon.json
@@ -15,8 +15,8 @@
     }
   ],
   "latestUpdate": {
-    "version": "1.35.0",
-    "temporaryNotice": "The maximum recording time has been increased to 10 minutes (600 seconds)."
+    "version": "1.41.0",
+    "temporaryNotice": "A new \"Video scale\" option has been added, to allow making high resolution recordings."
   },
   "versionAdded": "1.6.0",
   "enabledByDefault": false
--- a/addons/mediarecorder/userscript.js
+++ b/addons/mediarecorder/userscript.js
@@ -2,6 +2,8 @@ import downloadBlob from "../../libraries/common/cs/download-blob.js";
 
 export default async ({ addon, console, msg }) => {
   const LENGTH_LIMIT = 600;
+  const MIN_SCALE = 0.01;
+  const SCALE_LIMIT = 15;
   const DEFAULT_SETTINGS = {
     secs: 30,
     delay: 0,
@@ -9,8 +11,11 @@ export default async ({ addon, console, msg }) => {
     micEnabled: false,
     waitUntilFlag: true,
     useStopSign: true,
+    stageScale: 1,
   };
   const LOCALSTORAGE_ENTRY = "sa-record-options";
+
+  const vm = addon.tab.traps.vm;
 
   let isRecording = false;
   let isWaitingForFlag = false;
@@ -18,6 +23,7 @@ export default async ({ addon, console, msg }) => {
   let abortController = null;
   let stopSignFunc = null;
   let recordBuffer = [];
+  let stageScale = 1;
   let recorder;
   let timeout;
 
@@ -25,7 +31,7 @@ export default async ({ addon, console, msg }) => {
 
   const getStoredOptions = () => {
     try {
-      return JSON.parse(localStorage.getItem(LOCALSTORAGE_ENTRY)) ?? DEFAULT_SETTINGS;
+      return Object.assign(structuredClone(DEFAULT_SETTINGS), JSON.parse(localStorage.getItem(LOCALSTORAGE_ENTRY)) ?? DEFAULT_SETTINGS);
     } catch {
       return DEFAULT_SETTINGS;
     }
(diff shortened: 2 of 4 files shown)


WorldLanguages:
Does it actually increase the resolution of the recording? Or are all neighboring pixels the same color?

CST1229:
> Does it actually increase the resolution of the recording? Or are all neighboring pixels the same color?

It _should_ actually increase the resolution of the recording (it should have the same effect as e.g going into fullscreen does)

CST1229:
for comparison, here's a simple project recorded in 0.25x, 1 and 4x scale:
[0.25x.webm](https://github.com/user-attachments/assets/f639b4be-7763-425e-b814-491c83c3aae0)
[1x.webm](https://github.com/user-attachments/assets/a16bed24-b287-4984-842e-7297dba4ce37)
[4x.webm](https://github.com/user-attachments/assets/dbe28602-2907-4ee6-a227-075ce689ddda)

mxmou:
I think the setting would be easier to understand if it controlled the resolution directly, with 480x360 as the default. The width and height inputs would need to be linked so that changing one also updated the other.

CST1229:
> I think the setting would be easier to understand if it controlled the resolution directly, with 480x360 as the default. The width and height inputs would need to be linked so that changing one also updated the other.

yeah makes sense

WorldLanguages:
Should 480x360 still be the default? Why not something closer to fullscreen resolution?

CST1229:
> Should 480x360 still be the default? Why not something closer to fullscreen resolution?

idk. any recording made before this update would be 480x360 or lower (as that's the maximum stage scale you can get without fullscreen), and the fullscreen resolution depends on the window/screen size

addons/mediarecorder/userscript.js
```diff
@@ -188,6 +223,67 @@ export default async ({ addon, console, msg }) => {
         })
       );
 
+      // Video resolution.
```
DNin01, line 226: I think a better way of doing resolution settings would be to let users pick one of the standard quality measures. (480p, 720p, 1080p, etc.)

If you really wanted, there could also be a "custom" option where you get to choose the exact resolution.

Samq64: For the higher resolutions it won't be the standard width users expect since it's 4:3 and not 16:9.

addons/mediarecorder/addon.json
```diff
@@ -15,8 +15,8 @@
     }
   ],
   "latestUpdate": {
-    "version": "1.35.0",
-    "temporaryNotice": "The maximum recording time has been increased to 10 minutes (600 seconds)."
+    "version": "1.41.0",
```
mxmou, line 18:
```suggestion
    "version": "1.43.0",
```

addons/mediarecorder/userscript.js
```diff
@@ -2,22 +2,28 @@ import downloadBlob from "../../libraries/common/cs/download-blob.js";
 
 export default async ({ addon, console, msg }) => {
   const MAX_RECORD_TIME = 600; // seconds
+  const MIN_SCALE = 0.0001;
+  const SCALE_LIMIT = 20;
```
mxmou, line 6:
```suggestion
  const MAX_SCALE = 20;
```

addons/mediarecorder/userscript.js
```diff
@@ -2,22 +2,28 @@ import downloadBlob from "../../libraries/common/cs/download-blob.js";
...
   const DEFAULT_SETTINGS = {
     secs: 30,
     delay: 0,
     audioEnabled: true,
     micEnabled: false,
     waitUntilFlag: true,
     useStopSign: true,
+    stageScale: 1,
```
mxmou, line 14:
```suggestion
    stageScale: 2,
```

mxmou (changes requested):
The scaling seems to be wrong on high DPI displays. My display scale is set to 125%. I set the resolution to 524x393, but the video is actually 655x491 - almost exactly 1.25x the intended resolution.

DNin01:
> The scaling seems to be wrong on high DPI displays. My display scale is set to 125%. I set the resolution to 524x393, but the video is actually 655x491 - almost exactly 1.25x the intended resolution.

It might be scaling something redundantly then.

Issue #8016, opened 2024-12-20
Account Switcher addon


Elip100:
### Why this would be helpful

Many people have multiple Scratch accounts, and it can be annoying to have to log out and sign in with your credentials every time you want to check the messages of a different account.

### How this addon works

Under the account drop-down there would be a list of saved accounts, and you could click one to instantly switch to it (probably with a stored session token or something similar). It would be similar to how Google websites have an account switcher.

### Possible alternatives

Can't think of another addon that this could be built into

### Additional context

Simple mock-up (the real thing would look better than this; I made this in 5 minutes)

![image](https://github.com/user-attachments/assets/c34dd385-613e-4107-a1cd-b57d104e6592)

I intend to develop this if it is a well-received idea.

CubesterYT:
Related, and also closed, issue: https://github.com/ScratchAddons/ScratchAddons/issues/3309
It's still a great idea, but read the comments of that issue, it'll explain why it was denied the first time.

Samq64:
#1074 is also related and suggests making it a popup feature instead.

DNin01:
We'd need to encrypt the users' credentials for security.

Still though, anything account-login-related might be better suited for a password manager, but if there are some standout features in an account-switching addon, such as:
- Being able to peek at the unread message count of your other accounts
- Being able to view unshared projects from your other accounts
- An account switcher in the Scratch Messaging popup
- Passwordless login (using passkeys or something like that)

…then this could be a good fit for an addon — a credential manager of our own just for Scratch could integrate with the Scratch website's functions.

Samq64:
> Being able to peek at the unread message count of your other accounts

Last time I checked no authentication was required if it's just the count.

GarboMuffin:
> We'd need to encrypt the users' credentials for security.

How would you go about meaningfully encrypting that data and what threat model is being addressed?

GarboMuffin:
The only site-specific account-switching feature in a browser extension I'm aware of is Reddit Enhancement Suite which just stores usernames & passwords in plain text, including in backups of settings.

A possibly notable difference from Scratch & Reddit accounts is that Reddit supports 2FA. For those accounts RES will prompt for the 2FA code each time so the plain text username & password is slightly less dangerous. But I suspect a very small minority of Reddit accounts actually use this.

WorldLanguages:
Discussion about this feature is spread out between multiple issues, discussions, and Discord chats. It's quite messy.

[...]

PR #8032, opened 2024-12-25
Account switching draft

WorldLanguages:
I wrote some JavaScript functions that implement my idea from comment https://github.com/ScratchAddons/ScratchAddons/issues/8016#issuecomment-2558505794

### DIFF
--- a/addons/account-switching/addon.json
+++ b/addons/account-switching/addon.json
@@ -1,0 +1,16 @@
+{
+  "name": "Account switching",
+  "description": "-",
+  "popup": {
+    "icon": "../../images/icons/users.svg",
+    "name": "Accounts",
+    "fullscreen": false,
+    "script": "popup.js",
+    "html": "popup.html"
+  },
+  "tags": ["popup", "featured"],
+  "versionAdded": "1.41.0",
+  "enabledByDefault": false,
+  "dynamicEnable": true,
+  "dynamicDisable": true
+}
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -157,6 +157,7 @@
   "remaining-replies",
   "forum-user-agent",
   "no-sprite-confirm",
+  "account-switching",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/popups/account-switching/popup.html
+++ b/popups/account-switching/popup.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8" />
+  <title></title>
+</head>
+<body>
+
+</body>
+</html>
--- a/popups/account-switching/popup.html
+++ b/popups/account-switching/popup.html
@@ -1,10 +1,8 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8" />
-  <title></title>
-</head>
-<body>
-
-</body>
+  <head>
+    <meta charset="UTF-8" />
+    <title></title>
+  </head>
+  <body></body>
 </html>
--- a/popups/account-switching/popup.js
+++ b/popups/account-switching/popup.js
@@ -1,0 +1,67 @@
+const getAccountSwitchingCookie = async () => JSON.parse((await window.cookieStore.get("accountswitching")).value);
+const setAccountSwitchingCookieTo = (obj) => {
+  obj.extVersion = chrome.runtime.getManifest().version;
+  document.cookie = `accountswitching=${JSON.stringify(obj)}; Max-Age=31536000; Path=/; SameSite=Strict`;
+};
+
+async function addCurrentAccount() {
+  // Remembers the current sessionID for future use.
+  // This function must be executed in a webpage context (i.e. not by the service worker)
+
+  const scratchCookieInfo = await chrome.cookies.get({
+    url: "https://scratch.mit.edu/",
+    name: "scratchsessionsid",
+  });
+  const currentSessionId = scratchCookieInfo.value;
+
+  if ((await window.cookieStore.get("accountswitching")) === null) {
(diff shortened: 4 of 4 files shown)


WorldLanguages:
The actual cookie expiry should be around one week, and renewed each time the user opens the popup. This would be to consider that, if the user has disabled the extension (DISABLED, not uninstalled) the cookie will eventually expire.

DNin01 (changes requested):
I think there should be some way to encrypt or not store credentials.
- The credentials could be protected with a password or passkey.
- Remembering passwords could be an option. When off, switching accounts would sign you out, open Scratch's sign-in page, autofill your username, and have you enter your password.

Either one of the above or it should be noted that we do not encrypt their credentials and account access could be stolen if someone gains access to the device Scratch Addons is installed on.

We should also update our privacy policy if we need to clarify how sensitive information is used.

WorldLanguages:
> account access could be stolen if someone gains access to the device Scratch Addons is installed on

The existing privacy policy more or less considers this already. It mentions that malware and attackers can access extension storage.

No need to encrypt the session IDs, because the vanilla Scratch session ID is not encrypted.

Under no circumstances I will upload password manager software to the Chrome Web Store, so the only option left is storing session IDs.

DNin01:
> No need to encrypt the session IDs, because the vanilla Scratch session ID is not encrypted.

Some people sign out of accounts when they're done using them.

mxmou:
> > No need to encrypt the session IDs, because the vanilla Scratch session ID is not encrypted.
> 
> Some people sign out of accounts when they're done using them.

Then they probably shouldn't use this addon - it will be practically equivalent to being signed into multiple accounts at once.

WorldLanguages:
That was a misclick.

DNin01:
I suppose none of us are proficient enough to responsibly make a secure credential manager, nor should we be interested in turning Scratch Addons into one, so our best bet would be to leave a note on the addon explaining the way we do things and suggesting the use of a password manager instead for security concerns.

DNin01:
Things to think about:
- Will be automatically save accounts to the list or will we ask the user if they want us to save it when they sign in?
- How much info will be saved? Will we remember any credentials?
- How will we handle signing out and removing accounts from the list? Should removing an account also sign it out?
- What if a password changes?
- What if you switch accounts while a Scratch tab is open? Does it refresh? What about multiple tabs? Do they all refresh? Does one refresh and the rest close?

TheColaber:
So ideally, the UI would be straight on scratch's page (see #3327). The function of storing and switching between session tokens would of course not be done on the page, 
but best case scenario the addon's ui would:
 - show saved accounts when they are logged out (maybe somewhere in the sign in popup?)
 - add a single option in the username dropdown that would add/remove the current signed in account
 - add other saved accounts in the username dropdown to switch between
 - show message counts of other saved accounts

however, due to probably safety issues and its been made clear several times, it looks like it would be better to have this ui in the scratch addons popup. if thats the case, the ui would probably look a bit different and probably a bit more interesting... although it would look similar to all the ideas i mentioned above.

mxmou:
> > What if you switch accounts while a Scratch tab is open? Does it refresh? What about multiple tabs? Do they all refresh? Does one refresh and the rest close?
> 
> I don't think any tabs at all should reload. We can remind the user to reload after switching accounts.

If we don't reload tabs, I think we should at least update the account name displayed in the navbar to avoid the confusing state where the account Scratch thinks you're using and the account you're actually using are different. It's also important to tell the user to save their work _before_ switching accounts, because it isn't possible to save a project if you aren't logged into the account that owns it.

WorldLanguages:
True, I forgot about unsaved work in open editors.

TheColaber:
> @DNin01 @TheColaber My opinions
> 

I completely agree with everything stated.

> If we don't reload tabs, I think we should at least update the account name displayed in the navbar to avoid the confusing state where the account Scratch thinks you're using and the account you're actually using are different. It's also important to tell the user to save their work before switching accounts, because it isn't possible to save a project if you aren't logged into the account that owns it.

But can't "account switching" already be achieved by signing out from one account and logging into the other? Why should this addon act any differently than how scratch current acts. I understand that logically it makes more sense for tabs to update or reload, but then this addon would be doing two things at once, adding switching functionality, and adding functionality to update existing scratch tabs. I guess that could be okay, but it just feels weird that this addon would be controlling that.

mxmou:
> But can't "account switching" already be achieved by signing out from one account and logging into the other? Why should this addon act any differently than how scratch current acts. I understand that logically it makes more sense for tabs to update or reload, but then this addon would be doing two things at once, adding switching functionality, and adding functionality to update existing scratch tabs. I guess that could be okay, but it just feels weird that this addon would be controlling that.

Signing into another account on the Scratch website itself would reload the current tab. It can't reload other tabs, but the extension can and if it doesn't, the user will have to do it anyway.

DNin01:
> If we don't reload tabs, I think we should at least update the account name displayed in the navbar to avoid the confusing state where the account Scratch thinks you're using and the account you're actually using are different.

And make fields editable if you switch to the owner of a project and change the editor UI and ahhh... That's why I prefer the reload solution or reminding the user.

Or maybe it can reload situationally, like maybe we reload automatically unless a project is running (and might have unsaved progress) or you have unsaved changes or you have more than one tab open, then we don't reload but notify the user that they will have to reload their tabs to finish switching accounts.

mxmou:
> And make fields editable if you switch to the owner of a project and change the editor UI and ahhh... That's why I prefer the reload solution or reminding the user.

I haven't tested this, but I think it would happen automatically on scratch-www pages and in the editor if we update the Redux state. On scratchr2 pages a reload would probably be needed.

If there are unsaved changes, we need to tell the user about that before switching, when they're still logged into the right account to be able to save them.

WorldLanguages:
> And make fields editable if you switch to the owner of a project and change the editor UI and ahhh...

That can already happen if you manually switch accounts in a different tab

faretek1:
this would be quite useful

PR #8036, opened 2024-12-27
Add font-to-svg addon

GavinGoGaming:
### Changes

Added a tab with functionality to use google fonts or uploaded fonts to create SVG/vectors for usage within costumes. Visible text is localized (for "en" font)

image

image

### Reason for changes

I've found myself wanting to use custom fonts a lot on scratch (usually for good-looking UI) and I find myself going to one of few ad-filled websites to create svgs from my fonts, going back to scratch, and uploading them. This addon allows that process to happen all inside a single scratch tab!

### Tests/Notes

Tested on Brave and Chrome, both worked fine and supports usage alongside Variable Manager. Takes up tab id 4, var. man uses id 3.

Makerjs, Bezierjs, and Opentype are included in the file at the top. A comment explains what line to skip to in order to get past the library code. File is still only ~900kb, and I plan to remove some libraries (see below)

### Future plans
- "Add Costume" button to speed up process
- Improved UI (to fit the scratch style)
- Remove coloris and use native scratch color picker (one less library and fetch)
- Use cached(?) fonts instead of fetching from google


just-a-hriday:
This seems useful for sure, but the interface needs some work.
Also versionAdded should be 1.41.0.

GavinGoGaming:
> This seems useful for sure, but the interface needs some work. Also versionAdded should be 1.41.0.

Knew I did versionAdded wrong!
I already have a decent ui update made, that's my priority right now 100%.

just-a-hriday:
The tab icon doesn't quite match the other tabs. Maybe you could use the text icon from the costume editor?

mxmou:
> Also, perhaps this could be integrated directly into the costume editor - you could add a more fonts button to the font dropdown (similar to how it is in google docs) and add a pathify button next to it.

It would be even better if it wasn't necessary to click a "convert to path" button at all - the addon could convert the text automatically and store the original text as an attribute of the path element so that it can still be edited if the addon is enabled.

GavinGoGaming:
> Also, perhaps this could be integrated directly into the costume editor - you could add a more fonts button to the font dropdown (similar to how it is in google docs) and add a pathify button next to it.

> It would be even better if it wasn't necessary to click a "convert to path" button at all - the addon could convert the text automatically and store the original text as an attribute of the path element so that it can still be edited if the addon is enabled.

For now i'll keep it as a tab - but that's definitely a great addition to be added later to replace the planned "Add to costume" button.

> Could the libraries be moved to `/libraries/thirdparty` and the license info be filled out?

> The tab icon doesn't quite match the other tabs. Maybe you could use the text icon from the costume editor?

I'll fix both of those next push

GavinGoGaming:
> Could the libraries be moved to `/libraries/thirdparty` and the license info be filled out?

I noticed that those say "Note that these libraries are either from official release websites or from unpkg" - however in testing I had to modify parts of the libraries to get them to work (ex. Bezier.js used some weird syntax that broke when used without modification). Should I put the modified ones in /libraries/thirdparty?

Edit: new version does not need the weird tweaked libraries

WorldLanguages:
> however in testing I had to modify parts of the libraries to get them to work

Can you provide more details? Have you got a diff?

Samq64:
#8254 is an alternative that allows uploading fonts directly as a sprites with a built-in text engine script.

mxmou:
> #8254 is an alternative that allows uploading fonts directly as a sprites with a built-in text engine script.

I think these have different purposes. This one is for adding static text to a costume, like the costume editor's text tool, but with support for custom fonts. #8254 is for text engines, which can draw any text using code. We could add both, but I prefer W_L's suggestion to turn #8254 into a website.

GavinGoGaming:
i'm finally updating this pr again!
do you guys think this would fit in more as a btn in the costume editor? the seperate tab could be used for other svg generation stuff or, similar to #8254, splitting a font into multiple costumes - however the current solution might fit better in the costume editor

Samq64:
> Could the libraries be moved to /libraries/thirdparty and the license info be filled out?

Sorry, I meant  /libraries/thirdparty/**cs**. Also could the librairies be minified?

GavinGoGaming:
> Sorry, I meant /libraries/thirdparty/**cs**. Also could the librairies be minified?

Done

addons/font-to-svg/fontToSvg.js
```diff
@@ -0,0 +1,434 @@
...
+    return App;
+}());
+
+export default async function ({ addon, console, msg }) {
+  const vm = addon.tab.traps.vm;
+
+  let preventUpdate = false;
+  function addTabContent(holder) {
+	  const scripts = [
+        { src: "https://cdn.jsdelivr.net/gh/mdbassit/Coloris@latest/dist/coloris.min.js", type: "text/javascript" }
```
Samq64: It's more work but could Scratch's built-in colour picker be used instead?

libraries/license-info.json
Samq64, line 1: Could opentype.js be added here as well?

libraries/thirdparty/cs/maker.js
Samq64, line 1: Is a 100KB+ library necessary to render a font as an SVG, or am I missing something?

manifest.json
```diff
@@ -71,6 +71,7 @@
         "addon-api/*",
         "addons/*",
         "libraries/*/cs/*",
+        "libraries/*/f2svg/*",
```
Samq64: This is unused now.
```suggestion
```

Samq64:
> do you guys think this would fit in more as a btn in the costume editor?

If the UI could be simular to onion skinning with the text updating on the paint canvas then that would be great, but also a lot of work.

addons/font-to-svg/fontToSvg.js
mxmou: Or just `<input type="color">`.

mxmou:
> do you guys think this would fit in more as a btn in the costume editor?

It would be more intuitive to use if it was integrated into Scratch's existing text tool instead of having a separate button, as suggested in some earlier comments:

> > Also, perhaps this could be integrated directly into the costume editor - you could add a more fonts button to the font dropdown (similar to how it is in google docs) and add a pathify button next to it.
> 
> It would be even better if it wasn't necessary to click a "convert to path" button at all - the addon could convert the text automatically and store the original text as an attribute of the path element so that it can still be edited if the addon is enabled.

addons/font-to-svg/fontToSvg.js
```diff
@@ -0,0 +1,434 @@
...
+    App.prototype.render = function (fontIndex, variantIndex, text, size, union, filled, kerning, separate, bezierAccuracy, units, fill, stroke, strokeWidth, strokeNonScaling, fillRule) {
+        var _this = this;
+        var f = this.fontList.items[fontIndex];
+        var v = f.variants[variantIndex];
+        var url = f.files[v].replace('http:', 'https:');
+        if (this.customFont) {
+            this.callMakerJs(this.customFont, text, size, union, filled, kerning, separate, bezierAccuracy, units, fill, stroke, strokeWidth, strokeNonScaling, fillRule);
+        }
+        else {
+            opentype.load(url, function (err, font) {
```
Samq64: > DEPRECATED! migrate to: opentype.parse(buffer, opt) See: https://github.com/opentypejs/opentype.js/issues/675

libraries/thirdparty/cs/maker.js
GavinGoGaming: MakerJS is the library that the font2svg website used. It has a text constructor that is able to export to SVG and DXF data with a ton of customization. I could possibly remove any unused shapes/features from it

Samq64: If you used code from that website do you know what it is licensed under?

GavinGoGaming: Apache 2.0

mxmou: opentype.js has a [`getPath()`](https://github.com/opentypejs/opentype.js?tab=readme-ov-file#fontgetpathtext-x-y-fontsize-options) method for rendering text, and you can call [`toSVG()`](https://github.com/opentypejs/opentype.js?tab=readme-ov-file#pathtosvgoptions) on the result to get the SVG path element. Can you use those instead?

addons/font-to-svg/fontToSvg.js
```diff
@@ -0,0 +1,463 @@
+// Variable Manager's code from Jeffalo and GarboMuffin helped so much with the tabs on this ext! Big thanks to them.
+import MakerJs from "../../libraries/thirdparty/cs/maker.js";
+import * as opentype from "../../libraries/thirdparty/cs/opentype.min.mjs";
+import "../../libraries/thirdparty/cs/bezier.js";
```
mxmou: Is this needed for one of the other libraries to work?

GavinGoGaming: opentype loads the inputted font into the format used by maker.js, and bezier.js is another dependency of maker.js

Samq64: As mxmou mentioned in https://github.com/ScratchAddons/ScratchAddons/pull/8036#discussion_r2374848019 is there a reason why both of those libraries can't be dropped in favour of opentype's built-in methods?

GavinGoGaming:
- swapped out maker.js/bezier.js for opentype's direct API
- added "insert to sprite" button

PR #7288, opened 2024-03-22
Make `hide-backpack` adjust for `custom-zoom`'s autohide

mybearworld:
Resolves #7286

### Changes

Expand the `.sa-custom-zoom-area` when `hide-backpack` is active.

Note that this also makes hovering over the backpack icon show the zoom buttons.

### Reason for changes

Making zooming in possible while this addon is enabled

### Tests

Tested on Edge.

PR #8041, opened 2024-12-30
Fix `hide-backpack` incompatibility with auto-hiding zoom controls in `custom-zoom`

### LINKED ISSUE #8040: hide-backpack broken if auto-hiding zoom controls enabled
### Description

hide-backpack (replace backpack bar with button) doesn't work if the autohide zoom controls setting is enabled on custom-zoom.
The button is visible, but clicking it doesn't do anything.

I haven't checked yet, but this is likely just a layering issue.

### Steps to reproduce

1. Enable hide-backpack and custom-zoom, and enable the autohide zoom controls setting.
2. Open a project in the editor.
3. Try clicking the backpack button.
4. Disable the autohide zoom controls setting and reload.
5. Try clicking the button again. It should work this time.#

### Environment

Browser: Chrome 131.0.6778.205
Operating system: Windows

just-a-hriday:
Resolves #8040

### Changes

Added `z-index: 2` to stylesheet in hide-backpack to fix layering issue.

### Tests

Tested on Chrome 131.0.6778.205.

### DIFF
--- a/addons/hide-backpack/button.css
+++ b/addons/hide-backpack/button.css
@@ -29,6 +29,7 @@
   border: 2px solid rgba(87, 94, 117, 0.5);
   border-radius: 50%;
   cursor: pointer;
+  z-index: 2;
 }
 
 [dir="ltr"] .sa-backpack-button {
--- a/addons/hide-backpack/button.css
+++ b/addons/hide-backpack/button.css
@@ -29,7 +29,7 @@
   border: 2px solid rgba(87, 94, 117, 0.5);
   border-radius: 50%;
   cursor: pointer;
-  z-index: 2;
+  z-index: 2; /* above custom zoom controls */
 }
 
 [dir="ltr"] .sa-backpack-button {


Samq64 (approved):
There is another compatibility issue when the hide zoom setting is off where the zoom buttons are not aligned properly until the backpack is clicked but that is not caused by this PR.

image

just-a-hriday:
> There is another compatibility issue when the hide zoom setting is off where the zoom buttons are not aligned properly until the backpack is clicked but that is not caused by this PR.
> 
> image
> 

Might it be a good idea to make the backpack button hide with the zoom controls?

Samq64:
> Might it be a good idea to make the backpack button hide with the zoom controls?

No? The issue above fixes itself when backpack is opened and doesn't happen again after closing it.

addons/hide-backpack/button.css
```diff
@@ -29,6 +29,7 @@
   border: 2px solid rgba(87, 94, 117, 0.5);
   border-radius: 50%;
   cursor: pointer;
+  z-index: 2;
```
mxmou, line 32:
```suggestion
  z-index: 2; /* above custom zoom controls */
```

mxmou (changes requested):

mxmou:
> Might it be a good idea to make the backpack button hide with the zoom controls?

It intentionally doesn't hide: https://github.com/ScratchAddons/ScratchAddons/pull/7288#issuecomment-2016580298

mxmou:
> There is another compatibility issue when the hide zoom setting is off where the zoom buttons are not aligned properly until the backpack is clicked but that is not caused by this PR.

Can't repro. Does this happen consistently?

Samq64:
> Does this happen consistently?

Not anymore.

mxmou (approved):

PR #8058, opened 2025-01-04
New addon: Jump to unread posts in discussion forums

### LINKED ISSUE #8055: Forums: jump to unread posts when opening a topic
From Scratch comments:

> could we add an addon that allows you to jump to the unread posts of a topic just by clicking the topic, instead of "New posts" or "last post: undecember 32nd 2024 24:00:00 by kaj"

Explanation for non-forumers: currently, clicking the topic title opens the first page of the topic and clicking "New Posts" opens the oldest unread post. With this addon enabled, clicking the title would go to the unread post and the separate "New Posts" links would be removed.

image

just-a-hriday:
Resolves #8055

### Changes

Makes topic links on the discuss page go to the first unread post rather than the first post of the topic. Also removes the new posts links (which also removes the _Necropost?_ links created by the necropost-highlighter addon).

The addon name and description could probably be worded better.

### Tests

Tested on Chrome 131.0.6778.205.

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -157,6 +157,7 @@
   "remaining-replies",
   "forum-user-agent",
   "no-sprite-confirm",
+  "forums-jump-to-unread",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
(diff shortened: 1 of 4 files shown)


DNin01 (changes requested):
If the unread post has been dustbinned, attempting to load the page will result in a 403 error, because users don't have access to the dustbin.

If this happens, the normal topic link still works fine, however, if this addon is enabled, the topic link will be replaced with the unread post link and the topic will become completely inaccessible.

WorldLanguages:
I do not recommend running a userscript on `https://scratch.mit.edu/403` to fix that.

Sending a request to check if it's dustbinned, and then actually go to the page (causing a second request) sounds quite bad as well. I can't think of any other ideas.

PR #8062, opened 2025-01-05
Implement setting descriptions

### LINKED ISSUE #5320: Addon setting descriptions/info

Some settings can't be self-explained in just a name.

### What this change does

There will be a `?` or `i` button on some addon setting names that will display an extended setting description when clicked or hovered to further describe the setting if it is potentially vague.

### Possible alternatives

Not that I know of.

_No response_

Samq64:
Resolves #5320

### Changes

Adds a help button next to settings that shows its description in a tooltip when hovered like https://github.com/ScratchAddons/ScratchAddons/issues/5320#issuecomment-1533790219

This PR does not include any actual setting descriptions right now.

### Reason for changes

To display more information about an addon's settings without making them very long or using a notice instead.

### Tests

On setting names that wrap there is more padding than intended.

### DIFF
--- a/addons/block-cherry-picking/addon.json
+++ b/addons/block-cherry-picking/addon.json
@@ -30,6 +30,7 @@
     {
       "name": "Flip controls",
       "id": "invertDrag",
+      "description": "Make grabbing blocks individually the default behavior. Hold Ctrl to drag entire stacks.",
       "type": "boolean",
       "default": false
     }
--- a/addons/block-cherry-picking/addon.json
+++ b/addons/block-cherry-picking/addon.json
@@ -5,6 +5,10 @@
     {
       "text": "On macOS, use the Cmd key instead of the Ctrl key.",
       "id": "macContextDisabled"
+    },
+    {
+      "text": "If \"flip controls\" is enabled, grabbing blocks individually will be the default behavior. Hold Ctrl to drag the entire stack.",
+      "id": "flipControls"
     }
   ],
   "credits": [
@@ -26,7 +30,6 @@
     {
       "name": "Flip controls",
       "id": "invertDrag",
-      "description": "Make grabbing blocks individually the default behavior. Hold Ctrl to drag entire stacks.",
       "type": "boolean",
       "default": false
     }
--- a/addons/block-cherry-picking/addon.json
+++ b/addons/block-cherry-picking/addon.json
@@ -5,10 +5,6 @@
     {
       "text": "On macOS, use the Cmd key instead of the Ctrl key.",
       "id": "macContextDisabled"
-    },
-    {
-      "text": "If \"flip controls\" is enabled, grabbing blocks individually will be the default behavior. Hold Ctrl to drag the entire stack.",
-      "id": "flipControls"
     }
   ],
   "credits": [
--- a/webpages/settings/components/addon-setting.js
+++ b/webpages/settings/components/addon-setting.js
@@ -4,7 +4,6 @@ export default async function ({ template }) {
     template,
     data() {
       return {
-        isIframe: window.parent !== window,
         rowDropdownOpen: false,
         noResetDropdown: ["table", "boolean", "select"].includes(this.setting.type),
       };
--- a/webpages/settings/components/addon-setting.js
+++ b/webpages/settings/components/addon-setting.js
@@ -4,6 +4,7 @@ export default async function ({ template }) {
     template,
     data() {
       return {
+        isIframe: window.parent !== window,
         rowDropdownOpen: false,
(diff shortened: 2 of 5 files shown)


DNin01 (commented):
Depending on the screen size, tooltips can sometimes be positioned partially offscreen.

https://github.com/user-attachments/assets/0069ea8e-f907-4c8c-815c-d831936194f6

WorldLanguages:
> Depending on the screen size, tooltips can sometimes be positioned partially offscreen.

Should tooltips ever be positioned to the left of the help icon? That would, for most languages, hide the name of the setting because it is now behind the tooltip.
To avoid covering the name or potential values for the setting, we could consider positioning the tooltip above/below the help icon (for example, below the icon unless the edge case where there is not enough space due to the scrolling state)

Samq64:
The only position that will never overflow is above then which is a little strange. Is it worth adding JavaScript logic for the positioning instead?

WorldLanguages:
Yes I think JavaScript is needed

DNin01:
> Yes I think JavaScript is needed

I think we should make another PR that adds tooltip positioning capabilities as an API that lets us pop tooltips out to the left/right or above/below.

Samq64:
> > Yes I think JavaScript is needed
> 
> I think we should make another PR that adds tooltip positioning capabilities as an API that lets us pop tooltips out to the left/right or above/below.

There is already a class for positioning it to the left. I'm just trying to figure out how to conditionally add that class on hover without causing a flicker while it updates.

WorldLanguages:
Maybe this article can help: https://web.dev/articles/building/a-tooltip-component

Samq64:
Does someone know how to fix this?

[Screencast_20250106_134525.webm](https://github.com/user-attachments/assets/e7ad46d4-4436-4ab7-a4f1-108775d10e1a)

> Maybe this article can help: https://web.dev/articles/building/a-tooltip-component (Demo: https://gui-challenges.web.app/tooltips/dist/)

That demo doesn't handle automatically changing the position based on the window size.

webpages/settings/components/addon-setting.js
```diff
@@ -149,6 +150,10 @@ export default async function ({ template }) {
         this.addonSettings[this.setting.id] = newValue;
         this.updateSettings();
       },
+      checkOverflowing(e) {
+        this.isOverflowing = window.innerWidth - e.clientX < 250;
```
DNin01, line 154: A few issues:
- The mouse could be at a slightly different location on different function calls, causing the tooltip to pop out inconsistently.
  - Could we use `getBoundingClientRect()` instead of `MouseEvent.clientX`?
- There is still a scroll bar along the bottom of the page before this function is called to change the popout direction.
  - This could be fixed by setting `display: none` instead of `visibility: hidden` on the tooltip.

DNin01:
> Should tooltips ever be positioned to the left of the help icon? That would, for most languages, hide the name of the setting because it is now behind the tooltip.

Easy fix:
A tooltip popped out to the left of a label

WorldLanguages:
Great job!

Just one thing: if the addon is disabled, it looks like this.
image

Unfortunately in CSS there is no way to avoid inheriting opacity. So the only solution I can think of, is disabling setting descriptions if the addon is disabled.

WorldLanguages (approved):

WorldLanguages:
The tooltip is quite small and positioned above the help icon, so I think we could try to get it to work properly in the popup.

image

WorldLanguages:
Okay, setting descriptions are now showing on the popup

image

Prove me wrong...

Samq64:
> Prove me wrong...

It can overflow on the left side if the setting name is shorter than around 8 characters but I'm not sure if there are currently any that length in English.

WorldLanguages:
Hmm, that can become a problem in Japanese.

WorldLanguages:
The simplest solution might be adding a reasonable min-width to `.setting-label`

addons/block-cherry-picking/addon.json
```diff
@@ -30,6 +26,7 @@
     {
       "name": "Flip controls",
       "id": "invertDrag",
+      "description": "Make grabbing blocks individually the default behavior. Hold Ctrl to drag entire stacks.",
```
DNin01, line 29: You can keep this here or remove it before merging, but if you do keep it, wait until the manifest schema and translation system are updated to include this new property before merging.

DNin01 (approved):
Looks alright!

addons/block-cherry-picking/addon.json
WorldLanguages: I readded it to make sure translating this new field works properly. I didn't check the schema though

Issue #5160, opened 2022-10-04
Support for touchscreen laptops/Chromebooks


lisa-wolfgang:
We should discuss to what extent Scratch Addons should support touchscreen devices. There are many devices capable of running browsers that we fully support that have touchscreen functionality -- Chromebooks (certain models) and Microsoft Surface laptops are two of the most popular examples.

Many addons assume that the user has access to a keyboard, but some devices (or users' configurations) may not have physical keyboards, and their virtual equivalents may not support the Ctrl, Alt, or Shift keys. I haven't looked through our addons with this in mind yet, but I'm guessing that there are other instances (such as touch events not being supported in lieu of a mouse-pointer) where touchscreens have not been considered.

RedGuy12:
We also mention the Android Kiwi browser in the FAQ (unless that PR wasn't ever merged) and this would help simplify TurboWarp integration for Garbo.

DNin01:
Probably the first addon we should make touchscreen-compatible is 60 FPS project player mode. The higher framerate could be activated by touching and holding on the green flag and then selecting an option from a context menu.

CST1229:
> Probably the first addon we should make touchscreen-compatible is 60 FPS project player mode. The higher framerate could be activated by touching and holding on the green flag and then selecting an option from a context menu.

And mute as well.

WorldLanguages:
We could kill two birds with one stone: touch support would benefit laptop users, as well as mobile users.
iPad TurboWarp users too! Three birds :P

WorldLanguages:
I think this is yet one more item that should be added to a future page on the Scratch Addons docs, on things to consider when creating a new addon. It would apply to addons that already exist too, of course.

penguinmoose:
The individual block dragging addon would need to be made touchscreen-compatible. Maybe holding the block would be a replacement for a keypress.

DNin01:
These are, of what I found so far, the addons that need keyboard alternative function triggers:
- `60fps` Could possibly be a context menu item.
- `bitmap-copy` We'd have to support copying by clicking the Copy button.
- `block-duplicate` Might have to add an extra context menu item.
- `block-cherry-picking` Same thing here.
- `ctrl-click-run-scripts` Do we need to worry about this one?
- `editor-comment-previews` There's not really an alternative - clicking on comments expands them.
- `jump-to-def` Could probably be a context menu item, just like Scratch 2.0.
- `mute-project` Add a mute button.

[...]

PR #8064, opened 2025-01-05
"Go to definition" in context menu

DNin01:
One of the tasks in #5160

Right click menu of custom block with an option go to definition

### Changes

Adds an item to the context menu of custom block calls to jump to the definition of the custom block.

### Reason for changes

You can't middle click or Shift-click on a touch screen.

### Tests

Tested in Edge 131.


addons/jump-to-def/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Jump to custom block definition",
```
Samq64, line 2:
```suggestion
  "name": "Go to custom block definition",
```

Samq64 (approved):

addons/jump-to-def/addon.json
DNin01: I think it's okay for "jump" and "go" to be used interchangeably. "Jump" is a better summarizing word and "go" is more natural and works best in a verb phrase.

Samq64:
This breaks compatability with the modern Blockly editor since the context menu API isn't compatible with it yet.

DNin01:
> Acually this breaks compatability with the modern Blockly editor since the context menu API isn't compatible with it yet.

So then we'll update the API, right?

addons/jump-to-def/addon.json
```diff
@@ -1,6 +1,6 @@
...
-  "description": "Allows you to jump to a custom block's definition using the middle mouse button or Shift+Click on the block.",
+  "description": "Right click on a custom block for an option to go to its definition. You can also middle click or Shift+Click on a custom block to immediately go to its definition.",
```
mxmou, line 3:
```suggestion
  "description": "Jump to a custom block's definition by middle-clicking or Shift+clicking on the block or selecting \"Go to Definition\" in the right click context menu.",
```

addons/jump-to-def/addon.json
```diff
@@ -19,6 +19,10 @@
     }
   ],
   "versionAdded": "1.28.0",
+  "latestUpdate": {
+    "version": "1.41.0",
+    "temporaryNotice": "You can now use the context menu (open by right clicking or touching and holding) instead of the keyboard shortcut to jump to custom block definitions."
```
mxmou, line 24:
```suggestion
    "temporaryNotice": "You can now use the context menu (open by right clicking or touching and holding) to jump to custom block definitions."
```
I think it should be clear that the old way to go to the definition isn't being removed.

addons/jump-to-def/userscript.js
```diff
@@ -27,24 +43,26 @@ export default async function ({ addon, msg, console }) {
...
       }
     }
 
     _doBlockClick_.call(this);
   };
+
+  addon.tab.createBlockContextMenu(
+    (items, block) => {
+      if (!addon.self.disabled && block.type === "procedures_call") {
+        items.splice(3, 0, {
```
mxmou, line 58: This puts the new item before Edit in the code area and after Edit in the flyout:

image
image

I think it should always be after Edit, like in Scratch 2.0:

image

mxmou:
> Acually this breaks compatability with the modern Blockly editor since the context menu API isn't compatible with it yet.

It doesn't break the addon entirely - middle click and Shift-click still work. The context menu item doesn't appear but that can be fixed later.

mxmou:
I think this addon will need to be added to the `CONTEXT_MENU_ORDER` array in `addon-api/content-script/Tab.js`.

WorldLanguages:
By the way, wow, I forgot this was a 2.0 feature

medlans:
> By the way, wow, I forgot this was a 2.0 feature

It could be the fact that the Scratch Wiki never seemed to really document this for a while, and it was only added later.

Samq64:
May I push to this branch?

DNin01:
> May I push to this branch?

Sure

DNin01:
Good!

WorldLanguages:
Cool! Merging...

PR #8070, opened 2025-01-07
Conditional addon notices

### LINKED ISSUE #3930: Make `if` work for notices and warings

Samq64:
Resolves #3930

### Changes

Adds the ability to hide notices using the same structure/logic as `setting.if` and moves them into a component (but not update notices for now). The styles are still in `addon-body` but they can be moved if needed.

### Reason for changes

To hide notices when they aren't relevant while keeping them clearly visible when they are.

### Tests

Tested on Chromium.

### DIFF
--- a/addons/cat-blocks/addon.json
+++ b/addons/cat-blocks/addon.json
@@ -20,7 +20,10 @@
     {
       "type": "notice",
       "text": "The \"watch mouse cursor\" setting may impact performance when the editor is open.",
-      "id": "watch"
+      "id": "watch",
+      "if": {
+        "settings": { "watch": true }
+      }
     }
   ],
   "settings": [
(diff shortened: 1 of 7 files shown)


Samq64:
I already had this working locally before I thought of setting descriptions so I don't mind if this is rejected.

WorldLanguages:
I like this idea for "On MacOS,..." notices, but I don't think it replaces the need for confirmation when toggling "dangerous" or experimental settings.

Joeclinton1:
Is this done?

Samq64:
This doesn't support OS/browser conditions yet and some of the setting conditions would probbably be better as setting tooltips.

PR #8073, opened 2025-01-09
Update context menu code

mxmou:
### Changes

Updates the context menu API to work with the new version of Blockly. The updated API only supports block context menus - for workspace and comment menus, addons can call `Blockly.ContextMenuRegistry.registry.register()` directly.

### Tests

Tested block switching, blocks2image and devtools on Edge and Firefox. The menu items they add appear correctly, but most of them don't work - I'll fix them in a later PR. swap-local-global's menu callback currently throws an error in new Blockly - this will also be fixed later.

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -757,6 +757,30 @@ export default class Tab extends Listenable {
     createdAnyBlockContextMenus = true;
 
     this.traps.getBlockly().then((ScratchBlocks) => {
+      if (ScratchBlocks.registry) {
+        // new Blockly
+        const oldGenerateContextMenu = ScratchBlocks.BlockSvg.prototype.generateContextMenu;
+        ScratchBlocks.BlockSvg.prototype.generateContextMenu = function (...args) {
+          let items = oldGenerateContextMenu.call(this, ...args);
+          for (const { callback, blocks, flyout } of contextMenuCallbacks) {
+            let injectMenu =
+              // Block in workspace
+              (blocks && !this.isInFlyout) ||
+              // Block in flyout
+              (flyout && this.isInFlyout);
+            if (injectMenu) {
+              try {
+                items = callback(items, this);
+              } catch (e) {
+                console.error("Error while calling context menu callback: ", e);
+              }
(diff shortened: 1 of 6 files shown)


WorldLanguages:
> for workspace and comment menus, addons can call Blockly.ContextMenuRegistry.registry.register() directly.

As far as I can see, the order is determined by the `weight:` prop when using that method, so that is quite nice. This probably means we get to remove some addons from `CONTEXT_MENU_ORDER` in Tab.js

DNin01:
> > for workspace and comment menus, addons can call Blockly.ContextMenuRegistry.registry.register() directly.
> 
> As far as I can see, the order is determined by the `weight:` prop when using that method, so that is quite nice.

Much like our shared spaces API!

mxmou:
> > for workspace and comment menus, addons can call Blockly.ContextMenuRegistry.registry.register() directly.
> 
> As far as I can see, the order is determined by the `weight:` prop when using that method, so that is quite nice. This probably means we get to remove some addons from `CONTEXT_MENU_ORDER` in Tab.js

CONTEXT_MENU_ORDER is still needed for block context menu items and each addon adds at least one of those, so it unfortunately won't get any shorter after these changes. The reason why ContextMenuRegistry can't be used to add items to block context menus is that it doesn't allow us to add our items after vanilla items that are specific to certain blocks, such as "Edit" for custom blocks.

WorldLanguages:
> CONTEXT_MENU_ORDER is still needed for block context menu items and each addon adds at least one of those, so it unfortunately won't get any shorter after these changes.

Whoops, good point.

Samq64:
`block.getCategory()` is not defined on the new Blockly. `block.styleName_` seems to return the same values but I'm not sure if that's the correct way to check it.

mxmou:
> `block.getCategory()` is not defined on the new Blockly. There's `block.styleName_` but I don't think that's the correct way to check it.

I think the closest equivalent is `block.type.split("_")[0]`. I'll fix this in another PR because it's unrelated to context menus.

Samq64 (approved):

PR #8077, opened 2025-01-10
Use translated setting descriptions

Samq64:
See #8062 and https://github.com/ScratchAddons/l10n-script/pull/16.

Untested


WorldLanguages (approved):
Tested

PR #8078, opened 2025-01-11
Add one setting description for testing

WorldLanguages:
For testing purposes.


DNin01:
We could also branch out, you know? Ideally, checks should pass whenever pushing something to `master`. But it’s fine, ScratchAddons/manifest-schema#72 is getting merged soon…?

Samq64:
If this setting description is shipped the notice should be removed before releasing.

PR #8099, opened 2025-01-18
Disable pasting blocks with ctrl+v outside of workspace.

yuri-kiss:
(this is my first PR to this repo so if I did something wrong please tell me)

### Changes

https://youtu.be/sL9dNADnlmE

Disables ctrl+v when your mouse is not inside the workspace area.

### Reason for changes

This disables the annoying "bug" when you are on the canvas and press ctrl+v, whether it was by accident or on purpose.

### Tests

It works on the latest Firefox and Chromium versions, I haven't tested Safari or IE.

Polyfill testy thing I did: https://js.do/code/reiogjoertwisjhgwoiejiroe

#### QA

Q: Why not use the same detection method as the other keybinds?
A: https://github.com/ScratchAddons/ScratchAddons/pull/8099#issuecomment-2599814640

Q: What is the "bug"?
A: https://youtu.be/sL9dNADnlmE


Mistium:
:> thank u miyo, ur the best

GarboMuffin:
why not copy the pattern that is already used by the other shortcuts in the same addon

```js
if (document.activeElement.tagName === "INPUT") {
    return;
}
```

yuri-kiss:
> why not copy the pattern that is already used by the other shortcuts in the same addon
> 
> ```js
> if (document.activeElement.tagName === "INPUT") {
>     return;
> }
> ```

Because `document.activeElement` won't ever be the workspace or its children.

GarboMuffin:
I don't know what "the bug when you are on the canvas and press ctrl+v, whether it was by accident or on purpose" is so maybe that's a good place to start

yuri-kiss:
> I don't know what "the bug when you are on the canvas and press ctrl+v, whether it was by accident or on purpose" is so maybe that's a good place to start

I will record a video.

yuri-kiss:
> I don't know what "the bug when you are on the canvas and press ctrl+v, whether it was by accident or on purpose" is so maybe that's a good place to start

@GarboMuffin https://youtu.be/sL9dNADnlmE

mxmou:
This doesn't actually prevent the blocks from being pasted, it just makes them not stick to the mouse cursor. Is that intended?

yuri-kiss:
> This doesn't actually prevent the blocks from being pasted, it just makes them not stick to the mouse cursor. Is that intended?

weird, it prevented it for me and a tester, where are you trying to paste. Ill try to fix it

mxmou:
> > This doesn't actually prevent the blocks from being pasted, it just makes them not stick to the mouse cursor. Is that intended?
> 
> weird, it prevented it for me and a tester, where are you trying to paste. Ill try to fix it

If I press Ctrl+V when the mouse is inside the workspace, the block becomes attached to the mouse cursor. If the mouse is outside the workspace, the block is pasted next to its original position, as if the devtools addon wasn't enabled.

yuri-kiss:
> To fix this, you could probably override the Scratch/Blockly function that implements pasting or add a keydown event listener with `capture: true` that calls `preventDefault()`.

mmmm, Ill fix this

yuri-kiss:
fixed @mxmou

addons/editor-devtools/DevTools.js
```diff
@@ -43,6 +44,34 @@ export default class DevTools {
       this.initInner(root);
     }
   }
+
+  async patchCopy() {
```
mxmou, line 48: I think `patchPaste` would be a better name.

addons/editor-devtools/DevTools.js
```diff
@@ -43,6 +44,34 @@ export default class DevTools {
...
+    const devtools = this;
+    const blockly = await this.addon.tabs.traps.getBlockly();
```
mxmou, line 50:
```suggestion
    const blockly = await this.addon.tab.traps.getBlockly();
```

addons/editor-devtools/DevTools.js
```diff
@@ -43,6 +44,34 @@ export default class DevTools {
...
+    if (blockly.registry) {
+      const BlockPaster = blockly.registry.getClass(blockly.registry.Type.PASTER, 'block').constructor;
+      const oldPasteFunc = BlockPaster.prototype.paste;
+      BlockPaster.prototype.paste = function(copyData, workspace, coords) {
+        if (devtools.addon.settings.get("enablePasteBlocksAtMouse")) {
+          if (!this._canPaste(workspace)) return;
```
mxmou, line 56:
```suggestion
          if (!devtools._canPaste(workspace)) return;
```

addons/editor-devtools/DevTools.js
```diff
@@ -43,6 +44,34 @@ export default class DevTools {
...
+        }
+        return oldPasteFunc.call(this, copyData, workspace, coords);
+      };
+    } else {
+      const oldPasteFunc = blockly.WorkspaceSvg.prototype.paste;
+      blockly.WorkspaceSvg.prototype.paste = function (data) {
+        if (!(this.rendered && this.currentGesture_)) return;
```
mxmou, line 63:
```suggestion
```

`currentGesture_` will be `null` unless a block or comment is being dragged.

addons/editor-devtools/DevTools.js
```diff
@@ -43,6 +44,34 @@ export default class DevTools {
...
+        if (devtools.addon.settings.get("enablePasteBlocksAtMouse")) {
+          if (data.tagName.toLowerCase() !== "comment" && !devtools._canPaste(this)) {
+            this.currentGesture_.cancel();
```
mxmou, line 66:
```suggestion
```

addons/editor-devtools/DevTools.js
```diff
@@ -401,6 +430,18 @@ export default class DevTools {
     }
   }
 
+  _canPaste(workspace) {
+    // Don't paste if the mouse is outside the workspace SVG group
+    const group = workspace.svgGroup_;
+    const bounds = group.getBoundingClientRect();
```
mxmou, line 436: The group's bounds include blocks that are scrolled out of view:

image

You can get the actual bounds from the `.blocklySvg`:

image

```suggestion
    // Don't paste if the mouse is outside the workspace SVG
    const svg = document.querySelector(".blocklySvg");
    const bounds = svg.getBoundingClientRect();
```

mxmou (changes requested):

addons/editor-devtools/DevTools.js
yuri-kiss: oh my bad thats a typo

addons/editor-devtools/DevTools.js
yuri-kiss: woops sorry

addons/editor-devtools/DevTools.js
yuri-kiss: goofy

addons/editor-devtools/DevTools.js
yuri-kiss: this is what the function itself does?

addons/editor-devtools/DevTools.js
yuri-kiss: why

yuri-kiss: im following what hte blockly function does

addons/editor-devtools/DevTools.js
yuri-kiss: fixable I guess

addons/editor-devtools/DevTools.js
mxmou: Blockly only checks `rendered`, not `currentGesture_`, but the addon doesn't have to do either - it should simply call the Blockly function if `canPaste` is true and do nothing otherwise.

yuri-kiss: I don't see why it can't but okay, I will fix it

mxmou: This line (`this.currentGesture_.cancel()`) can be removed too. It stops the current drag, which isn't necessary here because nothing is being pasted.

mxmou (approved):

Samq64 (approved):
Tested on both editors

mxmou:
It's still failing for some reason. Can you allow edits by maintainers so that I can fix the formatting manually?

addons/editor-devtools/DevTools.js
```diff
@@ -43,6 +44,33 @@ export default class DevTools {
...
   }
+
+  async patchPaste() {
+    const devtools = this;
+    const blockly = await this.addon.tab.traps.getBlockly();
+    // Support for new blockly.
+    if (blockly.registry) {
+      const BlockPaster = blockly.registry.getClass(blockly.registry.Type.PASTER, 'block').constructor;
+      const oldPasteFunc = BlockPaster.prototype.paste;
+      BlockPaster.prototype.paste = function(copyData, workspace, coords) {
```
mxmou, line 55:
```suggestion
      const BlockPaster = blockly.registry.getClass(blockly.registry.Type.PASTER, "block").constructor;
      const oldPasteFunc = BlockPaster.prototype.paste;
      BlockPaster.prototype.paste = function (copyData, workspace, coords) {
```

mxmou (commented):
I'm not able to push the formatting changes Prettier wants to make (probably because the fork is owned by an org), so I'm posting them as a suggestion that you can accept.

addons/editor-devtools/DevTools.js
yuri-kiss: thank you so much

PR #8102, opened 2025-01-19
Remove flip controls notice

Samq64:
The setting description added in #8078 for testing is working fine and has been translated into multiple languages, so the notice is redundant now.


DNin01 (approved):
I guess this is fine if we're keeping the "testing" description for this release.

PR #8109, opened 2025-01-22
Don't run costume editor shortcuts with modifier keys

Samq64:
From feedback:

> I found a bug in Costume editor keyboard shortcuts addon. When i press ctrl + a to select all, it works but it also selects reshape tool. so better to make to selects tools when key ctrl || shift || alt isn't pressed

### Tests

Tested on Firefox. Ctrl+A switches to the select tool as expected.


DNin01 (approved):

WorldLanguages:
Whoops

PR #8116, opened 2025-01-24
Fix colored context menus on modern Blockly

Samq64:
### Changes

Fixes `editor-colored-context-menus` compatibility the new Blockly editor.

### Tests

Tested on Chromium with separators, Scratch's high contrast mode and `editor-theme3`'s colored on black setting.

### DIFF
--- a/addons/editor-theme3/theme3.js
+++ b/addons/editor-theme3/theme3.js
@@ -545,9 +545,15 @@ export default async function ({ addon, console, msg }) {
     };
   }
 
-  if (!Blockly.registry) {
-    // editor-colored-context menus compatibility
-    // doesn't support new Blockly yet
+  if (Blockly.registry) {
+    // New Blockly
+    const oldBlockShowContextMenu = Blockly.BlockSvg.prototype.showContextMenu;
+    Blockly.BlockSvg.prototype.showContextMenu = function (e) {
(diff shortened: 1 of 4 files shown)


addons/editor-theme3/theme3.js
```diff
@@ -879,6 +885,8 @@ export default async function ({ addon, console, msg }) {
         )
       );
       workspace.refreshTheme();
+      // used by editor-colored-context-menus
+      document.body.style.setProperty("--colour-text", uncoloredTextColor());
```
mxmou, line 889: This seems unnecessary - the variable is only referenced by colored context menus in a rule targeting `.blocklyMenuItem`, which only exists in new Blockly.

mxmou (approved):

addons/editor-theme3/theme3.js
Samq64: It's inside the `if (Blockly.registry)` block. I didn't see an existing `theme3` CSS variable that I could use but I can set `--sa-contextmenu-text` and use that on both editor versions if you prefer.

mxmou: Oops, I read that wrong. Setting `--colour-text` is fine, but you could also use `--sa-block-gray-text`.

PR #8133, opened 2025-02-01
"Instant read indicators on forum category pages" addon

### LINKED ISSUE #8127: Automatically update read status of forum topics on subforum pages
### Why this would be helpful

When going through topics like questions on QaS, I want to go through all the questions I haven't opened yet. I'll sometimes lose track about which topics I've opened and which I haven't. This would solve that.

### How this addon works

This addon would automatically unbold any forum topic links on the subforum page once that topic is opened in some tab.

### Possible alternatives

_No response_

### Additional context

_No response_

mybearworld:
Resolves #8127

### Changes

Adds a new addon that automatically marks topics as read on forum category pages.

### Reason for changes

When going through topics like questions on QaS, I want to go through all the questions I haven't opened yet. I'll sometimes lose track about which topics I've opened and which I haven't. This would solve that.

### Tests

Tested on Firefox

### DIFF
--- a/addons/live-read-topics/addon.json
+++ b/addons/live-read-topics/addon.json
@@ -1,0 +1,24 @@
+{
+  "name": "Instant read indicators on forum category pages",
+  "description": "Automatically shows a topic as read when clicking on it on a forum category's page without requiring a reload.",
+  "versionAdded": "1.42.0",
+  "tags": ["community", "forums"],
+  "credits": [
+    {
+      "link": "https://scratch.mit.edu/mybearworld",
+      "name": "mybearworld"
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "subforums.js",
+      "matches": ["https://scratch.mit.edu/discuss/*/"]
+    },
+    {
+      "url": "topics.js",
+      "matches": ["topics"]
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true
+}
--- a/addons/live-read-topics/addon.json
+++ b/addons/live-read-topics/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Instant read indicators on forum category pages",
+  "name": "Instant forum read indicators",
   "description": "Automatically shows a topic as read when clicking on it on a forum category's page without requiring a reload.",
   "versionAdded": "1.42.0",
   "tags": ["community", "forums"],
--- a/addons/live-read-topics/addon.json
+++ b/addons/live-read-topics/addon.json
@@ -19,6 +19,12 @@
       "matches": ["topics"]
     }
   ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["https://scratch.mit.edu/discuss/*/"]
+    }
+  ],
   "dynamicEnable": true,
   "dynamicDisable": true
 }
--- a/addons/live-read-topics/style.css
+++ b/addons/live-read-topics/style.css
@@ -1,4 +1,4 @@
-.tcl:not(highlighted-necropost) .sa-live-read-topics-new-posts {
+.tcl:not(.highlighted-necropost) .sa-live-read-topics-new-posts {
   display: none;
 }
 
--- a/addons/live-read-topics/style.css
+++ b/addons/live-read-topics/style.css
@@ -1,0 +1,11 @@
+.tcl:not(highlighted-necropost) .sa-live-read-topics-new-posts {
+  display: none;
+}
+
+/** necropost-highlight compatibility */
+.highlighted-necropost .sa-live-read-topics-new-posts {
+  color: currentcolor !important;
+  font-weight: normal !important;
+  cursor: text;
+  text-decoration: none !important;
+}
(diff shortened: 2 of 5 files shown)


addons/live-read-topics/addon.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Instant read indicators on forum category pages",
+  "description": "Automatically shows a topic as read when clicking on it on a forum category's page without requiring a reload.",
```
mybearworld, line 3: I'm not sure the name and description are good yet

WorldLanguages:
Cool, first use of BroadcastChannel on the project.

addons/live-read-topics/topics.js
```diff
@@ -0,0 +1,5 @@
+export default async function ({ addon, msg, console }) {
+  if (addon.self.disabled) return;
```
WorldLanguages, line 2: Is this needed? That is quite defensive. 😄 
Sure, you can get a situation where this returns, but that's unlucky

addons/live-read-topics/addon.json
WorldLanguages: Both look OK to me.

WorldLanguages (approved):

addons/live-read-topics/topics.js
WorldLanguages: IMO dynamicDisable:true means nothing for this addon, so it's fine to set dynamicDisable:true in the manifest without ever using addon.self.disabled or the disabled/reenabled events.

addons/live-read-topics/subforums.js
```diff
@@ -0,0 +1,11 @@
+export default async function ({ addon, msg, console }) {
+  new BroadcastChannel("sa-live-read-topics").addEventListener("message", (e) => {
+    const el = document.querySelector(`.tclcon a[href='/discuss/topic/${e.data}/']`);
+    if (!el) return;
+    el.parentElement.classList.add("topic_isread");
+    const icon = el.parentElement.parentElement.parentElement.querySelector(".inew");
```
WorldLanguages, line 6:
```suggestion
    const icon = el.closest(".tcl").querySelector(".inew");
```

addons/live-read-topics/topics.js
TheColaber: i dont think 'addon.self.disabled' can ever be true when the file first runs. if so that would be a mistake in our logic.

mybearworld: Ah, good point :P

addons/live-read-topics/addon.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Instant read indicators on forum category pages",
```
Samq64, line 2:
```suggestion
  "name": "Instant forum read indicators",
```

addons/live-read-topics/topics.js
Samq64: > IMO dynamicDisable:true means nothing for this addon

The `if (addon.self.disabled) return` could be re-added but inside the event listener.

Samq64 (commented):
Could the `(New Posts)` link be hidden?

Samq64 (approved):

addons/live-read-topics/topics.js
WorldLanguages: > The `if (addon.self.disabled) return` could be re-added but inside the event listener.

@Samq64 That would also make no sense. A message can only be received if the same addon sent it, and if the same addon sent it, then it's enabled.

addons/live-read-topics/addon.json
WorldLanguages: Sure, that is also fine

WorldLanguages: @mybearworld Up to you

mybearworld:
> Could the `(New Posts)` link be hidden?

Ah, good point. Doing that would cause issues with `necropost-highlighter` though, especially dynamically toggling that addon. I suppose a mutation observer could be used to figure out when `necropost-highlighter` gets disabled and then re-remove the unread indicator?

WorldLanguages:
That sounds complex, addons can be dynamically enabled or disabled for the benefit of the user, but that doesn't mean we should accept any level of complexity in exchange to that

mybearworld:
Latest commit does not seem to work on Zen (Firefox), works fine on Edge though. The "Necropost?" link `necropost-highlighter` adds gets `display: none` applied to it even though there's an explicit rule that it shouldn't get that. When you edit the `inline` in the CSS in the dev tools at all, it'll work - even if you set it back to `inline`. I can't figure out why it's doing that at all

Samq64:
> Latest commit does not seem to work on Zen (Firefox), works fine on Edge though.

Could it be a cache issue?

mybearworld:
> Could it be a cache issue?

Hopefully?

addons/live-read-topics/style.css
```diff
@@ -0,0 +1,11 @@
+.tcl:not(highlighted-necropost) .sa-live-read-topics-new-posts {
```
Samq64, line 1:
```suggestion
.tcl:not(.highlighted-necropost) .sa-live-read-topics-new-posts {
```

PR #8136, opened 2025-02-02
`auto-follow-topics`: use internal DjangoBB feature

### LINKED ISSUE #8135: `auto-follow-topics`: Use the internal djangobb checkbox instead of following via javascript
### Why this would be helpful

This would fix a few bugs with `auto-follow-topics` that exist already and follow the topic immediately. This would also provide compatibility with some new addon to add the checkbox.

### My suggestion

Instead of following using javascript, follow on the server's end by automatically selecting an invisible checkbox that does this.

### Possible alternatives

Just leave it as is.

### Additional context

_No response_

ampelctd:
Resolves #8135

### Changes

Uses the internal follow discussion checkbox instead

### Reason for changes

This makes following faster and fixes a few issues with our current implementation of the addon

### Tests

Tested using firefox 134.0.2. No known issues.

### DIFF
--- a/addons/auto-follow-topics/userscript.js
+++ b/addons/auto-follow-topics/userscript.js
@@ -1,22 +1,6 @@
 export default async function ({ addon, console, msg }) {
-  const getTopics = () => JSON.parse(sessionStorage.getItem("auto-follow-topics") || "[]");
-  const setTopics = (topics) => sessionStorage.setItem("auto-follow-topics", JSON.stringify(topics));
-  const postError = () => Boolean(document.querySelector(".errorlist"));
-  const topicID = location.href.split("/")[5];
-
-  document.querySelector("[name=AddPostForm]")?.addEventListener("click", (event) => {
-    const ids = getTopics();
-    ids.push(topicID);
-    setTopics(ids);
-  });
-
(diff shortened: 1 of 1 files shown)


mybearworld:
I didn't test this, but I'm assuming this doesn't handle dynamic disable right now?

ampelctd:
> I didn't test this, but I'm assuming this doesn't handle dynamic disable right now?

Nope, I don't believe so. I'll try working on that now though.

Samq64:
Could the checkbox be made visible and have a label?

ampelctd:
> Could the checkbox be made visible and have a label?

Maybe in another addon. This is just a behaviour change

Samq64 (approved):
> Maybe in another addon.

I think it belongs in this addon but it can be done later.

WorldLanguages:
@Chiroyce1 Do you have any opinions on this?

Samq64:
Showing the toggle could be also be an addon setting.

Chiroyce1:
@WorldLanguages LGTM, its a lot better than my approach which was just manually clicking the follow button, this also reduces the number of requests as its included with post submission.

PR #8145, opened 2025-02-09
blocks2image: improve UI

mxmou:
### Changes

* Makes the labels of buttons and context menu items shorter and changes them to title case to be more consistent with Scratch and other addons.
* Hides the modal while the image is being loaded to avoid layout shifts. A loading spinner is shown instead:
image
* Changes the modal title to be more similar to the name of the context menu item that opens the modal.
* Changes the button row so that there's only one primary button:
image
* Changes the copy icon in the copy-reporters addon to one that's more similar to Scratch's icons:
image

### Tests

Tested on Edge and Firefox.


Samq64:
The SVGs have rendering issues in some non-browser image viewers. Could PNG be the primary option?

WorldLanguages (approved):

Issue #8093, opened 2025-01-16
Use AGPL-3.0 license going forward


DNin01:
Based on discussion #7958

### Why?

- Same license as Scratch uses since about 2 months ago

WorldLanguages:
@Hans5958 Have you taken a look at the discussion?

Hans5958:
Yes, I do, and I I still don't think it applies here. We can still have GPL-3.0 even if Scratch does AGPL-3.0.

GarboMuffin:
Not a lawyer, haven't talked to one, could be entirely wrong. I'll just summarize what I believe:

1. Some addons (eg. folders) are very closely tied to Scratch internals, and some addons (eg. onion-skinning) copy and paste non-trivial amounts of code directly out of Scratch. I think it is reasonable to consider these derivative works of Scratch that would have to be licensed under AGPL-3.0 if they were written today based on AGPL-3.0 code from Scratch. If Scratch Addons were to stay on GPL-3.0 only, it might not be possible to update these addons with new AGPL-3.0 code.

2. Scratch Addons changing from GPL-3.0 to AGPL-3.0 for future contributions does not have downsides beyond the README requiring an extra sentence to explain the change.

WorldLanguages:
Hans, have you taken a look at the source code of some Scratch editor addons? I agree with you that there's no real reason to follow Scratch with their license change for code written by ourselves. But what about editor addons? More often than not, the monkeypatching needed to make an editor addon work properly requires borrowing some Scratch code into our own codebase.

Hans5958:
Garbo, the reason that I'm doubting myself is that you only mentioned them "[interact[ing] so closely](https://github.com/ScratchAddons/ScratchAddons/discussions/7958#discussioncomment-11850612)", which I'm thinking it as "linking an AGPL-3.0 library with a GPL 3.0 project". Several Stack Overflow answers pointed out that you have to put it on AGPL-3.0 (mixing allowed, etc), but having it just solely a dependency... I think it should be fine? I'm not convinced that doing `npm install` an AGPL-3.0 library would need making my project as AGPL-3.0 (or GPL-3.0 library + GPL-3.0 project, same thing)

[...]

PR #8147, opened 2025-02-10
Add Scratch packages to libraries

mxmou:
As I suggested in https://github.com/ScratchAddons/ScratchAddons/issues/8093#issuecomment-2614044040, this PR adds Scratch packages to the `libraries` array for addons that use icons or a significant amount of code from Scratch. I also added links to a few code comments that referenced Scratch code.


webpages/settings/index.html
```diff
@@ -378,7 +378,7 @@ <h1 v-cloak>{{ msg("settings") }}</h1>
         </p>
         <p>
           <a
-            href="./licenses.html?libraries=icu-message-formatter,vue,color-picker-web-component,comlink,Sora,fuse,idb,sortable,Roboto"
+            href="./licenses.html?libraries=icu-message-formatter,vue,color-picker-web-component,comlink,Sora,fuse,idb,sortable,Roboto,scratch-gui"
```
Samq64, line 381: All the other licenses listed here are for the settings page. Is there a specific reason for including `scratch-gui`?

WorldLanguages: Good catch

mxmou: The settings page uses the `download-blob.js` module from scratch-gui.

mxmou:
> Linking to the old licenses could become problematic in the future but I'm not sure what the solution is.

We can update the licenses if we use AGPL-licensed Scratch code in the future.

webpages/settings/index.html
WorldLanguages: @mxmou Can we remove that dependency instead? It's unnecessary :P 
The if statements in that file will never be truthy in environments where SA can run

DNin01:
Also, Scratch is moving all its packages into a monorepo called `scratch-editor`. We may need to update our references in the future.

libraries/licenses/GPL-2.0.txt
```diff
@@ -0,0 +1,677 @@
...
+
+  <signature of Ty Coon>, 1 April 1989
+  Ty Coon, President of Vice
+
+This General Public License does not permit incorporating your program into
+proprietary programs.  If your program is a subroutine library, you may
+consider it more useful to permit linking proprietary applications with the
+library.  If this is what you want to do, use the GNU Lesser General
+Public License instead of this License.                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
```
Samq64, line 340: Was the license accidentally pasted twice?

Samq64:
> Also, Scratch is moving all its packages into a monorepo called scratch-editor. We may need to update our references in the future.

At least for now each component is still structured as its own repository with the old licenses included..

libraries/licenses/GPL-2.0.txt
Samq64: Never mind, it's the same as scratch-flash.

Samq64 (approved):

libraries/licenses/GPL-2.0.txt
mxmou: I copied it from scratch-flash and didn't notice it was duplicated. I don't think it makes sense to keep the duplicated text because we might have other GPL-licensed dependencies in the future and they probably didn't make the same error.

DNin01 (approved):

PR #8155, opened 2025-02-14
fix-editor-comments: new Blockly

mxmou:
### Changes

Updates the "keep block comment connections straight" setting of "better editor comments" to work with modern Blockly. Other settings fix Scratch issues that are fixed in the new version of scratch-blocks.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/fix-editor-comments/userscript.js
+++ b/addons/fix-editor-comments/userscript.js
@@ -6,54 +6,73 @@ export default async function ({ addon, console }) {
   });
   const Blockly = await addon.tab.traps.getBlockly();
 
-  /*
-   * When a block with a comment attached is dragged, the comment now properly
-   * stores its updated position.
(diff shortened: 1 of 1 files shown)


Samq64:
The block dragging issue is still on modern Blockly in my testing.

mxmou:
> The block drag positioning issue is still there on modern Blockly in my testing, but it's possible I'm not on the latest commit.

You're right - looks like I didn't test it properly before.

mxmou:
I've done some more testing with block dragging. The bug is actually fixed in new Blockly - dragging a block emits comment move events - but the new position is sometimes not saved because of a different bug (gonfunko/scratch-blocks#84) that causes comments to get new random IDs when switching sprites.
image

This bug is quite serious, so I expect it to be fixed before the Scratch update is released. If not, we can fix it ourselves later.

PR #8165, opened 2025-02-20
editor-searchable-dropdowns: new Blockly

mxmou:
### Changes

Updates "block dropdown search" to be compatible with new Blockly. Also fixes two minor bugs in other addons so that the full-screen error popup doesn't appear in development versions of Scratch.

### Tests

Tested both Scratch versions on Edge and Firefox. Scrolling in long dropdowns is slightly broken on new Blockly, but this isn't caused by the addon.


Samq64 (approved):
Not a big deal but the extra height is removed on modern Blockly:
image

mxmou:
> Not a big deal but the extra height is removed on modern Blockly: image

I think it actually looks better without the empty space - but it might be a good idea to add a message to tell the user that nothing was found, otherwise it looks like all the items just disappeared.

mxmou:
I've added a "No results" message. Now it looks like this on new Blockly:

image

Old Blockly:

image

The spacing is intentionally weird - it's the same as on normal menu items.

Samq64 (approved):

PR #8180, opened 2025-02-28
custom-zoom: new Blockly

mxmou:
Resolves #

### DIFF
--- a/addons/custom-zoom/userscript.js
+++ b/addons/custom-zoom/userscript.js
@@ -1,4 +1,6 @@
 export default async function ({ addon, console }) {
+  const Blockly = await addon.tab.traps.getBlockly();
+
   const speeds = {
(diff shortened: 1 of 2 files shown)


addons/custom-zoom/userscript.js
```diff
@@ -18,6 +20,9 @@ export default async function ({ addon, console }) {
   customZoomAreaElement.addEventListener("mousedown", (e) => {
```
Samq64, line 20: This is not necessary on the new Blockly, right?

Samq64 (approved):

addons/custom-zoom/userscript.js
mxmou: Indeed. I'll add a comment so that we don't forget to remove it after the update is released.

PR #8181, opened 2025-02-28
reorder-custom-inputs: new Blockly

mxmou:
### Changes

Updates "rearrangeable custom block inputs" to work with modern Blockly. The important changes are all in `userscript.js`. `modified-funcs.js` only contains style changes to reflect the updated upstream code.

### Tests


Samq64 (changes requested):
Editing a custom block on modern Blockly after dynamically disabling the addon crashes the editor when calling `originalCreateAllInputs`.

There is also a bug on Scratch's side where editing a custom block clears all uses of its inputs, but that's something that needs to be fixed on their end.

Samq64 (approved):

PR #8183, opened 2025-03-01
editor-preview-comments: New Blockly support and move dark mode styling

Samq64:
### Changes

Adds modern Blockly support `editor-preview-comments`. It was already mostly working but hovering over collapsed comments and the middle section of custom block definitions are now fixed.

Also moves the dark mode support out of `editor-dark-mode` and into the addon.

### Tests

Tested on both editors on Chromium.

### DIFF
--- a/addons/editor-comment-previews/userstyle.css
+++ b/addons/editor-comment-previews/userstyle.css
@@ -17,7 +17,7 @@
   pointer-events: none;
 
   color: var(--editorDarkMode-input-text, #575e75);
-  background-color: var(--editorDarkMode-input-transparent90, rgba(255, 255, 255, 0.9));
+  background-color: var(--editorDarkMode-input-transparent90, rgb(255 255 255 / 90%));
   border-style: none;
   border-radius: 8px;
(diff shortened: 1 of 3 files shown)


addons/editor-comment-previews/userstyle.css
```diff
@@ -16,8 +16,8 @@
   white-space: pre-wrap;
   pointer-events: none;
 
-  color: rgb(87, 94, 117);
-  background-color: rgb(255 255 255 / 90%);
+  color: var(--editorDarkMode-input-text, #575e75);
+  background-color: var(--editorDarkMode-input-transparent90, rgba(255, 255, 255, 0.9));
```
mxmou, line 20:
```suggestion
  background-color: var(--editorDarkMode-input-transparent90, rgb(255 255 255 / 0.9));
```
I think it's better to keep this syntax for colors because it's also used elsewhere in this file.

mxmou (approved):

mxmou (approved):

PR #8200, opened 2025-03-11
Include settings and presets in search

Samq64:
> I wanted to find "separate lists and variables" to turn it off, though showing "Separate" didn't work.

Adds setting names, `select` setting values and preset names to the settings page search results.

Tested on Chromium in English only.

### DIFF
--- a/webpages/settings/data/fuse-options.js
+++ b/webpages/settings/data/fuse-options.js
@@ -29,10 +29,6 @@ export default {
       weight: 0.4,
     },
     {
-      name: "_english.settings.name",
-      weight: 0.2,
-    },
-    {
       name: "credits.name",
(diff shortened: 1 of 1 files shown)


WorldLanguages:
I approve the idea.

About the weights, I'm not sure if settings.name should be lower.

WorldLanguages:
Does _english.settings actually exist? I would double check that.

Samq64:
> Does _english.settings actually exist? I would double check that.

You're right, it's not. Should I add it or just remove it from the fuse options?

WorldLanguages:
Personally I would avoid adding additional _english properties

WorldLanguages (approved):

PR #8205, opened 2025-03-18
Scratch Potatoes

### LINKED ISSUE #8196: April Fools
TODO: will Scratch be updated with the frequent April Fools changes that are incompatible with some addons and may crash the editor?

About doing our own April Fools joke, this is what we've done each year so far. 

- 2021: dango rain on Scratch profiles
- 2022: Scratch Avocados and green accent color on popup
- 2023: "Featured dangos" row in front page
- 2024: Scratch Haddocks and blue accent color on popup
- 2025 ???

Samq64:
Resolves #8196

### Changes

image

### Reason for changes

April Fools!

### Tests

Tested on Chromium with `faketime`.


libraries/common/global-theme.js
```diff
@@ -6,6 +6,13 @@ export default function () {
     const favicon = document.getElementById("favicon");
     if (favicon) favicon.href = chrome.runtime.getURL("/images/icon-blue.png");
   }
+  const now = Date.now() / 1000;
+  if (now < 1743609600 && now > 1743436800 && ["settings", "popup"].includes(location.pathname.split("/")[2])) {
+    // March 31, 2025 12:00:00 GMT to April 2, 2025 12:00:00 GMT
```
WorldLanguages, line 11: Perfect! Our April Fools have always been 45 hours long

WorldLanguages:
Will release with enough time so that this reaches users on March 31th

Also need to check whether we want to disable some editor addons on the same period

libraries/common/global-theme.js
WorldLanguages: @Samq64 Actually isn't this off by 4 hours?

WorldLanguages (approved):
Perfect, just need to double check if the timestamps are using the correct timezone

libraries/common/global-theme.js
Samq64: Oops, it looks like the tool I used to generate the timestamps used local time.

WorldLanguages (approved):

Waakul:
The mention of the word "addons" or "addon" shoud've been replaced to potatoes...

PR #8207, opened 2025-03-21
Remove dead code from devtools addons

### LINKED ISSUE #7965: Remove Devtools Extension Code
There is still some code that is only relevant to the standalone devtools extension or for transitioning from it to Scratch Addons. Now that the extension is archived and unlisted and with Chrome disabling MV2 extensions that code should be removed eventually.

Samq64:
Resolves #7965

### Changes

Removes the devtools extension checks and includes most of the dead code removal from #7519 so it's easier to focus on the actual changes.

### Reason for changes

It's dead code and the devtools extension has been unlisted and unmaintained for a long time now.

### Tests

Tested on Chromium. While the 2 extensions can still be installed at the same time any remaining users will probably never install Scratch addons until the devtools extension completely breaks.


WorldLanguages (approved):

Samq64:
I forgot about the find bar.

WorldLanguages:
> I forgot about the find bar.

Do we need to reopen #7965 then?

Samq64:
No, I just missed some commented out code, it's not a big deal.

PR #8222, opened 2025-03-28
1.42 update notificaton

Samq64:
image


Samq64:
If there are no objections I'll merge this and #8224 in a few hours so the strings can be pulled by the action as soon as possible.

WorldLanguages:
Sure

PR #8224, opened 2025-03-28
Mention costume shortcut bug was fixed

Samq64:
Tested on Chromium

### DIFF
--- a/addons/costume-editor-shortcuts/addon.json
+++ b/addons/costume-editor-shortcuts/addon.json
@@ -15,5 +15,10 @@
   ],
   "versionAdded": "1.41.0",
   "tags": ["editor", "costumeEditor", "recommended"],
+  "latestUpdate": {
+    "version": "1.42.0",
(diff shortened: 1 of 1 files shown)


addons/costume-editor-shortcuts/addon.json
```diff
@@ -15,5 +15,10 @@
   ],
   "versionAdded": "1.41.0",
   "tags": ["editor", "costumeEditor", "recommended"],
+  "latestUpdate": {
+    "version": "1.42.0",
+    "isMajor": true,
+    "temporaryNotice": "The bug where the Ctrl key was ignored when reading shortcuts was fixed."
```
CST1229, line 21:
```suggestion
    "temporaryNotice": "The bug where shortcuts were still triggered while holding the Ctrl key was fixed."
```
maybe it should be reworded like this? something that somewhat more clearly says that this fixes Ctrl+A and stuff

PR #8228, opened 2025-03-29
discuss-button: Add extension settings preset

Samq64:
### Changes

Adds a Scratch Addons settings preset to the customizable navigation bar.

### Reason for changes

It's not obvious it's possible and it could be useful for those who don't pin the extension. I used the Scratch URL instead of the scratchaddons.com one to avoid the redirect message.

### Tests

Tested on Chromium.

### DIFF
--- a/addons/discuss-button/addon.json
+++ b/addons/discuss-button/addon.json
@@ -87,6 +87,10 @@
         {
           "name": "Wiki",
           "values": { "name": "Wiki", "url": "https://scratch-wiki.info/" }
+        },
+        {
+          "name": "Scratch Addons settings",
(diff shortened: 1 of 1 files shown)


WorldLanguages:
I don't think this makes sense, in order to pick the preset you have to use the settings page anyway

Users that don't pin the extension can search on google "scratch addons settings" every time

Samq64:
What about putting a direct link on https://scratchaddons.com/extension-settings/?

WorldLanguages:
@Samq64 That page already automagically redirects to the settings page, but only if the extension was installed through the store.

Samq64:
It doesn't on Firefox but that's fine.

PR #8235, opened 2025-04-02
Fix huge lag spike when opening insert blocks by name menu in TurboWarp with some extensions

Tacodiva:
This fixed an issue where with certain TurboWarp extensions installed, opening the middle click popup menu would cause a huge lag spike, up to 20-30 seconds. I forgot to put a limit on one of the loops, nothing in normal Scratch should be effected as the limit can't be hit by the vanilla blocks.

### DIFF
--- a/addons/middle-click-popup/WorkspaceQuerier.js
+++ b/addons/middle-click-popup/WorkspaceQuerier.js
@@ -762,7 +762,7 @@ class TokenTypeBlock extends TokenType {
 
   /**
    * @param {QueryInfo} query
-   * @param {number} idxi
+   * @param {number} idx
    * @param {number} depth
    * @returns
    */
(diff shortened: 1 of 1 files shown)


addons/middle-click-popup/WorkspaceQuerier.js
```diff
@@ -749,6 +751,13 @@ class TokenTypeBlock extends TokenType {
     };
 
     enumerateStringForms();
+
+    if (this.stringForms.length >= WorkspaceQuerier.MAX_STRING_FORMS) {
+      console.log(
+        "Warning: Block '" + this.block.id + "' has too many string forms. Search results may not be very good."
```
Samq64, line 757: The console wrapper works with warnings as well.
```suggestion
      console.warn(
        "Block '" + this.block.id + "' has too many string forms. Search results may not be very good."
```

Tacodiva: haha I actually didn't use `console.warn` for this same thing in multiple places in the original commit. I've fixed them up now. Reminds me of the classic addage;
> Ask a programmer to review 10 lines of code, they'll find 10 issues. Ask them to do 500 lines and they'll say it looks good.

WorldLanguages (approved):

PR #8241, opened 2025-04-03
blocks2image: new Blockly

mxmou:
### Changes

Makes the Save blocks as image addon compatible with modern Blockly.

### Tests

Tested both Scratch versions on Edge and Firefox. When "Save All as Image" is used on new Blockly, comments are included in the image, which I think makes sense.


Samq64:
When comments are expanded it just displays the placeholder text instead of the actual message.

Samq64 (approved):

PR #8242, opened 2025-04-03
Cleanup blocks improvements

### LINKED ISSUE #4131: "Clean Up Blocks +" and script-snap use different grids
I'm not sure what should be done here.
If we change the grids used by `Clean Up Blocks +`, cleaning up an old project that was already cleaned, would shift all blocks by a bit.
If we change the grid used by script-snap, it would stop using the vanilla Scratch dot grid, which is the grid that makes most sense. And it would also suffer of the above problem, but script-snap is a newer addon.

Samq64:
Resolves #4131
Progress on #6671

### Changes

- The unsued variable and related prompts are now optional and disabled by default
- Aligns scripts to the same grid as `script-snap`
- Don't modify the unused block messages directly
- Changes a bunch of `let`s to `const`s
- ~~Renames `unusedLocals` to `unusedVariables`~~
- Gets the workspace more directly
- Enables the addon by default

### Reason for changes

It's annoying getting so many prompts and cleanup not using the same grid.

### Tests

Tested on Chromium.

### DIFF
--- a/addons-l10n/en/editor-cleanup-plus.json
+++ b/addons-l10n/en/editor-cleanup-plus.json
@@ -1,6 +1,6 @@
 {
   "editor-cleanup-plus/orphaned": "{count, plural, one {Cleanup Plus: Delete 1 orphaned reporter block?} other {Developer tools: Delete # orphaned reporter blocks?} }",
-  "editor-cleanup-plus/unused-var": "{count, plural, one {Cleanup Plus: Delete 1 unused local variable? Here it is:\n} other {Developer tools: Delete # unused local variables? Here they are:\n} }",
-  "editor-cleanup-plus/unused-list": "{count, plural, one {Cleanup Plus: Delete 1 unused local list? Here it is:\n} other {Developer tools: Delete # unused local lists? Here they are:\n} }",
+  "editor-cleanup-plus/unused-var": "{count, plural, one {Cleanup Plus: Delete 1 unused local variable? Here it is:\n{names}} other {Developer tools: Delete # unused local variables? Here they are:\n{names}} }",
+  "editor-cleanup-plus/unused-list": "{count, plural, one {Cleanup Plus: Delete 1 unused local list? Here it is:\n{names}} other {Developer tools: Delete # unused local lists? Here they are:\n{names}} }",
   "editor-cleanup-plus/clean-plus": "Clean up Blocks +"
 }
--- a/addons-l10n/en/editor-cleanup-plus.json
+++ b/addons-l10n/en/editor-cleanup-plus.json
@@ -1,6 +1,6 @@
 {
   "editor-cleanup-plus/orphaned": "{count, plural, one {Cleanup Plus: Delete 1 orphaned reporter block?} other {Developer tools: Delete # orphaned reporter blocks?} }",
-  "editor-cleanup-plus/unused-var": "{count, plural, one {Cleanup Plus: Delete 1 unused local variable? Here it is:\n{names}} other {Developer tools: Delete # unused local variables? Here they are:\n{names}} }",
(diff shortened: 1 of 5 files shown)


addons-l10n/en/editor-cleanup-plus.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor-cleanup-plus/orphaned": "{count, plural, one {Cleanup Plus: Delete 1 orphaned reporter block?} other {Developer tools: Delete # orphaned reporter blocks?} }",
-  "editor-cleanup-plus/unused-var": "{count, plural, one {Cleanup Plus: Delete 1 unused local variable? Here it is:\n} other {Developer tools: Delete # unused local variables? Here they are:\n} }",
-  "editor-cleanup-plus/unused-list": "{count, plural, one {Cleanup Plus: Delete 1 unused local list? Here it is:\n} other {Developer tools: Delete # unused local lists? Here they are:\n} }",
+  "editor-cleanup-plus/unused-var": "{count, plural, one {Cleanup Plus: Delete 1 unused local variable? Here it is:\n{names}} other {Developer tools: Delete # unused local variables? Here they are:\n{names}} }",
+  "editor-cleanup-plus/unused-list": "{count, plural, one {Cleanup Plus: Delete 1 unused local list? Here it is:\n{names}} other {Developer tools: Delete # unused local lists? Here they are:\n{names}} }",
```
mxmou, line 4:
```suggestion
  "editor-cleanup-plus/unused-var": "{count, plural, one {Cleanup Plus: Delete 1 unused local variable? Here it is:\n{names}} other {Cleanup Plus: Delete # unused local variables? Here they are:\n{names}} }",
  "editor-cleanup-plus/unused-list": "{count, plural, one {Cleanup Plus: Delete 1 unused local list? Here it is:\n{names}} other {Cleanup Plus: Delete # unused local lists? Here they are:\n{names}} }",
```

addons/editor-cleanup-plus/addon.json
```diff
@@ -10,6 +10,14 @@
       "link": "https://scratch.mit.edu/users/Chrome_Cat/"
     }
   ],
+  "settings": [
+    {
+      "name": "Find unused variables",
```
mxmou, line 15: A setting description might be helpful here (to explain that this shows a prompt and that it also removes reporter blocks).

addons/editor-cleanup-plus/userscript.js
```diff
@@ -41,102 +41,95 @@ export default async function ({ addon, console, msg, safeMsg: m }) {
...
-          if (i > 0) {
-            message += ", ";
-          }
-          message += orphan.name;
-        }
-        if (confirm(message)) {
-          for (const orphan of unusedLocals) {
-            workspace.deleteVariableById(orphan.getId());
-          }
+    if (unusedVariables.length > 0) {
```
mxmou, line 99: I think the name `unusedLocals` makes sense - it includes both variables and lists, but only local (For this sprite only) ones.

mxmou (approved):

Samq64:
Should migration code be added to enable the setting for exsisting users?

mxmou:
> Should migration code be added to enable the setting for existing users?

I don't think that's necessary, but now that the addon is enabled by default, we need to add logic to disable it for people who have turned off the "enhance clean up blocks" setting in devtools.

WorldLanguages:
> Could someone take one more look at this (specificly the migration code) before merging?

How is the settings migration supposed to work exactly?

Samq64:
> How is the settings migration supposed to work exactly?

It disables the cleanup addon (which this PR enables by default) when the user has the devtools setting turned off.

WorldLanguages:
@Samq64 I just pushed new settings migration code and tested both cases.

My code considers the case where editor-devtools is disabled as well.

PR #8244, opened 2025-04-04
New Addon: Edit variables or lists from context menus

### LINKED ISSUE #8211: "Edit" option in variable block context menu
### Why this would be helpful

This would be useful when wanting to edit variables in the block editor without having to find the variable in the block flyout.

### How this addon works

This addon adds an "Edit" option to all variable and list block's context menus.

### Possible alternatives

This could be added onto the `swap-local-global` addon.
Also, if we have this addon, we should consider removing the `Convert to "For this sprite only"` option from the `swap-local-global` addon, or at least have it a togglable option #3773

### Additional context

Image

AeroKoder:
Resolves #8211 

### Changes

Added a "Edit" option to the context menus on the variables and list blocks.
Image

### Reason for changes

This would be useful when wanting to edit variables in the block editor without having to find the variable in the block flyout.

### Tests

Tested on Windows with Edge Version 134.0.3124.93 (Official build) (64-bit)


DNin01:
> Unsure if useful or just clutters context menu

It is an optional addon, so if you don't like it, you can leave it off. The question is, would the addon be worth adding?

FAReTek1:
> On, like, the block your mouse is hovering over?
You could do that. Or you could click and do keybind like how ctrl+c works.

PR #8254, opened 2025-04-07
Created "Upload Font" Addon 

### LINKED ISSUE #8253: Upload font as sprite for text engines
### Why this would be helpful

This would help users who build their own text systems and want to use different fonts without manually creating each character as a costume.

### How this addon works

This addons adds a menu item to the upload spite menu and allows the user to select a font. Then, it creates a new sprite where each character is split into a different costume along with some starter code to start displaying text.

### Possible alternatives

_No response_

### Additional context

Image

dylan-dang:
Resolves #8253

### Changes

Adds the ability to upload a font as a sprite for use in text engine by splitting each character into svg costumes.

https://github.com/user-attachments/assets/c3ec3f75-6e65-467c-80f2-c020dccd3c2e

### Reason for changes

This would help users who build their own text systems and want to use different fonts without manually creating each character as a costume.

### Tests

Tested on Edge


Samq64:
#8036 is an alternative which has more font customization but doesn't currently integrate well with the editor.

PR #8259, opened 2025-04-10
middle-click-popup: new Blockly fix

mxmou:
### Changes

Blockly has renamed the `flyout_` attribute of the toolbox to `flyout` without an underscore, which broke the middle click popup addon. This PR updates it.

### Tests

Tested on Edge.


WorldLanguages:
> Redux seems to be broken on modern Blockly on Firefox but that's a separate issue.

Whoops... Only on Firefox?

PR #8260, opened 2025-04-10
editor-devtools: new Blockly

mxmou:
### Changes

Updates "developer tools" to work with modern Blockly.

### Tests

Tested on Edge and Firefox.


Samq64 (commented):
The copy block and copy all context menu items do the same thing on the new Blockly. Could that either be fixed or the copy all option hidden on the new editor?

Samq64 (approved):

PR #8270, opened 2025-04-17
Forumoji addon

### LINKED ISSUE #4097: Emodjiango (might turn Scratchmoji soon)
### Additional context

I have a Scratch forum topic where we create Scratch-forum-style emojis, and we're trying to make it of all the Unicode emojis.

Since emojis are not supported on the forums, it'd be neat if we could add an addon that replaces the normal emojis with the Scratch ones.

I'm working on a JSON file with most of the codepoints of the current one, and the repository with all the emojis is here: https://github.com/lopste/emodjiango (most recent emojis)

And the scratch forum topic is here: https://scratch.mit.edu/discuss/topic/557083/

mybearworld:
Continuation of #6006, I've accidentally deleted my fork
Resolves #4097

### Changes

Add an addon called "Forumoji emoji when posting on the forums" (internally `forumoji`) that replaces emojis with their Forumoji variants.

The forumoji addon.

### Reason for changes

Two reasons:
1. It's more convenient than having to go to the Forumoji web page and having to search for the emoji you want.
2. It avoids accidentally putting an emoji into the post to mess up the page's markup.

### Tests

Tested on Microsoft Edge 114.0.1807.6 (Official build) dev (64-bit).


Joeclinton1:
Is this done?

addons/forumoji/userscript.js
```diff
@@ -0,0 +1,29 @@
+export default async function ({ addon, msg }) {
+  const postButton = document.querySelector("#djangobbwrap [method=post] [type=submit]");
+  const forumoji = Object.keys(
+    await (await fetch("https://raw.githubusercontent.com/vercte/forumoji/main/assets/emoji.json")).json()
```
Samq64, line 4: Conisdering this file is almost 100KB maybe the post should be checked for emojis first? But then again a lot of images are larger than that.

CST1229: that file could probably also just be compressed by a lot (e.g by minifying the json)

Samq64: I opened an issue for that.

Samq64: Could this be fetched from `https://gh.vercte.net/forumoji/assets/emoji.js` instead?

Issue #1199, opened 2021-01-04
Collapsible blocks


areeq0325:
**Describe the new addon**
Allows you to collapse hat, c, and reporter blocks.

**Link to the new addon**
I don't have one.

Explosion-Scratch:
I believe someone made this already (As a userscript)

Explosion-Scratch:
Here it is https://scratch.mit.edu/discuss/topic/206779/?page=1#post-2076486

areeq0325:
Maybe this addon could incorporate that.

ErrorGamer2000:
*continued from #2329* I meant the ENTIRE stack as one, not just individual blocks. It would leave the hat unchanged and shrink all of the children under it.

PR #8275, opened 2025-04-19
New addon:

RedPanda247:
Resolves #

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -159,6 +159,8 @@
   "no-sprite-confirm",
   "costume-editor-shortcuts",
   "live-read-topics",
+  "hide-blocks",
+
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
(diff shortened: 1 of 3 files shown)


Samq64:
I think an addon to collapse scripts (#1199) would be more useful.

RedPanda247:
> An addon to collapse scripts (#1199) would be easier to figure out how to get the blocks back compared to just hiding them entirely.

Yes, that was my first idea, but then i realized something... i have no idea at all how to do that😅 so i decided to start simple, but if someone knows how to do that and would be willing to help then id be all for it!

RedPanda247:
I am pausing work on this project for now due to a lack of available time to work on the project.
Feel free to pick up where I've left off and continue the project how you see fit, any help in development or suggestions for the project would be appreciated but for now i will not be working on it.

medlans:
> An addon to collapse scripts (#1199) would be easier to figure out how to get the blocks back compared to just hiding them entirely.

There was actually something like this in the 3.0 beta: https://scratch.mit.edu/discuss/topic/717857/?page=2#post-7585800

Though, it's not exactly the same it seems like

PR #8286, opened 2025-04-23
New feature for hide-delete-button addon: show trash icon when a key is holding

### LINKED ISSUE #8285: New setting for hide-delete-button addon: show trash icon when a key is holding
### Why this would be helpful

The hide-delete-button addon is very useful to avoid accidentally deleting assets, but it becomes tedious to delete them by always having to right-click and then "delete".

### My suggestion

A new setting with re-show of trash when shift or ctrl key is held. Like on Discord.

https://github.com/user-attachments/assets/880c77d1-7fa5-438d-b003-427c8a7f0ce7

### Possible alternatives

_No response_

### Additional context

_No response_

Valmontechno:
Resolves #8285

### Changes

Added a new setting to choose a key between Shift and Ctrl. When this key is held down, the trash button reappears.

https://github.com/user-attachments/assets/9d3d272f-b3ee-4ed0-8f2c-4a0fa76a2374

### Tests

Google Chrome: 135.0.7049.97 (Build officiel) (64 bits) (cohort: Control)
Microsoft Edge: 135.0.3179.85 (Version officielle) (64 bits)

### DIFF
--- a/addons/hide-delete-button/show-trash-shortcut.css
+++ b/addons/hide-delete-button/show-trash-shortcut.css
@@ -1,3 +1,4 @@
-.sa-show-trash div[class*="delete-button_delete-button_"] {
+.sa-show-trash div[class*="target-pane_target-pane_"]:hover div[class*="delete-button_delete-button_"],
+.sa-show-trash div[class*="selector_wrapper_"]:hover div[class*="delete-button_delete-button_"] {
   display: block !important;
 }


Samq64:
Does it make sense to have two shortcuts to choose from? I don't think there needs to be a setting at all, it could just slways be shift.

valmontechno:
@Samq64
I removed the Ctrl shortcut, but I'm wondering if I should replace the setting with a boolean or if I should leave it as a select.
I think that since the setting has a different function than the others, it should have a different form.

image

Samq64:
I don't think it needs to be optional at all but if you want to keep it a setting it should probbably just be a toggle.

DNin01:
> I removed the Ctrl shortcut, but I'm wondering if I should replace the setting with a boolean or if I should leave it as a select. I think that since the setting has a different function than the others, it should have a different form.

I think the select setting is good.

mxmou:
> I removed the Ctrl shortcut, but I'm wondering if I should replace the setting with a boolean or if I should leave it as a select. I think that since the setting has a different function than the others, it should have a different form.

I don't see a good reason to turn this off, so it probably shouldn't even be a setting. If it turns out many people would like to disable it, we can make it optional in a later update.

valmontechno:
Alright, if it is no longer a setting it must be indicated in the description of the addon. Like it's a new feature it should be indicated in `temporaryNotice`, however it'll desappear with the next update right? So maybe it should be indicated in `info`?

image

mxmou:
The last sentence of description explains how to delete sprites when this addon is enabled. It can be expanded to mention holding Shift:
"They can still be deleted using the right click context menu or by holding the Shift key."

The `temporaryNotice` could say something like this:
"You can now hold Shift to show delete buttons."

mxmou (approved):

Samq64 (approved):

Samq64:
Ideally the shift shortcut would only be shown when interacting with sprite/costume panes but this is also fine.

valmontechno:
@Samq64
I made it so that the shortcut is only activated when that sprite or costume pane is hovered.

PR #8297, opened 2025-04-26
New feature for folders addon: delete all assets in folder

Valmontechno:
### Changes

Add an option in the editor context menu: "Delete all in folder". It allow to delete all the sprites, costumes or sounds in the folders at the same time.
The deleted sprites can be restored by going to Edit → Restore.

[scrnli_XP7o3y8jtLk3C7.webm](https://github.com/user-attachments/assets/06ec534f-f415-4499-b681-41c5b20e2569)

### Reason for changes

Convenience, it's useful.

### Tests

Google Chrome: 135.0.7049.115
Microsoft Edge: 135.0.3179.98


addons/folders/userscript.js
```diff
@@ -1370,34 +1487,63 @@ export default async function ({ addon, console, msg }) {
...
-    verifyVM(vm);
-    patchSortableHOC(sortableHOCInstance.constructor, TYPE_SPRITES);
-    patchSpriteSelectorItem(spriteSelectorItemInstance.constructor);
-    sortableHOCInstance.saInitialSetup();
-    patchVM();
+    addon.tab
+      .waitForElement("[class^='sprite-selector_sprite-wrapper']", {
+        reduxCondition: (state) => !state.scratchGui.mode.isPlayerOnly,
+      })
+      .then((spriteSelectorItemElement) => {
```
Samq64, line 1494: Correct me if I'm wrong but I don't think this is necessary and it clutters the diffs. Same with `selectorListItem` below.

addons/folders/userscript.js
```diff
@@ -833,6 +865,83 @@ export default async function ({ addon, console, msg }) {
...
+            {
+              useEditorClasses: true,
+            }
+          )
+        ) {
+          if (type === "sprite") {
+            const targets = vm.runtime.targets;
+            restorationFunctions = [];
+
+            for (let i = targets.length - 1; i > -1; i--) {
```
Samq64, line 884: This could just be a `for...of` loop since only costumes need the first one preserved.

addons-l10n/en/folders.json
Samq64, line 1: Ideally the deletion messages would be combined but I'm not sure the best way to do that since passing a variable directly into msg it not usually a great idea. Also, nitpicky but could the diffs be cleaned up?

addons/folders/userscript.js
Samq64: Never mind, it's for the restore ordering.

mxmou: This is probably necessary because deleting a sprite removes it from `targets`.

addons/folders/userscript.js
Valmontechno: Sorry, I don't speak English very well and I'm not sure I understand what you mean.
If you mean why I replace `await` with `then`, it is because the `while (true)` loop must be executed at the beginning and `await` pauses the script execution.

addons/folders/userscript.js
Valmontechno: Yes it's necessary to avoid deleting the wrong asset because of the index shift when one is deleted.

addons/folders/userscript.js
```diff
@@ -1370,34 +1487,63 @@ export default async function ({ addon, console, msg }) {
...
+      if (!e.action || e.action.type !== "scratch-gui/restore-deletion/RESTORE_UPDATE") return;
+      restoreButtonMsg = null;
     });
-    const sortableHOCInstance = getSortableHOCFromElement(selectorListItem);
-    verifySortableHOC(sortableHOCInstance);
-    patchSortableHOC(sortableHOCInstance.constructor, TYPE_ASSETS);
-    sortableHOCInstance.saInitialSetup();
+
+    while (true) {
+      const restoreButton = await addon.tab.waitForElement(
```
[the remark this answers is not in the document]
Valmontechno: @Samq64 Thanks, I made the change.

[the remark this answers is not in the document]
Samq64: Sorry for not mentioning this earlier but could the 2 event listeners also be combined?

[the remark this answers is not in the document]
Valmontechno: @Samq64 Oh yeah, I hadn't thought of that.

addons-l10n/en/folders.json
mxmou: I don't think it's possible to combine them without making things more complicated for translators. `cleanup-plus` also has separate messages for variables and lists.

addons-l10n/en/folders.json
```diff
@@ -11,5 +11,13 @@
   "folders/name-prompt-title": "Create Folder",
   "folders/name-prompt": "Name of folder:",
   "folders/name-not-allowed": "Invalid folder name",
-  "folders/confirm-backpack-folder": "Save entire folder to backpack?"
+  "folders/confirm-backpack-folder": "Save entire folder to backpack?",
+  "folders/delete-folder-contents": "delete all in folder",
+  "folders/delete-folder-contents-prompt-title": "Delete all in folder",
```
mxmou:
```suggestion
  "folders/delete-folder-contents-prompt-title": "Delete All in Folder",
```

addons/folders/userscript.js
```diff
@@ -765,6 +768,35 @@ export default async function ({ addon, console, msg }) {
     });
   };
 
+  function getRestoreFun(type, lastCostumeDeleted) {
```
mxmou:
```suggestion
  function restore(type, lastCostumeDeleted) {
```

The name `getRestoreFun` implies that this returns a function.

Samq64 (changes requested):
There are still a few references to `getRestoreFun`.

PR #8310, opened 2025-05-03
dark-www: style studio page's loading state

mxmou:
### Changes

Before:

image

After:

image

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/dark-www/addon.json
+++ b/addons/dark-www/addon.json
@@ -1531,6 +1531,17 @@
       }
     },
     {
+      "name": "blue-studioPlaceholder",
+      "value": {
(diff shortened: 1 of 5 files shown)


Samq64 (approved):

DNin01 (commented):
`old-studio-layout` and `scratch3to2` change the background color (from blue) to gray, so if one of those addons are being used the loading background color should be blended with the gray background color.

mxmou:
The loading background is now gray if 3.0 → 2.0 is enabled - I think that fits the 2.0 theme better.

DNin01 (approved):

PR #8311, opened 2025-05-03
Add types to issue templates

DNin01:
### Changes

Each issue template now includes a `type` property so newly opened issues are tagged with the appropriate issue type. This utilizes GitHub's new issue types system.

This does not make any changes to labels.

### DIFF
--- a/.github/ISSUE_TEMPLATE/---bug.yml
+++ b/.github/ISSUE_TEMPLATE/---bug.yml
@@ -1,5 +1,6 @@
 name: 🐛 Bug
 description: Report a bug.
+type: Bug
 labels: ["status: needs triage", "type: bug"]
 body:
   - type: markdown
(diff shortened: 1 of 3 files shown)


Hans5958 (approved):
We should remove the `type:` labels at some point in the future, but this works for now.

DNin01:
It's working.

WorldLanguages:
Nice!

Hans5958:
I checked and I saw that removing the label (and possibly adding a type) shouldn't notify anyone, which means migrating it shouldn't be a nuisance like last time me closing discussions.

DNin01:
> @DNin01, I have removed labels on some bug and enhancement issues that you created. I hope it doesn't notify you.

It doesn't notify.

DNin01:
Maybe someday we could make more types like "problem" for inconvenience/problem reports and "checklist" for to-do lists.

PR #8316, opened 2025-05-05
horizontal-mystuff-tabs: change alignment

mxmou:
### Changes

Aligns the text of the first tab with the My Stuff header and the sort dropdown.

image
image

### Reason for changes

Previously, the border of the first tab was aligned with the header, which looked weird when the first tab wasn't selected:
image

### Tests

Tested on Edge.


DNin01 (approved):
Sure, why not?

Issue #6675, opened 2023-09-21
The forum down error page inconsistently gets affected by `dark-www`


mybearworld:
### Description

When the forums are down, a page is used, which is only sometimes affected by `dark-www`.

### Steps to reproduce

1. Go to the forums when they're down
2. Reload a lot

### Demonstration(s) and additional context

![the forum down image](https://github.com/ScratchAddons/ScratchAddons/assets/130385691/cc3a4dfc-d8a0-4ff3-8662-7210e6101202)
![the forum down image in dark](https://github.com/ScratchAddons/ScratchAddons/assets/130385691/8f5745a9-5fdf-490f-a081-655e31844fb8)

### Environment

Browser: Edge 117
Scratch Addons version: 1.34.1
Operating system: Windows

WorldLanguages:
> Reload a lot

bro :joy:

Relevant lines of code: https://github.com/ScratchAddons/ScratchAddons/blob/9733fdd890854f2023f8a37c947f08c2b6bf5802/background/get-userscripts.js#L269
https://github.com/ScratchAddons/ScratchAddons/blob/d34ee75a4e3684c292d2360739daac13240d502f/content-scripts/cs.js#L52

It's unlikely but this could be caused by Scratch sometimes returning 200 status code with an error page (probably not).
Keep in mind the Scratch Addons status detection is best-effort as there's no reliable way to know the HTTP status code of the current page through JavaScript.

mybearworld:
> > Reload a lot
> 
> bro 😂

What should I have written? That's how you reproduce this

mybearworld:
This'll be a little more difficult to test now that the forums are back up

Chiroyce1:
> This'll be a little more difficult to test now that the forums are back up

I suppose its worth fixing when the forums go down again then lol, this is honestly a non issue since most of the text is stil readable, and the contrast is good enough.

mybearworld:
> I suppose its worth fixing when the forums go down again then lol, this is honestly a non issue since most of the text is stil readable, and the contrast is good enough.

True. It's probably just some race condition thing where some addon expects something from the forums to exist on the page.

WorldLanguages:
[...]

PR #8318, opened 2025-05-06
Don't run most addons on error pages and My Classes

### LINKED ISSUE #6239: Educator gliches
### Description

There are some weird issues with scratchaddons in educator account

View my classes:
With the 3.0 Layout enabled
Before:
image
After:
image

### Steps to reproduce

1. Signin to a educator account
2.Go to my classes
3.Done

_No response_

### Environment

Browser: Brave
Scratch Addons version: 1.33.1
Operating system: Windows 10

### LINKED ISSUE #6853: We probably shouldn't run addons on scratch.mit.edu/emails/*
### Description

Addons shouldn't run on https://scratch.mit.edu/emails/welcome/thisiswhatimean/1/

### Steps to reproduce

1. Enable sticky footer
2. Go to https://scratch.mit.edu/emails/welcome/thisiswhatimean/1/
3. Weird layout issues with the footer

_No response_

### Environment

Any

### LINKED ISSUE #8300: Inconsistent Error Page Style
### Description

There are two different 500 errors that look very similar to each other. One can be accessed with [https://scratch.mit.edu/internalapi/backpack/kaj/get](https://scratch.mit.edu/internalapi/backpack/kaj/get), while the other can be accessed with [https://scratch.mit.edu/500/](https://scratch.mit.edu/500/) (notice how the first error page does not say "Windows, macOS, ChromeOS, or Android", while the second one does). However, the "Scratch 2.0 → 3.0" addon is inconsistent with changing the pages; the first page gets a yellow background to be consistent with other 3.0 pages, while the second one doesn't.

### Environment

Browser: Version 135.0.7049.115
Scratch Addons version: v1.42.0
Operating system: Sequoia Mac 15.4.1

mxmou:
Resolves #6239
Resolves #6853
Resolves #8300

### Changes

The `*` and `https://scratch.mit.edu/*` match patterns now exclude the following:
* all error pages
* everything under `/emails`
* My Classes (to fix #6239)

Addons can still run on these pages if they explicitly include them in `matches`. Specifically, 403 and 404 errors are still affected by dark mode and by addons for customizing the navigation bar (except the setting to disable the sticky navigation bar, which isn't very useful on error pages anyway).

### Tests

Tested on Edge and Firefox.


Samq64:
This doesn't fully resolve #8300 since only the first page is effected by dark mode, but I think it's close enough.

I suggest excluding 5xx but not 4xx and then removing the `#page-500` styles from `scratchr2_errors.css` instead.

Also this resolves #6675

mxmou:
> This doesn't fully resolve #8300 since only the first page is effected by dark mode, but I think it's close enough.

I don't think these two pages need to look exactly the same because they aren't the same error, but I can make both dark if you think that's better.

> I suggest not excluding 4xx pages and just removing the `#page-500` styles from `scratchr2_errors.css` instead.

So 2.0 -> 3.0 would style 4xx errors? I decided not to do that because of the reason described by Jeffalo in #8300.

WorldLanguages:
I believe the decisions taken by mxmou are fine

Samq64:
`scratch3to2` still runs on 404 project pages, probably because the page itself return 200?

Samq64 (approved):

mxmou:
> `scratch3to2` still runs on 404 project pages, probably because the page itself returns 200.

I think that's fine because it looks very different from the real 404 page, even with 3.0 → 2.0.

PR #8337, opened 2025-05-18
Add "related addons" relationships

WorldLanguages:
Continues #6104

Some notes:
1. Relationships are not bidrectional by design, but most times they will probably be
2. This is a work in progress based on [this table](https://github.com/ScratchAddons/ScratchAddons/pull/6104#issuecomment-1563406851)
3. `We should ensure a majority of the related-addon relationships we add aren't just subcategories. For example, I'm not sure if color picker addons should be related to each other. They aren't directly related - they simply affect the same area of the editor.` [source](https://github.com/ScratchAddons/ScratchAddons/pull/6104#issuecomment-1563429915)
4. Feedback is welcome, but let's try not to discuss excessively over links

### DIFF
--- a/addons/custom-zoom/addon.json
+++ b/addons/custom-zoom/addon.json
@@ -2,6 +2,7 @@
   "name": "Customizable code area zoom",
   "description": "Choose custom settings for the minimum, maximum, speed, and start size of the zoom of scripts in the code area, and autohide the controls.",
   "tags": ["editor", "codeEditor", "featured"],
+  "relatedAddons": ["items-per-row"],
   "credits": [
     {
       "name": "ErrorGamer2000",
--- a/addons/horizontal-mystuff-tabs/addon.json
+++ b/addons/horizontal-mystuff-tabs/addon.json
@@ -2,6 +2,7 @@
   "name": "Horizontal My Stuff tabs",
   "description": "Moves the My Stuff navigation buttons (\"All Projects\", \"My Studios\", \"Trash\") to the top of the page, instead of keeping them on the left.",
   "tags": ["community", "theme"],
+  "relatedAddons": ["scratchr2"],
   "credits": [
     {
       "name": "Maximouse",
--- a/addons/better-emojis/addon.json
+++ b/addons/better-emojis/addon.json
@@ -2,6 +2,7 @@
   "name": "Better emojis",
   "description": "Replaces the emojis in comments on the website with improved vector designs.",
   "tags": ["community", "comments"],
+  "relatedAddons": ["emoji-picker"],
   "credits": [
     {
       "name": "getbent",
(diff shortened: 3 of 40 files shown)


Samq64:
I was also working on this. I've just pushed my current changes so you can borrow from them if you want: https://github.com/Samq64/ScratchAddons/commit/4b3dde3efb7e8ce278ed2fcca7eb3de5ad438298

WorldLanguages (commented):
✅

addons/custom-zoom/addon.json
```diff
@@ -2,6 +2,7 @@
   "name": "Customizable code area zoom",
   "description": "Choose custom settings for the minimum, maximum, speed, and start size of the zoom of scripts in the code area, and autohide the controls.",
   "tags": ["editor", "codeEditor", "featured"],
+  "relatedAddons": ["items-per-row"],
```
mxmou: ?

addons/onion-skinning/addon.json
```diff
@@ -104,6 +104,7 @@
   ],
   "versionAdded": "1.5.0",
   "tags": ["editor", "costumeEditor", "recommended"],
+  "relatedAddons": ["paint-snap"],
```
mxmou: I don't think these two are more closely related than other costume editor addons.

addons/horizontal-mystuff-tabs/addon.json
```diff
@@ -2,6 +2,7 @@
   "name": "Horizontal My Stuff tabs",
   "description": "Moves the My Stuff navigation buttons (\"All Projects\", \"My Studios\", \"Trash\") to the top of the page, instead of keeping them on the left.",
   "tags": ["community", "theme"],
+  "relatedAddons": ["scratchr2"],
```
mxmou: This addon used to be a part of 2.0 → 3.0, but it shouldn't have been. The relation implies that making the tabs horizontal makes the My Stuff page more 3.0-like, even though we can't possibly know if that's true because there's no 3.0 version of the My Stuff page.

mxmou (approved):
I'm still not convinced that the relation between horizontal-mystuff-tabs and scratchr2 makes sense, but the rest looks good.

addons/horizontal-mystuff-tabs/addon.json
CST1229: wasn't the whole purpose of the addon to revert back to the old behavior of 2.0 -> 3.0 on the my stuff page? i think an addon basically descended from 2.0 -> 3.0 could be related to it

mxmou: > wasn't the whole purpose of the addon to revert back to the old behavior of 2.0 -> 3.0 on the my stuff page?

It was, but I made it a separate addon instead of adding the feature back to 2.0 → 3.0 specifically because it isn't really related to 2.0 → 3.0. I don't think it would make sense to make all addons that used to be a part of devtools related to each other either.

PR #8341, opened 2025-05-19
editor-cleanup-plus: new Blockly

mxmou:
### Changes

Updates "enhanced clean up tool" to work with modern Blockly.

### Tests

### DIFF
--- a/libraries/common/cs/devtools-utils.js
+++ b/libraries/common/cs/devtools-utils.js
@@ -60,16 +60,33 @@ export const getOrderedTopBlockColumns = (separateOrphans, workspace) => {
   let maxWidths = {};
 
   if (separateOrphans) {
-    let topComments = w.getTopComments();
+    if (w.getTheme) {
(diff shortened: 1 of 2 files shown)


libraries/common/cs/devtools-utils.js
```diff
@@ -60,16 +60,33 @@ export const getOrderedTopBlockColumns = (separateOrphans, workspace) => {
...
+      }
+    } else {
+      let topComments = w.getTopComments();
 
-    for (const comment of topComments) {
-      if (comment.setVisible) {
-        autoPositionComment(comment);
-        let right = comment.getBoundingRectangle().bottomRight.x;
+      for (const comment of topComments) {
+        if (comment.block_) {
```
mxmou, line 82: This is for old Blockly, but I changed it to make it more obvious what the if statement is for (it checks if the comment is a block comment).

Samq64 (approved):

Issue #8353, opened 2025-05-23
edtor-theme3: Comment color setting always shown


Samq64:
### Description

Customizable block colors's comments setting is always shown even when it's not relevant.

### Steps to reproduce

1. Set `editor-theme3`'s color mode to colored on black
2. Notice the comment color setting is still there
3. Open the editor, the comments are black

### Environment

Browser: Firefox
Scratch Addons version: 1.42.0
Operating system: Linux

mxmou:
It's meant to change the border color. If it doesn't, that should be fixed.

Maybe it would make sense to have two settings, "comment color" for when the text color is set to black/white and "comment border color" for colored text.

Samq64:
The color applies to stage comment borders as expected. I only tested with block comments before.

PR #8354, opened 2025-05-23
Add a couple setting descriptions to editor-theme3

Samq64:
Resolves https://github.com/ScratchAddons/ScratchAddons/issues/8353#issuecomment-2904449426

### Changes

Moves the insertion marker notice into a setting description and adds one about comments in the colored text mode. Also fixes the help icon alignment in the popup.

### Reason for changes

Although the notice was always visible it was far from the settings that effected it. Since there are multiple though this might not be any better.

### Tests

Tested on Brave.

### DIFF
--- a/addons/editor-theme3/addon.json
+++ b/addons/editor-theme3/addon.json
@@ -359,7 +359,7 @@
     },
     {
       "name": "Comments",
-      "description": "When using colored text on a black/white background this setting effects the borders of standalone comments.",
+      "description": "When using colored text on a black/white background this setting affects the borders of standalone comments.",
       "id": "comment-color",
       "type": "color",
       "default": "#FEF49C"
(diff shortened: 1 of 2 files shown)


addons/editor-theme3/addon.json
```diff
@@ -364,6 +359,7 @@
     },
     {
       "name": "Comments",
+      "description": "When using colored text on a black/white background this setting effects the borders of standalone comments.",
```
WorldLanguages, line 362:
```suggestion
      "description": "When using colored text on a black/white background this setting affects the borders of standalone comments.",
```

addons/editor-theme3/addon.json
```diff
@@ -364,6 +359,7 @@
     },
     {
       "name": "Comments",
+      "description": "When using colored text on a black/white background this setting affects the borders of standalone comments.",
```
mxmou, line 362: If we add support for showing descriptions conditionally in the future, this should only be shown if the text color is set to colored.

mxmou (approved):

PR #8356, opened 2025-05-24
Fix editor message count padding with custom-menu-bar

Samq64:
### Changes

Fixes this with editor messages enabled and custom-menu-bar set to labels only:
image

### Tests

Tested on Brave

### DIFF
--- a/addons/custom-menu-bar/menu-labels.css
+++ b/addons/custom-menu-bar/menu-labels.css
@@ -6,7 +6,8 @@
   justify-content: center;
 }
 
-[class*="menu-bar_hoverable"] > span:not(.sa-editormessages-count) {
+[class*="settings-menu_dropdown-label_"],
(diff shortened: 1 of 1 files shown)


addons/custom-menu-bar/menu-labels.css
```diff
@@ -6,7 +6,7 @@
   justify-content: center;
 }
 
-[class*="menu-bar_hoverable"] > span {
+[class*="menu-bar_hoverable"] > span:not(.sa-editormessages-count) {
```
mxmou, line 9: I think it's better to more precisely specify what the selector should match instead of excluding things it shouldn't.
```suggestion
[class*="settings-menu_dropdown-label_"],
[class*="menu-bar_collapsible-label_"] {
```

mxmou (approved):

PR #8363, opened 2025-05-29
Remove unused code and addon translations

Samq64:
### Changes

Removes some dead code from `find-bar` and the settings page that hasn't been used since 2021. Also deletes a bunch of translation files for removed addons. The action should be updated in the future to handle that though.

### Reason for changes

It's not being used for anything.

### Tests

The find bar and settings page work fine and if any extra translation files were accidentally deleted they will be regenerated.


WorldLanguages (approved):
Double check each deleted file before merging
(Actually nevermind, the l10n script should recreate any files if they were removed by accident)

PR #8365, opened 2025-05-30
`<is in editor?>` block for debugger

### LINKED ISSUE #8362: `<is in editor?>` block
### Why this would be helpful

Often when I'm coding, I create macros associated with keys to do different action like display the project thumbnail or show variables to debugging.

<br>

The problem is when I want share the project, I must remove manually all the macros.
So we would need an addon that only activates certain scripts in the project editor and not in the projet page.

### How this addon works

The solution would be have an  block.

It would work like the `<is compiled?>` and `<is TurboWarp?>` blocks.
It would return `0` by default and `true` if the addon is enabled and one is in the editor.

<br>

There should also be a button in the editor that allows you to switch the block value between `true` and `0`.

### Possible alternatives

It could be included in the debugger addon.

### Additional context

_No response_

Valmontechno:
Resolves #8362

### Changes

Add the `addReporterBlock()` function to the `addon.Tab` API. It adds a block that can return a value to the toolbox. Internally this is a special-cased argument reporter block like the "is compiled?" block of TurboWarp.

This allows to add the  block to the Debugger category.
You can test this block on this project https://scratch.mit.edu/projects/1182533047

### Reason for changes

View #8362

### Notice

The [`blocks.js`](https://github.com/Valmontechno/ScratchAddons/blob/a742be44d812bbb3075b2bdc5ce200f23bea0790/addon-api/content-script/blocks.js#L117C14-L117C31) module that allow to add blocks use the the `proccode` to identify the blocks, however the argument reporter blocks haven't `proccode` propriety and use instead the field value. I still use the "proccode" name because it used by all functions in the module. This should probably be changed.

### DIFF
--- a/addon-api/content-script/blocks.js
+++ b/addon-api/content-script/blocks.js
@@ -103,6 +103,7 @@ export const addBlock = (proccode, { args, callback, hidden, displayName }) => {
 
   const blockData = {
     id: proccode,
+    type: "procedures_call",
     args,
     handler: callback,
     hide: !!hidden,
@@ -110,6 +111,23 @@ export const addBlock = (proccode, { args, callback, hidden, displayName }) => {
   };
   customBlocks[proccode] = blockData;
   customBlockParamNamesIdsDefaults[proccode] = getNamesIdsDefaults(blockData);
+  queueToolboxUpdate();
+};
+
+export const addReporterBlock = (proccode, isBoolean, { callback, hidden }) => {
+  // The `proccode` parameter is not the proccode but the name of the argument
+  if (getCustomBlock(proccode)) {
+    return;
+  }
+
+  const blockData = {
+    id: proccode,
+    type: isBoolean ? "argument_reporter_boolean" : "argument_reporter_string_number",
+    handler: callback,
+    hide: !!hidden,
+    proccode,
+  };
+  customBlocks[proccode] = blockData;
   queueToolboxUpdate();
 };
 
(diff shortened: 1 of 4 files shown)


The2AndOnly:
I would like this!

addon-api/content-script/blocks.js
```diff
@@ -123,14 +141,19 @@ const generateBlockXML = () => {
...
+      const [names, ids, defaults] = getNamesIdsDefaults(blockData);
+      xml +=
+        '<block type="procedures_call" gap="16"><mutation generateshadows="true" warp="false"' +
+        ` proccode="${escapeHTML(proccode)}"` +
+        ` argumentnames="${escapeHTML(JSON.stringify(names))}"` +
+        ` argumentids="${escapeHTML(JSON.stringify(ids))}"` +
+        ` argumentdefaults="${escapeHTML(JSON.stringify(defaults))}"` +
+        "></mutation></block>";
+    } else {
+      xml += `<block type="${blockData.type}"><field name="VALUE">${blockData.id}</field></block>`;
```
Samq64: Could these values be passed into `escapeHTML`?

addons/debugger/addon.json
```diff
@@ -1,6 +1,13 @@
 {
   "name": "Debugger",
-  "description": "Adds a new \"debugger\" window to the editor. Allows for logging into the \"Logs\" tab of the debugger window using the \"log\", \"warn\" and \"error\" blocks. The \"breakpoint\" block will pause the project when executed. All running stacks of blocks can be viewed in the \"Threads\" tab of the debugger window, and when paused the \"Step\" button can be used to execute the next block. A graph of frames per second and number of clones can be viewed in the \"Performance\" tab.",
+  "description": "Adds a new \"debugger\" window to the editor. Allows for logging into the \"Logs\" tab of the debugger window using the \"log\", \"warn\" and \"error\" blocks. The \"breakpoint\" block will pause the project when executed. All running stacks of blocks can be viewed in the \"Threads\" tab of the debugger window, and when paused the \"Step\" button can be used to execute the next block. A graph of frames per second and number of clones can be viewed in the \"Performance\" tab. The \"is it in editor?\" block allows to check if the project is running in the editor.",
```
Samq64, line 3: All the blocks should probabbly be mentioned in the same sentance.

addons/debugger/addon.json
```diff
@@ -1,6 +1,13 @@
...
+  "info": [
+    {
+      "type": "notice",
+      "text": "Debug blocks only work for Scratch Addons users with this enabled. The \"is in editor?\" block cannot therefore it cannot be used as an anti-cheat method.",
```
Samq64:
```suggestion
      "text": "Debug blocks only work for Scratch Addons users with this enabled. The \"is in editor?\" block cannot therefore be used as an anti-cheat method.",
```

addons/debugger/userscript.js
```diff
@@ -68,6 +68,11 @@ export default async function ({ addon, console, msg }) {
       logMessage(content, thread, "error");
     },
   });
+  addon.tab.addReporterBlock("\u200B\u200Bis in editor?\u200B\u200B", true, {
+    callback: () => {
+      return document.body.classList.contains("sa-body-editor") ? true : 0;
```
Samq64, line 73: Why return different types here?

Samq64 (changes requested):

Samq64:
Sorry, I forgot this is a draft.

yuri-kiss:
this can be easilly abused can it not?
image
its already possible just with the debugger block (with limited functionality)

yuri-kiss:
also that project is private now, we cannot see it

Valmontechno:
> also that project is private now, we cannot see it

The new project https://scratch.mit.edu/projects/1245715078/

yuri-kiss:
> > also that project is private now, we cannot see it
> 
> The new project https://scratch.mit.edu/projects/1245715078/

do you have any thoughts on the abuse it could cause?

PR #8400, opened 2025-06-21
Fix setting layout in popup

### LINKED ISSUE #8399: Setting layout in the popup is wrong
### Description

Version 1.23 (specifically PR #3811), released in 2021, broke the CSS for setting labels in the popup and apparently nobody noticed.

v1.22.1 (note that Boolean settings are next to their labels, making it easy to see which label belongs to which switch):
Image

v1.23.0 (Boolean settings are now far from their labels and the long label wraps even though there's enough space):
Image

v1.42.0:
Image

### Steps to reproduce

1. Open the popup
2. Find an addon with many settings, for example Block switching.

_No response_

### Environment

Browser: Edge 137
Scratch Addons version: v1.42.0
Operating system: Windows

mxmou:
Resolves #8399

### Changes

image

### Reason for changes

* Putting settings closer to their labels makes it easier to see which setting is which.
* Labels that wrap unnecessarily use a lot of extra space.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -278,11 +278,13 @@ body.iframe .settings-column > .addon-setting {
   flex-direction: column;
   align-items: flex-start;
 }
-body.iframe .setting-label {
-  width: auto;
-}
 body.iframe .setting-label-container {
   margin-bottom: 5px;
+  width: auto;
+  max-width: calc(100% - 66px); /* prevent setting description tooltip from overflowing */
+}
+body.iframe .setting-label-container:not(:has(.tooltip)) {
+  max-width: none;
 }
 body.iframe .setting-input {
   flex-shrink: 0;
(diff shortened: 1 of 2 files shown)


webpages/settings/style.css
```diff
@@ -278,11 +278,13 @@ body.iframe .settings-column > .addon-setting {
...
-  width: auto;
-}
 body.iframe .setting-label-container {
   margin-bottom: 5px;
+  width: auto;
+  max-width: calc(100% - 66px); /* prevent setting description tooltip from overflowing */
+}
+body.iframe .setting-label-container:not(:has(.tooltip)) {
+  max-width: none;
 }
```
mxmou, line 288: In browsers that don't support `:has`, the `max-width` will be applied to all labels, which shouldn't cause any issues.

Samq64:
Keeping the toggles close to the labels does help but I'm not entirely convinced because of the misalignment in cases like this:

mxmou:
> Keeping the toggles close to the labels does help but I'm not entirely convinced because of the misalignment in cases like this

I don't think there's a simple way to fix that, but it's still better than the current version where all the switches are placed on the right edge. As I explained in the linked issue, the design proposed in this PR was already used at one point until it was accidentally broken in v1.23.

Samq64 (approved):

DNin01 (approved):

PR #8402, opened 2025-06-22
fix: prevent error when adding label to empty custom block (#8396)

### LINKED ISSUE #8396: Rearrangeable custom block inputs has error when adding a label to an empty custom block
Rearrangeable custom block inputs does not let you add a label to an empty custom block.

Steps to reproduce: 
1. Have the rearrangeable custom block inputs addon enable
2. Create a custom block and delete the "block name" text from the label
3. Click OK
4. Edit the block you just created
5. Try to add a label to the empty block. It will not work and this error will be in the console. 

Image

Joeclinton1:
Resolves #8396

### Changes

* Added optional chaining to safely access `lastInput.type` in `addLabelExternal`

### Reason for changes

* Fixes crash when adding a label to an empty custom block

### Tests

* Manually tested in **Chrome**

  * Verified label can be added to empty block without errors
  * Confirmed other custom block editing features still work as expected

### DIFF
--- a/addons/reorder-custom-inputs/userscript.js
+++ b/addons/reorder-custom-inputs/userscript.js
@@ -70,7 +70,7 @@ export default async function ({ addon, console }) {
       // We account for this with a hacky method of adding the delimiter at the end of the last label input
       if (fnName === "addLabelExternal") {
         const lastInput = proc.inputList[proc.inputList.length - 1];
-        if (lastInput.type === INPUT_DUMMY) {
+        if (lastInput?.type === INPUT_DUMMY) {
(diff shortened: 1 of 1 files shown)


Samq64 (changes requested):
Existing uses of the labeless custom block are not updated when editing it, only the definition and block palette.

Joeclinton1:
> labeless custom block are not updated when editing it, only the definition and block palette.

Separate bug, nothing to do with the first bug. In fact that is a Scratch bug, not a Scratch Addons bug.

PR #8418, opened 2025-07-02
invite-recent-users

VIGARPAST-777-2:
## Add "Invite Recent Users" Addon

**Addon ID:** `invite-recent-users`  
**Author:** @VIGARPAST_777  
**Category:** studio

### Description
This addon adds a button in studio pages to invite recent users to the studio.  
It fetches users from:
- Griffpatch's followers
- ScratchCat's followers
- Griffpatch's recent profile commenters (planned)
  
Features:
- Invite users one by one using Scratch's API
- Load more users
- Change data source from dropdown
- Lightweight UI that appears in the top right corner

Tested in latest Scratch UI (2025) and working.

Let me know if anything needs to be improved!


Samq64:
The files are not laid out out correctly, so this will not work. Also similar addons have been rejected in the past since they would cause a fair amount of spam for both users and the Scratch API.

CST1229:
> The files are not laid out out correctly

i feel like this pr is ai-generated.

please test your pull requests! because you very definitely did *not* test this at all...

PR #8424, opened 2025-07-05
copy-reporter: new Blockly

mxmou:
### Changes

Makes "copy reporter values" work with modern Blockly.

### Tests


Samq64:
When changing Scratch's color mode the report value is duplicated:
image

Samq64 (approved):

Issue #8144, opened 2025-02-07
New Addon: Show messages count on profile pages


50-scratch-tabs:
### Why this would be helpful

When looking at a user's profile page, knowing their messages count is useful for a variety of reasons, such as having an idea of when they last checked them, if they check them, and if they'll likely see your comment,

### How this addon works

This addon, when you load a profile page, will add a note near the top saying how many messages they have.

### Additional context

I wrote a userscript to do this a while ago in TamperMonkey, but I just had the thought of contributing it to Scratch Addons. It would obviously need some rewriting to fit in with Scratch Addons and its API (as well as testing with the profile page addons that I don't use), but that's pretty doable.

Here is the TamperMonkey userscript I made for it: [Scratch message display-2024-10-16.user.txt](https://github.com/user-attachments/files/18708818/Scratch.message.display-2024-10-16.user.txt)

Screenshot of what that userscript looks like: ![Image](https://github.com/user-attachments/assets/8be3b330-07e7-492b-ac97-93685ed7f9d9)

Also, I just discovered ScratchTools has it, but I don't like ScratchTools because it breaks on Firefox mobile.

AmpElectrecuted:
I'll try working on this nw even though I'm not a contributor, but hopefully my PR gets merged.

50-scratch-tabs:
> I'll try working on this now even though I'm not a contributor, but hopefully my PR gets merged.

I suggest starting with the code I uploaded.

mxmou:
This was already discussed in #3802. I don't think we should add it because message counts aren't meant to be seen by other users (if they were, this would be a Scratch feature).

50-scratch-tabs:
> This was already discussed in [#3802](https://github.com/ScratchAddons/ScratchAddons/pull/3802). I don't think we should add it because message counts aren't meant to be seen by other users (if they were, this would be a Scratch feature).

Ok then :(

PR #8428, opened 2025-07-06
Message count on profile page

VIGARPAST-777-2:
### Add new addon: Show Scratch message count on user profile

This PR adds a new Scratch Addons feature that displays the number of messages (notifications) a user has received on their Scratch profile page.

### Features

- Fetches message count from the Scratch API (`/users/<username>/messages/count`)
- Automatically displays the count below the user's name on their profile
- Styled with consistent design using `.sa-user-post-count`
- Supports both English and Spanish localization (if needed later)

### Motivation

This addon provides quick access to a user's current message count without needing to switch to the messages page. It's especially useful for checking activity at a glance while viewing a profile.

### Author

Created by **VIGARPAST_777** 🚀

Let me know if you'd like me to localize the UI text or add settings!


VIGARPAST-777-2:
@Samq64 I tried this with the developer mode, and it works, check it out and approve it if you can.

addons/user-post-count/addon.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "Show post count",
+  "description": "Adds the number of forum posts below the username on profile pages.",
```
Samq64: Could this be rewritten to describe what it actually does?

addons/user-post-count/userscript.js
```diff
@@ -0,0 +1,26 @@
...
+
+  usernameHeader.insertAdjacentElement("afterend", postCountSpan);
+
+  try {
+    const res = await fetch(`https://api.scratch.mit.edu/users/${username}/messages/count`);
+    if (!res.ok) throw new Error("Network response not ok");
+
+    const data = await res.json();
+
+    postCountSpan.innerText = `✉️ Messages: ${data.count ?? 0}`;
```
Samq64, line 21: The description says this is localized but it isn't. Could you follow https://scratchaddons.com/docs/localization/localizing-addons?

addons/addons.json
```diff
@@ -188,5 +188,6 @@
   "remove-curved-stage-border",
   "columns",
   "editor-compact",
-  "editor-theme3"
+  "editor-theme3",
+  "user-post-count"
```
Samq64: This should be moved above the line `// NEW ADDONS ABOVE THIS ↑↑` and renamed to better describe what it does.

Samq64 (commented):
This is also unlikely to be accepted since #3802 was closed.

If it is though I've left a review below.

addons/user-post-count/addon.json
VIGARPAST-777-2: sry, i tried to make another addon, and and I forgot to change some things

addons/addons.json
VIGARPAST-777-2: I did it

addons/user-post-count/addon.json
VIGARPAST-777-2: > Could this be rewritten to describe what it actually does?

done

mxmou:
We ask contributors to open an issue about their suggestions before submitting a PR, so that you don't waste time developing something that won't be accepted. This specific idea has a few problems:
* As discussed in #3802, it reveals information that's normally only visible to the account owner.
* It's not very useful. Your PR description says it's "useful for checking activity at a glance", but the What I've been doing section of the profile already exists and has the same purpose.

Since the same idea has previously been rejected, I'm closing this. If you want it to be reconsidered, please comment on the issue (#8144) and explain why.

PR #8438, opened 2025-07-11
custom-block-text: new Blockly

mxmou:
### Changes

Updates "customizable block text style" to work with modern Blockly.

### Tests

Tested on Edge and Firefox.


Samq64:
On new Blockly the shadow is also applied to text inputs while on the website it isn't.

mxmou:
Text inputs are now fixed. On old Blockly, dropdowns don't have a shadow either, but I don't think that's intentional.

Samq64 (approved):

PR #8445, opened 2025-07-14
Rename "full screen settings" to "settings page"

mxmou:
### Changes

Changes all references to "full screen settings" to say "settings page" instead.

### Reason for changes

The name "full screen settings" doesn't make sense on mobile devices because the popup is also full-screen.

### Tests

Tested on Edge.


Samq64 (approved):
Alternatively it could be called the "full settings page".

DNin01 (approved):
The previous wording was clearer about where you can find the settings page, but there are already some contextual clues that help you find your way from the popup page to the settings page.

DNin01:
Wait for a review from @Hans5958…?

mxmou:
> Wait for a review from @Hans5958…?

I just wanted a second review because this changes the text in a quite prominent part of the UI.

PR #8453, opened 2025-07-16
gamepad: move button after Set Thumbnail

mxmou:
### Changes

Before:
image

After:
image

Nothing is changed in the editor or on the current Scratch website.

### Tests

Tested locally on Edge.

PR #8451, opened 2025-07-15
dark-www & scratch3to2: style new "Set Thumbnail" button & modal

mxmou:
### Changes

image<br />
image<br />
image

### Reason for changes

Scratch will add a "Set Thumbnail" button on the project page.

### Tests

Tested locally on Edge.

The modal normally only opens when the button is clicked for the first time. To open it again, delete the `isFirstManualThumbnailUpdate` key in localStorage and reload the page.

PR #8456, opened 2025-07-17
animated-thumb: move button to dropdown

### LINKED ISSUE #8454: Scratch's new Set Thumbnail button: animated-thumb needs to be updated
Scratch will add a "Set Thumbnail" button to the project page:
image

> When you press the button, a snapshot of your project canvas is taken and saved as the project thumbnail. You can also use it while the project is playing.

If the thumbnail setter addon is also enabled, there will now be two "Set Thumbnail" buttons with overlapping functionality, which is confusing. My suggestion is to add a dropdown menu next to Scratch's button and move the addon's "Select File" option there. "Use Stage" is no longer needed because the new button does the same thing.

Scratch already shows a popup describing the new feature when the button is clicked for the first time. If we release an update before Scratch, we could add a message to that popup, explaining that the addon's UI has moved. If not, we can do something similar to the previous major update to this addon: https://github.com/ScratchAddons/ScratchAddons/pull/3603#issuecomment-961840230

mxmou:
Resolves #8454

### Changes

Changes the UI of the thumbnail setter addon if the upcoming Scratch update is detected:
image

Clicking "Upload Custom Thumbnail" opens the file picker. There's no "Use Stage" option because that's what Scratch's new Set Thumbnail button does. The "stop thumbnail auto-saving" checkbox is also no longer necessary - Scratch will remove automatic updating of thumbnails.

The addon adds the following message to the modal that appears when clicking "Set Thumbnail" for the first time:
image

If the user has already seen that modal, this message is shown instead:
image

### Tests

Tested locally on Edge and Firefox.


DNin01:
Nicely done!

DNin01:
It might actually be a good idea **not to add this dropdown menu the moment Scratch adds the Set Thumbnail button**. Why? Because it would look like the new Set Thumbnail button was either entirely a Scratch feature or entirely a Scratch Addons feature, which could create some confusion around where the feature is coming from.

For the combobutton to work out best for an unknowing user:
1. In the moment, I think we should leave the buttons alone; let Scratch's Set Thumbnail button be and keep our Set Thumbnail button where it is but at least rename ours to "Upload Thumbnail" to make it distinct.
2. After a while we can update the addon to transform our button, once users are used to having two independent features that set thumbnails.

So, is it okay if we plan to apply this update 1-2 weeks after Scratch's thumbnail button update?

Samq64:
Are there any special steps needed to test this? I'm logged in on localhost on the right `scratch-www` branch but I don't see the button.

WorldLanguages:
> So, is it okay if we plan to apply this update 1-2 weeks after Scratch's thumbnail button update?

I agree.

I even think it wouldn't be too bad if our thumbnail setter was broken/inaccessible for a little while. I vote to intentionally break it or alternatively to merge #8457 instead for this week's release.

` `

> scratchClass() can now also extract class names from stylesheets that use CSS nesting.

Very cool!

WorldLanguages:
> I even think it wouldn't be too bad if our thumbnail setter was broken/inaccessible for a little while.

For two reasons:

- The situation should be clearer to the average user. If Scratch and SA change their UIs drastically at the same time, it's not easy to understand which new UI is added by whom
- It's sometimes fine to have little features break and see most users don't care (and if they do it's also nice to see :P)

DNin01:
> intentionally break it

XD yeah

mxmou:
> Are there any special steps needed to test this? I'm logged in on localhost on the right `scratch-www` branch but I don't see the button.

The code that actually adds the button is in scratch-editor, so you need to switch to the right branch in both repos, run `npm run build` in the scratch-editor directory, and use `npm link`.

Samq64:
The order isn't adjusted with `editor-buttons-reverse-order` enabled so the upload button gets cut off:
image

mxmou:
Fixed:
image

addons/animated-thumb/upload.svg
Samq64, line 1: Was this file left here because it will replace the logo after the next release?

Samq64 (commented):
Our tooltip should be moved as well:
image

Also I can't get a thumbnail uploaded from localhost but that's probably to be expected.

addons/animated-thumb/upload.svg
mxmou: Yes, I decided not to remove the icon because the logo is meant to be temporary.

mxmou:
> Also I can't get a thumbnail uploaded from localhost but that's probably to be expected.

Indeed. I didn't change the code that saves the thumbnail, and the ST didn't change anything on the backend either, so everything should continue to work on the actual Scratch website.

addons/animated-thumb/addon.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Thumbnail setter",
```
DNin01: Because Scratch will have a thumbnail-setting feature, should we rename this addon to make it more distinct?
```suggestion
  "name": "Thumbnail uploader",
```

addons-l10n/en/animated-thumb.json
```diff
@@ -12,5 +13,7 @@
   "animated-thumb/error": "Thumbnail could not be uploaded.",
   "animated-thumb/too-big": "The file is too big.",
   "animated-thumb/gif": "Projects with an animated thumbnail might not appear on the Scratch front page and search results. Upload anyway?",
-  "animated-thumb/if-unsuccessful": "If this is not the thumbnail you uploaded just now, the Scratch server might be having a problem. Please wait a few hours and try again."
+  "animated-thumb/if-unsuccessful": "If this is not the thumbnail you uploaded just now, the Scratch server might be having a problem. Please wait a few hours and try again.",
+  "animated-thumb/info-box": "To upload a custom thumbnail using the thumbnail setter addon, click the arrow icon next to the \"Set Thumbnail\" button.",
+  "animated-thumb/info-tooltip": "To upload a custom thumbnail using the thumbnail setter addon from Scratch Addons, click the arrow icon."
```
DNin01: We could also use the message that @CST1229 suggested:

```suggestion
  "animated-thumb/info-tooltip": "The Scratch Team added this \"Set Thumbnail\" button, so Scratch Addons' thumbnail upload feature been moved here."
```

mxmou: Maybe, but I prefer the shorter message. Users will only see this if they've already clicked the Set Thumbnail button at least once - those who haven't will see the other message.

Samq64 (approved):

WorldLanguages:
Let's make a minor release out of this PR (and the one from yesterday that's already merged), what do you think?

mxmou:
> Let's make a minor release out of this PR (and the one from yesterday that's already merged), what do you think?

Sure. (There are two PRs from yesterday, #8451 and #8453.)

DNin01:
> Let's make a minor release out of this PR (and the one from yesterday that's already merged), what do you think? 

That's usually what I say. 😄

On top of the changes we made to support Scratch's Set Thumbnail button, here are some other changes we could cherry-pick for the minor update:
- 96d998d1b807682223e956790ec305254adb5f73
- e33de4763aa73d50afc19b8d3640561fcf585183

PR #8459, opened 2025-07-18
Fix help icon alignment in popup

Samq64:
### Changes

Moves the setting label margin back onto the parent when in narrow mode.


mxmou (changes requested):
Currently, this makes Boolean settings not aligned with their labels. In line 495 of `addon-body.html`, `.setting-label` should be changed to `.setting-label-container`.

mxmou (approved):

PR #8346, opened 2025-05-21
Compact user dropdown adjustments

Samq64:
Resolves #8067

### Changes

- Renamed `custom-menu-bar`'s class to prevent collisions
- Set the username's `line-height` to 0 on the website
- Fixed the username from disappearing when re-opening the project page
- Use a more specific menu item selector even though it's not really necessary

### Tests

Tested on Brave

PR #8460, opened 2025-07-18
discuss-button: Add redux condition for compact user dropdown

Samq64:
Resolves https://github.com/ScratchAddons/ScratchAddons/pull/8346#discussion_r2214358222

### Tests

Tested on Brave

### DIFF
--- a/addons/discuss-button/www.js
+++ b/addons/discuss-button/www.js
@@ -20,6 +20,7 @@ export default async function ({ addon }) {
   while (true) {
     menuItem = await addon.tab.waitForElement(".account-nav .dropdown > :first-child > :first-child", {
       markAsSeen: true,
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Thanks! I wonder if this should be automatic with an opt-out, it doesn't make sense to catch this on human review when it's our own API

mxmou:
> Thanks! I wonder if this should be automatic with an opt-out, it doesn't make sense to catch this on human review when it's our own API

<del>I don't think this should be default because most addons are editor addons and need the exact opposite condition. Making the option mandatory, so that you'd have to explicitly pass `null` for "no condition", would probably be enough.</del>

Actually, your idea is probably better. Very few addons have to run both inside and outside the editor, so these could explicitly set the condition to `null`. For the rest, "outside editor" makes sense as the default because most editor addons already specify the "inside editor" condition.

Issue #2739, opened 2021-06-11
Be able to export and load a certain addon settings/presets.


TheColaber:
This came from
> Exportable and importable color palettes for themes

mxmou:
Possible UI: there would be a "New preset" button at the end of the list of presets. Clicking it would open a modal showing the current settings of the addon. There would also be an input for the preset name. Once created, there would be edit and delete buttons next to the presets. Things I'm not sure about:
* Ideally it would be possible to reorder the custom presets. If so, there should be a larger margin between the last built-in preset and the first custom one.
* Should it be possible to add and edit presets in the popup?

TheColaber:
Great ideas. I think that the user should be told to go to the main settings page if they want to create presets. Sorta like what you get for trying to enable an addon with permissions.

Samq64:
> Ideally it would be possible to reorder the custom presets. If so, there should be a larger margin between the last built-in preset and the first custom one.

Another option could be to treat built-in presets as custom ones and let the user edit or even delete them.

mxmou:
> > Ideally it would be possible to reorder the custom presets. If so, there should be a larger margin between the last built-in preset and the first custom one.
> 
> Another option could be to treat built-in presets as custom ones and let the user edit or even delete them.

That would be possible, but it would make changing presets like #3839 did more difficult.

DNin01:
What if you could remove properties from the settings file to leave them as is? For instance, a settings file that only contains data of settings for the `dark-www` addon would only change anything in that addon - all other addons and settings would stay as they were configured before the import. This would allow users to, with some know-how, share presets of specific addons.

Just need to make the process more user-friendly and not sound like they're overwriting all of their settings even though it really wouldn't. It might also be more user-friendly to simply have some sort of import/export button on individual addons, for every addon that is thoroughly customizable, or some sort of preset manager.

WorldLanguages:
> It might also be more user-friendly to simply have some sort of import/export button on individual addons

Yes!

Samq64:
I've opened a PR to export and import the settings for an addon from a file but it would still be nice to have proper preset integration and use `chrome.storage` in the future.

> > Another option could be to treat built-in presets as custom ones and let the user edit or even delete them.
>
> That would be possible, but it would make changing presets like https://github.com/ScratchAddons/ScratchAddons/pull/3839 did more difficult.

[...]

PR #8472, opened 2025-07-25
Import and export individual addon's settings

Samq64:
See #2739

### Changes

Adds importing and exporting to addons with presets. This doesn't save custom presets to `chrome.storage`, only json files.

### Reason for changes

To be able to share themes without overwriting all the extension's settings.

### Tests

Tested on Firefox and Brave.

Todo:
- [x] Split dropdown styles from `addon-setting`
- [x] Close dropdown when clicking outside


webpages/settings/components/addon-body.js
```diff
@@ -54,6 +55,51 @@ export default async function ({ template }) {
...
+        const inputElem = Object.assign(document.createElement("input"), {
+          hidden: true,
+          type: "file",
+          accept: "application/json",
+        });
+        inputElem.addEventListener(
+          "change",
+          async (e) => {
+            const file = inputElem.files[0];
+            if (!file) {
```
Samq64: I copied this from `index.js`. Is there a case where this condition is met? Clicking cancel on the file picker doesn't trigger the `change` event.

mxmou:
I don't think these buttons will be used very often, so it might be better to put them into a dropdown:
image

webpages/settings/components/addon-body.js
WorldLanguages: Hmm, did you try all browser and OS combinations?
Keep the check to be consistent I guess. Kinda weird we have a translated string just for this potential edge case though.

WorldLanguages:
What should happen if I upload an addon configuration file into another addon? Can we provide a friendly error message for that case?

webpages/settings/components/addon-body.js
Samq64: It happens on both Firefox and Brave and leaks the input element so I've opened #8475

Samq64:
Right now this is duplicating a few functions and using styles directly from `addon-setting`. Should the dropdown be split into its own component, just the styles or left as it is?

mxmou:
> Right now this is duplicating a few functions and using styles directly from `addon-setting`. Should the dropdown be split into its own component, just the styles or left as it is?

The styles for `.addon-split-button-button` and `.addon-split-button-dropdown` are actually in `addon-body`, not `addon-setting`, so reusing them is fine. I'd remove the `.setting-table-dropdown` class though - it was meant for the table setting UI.

webpages/settings/components/addon-body.html
```diff
@@ -253,6 +267,10 @@
   [disabled] {
     cursor: initial !important;
   }
+  .addon-split-button {
+    align-self: flex-start;
```
mxmou: Not really important, but I think the `align-self` should stay specific to the table setting dropdown to make this class more reusable.

webpages/settings/components/addon-body.js
```diff
@@ -124,6 +177,13 @@ export default async function ({ template }) {
         this.$root.openRelatedAddons(this.addon);
         this.$root.blinkAddon(clickedAddon._addonId);
       },
+      toggleDropdown() {
+        this.isDropdownOpen = !this.isDropdownOpen;
+        this.$root.closePickers({ isTrusted: true }, null, {
+          callCloseDropdowns: false,
+        });
+        this.$root.closeResetDropdowns({ isTrusted: true }, this); // close other dropdowns
```
mxmou: This method should probably be renamed to `closeDropdowns` now that there's another kind of dropdown.

mxmou (approved):
Looks good, just a few small suggestions.

mxmou:
> What should happen if I upload an addon configuration file into another addon? Can we provide a friendly error message for that case?

We might also want to consider
1. a setting file for the whole extension being imported using an addon's Import settings option, and
2. an addon's setting file being imported using More Settings.

Currently, the "file is invalid" message is shown in both cases, which is fine, but I think it would be even better if the file was accepted. In the first case, other addons' settings should probably stay unchanged.

webpages/settings/index.js
```diff
@@ -261,9 +261,9 @@ let fuse;
       },
       closePickers(e, leaveOpen, { callCloseDropdowns = true } = {}) {
         this.$emit("close-pickers", leaveOpen);
-        if (callCloseDropdowns) this.closeResetDropdowns();
+        if (callCloseDropdowns) this.closeDropdowns();
       },
-      closeResetDropdowns(e, leaveOpen) {
+      closeDropdowns(e, leaveOpen) {
         this.$emit("close-reset-dropdowns", leaveOpen);
```
mxmou:
```suggestion
        this.$emit("close-dropdowns", leaveOpen);
```

webpages/settings/components/addon-body.html
Samq64: Could the style just be removed? Toggling it in devtools on 1.42 doesn't make a difference even in RTL.

Edit: I removed it

mxmou: Removing it makes the table setting dropdown appear hovered when the mouse is to the right of it:

split_button_align

webpages/settings/components/addon-body.html
```diff
@@ -19,11 +19,25 @@
...
+          >
+            <img src="../../images/icons/undo.svg" class="icon-type" draggable="false" />
+          </button>
+          <div v-click-outside="closeDropdowns">
+            <button class="addon-buttons addon-split-button-dropdown" @click="toggleDropdown">
+              <img src="../../images/icons/expand.svg" class="icon-type" draggable="false" />
+            </button>
+            <ul class="dropdown-list">
+              <li @click="importPreset">{{ msg('import') }}</li>
+              <li @click="exportPreset">{{ msg('export') }}</li>
```
mxmou: I've just noticed that the order here is different from the More Settings modal:

Image<br />

Image

mxmou (approved):

Samq64:
If an addon only has a single boolean or select setting it doesn't make sense to import or export its settings. Should the dropdown be removed from those addons or is it better to keep it for layout consistency?

And what about the popup? Originally I excluded the options there but changed that when switching to the dropdown design.

mxmou:
> If an addon only has a single boolean or select setting it doesn't make sense to import or export its settings. Should the dropdown be removed from those addons or is it better to keep it for layout consistency?

I think it's better to keep it consistent. Same with the popup - the dropdown button doesn't take up a lot of space, so there's no need to hide it.

Issue #1706, opened 2021-03-01
New addon: custom block category


ghost:
**Describe the new addon**
https://scratchblocks.github.io/#?style=scratch3&script=%2F%2F%20From%20this...%0A%0Adefine%20sa-motion%20%7C%20move%20(steps)%20steps%20backwards%0Amove%20((0)%20-%20(steps))%20steps%0A%0A%2F%2F%20to%20this!%0A%0Adefine%20(move%20(steps)%20steps%20backwards%20%3A%3A%20motion%20stack)%0Amove%20((0)%20-%20(steps))%20steps%0A%0A%2F%2F%20Well%2C%20%3Cem%3Esomething%3C%2Fem%3E%20like%20that...&

**Link to the new addon**
Idk

areeq0325:
I had meant "sa-motion | ", but that's fine too.

jeffalo:
i like this idea. we might want to phase out `sa-` because of the browser extenstion/userscript policy

WorldLanguages:
nah, `sa` can mean so many things

areeq0325:
Maybe abbreviations will work as well:
`mo-` would behave like `motion-`
`lo-` would behave like `looks-`
`et-` would behave like `etcetera-`
Since each category gets a unique abbreviation*, those can be included as well.

Endermaster3690:
> nah, `sa` can mean so many things

well maybe we still should? idk i would not use the `sa-` > because of the browser extenstion/userscript policy

TheColaber:
Feedback from before:
> a way to make custom blocks any color in a similar manner to how folders work

Mbrick2:
Maybe comments on define block

FlamedDogo99:
I'm not 100% sold on the implementation, but it could be interesting. I feel like keeping the SA-motion text in the block would be good if we go that way, in order to avoid having duplicates of native blocks. I've tentatively thrown this together:

https://github.com/user-attachments/assets/28936955-4dcf-4da9-8343-859178eb7711

I feel like a comment approach like [#4508](https://github.com/ScratchAddons/ScratchAddons/issues/4508) could be better?

PR #8476, opened 2025-07-27
New Addon: Recolor Custom Blocks

FlamedDogo99:
Partially implements #1706
### Changes
 - Created `recolor-custom-blocks` addons
   - Adds `userscript.js`, `addon.json` and `userstyle.css`
 - Updated `addons/addons.json` to include the new addon
 - Updated `addons/editor-theme3.js` to use custom block colors for procedure related blocks, and the recoloring of procedure_declaration's box_ element
#### Background
When dealing with a higher quantity of custom blocks in a project, it can quickly become difficult to sort through them all to find a particular block ([example](https://scratch.mit.edu/projects/1196842994/editor/)). While this can be partially mitigated by using colorful text such as emojis in block names, it seems like having a visually-uniform method of sorting blocks would be useful.

To address this issue, this addon allows users to color their custom blocks by prepending one of scratch's color categories (for example, `motion:` or `events:`) to it, or by selecting from a series of colored buttons in the edit procedure modal, pictured below.
| Category Color | Custom Color |
|--------|--------|
| Screenshot 2025-08-08 at 10 04 17 AM | Screenshot 2025-08-08 at 10 05 24 AM |

The user also can choose to hide or display prefixes for themselves in the editor

| Prefixes shown | Prefixes hidden |
|--------|--------|
| Screenshot 2025-08-08 at 10 08 39 AM | Screenshot 2025-08-08 at 10 08 56 AM | 

#### Rationale
1) Non-addon users accessibility
    - Coloring by the text at the beginning of the block allows users who don't use / disable the addon to still be able to read and understand the intended organization of a project which utilizes this feature. Additionally with prefixes shown, addon users can also still see how non-addon user's will see the layout project,
2) Flyout grouping.
    - As the custom blocks in the flyout are alphabetically sorted, blocks of the same color will already be next to each other
3) Block searching
   - This allows users to search / insert blocks by color, when prefixes are shown
4) Language accessibility
   - Non-english speakers can use the color menu without needing to understand the prefixes it appends.
#### Implementation
For this addon, my goal is to minimize the amount of runtime overhead and memory overhead. In general terms, this meant avoiding the use of `(async () => {while (true) {})` to listen for changes, and doing as little extra block bookkeeping as possible. In this implantation, the following is employed
 1) **Procedures Call**: `createAllInputs` is shimmed to handle hiding prefixes and recoloring before the normal method is called 
 2) **Procedures Prototype**: `createAllInputs` is shimmed to handle hiding prefixes and recoloring before the normal method is called 
 3) **Procedures Declaration**: As addon `reorder-custom-inputs` completely overrides the `init` and `createAllInputs` methods, we modify `Blockly.BlockSvg.prototype.initSvg` to modify procedures declarations's `onChangeFn` to recolor the block and update the modal's menu items. 
4) All color changes are saved to the block's `recolorCustomBlock`, which we modify `Blockly.BlockSvg.prototype.updateColour` to detect and color the blocks by, before calling the original function

Not conflicting with `theme3` continues to be the most challenging part of this addon. If this addon gets greenlit, a theme + block color api will be created so that addons will stop _directly editing the dom_ (which is really _really_ hard to detect from another addon)
### Tests
This was created and tested in on Brave (Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36) on macOS Sequoia Version 15.5, currently with no known errors


FlamedDogo99:
I would have to use mutation observers to change the coloring of ~~`middle-click-popup`~~ and `find-bar`, as their color values are hardcoded, so I think I'm going to call it for now.
The only thing left is possibly using `Blockly.BlockSvg.prototype.initSvg` for the coloring and only use `Blockly.BlockSvg.prototype.render` for the custom block editing modal, which would allow me to not need to listen to redux states. If anyone with experience can weigh in, it would be appreciated.

FlamedDogo99:
Shoot I just saw #7768, which will kill this completely. `Blockly.BlockSvg.prototype.render` doesn't even dispatch most of the time on the spork branch

mxmou:
> Shoot I just saw #7768, which will kill this completely. `Blockly.BlockSvg.prototype.render` doesn't even dispatch most of the time on the spork branch

There's almost certainly a way to make this work with the new Blockly version. You'll probably need to override a different method instead of `render` - maybe `applyColour`.

FlamedDogo99:
> You'll probably need to override a different method instead of `render` - maybe `applyColour`.

~~It’s looking like applyColour only is used on initialization?~~ I’ll definitely have to do some more research, as directly setting `style` seems to be unreliable (it somehow colors neighboring blocks in the same stack??)

Edit: Accidentally closed this, sorry 

Edit: This actually is very possible, and will be done soon

FlamedDogo99:
Todo
 - [x] Make procedures_declaration recolor inputs/fields when they're added
    - Fixed by dispatching updateColour to children  
 - [x] Fix define block arguments staying the previous color when the block color is updated
    - "Fixed" by manually setting the html element's stroke path _because this behavior is hard-coded_
 - [x] Sometimes procedures_declaration's onChangeFn isn't polluted?
    - The method previously used conflicted with the reorder-custom-inputs addon, and so we now pollute onChangeFn when we catch a procedure_declaration being rendered from Blockly.BlockSvg.prototype.initSvg

FlamedDogo99:
Am I allowed to submit changes to other addons with this? Tweaking the `middle-click-popup` and `find-bar` addons to read color values directly from the current block's coloring for custom procedures would be very nice is all.

Edit: I managed to get middle-click-popup working by just adding a stylesheet, so disregard this

FlamedDogo99:
Todo
  - [x] Make updateAllBlocks recalculate the color of colored custom procedures
  - [x] Somehow respond to theme3's manual coloring of the dom
      - [x] Procedure Declaration `box_`
      - [x] All text coloring
      - [x] Procedure Prototype svgPath stroke color 
      - If there's a way to get other addon's settings, this will be trivial to implement

WorldLanguages:
> Am I allowed to submit changes to other addons with this?

Sure

addons/recolor-custom-blocks/userscript.js
```diff
@@ -0,0 +1,320 @@
...
+    return {r: decimal >> 16, g: decimal >> 8 & 255, b: decimal & 255};
+  }
+  const RGBtoHex = (rgb) => {
+    return "#" + ((rgb.b | rgb.g << 8 | rgb.r << 16) | 1 << 24).toString(16).slice(1);
+  }
+  const multiply = (hexString, {r, g, b}) => {
+    const c = hexToRGB(hexString);
+    const multiplied = {r: r * c.r, g: g * c.g, b: b * c.b}
+    return RGBtoHex(multiplied);
+  }
```
Samq64: Similar functions already exist in [text-color.js](https://github.com/ScratchAddons/ScratchAddons/blob/master/libraries/common/cs/text-color.js), maybe those be used instead?

addons/recolor-custom-blocks/userscript.js
```diff
@@ -0,0 +1,320 @@
...
+    return customColor.colourPrimary === colors.colourPrimary &&
+        customColor.colourSecondary === colors.colourSecondary &&
+        customColor.colourTertiary === colors.colourTertiary &&
+        customColor.colourQuaternary === colors.colourQuaternary;
+  }
+  const setProcedureButtonColor = (iconElement, colors) => {
+    let svg = atob(iconElement.src.replace(uriHeader, ""));
+
+    // To change the menu item colors, we are inserting our own css after scratch's in the style tag
+    // We want to avoid adding unneeded styling, so we keep our changes after the "ccbc edit" comment
```
Samq64:
```suggestion
    // We want to avoid adding unneeded styling, so we keep our changes after the "rcb edit" comment
```

addons/recolor-custom-blocks/userscript.js
```diff
@@ -0,0 +1,320 @@
...
+    // This allows us to remove old changes, and not be negatively effected by theme3 recoloring them
+    const editedComment = "/*rcb edit*/";
+    const endStyleIndex = svg.indexOf("</style>");
+    const editedCommentIndex = svg.indexOf(editedComment);
+    const appendedStyles = editedComment + ".cls-3{fill:" + colors.colourPrimary + ";}.cls-3,.cls-4{stroke:" + colors.colourTertiary + ";}.cls-4{fill:" + colors.colourSecondary + ";}text.cls-4{fill:#fff;stroke:unset;}"
```
Samq64: This isn't necessary but could this be a backtick string with whitespace to improve readability?

addons/recolor-custom-blocks/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Recolor Custom Blocks",
+  "description": "Change the color of custom blocks by adding a block category (such as motion, events, etc) followed by a colon (:) ",
```
Samq64:
```suggestion
  "description": "Changes the colors of custom blocks that start with the name of a category followed by a colon (:), for example \"looks: animate\".",
```

Samq64:
And for localization just adding an addon notice mentioning that the category names are English only should be fine.

addons/recolor-custom-blocks/userscript.js
Samq64: Actually the correct import is [text-color.esm.js](https://github.com/ScratchAddons/ScratchAddons/tree/master/libraries/common/cs/text-color.esm.js).

addons/recolor-custom-blocks/addon.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "Recolor Custom Blocks",
```
Samq64: Addon names should be sentence case, see #2764 for the discussion.
```suggestion
  "name": "Recolor custom blocks",
```

addons/recolor-custom-blocks/addon.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "Change the color of custom blocks by adding a block category (such as motion, events, etc) followed by a colon (:) ",
+  "credits": [
+    {
+      "name": "FlamedDogo99",
+      "link": "https://github.com/FlamedDogo99"
+    }
+  ],
+  "tags": ["editor", "codeEditor"],
```
Samq64:
```suggestion
  "tags": ["editor", "codeEditor", "featured"],
  "relatedAddons": ["data-category-tweaks-v2", "editor-theme3"],
```

FlamedDogo99:
> Could the block text color and border in the definition be adjusted to match its uses when using the colored on white/black options?

This is proving to be more difficult than I expected. Here's the associated parts of editor-theme3
```js
  const textColor = (field) => {
    if (addon.self.disabled) return originalColors.text;
    if (textMode() === "white") return "#ffffff";
    if (textMode() === "black") return "#000000";
    if (field) {
      let block = field.sourceBlock_;
 // This is what's making the colors in the define block not work correctly in colored text mode :(
      if (block.isShadow() && block.getParent()) block = block.getParent();
      return block.getColourTertiary();
    }
    return "#000000";
  };
// ...
  const oldFieldLabelInit = Blockly.FieldLabel.prototype[fieldMethodName];
  Blockly.FieldLabel.prototype[fieldMethodName] = function () {
    oldFieldLabelInit.call(this);
// Directly setting the text color, without giving a way for other addons to notice
    if (this.textElement_) this.textElement_.style.fill = textColor(this);
  };
// ...
    const oldFieldVariableInit = Blockly.FieldVariable.prototype.init;
    Blockly.FieldVariable.prototype.init = function () {
      oldFieldVariableInit.call(this);
// They're even using important here, just to make our lives more difficult
// JUST USE MORE SPECIFICITY
      this.textElement_.style.setProperty("fill", textColor(this), "important");
    };
```
The current problem is that theme3 directly edits the dom, depending on the addon's internal state, which I don't have access to.

Samq64:
Feel free to make some adjustments to `editor-theme3` if needed.

> This is proving to be more difficult than I expected.

Don't worry then, the background still has a tint so it's not that important.

mxmou:
> The current problem is that theme3 directly edits the dom, depending on the addon's internal state, which I don't have access to.

It's probably easier to fix this on the theme3 side than in your addon. theme3 can access the `recolorCustomBlock` attribute of blocks, so you can change the definition of `textColor()` to behave differently if that attribute is present.

mxmou:
> And for localization just adding an addon notice mentioning that the category names are English only should be fine.

I disagree - people who use Scratch in another language might not know what the English names are because there are often multiple possible translations. Even for English-speaking users, it would be helpful to be able to use the name of the category, not the internal color ID, because there are some weird differences between names and IDs (sound/sounds, events/event).

We can add localization support in a later PR, but we probably shouldn't release the addon to users without it. The best solution is to do something similar to [how the `addon.tab.addBlock()` API handles translations](https://github.com/ScratchAddons/ScratchAddons/blob/2b84fc7/addon-api/content-script/blocks.js#L251-L262):
* When creating a block, replace the translated prefix (if there's one) with the color ID.
* When the block is displayed, replace the ID with the translation.

FlamedDogo99:
> It's probably easier to fix this on the theme3 side than in your addon. theme3 can access the recolorCustomBlock attribute of blocks, so you can change the definition of textColor() to behave differently if that attribute is present.

My goal has been to implement this addon in a way that doesn’t require adding specialized logic to other addons, however I will go this route if I continue to be stuck. The less resource intensive, the better.

> * When creating a block, replace the translated prefix (if there's one) with the color ID.
> * When the block is displayed, replace the ID with the translation.

Thanks for taking the time to provide some feedback! While I am all for making this addon accessible to other languages, I would like to take some time to think through the details of the implementation.

The main goal I don’t want to lose track of is accessibility for people who don’t have this addon enabled / don’t use scratch addons at all. To me, that means keeping visual changes minimal, if possible. Essentially, if the addon is disabled, can users still see the blocks in the same order in the flyout, and be able to understand the intended organization of the custom blocks? These changes are more meant to supplement an already existing strategy of custom-block users, instead of replacing it completely.

Another goal I want to keep in mind is keeping prefixes standardized between all addon users (as opposed to allow users to define their own prefixes and colors). The main rationale for this is avoiding conflicting prefix configurations for people viewing other user’s projects, and I’m happy to elaborate on this more if needed.

All that being said, I still think localization is still achievable without conflicting with the aforementioned design goals. The easiest solution that currently comes to mind is just allowing multiple prefixes (across languages) to represent the same coloring. While this will be more computationally involved (we might want to pre-compute these based off of the current language of something similar), it seems like a feasible solution to both of our goals.

Regardless, I would love to hear y’all’s thoughts on this, and I’m happy to rethink things if y’all disagree.

FlamedDogo99:
Out of curiosity, do the scratch forums support multiple languages for the blocks which use custom colors?

For example:

```sb
custom colors :: motion
```

CST1229:
> Out of curiosity, do the scratch forums support multiple languages for the blocks which use custom colors?

i don't think so (as in, you can only type `:: motion`, etc. in english)

mxmou:
Maybe there could be a dropdown menu in the Make a Block popup to select a color? Then localization wouldn't really be necessary - if someone doesn't know the English name of a category, they can still find it in the dropdown.

FlamedDogo99:
Alright, new path forward:
1) Formatting errors: Use `ScratchAddons.globalState` to update styling changes from `editor-theme3`
  - [x] Fix text color for colored text
  - [x] Fix box_ for procedure declaration
  - [x] Fix procedure prototype outline color for colored text
  - [x] Fix procedure prototype text color for colored text
  - This should hopefully fix these not updating during theme changes
2) Localization: Fully commit to using one method, and then worry about tweaking it later
  - [x] Implement @mxmou ‘s dropdown suggestion with colors displayed next to each item of the dropdown
  - [x] ~~Finish implementing prefixes for every language, not as the intended solution to localization but to allow people to see if they like it~~
3) Polishing + beyond: Once the addon is fully functional, experiment with other suggestions 
  - [x] Hiding the prefix when the addon is enabled
    - I think this will require registering new blocks (similar to how the debugger addon works) but there might be some other (easier) trickery that could allow this to work (without breaking the vm when a block is called)
  - [ ] Saving color data directly to the project json
     -  This goes against my goal of no-addon compatibility/parity, but is worth experimenting with
  - [ ] Extension icons on custom blocks
     - No idea, could be cool
  - [ ] Create a block color api so addons stop directly editing the dom of each block, and rewrite each addon which directly edits the dom
     - It is currently a pain to detect changes that other addons have made to blocks. Having a standardized method of interacting with them will allow for less conflicts. (This would make the zebra-striping addon be a lot easier to rework for new Blockly)

addons/recolor-custom-blocks/userscript.js
Samq64: I meant using [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) syntax but those newlines already help a lot and using concatenation leaves the extra whitespace out of the string.

addons/recolor-custom-blocks/userscript.js
```diff
@@ -0,0 +1,485 @@
...
+            [Blockly.Msg.CATEGORY_LOOKS, "looks"],
+            [Blockly.Msg.CATEGORY_SOUND, "sounds"],
+            [Blockly.Msg.CATEGORY_EVENTS, "event"],
+            [Blockly.Msg.CATEGORY_CONTROL, "control"],
+            [Blockly.Msg.CATEGORY_SENSING, "sensing"],
+            [Blockly.Msg.CATEGORY_OPERATORS, "operators"],
+            [Blockly.Msg.CATEGORY_VARIABLES, "data"],
+            [scratchAddons.l10n.messages["data-category-tweaks-v2/list-category"] ?? Blockly.Msg.LIST_MODAL_TITLE, "data_lists"],
+            [addon.tab.redux.state.locales.messages["gui.extension.pen.name"], "pen"],
+            ["-", "reset"],
```
Samq64: I understand not wanting to use the word custom here, so maybe "Default"? Reset could also work.

FlamedDogo99: Those strings (like reset) are what is used internally. What's displayed is the `Blockly.Msg.CATEGORY_VARIABLES` etc. (I've been trying to use scratch's translated strings for everything, so I don't have to worry about localization) How would you feel about "original" (which is `addon.tab.redux.state.locales.messages["gui.theme.default"]`) or "Custom Blocks" (which is `Blockly.Msg.CATEGORY_MYBLOCKS`)

mxmou: I like "original", but I don't think you should reuse Scratch's string for this because the translation might not be identical in every language. It's very easy to create a new string - see https://scratchaddons.com/docs/localization/localizing-addons/.

I also wouldn't recommend reading another addon's string like you do with Lists. Either duplicate it in this addon or move it to [`_general.json`](https://github.com/ScratchAddons/ScratchAddons/blob/master/addons-l10n/en/_general.json).

FlamedDogo99: I've been encouraged to move to a purely visual color system, so disregard the localization issues. Thanks for taking the time to leave feedback. I'll have the (hopefully final) large change out soon.

FlamedDogo99: It probably should have been left as a draft until now, but this should be the last _big_ revision. The UI should more or less be like this from now on
IMG_2204

addons/recolor-custom-blocks/userscript.js
```diff
@@ -0,0 +1,573 @@
...
+  }
+
+  const createColorButton = (colors) => {
+
+    const colorButton = document.createElement("button");
+    colorButton.classList.add("sa-rcb-colorButton")
+    colorButton.style.backgroundColor = colors.colourPrimary;
+    colorButton.style.borderColor = colors.colourTertiary;
+
+    const checkmarkSvg = document.createElementNS('http://www.w3.org/2000/svg', 'svg')
```
Samq64: Could this be split off into an svg file?

Samq64:
The text colour should be adjusted when on a light background to avoid bad contrast.
image

I don't think contrast with the workspace background matters though.

mxmou:
> The text colour should be adjusted when on a light background to avoid bad contrast.

I think the user should simply be expected to choose reasonable colors, at least in the initial version of this addon, because trying to ensure good contrast in all cases would be difficult. Forum scratchblocks don't change the text color depending on the background either.

FlamedDogo99:
Also I would want to have #8489 implemented first. It would take fighting tooth-and-nail with theme3 for the text color

addons/recolor-custom-blocks/userscript.js
FlamedDogo99: It's path is styled by scratch's `.blocklyFlyoutCheckboxPath` selector (in case an addon modifies the default checkbox). Can you still style non-inline svg elements with external css?

addons/editor-theme3/theme3.js
```diff
@@ -141,9 +141,9 @@ const defaultColors = {
 
 // From https://github.com/scratchfoundation/scratch-blocks/blob/2e3a31e/media/dropdown-arrow.svg
 const arrowPath =
-  "M6.36,7.79a1.43,1.43,0,0,1-1-.42L1.42,3.45a1.44,1.44,0,0,1,0-2c0.56-.56,9.31-0.56,9.87,0a1.44,1.44,0,0,1,0,2L7.37,7.37A1.43,1.43,0,0,1,6.36,7.79Z";
```
FlamedDogo99: My darn IDE keeps adding spaces. I'll try to fix this next commit

addons/recolor-custom-blocks/userscript.js
mxmou: > Can you still style non-inline svg elements with external css?

No, but I don't think it makes sense for other addons to affect these check icons anyway - they're quite different from those in the flyout.

addons/editor-theme3/theme3.js
mxmou: If you [enable Actions on your fork](https://github.com/FlamedDogo99/ScratchAddons/actions), it will automatically format the code after each commit.

FlamedDogo99:
This is good to review, however it and theme3 might need to be reworked later to use css instead of setColour + similar methods.
See: https://docs.google.com/document/d/1JLSuu0yqK9Yu9tx0zugBzlcA7qFcpX9bfHh15LDt29c/edit?tab=t.0#heading=h.mcmu9xr3zhjm

addons/editor-theme3/theme3.js
```diff
@@ -597,12 +604,18 @@ export default async function ({ addon, console, msg }) {
...
+      const updateBox_ = () => {
+        if (this.sourceBlock_.isShadow()) return;
+        // Labels in custom block editor
+        this.box_.setAttribute(
+          "fill",
+          isColoredTextMode() ? fieldBackground(this) : this.sourceBlock_.getColourTertiary()
+        );
+      };
+      if (this.box_) {
+        if (!this.box_) this.box_.updateBox_ = updateBox_;
```
Samq64: How could this condition be met? Is it supposed to check if `updateBox_` exists?

addons/recolor-custom-blocks/addon.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "Recolor custom blocks",
+  "description": "Changes the colors of custom blocks that start with the name of a category followed by a colon (:), for example \"looks: animate\".",
```
Samq64:
```suggestion
  "description": "Changes the colors of custom blocks that start with a hex code or category name followed by a colon (:), for example \"looks: animate\".",
```
Alternatively the prefix explanation could be removed since it's all accesible though the UI now.

addons/recolor-custom-blocks/assets/checkmark.svg
Samq64, line 1: There's only a single path so the styles can just be inlined.

addons/recolor-custom-blocks/userscript.js
```diff
@@ -0,0 +1,588 @@
...
+      ".cls-3{ fill: " +
+      colors.colourPrimary +
+      ";}" +
+      ".cls-3, .cls-4{ stroke: " +
+      colors.colourTertiary +
+      ";}" +
+      ".cls-4{ fill: " +
+      colors.colourSecondary +
+      ";}" +
+      "text.cls-4{ fill:#fff; stroke:unset;}";
```
Samq64: The formatter seems to have messed this up, so maybe it is better to use the other syntax?

addons/recolor-custom-blocks/userstyle.css
```diff
@@ -0,0 +1,56 @@
...
+  height: 13px;
+  padding: 0;
+  margin: 3px;
+  /*border-color: rgb(118, 118, 118);*/
+  /*border-color: light-dark(rgb(118, 118, 118), rgb(133, 133, 133));*/
+  border-radius: 3px;
+  border-style: solid;
+  border-width: 1px;
+  background: fieldtext;
+}
```
Samq64:
```suggestion
```

Samq64 (commented):
Could custom colour selections be compatible with `theme3`'s coloured on black/white setting?

Everything else looks good and are mostly just pitpicks.

addons/recolor-custom-blocks/addon.json
mxmou: I suggest something similar to the description of `folders`:
*Adds an option to choose a custom color when creating a block. This feature works by adding the color code or category name at the beginning of the block's label.*

addons/recolor-custom-blocks/assets/checkmark.svg
mxmou: We also usually optimize SVG files using svgo.

addons/editor-theme3/theme3.js
```diff
@@ -597,12 +604,18 @@ export default async function ({ addon, console, msg }) {
     Blockly.FieldTextInput.prototype.init = function () {
       // Text inputs
       oldFieldTextInputInit.call(this);
-      if (this.sourceBlock_.isShadow()) return;
-      // Labels in custom block editor
-      this.box_.setAttribute(
-        "fill",
-        isColoredTextMode() ? fieldBackground(this) : this.sourceBlock_.getColourTertiary()
-      );
+      const updateBox_ = () => {
```
mxmou, line 615: A comment explaining that this function is needed for compatibility with `recolor-custom-blocks` would be helpful here.

addons/editor-theme3/theme3.js
```diff
@@ -906,6 +919,12 @@ export default async function ({ addon, console, msg }) {
       if (textMode() === "colorOnWhite") Blockly.Colours.fieldShadow = "rgba(0, 0, 0, 0.15)";
       else Blockly.Colours.fieldShadow = originalColors.fieldShadow;
       Blockly.Colours.text = uncoloredTextColor(); // used by editor-colored-context-menus
+      if (addon.tab.redux.state.scratchGui.customProcedures.active) {
+        const declarationBlock = Blockly.getMainWorkspace()?.getTopBlocks?.()?.[0];
+        if (declarationBlock?.type === "procedures_declaration") {
+          declarationBlock.updateDisplay_();
```
mxmou, line 936: This should probably have a comment too.

addons/recolor-custom-blocks/addon.json
```diff
@@ -0,0 +1,35 @@
...
+  "credits": [
+    {
+      "name": "FlamedDogo99",
+      "link": "https://github.com/FlamedDogo99"
+    }
+  ],
+  "tags": ["editor", "codeEditor", "featured"],
+  "relatedAddons": ["data-category-tweaks-v2", "editor-theme3"],
```
mxmou, line 11: Could this addon be added to the `relatedAddons` list of the other two as well? Especially data category tweaks - I'm not sure about theme3 because it already has a lot of related addons.

addons/recolor-custom-blocks/userstyle.css
```diff
@@ -0,0 +1,56 @@
...
+.sa-rcb-colorButton {
+  flex-grow: 1;
+  aspect-ratio: 1;
+  width: unset;
+  height: unset;
+  overflow: hidden;
+  border: 2px solid;
+  transition: all 0.2s;
+  border-radius: 100%;
+  margin: 0 auto 0.125rem;
```
mxmou, line 26: Currently, the checkmark is very small compared to the button and not vertically centered. Reducing the padding would fix that:
```suggestion
  margin: 0 auto 0.125rem;
  padding: 4px;
```

addons/editor-theme3/theme3.js
FlamedDogo99: That is supposed to check if updateBox_ exists, which could also probably use a better name.

addons/recolor-custom-blocks/assets/checkmark.svg
FlamedDogo99: > We also usually optimize SVG files using svgo.

Is that part of the action that runs when during commits, or is that something I need to set up?

Samq64: It's not automatic. If you use VSCode the easiest way to install [the extension](https://marketplace.visualstudio.com/items?itemName=1000ch.svgo).

addons/recolor-custom-blocks/userstyle.css
FlamedDogo99: Were you intending to suggest I inline this?

FlamedDogo99: Cancel that, not sure why that's still there

addons/editor-theme3/theme3.js
FlamedDogo99: Sharing values between this and theme3 has been delegated to a module, with clearer naming

addons/recolor-custom-blocks/userscript.js
FlamedDogo99: Changed to use backticks, even though it adds unnecessary white-space

FlamedDogo99:
Implemented the suggestions, and did some cleanup. 
- Naming should be more consistent
- Methods should have clearer purposes
- Comments have been rewritten or added for clarity
- theme3 now has a module that it shares it's textColor method with, along with the addon's disabled state.
- Theme3 should no longer have conflicts with this addon
- The color menu is now hidden if the addon is disabled while editing a block 
- Prefixes now match the english category names, such as variables and lists
- The edit custom block add input buttons now use rgb colors so theme3 doesn't replace them
- The checkmark svg has been optimized with svgo
- Color buttons padding has been fixed

addons/editor-theme3/addon.json
```diff
@@ -184,22 +190,6 @@
           "settingId": "comment-color"
         }
       }
-    },
-    {
-      "name": "text-shadow",
```
Samq64: Was this intentionally removed?

FlamedDogo99: No, sorry. IDE issues. If you suggest an edit I’ll commit it, elsewise I’ll fix that tomorrow.

addons/editor-theme3/addon.json
```diff
@@ -184,22 +190,6 @@
...
-          "type": "settingValue",
-          "settingId": "text"
-        },
-        "options": {
-          "black": "rgba(0, 0, 0, 0.15)",
-          "colorOnWhite": "rgba(0, 0, 0, 0.15)",
-          "colorOnBlack": "#000000"
-        },
-        "default": "rgba(0, 0, 0, 0.4)"
-      }
```
Samq64, line 202:
```suggestion
```

addons/recolor-custom-blocks/addon.json
Samq64: I think you accidentally clicked the resolve button on this one since you upvoted mxmou's comment.

FlamedDogo99: Oh, I did end up going with his “Adds an option to choose a custom color when creating a block. This feature works by adding the color code or category name at the beginning of the block's label.” 
More concise than what I had written.

Samq64: It doesn't seem to be committed

Samq64 (approved):
Thanks! There's just a couple of mxmou's code comment suggestions still need to be committed and one more look from him.

addons/editor-theme3/theme3.js
FlamedDogo99: Oops, thats actually a fix for enabling editor-theme3 with the edit custom block modal already open

addons/editor-theme3/theme3.js
FlamedDogo99: Instead of storing a poorly named copy in each field, this function is now shared through a module imported in both addons. I've also added comments to explain this better

FlamedDogo99:
> Thanks! A couple of mxmou's code comment suggestions still need to be committed and then I'm waiting for one more look from him.

Thank you! And thanks for catching the missed commits, I (obviously 🤦) totally missed that.

FlamedDogo99:
Shoot I just realized that enabling the addon, disabling the addon, opening the edit custom block modal and then reenabling the addon makes the modal not update correctly, ~~so you might have to review again.~~

mxmou (approved):

WorldLanguages:
Will be merged for the next release (not the upcoming one)

addons/recolor-custom-blocks/addon.json
```diff
@@ -0,0 +1,35 @@
...
+    {
+      "name": "Hide color prefixes",
+      "id": "hideColorPrefix",
+      "type": "boolean",
+      "default": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.43.0"
```
Samq64:
```suggestion
  "versionAdded": "1.44.0"
```

PR #8484, opened 2025-08-04
Responsive update notification

mxmou:
### Changes

Updates the CSS for the update notification to support narrow screens.

I decided to hide the SA logo because it takes up a lot of space without adding any information. If we ever start adding [screenshots](https://github.com/ScratchAddons/ScratchAddons/pull/1016) to update notifications again (I think we should), this will need to be changed.

### Tests

Tested on Firefox for Android.

mobile_update_notif

### DIFF
--- a/content-scripts/cs.js
+++ b/content-scripts/cs.js
@@ -604,24 +604,6 @@ const showBanner = () => {
   const notifOuterBody = document.createElement("div");
   const notifInnerBody = Object.assign(document.createElement("div"), {
     id: "sa-notification",
-    style: `
-    position: fixed;
-    bottom: 20px;
-    right: 20px;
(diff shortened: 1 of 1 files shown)


Samq64 (approved):

DNin01:
Maybe the SA logo could be smaller and placed next to the notification title.

mxmou:
> Maybe the SA logo could be smaller and placed next to the notification title.

Let's do that if the v1.43 notification includes the SA logo - but I think it would be better to use an image that's actually related to the update.

WorldLanguages:
Where would you fit an image?

mxmou:
> Where would you fit an image?

Between the title and the rest of the text.

PR #8492, opened 2025-08-09
Fix debugger header buttons

### LINKED ISSUE #8216: Closing button is not displayed in the debugger
### Description

In French and undoubtedly in other language, the closing button is not displayed because of the lack of space. [Discord](https://discord.com/channels/806602307750985799/1083808657290039356)

### Steps to reproduce

1. Change the scratch language
2. Call a breakpoint
3. Open the debugger

https://github.com/user-attachments/assets/081dafe5-ddc5-4f79-8805-fae7fdf81032

### Environment

Browser: Google Chrome: 134.0.6998.118 (Build officiel) (64 bits) (cohort: Control) 
Operating system: Windows

Valmontechno:
Resolves #8216

### Changes

Fixes the bug where not all buttons in the debugger header were displayed in full.
Now, all the buttons are displayed correctly no matter the language.
The right buttons can be reduced if there is no enough space and they all will be the same dimension. In addition the buttons occupy alway the same size in all tabs.
<p>
image
image
</p>

Also, fixes a minor bug that made that the resume button isn't always highlighted on hover.

### Tests

Google Chrome: 139.0.7258.66
Microsoft Edge: 139.0.3405.86
OS: Windows 10

Tested languages: English, French, Spanish, Portuguese, Japanese, Korean

### DIFF
--- a/addons/debugger/style.css
+++ b/addons/debugger/style.css
@@ -107,7 +107,7 @@
   height: 20px;
 }
 
-.sa-debugger-unpause {
+.sa-debugger-unpause:not(:hover) {
   animation: saDebuggerUnpause 2s infinite alternate;
 }
 
--- a/addons/debugger/style.css
+++ b/addons/debugger/style.css
@@ -40,6 +40,7 @@
 .sa-debugger-interface [class*="card_header-buttons_"] {
   background-color: #29beb8;
   border-color: #3aa8a4;
+  text-wrap-mode: nowrap;
 }
 
 .sa-debugger-interface h1 {
@@ -87,6 +88,18 @@
 }
 .sa-debugger-tabs li.sa-debugger-tab-selected img {
   filter: none;
+}
+
+.sa-debugger-header-buttons {
+  width: 100%;
+  display: grid;
+  grid-auto-flow: column;
+  grid-template-columns: repeat(4, 1fr);
+  direction: rtl;
+}
+
+.sa-debugger-header-buttons > div {
+  min-width: 0;
 }
 
 .sa-debugger-header-buttons img {
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -514,11 +514,11 @@ export default async function ({ addon, console, msg }) {
     tabContentContainer.appendChild(tab.content);
 
     removeAllChildren(buttonContainerElement);
-    buttonContainerElement.appendChild(unpauseButton.element);
+    buttonContainerElement.prepend(unpauseButton.element);
     for (const button of tab.buttons) {
-      buttonContainerElement.appendChild(button.element);
+      buttonContainerElement.prepend(button.element);
     }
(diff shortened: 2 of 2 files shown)


addons/debugger/style.css
```diff
@@ -89,12 +90,24 @@
   filter: none;
 }
 
+.sa-debugger-header-buttons {
+  width: 100%;
+  display: grid;
+  grid-auto-flow: column;
+  grid-template-columns: repeat(4, 1fr);
+  direction: rtl;
```
Samq64, line 98: This is a little strange and might cause problems. The buttons were reversed at some point while testing but I can't replicate that right now.

mxmou: Also, from the CSS specification: "Authors should not use `direction` in HTML documents."

mxmou:
> Now, all the buttons are displayed correctly no matter the language.

Not in Portuguese:
image

Another problem with the approach proposed in this PR is that it assumes the translations and the width of the popup (see #5222 and #5386) won't change. To fully solve the issue, the tab bar either has to be scrollable or hide the labels when there isn't enough space.

Joeclinton1:
One idea I was thinking is to move the top buttons down to the bottom.

mxmou:
> One idea I was thinking is to move the top buttons down to the bottom.

There could also be two rows of buttons at the top: one for tabs and the close button and the other for the rest. Something like this:

image

Joeclinton1:
> There could also be two rows of buttons at the top: one for tabs and the close button and the other for the rest.

From a design point of view it unfortunately looks pretty ugly. You could centre it though and it would help, but still kind of ugly.
image
on my timing tab for debugger addon I have this second layer that you can pop out. So I would need to figure out how that reconciles with a second layer. I'll just mockup a version where the tools are placed at the bottom, so we can see that too.

Joeclinton1:
image

If you see what I did here, just making the debugger addon wider kind of solves the whole problem, and is by far the easiest solution

Samq64:
Some people already find the current size too big (#5222).

Joeclinton1:
image

Placing at bottom looks like this. Or in purple it looks like:

image

Joeclinton1:
image

Another design for super narrow that would require tooltips

And if you can do that then you can do the same for the header:
image

Samq64:
The labeless design would work well for `editor-compact` but I'm not sure if it makes sense as the default.

Could this PR be closed and discussion moved to an issue?

WorldLanguages:
> Could this PR be closed and discussion moved to an issue?

Hi, was it moved somewhere else? (Maybe it was but no one linked to it)

Joeclinton1:
> > Could this PR be closed and discussion moved to an issue?
> 
> Hi, was it moved somewhere else? (Maybe it was but no one linked to it)

Moved to the issue.https://github.com/ScratchAddons/ScratchAddons/issues/8216

PR #8499, opened 2025-08-11
Fix ESLint action

Samq64:
Removes `continue-on-error: true` from the ESLint action. Does anyone know why that was there since it's the last step?

### Tests

I'll see when I click create

### DIFF
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -30,4 +30,4 @@ jobs:
         run: |
           npm i globals @eslint/js @eslint/eslintrc
           npx eslint . --config eslint.config.mjs
-        continue-on-error: true
+


WorldLanguages:
> Should `--max-warnings=0` be passed or should the action also fail for warnings?

How many warnings do we have in current code?

Samq64:
None, i just thought I'd mention the flag since Dnin01 did in the issue.

WorldLanguages:
If it won't fail on current code then it wouldn't hurt to add it. I have no idea if it makes a difference at all with our eslint config, though.

mxmou:
> I have no idea if it makes a difference at all with our eslint config, though.

It does, but [only for a single rule](https://github.com/ScratchAddons/ScratchAddons/blob/8f2fa0d/eslint.config.mjs#L48). It might make more sense to change the config and make that an error.

DNin01:
If we import them, ESLint recommendations may also contain warning-level rules.

DNin01 (approved):
Let's try it.

PR #8504, opened 2025-08-13
Move restore messages to `_general.json`

mxmou:
### Changes

Moves the strings for restoring multiple items to `_general.json` to avoid duplication and to make sure they're translated identically in both folders and delete-others.

### Tests

Tested on Edge and Firefox.


WorldLanguages:
To be fair, Transifex has built tools to allow translators to reuse existing translations, so it's not a big deal unless it's repeated multiple times

mxmou:
> To be fair, Transifex has built tools to allow translators to reuse existing translations, so it's not a big deal unless it's repeated multiple times

True, but it's still easy to accidentally translate identical strings differently because changing one translation doesn't automatically update the others.

PR #8516, opened 2025-08-24
animated-thumb: Cleanup after Scratch update

Samq64:
### Changes

Cleans up the thumbnail setter by removing support for the version used before Scratch's update. Also replaces the logo with the upload icon and shows errors in an `alert()` instead of the modal. I've left the 2 added by SA messages as-is.

### Reason for changes

It's mostly dead code now and showing the error message as the only modal step doesn't seem any better than using `alert()` but that can be reverted.

### Tests

Tested on Chromium.


mxmou:
I think `alert()` should generally be avoided because it blocks the main thread. I suggest keeping the modal in this PR so that the removal of outdated code is separate from UI changes, but we should eventually change it to something similar to Scratch's alerts:

<p>image</p>

An `addon.tab.alert()` API might be useful for other addons as well.

mxmou (approved):
You can probably also remove the `info-box` and `info-tooltip` messages and the code that shows them - they won't be relevant anymore when the next update is released.

mxmou (approved):

PR #8518, opened 2025-08-24
Fix forum post coundown triggering on edit

### LINKED ISSUE #8517: `forum-post-countdown`: Editing a post counts as creating one
### Description

Editing a post leads to the 60s countdown being started.

### Steps to reproduce

1. Edit a post.

_No response_

### Environment

Browser: Zen 1.41.11b (Firefox 141.0.3)
Scratch Addons version: 1.43.0
Operating system: Windows

Samq64:
Resolves #8517

### Changes

Checks the success message to ensure it's not an edit ~~and run on `topics` instead of `newPostScreens` since it doesn't need those pages anymore.~~

### Tests

Tested by editing a post.

### DIFF
--- a/addons/forum-post-countdown/userscript.js
+++ b/addons/forum-post-countdown/userscript.js
@@ -1,10 +1,10 @@
 export default async function ({ addon, msg }) {
   const submitButton = document.querySelector("#djangobbwrap .form-submit [type=submit]");
 
-  let message = document.querySelector(".success");
+  const message = document.querySelector(".success")?.innerText;
 
-  // This message is not translated by Scratch
-  if (message && message.innerText !== "Post updated.") {
+  // These messages are not translated by Scratch
+  if (message === "Your reply saved." || message === "Topic saved.") {
     localStorage.setItem("sa-forum-post-countdown", Date.now());
   }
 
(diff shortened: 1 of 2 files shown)


CST1229:
does topic creation have the 60-second rule? (because if so then the countdown needs to appear there)

mxmou (approved):

mxmou:
Technically, the notices such as "Reply saved." can appear on any page because of the way Django's notice system works. The addon already works fine almost all the time, but a more reliable implementation would be to run the userscript on all forum pages (in case the notice appears on a different page) and compare the message with those for creating a topic and adding a reply (to exclude signature updates, for example).

Samq64:
Could someone give this one more quick look?

mxmou:
This is unfortunately hard to test without spamming the forums.

addons/forum-post-countdown/userscript.js
```diff
@@ -1,9 +1,10 @@
 export default async function ({ addon, msg }) {
   const submitButton = document.querySelector("#djangobbwrap .form-submit [type=submit]");
 
-  let message = document.querySelector(".success");
+  const message = document.querySelector(".success")?.innerText;
 
-  if (message) {
+  // These messages are not translated by Scratch
+  if (message === "Your reply saved." || message === "Topic saved.") {
```
mxmou, line 7: I've just verified that the message which appears after making a post is actually "Reply saved":

Image

PR #8536, opened 2025-09-03
New Blockly updates

mxmou:
### Changes

Updates to modern Blockly code:
* Uses Scratch's version of `clearWorkspaceAndLoadFromXml()` that loads local and cloud variables correctly.
* Uses Blockly's new built-in separator feature for context menus. Also updates the `createBlockContextMenu` API to make the same syntax work with old Blockly.
* devtools can now use a simpler workaround to copy multiple blocks.
* The code in theme3 that changes the color of matrix dropdowns is now only needed for old Blockly.

### Tests

Tested both new Blockly (locally) and the current Scratch website on Edge and Firefox.


Samq64:
The separators have less contrast then before on new Blockly when dark mode is enabled, and especially when colored context menus is as well.

PR #8538, opened 2025-09-04
`recolor-custom-blocks` & `reorder-custom-inputs`: support `editor-compact`

Samq64:
### Changes

Fixes compatability between recolor custom blocks and compact editor. I set the height to `1.9rem` since it's very close to what it was before. `!important` is used to override `[class*="modal_modal-content"] [class*="box_box"] button` but a higher specificity might be better.

Also fixes the alignment of the icons added by `reorder-custom-inputs` when `editor-compact` is enabled.

### Tests

Tested on Chromium


mxmou (commented):
> I set the height to `1.9rem` since it's very close to what it was before.

I suggest rounding it to 2rem. Everything else looks good to me.

mxmou (approved):

PR #8540, opened 2025-09-06
Added support for saving multiple gamepad settings per project

stephentine:
### Reason for changes

**Multiplayer Scenario**
Recently I've been working on a local multiplayer game where you connect two gamepads to control two different players. The current gamepad addon can only save one of those configurations, meaning I have to re-write one configuration every time

**Multiple Different Gamepads Scenario**
Another issue is that if I have a game where I save a configuration for one type of gamepad (ex. an N64 controller), but I sometimes want to play the game with a different gamepad (ex. a Nintendo Pro Controller), then I need to re-write the 2nd controller's configuration every time

### Changes

Saving your gamepad settings now saves _every_ connected gamepad's configuration independently. Whenever you connect a gamepad, it will try to use the configuration that best matches it (it will prefer configurations that match the gamepad's type and index). The index matching handles multiple gamepads of the same type (i.e. in the case of the _Multiplayer Scenario_ above) while the type matching handles multiple different gamepads (i.e. in the case of the other scenario).

If you only ever use one gamepad in a project, then the addon functions exactly the same as it did before these changes. If you had a configuration saved previously (before upgrading to this version), then those settings are kept and will be used for newly connected gamepads

**Configuration Selection Logic**
When a gamepad is connected, it will perform the following logic to choose its configuration:

If a saved configuration with the same type and index as this gamepad is found, use it
else, if a saved configuration with the same type as this gamepad is found, use it
else, use the configuration that was saved earliest (the oldest one in the project)

### Tests

- Connected two gamepads of the same type and saved unique configurations for each
  - After reconnecting the gamepads, verified that each gamepad picked up its respective configuration
- Connected two gamepads of different types (ex. Xbox & Nintendo Joy-Con) and saved unique configurations for each
  - After reconnecting the gamepads, verified that each gamepad picked up its respective configuration
- Saved a setting using the older version of the addon, then upgraded to this version
  - After connecting a new gamepad, verified that the old configuration was used

Tested in Chrome and Edge

### DIFF
--- a/addons/gamepad/userscript.js
+++ b/addons/gamepad/userscript.js
@@ -60,8 +60,7 @@ export default async function ({ addon, console, msg }) {
     }
     return null;
   };
-  const parseOptionsComment = () => {
-    const comment = findOptionsComment();
+  const parseOptionsComment = (comment) => {
     if (!comment) {
       return null;
     }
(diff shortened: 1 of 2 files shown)


addons/gamepad/userscript.js
```diff
@@ -71,27 +70,66 @@ export default async function ({ addon, console, msg }) {
...
+      // If a stored mapping exists for this exact gamepad id and index, use it
+      matchingOption = optionsArray.find((gamepads) => gamepads.id === gamepadId && gamepads.index === gamepadIndex);
+      if (!matchingOption) {
+        // Otherwise, if a stored mapping exists for this gamepad id at all, use it
+        matchingOption = optionsArray.find((gamepads) => gamepads.id === gamepadId);
+        if (!matchingOption) {
+          // Otherwise, use the first stored mapping no matter what it is
+          matchingOption = optionsArray.at(0);
+        }
+      }
```
DNin01, line 134: I would expect that the configuration of the same index be preferred over the first one even if the gamepad isn't the same. Oftentimes, a player's gamepad won't have the exact same ID as the project creator's gamepad, and it can even vary between browsers.

```suggestion
      // If a stored mapping exists for this exact gamepad id and index, use it
      matchingOption = optionsArray.find((gamepads) => gamepads.id === gamepadId && gamepads.index === gamepadIndex);
      if (!matchingOption) {
        // Otherwise, if a stored mapping exists for this gamepad id at all, use it
        matchingOption = optionsArray.find((gamepads) => gamepads.id === gamepadId);
        if (!matchingOption) {
          // Otherwise, use the stored mapping at the same index if it exists
          matchingOption = optionsArray.find((gamepads) => gamepads.index === gamepadIndex);
          if (!matchingOption) {
            // Otherwise, use the first stored mapping no matter what it is
            matchingOption = optionsArray.at(0);
          }
        }
      }
```

stephentine: > I would expect that the configuration of the same index be preferred over the first one even if the gamepad isn't the same. Oftentimes, a player's gamepad won't have the exact same ID as the project creator's gamepad, and it can even vary between browsers.

Good point, I didn't think about that. I just committed your changes

WorldLanguages:
@DNin01 One test is enough, I think.

WorldLanguages:
cc @GarboMuffin This changes the JSON schema ("protocol") for gamepad settings stored in-project.

PR #8545, opened 2025-09-07
Forum warning: mention where to disable extensions

Samq64:
### Changes

Mention `chrome://extensions` and `about:addons` in the forum warning when reporting a bug.

### Reason for changes

Some users don't know where to go to disable extensions.

### Tests

Tested on Firefox and Chromium

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -216,7 +216,7 @@
     "message": "report it here"
   },
   "forumWarning": {
-    "message": "Message added by the Scratch Addons extension: make sure the bug you're about to report still happens when all browser extensions are disabled from $2, including Scratch Addons. If you believe a bug is caused by Scratch Addons, please $1."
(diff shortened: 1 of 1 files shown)


_locales/en/messages.json
```diff
@@ -216,7 +216,7 @@
     "message": "report it here"
   },
   "forumWarning": {
-    "message": "Message added by the Scratch Addons extension: make sure the bug you're about to report still happens when all browser extensions are disabled, including Scratch Addons. If you believe a bug is caused by Scratch Addons, please $1."
+    "message": "Message added by the Scratch Addons extension: make sure the bug you're about to report still happens when all browser extensions are disabled from $2, including Scratch Addons. If you believe a bug is caused by Scratch Addons, please $1."
```
mxmou: A little more explanation might be helpful:
```suggestion
    "message": "Message added by the Scratch Addons extension: make sure the bug you're about to report still happens when all browser extensions are disabled. This can be done by going to $2 and turning every extension off, including Scratch Addons. If you believe a bug is caused by Scratch Addons, please $1."
```

DNin01:
We could also make the same mention on our feedback page, FAQ, etc.

Samq64:
Using an incognito window is another option assuming the user didn't grant extensions access there.

PR #8549, opened 2025-09-10
Move right debugger header buttons to footer

### LINKED ISSUE #8216: Closing button is not displayed in the debugger

Joeclinton1:
Resolves #8216

### Changes

Header buttons moved to a new footer, except for close button that remains.
image

### Reason for changes

Header was crowded. Was preventing us from letting users resize the debugger modal, aswell as causing bugs like #8216 where icons were pushed off when long translations went beyond the fixed width container which was designed for English.

### Tests

It doesn't do anything other than move the buttons, so it's purely an aesthetic change. I checked the buttons still worked and the on hover effects worked.

---

This is one possible design and more design's are discussed in the issue. I just thought the choice was quite clear so I proactively made this PR. It took me less than 10 minutes so if I to redesign it, I don't mind.

### DIFF
--- a/addons/debugger/style.css
+++ b/addons/debugger/style.css
@@ -262,3 +262,55 @@
 .sa-debugger-block-preview[data-shape="stacked"] {
   border-radius: 3px;
 }
+
+.sa-debugger-footer {
+  border-top: 1px solid rgba(0, 0, 0, 0.15);
+  display: flex;
+  justify-content: center;
+  padding: 8px 15px;
+  width: 100%;
+  box-sizing: border-box;
+}
+
+.sa-debugger-footer-buttons {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: 10px;
+}
+
+.sa-debugger-footer-buttons > div {
+  background-color: #29beb8;
+  border: 1px solid #3aa8a4;
+  border-radius: 4px;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding: 4px 8px;
+  color: white;
+  cursor: pointer;
+  gap: 4px;
+}
+
+.sa-debugger-footer-buttons > div:hover {
+  background-color: #3aa8a4;
+}
+
+.sa-debugger-footer-buttons img {
+  width: 16px;
+  height: 16px;
+  filter: brightness(0) invert(1);
+}
+
+.sa-debugger-footer-buttons span {
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.sa-debugger-header-buttons > div[title] span {
+  display: none;
+}
+
+.sa-debugger-footer-buttons .sa-debugger-unpause {
+  animation: none !important;
+}
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -110,6 +110,12 @@ export default async function ({ addon, console, msg }) {
   const tabContentContainer = Object.assign(document.createElement("div"), {
     className: "sa-debugger-tab-content",
   });
+  const interfaceFooter = Object.assign(document.createElement("div"), {
+    className: "sa-debugger-footer",
+  });
+  const footerButtonContainer = Object.assign(document.createElement("div"), {
+    className: "sa-debugger-footer-buttons",
+  });
 
   let isInterfaceVisible = false;
   const setInterfaceVisible = (_isVisible) => {
@@ -159,7 +165,8 @@ export default async function ({ addon, console, msg }) {
   interfaceHeader.addEventListener("mousedown", handleStartDrag);
 
   interfaceHeader.append(tabListElement, buttonContainerElement);
-  interfaceContainer.append(interfaceHeader, tabContentContainer);
+  interfaceFooter.appendChild(footerButtonContainer);
+  interfaceContainer.append(interfaceHeader, tabContentContainer, interfaceFooter);
   document.body.append(interfaceContainer);
   moveInterface(0, 0); // necessary to initialize position if running scratch-gui locally
 
@@ -220,6 +227,7 @@ export default async function ({ addon, console, msg }) {
   const closeButton = createHeaderButton({
     text: msg("close"),
     icon: addon.self.dir + "/icons/close.svg",
+    description: msg("close"),
   });
   closeButton.element.addEventListener("click", () => setInterfaceVisible(false));
 
@@ -514,11 +522,13 @@ export default async function ({ addon, console, msg }) {
     tabContentContainer.appendChild(tab.content);
 
     removeAllChildren(buttonContainerElement);
-    buttonContainerElement.appendChild(unpauseButton.element);
-    for (const button of tab.buttons) {
-      buttonContainerElement.appendChild(button.element);
-    }
     buttonContainerElement.appendChild(closeButton.element);
+    
+    removeAllChildren(footerButtonContainer);
+    footerButtonContainer.appendChild(unpauseButton.element);
+    for (const button of tab.buttons) {
+      footerButtonContainer.appendChild(button.element);
+    }
 
     if (isInterfaceVisible) {
       activeTab.show();
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -523,7 +523,7 @@ export default async function ({ addon, console, msg }) {
 
     removeAllChildren(buttonContainerElement);
     buttonContainerElement.appendChild(closeButton.element);
-    
+
     removeAllChildren(footerButtonContainer);
     footerButtonContainer.appendChild(unpauseButton.element);
     for (const button of tab.buttons) {


DNin01:
If the buttons at the bottom change depending on the selected tab (e.g., "clear" and "export" only available in Logs and "step" and "pause" only available in Threads), then the number of buttons on the popover's header bar would always stay the same, which could improve UX.

Joeclinton1:
> If the buttons at the bottom change depending on the selected tab (e.g., "clear" and "export" only available in Logs and "step" and "pause" only available in Threads), then the number of buttons on the popover's header bar would always stay the same, which could improve UX.

Unsure what you're trying to say here. But yes the number of buttons in the footer changes, and stays constant in the header.

addons/debugger/style.css
```diff
@@ -262,3 +262,55 @@
...
+  font-weight: 500;
+}
+
+.sa-debugger-header-buttons > div[title] span {
+  display: none;
+}
+
+.sa-debugger-footer-buttons .sa-debugger-unpause {
+  animation: none !important;
+}
```
Samq64: Could these be removed instead of hidden wih CSS?

addons/debugger/userscript.js
```diff
@@ -514,11 +522,21 @@ export default async function ({ addon, console, msg }) {
...
+    footerButtonContainer.appendChild(unpauseButton.element);
     for (const button of tab.buttons) {
-      buttonContainerElement.appendChild(button.element);
+      footerButtonContainer.appendChild(button.element);
     }
-    buttonContainerElement.appendChild(closeButton.element);
+
+    // Hide footer if all buttons are hidden
+    const allButtons = [unpauseButton.element, ...tab.buttons.map((b) => b.element)];
+    const hasVisibleButtons = allButtons.some((button) => {
```
Samq64: This doesn't always show the buttons and when it does on the threads tab resuming the project leaves the footer border behind. The easiest solution is to use CSS but `:has()` isn't supported in our minimum Chrome version.

Joeclinton1: I tried `:has()` but that doesn't work if the footer isn't empty but has items that are there just hidden. The best solution will be in the direction that I took.

DNin01: I fixed it!

Joeclinton1: @DNin01 Thanks! Was just about to do something similar to your solution, so you saved me some time. I tested your code and it works.

addons/debugger/style.css
Joeclinton1: I agree, no point hiding if we don't need it. I am also not entirely sure what the animation was even ever used for. My choice to do it with css was a bit lazy in retrospect. I think DNin01's commits already solve this. (It's good that you're paying attention to the details in the commit)

addons/debugger/style.css
```diff
@@ -94,17 +94,48 @@
...
+.sa-debugger-footer-buttons img {
+  width: 16px;
+  height: 16px;
+  filter: brightness(0) invert(1);
+}
+
+.sa-debugger-footer-buttons span {
+  font-size: 12px;
+  font-weight: 500;
 }
```
DNin01, line 164: In cleaning things up a bit, I removed an animation. I don't think it's necessary, but if you really want the resume button to be animated, here's a little something:
```suggestion

.sa-debugger-unpause img {
  animation: saDebuggerUnpause 1s ease-in-out infinite alternate;
}

@keyframes saDebuggerUnpause {
  0% {
    opacity: 1;
  }
  100% {
    opacity: 0.7;
  }
}

```

Joeclinton1: I didn't add the animation I was surprised when I moved the button to the footer and an animation started playing when I've never seen it before.

Joeclinton1:
https://github.com/ScratchAddons/ScratchAddons/issues/5222 We could even kill two birds with one stone if we made the modal resizable too. Run the below to test:

```js
const style = document.createElement("style");
style.textContent = `
.sa-debugger-tabs {
    overflow: hidden;
}
.sa-debugger-interface {
    resize: auto;
    min-width: 300px;
    min-height: 200px;
}
`;
document.head.appendChild(style);
```


image

addons/debugger/style.css
```diff
@@ -94,17 +94,48 @@
...
 
-.sa-debugger-unpause {
-  animation: saDebuggerUnpause 2s infinite alternate;
+.sa-debugger-footer {
+  border-top: 1px solid rgba(0, 0, 0, 0.15);
+  display: flex;
+  justify-content: center;
+  padding: 8px 15px;
+  width: 100%;
+  box-sizing: border-box;
```
mxmou, line 108:
```suggestion
  box-sizing: border-box;
  cursor: auto;
```
The footer currently inherits `cursor: move` from the parent element, even though it isn't draggable.

addons/debugger/style.css
```diff
@@ -94,17 +94,48 @@
...
-@keyframes saDebuggerUnpause {
-  0% {
-    background-color: rgba(0, 0, 0, 0.15);
-  }
-  100% {
-    background-color: rgba(0, 0, 0, 0);
-  }
+.sa-debugger-footer-buttons > div {
+  background-color: #29beb8;
+  border: 1px solid #3aa8a4;
```
mxmou: I like the way the buttons look, but removing the border and changing the label font weight to bold would make them more Scratch-like:
<p>Image</p>
If we do this, the tabs should probably also be bold for consistency.

addons/debugger/style.css
```diff
@@ -94,17 +94,48 @@
...
+}
+
+.sa-debugger-footer-buttons > div:hover {
+  background-color: #3aa8a4;
+}
+
+.sa-debugger-footer-buttons img {
+  width: 16px;
+  height: 16px;
+  filter: brightness(0) invert(1);
```
Samq64: These icons don't need any filters since they are already white.
```suggestion
```

Joeclinton1:
I've added the resizing, which technically should be a second pr, but if we approve it in this one we'll save everyone some time.

addons/debugger/style.css
```diff
@@ -55,7 +58,9 @@
   align-items: center;
   padding: 0 15px;
   font-size: 0.75rem;
+  overflow: hidden;
```
mxmou:
```suggestion
  overflow: auto;
  scrollbar-width: thin;
```
The scrollbar might be ugly but it would prevent the Performance tab from being (almost) invisible in languages with longer labels such as Portuguese:
Image
(Also, the text on the tab probably shouldn't wrap.)

addons/debugger/userscript.js
```diff
@@ -211,15 +218,21 @@ export default async function ({ addon, console, msg }) {
   });
   unpauseButton.element.classList.add("sa-debugger-unpause");
```
mxmou, line 219: This is now unused (unless you add the animation suggested by DNin01).

addons/debugger/userscript.js
mxmou, line 1: The `createHeaderButton` function should probably be renamed to just `createButton`.

addons/debugger/style.css
Joeclinton1: Partially the reason I did this PR is to get ready for the timing tools tab, which is going to be completely hidden on the right side of performance. So my thinking is if users resized the modal, it would be temporary then they'd resize back to the original to see the tabs again. I'll try your design and see what it looks like though.

Joeclinton1: Btw, the translation should be "Threads" https://pt.wikipedia.org/wiki/Thread_(computa%C3%A7%C3%A3o). A multi word translation shouldn't be allowed.

Samq64: The tabs that don't need blocks could be optional to free up even more header space for those who want to keep the window small, but that's for another PR.

Joeclinton1: image

This looked best in my testing.

addons/debugger/userscript.js
Joeclinton1: Executive decision, I'm going to add a variation of the animation as it is a little unclear what you're meant to do when the debugger appears the first time.

addons/debugger/style.css
Joeclinton1: .sa-debugger-unpause img {
  animation: saBorderPulse 1.2s ease-in-out infinite alternate;
}

@keyframes saBorderPulse {
  0%   { box-shadow: 0 0 4px 2px rgb(41 190 184 / 76%); }
  100% { box-shadow: 0 0 10px 5px rgb(41 190 184 / 0%); }
}

Went with this. It's a heartbeat box shadow, should attract more attention to the button than the small opacity change. The box shadow is similar to the breakpoint box shadow, so it should feel consistent.

addons/debugger/style.css
```diff
@@ -89,21 +95,78 @@
...
+  padding-right: 10px;
+}
+
 .sa-debugger-unpause {
-  animation: saDebuggerUnpause 2s infinite alternate;
+  animation: saBorderPulse 1.2s ease-in-out infinite alternate;
+}
+.sa-debugger-unpause img {
+  animation: saBorderPulse 1.2s ease-in-out infinite alternate;
 }
```
DNin01: The animation probably should not apply to the image within the button.

addons/debugger/style.css
```diff
@@ -89,21 +95,78 @@
...
+@keyframes saBorderPulse {
   0% {
-    background-color: rgba(0, 0, 0, 0.15);
+    box-shadow: 0 0 4px 2px rgb(41 190 184 / 76%);
   }
   100% {
-    background-color: rgba(0, 0, 0, 0);
+    box-shadow: 0 0 10px 5px rgb(41 190 184 / 0%);
   }
 }
```
DNin01: I think this animation is a little too distracting. Maybe slow it down or make it more subtle?
```suggestion
.sa-debugger-unpause {
  animation: saBorderPulse 1.5s ease-in-out infinite alternate;
}

@keyframes saBorderPulse {
  0% {
    box-shadow: 0 0 6px 2px rgb(41 190 184 / 0%);
  }
  100% {
    box-shadow: 0 0 6px 2px rgb(41 190 184 / 76%);
  }
}
```

addons/debugger/style.css
```diff
@@ -35,6 +35,9 @@
   background-color: white;
   width: 565px;
   height: 26rem;
+  resize: auto;
```
DNin01, line 38: On Chrome, the resize anchor is positioned next to the down icon on the scroll bar, causing the content to scroll down while resizing the popover, if there is a scroll bar.

addons/debugger/addon.json
```diff
@@ -84,9 +84,8 @@
   "tags": ["editor", "codeEditor", "featured"],
   "versionAdded": "1.16.0",
   "latestUpdate": {
-    "version": "1.33.0",
-    "temporaryNotice": "Graph animation is now disabled by default to increase performance. You can re-enable it if you want.",
-    "newSettings": ["fancy_graphs"]
+    "version": "1.44.0",
+    "temporaryNotice": "The debugger modal is now resizable."
```
DNin01: The debugger dialog is technically non-modal because it does not render the rest of the page non-interactive.

BTW, I think we say "modal" too much. 😁

```suggestion
    "temporaryNotice": "The debugger dialog is now resizable."
```

addons/debugger/style.css
Joeclinton1: Sorry that was a mistake

addons/debugger/style.css
Joeclinton1: Can't replicate sorry

Joeclinton1:
> Would it be possible to make the popover resizable in cardinal directions, not just by dragging the bottom-right edge?

It can't be done with native CSS. Would take 50 lines of code to do something that can be easily worked around with the existing corner resize handle.

addons/debugger/style.css
```diff
@@ -89,21 +95,75 @@
...
+  height: 100%;
+}
+
 .sa-debugger-header-buttons img {
   width: 20px;
   height: 20px;
 }
 
+.sa-debugger-footer {
+  border-top: 1px solid rgba(0, 0, 0, 0.15);
```
Samq64:
```suggestion
  border-top: 1px solid var(--editorDarkMode-border, rgba(0, 0, 0, 0.15));
```

addons/debugger/style.css
```diff
@@ -89,21 +95,75 @@
...
+  height: 16px;
+}
+
+.sa-debugger-unpause-container {
+  display: flex;
+  align-items: center;
+}
+
+.sa-debugger-unpause-container:not(:last-child) {
+  border-right: 1px solid rgba(0, 0, 0, 0.15);
```
Samq64:
```suggestion
  border-right: 1px solid var(--editorDarkMode-border, rgba(0, 0, 0, 0.15));
```

addons/debugger/style.css
Samq64: I can but only on the performace tab while unpaused, so it's not a big deal.

addons/debugger/style.css
```diff
@@ -53,9 +56,14 @@
   margin: 0;
   display: flex;
   align-items: center;
-  padding: 0 15px;
+  align-content: center;
+  flex-wrap: wrap;
+  gap: 10px;
+  padding: 7px 15px 7px 15px;
```
Samq64: Just a nitpick
```suggestion
  padding: 7px 15px;
```

Joeclinton1:
Thank you @DNin01 @mxmou @Samq64 for your help, without you guys I would have missed a lot of details. You've made it much better!

addons/debugger/style.css
Joeclinton1: Ah ok, I had it paused. I still can't replicate the scrolling whilst resizing problem, even if they are very close together. The only solution would be a permanent footer, which I can do easily of course, since there's code hiding it right now and that would just have to be disabled and instead we simple hide the border. But I think it looks better this way and uses space more conservatively.

Joeclinton1:
image

What do you guys think of using the modal api's close button for consistent ui? (I think it looks good and makes sense, but since the PR has already been approved and this is an unreviewed change on top of that, I'll wait before making the commit)

DNin01:
I think the close button should be consistent with Scratch, so let's update it.

addons/debugger/style.css
```diff
@@ -89,21 +95,70 @@
...
 
-@keyframes saDebuggerUnpause {
+@keyframes saBorderPulse {
   0% {
-    background-color: rgba(0, 0, 0, 0.15);
+    box-shadow: 0 0 6px 2px rgb(41 190 184 / 0%);
   }
   100% {
-    background-color: rgba(0, 0, 0, 0);
+    box-shadow: 0 0 6px 2px rgb(41 190 184 / 76%);
```
mxmou: A shadow without blur would be more similar to Scratch's own UI:
```suggestion
    box-shadow: 0 0 0 4px rgb(41 190 184 / 35%);
```

addons/debugger/style.css
```diff
@@ -89,21 +95,70 @@
...
 .sa-debugger-unpause {
-  animation: saDebuggerUnpause 2s infinite alternate;
+  animation: saBorderPulse 1.5s ease-in-out infinite alternate;
 }
 
-@keyframes saDebuggerUnpause {
+@keyframes saBorderPulse {
   0% {
-    background-color: rgba(0, 0, 0, 0.15);
+    box-shadow: 0 0 6px 2px rgb(41 190 184 / 0%);
```
mxmou:
```suggestion
    box-shadow: 0 0 0 4px rgb(41 190 184 / 0%);
```

addons/debugger/addon.json
mxmou: It's called a "window" in the addon description - maybe it would be better to use the same word for consistency?

Joeclinton1:
Gonna merge this in a couple hours if no-one else wants anything more changed. I think it's perfect.

@DNin01 Sorry for re-requesting a review twice. I just need approval on the close button change, so that we're being thorough with the review process.

addons/debugger/style.css
```diff
@@ -34,7 +34,10 @@
   z-index: 492;
   background-color: white;
   width: 565px;
-  height: 25rem;
+  height: 26rem;
+  resize: auto;
```
DNin01, line 38: Note that the `resize` property has no effect on Firefox for Android.

Not a big deal tho.

Joeclinton1: ok, I don't think I can fix this so will resolve this convo.

PR #8555, opened 2025-09-13
`pause`: New options for auto-pause and pause button visibility

### LINKED ISSUE #5836: Auto-pause on focus leave

DNin01:
Resolves #5836
Same as #5838 by @iqnite but a different implementation, and additional features.

### Automatic pausing

Adds a new setting that, when enabled, allows the project to pause automatically when one of the following things happens:

- Focus leaves the browser window
- You switch browser tabs
- You start typing a comment

...and automatically resume when focus returns.

However, it will not pause automatically if Cloud variables were updated in the last 15 seconds.

### Show/hide pause button

Another new setting was added, Show pause button, which hides the pause button when turned off. When hidden, pause can still be toggled via keyboard shortcut.

This is useful alongside the auto-pause feature if you for some reason only want the auto-pause feature and don't need a button to manually pause the project.

## Additional context

This feature was based on work that ended up becoming #8478, but here, the online feature detection has been stripped down from a complicated algorithm to a much simpler check for activity in the last 15 seconds, a process that ensures that projects with live user interaction aren't messed up by the auto-pausing.

## Tests

Everything works as expected on both projects with and without Cloud variables.

### DIFF
--- a/addons/pause/addon.json
+++ b/addons/pause/addon.json
@@ -1,6 +1,12 @@
 {
   "name": "Pause button",
   "description": "Adds a button to pause the project next to the green flag.",
+  "info": [
+    {
+      "text": "The project can also be paused with Alt+X (Option+X on macOS).",
+      "id": "keybind"
+    }
+  ],
   "credits": [
     {
       "name": "Jeffalo"
@@ -33,7 +39,6 @@
       "type": "boolean",
       "id": "pause-button",
       "name": "Show pause button",
-      "description": "When hidden, you can still pause by pressing Alt+X or Option+X.",
       "default": true
     },
     {
--- a/addons/pause/addon.json
+++ b/addons/pause/addon.json
@@ -40,7 +40,7 @@
       "type": "boolean",
       "id": "auto-pause",
       "name": "Pause automatically",
-      "description": "The project pauses automatically when the window does not have focus. May not work on all projects.",
+      "description": "The project pauses automatically when the window does not have focus. Does not apply to cloud projects.",
       "default": false
     }
   ],
--- a/addons/pause/addon.json
+++ b/addons/pause/addon.json
@@ -1,18 +1,16 @@
 {
   "name": "Pause button",
   "description": "Adds a button to pause the project next to the green flag.",
-  "info": [
-    {
-      "text": "The project can also be paused with Alt+X (Option+X on macOS).",
-      "id": "keybind"
-    }
-  ],
   "credits": [
     {
       "name": "Jeffalo"
     },
     {
       "name": "GarboMuffin"
+    },
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
     }
   ],
   "dynamicEnable": true,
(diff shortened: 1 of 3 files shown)


iqnite:
Should there be a toggle for auto-resume as well?

addons/pause/addon.json
```diff
@@ -30,10 +28,27 @@
...
+      "id": "pause-button",
+      "name": "Show pause button",
+      "description": "When hidden, you can still pause by pressing Alt+X or Option+X.",
+      "default": true
+    },
+    {
+      "type": "boolean",
+      "id": "auto-pause",
+      "name": "Pause automatically",
+      "description": "The project pauses automatically when the window does not have focus. May not work on all projects.",
```
CST1229: to me, "may not work on all projects" implies that this addon has a bug/issue of some kind, not that this behavior is automatically disabled on cloud projects. i think it should just say directly "Does not apply for cloud projects." or something similar.

DNin01:
> Should there be a toggle for auto-resume as well? 

If you were distracted and switch back or if your computer was asleep, that would be ideal so that you have time to get back into action. Are there any other cases where you wouldn't want it to auto-resume?

I could also experiment with having it only auto-resume if you come back **after less than a minute** since the last auto-pause. I think that would be a good balance, however, it wouldn't really work if the pause button was hidden—one of the other additions in this PR. Perhaps there could be something like the big green flag button when a project loads but instead it's a resume button.

iqnite:
> I could also experiment with having it only auto-resume if you come back **after less than a minute** since the last auto-pause. 

I don't think that's what users expect. Instead, there could be an auto-resume toggle, and if it's enabled, a number input could appear, allowing to customize that time.

However, this amount of customization might still be a bit overkill.

> Perhaps there could be something like the big green flag button when a project loads but instead it's a resume button.

This should definitely be a toggle. I pause projects to take a closer look at the stage, and a big button would obstruct the view.

DNin01:
> > I could also experiment with having it only auto-resume if you come back **after less than a minute** since the last auto-pause.
>
> I don't think that's what users expect. Instead, there could be an auto-resume toggle, and if it's enabled, a number input could appear, allowing to customize that time.
> 
> However, this amount of customization might still be a bit overkill.

Yeah, that could be confusing.

addons/pause/addon.json
```diff
@@ -30,10 +28,27 @@
       "matches": ["projects", "projectEmbeds"]
     }
   ],
+  "settings": [
+    {
+      "type": "boolean",
+      "id": "pause-button",
+      "name": "Show pause button",
```
Samq64, line 41: I doubt many users will disable this setting but currently the button always appears after the project finishes loading.

addons/pause/addon.json
```diff
@@ -1,18 +1,16 @@
 {
   "name": "Pause button",
   "description": "Adds a button to pause the project next to the green flag.",
-  "info": [
-    {
-      "text": "The project can also be paused with Alt+X (Option+X on macOS).",
```
Samq64, line 6: I think this should stay, it's not exclusive to the hide setting.

Samq64:
How about auto-resuming after an interaction? That could also be confusing though, especially if the project doesn't register that input.

DNin01:
I think auto-resuming is generally fine if it has only been a few moments since the project was paused and the device didn't go to sleep. Otherwise, I agree that resuming should be manual, **unless** the pause button is hidden, in which case, there wouldn't be an easy way to resume the project so maybe it should always auto-resume then. Complicated...?

> How about auto-resuming after an interaction? That could also be confusing though, especially if the project doesn't register that input.

I like that on top of my previous idea. We could also display a play-button overlay that is like the green flag overlay that shows when a project first loads, but that appears when the project is paused and/or the pause button is hidden. That would make more sense when the pause button is hidden.

PR #8556, opened 2025-09-13
custom-block-shape: new Blockly

mxmou:
### Changes

Updates the last remaining addon to modern Blockly!
name-backpack-items is also listed in the tracking issue, but it likely won't need any changes because it doesn't interact with Blockly.

### Tests


Samq64:
The two versions are pretty inconsistant when setting the padding to 50% while leaving the other settings at 100%:
|New Blockly|Old Blockly|
|-|-|
|image|image|

At one point the top of a hat block was cut off but I can't repro that anymore.

mxmou:
I don't think everything has to look exactly the same as before because the new block rendering code is very different. Some of the changes in these screenshots are actually improvements: the define hat now has the same amount of vertical padding as the normal hat and the notch shape is less weird.

I'll fix the width of number inputs and the height of the repeat block's mouth - those should be affected by the setting but they don't seem to be.

Samq64 (approved):

PR #8560, opened 2025-09-14
Allow dragging debugger on touchscreen

### LINKED ISSUE #5727: Cannot move debug window with touchscreen
### Description

When I try to drag and move the debugging window using my touchscreen on my windows laptop, it doesn't drag but when I use my touchpad it works. My touchscreen works with dragging the vanilla tutorials window.

### Steps to reproduce

1. Use a touchscreen on a windows laptop
2. Turn on the debugging addon
3. Try to drag the window

_No response_

### Environment

Browser: 
Scratch Addons version: 
Operating system:

DNin01:
Resolves #5727

- **Before changes:** The Debugger window could not be dragged on a touch screen, and doing so would scroll the page instead.
- **After changes:** The Debugger window can now be properly dragged by its title bar with both mouse and touch input.

### Tests

Tested with Edge DevTools Device Emulation.

### DIFF
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -163,7 +163,7 @@ export default async function ({ addon, console, msg }) {
     moveInterface(lastX, lastY);
   });
   interfaceHeader.addEventListener("pointerdown", handleStartDrag);
-  interfaceHeader.addEventListener("touchstart", (e) => e.preventDefault());
+  interfaceHeader.addEventListener("touchmove", (e) => e.preventDefault());
 
   interfaceHeader.append(tabListElement, buttonContainerElement);
   interfaceFooter.appendChild(footerButtonContainer);
--- a/addons/debugger/userscript.js
+++ b/addons/debugger/userscript.js
@@ -138,12 +138,12 @@ export default async function ({ addon, console, msg }) {
     mouseOffsetY = e.clientY - interfaceContainer.offsetTop;
     lastX = e.clientX;
     lastY = e.clientY;
-    document.addEventListener("mouseup", handleStopDrag);
-    document.addEventListener("mousemove", handleDragInterface);
+    document.addEventListener("pointerup", handleStopDrag);
+    document.addEventListener("pointermove", handleDragInterface);
   };
   const handleStopDrag = () => {
-    document.removeEventListener("mouseup", handleStopDrag);
-    document.removeEventListener("mousemove", handleDragInterface);
+    document.removeEventListener("pointerup", handleStopDrag);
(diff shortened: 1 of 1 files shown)


addons/debugger/userscript.js
```diff
@@ -162,7 +162,8 @@ export default async function ({ addon, console, msg }) {
   window.addEventListener("resize", () => {
     moveInterface(lastX, lastY);
   });
-  interfaceHeader.addEventListener("mousedown", handleStartDrag);
+  interfaceHeader.addEventListener("pointerdown", handleStartDrag);
+  interfaceHeader.addEventListener("touchstart", (e) => e.preventDefault());
```
Joeclinton1: e.preventDefault() is too broad here. Will prevent user being able to click the header buttons.
```suggestion
  interfaceHeader.addEventListener("touchstart", (e) => {
    // Only prevent default if not touching a button/interactive element
    if (!e.target.closest('button, [role="button"], .sa-debugger-tabs *, .sa-debugger-header-buttons *')) {
      e.preventDefault();
    }
  });
```

Joeclinton1 (changes requested):

addons/debugger/userscript.js
DNin01: Whoops!

Changing this to a `"touchmove"` listener also fixes this.

DNin01:
I changed `"touchstart"` to `"touchmove"`. Is that enough to fix it?

Joeclinton1:
> I changed `"touchstart"` to `"touchmove"`. Is that enough to fix it?

So now if you move the debugger whilst clicking a tab, it won't click when you let go. Whilst on desktop it does. Does that matter maybe not, but I'd still prefer the solution I gave even if it's more complex.

Joeclinton1 (approved):
The touchmove solution will allow you to click the buttons, but won't let you click the buttons whilst dragging. Since this behaviour might be unintended anyway on touchscreens, I'm going to approve it. 

The functionality of adding back the ability to drag works.

DNin01:
> > I changed `"touchstart"` to `"touchmove"`. Is that enough to fix it?
> 
> So now if you move the debugger whilst clicking a tab, it won't click when you let go. Whilst on desktop it does. Does that matter maybe not, but I'd still prefer the solution I gave even if it's more complex.

I would expect that dragging would not click the tab on desktop or mobile, only move the window.

With your solution, dragging from a tab does not move the window, which means you must grab it by an empty space on the title bar.

Samq64 (approved):
Tested on Android

PR #8561, opened 2025-09-14
New addon: `flag-menu`

### LINKED ISSUE #7230: Context menu for green flag (to toggle custom framerate, mute, etc.)
### Why this would be helpful

Two of our addons, `60fps` and `mute-project`, can be toggled by clicking the green flag while holding a certain key. That's not possible for users on touch screen devices.

### My suggestion

To fix this, we could create a custom context menu for the green flag. Users on mobile devices could long press the green flag to get a menu including all the toggleable runtime options that would normally be accessed via a keyboard shortcut.

This context menu would be available when addons that add green flag shortcuts are enabled (or maybe all the time).

Or (I think I like this better) we could make an addon that adds this context menu. Then someone who doesn't have any other addons enabled on a touch screen device can find this feature in the settings page, enable it, and then toggle turbo mode from the project page without a keyboard. And we could also reference the context menu addon from other addons that add green flag shortcuts using a "related addons" section (#5937) or by mentioning it with a note.

The context menu addon would probably be responsible for showing the different options based on which addons are enabled, or if we don't make this an addon, the context menu would act as a shared space for other addons to add items to by using the `addon.tab` API.

### Possible alternatives

Add these options to the Edit menu in the editor's menu bar.

### Additional context

Context menu on green flag with options "turn on turbo mode", "switch to 60 fps", and "mute project audio"

DNin01:
Resolves #7230

### Changes

- Adds a new addon which adds a context menu to the green flag. It lets you toggle turbo mode, higher framerate, and mute. Now mobile users can finally use turbo mode without having to enter the editor!
- Related addons have been updated to utilize their context menu items.
- Also fixed an issue that prevents the displayNoneWhileDisabled API from working on addons beginning with a number.

It looks like this:
Green flag context menu

*Context menu items (other than toggle turbo mode) only appear if their respective addons are enabled.

### Tests

Tested on Edge 140 with a mouse and with DevTools Device Emulation.

To do:
- [ ] Use `touchstart` instead of `contextmenu`
- [x] Localize
- [x] Dynamic disable, enable, and setting changes work for all updated addons
- [x] `editor-dark-mode` compatibility
- [x] `editor-compact` compatibility

### DIFF
--- a/addons-l10n/en/flag-menu.json
+++ b/addons-l10n/en/flag-menu.json
@@ -1,5 +1,6 @@
 {
-  "flag-menu/toggle-turbo": "toggle turbo mode",
+  "flag-menu/turbo-on": "turn on turbo mode",
+  "flag-menu/turbo-off": "turn off turbo mode",
   "flag-menu/set-fps": "switch to {fps} fps",
   "flag-menu/mute": "mute project",
   "flag-menu/unmute": "unmute project"
--- a/addons-l10n/en/flag-menu.json
+++ b/addons-l10n/en/flag-menu.json
@@ -1,0 +1,5 @@
+{
+  "flag-menu/toggle-turbo": "toggle turbo mode",
+  "flag-menu/set-fps": "switch to {fps} fps",
+  "flag-menu/toggle-mute": "toggle mute"
+}
--- a/addons-l10n/en/flag-menu.json
+++ b/addons-l10n/en/flag-menu.json
@@ -1,5 +1,6 @@
 {
   "flag-menu/toggle-turbo": "toggle turbo mode",
   "flag-menu/set-fps": "switch to {fps} fps",
-  "flag-menu/toggle-mute": "toggle mute"
+  "flag-menu/mute": "mute project",
+  "flag-menu/unmute": "unmute project"
 }
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -42,6 +42,7 @@
     }
   ],
   "tags": ["player", "featured"],
+  "relatedAddons": ["flag-menu"],
   "versionAdded": "1.1.0",
   "enabledByDefault": false,
   "libraries": ["scratch-gui"]
--- a/addons/flag-menu/context_menu.css
+++ b/addons/flag-menu/context_menu.css
@@ -4,39 +4,12 @@
   display: none;
 }
 
-/* Styles reused from https://github.com/scratchfoundation/scratch-gui/blob/develop/src/components/context-menu/context-menu.css */
-
-.flag-context-menu {
-  min-width: 130px;
-  padding: 5px 0;
-  margin: 2px 0 0;
-  font-size: 0.85rem;
-  text-align: left;
-  background-color: var(--editorDarkMode-input, white);
-  border: 1px solid var(--editorDarkMode-border, #00000026);
-  border-radius: calc(0.5rem / 2);
-  box-shadow: 0px 0px 5px 1px #00000026;
-  pointer-events: none;
-  transition: opacity 0.2s ease;
-  z-index: 10000;
+.sa-flag-context-menu {
   position: fixed;
   pointer-events: none;
 }
 
-.flag-context-menu.ctx-menu-open {
+.sa-flag-context-menu.sa-flag-menu-open {
   opacity: 1 !important;
   pointer-events: auto;
-}
-
-.flag-menu-item {
-  padding: 8px 12px;
-  color: var(--editorDarkMode-input-text);
-  white-space: nowrap;
-  cursor: pointer;
-  transition: 0.1s ease;
-}
-
-.flag-menu-item:hover {
-  background-color: var(--editorDarkMode-primary, #855cd6);
-  color: var(--editorDarkMode-primary-text, white);
 }
--- a/addons/flag-menu/context_menu.css
+++ b/addons/flag-menu/context_menu.css
@@ -1,0 +1,40 @@
+/* Hide context menu items related to other addons by default */
+#sa-flag-menu-fps, #sa-flag-menu-mute {
+  display: none;
+}
+
+/* Styles reused from https://github.com/scratchfoundation/scratch-gui/blob/develop/src/components/context-menu/context-menu.css */
+
+.flag-context-menu {
+  min-width: 130px;
+  padding: 5px 0;
+  margin: 2px 0 0;
+  font-size: 0.85rem;
+  text-align: left;
+  background-color: white;
+  border: 1px solid #00000026;
+  border-radius: calc(0.5rem / 2);
+  box-shadow: 0px 0px 5px 1px #00000026;
+  pointer-events: none;
+  transition: opacity 0.2s ease;
+  z-index: 10000;
+  position: fixed;
+  pointer-events: none;
+}
+
+.flag-context-menu.ctx-menu-open {
+  opacity: 1 !important;
+  pointer-events: auto;
+}
+
+.flag-menu-item {
+  padding: 8px 12px;
+  white-space: nowrap;
+  cursor: pointer;
+  transition: 0.1s ease;
+}
+
+.flag-menu-item:hover {
+  background: #855cd6;
+  color: white;
+}
--- a/addons/flag-menu/context_menu.css
+++ b/addons/flag-menu/context_menu.css
@@ -1,5 +1,6 @@
 /* Hide context menu items related to other addons by default */
-#sa-flag-menu-fps, #sa-flag-menu-mute {
+#sa-flag-menu-fps,
+#sa-flag-menu-mute {
   display: none;
 }
 
--- a/addons/flag-menu/context_menu.css
+++ b/addons/flag-menu/context_menu.css
@@ -6,10 +6,8 @@
 
 .sa-flag-context-menu {
   position: fixed;
-  pointer-events: none;
 }
 
 .sa-flag-context-menu.sa-flag-menu-open {
-  opacity: 1 !important;
-  pointer-events: auto;
+  visibility: visible !important;
 }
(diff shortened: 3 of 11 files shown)


addons/60fps/addon.json
mxmou, line 1: A notice explaining how to access 60 FPS mode on a touch screen would be helpful. Same with mute-project.

addons-l10n/en/flag-menu.json
```diff
@@ -0,0 +1,5 @@
+{
+  "flag-menu/toggle-turbo": "toggle turbo mode",
```
mxmou: Separate messages for turning it on and off would be more consistent with Scratch's UI and easier to translate into other languages. Similarly, "mute" and "unmute" would be better than "toggle mute".

addons/flag-menu/context_menu.css
```diff
@@ -0,0 +1,42 @@
+/* Hide context menu items related to other addons by default */
+#sa-flag-menu-fps,
+#sa-flag-menu-mute {
+  display: none;
+}
+
+/* Styles reused from https://github.com/scratchfoundation/scratch-gui/blob/develop/src/components/context-menu/context-menu.css */
```
mxmou: These styles will soon be outdated because Scratch's [React 18 PR](https://github.com/scratchfoundation/scratch-editor/pull/299) changes context menus. Can you use `addon.tab.scratchClass` so that the addon doesn't need to duplicate Scratch CSS?

mxmou (changes requested):
This addon should be enabled by default - otherwise touchscreen users can't use 60fps and mute-project at all without enabling it first.

After the React 18 update, it will be possible to navigate Scratch's context menus using arrow keys. It would be nice if the addon replicated that.

addons-l10n/en/flag-menu.json
DNin01: That was my original plan—I will add it later.

addons/flag-menu/context_menu.css
DNin01: Alright. It seems to work fine.

DNin01:
> This addon should be enabled by default.

Perhaps enabled by default only for mobile versions of the extension?

That's too much work, though. I'll just make it enabled by default.

> After the React 18 update, it will be possible to navigate Scratch's context menus using arrow keys. It would be nice if the addon replicated that.

Ideally, I would make the context menu an instance of a React context menu so that React can take care of this. Not sure how, though. May just have to stick with custom handlers.

We don't use React within our project, do we? (We probably should.)

addons/60fps/addon.json
DNin01: That addon already has so much notice text. 🤣

DNin01:
Okay, I got all those kinks worked out!

mxmou:
> Ideally, I would make the context menu an instance of a React context menu so that React can take care of this. Not sure how, though. May just have to stick with custom handlers.

Writing the event handlers is probably easier, especially because we currently need to support both versions of React. [That's what I did when updating the context menu API.](https://github.com/ScratchAddons/ScratchAddons/blob/master/addon-api/content-script/contextmenu.js#L19-L53) After the React 18 update is out, we can experiment with using Scratch's context menu component.

Speaking of React 18: if you don't mind, I'll push some changes to make the addon's styling work there.

DNin01:
> Speaking of React 18: if you don't mind, I'll push some changes to make the addon's styling work there.

Sure, you know more about it than I do.

addons/60fps/addon.json
mxmou: Maybe add this information to the description instead?

mxmou: Platform-specific descriptions, which I suggested in https://github.com/ScratchAddons/ScratchAddons/issues/3930#issuecomment-3263716029, would also be useful here.

WorldLanguages: If you want to add `"touchscreenOnly": true"` to notices within this PR it's fine to me.

addons/flag-menu/userscript.js
```diff
@@ -0,0 +1,83 @@
+export default async function ({ addon, console, msg }) {
+  let greenFlag;
+
+  const contextMenuClass =
+    addon.tab.scratchClass("context-menu_context-menu") || addon.tab.scratchClass("context-menu_context-menu-content"); // React 16 || React 18
```
DNin01: Scratch updated React to 18 with the Face Sensing update, so this can be simplified. The style of the context menu was also changed, so some things may need to be tweaked.

DNin01: I mean, Scratch's new context menu has no transition for fading in/out, it just appears immediately, plus there's arrow-key navigation and the width was reduced. Every other property applies automatically.

mxmou: > Scratch updated React to 18 with the Face Sensing update, so this can be simplified.

Done. I've also removed the opacity transition.

DNin01:
> After the React 18 update, it will be possible to navigate Scratch's context menus using arrow keys. It would be nice if the addon replicated that.

I don't know if Scratch uses this anymore, but [`react-contextmenu`](https://www.npmjs.com/package/react-contextmenu) is in the `package.json`. If we used that library...

I know how to program arrow-key navigation myself, though.

DNin01:
<details><summary>Still working on an implementation that skips over invisible menu items, but here's what I have now. I think it's alright.</summary>

```diff
  export default async function ({ addon, console, msg }) {
    let greenFlag;
+   let contextMenuOpen = false;

    const contextMenu = Object.assign(document.createElement("nav"), {
      role: "menu",
      className: addon.tab.scratchClass("context-menu_context-menu-content", { others: "sa-flag-context-menu" }),
    });
+   const setHighlightedItem = (menuitem) => {
+     const currentItem = contextMenu.querySelector("[data-highlighted]");
+     if (currentItem) {
+       currentItem.removeAttribute("data-highlighted");
+       currentItem.tabIndex = -1;
+     }
+     if (menuitem) {
+       menuitem.setAttribute("data-highlighted", "");
+       menuitem.tabIndex = 0;
+       menuitem.focus();
+     }
+   }
    const createItem = (id, textContent = "") => {
      const item = Object.assign(document.createElement("div"), {
        role: "menuitem",
        className: addon.tab.scratchClass("context-menu_menu-item"),
        id,
        textContent,
      });
-     item.addEventListener("mouseenter", () => item.setAttribute("data-highlighted", ""));
-     item.addEventListener("mouseleave", () => item.removeAttribute("data-highlighted"));
+     item.addEventListener("mouseenter", () => setHighlightedItem(item));
+     item.addEventListener("mouseleave", () => setHighlightedItem());
      return item;
    };
    contextMenu.append(
      createItem("sa-flag-menu-turbo", msg("turbo-on")),
      createItem("sa-flag-menu-fps"),
      createItem("sa-flag-menu-mute")
    );

    function closeContextMenu() {
+     contextMenuOpen = false;
+     setHighlightedItem();
      contextMenu.classList.remove("sa-flag-menu-open");
    }
    document.addEventListener("mousedown", (e) => {
      if (!e.target.closest("[role='menu']")) closeContextMenu();
    });
    document.addEventListener("keydown", (e) => {
-     if (e.key === "Escape") closeContextMenu();
+     if (!contextMenuOpen) return;
+     switch (e.key) {
+       case "Escape":
+       case "Tab":
+         closeContextMenu();
+       break;
+       case "Enter":
+       case " ":
+         if (contextMenu.querySelector("[data-highlighted]")) contextMenu.querySelector("[data-highlighted]").click();
+       break;
+       case "ArrowUp":
+         const previousItem = contextMenu.querySelector("[data-highlighted]")?.previousElementSibling ?? contextMenu.lastElementChild;
+         setHighlightedItem(previousItem, true);
+       break;
+       case "ArrowDown":
+         const nextItem = contextMenu.querySelector("[data-highlighted]")?.nextElementSibling ?? contextMenu.firstElementChild;
+         setHighlightedItem(nextItem, true);
+       break;
+       case "Home":
+       case "PageUp":
+         setHighlightedItem(contextMenu.firstElementChild, true);
+       break;
+       case "End":
+       case "PageDown":
+         setHighlightedItem(contextMenu.lastElementChild, true);
+       break;
+     }
    });

    contextMenu.querySelector("#sa-flag-menu-turbo").addEventListener("click", () => {
      closeContextMenu();
      greenFlag.dispatchEvent(new MouseEvent("click", { bubbles: true, shiftKey: true }));
    });
    contextMenu.querySelector("#sa-flag-menu-fps").addEventListener("click", () => {
      closeContextMenu();
      greenFlag.dispatchEvent(new MouseEvent("click", { bubbles: true, altKey: true }));
    });
    contextMenu.querySelector("#sa-flag-menu-mute").addEventListener("click", () => {
      closeContextMenu();
      greenFlag.dispatchEvent(new MouseEvent("click", { bubbles: true, ctrlKey: true }));
    });

    contextMenu.style.visibility = "hidden"; // Setting visibility here fixes a visual glitch on dynamic enable
    addon.tab.displayNoneWhileDisabled(contextMenu);
    addon.self.addEventListener("disabled", closeContextMenu);

    addon.tab.traps.vm.on("TURBO_MODE_ON", () => {
      contextMenu.querySelector("#sa-flag-menu-turbo").textContent = msg("turbo-off");
    });
    addon.tab.traps.vm.on("TURBO_MODE_OFF", () => {
      contextMenu.querySelector("#sa-flag-menu-turbo").textContent = msg("turbo-on");
    });

    while (true) {
      greenFlag = await addon.tab.waitForElement("[class^='green-flag_green-flag']", {
        markAsSeen: true,
        reduxEvents: ["scratch-gui/mode/SET_PLAYER", "fontsLoaded/SET_FONTS_LOADED", "scratch-gui/locales/SELECT_LOCALE"],
      });

      greenFlag.parentElement.appendChild(contextMenu);
      greenFlag.addEventListener("contextmenu", (e) => {
        if (addon.self.disabled) return;
        e.preventDefault();
+       contextMenuOpen = true;
        contextMenu.classList.add("sa-flag-menu-open");
        contextMenu.style.left = e.clientX + "px";
        contextMenu.style.top = e.clientY + "px";
      });
    }
  }
```

</details>

PR #8562, opened 2025-09-14
Paint addon settings buttons: use arrow icons

### LINKED ISSUE #8554: Use dropdown icons on costume editor addon buttons
These buttons, which currently look like this:
Costume editor tools with icon popup buttons

...can be represented better when they are dropdowns:
Costume editor tools with dropdown buttons

Compare these two designs:
- When both the toggle and the popup buttons have their own icons, they look like completely different tools.
- When the toggle buttons have icons and the popup buttons are dropdowns, it's clear that they are tied to their corresponding toggle button.

DNin01:
Resolves #8554

### Changes

Changes the dedicated icons for the `onion-skinning` and `paint-snap` settings buttons to arrows.

### Reason for changes

When both the toggle and the popup buttons have their own icons, they look like completely different tools, but when the popup buttons are dropdowns, it's clear that they are tied to their neighboring button.

https://github.com/user-attachments/assets/79008e88-071b-4b80-90d9-c8d7cb441fb0


addons/onion-skinning/style.css
```diff
@@ -12,6 +12,21 @@
   background-color: #855cd6;
 }
 
+/* Dropdown arrow */
+[class^="paint-editor_zoom-controls_"] ~ div > .sa-onion-button + .sa-onion-button .sa-onion-image {
```
DNin01: Using a selector like this is pretty hacky and could break if future changes are made. It would be ideal to change the code to allow dropdown-arrow styles to be implemented in a potentially simpler way.

Samq64:
`collapse-footer` and `sprite-properties` already use that icon pointing upwards so could it be moved to `/images/cs` and shared across all 4 addons?

DNin01:
> `collapse-footer` and `sprite-properties` already use that icon pointing upwards so could it be moved to `/images/cs` and shared across all 4 addons?

The arrow icon used in this pull request is 8x5 whereas the other two are both 12.7x8.79, so this would require a few alterations.

addons/onion-skinning/style.css
DNin01: Fixed.

addons/onion-skinning/style.css
```diff
@@ -12,6 +12,16 @@
   background-color: #855cd6;
 }
 
+.sa-onion-arrow .sa-onion-image {
+  width: 8px;
+  height: auto;
+  margin-top: 5px;
```
mxmou: The arrow doesn't look centered to me. A better way to center it might be
```css
.sa-onion-arrow {
  display: flex;
  align-items: center;
}
```

addons/onion-skinning/style.css
```diff
@@ -12,6 +12,16 @@
...
+}
+.sa-onion-button[data-enabled="false"] + .sa-onion-arrow:not([data-enabled="true"]) {
+  opacity: 0.33;
```
mxmou:
```suggestion
  opacity: 0.5;
```
Scratch's own buttons, such as undo and redo, have an opacity of 0.5 when disabled.

addons/onion-skinning/style.css
DNin01: Wow, it's that easy?

Samq64 (approved):

DNin01:
I should probably mention that this PR changes one thing about how you interact with the buttons:

Before, you could open the settings for onion skinning and snapping without first activating them, and changing a setting would turn them on for you. Now, you have to turn them on first as the dropdown buttons are grayed out when they're inactive.

Is that fine?

DNin01:
I think it's fine because the settings don't matter when the tools are turned off.

Issue #5249, opened 2022-10-23
Adding more utility CSS clases similar to existing `.sa-body-editor`


WorldLanguages:
### Why this would be helpful/solve a problem

For months, `content-scripts/inject/module.js` has been adding a class named `sa-body-editor` to the `<body/>` element of the page when entering the editor.
This class is used by CSS of many addons, and removes the need for the userscripts of those addons to handle a different class name each.
We could extend this idea and create other classes which are handled by core.

### What this change does

Some ideas:

- @Samq64 wanted to have a class that worked just like `sa-body-editor`, but only applied when editorMode is "editor", excluding "fullscreen"
- Many addons, such as mouse-pos, clones and now vol-slider have their own ways to detect whether project page fullscreen or project editor fullscreen is opened. Aditionally, each addon has its own code to hide in small stage mode.
- There could be an equivalent to `addon.tab.clientVersion` which would be accessible through 2 different classes, such as `.sa-body-scratchwww` and `sa-body-scratchr2` (cc @mxmou on whether this would be useful)

### Possible alternatives

Using CSS variables, but those are hard to use in order to conditionally apply styles.

mxmou:
Could these classes be set on the `documentElement`, not the body? It's sometimes necessary to set CSS properties on the root element.

> There could be an equivalent to `addon.tab.clientVersion` which would be accessible through 2 different classes, such as `.sa-body-scratchwww` and `sa-body-scratchr2` (cc @mxmou on whether this would be useful)

Probably not - these can already be differentiated using match patterns.

WorldLanguages:
> Could these classes be set on the `documentElement`, not the body?

Yes

> Probably not - these can already be differentiated using match patterns.

Oh, right 👍

WorldLanguages:
> Many addons, such as mouse-pos, clones and now vol-slider have their own ways to detect whether project page fullscreen or project editor fullscreen is opened. Aditionally, each addon has its own code to hide in small stage mode.

Consideration: editors such as TurboWarp would need to implement `.sa-body-smallstagemode` themselves. So I think it would be better to make this a module which is imported by the 3 addons, instead of a feature handled by non-addon code.

WorldLanguages:
> So I think it would be better to make this a module which is imported by the 3 addons, instead of a feature handled by non-addon code.

[...]

PR #8570, opened 2025-09-17
Modes Easter egg addon & `.sa-editor` class

### LINKED ISSUE #5129: An addon to bring back the "mode" heading tab in the editor from April Fools 22

Originally discussed in #5123

It would be nice to bring back the "mode" heading tab from April Fools 22. This includes Cat Blocks (Caturday mode) in addition to Old timey mode, Prehistoric mode, etc. It can be turned off by simply selecting the normal mode.

### What this addon does

This addon brings back the "mode" heading tab from April Fools 22.

### Possible alternatives

_No response_

Mode

mxmou:
Resolves #5129

### Changes

I started working on this in May 2022, then abandoned it, then forgot why I abandoned it. 3 years later, I've decided to finally finish it and make a PR.

* Adds the modes from April Fools 2020 (Mystery) and 2022 (90s, Old timey, Prehistoric). The mode can be selected in addon settings. Caturday mode isn't included because there's already a cat blocks addon.
* Adds a `.sa-editor` class, which is similar to `.sa-body-editor` but for the html element. The modes addon uses this to apply filters to the root element.

### Reason for changes

The modes are just for fun, but `.sa-editor` will likely be useful for other addons in the future.

### Tests

Tested on Edge and Firefox.


Samq64:
All the modes also effected fullscreen druing April Fools, right? Also this addon is 1.5MB which shouldn't be a problem but is something to keep in mind.

CST1229:
> All the modes also effected fullscreen druing April Fools, right? Also this addon is 1.5MB which shouldn't be a problem but is something to keep in mind.

over half of that size is from ~~projector.png~~ oldtimey.png alone. (i think the images could maybe be compressed to a different format. maybe projector.png could even be turned into a video? maybe, if possible?)

Samq64:
> i think the images could maybe be compressed to a different format.

I think mystery and prehistoric PNGs would have noticiable banding and jpeg doesn't support transparency.

mxmou:
> All the modes also effected fullscreen druing April Fools, right?

They didn't, but I think it makes more sense if they do.

mxmou:
> (i think the images could maybe be compressed to a different format. maybe projector.png could even be turned into a video? maybe, if possible?)

I've already reduced the size of the images - but, as @Samq64 pointed out, mystery.png and prehistoric.png can't be compressed too much without a visible quality loss. I'll try to do something with oldtimey.png.

content-scripts/inject/module.js
```diff
@@ -212,16 +212,25 @@ function onDataReady() {
...
   const pathname = location.pathname.toLowerCase();
   const split = pathname.split("/").filter(Boolean);
   if (!split[0] || split[0] !== "projects") return;
-  if (split.includes("editor") || split.includes("fullscreen")) document.body.classList.add("sa-body-editor");
-  else document.body.classList.remove("sa-body-editor");
+  if (split.includes("editor") || split.includes("fullscreen")) {
+    document.documentElement.classList.add("sa-editor");
+    document.body.classList.add("sa-body-editor");
+  } else {
+    document.documentElement.classList.add("sa-editor");
```
Samq64:
```suggestion
    document.documentElement.classList.remove("sa-editor");
```

Samq64:
> > All the modes also effected fullscreen druing April Fools, right?
> 
> They didn't, but I think it makes more sense if they do.

I disagree since it also applies to opening fullscreen from the project page and the 90s mode only changes the hue there which could be confusing.

mxmou:
> I disagree since it also applies to opening fullscreen from the project page and the 90s mode only changes the hue there which could be confusing.

I've tested Scratch's totally-normal-2022 branch again. 90s mode actually did the exact same thing in full screen as the addon does: the filter is applied but the background pattern isn't. This could be fixed by doing nothing at all in fullscreen or adding the pattern - the latter is easier.

addons/totally-normal-modes/assets/oldtimey.png
Samq64, line 1: I was able to get the original 1MB file down to 100KB by converting it to AVIF with [ezgif](https://ezgif.com/apng-to-avif)

addons/totally-normal-modes/addon.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "Editor modes",
+  "description": "Adds the 90s, Old timey and Prehistoric mode from April Fools 2022 and the Mystery Mode from 2020.",
```
Samq64:
```suggestion
  "description": "Adds the 90s, Old timey and Prehistoric modes from April Fools 2022 and the Mystery Mode from 2020.",
```

addons/totally-normal-modes/addon.json
```diff
@@ -0,0 +1,82 @@
...
+        },
+        {
+          "id": "mystery",
+          "name": "Mystery"
+        }
+      ],
+      "default": "90s"
+    }
+  ],
+  "tags": ["editor", "easterEgg"],
```
Samq64, line 87:
```suggestion
  "tags": ["editor", "easterEgg"],
  "relatedAddons": ["cat-blocks"],
```
This should only be a one-way relation since cat blocks is also an editor theme. Although it's not really needed at all since the Easter eggs category only has a few addons.

addons/totally-normal-modes/assets/oldtimey.png
mxmou: I'll do that. If someone has an older browser, they might not see the animation, but I think that's fine.

Samq64 (approved):
Above 1080p the overlays that follow the mouse aren't big enough. I don't think it's worth fixing if it was like that druing April Fools but the images could be bumped to 7680x4320 to support full screen 4K.

The assests are still a little large but it's much better than before. The projector sound could be shortened but then there would be audio cuts more often.

DNin01:
> Above 1080p the overlays that follow the mouse aren't big enough. I don't think it's worth fixing if it was like that druing April Fools but the images could be bumped to 7680x4320 to support full screen 4K.

There's probably another way to black out parts of the screen, without using images with large black masks.

WorldLanguages:
Resolves https://github.com/ScratchAddons/ScratchAddons/issues/5249#issuecomment-1288915309

mxmou:
> Above 1080p the overlays that follow the mouse aren't big enough.

I've fixed that by using `radial-gradient`s instead of image files. This also saves some space because `prehistoric.png` was quite big.

Regarding full screen, I think you were right - it's weird that this addon changes the project page full screen mode. I'll change that.

CST1229:
didn't Mystery mode also add a shadow effect on sprites based on the mouse position (which isn't in this addon)?

addons/totally-normal-modes/userscript.js
```diff
@@ -0,0 +1,82 @@
...
+    oldTimey.updateSound(addon, mode);
+    mystery.update(addon, mode);
+    updatePointerMoveListener();
+  };
+  addon.self.addEventListener("disabled", onSettingChange);
+  addon.self.addEventListener("reenabled", onSettingChange);
+  addon.settings.addEventListener("change", onSettingChange);
+  addon.tab.addEventListener("urlChange", () => {
+    oldTimey.updateSound(addon, mode);
+    mystery.update(addon, mode);
```
Samq64, line 70: This is in a `urlChange` listener so it does not update on `/projects/editor`

addons/totally-normal-modes/userscript.js
```diff
@@ -0,0 +1,82 @@
...
+    updatePointerMoveListener();
+  });
```
Samq64: This doesn't update properly on `/projects/editor` because it's a `urlChange`.

Samq64 (approved):
Should there be a note for the sound not starting until interacted with?

Also, in the future it could be useful to be able o listen to `editorMode` change events.

mxmou:
> Should there be a note for the sound not starting until interacted with?

I don't think so - I imagine most users interact with the editor quite quickly after opening it.

> Also, in the future it could be useful to be able to listen to `editorMode` change events.

I agree. Many editor addons that currently listen for `urlChange` events are actually trying to detect when the `editorMode` changes.

addons/totally-normal-modes/assets/oldtimey-projector.mp3
Samq64, line 1: Very minor but could this be renamed to `projector.mp3` or `oldtimey.mp3`

addons/totally-normal-modes/assets/90s-logo.svg
Samq64, line 1: And this one to `90s-logo.svg` for consistency?

addons/totally-normal-modes/assets/oldtimey-projector.mp3
mxmou: I'll rename them. The current names are the original ones from scratch-gui.

mxmou:
> The modes still exist in Scratch's source code but it's probably broken now and this implementation is better anyway.

The components are still there, but we can't access them because they're unreferenced - see https://github.com/scratchfoundation/scratch-gui/commit/43e0cb79c77cde0a96778c98ec899ceb70e26658.

content-scripts/inject/module.js
```diff
@@ -212,16 +212,29 @@ function onDataReady() {
   }
 }
 
-function bodyIsEditorClassCheck() {
-  if (isScratchGui) return document.body.classList.add("sa-body-editor");
-  const pathname = location.pathname.toLowerCase();
-  const split = pathname.split("/").filter(Boolean);
-  if (!split[0] || split[0] !== "projects") return;
```
DNin01, line 219: I don't think this check should have been removed. See #8653.

WorldLanguages:
@mxmou Should we actually add this? Github activity from the Scratch Team may suggest they want to add this as part of their monthly membership. Well, that's what I remember reading a month ago or so - I don't have any link or proof.

Samq64:
Was it specificly about these editor modes or just cat blocks which was also in the modes menu?

mxmou:
> @mxmou Should we actually add this? Github activity from the Scratch Team may suggest they want to add this as part of their monthly membership. Well, that's what I remember reading a month ago or so - I don't have any link or proof.

They've mentioned "themes for the code editor" for members. I don't think there's currently any information about what those themes will be (other than cat blocks), but I'd be surprised if they included these modes - unlike cat blocks, the modes just make the editor harder to use.

PR #8571, opened 2025-09-18
Display easter eggs cetegory name when open

Samq64:
When displaying easter egss the settings page now shows the category name in the header and when no search results are found.

### Tests

Tested on Chromium


mxmou (approved):
This might be a good opportunity to fix the horizontal alignment of the category name and the no results message:

Image

The horizontal margin of the message should be 20px, not 10px.

PR #8572, opened 2025-09-19
New addon: `blocks2text`

LuYifei2011:
### Changes

This pull request adds a new addon: Copy Blocks as Text.
It can copy blocks as scratchblocks text and paste it into forums or other places.

https://github.com/user-attachments/assets/5c49b9e9-f35d-43fe-ad6f-32c873afe6b4

### Reason for changes

It lets users copy scratchblocks text directly instead of writing it manually.

### Tests

Tested on Chrome 140.

> This description was translated by Google Translate and therefore may be somewhat inaccurate.
> Localization translated by GPT-5.
> I'm a newbie so please let me know if I'm doing something wrong.


Samq64:
This is the 3rd open pull request for this feature but they all take different approaches. This one uses the visible text, #7988 uses the block types (e.g. `motion_movesteps`) and #7748 uses the `parse-sb3-blocks` library.

addons/addons.json
```diff
@@ -189,5 +189,6 @@
   "remove-curved-stage-border",
   "columns",
   "editor-compact",
-  "editor-theme3"
+  "editor-theme3",
+  "blocks2text"
```
Samq64: Could this be moved above `// NEW ADDONS ABOVE THIS ↑↑`?

addons/blocks2text/addon.json
```diff
@@ -0,0 +1,18 @@
...
+      "matches": ["projects"]
+    }
+  ],
+  "credits": [{ "name": "LuYifei2011", "link": "https://github.com/LuYifei2011" }],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.44.0",
+  "enabledByDefault": false,
+  "permissions": ["clipboardWrite"],
+  "libraries": ["scratch-gui"]
```
Samq64:
```suggestion
  "versionAdded": "1.44.0"
```
`clipboardWrite` is only needed to copy images on older Firefox versions and the `scratch-gui` licence is for addons that modify its functions.

addons/blocks2text/userscript.js
```diff
@@ -0,0 +1,213 @@
...
+    // Add backslash escape before parentheses
+    return text.replace(/([()[\]{}])/g, "\\$1");
+  }
+
+  function serializeBlock(block) {
+    if (!block) {
+      return "";
+    }
+
+    if (block.type === "text") {
```
Samq64, line 20: Colour inputs should also be in square brackets.

Samq64:
Also custom block uses should have their categroy explicitly set since the definitions might not exist in the copied text.

addons/blocks2text/userscript.js
LuYifei2011: Sorry. I'm not entirely sure what your final intention was. The current code seems fine, but happy to revisit if there's more context you'd like to share.

addons/blocks2text/userscript.js
```diff
@@ -0,0 +1,215 @@
...
+
+  const imageNames = {
+    "green-flag": "@greenFlag",
+    "rotate-left": "@turnLeft",
+    "rotate-right": "@turnRight",
+  };
+
+  function escapeText(text) {
+    // Add backslash escape before parentheses
+    return text.replace(/([()[\]{}])/g, "\\$1");
```
Samq64:
```suggestion
    return text.replace(/([()[\]{}<>])/g, "\\$1");
```

addons/blocks2text/userscript.js
Samq64: It seems to have always used parentheses which is also fine so I'm not sure why the hex value wasn't surrounded with them before.

Joeclinton1:
The main issue I see with this, which #7748 solves (#7988 does also, but lacks multi language support as a consequence), is how to force English translations so that it can be compatible with the forums, which require English blocks?

this will require either mappings (in which case you should just use #7748), or changing editor language, rerendering, then converting back)

WorldLanguages:
For the record, discussion has moved to #2336

PR #8577, opened 2025-09-21
Settings page keyboard navigation

### LINKED ISSUE #2532: Addon settings page can not be used properly with a keyboard or other accessibility tools

Samq64:
Resolves #2532

### Changes

- The settings page and popup are now usable with the keyboard
- Converts the more settings modal into an HTML `<dialog>` element
- Consolidates the switch styles into its own file
- Consolidates `addon-group`, `btn-dropdown` and `back-button` into a single class
- Adds a dropdown compnent with arrow key navigation
- ~~Always shows the dismiss admin message link (regular comments are unaffected)~~

### Reason for changes

It's good for accessibility and not being able to toggle addons with the keyboard is annoying.

### Tests

Tested on Chromium

Todo:
- [x] Dropdows
- [x] Scratch messaging
- [x] Cloud games (already good)
- [x] Fix enabling addons with confirmations
- [ ] Better focus borders
- [x] Re-order table rows with the keyboard? #8934
- [ ] Navigate between addons with the arrow keys?


images/icons/x.svg
Samq64, line 1: This SVG was also rotated 360 degrees so I removed that.

mxmou: The transform was probably meant to fix a rendering bug. You can remove it if it doesn't cause a visible difference in the latest versions of browsers that SA supports.

DNin01 (commented):
Did some testing with Windows Narrator—there's still room for improvement with screen reader navigation, but, with a keyboard, things are at least usable now.

Some screen reader problems:
- Many icon buttons don't have a name
- Addon setting elements aren't linked to their labels
- Item lists like the category panel don't function like lists

Some keyboard navigation problems:
- Setting tooltips can't be accessed

DNin01:
I added keyboard navigation to dropdowns by simply setting `tabindex="0"` on them.

The button that opens the dropdown still needs to have an [`aria-haspopup` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-haspopup).

mxmou:
The correct keyboard behavior for dropdowns is that up and down arrows focus the previous/next item and Tab focuses the next element outside the dropdown.

DNin01:
> The correct keyboard behavior for dropdowns is that up and down arrows focus the previous/next item and Tab focuses the next element outside the dropdown.

I have before made a library with this exact functionality, so I'll look into this.

Samq64:
I have not done any screen reader testing myself yet.

> I have before made a library with this exact functionality, so I'll look into this.

I was hoping not to add another libary but it might be worth it if it needs more than a few lines of JS.

mxmou:
Blockly's menu implementation might be useful as a reference: https://github.com/google/blockly/blob/139fa2b/core/menu.ts#L407-L462

DNin01:
<details><summary>Maybe I should just tell you what I have in store...</summary>

```js
const shiftAmountsByKey = {
  ArrowUp: -1,
  ArrowDown: 1,
  ArrowLeft: -1,
  ArrowRight: 1,
  Home: -1000,
  End: 1000,
};

/** Makes a list-like element arrow-key navigated. */
export default class ArrowKeyList {
  constructor(element) {
    element.addEventListener("keydown", (e) => {
      if (e.ctrlKey || e.metaKey || e.altKey) return;
      if (e.key === "Tab") {
        if (e.shiftKey) {
          // Focus the first child, then the default behavior of Shift+Tab will move to the previous element
          element.firstElementChild.focus();
        } else {
          // Focus the last child, then the default behavior of Tab will move to the next element
          element.lastElementChild.focus();
        }
      } else {
        const shiftBy = shiftAmountsByKey[e.key];
        if (shiftBy) e.preventDefault(); else return;

        const oldFocusIndex = Array.from(element.children).indexOf(document.activeElement);
        // Move,
        const adjustedFocusIndex = oldFocusIndex + shiftBy;
        // clamp,
        const newFocusIndex = Math.min(Math.max(adjustedFocusIndex, 0), element.childElementCount - 1);
        // set focus!
        const targetElement = element.children[newFocusIndex];
        targetElement.focus();
      }
    });
  }
}
```

I'm not familiar with Vue at all, so I don't know how to work this code (which has been stripped down, by the way) into our project. (If you want to figure it out, feel free to use this snippet under the same license terms of our project.)

</details>

I think it's more important that we add ARIA attributes to our menus, though. `role`, `aria-haspopup`, and `aria-expanded` shouldn't be too hard to implement.

Or, I wonder if there's a library that handles ARIA attributes and navigation for us.

Samq64:
Menu arrow key naviagtion works now. I'm not sure if the menus should stay open when tabbing away, currently they doesn't.

> If you want to figure it out, feel free to use this snippet under the same license terms of our project.

Just to be clear you mean this repository's license, right?

DNin01:
It works! The dropdowns, menus, and menu items are announced as such:

> - "Button, collapsed, has popup"
>   - "Export settings, menu item, 1 of 2"
>   - "Import settings, menu item, 2 of 2"

However, icon-buttons still need readable labels—currently most of the images don't have alt text, and when there's no text, the default name is just "graphic" or "button".

> I'm not sure if the menus should stay open when tabbing away, currently they doesn't.

On GitHub, they collapse when you tab outside.

> > If you want to figure it out, feel free to use this snippet under the same license terms of our project.
> 
> Just to be clear you mean this repository's license, right?

Correct.

Samq64:
> However, icon-buttons still need readable labels—currently most of the images don't have alt text, and when there's no text, the default name is just "graphic" or "button".

I've added titles to more buttons but I say full screen reader support is out of the scope of this PR.

_locales/en/messages.json
```diff
@@ -152,6 +155,9 @@
   "resetToDefault": {
     "message": "Reset to default settings"
   },
+  "importExport": {
+    "message": "export/import settings"
```
mxmou:
```suggestion
    "message": "Export/import settings"
```

_locales/en/messages.json
```diff
@@ -176,6 +194,15 @@
   "preview": {
     "message": "Preview"
   },
+  "addRow": {
+    "message": "Add table row"
```
mxmou: I'm not sure if we should use the word table in the UI for table settings - they don't actually look like tables, so I doubt users think of them as such. I suggest using "item" instead of "table row".

_locales/en/messages.json
```diff
@@ -188,6 +215,9 @@
   "dark": {
     "message": "Dark"
   },
+  "toggleTheme": {
+    "message": "Toggle dark mode"
```
mxmou: It would be nice (and easier to translate) if there were separate "Switch to dark theme" and "Switch to light theme" messages. ("Theme", not "mode", for consistency with the text in More Settings.)

mxmou (commented):
Browsers render `<button>` text in a different font by default, which causes inconsistencies like this:

Image

A simple fix would be
```css
button {
  font: inherit;
}
```

_locales/en/messages.json
```diff
@@ -195,10 +195,10 @@
...
   "addRow": {
-    "message": "Add table row"
+    "message": "Add row"
   },
   "addPresetRow": {
-    "message": "Add preset table row"
+    "message": "Add preset row"
   },
   "deleteRow": {
     "message": "Delete table row"
```
mxmou:
```suggestion
    "message": "Delete row"
```

popups/scratch-messaging/popup.css
```diff
@@ -204,7 +206,7 @@ a.message-type-title-text {
 }
 .delete-btn {
   float: right;
-  opacity: 0;
+  visibility: hidden;
```
mxmou: What's the reason for this change? Is there any situation in which this makes the button unfocusable when it would otherwise be focusable?

popups/scratch-messaging/popup.css
```diff
@@ -226,14 +228,15 @@ a.message-type-title-text {
...
   color: var(--blue-text);
   font-weight: 500;
   user-select: none;
 }
-.comment:hover .reply-button-comment,
-.comment:hover .delete-btn {
-  opacity: 1;
+.comment:is(:hover, :focus-within) .reply-button-comment:not(.replying),
+.comment:is(:hover, :focus-within) .delete-btn,
+.message-type-admin .delete-btn {
```
mxmou: This line can be removed if the visibility change is reverted.

popups/scratch-messaging/popup.css
```diff
@@ -15,6 +15,8 @@ button {
   align-items: center;
   padding: 4px;
   border-radius: 4px;
+  background-color: transparent;
+  border: none;
```
mxmou, line 19:
```suggestion
  border: none;
  cursor: pointer;
```

popups/scratch-messaging/popup.html
```diff
@@ -103,14 +106,13 @@
             class="message-type-title hover-reverse"
             @click="messageTypeExtended.stMessages = !messageTypeExtended.stMessages"
           >
-            <div class="btn-dropdown">
+            <button class="btn-dropdown">
               <img
                 src="../../images/icons/expand.svg"
-                alt="v"
```
mxmou, line 109: This was removed here but not in the other instances of `.btn-dropdown`.

webpages/settings/components/addon-body.html
```diff
@@ -2,44 +2,34 @@
...
           </div><!-- no space --><addon-tag v-for="tag of addon.tags" :tag="tag"></addon-tag>
         </div>
       </div>
       <div class="addon-description" v-show="!expanded" dir="auto">{{ addon.description }}</div>
       <div class="addon-check">
-        <div
-          v-show="expanded && addon._enabled"
-          v-if="addon.settings"
-          class="addon-split-button dropdown-parent"
-          :class="{open: isDropdownOpen}"
```
mxmou, line 23: The open class was on this element, not on the dropdown itself, so that the button was highlighted when the dropdown was open (even if it wasn't hovered):

Image

The easiest way to restore this behavior is to change the relevant CSS selectors (`.addon-split-button.open`) to use `:has()`. They won't work in older browsers, but that's fine because it's just for aesthetics.

webpages/settings/components/addon-body.html
```diff
@@ -126,7 +116,7 @@
         <div class="addon-message">
           <span class="related-addons-text">{{msg('relatedAddons')}}</span>
           <span v-for="addonManifest of addon._relatedAddons">
-            <a @click="openRelated(addonManifest)">{{addonManifest.name}}</a>
+            <a href="#" @click="openRelated(addonManifest)">{{addonManifest.name}}</a>
```
mxmou: Navigating to `#` adds an entry to the navigation history, making it necessary to press the browser's back button twice to leave the page. The event handler should call `preventDefault()` to fix that.

webpages/settings/components/addon-body.html
```diff
@@ -184,7 +174,7 @@
   }
 
   .addon-icon {
-    margin-inline-start: 10px;
+    margin-inline-start: 2px;
```
mxmou:
```suggestion
    margin-inline-start: 1px;
```

On master, the spacing here is 9px:

Image

and here it's 10px:

Image

so the icon only has 1px of extra space.

webpages/settings/components/addon-body.js
```diff
@@ -140,9 +134,7 @@ export default async function ({ template }) {
             ? this.expanded
             : isIframe && !this.expanded && (this.addon.info || []).every((item) => item.type !== "warning")
               ? false
-              : event.shiftKey
-                ? false
-                : newState;
+              : newState;
```
mxmou: This removes the ability to enable addons without expanding them (#1484). I don't think the feature is very useful, but since it's so easy to implement I suggest keeping it and documenting it somewhere (at least in a code comment).

webpages/settings/components/addon-setting.html
```diff
@@ -38,41 +43,45 @@
...
-              <img class="icon-type" src="../../images/icons/expand.svg" draggable="false" />
-            </button>
-            <ul class="dropdown-list align-start">
-              <li v-for="preset of setting.presets" @click="addTableRow(preset.values)">{{ preset.name }}</li>
-            </ul>
-          </div>
+          <dropdown
+            button-class="addon-buttons addon-split-button-dropdown"
+            :is-disabled="!addon._enabled"
+            :title="msg('addPresetRow')"
```
mxmou:
```suggestion
            :button-title="msg('addPresetRow')"
```

webpages/settings/components/dropdown.js
```diff
@@ -0,0 +1,73 @@
+export default async function ({ template }) {
+  const Dropdown = Vue.extend({
+    props: ["buttonClass", "buttonTitle", "isDisabled", "alignStart"],
```
mxmou: This component has an `is-disabled` prop, reset-dropdown has `enabled`, and the color picker has `disabled`. They should probably be consistent.

webpages/settings/components/dropdown.js
```diff
@@ -0,0 +1,73 @@
...
+    data() {
+      return {
+        isOpen: false,
+        shiftAmountsByKey: {
+          ArrowUp: -1,
+          ArrowDown: 1,
+          ArrowLeft: -1,
+          ArrowRight: 1,
+          Home: -1000,
+          End: 1000,
```
mxmou:
```suggestion
          Home: -Infinity,
          End: Infinity,
```
There will probably never be a dropdown with more than 1000 items, but I think using Infinity makes the intent clearer.

webpages/settings/components/dropdown.js
```diff
@@ -0,0 +1,73 @@
...
+    methods: {
+      toggle() {
+        this.isOpen = !this.isOpen;
+        this.$root.closePickers({ isTrusted: true }, null, {
+          callCloseDropdowns: false,
+        });
+        this.$root.closeDropdowns({ isTrusted: true }, this); // close other dropdowns
+      },
+      listClick(e) {
+        if (e.target.tagName === "LI") {
```
mxmou:
```suggestion
        if (e.target.closest("li")) {
```
In a color setting's reset dropdown, the target might be the color preview.

webpages/settings/components/modal.html
```diff
@@ -1,63 +1,62 @@
...
+
   .close {
-    float: right;
+    padding: 0;
+    background-color: transparent;
+    border: none;
     cursor: pointer;
+  }
+
+  close > img {
```
mxmou:
```suggestion
  .close > img {
```

webpages/settings/components/reset-dropdown.js
mxmou: This line and line 14 throw an error because `toggle()` no longer exists.

webpages/settings/index.html
```diff
@@ -299,7 +306,7 @@ <h1 v-cloak>{{ msg("settings") }}</h1>
         </div>
       </div>
     </div>
-    <modal class="more-settings" :is-open.sync="moreSettingsOpen" :title="msg('moreSettings')" v-cloak>
+    <modal class="more-settings" :is-open.sync="moreSettingsOpen" :title="msg('moreSettings')" v-el:moreSettings>
```
mxmou: The `v-cloak` attribute should stay - otherwise the content of the modal can briefly appear at the bottom of the popup after switching from a different tab to Addons:

Image

webpages/settings/style.css
```diff
@@ -133,22 +137,28 @@ h1 {
...
+  background-color: transparent;
+  border: none;
   cursor: pointer;
-  transition: 0.2s ease;
-  filter: var(--content-icon-filter);
+  transition: background-color 0.2s ease;
 }
 
-.back-button:hover {
+.arrow-button:hover {
```
mxmou:
```suggestion
.arrow-button:hover,
.clickable-area:hover .arrow-button,
.addon-group:hover .arrow-button {
```

popups/scratch-messaging/popup.html
Samq64: I have decided to leave it in Scratch messaging for now because the buttons don't currently have proper collapse/expand messgaes like the settings page. I will commit in a few hours.

mxmou (commented):
There's one remaining bug: on Firefox, the focus outlines of `.filter-option`s and popup tabs aren't fully visible because the parent has `overflow: hidden`. A negative `outline-offset` would fix that.

mxmou (approved):
`.filter-option` still needs a negative `outline-offset`.

webpages/settings/components/dropdown.js
DNin01, line 1: It would be practical for the first list item to be focused when the dropdown is opened. That way, you don't have to press another key to enter the list.

webpages/settings/components/addon-setting.html
```diff
@@ -38,41 +43,45 @@
...
           class="filter-option"
           v-for="option of setting.potentialValues"
           :class="{'sel' : addonSettings[setting.id] === option.id, 'disabled': !addon._enabled}"
+          :disabled="!addon._enabled"
           @click="updateOption(option.id);"
         >
           {{ option.name }}
-        </div>
+        </button>
       </div>
```
DNin01: Arrow-key navigation would also be a good fit for select options. They're like radio buttons.

DNin01:
> Navigate between addons with the arrow keys?

<details><summary>I managed to get this working, but it relies on the addon cards' style attributes being an exact value (to skip over addons in collapsed groups).</summary>

<p>

`addon-body.html`
```diff
  <template>
- <div class="addon-body" v-show="shouldShow" :id="'addon-' + addon._addonId">
+   <div
+     class="addon-body"
+     v-show="shouldShow"
+     :id="'addon-' + addon._addonId"
+     tabindex="0"
+     @keydown="handleKeys"
+   >
      <div class="addon-topbar">
```

`addon-body.js`
```diff
        },
+       handleKeys(e) {
+         if (!e.target.classList.contains("addon-body")) return;
+         if (e.ctrlKey || e.metaKey || e.altKey) return;
+         const cards = document.querySelectorAll(".addons-container .addon-body:not([style='display: none;'])");
+         const index = Array.prototype.findIndex.call(cards, (i) => i === document.activeElement);
+         if (e.key === "ArrowUp") {
+           if (!cards[index - 1]) return;
+           e.preventDefault();
+           cards[index - 1].focus();
+         } else if (e.key === "ArrowDown") {
+           if (!cards[index + 1]) return;
+           e.preventDefault();
+           cards[index + 1].focus();
+         }
+       },
        toggleAddonRequest(event) {
```

</p>

There's probably a better way to do this. Maybe we could revisit it another time.</details>

DNin01:
The only change needed to make badge tooltips reveal with the keyboard is:

`addon-tag.html`
```diff
  <template>
    <div
      class="badge"
      v-if="shouldShow"
+     :tabindex="tagInfo.tooltipText ? 0 : -1"
      :class="{
        'tooltip': tagInfo.tooltipText,
```

And...

Recommended label in settings page

DNin01:
> Not sure if this is any better than just manully handling keypresses

That's even better! The options are announced as radio buttons by the screen reader.

webpages/settings/index.html
DNin01: This here, the other sidebar-toggle button, still needs to be labeled and focusable.

DNin01 (approved):
Everything so far looks OK to me now. The remaining todos could easily be saved for another PR, if you want.

Samq64:
Should the prevoius element before the dropdown button become focused when shift+tabbing inside one instead of the button itself? It's a simple change.

webpages/settings/components/addon-setting.html
```diff
@@ -73,15 +73,18 @@
...
-        <button
-          class="filter-option"
-          v-for="option of setting.potentialValues"
-          :class="{'sel' : addonSettings[setting.id] === option.id, 'disabled': !addon._enabled}"
-          :disabled="!addon._enabled"
-          @click="updateOption(option.id);"
-        >
-          {{ option.name }}
-        </button>
+        <div v-for="option of setting.potentialValues">
```
mxmou, line 76: The `<label>` elements have a `flex-grow` property which now doesn't work:

image

It should be applied to this outer `<div>` instead.

webpages/styles/components/filter.css
```diff
@@ -44,7 +44,7 @@
 .filter-options input:not(:checked) + label:hover {
   cursor: pointer;
 }
-.filter-option:last-child {
+.filter-options > :last-child > .filter-option {
```
mxmou, line 47: This change can then be reverted.

mxmou (approved):

Samq64:
Does this need to wait for v1.45?

DNin01:
Now that this is done, I'll try to work on improving screen reader support and some other more specific UI interactions.

PR #8579, opened 2025-09-21
Fix not all text being changed to white when timing tools heatmap enabled

Joeclinton1:
I noticed a bug in my timing tools PR: my css selection for textElements to force to be white (needed for when users have high contrast on, which has black text which doesn't contrast well with heatmap enabled) will exclude certain text elements meaning not all text gets changed to white.

I need to change lines 14-16 of addons\debugger\timing\HeatmapManager.js from 
```js
const textElements = Array.from(block.svgGroup_.children)
    .filter((el) => !el.classList.contains("blocklyDraggable"))
    .flatMap((el) => Array.from(el.querySelectorAll("text")));
```

to

```js
  const textElements = block.svgGroup_.querySelectorAll(
    ":scope > :not(.blocklyDraggable) text, :scope > text:not(.blocklyDraggable)"
  );
```

| Before | After |
|--------|-------|
| Before | image |

---
I additionally have now thrown in a second bug fix, that will make the realtime heatmap updates take effect even if the project runs for less than 100ms.

### DIFF
--- a/addons/debugger/timing/HeatmapManager.js
+++ b/addons/debugger/timing/HeatmapManager.js
@@ -113,7 +113,7 @@ class HeatmapManager {
   startRealtimeUpdates(heatmapMax) {
     this.stopRealtimeUpdates(); // Clear any existing interval
     this.realtimeUpdateInterval = setInterval(() => {
-      if (this.config.showHeatmap && this.isProjectRunning() && this.modifiedTimers.size > 0) {
+      if (this.config.showHeatmap && this.modifiedTimers.size > 0) {
         this.updateHeatmapColors(this.currentHeatmapMax, true); // true = only modified timers
       }
     }, 100); // Update every 100ms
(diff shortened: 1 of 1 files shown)


Samq64:
The block definition name is also white when it shouldn't be effected:
image

Joeclinton1:
> The block definition name is also white when it shouldn't be effected:

Ah ok, so this is going to need an even more complicated selector.

Joeclinton1:
@Samq64 Works now in my testing. Thanks for catching that edge case.

I also made it so that it will update the heatmap even when the project isn't running if it finds any block has been modified since last update. Because during my testing for this bug it was annoying me that I'd click the green flag with heatmap and profiling enabled and it wouldn't colour the blocks until I actually enabled the heatmap. (because if the project runtime is less than the update interval it won't hit the update)

Samq64 (approved):

PR #8584, opened 2025-09-25
Fix -0 being logged as 0 in debugger

Tacodiva:
In floating point math, there are two distinct values for 0 and -0. This difference is subtle, but it does make a difference in some obscure cases (eg `1/-0 === -Infinity` instead of `Infinity`). Because `(-0).toString() === "0"` when the debugger tries to log a statement which reports -0, it logs "0". [This can cause some confusion](https://github.com/TurboWarp/scratch-vm/issues/224). Making this debug tool print the value accurately (even when Scratch itself doesn't) makes sense to me.

Before: 
image

After:
image

### DIFF
--- a/addons/debugger/logs.js
+++ b/addons/debugger/logs.js
@@ -133,7 +133,7 @@ export default async function createLogsTab({ debug, addon, console, msg }) {
 
   const addLog = (text, thread, type) => {
     const log = {
-      text,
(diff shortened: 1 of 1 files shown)


mxmou (approved):

Joeclinton1:
Click the merge button my dude.

PR #8586, opened 2025-09-25
Fix debugger blocks not being coloured sometimes on project load

### LINKED ISSUE #8585: Debugger blocks are sometimes not coloured on initial project loading
### Description

When the editor is opened directly (for example from My Stuff, not from a project page), debugger blocks in the project sometimes have the wrong color. If a new block is added from the palette, it will have the correct color. Switching sprites fixes the issue.

### Steps to reproduce

1. Create a project that contains debugger blocks and save it.
2. Go to My Stuff.
3. Click the "See inside" button for that project.

Image

### Environment

Browser: Firefox 144.0b5
Scratch Addons version: 04b7130cabcdb7e63dedde7e6e8a9df54c238b5f
Operating system: Widows

Joeclinton1:
Resolves #8585

### Changes

Adds a recolouring step after initial injection to resolve the case in which the injection happens after blocks already created.

### Reason for changes

See issue.

### Tests

https://github.com/user-attachments/assets/e11f57e8-c741-47bf-8fdc-850ac5fa5987

### DIFF
--- a/addon-api/content-script/blocks.js
+++ b/addon-api/content-script/blocks.js
@@ -186,6 +186,13 @@ const injectWorkspace = (ScratchBlocks) => {
     };
   }
 
+  // recolour existing procedure blocks incase this injection is called after the blocks have already been created
+  for (const block of ScratchBlocks.mainWorkspace.getAllBlocks(false)) {
+    if (block.type === "procedures_call" || block.parentBlock_?.type === "procedures_call"){
+      block.applyColour?.() || block.updateColour?.();
+    }
+  }
+
   // We use Scratch's extension category mechanism to create a new category.
   // https://github.com/scratchfoundation/scratch-gui/blob/ddd2fa06f2afa140a46ec03be91796ded861e65c/src/containers/blocks.jsx#L344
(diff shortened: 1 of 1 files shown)


addon-api/content-script/blocks.js
```diff
@@ -186,6 +186,13 @@ const injectWorkspace = (ScratchBlocks) => {
     };
   }
 
+  // recolour existing procedure blocks incase this injection is called after the blocks have already been created
+  for (const block of ScratchBlocks.mainWorkspace.getAllBlocks(false)) {
```
mxmou: `mainWorkspace` can be the wrong workspace if the Make a Block modal is open, which might become an issue if the debugger addon gets dynamic enable in the future.

```suggestion
  for (const block of addon.tab.traps.getMainWorkspace().getAllBlocks(false)) {
```

Joeclinton1: So that exact code isn't going to work because we don't have addon. Also needs to be getWorkspace(). But also this problem will never occur as you can't have the modal open when project starts. I will make a change which will work, but it would also need a dynamic disable recolouring thing too, for full utility.

mxmou: You don't need await. This should work:
```suggestion
  for (const block of tabAPI.traps.getWorkspace().getAllBlocks(false)) {
```

mxmou (approved):

PR #8589, opened 2025-09-29
Fix welcome page & remove some dead code

Samq64:
Specifically some commented out `find-bar` code, the `key` setting type that was never added to the schema and the code that opened the Scratch messaging extension transion page.

Also fixes a bug that caused the welcome page to never open on install.

### DIFF
--- a/background/transition.js
+++ b/background/transition.js
@@ -6,13 +6,10 @@ chrome.runtime.onInstalled.addListener(async (details) => {
   const developerMode = (await chrome.management.getSelf()).installType;
   const currentVersion = chrome.runtime.getManifest().version;
   const [major, minor, _] = currentVersion.split(".");
-  if (details.previousVersion && details.previousVersion.startsWith("0")) {
-    chrome.tabs.create({ url: `https://scratchaddons.com/${localeSlash}scratch-messaging-transition/?${utm}` });
-  } else if (details.reason === "install" && !developerMode) {
-    chrome.tabs.create({ url: `https://scratchaddons.com/${localeSlash}welcome/?${utm}` });
-  }
-
   if (details.reason === "install") {
+    if (!developerMode) {
+      chrome.tabs.create({ url: `https://scratchaddons.com/${localeSlash}welcome/?${utm}` });
+    }
     chrome.storage.local.set({
       bannerSettings: { lastShown: `${major}.${minor}` },
     });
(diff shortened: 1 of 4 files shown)


background/transition.js
```diff
@@ -6,13 +6,10 @@ chrome.runtime.onInstalled.addListener(async (details) => {
   const developerMode = (await chrome.management.getSelf()).installType;
```
DNin01: If this is supposed to be a boolean variable, then it should be set to `(await chrome.management.getSelf()).installType === "development"`.

Samq64: It looks like the welcome popup might have been broken since #7157.

DNin01: Oh, yeah...

Welcome page on analytics

Luckily, I haven't seen anyone complain. Maybe we don't even need the welcome page anymore? :P

DNin01: Chrome already pops up a message telling you how to activate an extension you just installed, so that might be helping. On top of that, the default popup tab now is Addons, which probably also helps point new users in the right direction.

DNin01: ...Or maybe not.

Farewell page on analytics

Occasionally we get the rare review complaining about discoverability specifically, I've only seen two:

> - amazing but how do I turn something on? it wont work
>
> - i forgot how to open all the addons can anyone tell me how?

background/transition.js
DNin01, line 20: The install-URL- and uninstall-URL-setting statements could also be combined:

```suggestion
  if (details.reason === "install") {
    if (!developerMode) {
      chrome.tabs.create({ url: `https://scratchaddons.com/${localeSlash}welcome/?${utm}` });
    }
    chrome.storage.local.set({
      bannerSettings: { lastShown: `${major}.${minor}` },
    });
  }
  if (!developerMode) {
    chrome.runtime.setUninstallURL(`https://scratchaddons.com/${localeSlash}farewell?${utm}`);
  }
});
```

DNin01: Not sure why the suggestion isn't rendering correctly, but yeah.

DNin01:
@WorldLanguages I'd like to report how our welcome and farewell page rates have changed, but the analytics seem to be down—there's no data on the last 5 days.

DNin01:
`https://analytics.scratchaddons.com/api/plausibleevent` is returning a 404 error instead of logging a page view.

PR #8593, opened 2025-10-03
Specificity changes to support Face Sensing update

mxmou:
Scratch has increased the specificity of some of their CSS selectors.


Samq64 (approved):

WorldLanguages:
@mxmou Is this all we need for Face Sensing then?

mxmou:
> @mxmou Is this all we need for Face Sensing then?

Yes - unless the ST makes some additional changes before the update.

PR #8595, opened 2025-10-04
Do not reload when changing forceEnglish

DNin01:
### Changes

Changing the "Show addon names and descriptions in English" setting now only requires a settings page reload instead of a full extension reload.

https://github.com/user-attachments/assets/c9988f92-3741-4134-beb1-cff5b504b632

### Reason for changes

It's better to reload only the resources that are necessary.

### Tests

Tested in Edge 141. Addons still work after toggling setting.

Help me find edge cases?

### DIFF
--- a/background/load-addon-manifests.js
+++ b/background/load-addon-manifests.js
@@ -41,7 +41,7 @@ const localizeSettings = (addonId, setting, tableId) => {
   }
 };
 
-(async function () {
+async function loadAddonManifests() {
   const forceEnglish = await new Promise((resolve) => {
     chrome.storage.local.get("forceEnglish", (obj) => {
       resolve(!!obj.forceEnglish);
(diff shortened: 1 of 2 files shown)


WorldLanguages:
Probably no edge cases at the moment, but it could become a problem in the future.

Maybe you can achieve this exact same user experience (i.e. the tab not closing) while also reloading the extension. For example, by redirecting the user to a `data:` URL (or `about:blank` if you want) before reloading, and then sending back the user to the settings page as soon as possible.
This way you end up with the same user experience but less development headaches.

background/load-addon-manifests.js
```diff
@@ -239,4 +239,18 @@ const localizeSettings = (addonId, setting, tableId) => {
...
+// Data reload required when toggling the forceEnglish setting
+chrome.runtime.onMessage.addListener(async function (request, sender, sendResponse) {
+  if (request === "reloadAddonManifests") {
+    // Only attempt to reload manifests if they are already loaded
+    if (!scratchAddons.localState.ready.manifests) return;
+
+    scratchAddons.localState.ready.manifests = false;
+    scratchAddons.manifests = [];
+    await loadAddonManifests();
+    sendResponse("done");
```
Samq64, line 254: It looks like this respose is never received.
```
Unchecked runtime.lastError: The message port closed before a response was received.
```

WorldLanguages: It is not necessary to send a response.

If you really want to send one, then you need to return true; and avoid the use of async/await in favor of normal promises. (this is how asynchronously sending a response works with the chrome messaging API)

PR #8596, opened 2025-10-04
v1.44 update notification

WorldLanguages:
image


WorldLanguages:
Feedback on the update message?

DNin01:
This is another case where I would release a minor update with compatibility fixes first. Scratch said the Face Sensing blocks are coming "next week". That's probably next Friday, based on how they have previously timed announcements and releases.

WorldLanguages:
There's not a lot of commits on the master branch, so I don't think it hurts to ship a few new features

mxmou:
> Feedback on the update message?

I think it's good. For the image, something language-independent would be better - perhaps like this:

recolor_custom_blocks_notif

DNin01:
> There's not a lot of commits on the master branch, so I don't think it hurts to ship a few new features

Well, let's wrap it up by this weekend, then.

It's okay if we're not right on time because the breakage isn't too bad. It looks like only `columns` and `hide-flyout` would throw errors without the fixes, due to not being able to find the renamed extension button.

WorldLanguages:
I will change the image to the one suggested by mxmou and will also add more details about "improvements to debugger".

DNin01:
Perfect! Those updates were exciting.

DNin01:
We were too late...

PR #8603, opened 2025-10-05
debugger: add default label to timer blocks

mxmou:
### Changes

image

### Reason for changes

To make the input's purpose more obvious.

### Tests

Tested on Edge with old & new Blockly.

### DIFF
--- a/addon-api/content-script/blocks.js
+++ b/addon-api/content-script/blocks.js
@@ -46,7 +46,7 @@ export const getCustomBlock = (proccode) => {
 const getArgumentId = (index) => `arg${index}`;
 
 const getNamesIdsDefaults = (blockData) => [
-  blockData.args,
+  blockData.args.map((arg) => arg.name),
   blockData.args.map((_, i) => getArgumentId(i)),
   blockData.args.map(() => ""),
 ];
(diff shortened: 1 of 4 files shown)


mxmou:
@WorldLanguages Do you think this could be merged for v1.44? It would probably make the new debugger blocks slightly easier to understand for first-time users.

WorldLanguages:
Sure

Joeclinton1:
Btw, if we were to do this properly, we'd do it the same way as messages with a drop down for new timer and rename timer. But this works I guess for now. If you guys think a dropdown thing is needed I can make it happen.

mxmou:
> Btw, if we were to do this properly, we'd do it the same way as messages with a drop down for new timer and rename timer. But this works I guess for now. If you guys think a dropdown thing is needed I can make it happen.

I'm not sure what's better - the disadvantage of a dropdown is that you can't type into it directly, making it much slower to create many timers. Perhaps a text input with an added dropdown menu is the best option. If we do something like that, it should probably be in the next release so that this one isn't delayed.

DNin01:
I think the text inputs are fine for these blocks.

PR #8606, opened 2025-10-06
Add relationships to 6 addons

DNin01:
### Changes

Adds the following addon relations to help users find what they're looking for:

- #### `ctrl-click-run-scripts` → `fix-pasted-scripts`
  
  In case users find that addon while looking for an addon that prevents scripts from running when duplicated blocks are clicked into place.
  
- #### `turbowarp-button` ↔ `60fps`
  
  Because TurboWarp also has a frame rate control.
  
- #### `debugger` ↔ `fps` ↔ `clones`
  
  Because they all make measurements.

### DIFF
--- a/addons/60fps/addon.json
+++ b/addons/60fps/addon.json
@@ -42,6 +42,7 @@
     }
   ],
   "tags": ["player", "featured"],
+  "relatedAddons": ["turbowarp-player"],
   "versionAdded": "1.1.0",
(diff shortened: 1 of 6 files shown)


DNin01:
> Maybe also `turbowarp-player` <--> `live-featured-project`?

I don't think users usually associate those two addons; each one serves a very different purpose despite them having TurboWarp integration in common.

Issue #8609, opened 2025-10-07
Scratch's top bar on Fullscreen Mode


ElJuanEn55:
### Description

In fullscreen mode, there's a bug, where the top bar doesn't hide, also i'm unable to press the green flag, pause or stop buttons (I know isn't a lot of info. but that's how i can describe it).

### Steps to reproduce

1. Enable Fullscreen Mode addon.
2. Run a project and enter fullscreen
3. The Scratch's top bar (the top purple one) will not hide and also can't press the Green Flag, Pause and Stop buttons

### Demonstration(s) and additional context

I can't upload the video via GitHub, so please look at this link:
https://drive.google.com/file/d/1cehMvHIXeoqOekIdGP2JHBKQqofE3RDi/view?usp=sharing

EDIT: I discovered that only happens when you are in the editor, maybe this info can make the fix easier.

### Environment

Browser: Ver. 140.0.7339.208 (x64 version)
Scratch Addons version: 1.43.0
Operating system: Windows 10

Samq64:
I can't repro. Here's a frame from the attatched video:

<img width="1012" height="565" alt="Image" src="https://github.com/user-attachments/assets/d7a0e735-230f-4c21-acdd-f8d2de67c471" />

mxmou:
Does clicking the Add extension button, so that the "there's a new extension" popup disappears, fix it?

ElJuanEn55:
Nope, clicking the extensions button doesn’t work, it’s still there

ElJuanEn55:
Wait a minute, I tested now and is not the fullscreen addon, but I think other addon isn't allowing the top bar to disappear (Scratch without any addons hides the top bar, so is not a Scratch Bug)

ElJuanEn55:
I found the problem: it was the 'Auto-hide block palette'. This is preventing the top bar from hiding. Also, I can no longer pick up blocks from the palettes because they are not touching the left-hand sidebar containing all types of blocks (Motion, Sensing, Sound, etc.).
@Samq64 (just in case 😅)

FlamedDogo99:
See #8614

PR #8610, opened 2025-10-07
Fix better-img-uploads and hide-delete-button CSS selectors

Samq64:
### Changes

Fixes some CSS selectors after Scratch's face sensing extension update.

### Tests

Tested on Chromium


mxmou (approved):
I suggest using `*=` instead of `^=` in case a future update adds another class before `gui_tabs`. Otherwise this looks good.

WorldLanguages:
Are more PRs incoming, or are we good to release now?

Samq64:
This one should be the last unless there is more info on #8609.

PR #8591, opened 2025-10-02
Prepare for upcoming Face Sensing release

mxmou:
### Changes

Changes related to the new Face Sensing extension:
* Updates editor-dark-mode and editor-theme3 to recolor the face sensing icon. Note that theme3 doesn't recolor any extension icons in the category menu in high contrast mode yet - I'll probably implement that in a future PR.
* The Add Extension button has been moved to a separate component, which requires changes to several addons that refer to it using the class name.
* Adds styling for the pulsing animation that appears on the Add Extension button and the Face Sensing extension in the library. To test this, change the value of [this prop](https://github.com/scratchfoundation/scratch-editor/blob/7f483ae/packages/scratch-gui/src/components/gui/gui.jsx#L485) to true by editing the file. The animation only appears when the editor is opened for the first time - to see it again, delete the localStorage key `hasIntroducedFaceSensing`.

### Tests

Tested the [face sensing branch of scratch-editor](https://github.com/scratchfoundation/scratch-editor/tree/UEPR-282-face-sensing) on Edge and Firefox.

PR #8611, opened 2025-10-07
Remove outdated code after React update

mxmou:
### Changes

Removes code that isn't needed anymore because Scratch has updated to React 18.

### Tests

Tested on Edge and Firefox. Verified that addons which use the context menu API (e.g. move-to-top-layer) or `getPaper()` (e.g. onion-skinning) still work.

### DIFF
--- a/addons/columns/style.css
+++ b/addons/columns/style.css
@@ -64,7 +64,6 @@
   align-items: center;
 }
 
-[class*="gui_extension-button-container_"],
 [class*="extension-button_extension-button-container_"] {
   top: var(--sa-add-extension-button-y);
   bottom: auto;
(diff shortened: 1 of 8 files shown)


Samq64:
#8591 also left some selectors for compatibility that didn't end up being necessary.

addon-api/content-script/Trap.js
```diff
@@ -29,14 +29,14 @@ export default class Trap extends Listenable {
    * @private
    */
   get REACT_INTERNAL_PREFIXES() {
-    return ["__reactInternalInstance$", "__reactFiber$"];
+    return ["__reactFiber$"];
```
Samq64, line 32: In cases like this does it make sense to keep using an array?

mxmou: I think it does - arrays will be easier to update if/when React renames these again.

PR #8622, opened 2025-10-09
Fix `paint-by-default`

### LINKED ISSUE #8615: `paint-by-default` broke
```text
The "paint costume by default" doesn't work anymore - it says "paint" when I
hover over it, but when I click, it goes to the library. This is getting on my
nerves, I know it's just a tiny bit more work to just go a little bit up but
it's annoying when I accidentally select the library. Please fix this, I believe
this was caused by a recent update released like under 5 hours ago
```
Scratch Addons version: 1.43.0

Samq64:
Resolves #8615

### Changes

~~Moves `paint-by-default`'s event listeners from the body to the individual buttons.~~

### Reason for changes

After the React 18 update using `stopPropagation()` in the click listener on the body isn't enough to prevent the asset library from opening. ~~I have done the same to `mousemove` for consistency.~~

### Tests

Tested on Chromium

### DIFF
--- a/addons/paint-by-default/userscript.js
+++ b/addons/paint-by-default/userscript.js
@@ -111,7 +111,6 @@ export default async function ({ addon, console }) {
       elementToClick.click();
     },
     {
-      bubble: true,
       capture: true,
     }
   );


mxmou:
This adds a `waitForElement` call with no `reduxEvents`, which isn't ideal for performance. Have you tried listening to `pointerup` instead of `click`, like `no-sprite-confirm`?

Samq64:
> Have you tried listening to `pointerup` instead of `click`, like `no-sprite-confirm`

It doesn't prevent the asset library from opening either so I have added some `reduxEvents` for now.

Samq64:
> maybe adding `capture: true` to the listener

Thanks! That was all that was necessary.

PR #8623, opened 2025-10-09
Only welcome users on store installs

### LINKED ISSUE #8597: Do not open `/welcome` if extension was installed without user consent
According to Chrome Web Store data, some users didn't install the extension by their own, but instead it was the device admins that did.

According to an LLM this is example code we can use to detect that case, and avoid disturbing the user with a tutorial they didn't ask for:
```js
export async function getInstallContext() {
  const { installType, mayDisable } = await chrome.management.getSelf();
  return {
    installType,
    installedByAdmin: installType === 'admin',
    userInstalled: installType === 'normal',
    sideloaded: installType === 'sideload',
    userCanRemove: !!mayDisable,
  };
}

```

DNin01:
Resolves #8597

### Changes

The welcome page and uninstall URL are now only used on [normal installs](https://developer.chrome.com/docs/extensions/reference/api/management#type-ExtensionInstallType:~:text=normal%3A%20The%20extension%20was%20installed%20normally%20via%20a%20.crx%20file) (directly from an extension store), so users won't see these pop up if Scratch Addons was installed through developer mode, an enterprise policy, or some other means.

### DIFF
--- a/background/transition.js
+++ b/background/transition.js
@@ -3,18 +3,18 @@ const utm = `utm_source=extension&utm_medium=tabscreate&utm_campaign=v${chrome.r
 const uiLanguage = (chrome.i18n.getUILanguage && chrome.i18n.getUILanguage()) || navigator.language;
 const localeSlash = uiLanguage.startsWith("en") ? "" : `${uiLanguage.split("-")[0]}/`;
 chrome.runtime.onInstalled.addListener(async (details) => {
-  const developerMode = (await chrome.management.getSelf()).installType === "development";
+  const normalInstall = (await chrome.management.getSelf()).installType === "normal";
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Thanks!

WorldLanguages:
How did you merge without approvals? That means the master branch is unprotected :O
I'll fix that

DNin01:
> How did you merge without approvals? That means the master branch is unprotected :O I'll fix that

@ScratchAddons/ability-to-merge-any-pull-request can bypass rules ;)

Maybe we should only be able to merge with 1 approval, instead of the usual 2.

WorldLanguages:
To clarify, by unprotected I meant force-pushing into the master branch was allowed. Whoops.

PR #8626, opened 2025-10-10
editor-dark-mode: style face sensing tooltips

Samq64:
### Changes

Applies `editor-dark-mode`'s colours to the face sensing tooltips.

### Reason for changes

Might as well sense someone sent feedback about it and we'll need to publish a patch release anyway.

### Tests

Tested on Chromium. The weird selectors are because the the tooltip in the library has a `-modal` prefix while the other one doesn't

### DIFF
--- a/addons/editor-dark-mode/experimental_editor.css
+++ b/addons/editor-dark-mode/experimental_editor.css
@@ -484,8 +484,11 @@ img[class*="tool-select-base_tool-select-icon_"],
 .driver-popover[class*="tooltip-face-sensing"] {
   border-color: var(--editorDarkMode-primary);
 }
-.driver-popover[class*="tooltip-face-sensing"] .driver-popover-arrow-side-right {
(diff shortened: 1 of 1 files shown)


addons/editor-dark-mode/experimental_editor.css
```diff
@@ -479,9 +480,13 @@ img[class*="tool-select-base_tool-select-icon_"],
 [class*="stage-selector_stage-selector_"]:hover,
 [class*="drag-layer_image_"],
 [class*="sound-editor_button_"]:focus::before,
-[class*="library-item_library-item_"]:hover {
+[class*="library-item_library-item_"]:hover,
+.driver-popover[class*="tooltip-face-sensing"] {
   border-color: var(--editorDarkMode-primary);
 }
+.driver-popover[class*="tooltip-face-sensing"] .driver-popover-arrow-side-right {
```
mxmou: In the modal, there can also be a `.driver-popover-arrow-side-left` with a left border.

PR #8634, opened 2025-10-13
Rename sprite folders to asset folders

ampelectrecuted:
### Changes

Change sprite folders to asset folders.

### Reason for changes

Sprite folders is a misleading name as it implies that the extension only creates folders for sprites.

### Tests

Will test soon.

### DIFF
--- a/addons/folders/addon.json
+++ b/addons/folders/addon.json
@@ -1,5 +1,5 @@
 {
-  "name": "Sprite folders",
+  "name": "Asset folders",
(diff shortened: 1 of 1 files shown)


mxmou:
> Sprite folders is a misleading name as it implies that the extension only creates folders for sprites.

The proposed name isn't much better: in Scratch, the word "asset" only refers to costumes and sounds, not sprites. I think the current name is fine because it describes the addon's main use case.

WorldLanguages:
This should be an issue, not a pull request.

PR #8646, opened 2025-10-16
`mediarecorder`: Allow choosing preferred format

### LINKED ISSUE #8074: video recorder looking weird when i watch videos recorded by the feature.
### Description

when i watch the videos i recorded using the video recorder, the video is weird, its either because of my laptop or the video recorder, please help!!

i get this
videoframe_22920

instead of this
download (14)

### Steps to reproduce

1.
2.
3.

_No response_

### Environment

Browser: 
Scratch Addons version: 
Operating system:

### LINKED ISSUE #8087: Another Bug I Found (NEEDS FIXED BADLY)
### Description

I Cant Record Project While Im In It. It Worked Before Though. It Was How I Made The Video HD.

### Steps to reproduce

1.
2.
3.

_No response_

### Environment

Browser: Google Chrome,
Scratch Addons version: Newest.
Operating system: Windows.

Samq64:
Resolves #8087
Resolves #8074

### Changes

Allows selecting between mp4 and webm as the preferred format from the settings page while falling back to the other if unsupported by the browser.

### Reason for changes

Neither format is perfect so giving the choice of which format to use could might help some with recording issues but I didn't see a reason to include it in the editor UI.

### Tests

Tested on Chromium and Firefox

### DIFF
--- a/addons/mediarecorder/userscript.js
+++ b/addons/mediarecorder/userscript.js
@@ -55,11 +55,11 @@ export default async ({ addon, console, msg }) => {
         "video/webm",
         // Safari only supports encoding H264 as mp4
         "video/mp4",
-      ]
-    }
+      ],
+    };
     mimeType = mimeOrder[addon.settings.get("preferred")].find((i) => MediaRecorder.isTypeSupported(i));
     fileExtension = mimeType.split(";")[0].split("/")[1];
-  }
+  };
 
   addon.settings.addEventListener("change", setMimeType);
   setMimeType();


DNin01:
Friendly competition: #8647 😉

I actually created my own format picker a while ago but never proposed it via pull request.

DNin01:
In your pull request, the user gets to choose their preferred recording format through the addon settings, which includes recording formats that may be unsupported on the user's browser.

In my pull request, the user gets to choose a format every time they record, from the formats available to them.

I kind of like both ways.

mxmou:
I think it's more convenient if the format can be chosen in the same place as other options.

DNin01:
> I think it's more convenient if the format can be chosen in the same place as other options.

That's what I would say, too.

PR #8656, opened 2025-10-18
Fix studio navigation

### LINKED ISSUE #8653: Unnecessary studio browser tab reloads
### Description

When navigating between studio tabs for the first time the page is reloaded when Scratch Addons is running (even without addons) but this does not happen when the extension is entirely disabled.

### Steps to reproduce

1. Disable Scratch Addons from the browser's settings
2. Open a studio
3. Navigate between the tabs
4. Do the same with Scratch Addons enabled

> Uncaught TypeError: can't access property "mode", state.scratchGui is undefined

This doesn't happen in v1.44.1 and #8570 is the only PR that changed `/content-scripts/inject/module.js` since then.

### Environment

Browser: Chromium 141
Operating system: Linux

mxmou:
Resolves #8653

### DIFF
--- a/content-scripts/inject/module.js
+++ b/content-scripts/inject/module.js
@@ -213,6 +213,9 @@ function onDataReady() {
 }
 
 function editorClassCheck() {
+  const pathname = location.pathname.toLowerCase();
+  const split = pathname.split("/").filter(Boolean);
+  if (!isScratchGui && split[0] !== "projects") return;
   let isInEditor = false;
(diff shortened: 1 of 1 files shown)


DNin01 (approved):

WorldLanguages:
How on earth did the previous code cause reloads? This might be the most confusing bug I've ever seen.

WorldLanguages:
Oh, we were throwing an exception inside a polluted method. Ok

PR #8657, opened 2025-10-18
`player-thumb`: Use `background-image` instead of an `<img>` element

### LINKED ISSUE #8654: `player-thumb`: Thumbnail doesn't appear during loading if `editor-dark-mode` is enabled
### Description

The whole point of `player-thumb` is to show the thumbnail of the project you're loading into. However, `editor-dark-mode` applies a solid background color to the loading screen, which overrides `player-thumb`'s setting of the background image.

### Steps to reproduce

1. Enable `player-thumb` and open a project page.
2. Observe that the thumbnail is visible under the loading screen.
3. Enable `editor-dark-mode` and reload.
4. Observe that the loading screen has a solid background color.

_No response_

### Environment

Scratch Addons version: v1.44.1

Samq64:
Resolves #8654

### Changes

Applies a `background-image` and a single colour gradient instead of inserting a new `` element. Also allows dynamically enabling or disabling the addon as well as changing how much the thumbnail is darkened.

### Reason for changes

It's simpler and less finicky. Also someone sent feedback about bad contrast (although in that case the proposed solution was to blur it).

### Tests

Tested on Chromium with a few addon combinations, but this still needs more testing.

### DIFF
--- a/addons/player-thumb/scratch3to2.css
+++ b/addons/player-thumb/scratch3to2.css
@@ -3,6 +3,6 @@
 }
 :root:not(.sa-editor) .sa-loader-background {
   border-radius: 0;
-  top: 38px;
-  height: calc(100% - 39px);
+  top: 40px;
+  height: calc(100% - 41px);
 }


addons/player-thumb/addon.json
```diff
@@ -31,6 +31,22 @@
       "runAtComplete": false
     }
   ],
+  "settings": [
+    {
+      "name": "Darken (%)",
```
mxmou:
```suggestion
      "name": "Darken thumbnail (%)",
```

addons/player-thumb/addon.json
```diff
@@ -31,6 +31,22 @@
...
+      "id": "darken",
+      "type": "integer",
+      "min": 0,
```
mxmou: I don't think it should be possible to set this to zero - it makes the loading text impossible to read if the thumbnail is mostly white.

addons/player-thumb/userscript.js
```diff
@@ -2,47 +2,27 @@ export default async function ({ addon, console }) {
   if (addon.tab.editorMode === "editor") return;
 
   const projectId = location.pathname.split("/")[2];
-  const thumb = document.createElement("img");
-  thumb.src = `https://uploads.scratch.mit.edu/get_image/project/${projectId}_480x360.png`;
-  thumb.classList = "sa-project-thumb loading";
-  thumb.draggable = false;
+  document.documentElement.style.setProperty(
+    "--thumb-src",
```
mxmou:
```suggestion
    "--sa-player-thumb-src",
```

addons/player-thumb/userscript.js
```diff
@@ -2,47 +2,27 @@ export default async function ({ addon, console }) {
...
+    `url(https://uploads.scratch.mit.edu/get_image/project/${projectId}_480x360.png)`
+  );
 
   const stageWrapper = await addon.tab.waitForElement('div[class*="stage-wrapper_stage-wrapper_"]');
   addon.tab.redux.initialize();
-
-  // It's possible this runs after the project loads even without dynamic enable
-  if (addon.tab.redux.state?.scratchGui?.projectState?.loadingState === "SHOWING_WITH_ID") return;
-  // Second check in case of a URL change
-  if (addon.tab.editorMode === "editor") return;
```
mxmou, line 16: Is this really not needed?

addons/player-thumb/userstyle.css
```diff
@@ -1,56 +1,49 @@
...
+  width: calc(100% - 2px);
   border-radius: calc(0.5rem - 1px);
 }
 
-:root:not(.sa-editor) .sa-loader-background,
-:root:not(.sa-editor) .sa-project-thumb.loading {
+:root:not(.sa-editor) .sa-loader-background {
   top: calc(2.75rem + 1px);
   height: calc(100% - 2.75rem - 2px);
 }
```
mxmou, line 28: The way this is currently written, it isn't very easy to see which styles get applied to which element. It might be better to use a separate rule for each element, even if that results in some duplication.

mxmou:
> Also someone sent feedback about bad contrast (although in that case the proposed solution was to blur it).

Another option is to put the loading message and progress bar into a transparent black container. Then both the thumbnail and the foreground elements will be clearly visible.

addons/player-thumb/scratch3to2.css
```diff
@@ -1,12 +1,11 @@
 :root:not(.sa-editor) [class*="stage-wrapper_stage-wrapper"] {
   position: relative;
 }
-:root:not(.sa-editor) .sa-loader-background,
-:root:not(.sa-editor) .sa-project-thumb {
-  border-radius: 0;
+:root:not(.sa-editor) div[class*="stage_green-flag-overlay-wrapper_"] {
+  margin: 0;
```
mxmou: Could `userstyle.css` set margin instead of top and left for this element to make things a little simpler (this override wouldn't be needed)? Alternatively, scratch3to2 could be changed to use top and left.

addons/player-thumb/userscript.js
Samq64: If it reaches this point it will just throw an error in the console when querying the loader background but I can re-add the check.

addons/player-thumb/scratch3to2.css
```diff
@@ -1,12 +1,8 @@
...
 }
-:root:not(.sa-editor) .sa-loader-background,
-:root:not(.sa-editor) .sa-project-thumb {
+:root:not(.sa-editor) .sa-loader-background {
   border-radius: 0;
-}
-:root:not(.sa-editor) .sa-loader-background,
-:root:not(.sa-editor) .sa-project-thumb.loading {
   top: 38px;
   height: calc(100% - 39px);
```
mxmou: I wrote this part of the code, but it's wrong and I think this PR is a good opportunity to fix it:
```suggestion
  top: 40px;
  height: calc(100% - 41px);
```

mxmou (approved):

PR #8660, opened 2025-10-20
dark-www & scratch3to2: update ToU modal styles

mxmou:
### Changes

Updates the styles for Scratch's Terms of Use modal, which were first added in #8232. The modal will be shown when the terms are updated.

image

### Tests

Tested locally on Edge.


Samq64:
Will this modal be shown to users soon?

mxmou:
> Will this modal be shown to users soon?

I don't think so, but the ST hasn't said anything about the date.

WorldLanguages (approved):
Untested

PR #8665, opened 2025-10-26
Use independent setting names

DNin01:
### Changes

Some addon settings contained fragmented setting names which are fulfilled by setting values:
> - Show... (icon and text/icon only/text only)
> - Show menu... (icons and labels/icons only/labels only)
> - Show statuses... (everywhere/only on others' profiles)

When this happens, the setting name can't stand alone, and it is difficult to understand its entire meaning before reading the setting values. This is more of a problem for screen reader users (who may not see the punctuation and may not immediately notice the full context) and possibly when localizing in other languages.

I fixed these by completing the setting names or changing them entirely.

Note: "Show statuses on..." is still indefinite, but the incomplete verb phrase still contains a subject and it's clear that the next word would be a type of noun, so it should serve fine as a setting header.

### DIFF
--- a/addons/custom-menu-bar/addon.json
+++ b/addons/custom-menu-bar/addon.json
@@ -30,7 +30,7 @@
   "relatedAddons": ["discuss-button", "editor-compact"],
   "settings": [
     {
-      "name": "Menu item appearance",
+      "name": "Menu button appearance",
       "id": "menu-labels",
       "type": "select",
       "potentialValues": [


addons/custom-menu-bar/addon.json
```diff
@@ -30,7 +30,7 @@
   "relatedAddons": ["discuss-button", "editor-compact"],
   "settings": [
     {
-      "name": "Show menu...",
+      "name": "Menu item appearance",
```
mxmou: This setting controls the menus themselves, not their items. A better name might be "menu appearance" or "dropdown appearance".

DNin01: What is an item on a "menu bar" with items that are menus that themselves have menu items? Lol

Yeah, probably "dropdown appearance".

mxmou: It could also be "dropdown menu appearance".

DNin01: That could refer to either the button that opens the menu or the whole menu itself.

Do these kinds of buttons even have a name?

PR #8669, opened 2025-10-29
Remove `.sa-body-editor` and use `.sa-editor` instead

mxmou:
### Changes

Removes the `.sa-body-editor` class.

### Reason for changes

The new `.sa-editor` class, which is added to the root element instead of the body, makes it redundant.

### Tests

Tested on Edge.


Samq64:
Could the more complicated full screen selectors be replaced with `.sa-fullscreen` since you are already changing many of those lines?

PR #8676, opened 2025-11-12
discuss-button: support new Membership link

mxmou:
### Changes

Scratch will remove the About link from the navigation bar and add a new Membership link instead, which will only be visible when logged out. This PR updates the customizable navigation bar addon accordingly:

* Rewrites the userscript for scratch-www pages to fix a crash
* Adds a new setting controlling whether each link is always visible or only when (not) logged in
* Updates the default settings

### Tests

Tested locally on Edge.

### DIFF
--- a/background/get-addon-settings.js
+++ b/background/get-addon-settings.js
@@ -744,6 +744,12 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
             // Respect the value of enabledByDefault (by doing nothing)
           }
         }
+
+        if (addonId === "discuss-button" && settings.showMembership === undefined) {
+          if (settings.items) settings.items = settings.items.filter((i) => i.url !== "/about");
+          settings.showMembership = true;
+          madeAnyChanges = madeChangesToAddon = true;
+        }
       }
 
       if (addonsEnabled[addonId] === undefined) addonsEnabled[addonId] = !!manifest.enabledByDefault;


addons/discuss-button/items_scratchr2.js
mxmou, line 1: This file is mostly the same as the old userscript.js.

addons/discuss-button/items_scratchwww.js
mxmou, line 1: The old code crashes on www pages because:
- When the page is initially loaded, the Membership link is shown.
- The addon runs and removes all the links, replacing them with new elements.
- Then, when the session data is loaded, React tries to remove the Membership link. Since the element has already been removed by the addon, this causes a crash.

I've rewritten it to override the navbar's render function instead of modifying the DOM directly.

addons/discuss-button/addon.json
```diff
@@ -61,32 +66,56 @@
...
+              "id": "loggedIn",
+              "name": "When logged in"
+            },
+            {
+              "id": "loggedOut",
+              "name": "When not logged in"
+            }
+          ],
+          "default": "always"
         }
```
DNin01: Sorry, but I don't think adding this option to every item is the best way to do this. It's unlikely that users will be signing in/out constantly and just to have different navigation bar items. It makes this option more obscure.

Here are some alternative solutions:
- Do not modify the nav bar when signed out; only apply nav bar customizations when signed in.
- Ignore "Membership"; let it show when signed out regardless of addon settings.
- Add a setting "Remove default navigation bar items". When off, custom nav bar items would be appended to the default nav bar. When on, the default items would be replaced completely by the custom items (the current behavior).

(If we're in a hurry, we could do option 1 or 2 now and option 3 later.)

DNin01 (changes requested):

addons/discuss-button/addon.json
CST1229: Or:
- Don't change the navbar while logged out with the addon at all (always keep the About button instead)

DNin01: Well, I think Scratch is removing the About link from the nav bar entirely. For everyone. I think that's not a _bad_ idea, considering:
- For signed out users, "About Scratch" is already a link on the front page
- The About page is most useful for parents and educators, not signed-in kids
- It takes up space on the nav bar

If people want to bring back the About link by using this addon, I don't know why they would, but that should still be possible.

addons/discuss-button/items_scratchwww.js
DNin01: What kind of crash does the old code trigger? Does the page stop working?

mxmou: > What kind of crash does the old code trigger? Does the page stop working?

It shows the full-screen "Scratch has crashed" message.

addons/discuss-button/addon.json
mxmou: Maybe there could be another setting, separate from the list of items, that would allow the user to choose whether to show the Membership link.

DNin01: > Maybe there could be another setting, separate from the list of items, that would allow the user to choose whether to show the Membership link.

That's the only other thing I've thought of so far.

mxmou: I've added the setting.

mxmou:
We'll need to decide what to do with existing users' settings: should they be kept (in that case, the new setting should be turned off for them) or should the About item be removed?

WorldLanguages:
Please ping me when this is done so that we can release a new version

addons/discuss-button/items_scratchwww.js
Samq64: I can't repro with just `discuss-button` or all addons enabled.

DNin01:
> We'll need to decide what to do with existing users' settings: should they be kept (in that case, the new setting should be turned off for them) or should the About item be removed?

I quickly added a setting update: if after the update the addon settings contain an item with a URL of "/about", it is removed.

(If we release this patch before Scratch adds the Membership button, then if a user has this addon enabled, that button will appear on their nav bar early. It'll also go to a "not found" page, but that's better than crashing the site.)

DNin01:
@WorldLanguages, I think the PR is ready.

Things to cherry-pick for this next hotfix (v1.44.2):
- https://github.com/ScratchAddons/ScratchAddons/commit/4622c000afafd67bb1da2ff760f6b2756ab43009
- https://github.com/ScratchAddons/ScratchAddons/commit/c1b9c89d1c348c343a1f6dfe9c6dd925a24bc6ca
- This PR

DNin01:
@WorldLanguages I've already done the cherry-picking and prepared the next hotfix on [branch `v1.44.x`](https://github.com/ScratchAddons/ScratchAddons/tree/v1.44.x) for you to release.

background/get-addon-settings.js
```diff
@@ -744,6 +744,12 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
             // Respect the value of enabledByDefault (by doing nothing)
           }
         }
+
+        if (addonId === "discuss-button" && settings.showMembership === undefined) {
```
mxmou, line 748: This doesn't work: `showMembership` can't possibly be undefined here because all undefined settings are set to their default value on line 213.

DNin01: Oh well

WorldLanguages: This is why you test things like this 😄

mxmou: Fixed by #8685

PR #8678, opened 2025-11-14
Fix select settings inside table items

mxmou:
Moved from #8676. Makes select settings work correctly when they're inside a table row by giving them unique names and IDs.


WorldLanguages (approved):
Untested

PR #8684, opened 2025-11-19
Don't check `typeof browser` to detect Firefox

### LINKED ISSUE #8682: Chromium will soon support `window.browser`
We check if `window.browser` exists in several places to determine whether the extension is running on Firefox. This will soon stop working because the `browser` namespace will be available in Chromium as well, causing the Firefox-only code to run on both browsers.

On the latest build of Chromium, the service worker fails to start with the following error message:
Image

And in Edge 144 (dev channel):
Image

mxmou:
Resolves #8682

### Changes

Changes the way background and settings page code detects that it's running on Firefox: instead of checking if `window.browser` exists, it now checks if a Firefox-specific API is available. I've made a shared module to detect the browser so that it's easy to update in the future if needed.

### Reason for changes

The `window.browser` namespace will soon exist in both Chromium and Firefox.

### Tests

Tested on Edge v144 (dev channel) and Firefox.

### DIFF
--- a/addon-api/common/Self.js
+++ b/addon-api/common/Self.js
@@ -1,4 +1,5 @@
 import Listenable from "./Listenable.js";
+import { isFirefox } from "../../libraries/common/cs/detect-browser.js";
 
 /**
  * Represents information about the addon.
@@ -13,8 +14,7 @@ export default class Self extends Listenable {
     this._addonId = info.id; // In order to receive fireEvent messages from background
     this.id = info.id;
     this._addonObj = addonObj;
-    // catches both Chrome and Chromium
-    this.browser = /Chrom/.test(navigator.userAgent) ? "chrome" : "firefox";
+    this.browser = isFirefox() ? "firefox" : "chrome";
     this.disabled = false;
     this.addEventListener("disabled", () => (this.disabled = true));
     this.addEventListener("reenabled", () => (this.disabled = false));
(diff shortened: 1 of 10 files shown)


yuri-kiss:
what if chrome goes and adds this api too though, also maybe try more strict checking?

stuff like `Error().fileName != (void 0)` has worked for a long time to detect firefox.

mxmou:
My first idea was to call the `getBrowserInfo()` function if it exists and check the result. Unfortunately, it's async, so that would require additional changes in each place that needs to check the browser.

> stuff like `Error().fileName != (void 0)` has worked for a long time to detect firefox.

That's probably better. We could also just check the `userAgent` - some addons already do that.

yuri-kiss:
> My first idea was to call the `getBrowserInfo()` function if it exists and check the result. Unfortunately, it's async, so that would require additional changes in each place that needs to check the browser.
> 
> > stuff like `Error().fileName != (void 0)` has worked for a long time to detect firefox.
> 
> That's probably better. We could also just check the `userAgent` - some addons already do that.

yeah that is what I do when I am to bored to get feature based detection, I do the error thing and the userAgent. (though the user agent lies sometimes)

mxmou:
I've changed the detection code so that it works outside the extension context. I used `Error.fileName` as suggested by @yuri-kiss, but that can be easily changed if needed.

yuri-kiss (approved):
ik my review means nothing but, this works on older chromium versions and firefox too. :3

PR #8689, opened 2025-11-22
Add keyboard controls to table row handles

DNin01:
Resolves (from #8577):
> ` ` Re-order table rows with the keyboard?

### Changes

Before, the handles used to reorder table rows in addon settings only responded to mouse input. Now, you can focus the handle and press the up and down arrow keys to move the table row. When using a screen reader, the handle is supposed to work like a [spin button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/spinbutton_role), which also allows items to be moved incrementally.

### Reason for changes

For better accessibility.

I don't know if this is usually how moving items works, but it'll do.

### Tests

Changes tested using Windows Narrator and Edge.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -203,6 +203,12 @@
   "deleteRow": {
     "message": "Delete row"
   },
+  "moveRow": {
+    "message": "Move row"
+  },
+  "dragRow": {
+    "message": "Drag or use the arrow keys to move"
+  },
   "scratchAddonsTheme": {
     "message": "Scratch Addons theme"
   },
--- a/webpages/settings/components/addon-setting.html
+++ b/webpages/settings/components/addon-setting.html
@@ -28,7 +28,18 @@
               >
                 <img class="icon-type" src="../../images/icons/close.svg" />
               </button>
-              <button :disabled="!addon._enabled" class="addon-buttons handle">
+              <button
+                v-if="addonSettings[setting.id].length > 1"
+                :disabled="!addon._enabled"
+                role="spinbutton"
+                class="addon-buttons handle"
+                :title="msg('dragRow')"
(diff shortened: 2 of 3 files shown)


Samq64:
I was thinking of replacing the handle with up and down arrows and moving the X to the top right but this also works.

Samq64 (approved):

DNin01:
> I was thinking of replacing the handle with up and down arrows

The reason I tried out a spin button is so that the UI wouldn't have to change depending on whether you were using a keyboard or a mouse—it was a creative solution.

mxmou:
> I don't know if this is usually how moving items works, but it'll do.

I think it would be a good idea to compare this implementation with a few other apps and websites - if there's an established pattern, we should follow it instead of inventing something else.

For example, in the Language & region page in Windows settings, the drag handles don't get keyboard focus. Instead, there's a dropdown menu with "move up" and "move down" options:
image

DNin01:
A pair of arrow buttons might work better with older browsers which don't support ARIA.

mxmou:
I think the handle should still be available to mouse users, even arrow buttons are added. Perhaps like this:

image

DNin01:
@mxmou, could you open a PR with that design implemented but with the up and down arrow buttons on the left, above and below the drag handle respectively? And maybe change the mouse cursor for the drag handle to a hand.

DNin01:
Thanks! Superseded.

Samq64:
> ` ` Navigate between addons with the arrow keys

By the way what I meant here was [DuckDuckGo](https://duckduckgo.com/?q=scratch+addons)'s arrow key navigation but I don't think it is worth implementing in our case any more.

Issue #8563, opened 2025-09-15
New Addon: Pixel Art tools


Joeclinton1:
### Why this would be helpful

People tend to do their pixel art projects outside of scratch and then import them back in because of poor support. We could make Scratch a quite a bit more useful for pixel art with some small additions to the bitmap editor. 

 By reusing scratch's existing bitmap editor and our other addons like onion skinning, this can make it quite useful for people just starting out.

### How this addon works

We would add:

1. A colour palette. It should be editable, and reuse scratch's existing colour picker.
2. A much more zoomed in default canvas with the ability to pick a size. defaulting to square. Canvas has a border and outside the canvas we make it darker, to visually show where the canvas is.
3. Resized checkerboard grid to match the zoomed in canvas
4. Pixel mode button, for the paint brush which when enabled only allows squares to be drawn, of sizes 1,2,3,4 which you can select via existing brush size.

This would pair very well with https://github.com/ScratchAddons/ScratchAddons/issues/4067 but that should be a separate addon.

### Possible alternatives

Use external pixel art editors, but I view this as a stepping stone to introduce kids to pixel art.

Joeclinton1:
Just to say I don't plan on really making this, but I'm making an issue just so I don't forget. I might do it one day.

Joeclinton1:
I finished making it! Please check out PR #8574

PR #8700, opened 2025-12-03
`disable-auto-save`: Allow leaving editor without saving

DNin01:
Previously, when you clicked "See Project Page" with `disable-auto-save` enabled, you would receive a confirmation asking you if you want to return to the project page _and_ save your changes.

This PR adds a custom alert to this addon. It introduces a third option, "Don't save", which returns you to the project page _without_ saving any changes.

Disable autosave addon prompt

### Tests

All cases tested.

### Thanks to…

- @Joeclinton1 for the modal code

### DIFF
--- a/addons-l10n/en/disable-auto-save.json
+++ b/addons-l10n/en/disable-auto-save.json
@@ -1,3 +1,7 @@
 {
-  "disable-auto-save/save-and-leave": "Message by \"disable auto-save\" addon from Scratch Addons: going to the project page will save your changes. Do you want to save the project and leave the editor?"
+  "disable-auto-save/alert-title": "Disable autosave addon",
+  "disable-auto-save/unsaved-changes": "Do you want to save your changes before leaving the editor?",
+  "disable-auto-save/save": "Save",
+  "disable-auto-save/discard": "Don't save",
+  "disable-auto-save/cancel": "Back"
 }
--- a/addons-l10n/en/disable-auto-save.json
+++ b/addons-l10n/en/disable-auto-save.json
@@ -1,5 +1,6 @@
 {
-  "disable-auto-save/alert-title": "\"Disable auto-save\" addon",
+  "disable-auto-save/alert-title": "Exit to Project Page",
+  "disable-auto-save/alert-subtitle": "Message from \"Disable auto-save\" addon",
   "disable-auto-save/unsaved-changes": "Do you want to save your changes before leaving the editor?",
   "disable-auto-save/save": "Save",
   "disable-auto-save/discard": "Don't save",
--- a/addons-l10n/en/disable-auto-save.json
+++ b/addons-l10n/en/disable-auto-save.json
@@ -1,5 +1,5 @@
 {
-  "disable-auto-save/alert-title": "Disable autosave addon",
+  "disable-auto-save/alert-title": "\"Disable auto-save\" addon",
   "disable-auto-save/unsaved-changes": "Do you want to save your changes before leaving the editor?",
   "disable-auto-save/save": "Save",
   "disable-auto-save/discard": "Don't save",
--- a/addons/disable-auto-save/addon.json
+++ b/addons/disable-auto-save/addon.json
@@ -11,10 +11,24 @@
   "credits": [
     {
       "name": "RustingRobot"
+    },
+    {
+      "name": "DNin01",
+      "link": "https://github.com/DNin01"
+    },
+    {
+      "name": "Chrome_Cat",
+      "link": "https://scratch.mit.edu/users/Chrome_Cat/"
     }
   ],
   "dynamicEnable": true,
   "dynamicDisable": true,
+  "userstyles": [
+    {
+      "url": "modal.css",
+      "matches": ["projects"]
+    }
+  ],
   "userscripts": [
     {
       "url": "userscript.js",
--- a/addons/disable-auto-save/userscript.js
+++ b/addons/disable-auto-save/userscript.js
@@ -27,6 +27,7 @@ export default async ({ addon, console, msg }) => {
     });
     modal.container.classList.add("unsavedChangesDialog");
 
+    appendChildWithAttributes(modal.content, "b", { textContent: msg("alert-subtitle") });
     appendChildWithAttributes(modal.content, "p", { textContent: msg("unsaved-changes") });
     const buttonRow = appendChildWithAttributes(modal.content, "div", {
       className: addon.tab.scratchClass("prompt_button-row", { others: "unsavedChangesDialog-actions" }),
@@ -39,6 +40,7 @@ export default async ({ addon, console, msg }) => {
       });
     }
     buttons.save.classList.add(addon.tab.scratchClass("prompt_ok-button"));
+    buttons.save.focus();
 
     // On any button click:
     buttonRow.addEventListener(
@@ -64,6 +66,10 @@ export default async ({ addon, console, msg }) => {
     );
 
     modal.closeButton.addEventListener("click", modal.remove);
+    modal.container.parentElement.addEventListener("click", modal.remove);
+    modal.container.addEventListener("keydown", (e) => {
+      if (e.key === "Escape") modal.remove();
+    });
   }
 
   while (true) {
--- a/addons/disable-auto-save/userscript.js
+++ b/addons/disable-auto-save/userscript.js
@@ -48,7 +48,7 @@ export default async ({ addon, console, msg }) => {
       // All buttons close the modal
       modal.remove();
 
-      if (selection === "discard") {
+      if (e.target.value === "discard") {
         // Mark changes as saved to avoid a save-on-navigation
         setProjectChanged(false);
         // Wait for the state change to register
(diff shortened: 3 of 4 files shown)


addons/disable-auto-save/addon.json
DNin01, line 2: Could we also rename the addon from "Disable auto-save" to "Disable autosave" for grammatical reasons?

addons/disable-auto-save/userscript.js
```diff
@@ -1,13 +1,67 @@
...
+        if (!e.target.matches("button")) return;
+        // All buttons close the modal
+        modal.remove();
+
+        if (e.target.value === "discard") {
+          // Mark changes as saved to avoid a save-on-navigation
+          addon.tab.redux.dispatch({
+            type: "scratch-gui/project-changed/SET_PROJECT_CHANGED",
+            changed: false,
+          });
```
mxmou: Setting `changed` to false will disable the confirmation message that normally appears when closing the page. You should set it back to true after switching to the project page.

mxmou (changes requested):

addons/disable-auto-save/addon.json
mxmou: What are the reasons?

DNin01: I'm not actually entirely sure, but we can save this for another PR.

addons/disable-auto-save/userscript.js
DNin01: > Setting `changed` to false will disable the confirmation message that normally appears when closing the page. You should set it back to true after switching to the project page.

Done. (c45a617ecbc64be6917a457f189cb41d21e01fef)

Now, if you try to navigate, reload, or close the tab, you'll get the browser confirmation message.

@mxmou

addons-l10n/en/disable-auto-save.json
```diff
@@ -1,3 +1,7 @@
 {
-  "disable-auto-save/save-and-leave": "Message by \"disable auto-save\" addon from Scratch Addons: going to the project page will save your changes. Do you want to save the project and leave the editor?"
+  "disable-auto-save/alert-title": "\"Disable auto-save\" addon",
```
mxmou: The modal title usually summarizes what the modal is for - it could say "Save Project", for example. The addon that created the message could be mentioned in the modal body instead:

Image

mxmou (approved):

Joeclinton1:
Will review if someone tests #8563 thanks

addons/disable-auto-save/userscript.js
```diff
@@ -1,13 +1,79 @@
...
+        if (!e.target.matches("button")) return;
+        // All buttons close the modal
+        modal.remove();
+
+        if (e.target.value === "discard") {
+          // Mark changes as saved to avoid a save-on-navigation
+          setProjectChanged(false);
+          await new Promise((resolve) => setTimeout(resolve, 0));
+          document.querySelector(PAGE_BUTTON_SELECTOR).click();
+          await new Promise((resolve) => setTimeout(resolve, 0));
```
Joeclinton1: this setTimeout promise thing is weird. Explain why you had to use it in a comment.

addons/disable-auto-save/userscript.js
```diff
@@ -1,13 +1,79 @@
...
+          setProjectChanged(true);
+        }
+        if (e.target.value === "save") {
+          document.querySelector(PAGE_BUTTON_SELECTOR).click();
+        }
+      },
+      { bubble: true }
```
Joeclinton1: This is not a valid option for addEventListener (see https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener), bubbling is default behaviour so can be removed.

addons/disable-auto-save/userscript.js
```diff
@@ -1,13 +1,79 @@
...
+  const PAGE_BUTTON_SELECTOR = "[class*='community-button_community-button_']";
+
   addon.tab.redux.initialize();
   addon.tab.redux.addEventListener("statechanged", ({ detail }) => {
     if (addon.self.disabled || detail.action.type !== "timeout/SET_AUTOSAVE_TIMEOUT_ID") return;
     clearTimeout(detail.next.scratchGui.timeout.autoSaveTimeoutId);
     console.log("Pending autosave prevented.");
   });
 
+  function showSaveAlert() {
```
Joeclinton1, line 23: This modal implementation looks great and works well! However, as you credited it shares many similarities with `asset-conflict-dialog` addon's modal.

 Since we now have two addons using this same modal pattern (createModal + button row + scratch
  classes), would it make sense to abstract this into a shared utility or add it to the `addon.tab`       
  API? This would ensure code remains DRY, is consistent and makes future similar modals easier.
  
   Something like:
   ```js
  addon.tab.createPromptModal(title, {
    subtitle: "...",
    message: "...",
    buttons: [
      { value: "cancel", label: msg("cancel") },
      { value: "save", label: msg("save"), primary: true }
    ],
    onButtonClick: (value) => { /* handle */ }
  });
  ```

Joeclinton1 (changes requested):

Joeclinton1:
Tested the features and it works well, but have requested some changes.

addons/disable-auto-save/userscript.js
DNin01: Well, is there, like, a way to wait for the Redux state to change instead of using timeouts here?

These timeouts were just used to wait (for one JS event cycle) for the project-changed state to update before attempting to click the "See Project Page" button, or else the project would save anyway. [`no-sprite-confirm` also uses a similar timeout](https://github.com/ScratchAddons/ScratchAddons/blob/master/addons/no-sprite-confirm/userscript.js).

Joeclinton1: If you could just add a comment, it's fine.

DNin01: How about two?
```js
setProjectChanged(false);
await new Promise((resolve) => setTimeout(resolve, 0)); // wait
document.querySelector(PAGE_BUTTON_SELECTOR).click();
await new Promise((resolve) => setTimeout(resolve, 0)); // wait
setProjectChanged(true);
```

addons/disable-auto-save/userscript.js
DNin01: Yes! There are some other things I want to improve in the modal API, so should I open a separate issue/PR for these additional changes?

addons/disable-auto-save/userscript.js
Joeclinton1: That comment is not very informative. We already know that it is for waiting, but the comment should say why you need to wait.

addons/disable-auto-save/userscript.js
Joeclinton1: I personally would do it in the same PR, otherwise you're making technical debt that is to be resolved later. @Samq64 What do you think?

Samq64: I think either way is fine. `animated-thumb`, `studio-followers` and `mediarecorder` could also work with your example. `asset-conflict-dialog` uses a different class for non-primary buttons compared to other addons though.

`blocks2image` has button icons and one that is left-aligned. Does it make sense to expand the API to allow that?

DNin01: Does `animated-thumb` still have a modal after the... you know, the Scratch update?

Anyway, another option is I could open a PR that extends the API and then this PR would be pending until it is finished.

Samq64: > Does animated-thumb still have a modal

Yes, but only after the upload either succeeds or fails. The only thing left in the button row now is the close button.

addons/disable-auto-save/userscript.js
DNin01: How about this?
```js
// Mark changes as saved to avoid a save-on-navigation
setProjectChanged(false);
// Wait for the state change to register
await addon.tab.redux.waitForState((state) => state !== "scratch-gui/project-changed/SET_PROJECT_CHANGED");
document.querySelector(PAGE_BUTTON_SELECTOR).click();
// Then we can mark it as changed again
setProjectChanged(true);
```
A few comments were added to this updated code, which waits for a condition instead of a timeout, so the code also explains itself more.

Joeclinton1 (approved):
Gonna approve as it works and adheres to the two changes I suggested apart from the idea of making a shared prompt modal addon api, but under the condition that we later make a pr to do the shared modal addon api.

DNin01:
> under the condition that we later make a pr to do the shared modal addon api

I'm already working on it.

addons/disable-auto-save/userscript.js
mxmou: We already have a method (`addon.tab.confirm()`) that creates a modal with a message and a button row. Maybe that could be extended to allow customizing the list of buttons.

For modals that have a button row but aren't simple confirmations (such as `studio-followers` and `mediarecorder`), `createModal()` could have an option to specify a list of buttons, or there could be a new `createButtonRow()` method that would return a `<div>` that you could add to the modal yourself. The advantage of the latter is that the addon could choose where exactly in the DOM to insert it.

DNin01: There sure is a bit of room for general improvement in the modal API.

WorldLanguages:
image

Reminds me of this 😄

PR #8701, opened 2025-12-05
dark-www: style membership features

mxmou:
### Changes

Adds styles for new Scratch Membership features on the website:
* A "Member" label. This will be shown next to members' usernames in places where there's no profile picture. Elsewhere, the profile picture will have cat ears.
  image
* A new FAQ section on the Membership page.

### Tests

Tested locally on Edge.

### DIFF
--- a/addons/dark-www/experimental_scratchwww.css
+++ b/addons/dark-www/experimental_scratchwww.css
@@ -526,7 +526,7 @@ p.callout,
 .banner-outer:not(.banner-danger) .banner-button /* Share */ {
   background-color: #ffab1a;
 }
-.membership-label-text  {
+.membership-label-text {
   color: var(--darkWww-button-text);
 }
 .row .col-sm-9 input[type="radio"]:checked::after {


Joeclinton1:
I haven't seen a single membership badge, is this what it looks like?

CST1229:
> I haven't seen a single membership badge, is this what it looks like?

i think it's what they will look like once they're added (since they're not yet)

mxmou:
> i think it's what they will look like once they're added (since they're not yet)

Indeed - I should've clarified that in the PR description. Members' profile pictures will have cat ears, and in places where the picture isn't shown, a badge will be added to the username instead.

Joeclinton1:
> > i think it's what they will look like once they're added (since they're not yet)
> 
> Indeed - I should've clarified that in the PR description. Members' profile pictures will have cat ears, and in places where the picture isn't shown, a badge will be added to the username instead.

Guess I can't get membership now, because two sets of cat ears would be a bit strange haha.

DNin01 (approved):
Untested

PR #8702, opened 2025-12-05
Ignore pause shortcut in input fields

### LINKED ISSUE #8693: "Pause button" addon's shortcut prevents keyboard character input.
### Description

As the title says, the addon "Pause button" adds a shortcut to pause and unpause the project and sets this to "Alt + x" or "Option + x" on MacOS. On many keyboard layouts, the alt / option key is used to input alternative characters, (For example, on my keyboard "Option + x" prints the letter ʒ, and on the default MacOS layout, prints ≈.) but when this addon is enabled, these characters cannot be input, as the project will pause / unpause instead of the character being printed, leaving the text field unchanged. This applies to both the shifted and unshifted variants.

Ideally, either a different shortcut should be used here or the program should be able to recognize whether the user is in a text field or not and enable / disable the shortcut when appropriate.

### Steps to reproduce

1. Enable "Pause button."
2. Input option + x.

_No response_

### Environment

Browser: Firefox 145.0.1
Scratch Addons version: 1.44.3
Operating system: MacOS Tahoe 26.1

DNin01:
Resolves #8693

### Changes

Allows <kbd>option</kbd><kbd>x</kbd> to type a special character in an input box in macOS instead of pausing/resuming the project without typing anything.

### Reason for changes

See issue

### Tests

In theory, it works.

> [!NOTE]
> On Windows `e.key.toLowerCase() === "x"`, which is on the left-hand side of the logical OR operator, evaluates to true so the behavior doesn't change. In other words, the updated behavior is exclusive to macOS.

### DIFF
--- a/addons/pause/userscript.js
+++ b/addons/pause/userscript.js
@@ -21,11 +21,12 @@ export default async function ({ addon, console, msg }) {
   document.addEventListener(
     "keydown",
     function (e) {
+      if (addon.self.disabled) return;
       // e.code is not enough because that corresponds to physical keys, ignoring keyboard layouts.
       // e.key is not enough because on macOS, option+x types ≈ and shift+option+x types ˛
       // e.keyCode is always 88 when pressing x regardless of modifier keys, so that's how we'll handle macOS.
       // Because keyCode is deprecated we'll still check e.key in case keyCode is not as reliable as we think it is
-      if (e.altKey && (e.key.toLowerCase() === "x" || e.keyCode === 88) && !addon.self.disabled) {
+      if (e.altKey && (e.key.toLowerCase() === "x" || (e.keyCode === 88 && !e.target.matches(":read-write")))) {
         e.preventDefault();
         e.stopImmediatePropagation();
         setPaused(!isPaused());


DNin01:
@Joeclinton1, do you think this is reasonable?

Joeclinton1:
> @Joeclinton1, do you think this is reasonable?

Yes, seems reasonable. I see no exception where a user might want to pause unpause whilst also typing. The typing special characters functionality takes priority over a shortcut which has a gui alternative.

Joeclinton1:
I can't approve because I don't have mac for testing.

DNin01:
We need @GarboMuffin 😄

addons/pause/userscript.js
```diff
@@ -21,11 +21,12 @@ export default async function ({ addon, console, msg }) {
   document.addEventListener(
     "keydown",
     function (e) {
+      if (addon.self.disabled) return;
       // e.code is not enough because that corresponds to physical keys, ignoring keyboard layouts.
       // e.key is not enough because on macOS, option+x types ≈ and shift+option+x types ˛
       // e.keyCode is always 88 when pressing x regardless of modifier keys, so that's how we'll handle macOS.
       // Because keyCode is deprecated we'll still check e.key in case keyCode is not as reliable as we think it is
-      if (e.altKey && (e.key.toLowerCase() === "x" || e.keyCode === 88) && !addon.self.disabled) {
+      if (e.altKey && (e.key.toLowerCase() === "x" || (e.keyCode === 88 && !e.target.matches(":read-write")))) {
```
mxmou, line 29: Some other addons already use different ways to detect an input being focused, for example [checking the element's `tagName`](https://github.com/ScratchAddons/ScratchAddons/blob/4c4996b/addons/find-bar/userscript.js#L147). Using this selector is probably more reliable - maybe the other places should be updated to do the same.

WorldLanguages: I didn't know about `:read-write`. Interesting!

PR #7833, opened 2024-09-24
editor-theme3: Add 2.0 Style Block Bevel Filter

ZenithRogue:
Resolves #7789

### Changes

Adds an option to editor-theme3 for a Scratch 2.0 styled bevel filter
image
image

### Reason for changes
I love the Scratch 2.0 block style

### Tests
Everything seems to work just fine. Relatively minor changes tbh. Edgecases for drag and drop block inputs has been resolved since the initial CSS I showed off in the initial issue.

PR #8706, opened 2025-12-07
Add "Compact dropdowns" addon

sandetsun:
image
(Coshnaut is me)

### Changes

Modifies the layout of the dropdown input menus to be more compact and similar to Scratch 2.0
image
image

Compatible with the dropdown search addon
image

### Options
**Enable shadows:**
Gives a drop-shadow to the menu
image
With dropdown search, the search box looks like a 2.0 input
image

**Enable check-marks:**
Enables or disables the presence of checkmarks
On:
image
Off:
image

### Reason for changes

The current block dropdown menu is quite large, and might not use the space very well.
Also, the options to enable the shadows would work nicely for a user looking to make their editor look and function like Scratch 2.0's

### Tests

I have tested on Google Chrome 142.

### DIFF
--- a/addons/scratch2-dropdowns/addon.json
+++ b/addons/scratch2-dropdowns/addon.json
@@ -17,7 +17,7 @@
       "type": "boolean",
       "default": false
     },
-	{
+    {
       "name": "Enable check-marks",
       "id": "checkmark",
       "type": "boolean",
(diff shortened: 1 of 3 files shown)


Samq64:
Related: #7833

sandetsun:
Im sorry I've forgotten to add the labels, and I didn't mean to edit that other json file

mxmou:
I like this, but I'm not sure if it makes sense as a standalone addon - someone who enables it will probably want the blocks to be 2.0-styled as well, not just the dropdowns. Perhaps this and #7833 could be combined into a single addon.

sandetsun:
> I like this, but I'm not sure if it makes sense as a standalone addon - someone who enables it will probably want the blocks to be 2.0-styled as well, not just the dropdowns. Perhaps this and #7833 could be combined into a single addon.

I see what you mean but I disagree, #7833 is already proposed as a option inside of another addon. Bundling this inside a toggle on an unrelated addon would feel out of place, and I think this already changes enough to be alone

Samq64:
> https://github.com/ScratchAddons/ScratchAddons/pull/7833 is already proposed as a option inside of another addon

Yes but I think splitting it into a new addon along with this would be a better fit.

PR #8663, opened 2025-10-23
Webpages: Improve screen reader navigation experience

DNin01:
### Changes

These changes are the start of improvements to screen reader accessibility.

Screen reader users will notice the following enhancements:
- Landmarks have been added.
- Semantics have been added to the addon list.
- More icons have been assigned labels or alt text.
- Groups of related content and important region markers have been labeled.

Let me know if I missed anything.

### Commit summary

- **Initial changes**
  1. 🏷️ [**Add labels to icons**](https://github.com/ScratchAddons/ScratchAddons/pull/8663/commits/c115e22e02b5195dc67b874732739030edd9238b) — adds either alt text or tooltips to several graphics.
  2. 🏳️ [**Label groups and landmarks**](https://github.com/ScratchAddons/ScratchAddons/pull/8663/commits/35ac77a8454dfdc998a5bf5f85a9e9b300184ad4) — adds landmarks to the banner, navigation, and main content, and adds ARIA attributes to blocks of content like addon bodies, addon group separators, and info boxes.
  3. 🫧 [**Popup updates**](https://github.com/ScratchAddons/ScratchAddons/pull/8663/commits/eb83ba1ada50323ba95aca966593cb0a816533d8) — applies the above changes to the extension's popup.
  4. 🔖 [**Label blocks and inputs**](https://github.com/ScratchAddons/ScratchAddons/pull/8663/commits/09c738b21e93c37328dc525fa4291b4d48bf85b4) — adds ARIA labels to addon checkboxes/inputs, setting description tooltip containers, and "new features" message boxes.
  5. 📂 [**Semantics in addon list**](https://github.com/ScratchAddons/ScratchAddons/pull/8663/commits/dfaca92c8cc63154aa973de001e1ac999c2f7fda) — turns some elements in the addon listing page into headings.
- **Revisions since pull request open**
  1. 🔘 [**Fix addon switch description**](https://github.com/ScratchAddons/ScratchAddons/pull/8663/commits/cb3302dfbf894fd32d726187afceb0abc4c1dafe) — corrects a mistake in a previous change that added descriptions to addon toggle switches.
  2. 📍 [**Change addon name labels to level 3 headings**](https://github.com/ScratchAddons/ScratchAddons/pull/8663/commits/92d783a7954c46ff2be5ed89a4f1476d19000fd9) — changes addon names to headings as well.
[...]

PR #8718, opened 2025-12-13
Webpages: Add icon descriptions

DNin01:
Moved from #8663

### Changes

Adds either alt text or tooltips to icons in the settings page which didn't have them before that convey extra information not otherwise available to screen reader users. These icons include:
- Addon type icons
- The `x` button on the search bar
- The icons next to addon information blocks

It's also just nice to have a label on some icons.

Also, the icons in the More Settings menu now have `aria-hidden="true"` because they don't convey any useful information for screen reader users.

### Tests

Tested with:
- Edge 143
- Windows Narrator

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -128,6 +128,24 @@
   "easterEggs": {
     "message": "Easter Eggs"
   },
+  "editorFeature": {
+    "message": "Editor feature"
+  },
+  "playerFeature": {
+    "message": "Project player feature"
+  },
+  "websiteFeature": {
+    "message": "Website feature"
+  },
+  "themeAddon": {
+    "message": "Theme"
+  },
+  "popupFeature": {
+    "message": "Extension popup feature"
+  },
+  "easterEgg": {
+    "message": "Easter egg"
+  },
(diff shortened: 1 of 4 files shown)


_locales/en/messages.json
```diff
@@ -128,6 +128,24 @@
...
   },
+  "editorFeature": {
+    "message": "Editor feature"
+  },
+  "playerFeature": {
+    "message": "Project player feature"
+  },
+  "websiteFeature": {
+    "message": "Website feature"
+  },
```
mxmou, line 139: Could the existing strings "For editor", "For project player", and "For website" be used?

DNin01: I think noun phrases work better as icon descriptions, but we could reuse _these_ strings on the "For editor", "For website", and "For project player" badges—if they are even necessary after adding icon tooltips showing the same information.

mxmou: The badges are still needed because they're shown on theme addons, which have the theme icon, so the badge is the only indication of which part of Scratch the addon affects.

mxmou (approved):

PR #8724, opened 2025-12-14
`fix-costume-drag`: Delta time

DNin01:
### Changes

Updates @The2AndOnly's addon (again):
- Added delta timing to ensure the auto scroll speed remains consistent between different refresh rates.
  - Maximum delta is 0.1 (1/10 sec.)
- Also slightly sped up auto scrolling (changed `SPEED_PRESETS`) so you can move costumes faster.

### Videos

| Before | After |
|---|---|
| <video src=https://github.com/user-attachments/assets/c935cc90-5223-48a4-b3a7-17c64a36b89c> | <video src=https://github.com/user-attachments/assets/4218dd2c-8e36-42e6-a4af-ed641b52bf79> |

<sup>Recording 60 FPS and screen refreshing at 120 Hz in both. Note that the `fps` addon is a slightly inaccurate counter.</sup>

### Tests

Tested on Edge 143 using both 60 Hz and 120 Hz settings.

### Known issues

- Scrolling will begin to slow down if the project FPS drops below 10 (you can notice this in the video). This is caused by the delta maximum.
- Measurements only occur every drag update (including from auto scrolling), so if your mouse approaches the edge slowly, delta will have maxed out which causes a jump. However, this isn't a noticeable inconvenience under normal use, especially since delta is limited as mentioned before.

### DIFF
--- a/addons/fix-costume-drag/userscript.js
+++ b/addons/fix-costume-drag/userscript.js
@@ -133,8 +133,11 @@ export default async function ({ addon, console }) {
 
           // Auto scroll
           const edgeSize = 30; // Distance from the top/bottom to trigger scroll
-          const delta = performance.now() - lastTime; // Calculate delta time
-          const scrollAmount = (scrollSpeed * Math.min(delta, 100)) / 15;
+          const deltaTime = performance.now() - lastTime;
+          // Drag updates happen whenever you move your mouse or the list is scrolled, so if your mouse
+          // approaches the auto-scroll zone very slowly, the rate of drag updates will decrease which
+          // means delta time will be greater. That's why it is capped at 100ms:
+          const scrollAmount = (scrollSpeed / 15) * Math.min(deltaTime, 100);
           if (this.props.dragInfo.currentOffset.y < containerRect.top + edgeSize) {
             scrollContainer.scrollTop -= scrollAmount;
           } else if (this.props.dragInfo.currentOffset.y > containerRect.bottom - edgeSize) {


addons/fix-costume-drag/userscript.js
```diff
@@ -132,13 +133,16 @@ export default async function ({ addon, console }) {
 
           // Auto scroll
           const edgeSize = 30; // Distance from the top/bottom to trigger scroll
+          const delta = performance.now() - lastTime; // Calculate delta time
```
mxmou: This comment isn't very useful - I'd prefer a short explanation of why the 100 ms maximum on the next line is needed.

mxmou (approved):

addons/fix-costume-drag/userscript.js
```diff
@@ -132,13 +133,16 @@ export default async function ({ addon, console }) {
...
+          const scrollAmount = (scrollSpeed * Math.min(delta, 100)) / 15;
```
DNin01: > @mxmou: This comment isn't very useful - I'd prefer a short explanation of why the 100 ms maximum on the next line is needed.

Is this better?

```suggestion
          const deltaTime = performance.now() - lastTime;
          const scrollAmount = (scrollSpeed / 15) * Math.min(deltaTime, 100); // Maximum delta time: 100ms
```

The2AndOnly: But why

DNin01: > But why

Why that comment was unnecessary or why this comment was important?

mxmou: > > But why
> 
> Why that comment was unnecessary or why this comment was important?

Why is the maximum delta 100 ms - your new comment still doesn't explain that.

DNin01: > Why is the maximum delta 100 ms - your new comment still doesn't explain that.

What would you say, then?

mxmou: I don't know why you added the maximum - that's why I think there should be a comment clarifying it. I assume it's related to the fact that, as you said in the description, "measurements only occur every drag update".

DNin01: So, because the delta time calculation is only done within the drag-update function, that calculation is essentially the time difference between now and the time of the last drag update. Drag updates happen whenever you move your mouse or the list is scrolled, so if your mouse approaches the auto-scroll zone very slowly, the rate of drag updates will decrease which means delta time will be greater.

This manifests as jump-scrolls. Without capping delta time at 0.1s, this happens:

https://github.com/user-attachments/assets/9f1b69ce-48b1-47d9-ade2-dc25cb15984f

There are definitely other ways to do the calculation that would solve the issue, but still, it's generally a good safeguard to have a maximum amount of time to move forward. In games, for example, you don't want your character to uncontrollably zip ahead multiple seconds just because your screen froze for a moment. The same may apply for auto-scrolling if you're playing a Scratch project that slows to a crawl. Auto-scrolling should be consistent as long as it is controllable.

mxmou: I suggest adding this part as a comment in the code:

> Drag updates happen whenever you move your mouse or the list is scrolled, so if your mouse approaches the auto-scroll zone very slowly, the rate of drag updates will decrease which means delta time will be greater.

addons/fix-costume-drag/userscript.js
DNin01: OK, I updated it.

PR #8742, opened 2025-12-23
cat-blocks: disable Scratch's Theme submenu

mxmou:
### Changes

* Disables the member-only Theme submenu if the Cat blocks addon is enabled.
* Fixes editor-theme3 not disabling the Color Mode submenu if the user was a Scratch member.
* Fixes links to settings for specific addons (`#addon-<addon ID>`). They were broken by #6104.

image

### Tests

Tested on Edge and Firefox. I set the `has_active_membership` session flag to true to make the submenu appear.

### DIFF
--- a/addons/cat-blocks/userscript.js
+++ b/addons/cat-blocks/userscript.js
@@ -433,7 +433,7 @@ export default async function ({ addon, console, msg }) {
   const setTheme = async (newTheme) => {
     if (!addon.tab.redux.state) return;
     const currentTheme = getTheme();
-    if (newTheme != currentTheme) {
+    if (newTheme !== currentTheme) {
       addon.tab.redux.dispatch({
         type: "scratch-gui/settings/SET_THEME",
         theme: newTheme,


DNin01:
What should we do about our Cat Blocks addon now that they added cat blocks as a membership benefit?

I think we can still _have_ one—I don't think they'll mind because cat blocks are just a fun bonus gift members get, not a selling point (not that they're deliberately trying to paywall anything, and they can't lock down open source material anyway), but maybe—IDK if I'm overthinking—we could at the least make our addon easter-egg only?

Or change it to dog blocks 😆

DNin01:
I'll be able to test this genuinely once my Christmas wish comes true. 💫

addons/cat-blocks/style.css
```diff
@@ -0,0 +1,15 @@
+@import url("../../libraries/common/cs/setting-managed-by-sa.css");
+
+.sa-theme-submenu li:nth-child(1),
+.sa-theme-submenu li:nth-child(2) {
```
DNin01: Hopefully no list items are added later (or this wouldn't work).

mxmou: I copied this from theme3. It could probably be changed to something like `li:not(:last-child, :nth-last-child(2))` to be more reliable.

medlans:
> What should we do about our Cat Blocks addon now that they added cat blocks as a membership benefit?
> 
> I think we can still _have_ one—I don't think they'll mind because cat blocks are just a fun bonus gift members get, not a selling point (not that they're deliberately trying to paywall anything, and they can't lock down open source material anyway), but maybe—IDK if I'm overthinking—we could at the least make our addon easter-egg only?
> 
> Or change it to dog blocks 😆

Maybe keep it and add more "themes" like that XD (okay I feel like cat blocks would be better called mode like in April Fools because I feel like theme implies stuff like custom colors or dark mode, but whatever)

addons/cat-blocks/userscript.js
DNin01, line 534: This is not a big deal, but the Watch Mouse Cursor feature doesn't work when cat blocks are enabled in both Scratch and this addon.

Steps to reproduce:
1. Activate Scratch Membership features
2. Disable the `cat-blocks` addon
3. Switch the editor theme to "Cat Blocks"
4. Enable the `cat-blocks` addon

mxmou: I'll try to disable Scratch's cat blocks when the addon is enabled. That's probably the simplest solution.

Samq64: `Blockly.setTheme(Blockly.Themes.CLASSIC)` should handle that.

DNin01: > I'll try to disable Scratch's cat blocks when the addon is enabled. That's probably the simplest solution.

…and when the user turns off the Cat Blocks addon, always ensures the blocks go back to normal.

addons/cat-blocks/userscript.js
```diff
@@ -428,7 +429,34 @@ export default async function ({ addon, console }) {
...
+        scratchTheme = addon.tab.redux.state.scratchGui.settings.theme;
+        addon.tab.redux.dispatch({
+          type: "scratch-gui/settings/SET_THEME",
+          theme: "default",
+        });
+      } else {
+        addon.tab.redux.dispatch({
+          type: "scratch-gui/settings/SET_THEME",
+          theme: scratchTheme,
+        });
```
DNin01: This is kind of a nitpick, but disabling a dynamic addon usually reverts its effects on the interface, so should the Theme setting be automatically reset to "Default" whenever the addon is disabled? Currently, the user's theme choice is always preserved after the addon is disabled... which _is_ fine; users would probably realize that they also have to use the Theme menu to turn off cat blocks, but... just to be extra helpful. 😄

DNin01 (commented):
Now the cat blocks watch the mouse cursor as expected, but for some reason I also have to leave and re-enter the editor (or reload the page or change Scratch's "Theme" setting) whenever I toggle the addon for the cats to appear or disappear.

addons/cat-blocks/userscript.js
mxmou: Maybe it could show a message reminding the user to change the Scratch theme as well - I don't think addons should change settings without asking the user first.

addons/cat-blocks/userscript.js
```diff
@@ -428,7 +429,34 @@ export default async function ({ addon, console }) {
...
+      // wait for new workspace to render
+      await new Promise((resolve) => {
+        const oldInject = Blockly.inject;
+        Blockly.inject = function (...args) {
+          Blockly.inject = oldInject;
+          resolve();
+          return oldInject.call(this, ...args);
+        };
+      });
+    }
```
DNin01, line 450: For me, this promise never resolves, meaning I have to switch sprites or views or reload the page for the cat blocks to appear or disappear after dynamically enabling or disabling the addon. Otherwise, it seems to work as expected.

<sub>scratch.mit.edu on Edge 143</sub>

DNin01 (changes requested):

addons/cat-blocks/userscript.js
mxmou: I'll push a fix soon. The issue is that if the Scratch theme was already `default`, no new workspace is created and `inject()` is never called.

addons/cat-blocks/userscript.js
mxmou: I've added this message:

image

addons/cat-blocks/userscript.js
DNin01: Wait, I'm sorry. I didn't pull 82f62d11f025f09f61e6cc9e8413d33b00448f10 before reviewing.

> I'll push a fix soon. The issue is that if the Scratch theme was already `default`, no new workspace is created and `inject()` is never called.

No need, this snippet is inside an if-`newTheme !== currentTheme` block, so it doesn't even enter the `await` statement in that case... right?

DNin01:
Wow, is this the first time an addon has ever displayed a banner message in Scratch's style?

DNin01 (approved):

addons/cat-blocks/userscript.js
mxmou: > Wait, I'm sorry. I didn't pull [82f62d1](https://github.com/ScratchAddons/ScratchAddons/commit/82f62d11f025f09f61e6cc9e8413d33b00448f10) before reviewing.

The PR timeline is sometimes confusing because it shows when a commit was made locally, not when it was pushed to GitHub. I pushed this commit after your review.

mxmou:
> Wow, is this the first time an addon has ever displayed a banner message in Scratch's style?

Yes, but the code for that will be useful for quite a few addons, especially those that currently use `window.alert()`. I'll move it to the addon API in a future PR.

Samq64:
The added elements seem to be duplicated and overflow the menu:

PR #8743, opened 2025-12-24
feat: Add new settings page with dedicated HTML, CSS, JavaScript, and…

chi20253032:
… localization.

Resolves #


Samq64:
This seems to add a settings profile manager but it does not work and I don't see a real advantage over exporting and importing settings files or using another browser profile. This PR also changes the logo and code formatting unnecessarily.

CST1229:
> This seems to add a settings profile manager but it does not work and I don't see a real advantage over exporting and importing settings files or using another browser profile. This PR also changes the logo and code formatting unnecessarily.

i wonder if this was a pr meant for a fork but submitted here by accident (for some reason this happens way too much on github)

PR #8758, opened 2026-01-02
Local project share addon

### LINKED ISSUE #8757: [New Addon] Share Scratch projects directly from the editor over the local network
### Why this would be helpful

The original idea behind this addon was to make it easier to send projects directly from the Scratch editor to other devices on the local network.

A concrete use case is ScratchEverywhere! (https://github.com/ScratchEverywhere/ScratchEverywhere), where projects often need to be transferred to another device or console. With this addon, there is no need to open a separate application (for example for FTP), or to remove and copy files via an SD card.

It’s mostly a convenience / nice-to-have feature, but it can save a few steps and make workflows smoother. Since it simply sends the project over HTTP, it could also be useful for other local setups and tools, not just ScratchEverywhere!.

### How this addon works

The addon adds a “Local Share” button to the File menu in the Scratch editor.

When clicked, it opens a small modal where the user can enter the recipient’s IP address and port. The current project is then exported and sent directly over HTTP to the target device. A progress bar is shown during the upload so the user can see the transfer status.

### Possible alternatives

I looked through existing addons and did not find anything with similar functionality. If there already is an addon that overlaps with this idea, I would be open to adapting or merging the concept instead.

### Additional context

This is my first Scratch Addon. I tried to follow existing addon patterns and based the UI styling on the Media Recorder addon. I’m very open to feedback, suggestions, or alternative approaches.

Br0tcraft:
Resolves #8757

### Changes

This PR adds a small addon that lets you share Scratch projects directly from the editor.

It adds a “Local Share” button to the File menu, which opens a simple modal where you can enter the recipient’s IP address and port. The current project is then exported and sent to another device over HTTP, with a progress indicator shown during the upload.

I tried to keep the addon as general as possible so it can be used in different local network setups. I looked through existing addons and didn’t find anything with similar functionality. If there *is* overlap with another addon, I’m happy to adapt or adjust this to avoid duplication.

This is my first Scratch Addon, so I mainly followed the structure of existing addons and based the UI on the “Media Recorder” addon. I’m not fully sure if everything is done in the best possible way, but it works and follows existing patterns as closely as I could.

### Reason for changes

The original idea behind this feature was to make it easier to send projects directly from the Scratch editor to ScratchEverywhere!  
(https://github.com/ScratchEverywhere/ScratchEverywhere)

Without this, transferring projects often requires a separate app (for example FTP) or manually copying files via an SD card. This addon is mostly a convenience / nice-to-have feature that removes a few of those steps.

The feature does not necessarily have to be added, as I will also integrate it directly into the SE! editor.
(https://editor.scratchbox.dev/editor.html).  
However, it would be really nice if this functionality were also available through Scratch Addons for general use.

### Tests

Tested in Chrome and Firefox.
(In Forefox, however, you have to disable `security.mixed_content.block_active_content` for HTTP requests.)

I’m keeping this as a draft PR for now so I can get feedback on improvements, better ways to handle things, or anything I might have missed. Open to criticism and suggestions!

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -163,6 +163,7 @@
   "fix-costume-drag",
   "recolor-custom-blocks",
   "totally-normal-modes",
+  "local-share",
 
   "// NEW ADDONS ABOVE THIS ↑↑",
   "// Note: these themes need this exact order to work properly,",
--- a/addons/local-share/addon.json
+++ b/addons/local-share/addon.json
@@ -1,0 +1,38 @@
+{
+  "name": "Local Project Share",
+  "description": "Send Scratch projects to a local device via HTTP.",
+  "tags": ["editor", "editorMenuBar"],
+  "versionAdded": "1.45.0",
+  "enabledByDefault": false,
+  "settings": [
+    {
+      "name": "Host to send to",
+      "id": "host",
+      "type": "string",
+      "default": ""
+    },
+    {
+      "name": "Port to use",
+      "id": "port",
+      "type": "positive_integer",
+      "default": 3000
+    }
+  ],
+  "userscripts": [
+    {
+      "url": "userscript.js",
+      "matches": ["projects"]
+    }
+  ],
+  "userstyles": [
+    {
+      "url": "userstyle.css",
+      "matches": ["projects"]
+    }
+  ],
+  "credits": [
+    {"name": "Br0tcraft"}
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true
+}
--- a/addons/local-share/addon.json
+++ b/addons/local-share/addon.json
@@ -16,6 +16,12 @@
       "id": "port",
       "type": "positive_integer",
       "default": 3000
+    },
+    {
+      "name": "Use HTTPS",
+      "id": "useHttps",
+      "type": "boolean",
+      "default": false
     }
   ],
   "userscripts": [
--- a/addons/local-share/addon.json
+++ b/addons/local-share/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Local Project Share",
-  "description": "Send Scratch projects to a local device via HTTP.",
+  "description": "Adds a \"Share locally\" button to the File dropdown menu, allowing you to send the current project to a local device via HTTP.",
   "tags": ["editor", "editorMenuBar"],
   "versionAdded": "1.45.0",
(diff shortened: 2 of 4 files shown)


Samq64:
> The current project is then exported and sent to another device over HTTP

This is pretty vague. Could you explain the requirements and limitations?

Br0tcraft:
@Samq64 
The addon requires the receiving device to be on the same local network with an http(s) server ready to accept the project. When you click the “Local Share” button, the project is exported as a .sb3 blob and sent via a POST request to http(s)://IP:Port/send, with a progress bar showing upload status. It’s mainly a convenience feature to quickly share projects between devices without manually transferring files or using another app.
(I kept it very simple. It doesn't have many features, and the error handling isn't particularly comprehensive.)

Samq64:
> an http(s) server ready to accept the project.

Doesn't that defeat the convenience aspect? Not everyone knows how to setup an HTTP server or has Python installed.

Br0tcraft:
Of course, setting up a server is some initial effort, and if this feature isn’t added, that’s completely fine. The main idea behind this addon is to enable sending projects directly to a backend integrated within the SE! runtime. In that case, the user wouldn’t need to do anything besides clicking a button in SE! on the console of their choice to receive the project straight from the Scratch editor.
For other use cases or different receivers, a compatible server needs to exist, which requires at least some programming knowledge. Once it’s set up, though, you don’t have to manually transfer the project file every time you make a change via third-party apps, USB sticks, or directly from the SD card.
If helpful, I could also create a small receiver app or script for Windows and Linux to simplify the process.

CST1229:
This feels rather specific (requiring a cooperating application), maybe it could work better as a standalone userscript or browser extension.

Br0tcraft:
Yes, it is quite specific (but I could write the necessary scripts/programs to make setting up a local server as easy as possible to keep it as general as possible, and then link the Code/project in the added addon if allowed). 
My Addon doesn't necessarily have to be integrated into the Scratch addon. I'm not demanding that. 
As already mentioned, this feature will definitely be built into the SE editor. 
My pull request was more of a question about whether something like this could be made directly accessible via the Scratch addon. 
If the addon is too specific, that's perfectly fine. I'll just delete this pr.

WorldLanguages:
Yeah, this is probably too specific to be added. Interesting project though!

PR #8812, opened 2026-01-31
Make sure option element IDs aren't duplicated

### LINKED ISSUE #8749: Only one of repeated addon `select` settings is displayed correctly
### Description

When the same addon appears more than once in the addon list, a visual bug, regressed NOT by #8678, surprisingly, but #8577, affects how its setting values are displayed. This only seems to affect `"select"` settings, where only one occurrence of an identical option will render as highlighted. If the first one is offscreen, changing the setting also auto-scrolls the page to reveal it.

It's notable that new or updated addons always appear twice in the list.

### Steps to reproduce

1. Open the settings page.
2. Enable and change the settings of the `"Improved dragging of costumes and sounds"` addon.

https://github.com/user-attachments/assets/fbed56a7-4a89-4690-a05a-669abd572b20

### Environment

Browser: Edge 143
Scratch Addons version: v1.45.0-pre 751866985acdbb5dc1cf1f22e370983ffcfac03b
Operating system: Windows

DNin01:
Resolves #8749

This works now:

https://github.com/user-attachments/assets/f6393176-b676-4f2b-884a-62b2da6170e4

Tested in Edge 144.

### DIFF
--- a/webpages/settings/components/addon-setting.js
+++ b/webpages/settings/components/addon-setting.js
@@ -80,7 +80,7 @@ export default async function ({ template }) {
         });
       },
       selectOptionId(option) {
-        return `${this.groupId}-${this.selectName}-${option.id}`;
+        return `${this.selectName}-${option.id}`;
       },
       checkValidity() {
         // Needed to get just changed input to enforce it's min, max, and integer rule if the user "manually" sets the input to a value.


webpages/settings/components/addon-setting.js
```diff
@@ -6,7 +6,7 @@ export default async function ({ template }) {
       return {
         noResetDropdown: ["table", "boolean", "select"].includes(this.setting.type),
         tableChild: this.settingPath.length > 1,
-        selectName: `${this.addon._addonId}-${this.settingPath.join("-")}`,
+        selectName: `${this.$parent.groupId}-${this.addon._addonId}-${this.settingPath.join("-")}`,
```
mxmou: I don't think accessing the parent component's prop through `$parent` like this is a good idea. The `<addon-body>` should explicitly pass it to the `<addon-setting>` so that it's easier to see where it's used when reading the code.

DNin01: Do you know how to do that?

mxmou: * Add `groupId` to the `props` array near the top of this file
* Add `:group-id="groupId"` in every place where an `<addon-setting>` component is created (once in `addon-body.html` and once in `addon-setting.html`)
* You can then access it as `this.groupId`

webpages/settings/components/addon-setting.js
```diff
@@ -80,7 +80,7 @@ export default async function ({ template }) {
         });
       },
       selectOptionId(option) {
-        return `${this.selectName}-${option.id}`;
+        return `${this.groupId}-${this.selectName}-${option.id}`;
```
mxmou:
```suggestion
        return `${this.selectName}-${option.id}`;
```
This part isn't needed because the `selectName` already includes the `groupId`.

mxmou (approved):

PR #8828, opened 2026-02-03
`better-featured-project` code cleanup

DNin01:
### Changes

The Profile Page Banner addon has been updated. The code is now easier to read, and the content now loads earlier.

### Reason for changes

Before, the userscript did not have `"runAtComplete": false`, so the featured project image could take several seconds to appear, particularly on profiles with a lot of comments, due to images having to load first.

### Notes

Tested in Edge 144. Works with `live-featured-project`.

Make sure to check the Commits tab to understand the changed code.

@mxmou, you can review this PR if you want—you updated this addon recently.

To help with testing with the membership badge, I've also made [`test-membership`](https://github.com/DNin01/ScratchAddons/tree/test-membership/addons/test-membership).

### DIFF
--- a/addons/better-featured-project/userscript.js
+++ b/addons/better-featured-project/userscript.js
@@ -14,16 +14,14 @@ export default async function ({ addon, console, msg }) {
     if (document.querySelector(".user-content .player .title a").innerText.trim().length) {
       var featuredProjectName = document.createElement("div");
       featuredProjectName.id = "better-featured-project-name";
-
-      const [h2, h3] = [document.createElement("h2"), document.createElement("h3")];
+      const h2 = document.createElement("h2");
+      const h3 = document.createElement("h3");
       h2.innerText = featuredHeading;
       h3.innerText = featuredTitle;
       featuredProjectName.append(h2, h3);
-
       headerText.insertAdjacentElement("afterend", featuredProjectName);
       addon.tab.displayNoneWhileDisabled(featuredProjectName);
     }
-
     boxHead.insertAdjacentElement("afterbegin", document.createElement("a")).id = "better-featured-project-overlay";
     document.getElementById("better-featured-project-overlay").href = featuredLink;
     addon.tab.displayNoneWhileDisabled(document.getElementById("better-featured-project-overlay"));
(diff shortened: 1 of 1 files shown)


addons/better-featured-project/userscript.js
```diff
@@ -3,46 +3,44 @@ export default async function ({ addon, console, msg }) {
...
-          .setAttribute("id", "fpb-name");
-        document.querySelector("#fpb-name").appendChild(document.createElement("h2"));
-        document.querySelector("#fpb-name").appendChild(document.createElement("h3"));
+        headerText.insertAdjacentElement("afterend", document.createElement("div")).id = "fpb-name";
+        document.getElementById("fpb-name").append(document.createElement("h2"), document.createElement("h3"));
         document.querySelector("#fpb-name h2").innerText = featuredHeading;
         document.querySelector("#fpb-name h3").innerText = featuredTitle;
       }
-      if (document.querySelector('#featured-project [data-control="edit"]') !== null) {
+      if (document.querySelector('#featured-project [data-control]')) {
```
mxmou: This is very minor, but I'd keep the `="edit"` part: it has no impact on the specificity or behavior, but I think it makes the selector a little more self-explanatory.

mxmou (changes requested):
Dynamic enable/disable works and some of the code changes look good, but there are also some I disagree with, especially those that look like arbitrary style decisions. Because we don't have a project-wide style guide, I don't see a reason not to stick to the style used by the original addon author unless the change significantly improves the code (makes it more readable, which isn't necessary the same thing as making it shorter or simpler).

In particular, I don't think element IDs should be abbreviated: it makes them easier to write but harder to understand, so it's only worth it if the abbreviation is used in many places. I'm also not convinced that replacing `querySelector()` with `getElementById()` and similar changes are justified.

DNin01:
Pretty sure lint wasn't failing before, but let's disregard that for now.

DNin01:
> @mxmou · there are also some I disagree with, especially those that look like arbitrary style decisions. Because we don't have a project-wide style guide, I don't see a reason not to stick to the style used by the original addon author unless the change significantly improves the code (makes it more readable, which isn't necessary the same thing as making it shorter or simpler).

The large if-then block was replaced with a conditional return statement to reduce the maximum indentation level, but I've reverted some of the other styling changes.

> In particular, I don't think element IDs should be abbreviated: it makes them easier to write but harder to understand, so it's only worth it if the abbreviation is used in many places.

I've reverted the names.

> I'm also not convinced that replacing `querySelector()` with `getElementById()` and similar changes are justified.

I used `getElementById()` where possible because it can be slightly faster than doing a query.

Any other feedback?

mxmou:
> The large if-then block was replaced with a conditional return statement to reduce the maximum indentation level, but I've reverted some of the other styling changes.

I think that's a good change - the logic is easier to follow if there's less nesting.

> I used `getElementById()` where possible because it can be slightly faster than doing a query.

I doubt there's a measurable performance difference unless you're calling in a loop.

mxmou (approved):

PR #8832, opened 2026-02-06
Website themes: better support for deleted and reported comments

### LINKED ISSUE #8727: Deleted Comment Changes Color (Reverts to Light Style even in dark mode)
### Description

A deleted comment should have a dark background, but in studios and projects, deleted comments have a light background (an artifact from light mode). 

### Steps to reproduce

1. Enable Scratch Addons Dark Mode
2. Go to a studio or project you own (this problem does not happen on profile pages)
3. Post a comment 
4. Delete the comment by clicking the "delete" option just above the comment

Image

### Environment

Browser: Microsoft Edge Version 143.0.3650.80 (Official build) (64-bit)
Scratch Addons version: 1.44.4
Operating system: Windows 11 Home 25H2

mxmou:
Resolves #8727

#### Project and studio pages with dark-www

Before:
image

After:
image

#### Profile page with scratchr2

Before:
image

After:
image

(I didn't actually delete or report either of those comments)

### Tests

Tested on Edge with different combinations of addons.

### DIFF
--- a/addons/scratchr2/scratchr2_comments.css
+++ b/addons/scratchr2/scratchr2_comments.css
@@ -123,7 +123,7 @@
   color: var(--darkWww-link-hover, #7854c0);
 }
 #comments .comment .info .time {
-  color: var(--darkWww-box-transparentText, #b3b3b3);
+  color: var(--darkWww-box-transparentText, rgba(0, 0, 0, 0.3));
 }
 #comments .comment .info .reply {
   float: right;


pufferfish101007:
The contrast of '7 days ago' text looks to me as if it is quite a bit worse on the scratchr2 example than it was before. Would it be possible to alter the colour of the text (or the background) to improve this? (I'm not sure how this related to any vanilla scratch theming, if any - if scratch already does this then please ignore)

(However it is much better in the dark-www example!)

mxmou:
The contrast is the same as on vanilla 3.0-styled pages, but I can improve it slightly by using transparent black instead of light gray (so that it will automatically be darker on the red background than on white).

mxmou:
Now it looks like this:
image

PR #8836, opened 2026-02-11
Fix unused assignments

DNin01:
### Problem

The ESLint rule `no-useless-assignment` recently became recommended and is now checked for in our repository. As a result, some code now creates error messages. These errors happen when a variable is **assigned a value that is never read later** or is guaranteed to be **reassigned before it is read**.

You can learn more about this on the [ESLint Docs](https://eslint.org/docs/latest/rules/no-useless-assignment).

### Changes

These errors were fixed. In the process, a bug in `editor-theme3` where the addon always use the old Blockly reference for `FieldNote` was fixed. The 10 other changes should have no effect.

Since opening this PR, ESLint was pinned to version 9.x, so this PR explicitly enables the rule that is recommended on v10.

### Tests

ESLint checks succeed on the changes.

Webpages tested in Edge 145.

### DIFF
--- a/background/l10n.js
+++ b/background/l10n.js
@@ -26,17 +26,14 @@ export default class BackgroundLocalizationProvider extends LocalizationProvider
 
     localeLoop: for (const locale of locales) {
       for (const addonId of addonIds) {
-        let resp;
-        let messages;
         const url = `/addons-l10n/${locale}/${addonId}.json`;
         try {
-          resp = await fetch(url);
-          messages = await resp.json();
+          const resp = await fetch(url);
+          const messages = await resp.json();
+          this.messages = Object.assign(messages, this.messages);
         } catch (_) {
           if (addonId === "_general") continue localeLoop;
-          continue;
         }
-        Object.assign(this.messages, messages);
       }
     }
     this._reconfigure();
--- a/content-scripts/inject/l10n.js
+++ b/content-scripts/inject/l10n.js
@@ -13,20 +13,17 @@ export default class UserscriptLocalizationProvider extends LocalizationProvider
     }
     let addonMessages = {};
     for (const dir of this._urls) {
-      let resp;
-      let messages;
       const url = `${dir}/${addonId}.json`;
       try {
-        resp = await fetch(url);
-        messages = await resp.json();
+        const resp = await fetch(url);
+        const messages = await resp.json();
+        addonMessages = Object.assign(messages, addonMessages);
+        this.messages = Object.assign(messages, this.messages);
       } catch (_) {
         if (addonId === "_general") {
           this._urls.delete(dir);
         }
-        continue;
       }
-      Object.assign(addonMessages, messages);
-      Object.assign(this.messages, messages);
     }
     if (addonId === "_general") {
       this._reconfigure();


Samq64:
It might be a good idea to pin the ESLint version to 10.x and update it on our own schedule.

addons/editor-theme3/theme3.js
```diff
@@ -802,7 +802,7 @@ export default async function ({ addon, console, msg }) {
 
   let FieldNote;
   if (Blockly.registry) FieldNote = Blockly.registry.getClass(Blockly.registry.Type.FIELD, "field_note");
-  FieldNote = Blockly.FieldNote;
+  else FieldNote = Blockly.FieldNote;
```
DNin01, line 805: NOW all addons support the new Blockly. 😄

background/l10n.js
```diff
@@ -27,7 +27,7 @@ export default class BackgroundLocalizationProvider extends LocalizationProvider
     localeLoop: for (const locale of locales) {
       for (const addonId of addonIds) {
         let resp;
-        let messages = {};
+        let messages;
```
mxmou: I don't think I understand the reason for this change (and the one in the other `l10n.js`). The rest look good.

DNin01: It's weird, but ESLint gets upset if I don't do this. I think it's because there's a `continue` statement in the `catch` block (in which case, the scope exits early and the empty object is technically never accessed).

See [https://eslint.org/play/...](https://eslint.org/play/#eyJ0ZXh0IjoiLyogZXNsaW50IG5vLXVzZWxlc3MtYXNzaWdubWVudDogXCJlcnJvclwiICovXG5mb3IgKGxldCBhIG9mIGIpIHtcbiAgbGV0IGEgPSB7fTtcbiAgdHJ5IHtcbiAgICBhID0ge307XG4gIH0gY2F0Y2gge1xuICAgIGNvbnRpbnVlOyAvLyBUcnkgY29tbWVudGluZyB0aGlzIG91dFxuICB9XG4gIE9iamVjdC5hc3NpZ24oYSwgYyk7XG59Iiwib3B0aW9ucyI6eyJydWxlcyI6e30sImxhbmd1YWdlT3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19)

mxmou: Oh, that makes sense. So the `|| {}` you added is also technically not needed?

DNin01: Oh, you're right. `Object.assign` works with null `source` arguments, it just assigns nothing.

DNin01: Wait, actually, no. Here, the first argument, `target`, which has to be an object, is `messages`.
https://github.com/ScratchAddons/ScratchAddons/blob/b66574f00fb8ffd64fc607653dab42b464ab9d0c/background/l10n.js#L39
It might be possible to switch this around, though (and then we wouldn't have to reassign `this.messages`).

DNin01: I've changed the reassignment to in-place accumulation by just swapping the two arguments. This may slightly improve speed and memory usage, too.

mxmou: I meant that the null check isn't needed because `messages` is never null (either the request is successful or the `continue` is reached).

DNin01: I don't see a null check...

mxmou: The `|| {}`.

DNin01: Oh, I changed the code again so it isn't needed anyway.

mxmou (approved):

background/l10n.js
```diff
@@ -36,7 +36,7 @@ export default class BackgroundLocalizationProvider extends LocalizationProvider
           if (addonId === "_general") continue localeLoop;
           continue;
         }
-        this.messages = Object.assign(messages, this.messages);
+        Object.assign(this.messages, messages);
```
DNin01: Uh oh, it seems like this change makes addons always display in English because the locales are now accumulated in reverse. That's because there is one functional difference here...

- **Before:** `messages` entries are replaced by **old** ones from `this.messages`.
- **After:** `this.messages` entries are replaced by **new** ones from `messages`.

I'm on a fix.

DNin01: Aha! So, we didn't even need to reassign the variables.

background/l10n.js
DNin01: > I meant that the null check isn't needed because `messages` is never null (either the request is successful or the `continue` is reached).

I see now!

DNin01:
No opinions on the `no-useless-assignment` rule yet, but it's recommended by the latest version of ESLint and even helped us fix a bug, so I say, let's give it a try.

PR #8849, opened 2026-02-19
Fix stage header when `dark-www` and `scratch3to2` are active simultaneously

iqnite:
Resolves https://discord.com/channels/806602307750985799/1469856099590213683

### Changes

Updated the `scratch3to2` addon's userstyle to apply the page background and border colors from `dark-www` on the stage header, if the latter is enabled. If `dark-www` is enabled, the gradient uses the colors from "Page background" and a darker variant.

### `dark-www` disabled

image

### `dark-www` enabled

image

### Reason for changes

This change ensures that colors are consistent when both `dark-www` and `scratch3to2` are enabled.

### Tests

Tested on Microsoft Edge 145.0.3800.65.

### DIFF
--- a/addons/scratch3to2/project.css
+++ b/addons/scratch3to2/project.css
@@ -186,7 +186,10 @@
 }
 :root:not(.sa-editor) [class*="stage-header_stage-header-wrapper_"],
 :root:not(.sa-editor) [class*="stage-header_stage-header-wrapper-overlay_"] /* embeds */ {
-  background-image: linear-gradient(var(--darkWww-page, white), color-mix(in srgb, var(--darkWww-page, white), black 10%));
+  background-image: linear-gradient(
+    var(--darkWww-page, white),
+    color-mix(in srgb, var(--darkWww-page, white), black 10%)
+  );
   border: 1px solid var(--darkWww-border-15, #d0d1d2);
   border-bottom: none;
   border-radius: 8px 8px 0 0;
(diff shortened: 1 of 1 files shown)


DNin01 (approved):

iqnite:
@mxmou I've now updated it to use the colors from the `page` and `gray` settings. I feel it's more flexible than arbitrarily determining light/dark, looks better on colored backgrounds, and there's no notable difference when using the Scratch 2.0 theme preset.

mxmou:
That's better, but the gradient still isn't as noticeable as before with the 2.0 preset. Instead of `page` and `gray`, I suggest using two different shades of `page`, which can be generated using CSS like this:
* `var(--darkWww-page, white)`
* `color-mix(in srgb, var(--darkWww-page, white), black 10%)`

addons/scratch3to2/project.css
mxmou: This should be changed as well for consistency.

mxmou: And this.

iqnite:
> `color-mix(in srgb, var(--darkWww-page, white), black 10%)`

That's funny, I searched for this for 10 minutes without any results. Thank you.

mxmou (approved):

WorldLanguages:
> > `color-mix(in srgb, var(--darkWww-page, white), black 10%)`
> 
> That's funny, I searched for this for 10 minutes without any results. Thank you.

Modern CSS is quite cool.

PR #8806, opened 2026-01-30
Fix CSS selectors after update

Samq64:
Resolves #8799
Resolves #8800
Resolves #8801
Resolves #8802

### Changes

Fixes some CSS selectors after a Scratch update.

### Tests

Tested on Helium.

PR #8850, opened 2026-02-19
Fix search-sprites, forward-compatible CSS selectors

mxmou:
### Changes

Fixes the Sprite pane search addon, which was broken by a Scratch update.

Also updates CSS selectors in other addons in a similar way so that such issues are less likely to happen in the future. See https://github.com/ScratchAddons/ScratchAddons/pull/8806#discussion_r2747654851.

### Tests

Tested the changed addons on Edge.

### DIFF
--- a/addons/fullscreen/userscript.js
+++ b/addons/fullscreen/userscript.js
@@ -40,7 +40,7 @@ export default async function ({ addon, console }) {
       addon.settings.get("toolbar") === "hover"
     ) {
       const canvas = await addon.tab.waitForElement('[class*="stage_full-screen"] canvas');
-      const header = await addon.tab.waitForElement('[class*="stage-header_stage-header-wrapper_"]');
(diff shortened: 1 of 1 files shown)


WorldLanguages:
Thanks!

addons/expanded-backpack/backpack.css
Samq64, line 1: Do the class names in this file not contain an underscore after?

mxmou: I don't think the underscore is as important as using `*=`, but I can add it.

addons/fullscreen/userscript.js
DNin01, line 1: `fullscreen` breaks and does this when "toolbar visibility" is set to "When hovered":

Project page with broken toolbar

It fixes when I change settings or disable the addon. No errors, it just glitches out. Not sure what this has to do with.

PR #8856, opened 2026-02-27
add addon `Bool variable`

BlockerNagnto:
Resolves #

Scratch's code is not concise enough, often need to use <(var)=(true)> to allow a variable to be input into the discriminant.

### Changes

Allow variables to be placed into the Boolean input slot just like list.

### Reason for changes

Variables can hold the same things as lists, but they don't receive the same treatment.

I don't want to use lists with only one item.

### Tests

only test on chrome.
known issue(or function?):the bool slot with text will make a lot of block disapear in editor if this addon was disabled.
this won't stop the code(only tested in chrome and scratch(android app,3.0.66-minSdk26))

### DIFF
--- a/addons/bool-variable/userscript.js
+++ b/addons/bool-variable/userscript.js
@@ -38,6 +38,15 @@ export default async function ({ addon }) {
       return true;
     }
 
-    return originalCheckType.call(this, otherConnection);
+    // B. 修復顯示報錯邏輯：處理原本會崩潰的 String/Boolean 衝突
+    const result = originalCheckType.call(this, otherConnection);
+    if (!result) {
+      const isThisBool = this.check_ && this.check_.includes("Boolean");
+      const isOtherBool = otherConnection.check_ && otherConnection.check_.includes("Boolean");
+      if (isThisBool || isOtherBool) {
+        return true;
+      }
+    }
+    return result;
   };
 }


BlockerNagnto:
甚麼情況
怎麼過了

CST1229:
i don't think we'd add this addon, BECAUSE it allows making scripts that break the vanilla Scratch editor.

and, you shouldnt add the non-english localization strings yourself.

BlockerNagnto:
> and, you shouldnt add the non-english localization strings yourself.

even non-english localization strings actaully is orignal text?
bot translation is not always perfect

CST1229:
> > and, you shouldnt add the non-english localization strings yourself.
> 
> even non-english localization strings actaully is orignal text? bot translation is not always perfect

localizations are done sorta-outside this repo (on [Transifex](https://scratchaddons.com/docs/localization/joining-the-localization-team/)) and pulled into it via a bot

also, you don't need an English json file if there aren't any strings *from the addon userscript* that need translating. so if it's only string IDs that start with `@` then you can remove the English file

CST1229:
> i don't think we'd add this addon, BECAUSE it allows making scripts that break the vanilla Scratch editor.
> 
> and, you shouldnt add the non-english localization strings yourself.

(but due to the above reason i don't think this addon will get merged anyways, so this is mostly just advice for any future addons you might want to make)

see also: https://scratchaddons.com/docs/localization/localizing-addons/

BlockerNagnto:
>BECAUSE it allows making scripts that break the vanilla Scratch editor.

code will still work on vanilla Scratch,just won't be visable at editor

BlockerNagnto:
🙃

BlockerNagnto:
maby i can turn this into "hiding block" function

BlockerNagnto:
>BECAUSE it allows making scripts that break the vanilla Scratch editor.

https://scratch.mit.edu/projects/382446179/
breaking thing is a function
and that's why i put warning tag

iqnite:
> breaking thing is a function and that's why i put warning tag

Warning tag or not, Scratch Addons are not supposed to break things.

Off-topic, but is it actually possible to achieve this functionality without transpilation or breaking anything?

DNin01:
> Off-topic, but is it actually possible to achieve this functionality without transpilation or breaking anything?

[Magic equals sign.webm](https://github.com/user-attachments/assets/da93da6d-bc3b-44f1-a501-0955da276cd7)

BlockerNagnto:
> Off-topic, but is it actually possible to achieve this functionality without transpilation or breaking anything?

whith out transpilation it,scratch will found a wrong input type and reject showing it

> > Off-topic, but is it actually possible to achieve this functionality without transpilation or breaking anything?
> 
>  Magic.equals.sign.webm

this one just make old trick automatic,still not did the thing i want(Scratch's code is not concise enough)
maby i really need to make the bool var list

DNin01:
> > Off-topic, but is it actually possible to achieve this functionality without transpilation or breaking anything?
>
> whith out transpilation it,scratch will found a wrong input type and reject showing it

Are there no other ways to indirectly connect a variable to a boolean slot? I wonder if there's a "hacked" block with a boolean shape but a round input? I mean like in this TurboWarp extension: https://turbowarp.org/editor?extension=https://extensions.turbowarp.org/true-fantom/couplers.js

CST1229:
> Are there no other ways to indirectly connect a variable to a boolean slot? I wonder if there's a "hacked" block with a boolean shape but a round input? I mean like in this TurboWarp extension: https://turbowarp.org/editor?extension=https://extensions.turbowarp.org/true-fantom/couplers.js

idk.

for casting: `not <(variable) = <false block>>` (where `<false block>` is any block that returns a *boolean* false, like `<> and <>`. string false, like typing in the input directly, won't work properly for some cases) seems to match the closest to TurboWarp's boolean cast block, but it still doesn't match up for the empty string, undefined (hacked blocks/Translate errors only) and number `NaN` (works for string `NaN` though)
[bool cast.zip](https://github.com/user-attachments/files/25694400/bool.cast.zip) (rename file extension to .sb3 then load in turbowarp)

BlockerNagnto:
> > > Off-topic, but is it actually possible to achieve this functionality without transpilation or breaking anything?
> > 
> > 
> > whith out transpilation it,scratch will found a wrong input type and reject showing it
> 
> Are there no other ways to indirectly connect a variable to a boolean slot? I wonder if there's a "hacked" block with a boolean shape but a round input? I mean like in this TurboWarp extension: https://turbowarp.org/editor?extension=https://extensions.turbowarp.org/true-fantom/couplers.js

don't have one in scratch, that's why turbowarp needs to add it

> > Are there no other ways to indirectly connect a variable to a boolean slot? I wonder if there's a "hacked" block with a boolean shape but a round input? I mean like in this TurboWarp extension: https://turbowarp.org/editor?extension=https://extensions.turbowarp.org/true-fantom/couplers.js
> 
> idk.
> 
> for casting: `not <(variable) = <false block>>` (where `<false block>` is any block that returns a _boolean_ false, like `<> and <>`. string false, like typing in the input directly, won't work properly for some cases) seems to match the closest to TurboWarp's boolean cast block, but it still doesn't match up for the empty string, undefined (hacked blocks/Translate errors only) and number `NaN` (works for string `NaN` though) [bool cast.zip](https://github.com/user-attachments/files/25694400/bool.cast.zip) (rename file extension to .sb3 then load in turbowarp)

as i said on top, scratch can do it most of time like list - only 1、true(Regardless of capitalization) turn into true output, other are all false
it's safe enough

e016:
> for casting: not <(variable) = <false block>> (where <false block> is any block that returns a boolean false, like <> and <>. string false, like typing in the input directly, won't work properly for some cases) seems to match the closest to TurboWarp's boolean cast block

You can use `<(variable) > [0]>` which correctly checks for 1, true, and empty!

CST1229:
> You can use `<(variable) > []>` which correctly checks for 1, true, and empty!

that one doesnt match up for false (both boolean and string false), 0, NaN and undefined. so that might be worse

e016:
> that one doesnt match up for false (both boolean and string false), 0, NaN and undefined. so that might be worse

I've tested it again, and `<(variable) > (0)>` _does_ work!

DNin01:
> > that one doesnt match up for false (both boolean and string false), 0, NaN and undefined. so that might be worse
> 
> I've tested it again, and `<(variable) > (0)>` _does_ work!

But the string `"false"` is greater than 0, too.

e016:
Yeah, I figured that out. There's no good way to check if something's truthy!!

BlockerNagnto:
> Yeah, I figured that out. There's no good way to check if something's truthy!!

scratch is fun

try<(var)=(1)>

and block opcode "argument_editor_boolean"

they work

e016:
> try<(var)=(1)>

yep! that works!!!!!!! (the string "1" not 1 the number)

> and block opcode "argument_editor_boolean"

wait, i found it, lemme add it to a project

e016:
The `argument_editor_boolean` works well for inserting a constant "true"/"false" string, but it doesn't work as a coupler:

image

Plus, `<[1] = <true>>` gives false :(

BlockerNagnto:
> The `argument_editor_boolean` works well for inserting a constant "true"/"false" string, but it doesn't work as a coupler:
> 
> image
> 
> Plus, `<[1] = <true>>` gives false :(
> 

i remember "1" in that will return true

PR #8860, opened 2026-03-01
New Addon: `Text2Blocks`

### LINKED ISSUE #8573: New Addon: Convert scratchblocks Text to Blocks
### Why this would be helpful

Currently, when we receive scratchblocks-formatted text code from LLMs or other sources, we have to manually convert it into visual blocks. This process is not only time-consuming but also prone to errors. Adding an addon to automate this step would significantly improve efficiency—especially useful for teaching, project sharing, or rapid prototyping.

### How this addon works

This addon would add a button to the editor. When clicked, it opens a popup window with the following tabs:

- **Code**: An input field where users can enter or paste scratchblocks text code.
- **Variables/Lists**: If the code includes variables or lists, users should be able to select existing ones or create new ones (with an option to make them _for this sprite only_).
- **Events, Costumes, Sounds, Custom Blocks**: Same logic as “Variables/Lists.”
- **Extensions**: Displays which extensions will be added based on the code.

### Possible alternatives

This addon shares some similarities with the existing `middle-click-popup` addon, which is designed for quickly inserting a small number of blocks manually. However, the proposed addon focuses on bulk automatic generation—its goal is to convert complete scratchblocks text into blocks, making it ideal for importing full scripts from AI or other text sources. While both addons deal with block insertion, their use cases and target audiences differ.

### Additional context

This feature would be especially valuable for educators, content creators, and users leveraging AI-assisted programming.

Note:
This addon is currently just a proposal. While I may explore building it myself, I'd be genuinely happy if someone else is interested in developing it or contributing ideas. Whether it's implementation suggestions, design feedback, or taking on the development entirely—any form of collaboration is warmly welcomed!

LuYifei2011:
Resolves #8573

### Changes

https://github.com/user-attachments/assets/51b8a0b3-7ed6-443c-bcd3-1ebb3d265af6

### Reason for changes

Allow users to paste code from forums or other sources and convert it into blocks.

### Tests

On Chrome 147.
[test samples](https://gist.github.com/LuYifei2011/2d7190b62de383c99750bbf1800b8534?permalink_comment_id=6091577#gistcomment-6091577)

### Future work

* [x] Complete custom block support
* [ ] Fix the issue when collecting all variable and list names.
* [ ] Make error messages more human-friendly
* [ ] Inspect and improve dropdown menu behavior and validators, especially dynamic ones
* [ ] Test more blocks

The following may or may not be implemented:

* [ ] Support comments
* [ ] Lenient syntax for math and logic operations
* [ ] Display which extensions will be added
* [ ] Events, costumes, and sounds tabs that can create new ones (for events) or select existing ones, similar to the variables/lists tab
* [ ] Support Scratch 2, TurboWarp (e.g., `is TurboWarp?`), SA debugger, and hidden blocks/values (dropdowns and inputs)


BlockerNagnto:
what if make this also support copy block as text?

that'll make us easier to copy code from project to project

Samq64:
> what if make this also support copy block as text?

There are already 3 open pull requests that implement that in different ways. The main tracking issue is #2336.

addons/editor-text2blocks/scratchblocks/blocks-info.js
```diff
@@ -0,0 +1,2617 @@
...
+  data_listcontents: {
+    shape: "reporter",
+    category: "list",
+    params: [],
+    skipLocaleBuild: true,
+  },
+  data_addtolist: {
+    shape: "stack",
+    category: "list",
+    params: [
```
FlamedDogo99, line 1047: I think you have the order switched here. Should be
```
params: [
      {
        name: "LIST",
        type: FIELD_DROPDOWN,
        dynamicOptions: "lists",
      },
      {
        name: "ITEM",
        opcode: "text",
      },
    ],
```

addons/editor-text2blocks/scratchblocks/syntax.js
```diff
@@ -0,0 +1,858 @@
...
+        }
+        return;
+      }
+
+      let name, info;
+      if (block.isReporter && block.info.category === "variables" /*&& block.info.categoryIsDefault*/) {
+        block.info.opcode = "data_variable";
+        name = blockName(block);
+        info = block.info;
+      } else if (block.isReporter && block.info.category === "list" /*&& block.info.categoryIsDefault*/) {
```
FlamedDogo99, line 822: This breaks `item (x) of [list v]`, maybe try something more strict like this?

```js
} else if (block.isReporter && block.info.category === "list" /*&& block.info.categoryIsDefault*/) {
  if (block.children.filter(c => c.isInput).length === 0) {
    block.info.opcode = "data_listcontents"
  }
  name = blockName(block);
  info = block.info;
}
if (!name) {
  return;
}

// list reporters
if (listNames.has(name)) {

  info.category = "list";
  info.categoryIsDefault = false;
  if (block.children.filter(c => c.isInput).length === 0) {
    info.opcode = "data_listcontents";
  }
}
```

addons/editor-text2blocks/scratchblocks/blocks-info.js
LuYifei2011: Pardon. I don't see any difference between your code and the original one

PR #8867, opened 2026-03-05
Rename Blockly renderer from scratch to scratch_classic

Samq64:
### Changes

Renames a couple Blockly renderer class references from `scratch` to `scratch_classic` to fix those addons on new Blockly.

### Tests

Tested on Helium. I did not test with the Scratch membership and additional might need to be made for those with its variant of cat blocks enabled.

### DIFF
--- a/addons/cat-blocks/userscript.js
+++ b/addons/cat-blocks/userscript.js
@@ -323,7 +323,7 @@ export default async function ({ addon, console, msg }) {
       return hat;
     };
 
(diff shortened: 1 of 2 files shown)


mxmou:
I haven't been able to test this with Scratch's cat block renderer yet, but the addons will probably have to override the methods of both renderers.

Samq64:
I have just tested with the membership flag and this seems to work fine as it is.

PR #8873, opened 2026-03-07
cat-blocks & editor-theme3: fix duplicated menu items

mxmou:
Updates some selectors to fix menu items appearing twice (https://github.com/ScratchAddons/ScratchAddons/pull/8742#issuecomment-4013938960).


DNin01:
I see what happened—it found two `<ul>` elements!

mxmou:
> Another way to narrow down the selector would be to use a class selector like:
> 
> ```css
> ul[class*="menu_menu_"]
> ```
> 
> (Its parent element has a class of `menu_submenu_`.)

Good idea - that will work even if Scratch changes the DOM structure again.

PR #8874, opened 2026-03-07
v1.45 update notification

DNin01:
Usually @WorldLanguages makes the update notifications, but I had more time for this one.

What do you think?

Update notification for v1.45

The graphic used on the costume cards in the image is a cat I created for a mockup. It's what I imagine as our mascot, but that's beside the point. 😆

To test, just checkout then go to https://scratch.mit.edu/#sa-update-notif.


_locales/en/messages.json
```diff
@@ -296,11 +296,11 @@
...
   },
-  "extensionUpdateImageAlt_v1_44": {
-    "message": "Screenshot of the \"recolor custom blocks\" addon in action"
+  "extensionUpdateImageAlt_v1_45": {
+    "message": "Screenshot of dragging a costume in the costume list"
   },
-  "extensionUpdateInfo1_v1_44": {
-    "message": "The new \"recolor custom blocks\" addon adds an option to choose a custom color when creating a block. Go to $1 to enable it.",
+  "extensionUpdateInfo1_v1_45": {
+    "message": "The \"Improved dragging of costumes and sounds\" addon, new and enabled by default, makes the costume and sound lists scroll automatically when reordering items. Check it out in $1.",
```
mxmou:
```suggestion
    "message": "The new \"Improved dragging of costumes and sounds\" addon, which is enabled by default, makes the costume and sound lists scroll automatically when reordering items. Check it out in $1.",
```

mxmou (approved):
I like the image. The text is also good, but there's one sentence I'd change slightly - see below.

Hexagonl:
When do you guys plan on releasing the update? I'd like to make the update video before the update releases so it can be put into this update notification.

DNin01:
> When do you guys plan on releasing the update? I'd like to make the update video before the update releases so it can be put into this update notification.

I'd say probably in about two weeks, the end of this month at the latest.

@WorldLanguages, do you have an answer?

WorldLanguages:
Nice image to describe the new addon @DNin01

@Hexagonl We could decide on a date no problem, but for this specific update notification, the image by DNin01 looks superior in terms of telling the average user what's new, compared to a YouTube thumbnail for the small % of users that are really into the details of every update and will click. 

Anyone have ideas on how we could advertise the YouTube video without compromising the image? Or if you don't think keeping the image is a big deal, feel free to say that as well.

mxmou:
> Anyone have ideas on how we could advertise the YouTube video without compromising the image? Or if you don't think keeping the image is a big deal, feel free to say that as well.

I suggest adding a button next to the notification text. A few mockups of how it could be placed:

image

image

image

DNin01:
We could also just link to `youtube.com/@ScratchAddons` so that we don't have to hardcode the video URL every time, if that would be easier.

Hexagonl:
> We could also just link to `youtube.com/@ScratchAddons` so that we don't have to hardcode the video URL every time, if that would be easier.

I think that should exist as a backup, as I do think directly linking to the video is preferable yet I do acknowledge that sometimes the video isn't always ready by the time the update is released

I like the mockup, though perhaps something like this is cleaner (courtesy of photoshopping):

image

DNin01:
I think I'll merge this now and add the video button a little later. Sound good, @WorldLanguages? I know you usually like to allow some time for translations.

(I can at least add the string for the button now so it can be translated ahead of time.)

@mxmou, how did you make the button? Or do you want to open a PR to add it?

mxmou:
I'll make a PR for the button. Which design should I use?

DNin01:
> I'll make a PR for the button. Which design should I use?

I would say [Hexagonl's](https://github.com/ScratchAddons/ScratchAddons/pull/8874#issuecomment-4033137767), except that there's probably not enough space for the button to fit between wider screenshots and the notification's footer without necessitating wrapping or resizing, but we could try it for this update notification and adjust it for future ones as needed.

_locales/en/messages.json
```diff
@@ -296,11 +296,11 @@
...
   },
-  "extensionUpdateImageAlt_v1_44": {
-    "message": "Screenshot of the \"recolor custom blocks\" addon in action"
+  "extensionUpdateImageAlt_v1_45": {
+    "message": "Screenshot of dragging a costume in the costume list"
   },
-  "extensionUpdateInfo1_v1_44": {
-    "message": "The new \"recolor custom blocks\" addon adds an option to choose a custom color when creating a block. Go to $1 to enable it.",
+  "extensionUpdateInfo1_v1_45": {
+    "message": "The new \"Improved dragging of costumes and sounds\" addon, which is enabled by default, makes the costume and sound lists scroll automatically when reordering items. Check it out in $1.",
```
mxmou, line 303: This isn't a new issue, but I sometimes forget to correct the Slovenian translation of the "Scratch Addons settings" link when the grammatical context changes. For translators, it would be much better if the link text was part of this string - perhaps like this if there's no better way:

```
Check it out in $1Scratch Addons settings$2.
```

DNin01: One of my ideas was to use either tab characters or \<brackets> to mark the start and end of the link.

Games with tokenized dialogue probably do this all the time.

mxmou: I think it's better use something that Transifex recognizes as special and will warn translators if they leave it out. This includes placeholders like $1 and HTML tags: https://help.transifex.com/en/articles/6240403-translating-html-content

PR #8876, opened 2026-03-07
hide-backpack: spork updates

mxmou:
### Changes

The design of spork's zoom buttons has been changed to be more similar to the current version of Scratch. This PR updates the code that moves those buttons to make space for the backpack button. It also updates the backpack button's position.

### Tests

Tested https://scratchfoundation.github.io/scratch-editor/develop/scratch-gui/ on Edge.

### DIFF
--- a/addons/hide-backpack/userscript.js
+++ b/addons/hide-backpack/userscript.js
@@ -1,4 +1,13 @@
+let RESIZE_BUTTON_OFFSET;
+
 export default async function ({ addon, console }) {
+  if (document.querySelector(".blocklyToolboxCategoryGroup")) {
+    // new Blockly
+    RESIZE_BUTTON_OFFSET = 31;
+  } else {
+    RESIZE_BUTTON_OFFSET = 36;
+  }
+
   let originalBackpack;
 
   // Event listeners that add dynamic enable/disable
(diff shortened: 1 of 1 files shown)


addons/hide-backpack/userscript.js
```diff
@@ -88,7 +88,8 @@ function moveResizeButtons(addon, distance) {
   if (document.querySelector(".blocklyToolboxCategoryGroup")) {
     // new Blockly
     const workspace = addon.tab.traps.getWorkspace();
-    workspace.zoomControls_.MARGIN_VERTICAL = 20 + distance;
+    const zoomControls = workspace.getComponentManager().getComponent("zoomControls");
+    zoomControls.MARGIN_VERTICAL = 20 + distance;
```
Samq64, line 101: 15 is about to the distance between the zoom out and reset buttons.
```suggestion
    zoomControls.MARGIN_VERTICAL = 15 + distance;
```

mxmou: The 20 here is correct - if `distance` is zero, which happens when the addon is disabled, the margin needs to be reset to the default value of 20. I'll fix the spacing in a different way.

PR #8879, opened 2026-03-08
editor-dark-mode: fix menu scrollbars

mxmou:
### Changes

Before:
image

After:
image

Also removes the CSS for face sensing notifications.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/editor-dark-mode/experimental_editor.css
+++ b/addons/editor-dark-mode/experimental_editor.css
@@ -469,6 +469,8 @@ img[class*="tool-select-base_tool-select-icon_"],
   border-color: var(--editorDarkMode-primary);
 }
 [class*="input_input-form_"]:not([class*="project-title-input_title-field_"]):hover,
+[class*="react-tabs_react-tabs__tab_"]:focus-visible,
+[class*="react-tabs_react-tabs__tab_"][class*="gui_tab_"][class*="gui_is-selected_"]:focus-visible,
 [class*="toggle-buttons_button_"]:focus::before,
 [class*="stage_frame_"],
 [class*="sprite-selector-item_sprite-selector-item_"]:hover,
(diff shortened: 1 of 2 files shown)


addons/editor-dark-mode/experimental_editor.css
```diff
@@ -762,3 +736,10 @@ div[class*="delete-confirmation-prompt_body_"] /* <-- specificity */ [class*="de
   /* The <object> gets a white background if color-scheme is dark. */
   color-scheme: light;
 }
+
+[class*="react-tabs_react-tabs__tab_"]:focus {
+  outline: revert;
+}
+[class*="react-tabs_react-tabs__tab_"]:focus::after {
+  display: none;
+}
```
DNin01: The focus underline on editor tabs is vanilla Scratch behavior. It shouldn't have to be changed to a different type of indicator by this addon.

mxmou: The underline is almost invisible in vanilla:
image

(it's on the Costumes tab in this image)

The CSS for that is from the react-tabs library and Scratch should probably override it to match the design of the tabs, but it doesn't. If we want the addon to look like vanilla Scratch, it should change the white underline to a dark one in dark mode, but I think it makes more sense to use something that isn't impossible to see.

DNin01: The addon could also change the underline to a light one in dark themes and a dark one in light themes, or to the highlight color, so that the shape of Scratch's current design stays the same but the colors closer match the user's expectations of the theme they configured to change them.

mxmou: I think changing the border color and adding an outline would fit Scratch's design best - that's what focused inputs and the stage size buttons do, for example.

mxmou: I've changed the focused tab design to this:

image

DNin01: @mxmou focused tabs now look like this officially:

Outlined costume editor tab

mxmou: The white underline is has now been fixed by Scratch, so I'll revert that part of the PR.

DNin01 (approved):

PR #8865, opened 2026-03-03
Added a "previous costume" dropdown in `editor-random-direction-block`

e016:
Resolves #8778

### Changes

Adds an option to add a "previous costume" option in the "switch costume to ()" block

### Reason for changes

Current workarounds with (join () ()) and (costume [#] - 1) are cumbersome. It's easier to use the dropdown than to make the block yourself.

### Tests

I've only tested in chrome.

PR #8884, opened 2026-03-10
Add more menu items with `editor-random-direction-block`

### LINKED ISSUE #6761: New addon: add "self" to sprite dropdowns in blocks
### Why this would be helpful

> I have a suggestion that I don't think is already a thing, that being to allow selecting blocks for a sprite within the same sprite. For example, being able to choose "go to [Sprite1]" or "[x position] of [Sprite1]" inside of Sprite1's code, instead of needing to get those blocks from another sprite and copy them over. These blocks are incredibly useful when dealing with clones (as they will always output the data of the original sprite, even to clones) so being able to select them more easily would be nice.

### How this addon works

Simply adds the option in the dropdown so that it's not necessary to use the backpack/drag-and-drop for this.
Useful for clones.

### Possible alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #8778: Back Costume
### Why this would be helpful

I am creating a custom block, and would find it easier to use a reverse 'Next Costume' block for my animate cblock but i cant do that.
@Facerig_Rig on scratch.

### How this addon works

Works similar to the already existing Next Costume block, once activated sends the costume back one, basically being a reverse Next Costume block.

### Possible alternatives

not sure.

### Additional context

@Facerig_Rig on scratch.

DNin01:
Resolves #6761
Resolves #8778
Supersedes #8865

### Changes

You can now add "next costume" and "previous costume" items to the "switch costume to" block menu and select the current sprite in sprite menus using the `editor-random-direction-block` addon.

### Reason for changes

- Scratch does not let you select the current editing target in sprite menus, which makes it harder to use the parent sprite of a clone as a reference.
- Selecting "previous costume" can be faster than setting up the blocks to decrement the costume number by 1.

### Notes

Tested in Edge 145.

Minimum release version: 1.46.0

Credits:
- Based on the work done by @e016 in #8865.

### DIFF
--- a/addons/editor-random-direction-block/userscript.js
+++ b/addons/editor-random-direction-block/userscript.js
@@ -1,27 +1,68 @@
-export default async function ({ addon, global, cons, msg }) {
+export default async function ({ addon, console, msg }) {
   const ScratchBlocks = await addon.tab.traps.getBlockly();
+  const runtime = await addon.tab.traps.vm.runtime;
 
-  function appendRandomOption(menuOptions) {
-    if (!addon.self.disabled) {
-      if (!menuOptions.find((option) => option[1] === "_random_"))
-        menuOptions.push([ScratchBlocks.Msg.MOTION_POINTTOWARDS_RANDOM, "_random_"]);
+  function extendPointTowardsMenu(menuOptions) {
+    if (!addon.self.disabled && addon.settings.get("randomDirection")) {
+      if (!menuOptions.find((option) => option[1] === "_random_")) {
+        menuOptions.splice(1, 0, [ScratchBlocks.Msg.MOTION_POINTTOWARDS_RANDOM, "_random_"]);
+      }
+    }
+    return menuOptions;
+  }
+  function extendSwitchCostumeMenu(menuOptions) {
+    if (addon.self.disabled) return menuOptions;
+    if (addon.settings.get("nextCostume")) {
+      if (!menuOptions.find((option) => option[1] === "next costume")) {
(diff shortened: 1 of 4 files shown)


e016:
Wow!!

I didn't even know "next costume" worked on the switch to costume block!!!

addons/editor-random-direction-block/userscript.js
```diff
@@ -1,50 +1,100 @@
...
+    if (addon.settings.get("nextCostume")) {
+      if (!menuOptions.find((option) => option[1] === "next costume")) {
+        menuOptions.push(["next costume", "next costume"]);
+      }
+    }
+    if (addon.settings.get("previousCostume")) {
+      if (!menuOptions.find((option) => option[1] === "previous costume")) {
+        menuOptions.push(["previous costume", "previous costume"]);
+      }
+    }
```
DNin01, line 24: These items are currently not localized, but they can be changed without issue, except that they change back to the unlocalized names if you disable the addon (which happens with the random direction block too). Scratch doesn't have built-in messages for these, so we'd have to use our own.

e016: Yeah, my PR originaly used an addon `msg("previous costume")` for the previous costume..

DNin01: Done! (Doesn't work immediately after clicking "See inside" for some reason, but that's probably not a new issue.)

Samq64:
There is also #4636 which is still open but abandoned.

PR #8893, opened 2026-03-16
Enable "curly" rule for multiline statements

DNin01:
### Changes

This pull request proposes adding the ESLint rule [`curly`](https://eslint.org/docs/latest/rules/curly) to enforce using braces in control statements that span multiple lines.

### Motivation

Single-line statements without braces as shown below is okay if it makes the code more compact.
```js
if (!foo) foo = 42;
```
But although some braces are optional, it's best practice to use them for control statements in general to avoid technical debt, especially when there are multiple blocks or lines. Not doing so can make it harder to understand or modify the code in the future. For example, modifying them without caution can create bugs:
```js
if (foo)
  foo++;
else
  foo = 42;
  console.log("foo declared."); // BUG: Outside the else statement!
```
```js
if (yes) confirm();
cancel(); // BUG: This is supposed to be next to an else statement!
```
Case in point: https://github.com/ScratchAddons/ScratchAddons/pull/8836#pullrequestreview-3787129416

It's easier when the statement looks like this:
```js
if (foo) {
  foo++;
} else {
  foo = 42;
}
```

This style can be enforced by using the ESLint rule `curly`, or only for multi-line statements using the rule `curly: ["multi-line"]`.

[Learn more on the ESLint documentation.](https://eslint.org/docs/latest/rules/curly)

I don't think there's any reason not to do this, but let me know what you think.

### DIFF
--- a/addons/middle-click-popup/WorkspaceQuerier.js
+++ b/addons/middle-click-popup/WorkspaceQuerier.js
@@ -1274,13 +1274,9 @@ export default class WorkspaceQuerier {
     function searchToken(token) {
       const subtokens = token.type.getSubtokens(token, query);
       if (subtokens) {
-        for (const subtoken of subtokens) {
-          searchToken(subtoken);
-        }
+        for (const subtoken of subtokens) searchToken(subtoken);
       } else if (!(token.type instanceof TokenTypeStringLiteral) && token.isProper && !token.isTruncated) {
-        for (let i = token.start; i < token.end; i++) {
-          canBeString[i] = false;
-        }
+        for (let i = token.start; i < token.end; i++) canBeString[i] = false;
       }
     }
     for (const result of results) searchToken(result.token);


DNin01:
I didn't expect there to be 94 contradictions...

Many of them are just `if` statements without an `else`, which... I guess make sense to change? A little annoying, but it's a trade-off.

DNin01:
It reminds me of how trailing commas are useful, I guess.

DNin01:
It seems like a lot of these may have been introduced by Prettier formatting turning single-line if-then statements into double-line ones, without braces… which is not my favorite, if you couldn't tell.

mxmou:
I agree with the intent of this change, but I'm afraid it will be annoying to have to fix things after Prettier.

DNin01:
> I agree with the intent of this change, but I'm afraid it will be annoying to have to fix things after Prettier.

We could just enforce braces on all control statements, multiline or not, so that when Prettier formats them they'll always be in the correct style. The tradeoff is that every control statement would have to be multiline.

Another option is to use ESLint's `--fix` option as part of auto-format. ESLint would add the missing braces, Prettier would format, and all would be well.

DNin01:
> Another option is to use ESLint's `--fix` option as part of auto-format. ESLint would add the missing braces, Prettier would format, and all would be well.

At worst, this would require three steps: format, lint, and format again, because ESLint doesn't space the braces it adds with the `--fix` option. Doable with workflows, but potentially pretty rudimentary.

We might have to revisit this if and once we rewrite our workflows.

WorldLanguages:
At first glance this doesn't look worth it just because it would have caught 1 single bug

Hans5958:
Actually I kinda have an opposing opinion on these kind of ESLint rule changes.

All rules have their benefits. There are even various set of recommended rules to help you avoid bugs. However, if these changes result in decreased productivity, the benefits may not be worth it. When it comes to this rule, we didn't have much of a problem where we forgot the curly brackets or so, so I don't think it is worth enabling.

DNin01:
We won't change the rules right now, then.

PR #8895, opened 2026-03-17
Do not capture Escape key when search empty

### LINKED ISSUE #8864: Search bar prevents Escape key from closing the extension popup
### Description

Usually, Chrome extension action popups can be closed by pressing Escape, but because the search bar is autofocused and has a listener that overrides the default behavior of Escape, this doesn't work on Scratch Addons' popup.

A simple fix would be to not prevent the default behavior if the search query is empty, as there is nothing to clear and no reason to prevent a default action.

### Steps to reproduce

1. Activate Scratch Addons
2. Open the "Addons" tab
3. Press <kbd>Escape</kbd>

_No response_

### Environment

Browser: Edge 145
Scratch Addons version: 8433f3921e5ab290a311f84c12c02cf383a2aae9
Operating system: Windows 11

DNin01:
Resolves #8864

Currently, pressing Escape clears the search box. This change allows the Escape key to close the extension's popup when the search box is already empty.

### Tests

Tested in Edge 146.

### DIFF
--- a/webpages/settings/index.js
+++ b/webpages/settings/index.js
@@ -673,8 +673,11 @@ let fuse;
       document.querySelector("#searchBox").focus();
     } else if (e.key === "Escape") {
       if (document.activeElement === document.querySelector("#searchBox")) {
-        if (vue.searchInputReal.length > 0) e.preventDefault();
-        vue.searchInputReal = "";
+        if (vue.searchInputReal.length > 0) {
+          // Escape is used to close extension popups, so we should only prevent it if there's input text to clear
+          e.preventDefault();
+          vue.searchInputReal = "";
+        }
       } else if (vue.categoryOpen && vue.smallMode) {
         vue.categoryOpen = false;
       } else {


webpages/settings/index.js
```diff
@@ -673,7 +673,7 @@ let fuse;
       document.querySelector("#searchBox").focus();
     } else if (e.key === "Escape") {
       if (document.activeElement === document.querySelector("#searchBox")) {
-        e.preventDefault();
+        if (vue.searchInputReal.length) e.preventDefault();
```
Samq64: Because this is a number being explicit might be better here
```suggestion
        if (vue.searchInputReal.length > 0) e.preventDefault();
```

DNin01:
> Looks good. A code comment explaining that we want Esc to exit the popup might be helpful.

How's this?

```js
// Escape is used to close extension popups, so we should only prevent it if there's input text to clear
```

mxmou:
I think that's good.

PR #8896, opened 2026-03-17
Cleanup some `webpages` code and styles

Samq64:
The changes are listed in the individual commits. ~~I am not entirely sure if there was another reason for `calculatePopupSize()` besides compatibility with older versions of Firefox .~~

### Tests

Tested on Firefox and Helium (Chromium-based) although this should also get some testing on mobile.

### DIFF
--- a/webpages/popup/index.js
+++ b/webpages/popup/index.js
@@ -3,7 +3,7 @@ import globalTheme from "../../libraries/common/global-theme.js";
 globalTheme();
 
 function calculatePopupSize() {
-  /* For mobile or when the default browser zoom is over 100% */
+  /* For mobile or when there isn't enough space for the full popup size */
   if (!window.innerWidth || !window.innerHeight) {
     setTimeout(calculatePopupSize, 0);
     return;


mxmou:
> I am not entirely sure if there was another reason for `calculatePopupSize()` besides compatibility with older versions of Firefox .

It's still needed to fix #600 and to support mobile devices where we can't use a fixed size.

webpages/popup/index.js
Samq64: @mxmou Do you remember what this magic number is for and why the fallback is 599px instead of 600px?

mxmou: For some reason, the magic number is needed to fix #600. 2 was enough on Chrome, but I had to increase it to 3 to support Firefox as well (see #1145).

I don't remember why I used 599px as the default height, but I don't think the exact number matters - if it's too big, this code will set `--height` to a smaller value anyway.

DNin01: I find that using `vw` and `vh` to hold up the outer layout lead to situations like this where it's harder to create responsive UI. I always suggest using flexboxes for this purpose.

I can look into this more later.

DNin01: Or is it because of iframes?

Samq64: With this PR there are no other manual height calculations in the popup. Both Firefox and Chromium behave for me at 110% zoom even after removing `overflow: hidden` from the body and the `3px`.

mxmou: You can remove those things then - note that `overflow: hidden` is defined twice for some reason. Interestingly, it looks like Firefox doesn't even use the default zoom for extension popups anymore - you can clearly see this if you set it to something like 150%.

webpages/settings/style.css
```diff
@@ -74,7 +75,8 @@ h1 {
   padding-top: 10px;
   scrollbar-width: thin;
   scrollbar-color: var(--gray-text) transparent;
-  height: 100vh;
+  height: 1px; /* flex-grow scroll hack */
```
Samq64: I did this because `addons-container` is a child of another `flex-grow` container, not the body which has a set height of 100vh.

mxmou: I'd use the `flex-basis` property instead of `height` in this case, to make it clear that it's just the base size that `flex-grow` may change. Then the comment would probably be redundant.

webpages/popup/index.js
```diff
@@ -3,25 +3,26 @@ import globalTheme from "../../libraries/common/global-theme.js";
 globalTheme();
 
 function calculatePopupSize() {
+  /* For mobile or when the default browser zoom is over 100% */
```
mxmou:
```suggestion
  /* For mobile or when there isn't enough space for the full popup size */
```

There are other situations where the popup width or height is limited. For example:

Image

WorldLanguages:
Regression: #8927

PR #8913, opened 2026-03-27
Add paint-canvas-pan addon

griffpatch:
### Changes

Adds a paint-canvas-pan addon that lets users pan the vector costume editor canvas by holding Space and dragging, or using middle-click drag. Zoom to cursor position with Alt+scroll.

Fully supports dynamic enable/disable
No dependencies on other addons

### Reason for changes

Standard navigation controls for a paint editor


mxmou:
> Zoom to cursor position with Alt+scroll.

Doesn't Ctrl+scroll already do that? Or is that a platform-specific thing?

griffpatch:
> > Zoom to cursor position with Alt+scroll.
> 
> Doesn't Ctrl+scroll already do that? Or is that a platform-specific thing?

It's possible... but I was making it consistent with the usual key used for this in illustrating apps, and making it a bit more responsive than the oddly slow default zoom

mxmou:
> It's possible... but I was making it consistent with the usual key used for this in illustrating apps, and making it a bit more responsive than the oddly slow default zoom

I think it would be nice if the addon improved the default zoom functionality in addition to adding the Alt key as an alternative to Ctrl. That way, users could use whichever modifier key they prefer (or even pinch-to-zoom on a touch screen) and still get the more responsive behavior. (I'm used to Ctrl since the only other drawing app I use is Inkscape.)

griffpatch:
Maybe the zoom should be a choice... it's the panning that I really wanted
to add

PR #8914, opened 2026-03-27
Feature/paint stroke options

griffpatch:
### Changes

Adds a new addon [paint-stroke-options](vscode-file://vscode-app/c:/Users/andy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) that injects stroke join and cap controls into the outline colour picker in the vector costume editor.

Corner: Miter or Round stroke join
Cap: Butt, Round, or Square stroke cap
Controls appear at the bottom of the stroke colour popup, match the Scratch colour picker visual style, and highlight the currently active option. Changes are applied to all selected items and trigger an undo snapshot.

image

### Reason for changes

Paper.js exposes [strokeJoin](vscode-file://vscode-app/c:/Users/andy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [strokeCap](vscode-file://vscode-app/c:/Users/andy/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) on path items but Scratch's paint editor provides no UI to change them. This addon fills that gap for users who want more control over how strokes render on vector shapes, particularly useful for custom block designs and illustrations.

### Tests

Tested in Chrome 146. Verified:

Controls appear only in the stroke (Outline) popup, not the fill popup
Clicking each option applies immediately to selected items
Active option is highlighted correctly
Undo works after applying a change
Addon enable/disable works without page reload


CST1229:
> Paper.js exposes strokeJoin and strokeCap on path items but Scratch's paint editor provides no UI to change them

*except for the line tool being able to change caps to round if you connect a line to an existing shape (e.g a rectangle with a part erased, which uses a butt cap)

griffpatch:
> > Paper.js exposes strokeJoin and strokeCap on path items but Scratch's paint editor provides no UI to change them
> 
> *except for the line tool being able to change caps to round if you connect a line to an existing shape (e.g a rectangle with a part erased, which uses a butt cap)

Quite so, an slightly inexplicably so :)

addons/paint-stroke-options/userscript.js
```diff
@@ -0,0 +1,129 @@
...
+    for (const item of items) {
+      item[prop] = value;
+    }
+    // Trigger undo snapshot via React fiber walk.
+    const canvasContainer = document.querySelector("[class*='paint-editor_canvas-container_']");
+    if (!canvasContainer) {
+      return;
+    }
+    let fiber = canvasContainer[addon.tab.traps.getInternalKey(canvasContainer)];
+    while (fiber && typeof fiber.stateNode?.handleUpdateImage !== "function") {
```
pufferfish101007, line 70: Is this loop guaranteed to terminate? Could you add a comment explaining why?

addons/paint-stroke-options/userscript.js
```diff
@@ -0,0 +1,129 @@
...
+      fiber = fiber.return;
+    }
+    if (fiber?.stateNode?.handleUpdateImage) {
```
pufferfish101007: Because of the loop condition above, can this be simplified to `if (fiber) {...}`?

pufferfish101007 (changes requested):

pufferfish101007 (changes requested):
Seems to work well with its intended use cases.

However, I would expect to be able to select the stroke options before drawing a shape (for instance while I have the rectangle or line tool selected, I might change the stroke colour before drawing, and would expect the visible stroke buttons to be useable).

Dark mode styling probably needs a bit of work - the text contrast is quite low and the buttons don't really match up with other buttons around them:


Looks good in light mode though.

Only a couple of very minor code comments and otherwise l;ooks great.

addons/paint-stroke-options/addon.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Stroke join and cap options",
+  "description": "Adds stroke join (miter/round) and cap (butt/round/square) controls to the color picker in the costume editor.",
```
pufferfish101007: Could it be specified that this is the outline colour picker (as opposed to the fill colour picker)?

addons/paint-stroke-options/userscript.js
griffpatch: Yeah it is.. I've added a comment

addons/paint-stroke-options/addon.json
griffpatch: Done

addons/paint-stroke-options/userscript.js
griffpatch: Yep thx

griffpatch:
> I would expect to be able to select the stroke options before drawing a shape (for instance while I have the rectangle or line tool selected, I might change the stroke colour before drawing, and would expect the visible stroke buttons to be useable).

Yeah, I didn't want to overcomplicate this PR - but you are right - I've added support for this now too, but it will need some good testing! :)

sad-birds:
This is cool! Would it make sense to also add UI for dashed lines? https://paperjs.org/reference/style/#dasharray

pufferfish101007:
The new changes seem to work great, thank you!

I've tested this out with the new round corners addon and it seems to work well which is good.

Minor problems with compat with the boolean shape operations addon - which could perhaps be fixed in a followup PR? - are that combine & expand cause round joins to become mitered.

Also re. name and description - could 'stroke' maybe be replaced with 'outline' if that is correct? That is probably more accessible to most scratchers. 'Corner' and 'end' might be more understandable that 'join' and 'cap', but again not sure if they're actually the same thing. Also a note explaining what a miter would probably be useful.

griffpatch:
Thanks for testing this and for the wording suggestions! I've updated the addon to use Scratch-facing terms. It's now called "Outline corners and ends", and the visible headings are "Corner" and "End". The icon tooltips explain the technical terms, including "Sharp corner (miter)" and "Flat end (butt)".

I agree the boolean shape operations issue is best handled in a follow-up PR. Those operations create replacement paths, so they should preserve the original outline's corner and end styles when doing so.

mxmou:
I have a very minor design suggestion: the added buttons might look better with no gap between them.

image

This would make them more similar to other button groups in Scratch, such as the zoom buttons in the paint editor and the stage size buttons in the stage header.

griffpatch:
> I have a very minor design suggestion: the added buttons might look better with no gap between them.
> 
> image
> This would make them more similar to other button groups in Scratch, such as the zoom buttons in the paint editor and the stage size buttons in the stage header.

As you wish!

griffpatch:
> Minor problems with compat with the boolean shape operations addon - which could perhaps be fixed in a followup PR? - are that combine & expand cause round joins to become mitered.

@pufferfish101007 - Talking of follow up PR for the shaping tool - did you want one for inclusion in this coming release if it's small or a followup release - I can look into fixing it sooner than later if that's what you meant :)

In case you want it - here it is: https://github.com/ScratchAddons/ScratchAddons/pull/9149

pufferfish101007:
Sorry, more minor problems with grouping and multiselection!

When objects are grouped together, you can still sort of select them individually using the reshape tool (or at least, the visual feedback tells me that I've only selected one object). However, trying to change the outline corner/end styles when such an object is selected doesn't work. Other operations such as changing the outline thickness do work as expected.

If it's difficult to make this work properly, I wouldn't be opposed to disabling the addon's buttons if such a selection is detected.

I also was experiencing the following: when multiple objects are selected (not necessarily grouped), the corner/end style can only be changed if all of the objects match on that property. However, I can't reproduce it any more! So either I'm hallucinating things (not unlikely), or something's working a little inconsistently.

griffpatch:
Thanks for spotting this! You can now change the outline corners and ends of an individual shape selected with the reshape tool inside a group. This also works for compound paths.

I couldn't reproduce the problem with changing mixed styles on ungrouped shapes, but I did find that the buttons could show a setting that wasn't shared by the whole selection. They now leave that setting unselected until you choose one.

I've also fixed something related: selecting an existing shape now remembers its corner and end settings for the next shape you draw. Previously, those settings were only remembered when you clicked the buttons yourself.

addons/paint-stroke-options/addon.json
```diff
@@ -0,0 +1,28 @@
...
+  ],
+  "userstyles": [
+    {
+      "url": "style.css",
+      "matches": ["projects"]
+    }
+  ],
+  "libraries": ["scratch-paint"],
+  "enabledByDefault": false,
+  "versionAdded": "1.45.0"
```
pufferfish101007, line 27: will need to be changed to 1.46.0 (or whatever's appropriate) when merged

pufferfish101007 (approved):

PR #8915, opened 2026-03-27
Vue 3 take 3

Samq64:
### Changes

Similar to #6272 (Thanks @GrahamSH-LLK!) except `dist` is not committed and is in the project root as `webpages-dist` instead so `webpages` may easily be removed with the packer script. This PR tries to keep changes and restructuring to a minimum, so no single file components (except for the license and permission pages) or moving everything to `src`.

### Reason for changes

Vue 3 has better performance and is the first major step in reducing tech dept in the settings page.

Even though I'm submitting this I still wonder if a rewrite is a better approach, even if also in Vue.

### Tests

Tested on Helium (based on Chromium 146). There are still some bugs that need to be ironed out.

### To do:
- [x] Find another colour picker
- [x] Fix setting components sometimes not rendering (also happened on the previous PR)
- [ ] Fill out new library licenses
- [ ] Load light theme and fonts earlier
- [ ] Check paths again, especially outside `webpages/`
- [ ] Find out why Prettier re-formats everything (or run it later)


webpages/settings/permissions.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+    <div class="navbar">
+      <a href="./settings.html"
+        ><img src="../../images/icon-transparent.svg" class="logo header-button" alt="Logo" draggable="false"
+      /></a>
+      <h1 v-cloak>{{ msg("permissions") }}</h1>
```
pufferfish101007: I don't think `v-cloak` is necessary inside template files as it's all precompiled

Samq64: I have removed it from the license and permission pages. The rest I will do in a followup PR.

Samq64:
The 2nd one is the current colour picker replacement:

color-picker-old
color-picker-new

Samq64:
Realistically I think the only viable option at this point is to merge a rewrite/redesign alongside the Vue 1 version and provide users with a way to switch between them for a couple releases.

mxmou:
> Realistically I think the only viable option at this point is to merge a rewrite/redesign alongside the Vue 1 version and provide users with a way to switch between them for a couple releases.

I don't think I understand why that would be better - can you explain?

Samq64:
You are right that is a little extreme and would add even more complexity and tech debt. My thought was it would eliminate the merge conflicts every time webpages/ is touched and it could be shipped without being perfect at first. Maybe it would be better to consider something like [vue3-sfc-loader](https://github.com/FranckFreiburger/vue3-sfc-loader) though or introduce a build step while keeping the current library versions first.

mxmou:
Some merge conflicts could be avoided by pausing other setting page work when the Vue 3 port is nearly complete. Or maybe even sooner - I'm not aware of any high-priority issues with the settings page right now, so focusing on addons instead for a while shouldn't be a big problem.

DNin01:
> Some merge conflicts could be avoided by pausing other setting page work when the Vue 3 port is nearly complete. Or maybe even sooner

To do that, we just need a plan—to make sure everyone knows when development for the Vue 1 webpages will close, how to use Vite, etc. I would relate to any development resource because I have barely used Node.js lol.

webpages/package.json
DNin01, line 1: I would personally add a top-level package.json to make it easier to add build steps to other parts of the extension in the future (you never know if we may add TS or Sass).

webpages/vite.config.js
DNin01, line 1: Changing where the webpages are found would break links, which isn't that much of a problem—every extension installation has a different base URL so it's impossible to externally link to them anyway—but if it changes, we will need to make sure all of our own links and redirects are correct.

We could also use the old directories for pages that redirect to new URLs.

DNin01 (commented):
Some suggestions for next time...

DNin01:
> Maybe it would be better to consider something like [vue3-sfc-loader](https://github.com/FranckFreiburger/vue3-sfc-loader) though

I'm afraid anything like that would require string evaluation which is blocked by Chrome's CSP for extensions.

Samq64:
Part of the problem is trying to keep the rest of the project build-free. I know this has been brought up many times but at our scale a build step will still be useful even without Vue 3 or language extensions. For example being able to `npm run lint` or `npm run format` without worrying about local tools, easier library management, removing automated translations from the source tree, etc. Yes, installing Node, running a couple of commands and the size of `node_modules ` may be annoying and it does raise the barrier of entry, but with hundreds of thousands of users and only a small team of volunteers no one really has time to teach new contributors basic developer skills any more. It's too bad because that low entry point is how I made it here and while I like to keep things simple when possible at some point the extra tooling outweighs that simplicity.

PR #8916, opened 2026-03-27
Feature/paint round corners

### LINKED ISSUE #4999: An addon that makes rounding corners in the costume editor easier

Right now, the only way to make rounded corners in shapes in the costume editor is to use the point editor tool, make two rounded (Bézier curve) points an equal distance from the sharp corner per non-rounded point, then delete the sharp corners. This can be a bit hard and time-consuming - it isn’t easy to get all the rounded points an equal distance from the sharp corner.

I always need to make rounded corners in my projects, especially in buttons and pop-ups.

I think there should be an addon that does exactly that for me - making a shape in the costume editor have rounded corners. The addon can be a button next to all the options (forward, backward, bring to front, etc) on top of the canvas in the costume editor. Upon clicking the button, I’m thinking a slider would show in which you can choose how rounded corners should be.

- A somewhat simpler solution would be to have a button next to “curved” and “pointed” shown when a pointed corner is selected that makes that specific point rounded (with an option controlling how much so).

### Additional context

None

griffpatch:
> ***Note added by @DNin01***
>
> This pull request incorporates code generated by Claude Sonnet.

# Changes

Adds a new [paint-round-corners] addon that lets you round the corners of vector shapes directly from the paint editor's **Reshape** tool - no separate mode or toolbar button.

image

Select a shape's points with Reshape as usual; wherever eligible corners exist among your selected points, a small draggable handle appears. Drag one to round that corner by inserting smooth Bézier control points. Handles follow Reshape's own point selection:
- Select specific points → only those corners' handles show, and dragging one moves just the selected set.
- Nothing selected (just the shape) → every corner's handle shows, and dragging any one rounds all of them together.
- Shift-click a handle to build/adjust a multi-selection.

Existing rounded corners can be re-opened and adjusted the same way. The overlay is inactive in bitmap mode.

*(Edit: this replaces the original design - a dedicated "Round Corners" toolbar button/mode - with a passive overlay integrated into Reshape, based on review feedback below about friction switching between the two tools.)*

# Reason for changes

Rounding corners on vector shapes currently requires manually placing Bézier handles on every node, which is tedious. This addon makes it a single drag operation per corner, and integrating it into Reshape means there's no extra tool to learn or switch to/from.

# Tests

Tested in Chrome:

- Round corners on a simple rectangle ✓
- Round corners on a complex polygon ✓
- Round individual corners independently, and all together when nothing is selected ✓
- Rounding is capped at max rounding distance calculated from distance between lines ✓
- Undo/redo works correctly ✓
- Switching away from Reshape (or tabs) hides the overlay cleanly; returning restores it ✓
- Whole-shape moves and point drags stay in sync with the overlay ✓
- Converting to bitmap hides the overlay ✓
- Dynamically enabling/disabling the addon works ✓


Joeclinton1 (changes requested):
It works great! Passed (almost) all the tests I threw at it. Love the user experience, clearly well thought out. The isArcPair math is quite nice.

There are two small issues though:

1. `Navigating away (code tab) and back restores correctly` this was not the case when I tried. What happened was that when I went to the code tab and then back it changed the tool, and then stopped me being able to select the tool. Trying to select the tool would select the circle tool instead.

2. If you change the rounding on any of the corners, and then you click on a single node. Sometimes it will select just that node, other times though it will not the select the node and just reset the rounding for all of the corners. The later case happening is annoying.

griffpatch:
> @Joeclinton1 It works great! Passed (almost) all the tests I threw at it. Love the user experience, clearly well thought out. The isArcPair math is quite nice.
> 
> There are two small issues though:
> 
> 1. `Navigating away (code tab) and back restores correctly` this was not the case when I tried. What happened was that when I went to the code tab and then back it changed the tool, and then stopped me being able to select the tool. Trying to select the tool would select the circle tool instead.

You are so right - great catch. I think it got introduced as I added the enable disable code before committing! - Should be fixed now (although, I auto re-select the rounding tool if you come back to the editor, but I'm not too bothered about that one right now)

> 2. If you change the rounding on any of the corners, and then you click on a single node. Sometimes it will select just that node, other times though it will not the select the node and just reset the rounding for all of the corners. The later case happening is annoying.

I suspect this is a miss click? - I've added a much larger hit zone to each node now to avoid the issue (I'm hoping)

addons/paint-round-corners/userstyle.css
```diff
@@ -0,0 +1 @@
+/* No custom styles needed: active state uses the native tool-select-base_is-selected class */
```
pufferfish101007: Redundant file can be removed unless the `cssText` for the SVGs is moved here

pufferfish101007 (changes requested):

Joeclinton1 (changes requested):
Bugs fixed:

1. The bug where it would deselect sometimes seems fixed by the larger hitbox so perhaps it was indeed a misclick.
2. The bug where changing to code tab and back would prevent you being able to use the Round Corners tool is fixed.

Bugs:

1. When clicking to the code editor and back, the tool does not stay as the Round Corners tool. (instead it changes you to the last tool you used before Round Corners)
2. I just discovered that zooming whilst using the Round Corners tool, will cause the overlaid nodes to not match the zoom. The behaviour should match that of the Reshape tool.

griffpatch:
> @Joeclinton1  Bugs fixed:
> Bugs:
> 
> 1. When clicking to the code editor and back, the tool does not stay as the Round Corners tool. (instead it changes you to the last tool you used before Round Corners)

Okay, I have added code to try to address this - We may need to rethink this into a more universal mechanism if we ever add more tools, but it will do the job here.

> 2. I just discovered that zooming whilst using the Round Corners tool, will cause the overlaid nodes to not match the zoom. The behaviour should match that of the Reshape tool.

Great catch - I fixed this on the gradient tool, but not for this one. Fixed now.

WorldLanguages:
For the record, and specially those that aren't on Discord, I want to highlight that griffpatch is developing these pull requests through Claude Code (an AI assisted coding tool). cc @mxmou @DNin01 

(so creating addons now costs money 😄)

griffpatch:
Yes, I am using Clause Sonnet 4.6 extensively in my development - I would
warn you though that it can appear like magic at giving you what you want,
and it truely is, BUT unless you know what you are doing, can instruct it
well, and understand what it has done and why, then just believing it has
done something right is a scary place to go. It requires a very good amount
of directing and prepping to ensure it can do what you want, and a lot of
coordination to help it debug along with you. Vibe coding is still in its
early days, but it's maturing fast. I find claude way way way more reliable
than ChatGPT was, and it really does a good job learning codebases and
keeping track of what you are doing.  The important thing is to know what
you 'are' doing before you get claude to do it for you, as like I say, it
will... and it may not work as you imagined. It will also quite
happily destroy your whole repo and change things it shouldn't if you don't
first tell it that it shouldn't. Before I began I went through a phase of
teaching it about scratch addons, and how I wanted to code.

DNin01:
Personally, I like the thought process of writing my own code and to be able to say that I did so, but the speed at which you can create software with AI is intriguing.

For our project, if licensing or attribution issues are a concern, I think that at the least we should ask that contributors disclose if they directly used AI output in their PRs.

griffpatch:
I am right there with you. There's nothing more that I love than planning
out and developing some great code and to know it's excellent and well
thought out. Using AI to code felt so weird when I started not long ago,
but it's something that has really taken hold in the corporate world, and
we are very much actively encouraged to use it.
I was intrigued by how well it could work with a repository like Scratch
Addons, and I had a number of addons I always wanted to investigate, so
that's where my adventure began. I tried to ensure they were fully
contained and didn't impact other parts of the repo, and have been careful
to be very specific about what they do and how they do it.
Regarding attribution, your idea is sound. I believe many projects now
require developers to disclose their use of AI in pull requests (PRs) and
to add this requirement to their policies.

Joeclinton1 (approved):
The two bugs are now fixed, so it looks good to me!

addons/paint-round-corners/userscript.js
```diff
@@ -0,0 +1,844 @@
...
+        ? outDirVec.normalize()
+        : nextB.handleOut.length > EPS
+          ? nextB.handleOut.normalize()
+          : null;
+    if (!outDir) return false;
+    if (outDir.dot(B.handleIn.multiply(-1).normalize()) < 0.85) return false;
+    return true;
+  };
+
+  // Returns true when segment A and the following segment B form a rounded corner
```
pufferfish101007: This comment doesn't match the actual return value(s) of the function

pufferfish101007 (commented):
A few bits of unexpected behaviour here:

What's the intended behaviour for concave corners? The addon seems to inconsistently detect them, and sometimes gives some weird artifacts when 'rounding' them. (Video: note artifacts when changing the concave corner, and notice also the concave corner on the left of Scratch Cat's head that is not treated as a corner by the addon)

https://github.com/user-attachments/assets/df25d2c3-32a2-46c3-afcd-47ec30d7be89

---

If I make manually rounded corners using the existing 'Reshape' tool, even if I make these quite sharp, they don't seem to get detected by the addon:

Image

---

If I select a shape with the rounding corners tool, then switch costumes, the corner selectors are still visible:

https://github.com/user-attachments/assets/48439153-8367-4168-acb5-3eb8feb885d6

---

Switching between reshape and round corners is a bit strange: going from reshape to round corners, the reshape handles are still visible but don't do anything; going from round corners to reshape, the handles don't appear:

https://github.com/user-attachments/assets/de1c364b-a965-4066-8b8a-46a95d71d888

addons/paint-round-corners/userscript.js
griffpatch: Oops, that was a hanging comment from a deleted function

griffpatch:
> A few bits of unexpected behaviour here:
> 
> What's the intended behaviour for concave corners? The addon seems to inconsistently detect them, and sometimes gives some weird artifacts when 'rounding' them. (Video: note artifacts when changing the concave corner, and notice also the concave corner on the left of Scratch Cat's head that is not treated as a corner by the addon)

@pufferfish101007 The addon only detects corners where two straight lines are connected using a single curved line. The examples where you can't see the corner detected are due to the corner meeting another curved line. This is expected behavior because you can't extend a corner curve into another curved line with consistent results.

pufferfish101007:
Just FYI I am unlikely to have much time to properly review this or #8917 for a couple of weeks - but I will review once I find time if nobody else has by then :) (I don't want to delay these getting merged, but I also don't want to rush a review.)

pufferfish101007:
Thanks for the clarification for why some corners aren't editable.

The core functionality works well but there are a couple of minor bugs still remaining:

> If I select a shape with the rounding corners tool, then switch costumes, the corner selectors are still visible:
>
> https://github.com/user-attachments/assets/48439153-8367-4168-acb5-3eb8feb885d6

> Switching between reshape and round corners is a bit strange: going from reshape to round corners, the reshape handles are still visible but don't do anything; going from round corners to reshape, the handles don't appear:
>
> https://github.com/user-attachments/assets/de1c364b-a965-4066-8b8a-46a95d71d888

In particular going from reshape to round corners still doesn't remove the reshape handles, and when going from round corners to reshape, the reshape handles don't appear consistently (sometimes I have to move my mouse or pan the canvas to get them to appear).

Also a bug that's either new or that I just didn't notice before, is that the round corner nodes appear on top of the editor controls if panned up or across too far. Hopefully this is just a z-index problem and shouldn't need a more complicated fix.

image

---

@griffpatch if you'd like to try to get this merged soon but don't have time to look at these I'd be happy to try and write up fixes this weekend (in theory I will have some time). (If you'd prefer to do it yourself then that's also cool; I won't work on anything unless you would like me to)

pufferfish101007:
Also if you're preferring to leave any weird edge cases unfixed because the vanilla paint editor is weird enough (as mentioned in https://github.com/ScratchAddons/ScratchAddons/pull/8917#issuecomment-4292678458), and other core members are happy with that, then I will happily desist from going on about tiny bugs :)

griffpatch:
No thanks for these - I will take a look when I can!

griffpatch:
@pufferfish101007 - I have fixed the nodes being visible as they move outside the canvas area

griffpatch:
@pufferfish101007 Thanks again for the detailed testing - all three of these are fixed now:

1. Costume switch no longer leaves the corner handles stuck on screen.
2. The Reshape ↔ Round Corners handle issues (stale reshape handles, handles not reappearing) are gone - see below, this whole class of bug went away.
3. The z-index issue when panned/zoomed far was an `overflow:visible` on the widget's SVG container; fixed by clipping it to the canvas bounds like the native canvas already does.

While fixing #2 I ended up going further: I've merged corner rounding directly into the **Reshape** tool instead of keeping it as a separate mode/toolbar button. There's no more "Round Corners" button to switch to/from at all now - select a shape's points with Reshape as usual, and a draggable handle appears wherever there's an eligible corner among your selected points:

- Select specific points → only those corners' handles show; dragging one moves just that selection.
- Nothing selected (just the shape) → every corner's handle shows, and dragging any one rounds them all together.
- Shift-click to build/adjust a multi-selection, same as before.

Since there's no longer a second mode to switch into/out of, the stale-handle and mode-switch bugs are structurally impossible now rather than patched over. I've updated the PR description above to match. Would appreciate another pass whenever you have time - no rush, and thanks for your patience on this one!

CST1229:
> —

> —

> —

> —

what

em dashes???

addons/paint-round-corners/userscript.js
```diff
@@ -295,17 +377,37 @@ export default async function ({ addon }) {
...
+            if (c) {
+              pathCornersAll.push(c);
+              if (!anyNodeSelected || path.segments[i].selected) {
+                pathCornersSelected.push(c);
+              }
+            }
           }
         }
+
+        if (!anyNodeSelected) {
```
pufferfish101007: This and last branch are the same, could they be reordered to avoid duplication?

addons/paint-round-corners/addon.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Round path corners",
+  "description": "Adds interactive corner-rounding handles to vector shapes in the paint editor's Reshape tool. Select a shape's points with Reshape to reveal draggable handles at its eligible corners, then drag one to round it. Selecting specific points rounds just those corners; with no points selected, dragging any handle rounds every corner together.",
```
pufferfish101007: Could you add an explanation/note about how to select nodes (shift click)? I don't think that's necessarily easy to figure out on your own

pufferfish101007 (changes requested):
One minor code comment and an addon description request. [d7941ad](https://github.com/ScratchAddons/ScratchAddons/pull/8916/commits/d7941ad8a2a4fc8e589ba7640469a91b0f00d5f7) & [a977194](https://github.com/ScratchAddons/ScratchAddons/pull/8916/commits/a977194018bb62643f7e3f859c0495545dafc87e) are too involved with the inner workings of the addon for me to properly think about, but they don't look as if they'll break anything; I'm happy with all the other changes so I'm not going to do another sweep through all of the code.

This behaviour is nice and seems mostly bug-free! The only problem I've found is that if I have the reshape tool selected, disable the addon, and reenable it, then I need to switch to another tool before the rounding handles appear again:

https://github.com/user-attachments/assets/1d5c540d-0320-4852-932c-0e7c60592703

This isn't a big problem so I'm happy to give an approving review once the addon description comment is addressed :)

addons/paint-round-corners/userscript.js
griffpatch: I will update this, good catch

addons/paint-round-corners/addon.json
griffpatch: Done

pufferfish101007 (approved):
Thanks! Am happy to approve.

@Joeclinton1 are you happy for your approval to still stand after the changes described in https://github.com/ScratchAddons/ScratchAddons/pull/8916#issuecomment-4876614600? If so this can be merged

Joeclinton1:
> Thanks! Am happy to approve.
> 
> @Joeclinton1 are you happy for your approval to still stand after the changes described in [#8916 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/8916#issuecomment-4876614600)? If so this can be merged

I'm very busy right now, but it would be wrong of me to approve this without looking at the update changes. I'm sorry @griffpatch the correct thing for me to do is remove my review and review again. Fortunately, I am going to update this task to my highest priority, to avoid further delays.

Codex says these are the changes since my review so will be focusing on them:

- Major redesign: Removed the separate Round Corners tool and integrated rounding handles directly into the native Reshape tool.

- Selection behaviour: Specific selected nodes affect only those corners; no selected nodes means all corners move together. Selection is now preserved when corners split into or merge from Bézier arcs.

- Geometry improvements: Better support for curved neighbouring edges, adjacent rounded corners, different starting radii, and maximum-radius constraints.

- UI/lifecycle fixes: Handles now stay inside the canvas and update correctly across costume changes, tabs, undo/redo, zooming, panning and bitmap conversion.

- Contributor requests: Added Shift-click instructions and removed duplicated code. All requested review threads are resolved.

- Review priority: Carefully review d7941ad and a977194; pufferfish did not deeply assess their selection and geometry logic.

- Known minor bug: Re-enabling the addon while Reshape is already selected may require switching tools before handles return. Pufferfish considered this non-blocking and approved.

Joeclinton1:
Have tested all the changes, read through the description, considered the ui and got codex to do a review too. 

There is
1. the known bug of the dynamic re-enable not working if you are already on reshape. 
2. Bug where if you make a corner rounded then alter the curve with bezier handles, the corner handle still exists. If you then try to use it, it undoes the curve change, and re-evens the two sides to be a clean arc again.

These are both quite minor. @pufferfish101007 should we just ignore these? I think fixing bug 2 is harder than it sounds. and for little gain.

griffpatch:
> Have tested all the changes, read through the description, considered the ui and got codex to do a review too.
> 
> There is
> 
> 1. the known bug of the dynamic re-enable not working if you are already on reshape.

Fixed in last push - thanks

> 2. Bug where if you make a corner rounded then alter the curve with bezier handles, the corner handle still exists. If you then try to use it, it undoes the curve change, and re-evens the two sides to be a clean arc again.

This is by design - I tried to make it resilient for Scratchers who might not understand that reshaping might break the corner rounding.

Joeclinton1 (approved):
Thank you for that change and for clarification about the apparent second bug. I am happy to approve :)

addons-l10n/en/paint-round-corners.json
DNin01: This file can be removed.

pufferfish101007:
Some final thoughts before merging:
- The average scratch user probably doesn't know what a path is. Similar addons mention the costume editor explicitly in the name, so I'd suggest something like `Round corners in costume editor`
- The description is quite long compared to similar addons. We should also probably refer to 'costume editor' rather than 'paint editor' for consistency with other addons. I would suggest something like `Use the Reshape tool in the costume editor to reveal draggable handles for rounding a shape's corners. Shift-click to select specific corners; with none selected, all corners are rounded together.` This is similar in style to `paint-skew`, which reads more as instructions rather than an extended description, and is also a more similar length to other addon descriptions.
- I think this should have `"relatedAddons": ["paint-skew"]` in the addon manifest
- Worthy of `featured` tag? not sure what the process of getting that is

Objections/affirmations/suggestions would be appreciated :)

DNin01 (commented):
Here are my thoughts:

- It feels **a little advanced** for the Scratch paint editor but is not a bad idea for an addon. If it didn't exist, I already have ways of creating round-ish angles and would probably look for a more professional graphic editor if I needed to do something more advanced.
- The tool makes sense with shapes like rectangles and regular polygons, but on **more complex shapes with more curvature**, it's hard to tell what will get recognized as a "corner" angle or if the rounding will work in the way you want.
- I don't really like how all the angles get rounded at once if you don't first single- or multi-select those that you want to adjust, **particularly on more complex shapes where the angles are not all congruent**. I would rather explicitly select all the angles I want to adjust. That would be less efficient but more intuitive, less destructive, and more engaging for designers like me who are fond of manual creative processes. A "select all corners" button as well would be fine, though.

DNin01:
> * Worthy of `featured` tag? not sure what the process of getting that is

I don't think it's something that the majority of Scratchers need or have thought to ask for, so I am personally not sure about making it featured.

I even think it wouldn't be a bad idea to make it beta until we collect some feedback.

Joeclinton1:
> > * Worthy of `featured` tag? not sure what the process of getting that is
> 
> I don't think it's something that the majority of Scratchers need or have thought to ask for, so I am personally not sure about making it featured.
> 
> I even think it wouldn't be a bad idea to make it beta until we collect some feedback.

I have wanted it for a while

Samq64:
Some addons have usage documentation on the website but there is no UI on the settings page to open it yet.

Joeclinton1:
> Here are my thoughts:
> 
> * It feels **a little advanced** for the Scratch paint editor but is not a bad idea for an addon. If it didn't exist, I already have ways of creating round-ish angles and would probably look for a more professional graphic editor if I needed to do something more advanced.
> * The tool makes sense with shapes like rectangles and regular polygons, but on **more complex shapes with more curvature**, it's hard to tell what will get recognized as a "corner" angle or if the rounding will work in the way you want.
> * I don't really like how all the angles get rounded at once if you don't first single- or multi-select those that you want to adjust, **particularly on more complex shapes where the angles are not all congruent**. I would rather explicitly select all the angles I want to adjust. That would be less efficient but more intuitive, less destructive, and more engaging for designers like me who are fond of manual creative processes. A "select all corners" button as well would be fine, though.

You cannot come in right at the end of development and suggest anything about ui. That's not how things are done. The review right now is about the code and if the code works and the most important stuff, like the description/info/tags.

mxmou:
> I don't really like how all the angles get rounded at once if you don't first single- or multi-select those that you want to adjust, **particularly on more complex shapes where the angles are not all congruent**. I would rather explicitly select all the angles I want to adjust. That would be less efficient but more intuitive, less destructive, and more engaging for designers like me who are fond of manual creative processes. A "select all corners" button as well would be fine, though.

Changing a single corner by default would probably be more intuitive. But the most common use case for this addon will be rounded squares and rectangles, where all corners typically have the same radius.

> You cannot come in right at the end of development and suggest anything about ui. That's not how things are done. The review right now is about the code and if the code works and the most important stuff, like the description/info/tags.

I'd argue UI is one of the things that can be tweaked near the end of the development because it doesn't usually require big code changes. It's also really important to get UI right, because once it's released, it becomes hard to change without making it confusing for people who have already learned how to use the initial version.

Joeclinton1:
> I'd argue UI is one of the things that can be tweaked near the end of the development because it doesn't usually require big code changes. It's also really important to get UI right, because once it's released, it becomes hard to change without making it confusing for people who have already learned how to use the initial version.

Design comes first, then code. 3 months ago was the time to propose changes to the ux, ui. We are literally about to merge and people come in asking for changes. Unprofessional. Ux and Ui wise we approved it.

pufferfish101007:
> * It feels **a little advanced** for the Scratch paint editor but is not a bad idea for an addon.

> I don't think it's something that the majority of Scratchers need or have thought to ask for, so I am personally not sure about making it featured.

FWIW, this is the 3rd most upvoted suggestion on the archive from the old community discord server, so isn't without precedent: https://sa-community.github.io/suggestions-archive/


> * I don't really like how all the angles get rounded at once if you don't first single- or multi-select those that you want to adjust... I would rather explicitly select all the angles I want to adjust. ... A "select all corners" button as well would be fine, though.

Perhaps a setting to toggle this behaviour? I can understand why people might want it either way.

> * The tool makes sense with shapes like rectangles and regular polygons, but on **more complex shapes with more curvature**, it's hard to tell what will get recognized as a "corner" angle or if the rounding will work in the way you want.

A note in the addon manifest would be useful here if you are able to concisely describe the process by which corners are detected @griffpatch? Alternatively as @Samq64 mentioned, addon docs on the website might be worthwhile for this (if we also add a way to link to the docs from the settings page)

DNin01:
> Design comes first, then code. 3 months ago was the time to propose changes to the ux, ui.

True, but it's not the only time. Design iterations often don't happen in one phase.

Design first is a useful principle, but sometimes you'll finish the code and then your testers will realize something in the design that isn't quite right, something that wouldn't have been caught in the design phase. Things come up, and it's OK if something important enough causes a small delay.

And don't panic—my suggestion to make the rounding tool a little more beginner friendly _is_ just one button on top of the UI that was already made, not a full UI rework. And since everything up until then was approved, we don't have to rethink anything else.

DNin01:
> > * I don't really like how all the angles get rounded at once if you don't first single- or multi-select those that you want to adjust... I would rather explicitly select all the angles I want to adjust. ... A "select all corners" button as well would be fine, though.
> 
> Perhaps a setting to toggle this behaviour? I can understand why people might want it either way.

Sure. I would name it "automatically select all corners" and have it off by default. That wouldn't require any UI changes within the addon except for all the handles to start out solid blue.

griffpatch:
I wouldn't assume off by default is the best choice... It was carefully
designed to work like it does for the normal case. It follows how the same
feature works in other applications and once you get used to it will become
second nature.

DNin01:
> It follows how the same feature works in other applications and once you get used to it will become second nature.

In that case, maybe the addon description could refer by name to another application that has this feature (e.g., "like in Affinity Designer") to tell those users to expect a familiar experience and imply to other users that this is a more advanced feature.

griffpatch:
Thanks for all the feedback. I've updated the manifest to:

- rename the addon to `Round corners in costume editor`
- shorten the description and use "costume editor" consistently
- add `paint-skew` as a related addon

I've left the current selection behaviour as it is. Rounding all corners together is the common case this was designed around, particularly for rectangles, and the description now explains how to select specific corners when needed.

I have also left out the featured or beta tags and the Adobe illistrator reference for now.

DNin01 (approved):

Joeclinton1:
Woop! Good job on getting it merged!

griffpatch:
Woohoo!!! :) - How exciting! - Thanks everyone for the reviews, suggestions
and support

PR #8917, opened 2026-03-27
Feature/paint boolean ops

### LINKED ISSUE #8052: SVG set operations
Union, difference, intersection and more. Would help with many things.

griffpatch:
> ***Note added by @DNin01***
>
> This pull request incorporates code generated by Claude Sonnet.

### Changes

Adds a new **Boolean shape operations** addon to the vector costume editor.

A **Shaping** section is injected into the editor toolbar with two layout modes: inline (wide toolbar, five buttons sit directly in the bar) and collapsed (narrow toolbar, a single trigger button opens a hover dropdown panel).

image

image

**Operations added:**

| Button | Shortcut | Behaviour |
|---|---|---|
| Unite | click | Merges all selected shapes into one, keeping the backmost shape's fill/stroke |
| Subtract | click | Cuts the upper shape(s) out of the backmost shape; Shift = Punch Through (cuts all other shapes); Alt = keep cutter |
| Intersect | click | Keeps only the overlapping region; Alt = Divide (splits all shapes into every distinct non-overlapping fragment, Illustrator-style) |
| Combine / Release | click | Combines selected paths into a CompoundPath with even-odd fill (punch-through holes), or releases a CompoundPath back into individual paths |
| Expand | click | Expands each shape by half its stroke width, converting the visual stroke to a filled outline |
| Open / Close Path | Reshape mode context bar | Opens a closed path at the selected anchor, duplicating the break-point so both endpoints overlap at the same position; closes an open path, merging endpoints if they are within 5 units |

**Additional details:**

- Text-to-path: selected text items are automatically converted to path outlines before any boolean operation, using a rasterise → threshold → marching-squares trace → Bézier curve-fit pipeline (we could add a hefty library to the dependencies to do this natively?)
- Group awareness: Subtract, Intersect, and Punch Through descend recursively into paper.js Groups, preserving the original nesting structure
- Style preservation: fill/stroke/strokeWidth are snapshotted before and restored after every operation (paper.js boolean ops silently lose styles)
- Scratch open-circle fix: Scratch stores circles as open paths whose last point coincides with the first; a pre-processing step detects and closes these before boolean ops to avoid corrupt results
- Dynamic enable/disable: the shaping section and mode-tools button are removed when the addon is disabled; button enabled states update live from the current paper.js selection

### Reason for changes

The vector costume editor has no boolean shape tools. Creating complex shapes — cutouts, compound letters, logos, overlapping fills — requires tediously cutting paths by hand with the Reshape tool. Every professional illustration app (Illustrator Pathfinder, Inkscape Path menu, Figma boolean groups) provides these operations as first-class features. This addon brings the full set directly into the editor toolbar.

### Tests

Tested in Chrome:

- Unite, Subtract, Intersect on two plain paths
- Subtract+Shift (Punch Through) on a stack of three shapes
- Subtract+Alt / Intersect+Alt (keep cutter / Divide)
- Combine and Release round-trip on a CompoundPath
- Expand on paths with and without a stroke
- Open/Close Path in Reshape mode, including handle preservation at the join and the 5-unit merge snap
- Text-to-path conversion triggered via Unite and Subtract on PointText items
- Group-aware Subtract where the target is a paper.js Group
- All operations on Scratch circles (pseudo-open paths) — no corrupt output
- Dynamic enable/disable via the addon toggle
- Inline ↔ collapsed toolbar layout in response to toolbar width


TSFBCE24RhythmHeaveners:
Yay
It's a Scratch Tools Addon Ported Intro Scratch Addones :smiley:

griffpatch:
None of these are ports, I'd be interested to see how they compare though!
I spent a lot of time getting mine to do some cool stuff!

Joeclinton1 (changes requested):
Everything (almost) works! Amazing addon. I'm super impressed with the code inside and how this handles the edge cases. 

Bugs:
1. Alt drag should duplicate an object and drag the duplicate only, but alt drag on a shape with a hole, duplicates and drags both parts.
2. if you click an object and then deselect it by clicking itself with shift (or sometimes ctrl) held down, the boolean operations will be grayed out the next time you click a shape. This doesn't always happen, but it is reproducible.

Ux change:
Personally the functionality of combine/release/expand is not clear without reading the PR description. In the addon description could you describe the different boolean features? Also maybe the tool tip could give some info.

TSFBCE24RhythmHeaveners:
> None of these are ports, I'd be interested to see how they compare though! I spent a lot of time getting mine to do some cool stuff!

No. It's not a port, said by Griff. It's a recreation, but half a port.

griffpatch:
> @Joeclinton1 Everything (almost) works! Amazing addon. I'm super impressed with the code inside and how this handles the edge cases.
> 
> Bugs:
> 
> 1. Alt drag should duplicate an object and drag the duplicate only, but alt drag on a shape with a hole, duplicates and drags both parts.

This is actually a Scratch bug - If you import any SVG with a compound share, alt dragging it will show a similar issue - It's also present in Penguin Mod and other costume editors. However, I hate that bug too and I've found a solution so I've included a patch to Scratch's paint editor bug in this addon (can be turned off if wanted with a toggle)

> 2. if you click an object and then deselect it by clicking itself with shift (or sometimes ctrl) held down, the boolean operations will be grayed out the next time you click a shape. This doesn't always happen, but it is reproducible.

I've not been able to reproduce this one myself yet.

> Ux change: Personally the functionality of combine/release/expand is not clear without reading the PR description. In the addon description could you describe the different boolean features? Also maybe the tool tip could give some info.

Agreed... I just didn't want to over-do it with lots of explanation text. I'll have a think

griffpatch:
> > None of these are ports, I'd be interested to see how they compare though! I spent a lot of time getting mine to do some cool stuff!
> 
> No. It's not a port, said by Griff. It's a recreation, but half a port.

Like I said - not a port, not one line in fact and I'd not seen Penguin mod until after this PR was raised.

TSFBCE24RhythmHeaveners:
That's not nice. Is it like you are using PenguinMod? or is it original?

CubesterYT:
> That's not nice. Is it like you are using PenguinMod? or is it original?

This PR is completely unrelated to PenguinMod, he stated he wasn't aware of the mod till now. This is a completely original implementation. Please stop bugging him.

addons/paint-boolean-ops/text-to-path.js
```diff
@@ -0,0 +1,654 @@
...
+  static convert(textItem, paper) {
+    try {
+      return this.#run(textItem, paper);
+    } catch (e) {
+      console.error("[TextToPath]", e);
+      return null;
+    }
+  }
+
+  // ── Private ─────────────────────────────────────────────────────────────
```
pufferfish101007: This comment isn't particularly helpful as there are private items both above and below

addons/paint-boolean-ops/text-to-path.js
```diff
@@ -0,0 +1,654 @@
...
+    // The canvas spans [bounds.x, bounds.x + bounds.width] × [...height]
+    const ox = bounds.x,
+      oy = bounds.y;
+    const sx = cw / bounds.width,
+      sy = ch / bounds.height;
+
+    // 2. Binary grid: 1 = inside (alpha above threshold) ─────────────────
+    // Optional: pre-blur alpha with a separable 3×3 Gaussian (σ≈0.7px) before
+    // thresholding. Cuts marching-squares staircase amplitude but can soften
+    // sharp corners slightly. Controlled by #PRE_BLUR.
```
pufferfish101007: Looks like `PRE_BLUR` is always false, so one branch is dead code

addons/paint-boolean-ops/text-to-path.js
```diff
@@ -0,0 +1,654 @@
...
+  // Enable a separable 3×3 Gaussian pre-blur (σ≈0.7px) before thresholding.
+  // Smooths the anti-aliased fringe, reducing staircase noise but can soften
+  // sharp corners very slightly. Set to false to threshold raw alpha directly.
+  static #PRE_BLUR = false;
+
+  // Set to true to log detailed per-corner diagnostics to the browser console.
+  // Shows raw marching-squares pts, post-smooth pts, corner dot products, and
+  // the final bezier segments produced — making it easy to see which stage
+  // introduces corner rounding or spurious anchors.
+  static #DEBUG = false;
```
pufferfish101007: Should this be removed before merging?

addons/paint-boolean-ops/text-to-path.js
```diff
@@ -0,0 +1,654 @@
...
+    if (pts.length < 3) return pts;
+    const [x1, y1] = pts[0];
+    const [x2, y2] = pts[pts.length - 1];
+    const len = Math.hypot(x2 - x1, y2 - y1);
+    let maxD = 0,
+      maxI = 0;
+    for (let i = 1; i < pts.length - 1; i++) {
+      const [px, py] = pts[i];
+      const d =
+        len < 1e-10
```
pufferfish101007, line 553: Would it be possible to add a comment to justify why this small number rather than any other small number, and why does using this particular formulation of `Math.hypot` make sense here?

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+  shapingLbl.className = "sa-shaping-label";
+  shapingLbl.textContent = msg("shaping");
+  shapingBtn.appendChild(shapingIcon);
+  shapingBtn.appendChild(shapingLbl);
+  shapingSection.appendChild(shapingBtn);
+
+  // Dropdown panel — fixed-position popup shown on hover in collapsed mode.
+  const shapingDropdown = document.createElement("div");
+  shapingDropdown.className = "sa-shaping-dropdown";
+  shapingDropdown.style.cssText = [
```
pufferfish101007: Could this be put in a separate css file that then uses the class selector?

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+    if (closeTimer) {
+      clearTimeout(closeTimer);
+      closeTimer = null;
+    }
+    const r = shapingBtn.getBoundingClientRect();
+    shapingDropdown.style.left = r.left + "px";
+    shapingDropdown.style.top = r.bottom + 2 + "px";
+    shapingDropdown.style.display = "flex";
+  };
+  const scheduleClose = () => {
```
pufferfish101007: If the `mouseenter` event checks if the addon is disabled, should this also check if the addon is disabled?

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+    else performBooleanOp(op, e.altKey);
+  };
+  shapingSection.addEventListener("click", handleClick);
+  // Note: shapingDropdown is a DOM child of shapingSection so clicks already
+  // bubble up to the section listener above — no second listener needed.
+
+  // ── Enable/disable buttons based on current paper.js selection ─────────
+  const updateButtonStates = async () => {
+    if (!modDisabledClass) {
+      const anyDisabled = document.querySelector("[class*='button_mod-disabled']");
```
pufferfish101007:
```suggestion
      const anyDisabled = document.querySelector("[class*='button_mod-disabled_']");
```

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+    // unite: optionally convert any selected text items to path outlines first.
+    let convertedText = 0;
+    if (opName === "unite") convertedText = convertTextItems(paper);
+    // flat selection, chain back→front into one result.
+    const selected = getPaintingSelected(paper);
+    if (selected.length < 2) {
+      // Single converted text item: no unite needed, just save.
+      if (convertedText > 0 && selected.length === 1) {
+        triggerUpdateImage();
+        setTimeout(updateButtonStates, 50);
```
pufferfish101007: Why a 50ms timeout here? If there's something in particular that needs to be waited for, is it possible to trigger the button update directly when the state is ready?

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+      }
+    }
+
+    triggerUpdateImage();
+    setTimeout(updateButtonStates, 50);
+  };
+
+  // ── Trigger scratch-paint undo snapshot ───────────────────────────────
+  const triggerUpdateImage = () => {
+    const canvasContainer = document.querySelector("[class^='paint-editor_canvas-container']");
```
pufferfish101007:
```suggestion
    const canvasContainer = document.querySelector("[class*='paint-editor_canvas-container_']");
```

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+  const injectModeToolsBtn = () => {
+    const mode = addon.tab.redux.state?.scratchPaint?.mode;
+    if (mode !== "RESHAPE") {
+      modeToolsOCBtn.remove();
+      return;
+    }
+
+    // Find the dashed-border group (contains Curved + Pointed in Reshape mode).
+    const dashedGroup = document.querySelector(
+      "[class*='mode-tools_mode-tools'] [class*='mode-tools_mod-dashed-border']"
```
pufferfish101007:
```suggestion
      "[class*='mode-tools_mode-tools_'] [class*='mode-tools_mod-dashed-border_']"
```

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+  // ── Inject both sections into the fixed-tools row ─────────────────────
+  // We wait for the Front/Back sub-row specifically, then insert our sections
+  // immediately after it. Class names are read from live DOM elements at
+  // runtime so any future hash changes are handled automatically.
+  const toolsLoop = async () => {
+    let hasRunOnce = false;
+    let lastFrontBackRow = null;
+    const updateLayout = () => (window.innerWidth >= 1600 ? setInlineMode() : setCollapsedMode());
+    while (true) {
+      const frontBackRow = await addon.tab.waitForElement("[class^='fixed-tools_row'][class*='input-group']", {
```
pufferfish101007:
```suggestion
      const frontBackRow = await addon.tab.waitForElement("[class*='fixed-tools_row_'][class*='input-group_']", {
```

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1169 @@
...
+  //
+  // window blur / visibilitychange do not reliably fire when Alt+Tab shifts
+  // OS focus away in Chrome on Windows, so we can't use those.
+  //
+  // Instead, intercept mousedown in capture phase — which runs before
+  // paper.js reads modifiers — and sync paper.Key.modifiers against the
+  // native event's real hardware state. MouseEvent.altKey/shiftKey etc.
+  // always reflect actual key state regardless of what paper.js cached.
+  // Guarded by the "fix-stuck-modifiers" setting.
+  {
```
pufferfish101007: Redundant blocks here and below

pufferfish101007 (changes requested):
This is very exciting. I've left some comments, primarily regarding code style. I think the standard is for class selectors to be written as `[class*="class_class-name_"]` with a trailing underscore but I didn't make a comment for every occurrence. I trust that the various algorithms work!

TSFBCE24RhythmHeaveners:
I tested it and It felt like PenguinMod's.

griffpatch:
> I tested it and It felt like PenguinMod's.

It's not surprising, boolean operations are a standard feature in vector paint editors, and the core functionality is actually already built into the paper.js framework that scratch uses for the vector paint editor in Scratch. Both of our addons / extensions are mostly just surfacing that functionality that Scratch never did.  Mine (I hope) just does it in a more useable / reliable way.

TSFBCE24RhythmHeaveners:
Oh. :sob:

addons/paint-boolean-ops/userscript.js
griffpatch: No, it doesn't need the guard - closing is always valid and desirable (including when the addon is disabled mid-session). The mouseenter check specifically prevents opening the dropdown after disable, which is the only case requiring the guard.

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1161 @@
...
+      const fixedToolsRow = frontBackRow.parentElement;
+      closeDropdown(); // ensure panel is closed on each reinsertion
+      frontBackRow.after(shapingSection);
+      updateLayout(); // re-evaluate inline vs collapsed after (re)insertion
+
+      if (!hasRunOnce) {
+        hasRunOnce = true;
+
+        // Extract layout and separator classes from native toolbar elements.
+        const nativeDashedGroup = fixedToolsRow.querySelector("[class*='mod-dashed-border']");
```
pufferfish101007: Another trailing underscore needed here

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1161 @@
...
+  // OS focus away in Chrome on Windows, so we can't use those.
+  //
+  // Instead, intercept mousedown in capture phase — which runs before
+  // paper.js reads modifiers — and sync paper.Key.modifiers against the
+  // native event's real hardware state. MouseEvent.altKey/shiftKey etc.
+  // always reflect actual key state regardless of what paper.js cached.
+  // Guarded by the "fix-stuck-modifiers" setting.
+  const _syncModifiers = (e) => {
+    if (addon.self.disabled) return;
+    if (!addon.settings.get("fix-stuck-modifiers")) return;
```
pufferfish101007: Is there ever a case where we wouldn't want this setting to be enabled?

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,1161 @@
...
+
+    if (_selectedDesc?.set) {
+      const _origSet = _selectedDesc.set;
+      const _origGet = _selectedDesc.get;
+      const _addon = addon;
+      Object.defineProperty(_fixPaper.CompoundPath.prototype, "selected", {
+        get: _origGet,
+        set: function (selected) {
+          _origSet.call(this, selected);
+          if (!selected && !_addon.self.disabled && _addon.settings.get("fix-compound-deselect")) {
```
pufferfish101007: Same with this setting

pufferfish101007:
Selecting scratch cat's head and using "expand" makes the selected group invisible. Not sure why.


https://github.com/user-attachments/assets/105e694f-8baa-4836-8d46-58e06c47a8cb

This doesn't happen if that group is ungrouped.

I can inconsistently reproduce this on a smaller sample with two shapes of the same colour, where one overlaps entirely with the other:


https://github.com/user-attachments/assets/906ee70f-6323-45c7-8eda-6ef796783589

pufferfish101007:
When combining scratch cat's head, the 'combine' button doesn't change to 'release' until the selected shapes are deselected and then reselected.

https://github.com/user-attachments/assets/f5a85f9d-27d6-41ab-a50f-3943db59adfd

pufferfish101007:
A super minor point but the disabling/enabling of the shaping button is noticeably delayed from the buttons next to it. I suspect there's nothing that can be done about that without full access to the react internals but it'd be cool if there is.

addons/paint-boolean-ops/userscript.js
Samq64: The trailing underscore doesn't matter much, it's just what addons generally do when matching react classes.

addons/paint-boolean-ops/userscript.js
griffpatch: So I would expect not, except that I just wanted to be careful in case it ever broke anything upstream... - I'm happy to remove the fix settings and just fix them quietly...? - Fixing bugs in Scratch is a slightly dangerous business!!!

griffpatch:
> Selecting scratch cat's head and using "expand" makes the selected group invisible. Not sure why.
> I can inconsistently reproduce this on a smaller sample with two shapes of the same colour, where one overlaps entirely

I think I have got to the bottom of these now (it was due to the shapes being recreated using the groups outline and fill... which were nothing!

griffpatch:
> A super minor point but the disabling/enabling of the shaping button is noticeably delayed from the buttons next to it. I suspect there's nothing that can be done about that without full access to the react internals but it'd be cool if there is.

I have to delay by one animation frame to wait for the react DOM to be updated. I can't sync in in real time before the update sadly. But, I think I have at least improved the responsiveness so it is more consistently enabled/disabled.

Joeclinton1 (approved):
I was unable to reproduce the bug I identified from before where deselecting with shift grayed out the boolean operations, so either it's triggered by some specific previous set of steps, caused by Chrome acting up, or it has actually just been fixed by one of your changes.

The issue with Alt duplicating a compound shape which you pointed out is a scratch bug, is fixed by your patch.

The new tooltip and info in the description is useful and made using the addon much easier. Especially with the addition of the modifier keys.

Checking through the code, is a bit challenging to go through properly for everything because it is so long but general code style seems good. Well commented. The algorithms for path expanding, and changing the text to curves were fun to read. [pufferfish101007](https://github.com/pufferfish101007)'s comments cover the main stuff.

Since every feature works and the ui looks good, and I cannot produce a bug, I would say it looks good to me!

Joeclinton1:
I forgot to check compatibility with other addons. Can someone do that.  (particularly dark mode and compact editor)

Samq64:
For compact editor support I noticed the buttons collapse way before running out of space and the border is missing on two of the buttons and as well as the spacing where there is normally a separator:
image

For dark mode the shaping menu still needs some work:
image

Also it only opens if there is already an object which is a little odd and is inconsistent with other options in general when the window is narrow. Is it worth merging it into he existing overflow menu instead?

And one more tiny thing is the lack of a separator between back and unite.

addons/paint-boolean-ops/userscript.js
pufferfish101007: That's quite reasonable - I didn't fully appreciate that these were bugs present in vanilla scratch. I don't have a strong opinion either way!

pufferfish101007:
No more code comments from me, looks good!

This bug is still remaining:
> When combining scratch cat's head, the 'combine' button doesn't change to 'release' until the selected shapes are deselected and then reselected.

I also agree with @Samq64 that the Shaping button should be put into the 'more' overflow menu, when the viewport gets too narrow - on my laptop, enabling this addon causes the page to overflow horizontally (you can see this in the videos in my previous comments).

addons/paint-boolean-ops/userscript.js
griffpatch: Going to remove the config and just auto apply. No users will know to turn these off anyhow!

griffpatch:
> No more code comments from me, looks good!
> 
> This bug is still remaining:
> 
> > When combining scratch cat's head, the 'combine' button doesn't change to 'release' until the selected shapes are deselected and then reselected.

This is a weird one, I think I'll not worry to much about that edge case as there are lots of other weirdness's in the Scratch pain editor like that.

> I also agree with @Samq64 that the Shaping button should be put into the 'more' overflow menu, when the viewport gets too narrow - on my laptop, enabling this addon causes the page to overflow horizontally (you can see this in the videos in my previous comments).

That was really hard to achieve, but I think I have a good enough solution to it. See what you think?

griffpatch:
I have split the code into multiple files and tried to add a bit more plain English documentation to the methods.
I also added some hard won code to fight into the menu bar resizing and that MORE menu to make the toolbar feel more Scratch like and interactive.  It is compatible with the Compact Editor Addon.

pufferfish101007 (approved):
Thanks for doing the menu collapsing - it really makes a difference for small screens. It is not cleared up when the addon is dynamically disabled but that doesn't affect normal reusability at all, and will probably be a very rare occurrence in the wild.

Overall looks good to me; as usual, my approval doesn't count towards the required number, but I'll give an approving review anyway.

pufferfish101007:
I have just noticed that @Samq64's concerns re. compact editor compatibility from https://github.com/ScratchAddons/ScratchAddons/pull/8917#issuecomment-4165554071 haven't been addressed

Samq64:
Wihtout `editor-compact`: Shrinking then expanding the window causes the tools to disappear entirely until it becomes wide enough to display all of them without the overflow menu.

With `editor-compact`: While shrinking the tools appear both in the toolbar and in the overflow menu. Expanding works as expected.

griffpatch:
> Wihtout `editor-compact`: Shrinking then expanding the window causes the tools to disappear entirely until it becomes wide enough to display all of them without the overflow menu.
> 
> With `editor-compact`: While shrinking the tools appear both in the toolbar and in the overflow menu. Expanding works as expected.

I have separated off the toolbar integrating code from the main addon code which is a big code quality improvement. And have reworked that scripting to be more resilient. It should now collapse and expand robustly as well as react to enabling / disabling consistently.

Thanks for your bug reports!

Samq64:
I have not re-tested yet but one more minor thing from yesterday: with `editor-compact` could the overflow menu's padding be reduced to match the height of the tools?

griffpatch:
> I have not re-tested yet but one more minor thing from yesterday: with `editor-compact` could the overflow menu's padding be reduced to match the height of the tools?

This isn't an issue with this extension, but would require a change to the editor-compact addon I believe?
The clean fix belongs in editor-compact/userstyle.css: reduce vertical padding for [class*="fixed-tools_mod-menu-item_"]. That will compact Front/Back and our operations consistently.

addons/paint-boolean-ops/addon.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "Boolean shape operations",
+  "description": "Adds a Shaping panel to the vector costume editor: Unite · Subtract · Intersect · Combine/Release · Expand",
```
mxmou: The name Shaping doesn't seem to be used in the UI anymore, so it probably shouldn't be capitalized here. The following would be more consistent with other addon descriptions, which are written as full sentences:

```suggestion
  "description": "Adds shape operations to the vector costume editor: Unite, Subtract, Intersect, Combine/Release, and Expand.",
```

addons/paint-boolean-ops/addon.json
```diff
@@ -0,0 +1,50 @@
...
+  "info": [
+    {
+      "type": "info",
+      "text": "Unite — merge all selected shapes into one filled shape",
```
mxmou: Info messages are also written as sentences in other addons:
```suggestion
      "text": "\"Unite\" merges all selected shapes into one filled shape.",
```

addons/paint-boolean-ops/addon.json
```diff
@@ -0,0 +1,50 @@
...
+      "id": "infoIntersect"
+    },
+    {
+      "type": "info",
+      "text": "Combine / Release — Combine groups shapes into a compound path where overlaps become transparent holes; Release splits it back into separate shapes",
+      "id": "infoCombine"
+    },
+    {
+      "type": "info",
+      "text": "Expand — grows the shape outward by half its stroke width, making the fill match the visual edge of the stroke",
```
mxmou: This should say "outline", not "stroke", like Scratch's UI. It might also be a good idea to clarify that the edge of the fill after the operation will be where the edge of the stroke was before the operation.

addons/paint-boolean-ops/addon.json
```diff
@@ -0,0 +1,50 @@
...
+      "id": "infoExpand"
+    }
+  ],
```
mxmou, line 40: Could there be an additional notice explaining that the added buttons might be in the "More" menu on smaller screens?

mxmou (commented):
I have a few suggestions for the addon's presentation on the settings page.

griffpatch:
Thanks, I have made the requested wording changes in 392da5ba. The description and info text now use full sentences, the Expand text uses Scratch's outline wording and explains the fill edge, and there is a notice that the buttons may move into the More menu on smaller screens. I also brought the tooltips into line with the same wording. JSON and whitespace checks pass.

addons/paint-boolean-ops/addon.json
griffpatch: Updated the description and removed the unused "Shaping" localisation.

addons/paint-boolean-ops/addon.json
griffpatch: Updated all of the info messages so they read as full sentences.

addons/paint-boolean-ops/addon.json
griffpatch: Changed this to use the Scratch UI term "outline" and clarified where the fill edge ends up.

addons/paint-boolean-ops/addon.json
griffpatch: Added a notice explaining that the buttons may move into the "More" menu on smaller screens.

pufferfish101007:
Just noticed that the PR description mentions the "Open / Close Path" tool, which isn't mentioned in the addon description. Could some explanation of that be added too, preferably with a more scratch-friendly name?

addons/paint-boolean-ops/userscript.js
```diff
@@ -0,0 +1,747 @@
...
+    inlineSectionClass: "sa-shaping-section",
+    inlineItemSelector: ".sa-shaping-item",
+    inlineIconSelector: ".sa-shaping-item-icon",
+    inlineLabelSelector: ".sa-shaping-item-label",
+    overflowItems: allMoreItems,
+    onNativeClasses: ({ dashedBorderClass: nativeDashedBorderClass, disabledClass }) => {
+      dashedBorderClass = nativeDashedBorderClass;
+      if (dashedBorderClass) {
+        intersectBtn.classList.add(dashedBorderClass);
+        compoundBtn.classList.add(dashedBorderClass);
```
pufferfish101007: Is there a good reason why the intersect and compound buttons have this class? It makes the buttons look less consistent (see Samq64's previous comment)

griffpatch: They are different to the other Boolean operations, not actually being boolean operations but still classed as shaping tools. But perhaps I should just separate off the last two rather than having them as individuals... I've committed a fix for that

griffpatch:
I made a small change to have the correct bounding box drawn after a boolean operation

addons/paint-boolean-ops/userscript.js
pufferfish101007: Sorry, I hadn't realised that that was the class that controlled the separation. The problem isn't with the usual separation but rather when compact-editor is enabled, that class removes border, padding and margin so we end up with this
image

I think the correct usage of the class - based on inspecting the existing buttons - would be to wrap each input group in its own div.input-group_inout-group_* (i.e. (unit subtract intersect) (combine) (expand)), and apply the `fixed-tools_mod-dashed-border_*` class to the div rather than the span. That way, the dashed separators should still appear when compact-editor is disabled, and when it is enabled, the spacing should be correct (which it currently isn't because the spacing separator only applies to the input-group class), and the borders of the buttons should not be overridden.

griffpatch:
> Just noticed that the PR description mentions the "Open / Close Path" tool, which isn't mentioned in the addon description. Could some explanation of that be added too, preferably with a more scratch-friendly name?

Quite right - I've changed the label and added tooltips

addons/paint-boolean-ops/userscript.js
griffpatch: > Sorry, I hadn't realised that that was the class that controlled the separation. The problem isn't with the usual separation but rather when compact-editor is enabled, that class removes border, padding and margin so we end up with this image
> 
> I think the correct usage of the class - based on inspecting the existing buttons - would be to wrap each input group in its own div.input-group_inout-group_* (i.e. (unit subtract intersect) (combine) (expand)), and apply the `fixed-tools_mod-dashed-border_*` class to the div rather than the span. That way, the dashed separators should still appear when compact-editor is disabled, and when it is enabled, the spacing should be correct (which it currently isn't because the spacing separator only applies to the input-group class), and the borders of the buttons should not be overridden.

Ahh! I get what you mean now sorry, ok cool, I think I can fix that. I've added a small patch to the toolbar class and correctly wrapped the buttons as they should have originally been handled. I was just lazy - thanks for spotting it!

pufferfish101007:
So nearly there! Unfortunately it looks like [d2f7547](https://github.com/ScratchAddons/ScratchAddons/pull/8917/commits/d2f75475aaccfedcf7691595868b37086c2b7322) sometimes overcompensates (by quite a lot!)


https://github.com/user-attachments/assets/2fd2382c-2a84-41bb-8588-3052d8de00d5

(I checked and that commit is the culprit)

If I fully ungroup all of the subgroups by pressing the ungroup button again, the bug doesn't occur.

griffpatch:
> So nearly there! Unfortunately it looks like [d2f7547](https://github.com/ScratchAddons/ScratchAddons/pull/8917/commits/d2f75475aaccfedcf7691595868b37086c2b7322) sometimes overcompensates (by quite a lot!)
> 
>  2026-08-03.19-26-10.mp4 
> (I checked and that commit is the culprit)
> 
> If I fully ungroup all of the subgroups by pressing the ungroup button again, the bug doesn't occur.

Oh wow - you are great at finding gotchas! - This wasn't a bug in the latest commit, but it did reveal a bug that I hadn't tracked down previously but had seen... that when you combine shapes, sometimes it just didn't look right... It was combining shapes with nested groupings, the nested transformations were not being undone so some of the inner shapes got transported far away! hence the weird selection you saw! I've patched this and WOW does the combine look so much better now! Thanks for spotting that!

pufferfish101007:
Hopefully final things (all of which I'm happy to push):
- versionAdded needs updated to 1.46.0
- icons/exclude.svg is unused
- some other addons that use the alt key specify that this is option on mac (see the first few results of https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons+%22alt%22+language%3AJSON+path%3Aaddons%2F&type=code) - I think we should do the same here, in both the addon description and the tooltips, probably inline (`...Hold Alt (Option on macOS)...`)

griffpatch:
> Hopefully final things (all of which I'm happy to push):
> 
> * versionAdded needs updated to 1.46.0
> * icons/exclude.svg is unused
> * some other addons that use the alt key specify that this is option on mac (see the first few results of https://github.com/search?q=repo%3AScratchAddons%2FScratchAddons+%22alt%22+language%3AJSON+path%3Aaddons%2F&type=code) - I think we should do the same here, in both the addon description and the tooltips, probably inline (`...Hold Alt (Option on macOS)...`)

Done

PR #8918, opened 2026-03-27
Add paint-gradient-editor addon: gradient stop editor for vector cost…

### LINKED ISSUE #4953: An addon for changing the angle/direction of gradients
### What this addon does

Sometimes I want to make a gradient coming from a specific angle, not just horizontal and vertical ones. I know there's a little hack where you can get angles at 45° intervals by holding shift and rotating the shape you're coloring, but what if you want to get more precise (like 58°) without having to rotate the shape perfectly? That's where my suggestion comes in.

### How it works

After applying/selecting a horizontal or vertical gradient, you'd have a new number input to change its angle (where the gradient is coming from).
Color sliders popup with a slider to change the angle

I don't know of any

Here's what I mean by angle:
<picture>
  <source media="(prefers-color-scheme: light)" srcset="https://user-images.githubusercontent.com/106490990/182940446-f44226b0-9215-47d3-adf8-84fd2dfb9112.png">
  <source media="(prefers-color-scheme: dark)" srcset="https://user-images.githubusercontent.com/106490990/182940496-aaeb5544-a12c-4e83-a03c-4913644f8aa5.png">
  0 degrees is a vertical gradient, 45 degrees is a diagonal gradient, and 90 degrees is a horizontal gradient.
</picture>

Also, direction might be a better word.

griffpatch:
### Resolves
Will close #4953

### Changes

Adds a new **Gradient stop editor** addon to the vector costume editor.

When a linear or radial gradient fill or stroke is active, a gradient bar is injected directly onto the canvas as a live SVG overlay. The overlay is positioned and scaled to match the current paper.js view (including zoom and pan) and updates via a `requestAnimationFrame` loop.

image

image

## How to use

Select a shape with a gradient fill or stroke and open its colour picker. A line with draggable handles appears on the canvas showing the gradient axis.

- **Move the gradient** - drag the start or end handle (filled circles) to reposition the axis
- **Change a colour** - click any handle to open a colour picker for that stop
- **Add a stop** - click anywhere on the axis line
- **Move a stop** - drag a handle along the axis
- **Delete a stop** - drag a handle sideways off the axis; release to confirm
- **Radial gradients** - drag the ring handle at the centre to move the whole gradient

### Fill (paint bucket) tool

- **Copies the exact last-selected fill** - switching to the Fill tool now always matches whatever shape you last selected (solid colour or gradient, including its radial/linear axis and any extra stops), rather than Scratch's own behaviour of reusing whatever gradient type/axis was last used in the Fill tool specifically.
- **Shift-key eyedropper** - hold **Shift** while the Fill tool is active to turn it into a colour picker: the cursor changes, and hovering any shape's fill or outline temporarily previews that colour (gradient stops included) in the fill palette. Click to pick it, or release Shift without clicking to cancel and restore the previous fill.

### Reason for changes

The built-in Scratch gradient tool only lets you set the two end colours. There is no way to reposition the gradient stops, add intermediate stops, adjust the midpoint bias, or shift the gradient axis without leaving the colour picker. This addon exposes those controls directly on the canvas, making gradient editing practical for complex artwork.

The Fill tool additions round out the workflow: they let you carry a complex, hand-tuned gradient (or a plain colour) from one shape to another without re-building it stop-by-stop, matching the eyedropper-style workflow other paint tools offer.

### Tests

Tested in Chrome:

- Linear gradient: drag start and end handles to reposition axis
- Linear gradient: click axis to add an intermediate stop; drag to reposition; drag off-axis to delete
- Radial gradient: drag midpoint bias handle
- Colour picker: hue/saturation/alpha sliders, hex input, eyedropper picks colour from a canvas shape
- Switch between fill gradient on one item and stroke gradient on another
- Dynamic disable via addon toggle - overlay removed immediately
- Dynamic re-enable - overlay reconstructed on next gradient popup open
- Fill tool: selecting a solid-coloured shape then switching to Fill tool applies that exact colour
- Fill tool: selecting a multi-stop gradient shape then switching to Fill tool applies the full gradient (axis, type, and all extra stops) to the next shape filled
- Shift-eyedropper: hover a shape's fill and its outline separately - palette previews whichever is closer to the cursor
- Shift-eyedropper: click to commit a sampled colour, then use the Fill tool normally - the picked colour/gradient carries over correctly, including multi-stop gradients
- Shift-eyedropper: release Shift without clicking - fill palette and on-canvas preview both revert to the colour shown before Shift was pressed
- Shift-eyedropper: click once (e.g. an outline), keep Shift held and hover elsewhere (e.g. the shape body) to preview further colours, then release - palette correctly reverts to the last *clicked* colour, not the last hovered one


Joeclinton1 (commented):
This is a really useful addon, and I've wanted something like this for a while. Particularly the ability to fade out shapes at the edges.

The main functionality works well but there are some bugs though. I've commented directly at the lines that cause these bugs.

Additionally I think the addon description needs to describe how to use the addon in more detail.

Also the userscript file is very long and it might benefit from being split up into smaller files if possible.

addons/paint-gradient-editor/userscript.js
```diff
@@ -0,0 +1,1521 @@
...
+    };
+
+    // Generic inline HSV + alpha colour picker.
+    // openColorPicker(initialCss, onCommit, clientX, clientY, handleGroup?, ringRadius?)
+    //   initialCss    — starting colour as any CSS string
+    //   onCommit(css) — called live on every change with the new CSS colour string
+    //   handleGroup   — SVG <g> handle group to highlight while picker is open (optional)
+    //   ringRadius    — ring radius in SVG units (default 11 for p0/p1 handles)
+    // openExtraColorPicker(idx, …) is a thin wrapper around this for extra stops.
+    const openColorPicker = (initialCss, onCommit, clientX, clientY, handleGroup = null, ringRadius = 11) => {
```
Joeclinton1: I am unsure about the choice to implement a new colour picker. We already have third party colour picker vendor at libraries/thirdparty/color-picker.js, which is used for colour picking in the settings so to be more DRY I would normally say find a way to use that instead of implementing your own.

Secondly there is an addon 2D-colour picker which if it was extended a little bit, would provide the necessary colour picker directly in the existing colour selection window. So these overlap a bit. Most likely though it does need to be a pop out window, and so the third party colour picker provides a better option.

Although the eyedropper tool is not implemented in the third party one, so that would have to be added. so would some of the styling choices.

if you do keep it, I would say one thing is it shouldn't be dark mode by default. Only dark mode when the editor dark mode addon is enabled otherwise it needs to match scratch's existing styling.

addons/paint-gradient-editor/userscript.js
```diff
@@ -0,0 +1,1521 @@
...
+      const bg = Math.round(lc[1] + (rc[1] - lc[1]) * lerpT);
+      const bb = Math.round(lc[2] + (rc[2] - lc[2]) * lerpT);
+      const ba = lc[3] + (rc[3] - lc[3]) * lerpT;
+      const hex2 = (v) =>
+        Math.round(clamp(v, 0, 255))
+          .toString(16)
+          .padStart(2, "0");
+      const blended =
+        ba >= 0.999 ? `#${hex2(br)}${hex2(bg)}${hex2(bb)}` : `rgba(${br},${bg},${bb},${Math.round(ba * 1000) / 1000})`;
+      const newStop = { color: blended, offset: clamp(t, stops.p0 + 0.01, stops.p1 - 0.01) };
```
Joeclinton1: The add-stop path stores the new stop at the raw projected axis fraction, but the overlay renders stop handles using crampedFrac(...). That means creation and display are using different mappings, so a newly added stop can appear at a different location from the click point.

addons/paint-gradient-editor/userscript.js
```diff
@@ -0,0 +1,1521 @@
...
+    // addon's prevEventHandler re-reads and repositions its handle.
+    const syncPickerColorToRedux = (colorIndex) => {
+      const items = selectedLayers();
+      if (!items.length) return;
+      const cp = colorProp();
+      // Temporarily collapse every selected item to its 2 outer stops so that
+      // getColorsFromSelection reads the correct rgba from paper.js.
+      for (const item of items) {
+        const grd = item[cp]?.gradient;
+        if (!grd?.stops?.length >= 2) continue;
```
Joeclinton1: This guard is broken because ! binds before >=, so it does not actually skip anything. In mixed
selection, grd can be undefined, the loop falls through, and grd.stops throws. I can reproduce this as
Uncaught TypeError: Cannot read properties of undefined (reading 'stops') from syncPickerColorToRedux.

The behaviour is that if you select one shape with gradient and one shape without and you try to change the colour of a stop with a gradient, when you click off and go back you'll see the stop has been deleted aswell as an error in the console.

addons/paint-gradient-editor/userscript.js
```diff
@@ -0,0 +1,1521 @@
...
+      sync: syncOverlay,
+      destroy: () => {
+        active = false;
+        svg.remove();
+      },
+    };
+  };
+
+  // ── Main loop ──────────────────────────────────────────────────────────────────
+  addon.self.addEventListener("disabled", () => {
```
Joeclinton1, line 746: This destroys the SVG overlay, but it does not also close the custom floating color picker. Disabling the addon mid-interaction leaves this stale UI behind.

addons/paint-gradient-editor/userscript.js
```diff
@@ -0,0 +1,1521 @@
...
+      activeOverlay = buildOverlay(paper, canvasContainer, overlayCanvas);
+    }
+
+    // scratch-paint doesn't always dispatch CLOSE_MODAL when the picker closes
+    // (e.g. clicking outside the popover), but Redux modal state does update.
+    // Poll via rAF to detect the close within one frame, without DOM observation.
+    const pollPickerClose = () => {
+      const spModals = addon.tab.redux.state?.scratchPaint?.modals;
+      const stillOpen = activeColorMode === "stroke" ? spModals?.strokeColor : spModals?.fillColor;
+      if (!stillOpen) {
```
Joeclinton1, line 832: This only hides/syncs the overlay when the picker closes; it does not destroy it. As written, the overlay’s requestAnimationFrame loop can keep running after close.

Joeclinton1 (changes requested):
This is a really useful addon, and I've wanted something like this for a while. Particularly the ability to fade out shapes at the edges.

The main functionality works well but there are some bugs though. I've commented directly at the lines that cause these bugs.

Additionally I think the addon description needs to describe how to use the addon in more detail. (for example the way you delete is definitely not something you can know without instructions)

Consider if drag off line to delete is actually better than double click to delete (the way scratch does it on their node editor)

Also the userscript file is very long and it might benefit from being split up into smaller files if possible.

addons/paint-gradient-editor/userscript.js
Samq64: > We already have third party colour picker vendor at libraries/thirdparty/color-picker.js

Please don't. That one needs to be replaced anyway because it is for Vue 1 and the source code is no longer available.

WorldLanguages:
Before implementing a color picker, consider whether `<input type="color">` is good enough.

griffpatch:
No it wasn't - I needed transparency, and the stage colour picker too... I
did try it with the standard one first, but didn't like it.

addons/paint-gradient-editor/userscript.js
griffpatch: I have now separated the color picker from the main userscript.js, and started to generify it a bit more. I've also changed it to default to a light theme (came out okay, originally I was worried to blended too much into the white of the editor which is why I went dark theme), and of course, enabled a dark theme too when that is applied. See what you think.

addons/paint-gradient-editor/userscript.js
griffpatch: Dang... you are right. I've patched it.

addons/paint-gradient-editor/userscript.js
griffpatch: Hmm, I wasn't able to to reproduce this, unless another fix has guarded it, or I'm doing it wrong, however I can see the potential, so have patched it (I hope - If you could retest I'd appreciate it thanks)

addons/paint-gradient-editor/userscript.js
griffpatch: Fixed

addons/paint-gradient-editor/userscript.js
griffpatch: Fixed

Joeclinton1 (changes requested):
So, the new behaviour is that if I select a shape with a gradient and no gradient the stop points go away which I guess fixes the guard bug. Either way I can't test that anymore. The other bugs seem fixed. And the new style matches the light mode of scratch better and reacts to dark mode.

**Some bugs**:

1. The stop points are added at the right place if you click between the two existing nodes, however if you click outside those two it will place a stop point inside the two existing stop points rather than outside. The desired behaviour would be either to alert the user that you can't place a node outside the two starting stop points, or to just let the user do that instead of relocating it back to inside.

https://github.com/user-attachments/assets/e132aa50-3542-4973-9f6d-754eb49a29a0

2. There is a bug where the fill preview for the shape begins with the generic multi shape selected icon, and then when you click the fill it shows the right preview. It should show that preview from the start. Image
3. Sometimes drawing a new shape after interacting with a gradient will result in that shape having an empty fill

https://github.com/user-attachments/assets/02cabde9-44b3-4e2e-afa1-8bd428f96888

**UX**:

1. Would it be possible to double click a shape to enter gradient stop edit mode?
2. I would still like the addon description updated to tell the user better how to use the addon.
3. Consider a tool tip that appears after some delay of hovering a stop point to tell the user how to delete it.

griffpatch:
> **Some bugs**:
> 
> 1. The stop points are added at the right place if you click between the two existing nodes, however if you click outside those two it will place a stop point inside the two existing stop points rather than outside. The desired behaviour would be either to alert the user that you can't place a node outside the two starting stop points, or to just let the user do that instead of relocating it back to inside.

Okay, I fixed this now - clicking 'before' the first stop in a radial fill creates a new duplicate stop (which is hopefully less confusing - although you then can't delete that one as it's the first stop... They'll just have to lean that for now.

> 2. There is a bug where the fill preview for the shape begins with the generic multi shape selected icon, and then when you click the fill it shows the right preview. It should show that preview from the start.

Indeed. I think I won't try to solve that one at present. The fill is a mixed style fill so seeing it as that on the fill icon is acceptable I think?

> 3. Sometimes drawing a new shape after interacting with a gradient will result in that shape having an empty fill
Ouch... It's probably not going to be easy to fix all of Scratch paints bugs on this. They are existing issues with any imported SVGs with similar fills. I think we'll have to leave that one for now.

> 1. Would it be possible to double click a shape to enter gradient stop edit mode?
I would leave that idea for now as we need to be able to enter for the fill and the outline... double clicking may become used for something else at some point.

> 2. I would still like the addon description updated to tell the user better how to use the addon.
Agreed - Updated

> 3. Consider a tool tip that appears after some delay of hovering a stop point to tell the user how to delete it.
You can now double click to delete and drag off, I think a user will find it okay now.

Joeclinton1 (dismissed):
- I like the solution to clicking outside the starting node. Clever. 
- The new addon description is good.
- The reasoning for not doing the double click to edit is fine.
- The reasoning for not doing a tooltip is fine. And double clicking to remove is nice.

I do not think I can approve it with the glaring bugs of a new shape having an empty fill and the wrong preview. @Samq64 since you've been requested a reviewer on this what is your opinion? I'm going to give it a try myself to see how hard to achieve it is. If it's easy obviously I'll want it included. If I can't do it with reasonable effort then perhaps we can overlook this flaw.

Joeclinton1:
I made the changes that fix the two bugs I mentioned. There is still a bug in that the preview is a simplified two stop version unsure whether to fix that too, as requires dom manipulation, due to a lack of support in the redux. 
... I have also added myself to the credits ;)

Joeclinton1:
I didn't realize you can have both fill and outline gradients that's cool. However it led to a bug that I had to fix. I found I can actually get a full (not two stop simplified) preview to work by directly editing the preview in the dom overriding the redux, but it was a bit buggy (and the code was getting unreadably complex), so i'm stopping here.

Edit: I decided not to stop there and push a change that fixes the preview to use the complex gradient. The new shapes still use the redux simplified gradients though, and nothing I can do about that without patching redux.

Also I can't really approve this anymore because I've contributed code.

griffpatch:
Thanks @Joeclinton1, I've modified things a bit more to keep the orientation of the fill in the fill color swatch too. And I've gone ahead and implemented using the FILL tool to be able to use the last used fill (or stroke gradient) to fill subsequent shapes.

Joeclinton1 (changes requested):
The addon is almost done.

- The user script is now very long so I wonder if it would be possible to split up the userscript into separate files so that the userscript becomes more like a manager to make it easier to maintain. (I know this is mostly my fault for pushing for perfection on features that require basically hacking Scratch to achieve, and I'm worried that this was the wrong call, but the end result we currently have here is quite nice if you ignore the tremendous complexity required to achieve it)

- The linear gradients have a bug where the gradient is not correct given the position of the nodes, and it's even possible to have the extra stop swap places with the first stop by moving less than 25% of the range.

Image 

Image

griffpatch:
> The addon is almost done.
> 
> * The user script is now very long so I wonder if it would be possible to split up the userscript into separate files so that the userscript becomes more like a manager to make it easier to maintain. (I know this is mostly my fault for pushing for perfection on features that require basically hacking Scratch to achieve, and I'm worried that this was the wrong call, but the end result we currently have here is quite nice if you ignore the tremendous complexity required to achieve it)

Yes all your fault haha - Yeah, it has ended up getting longer and longer. If I refactor it then it will probably mean it will need approving all over again... but I guess that's ok... better to do it right! I'll have a think and see if I can do anything about it.

> * The linear gradients have a bug where the gradient is not correct given the position of the nodes, and it's even possible to have the extra stop swap places with the first stop by moving less than 25% of the range.

Are you able to give reproduction steps for this bug, I am having difficulty replicating!

addons/paint-gradient-editor/userscript.js
```diff
@@ -0,0 +1,1404 @@
...
+      .join("");
+
+  // Parse "#rrggbb", "rgb(r,g,b)" or "rgba(r,g,b,a)" → [r,g,b,a] (a is 0–1), or null.
+  const parseColor = (c) => {
+    if (!c || typeof c !== "string") return null;
+    if (c.startsWith("#") && c.length >= 7)
+      return [...[0, 1, 2].map((i) => parseInt(c.slice(1 + i * 2, 3 + i * 2), 16)), 1];
+    const m = c.match(/rgba?\(\s*([\d.]+)\s*,\s*([\d.]+)\s*,\s*([\d.]+)(?:\s*,\s*([\d.]+))?\s*\)/);
+    return m ? [+m[1], +m[2], +m[3], m[4] !== undefined ? +m[4] : 1] : null;
+  };
```
Samq64: This function duplicates the one in `stop-color-picker.js` but is it needed at all? Other addons already use [TinyColor](https://github.com/bgrins/TinyColor).

Joeclinton1:
> Yes all your fault haha - Yeah, it has ended up getting longer and longer. If I refactor it then it will probably mean it will need approving all over again... but I guess that's ok... better to do it right! I'll have a think and see if I can do anything about it.
> 

I'll make a commit now which refactors into the files. You can look at it and decide if you just want to roll back or not. But these ai coding tools make a refactor so easy that it's easier to just do it than explain the full structure.

> Are you able to give reproduction steps for this bug, I am having difficulty replicating!

Ok so I discovered it happens only if you first change to radial, move the start stop and then go back to linear.
(EDIT: I have fixed this bug in the latest commit)

https://github.com/user-attachments/assets/98384ab6-51c5-4dca-a61c-01fddfee3b9c

addons/paint-gradient-editor/userscript.js
Joeclinton1: I pushed a commit that addresses this

griffpatch:
@Joeclinton1 - Just to warn you - I was already underway refactoring the addon into modules... I'll review what you did once I've done to see if it's worth continuing, but I haven't got time tonight.

Joeclinton1:
> @Joeclinton1 - Just to warn you - I was already underway refactoring the addon into modules... I'll review what you did once I've done to see if it's worth continuing, but I haven't got time tonight.

Oh, i'm really sorry! I wanted to just get this done, so that it can be approved and we can move on. Didn't want to bother you with more work than is needed. After all it's only me really pushing for a big refactor.

Joeclinton1:
I've made 5 commits. This effectively finishes the addon. And I'd be prepared to approve it at this point. 

The changes include a large refactor, improving the commenting, splitting up the files into focused modules, remove defensive coding, using tinycolor. They also fix the bug I pointed out when you go from linear to radial and back again. It also makes it responsive to editor dark mode.

You can revert all of it if you want, because maybe you like the way you've refactored it better. But I think the easier thing would be to drop your on going module split up and build upon the changes i've made. If you do decide to revert all the commits to your last good commit, but want the bug fix commit and dark mode commit still, then ask an ai to find the commit difs and implement them back ontop of your current code.

Sorry again if I've wasted any of your time. I just really want to push this through, as I find it hard to stop thinking about things once I start something.

griffpatch:
> Sorry again if I've wasted any of your time. I just really want to push this through, 

No not at all, and I am super appreciative that you came on board on this log, and for your fixes.

> as I find it hard to stop thinking about things once I start something.

I know the feeling! - So, I did complete my refactor, and of course pull in your fixes... I'm torn now, I don't really know which branch to go with. If you want to see what I had done I've popped it in a new feature branch: https://github.com/griffpatch/ScratchAddons/tree/andy/paint-gradient-editor-refactor/addons/paint-gradient-editor

If you don't feel this was worth the effort, then I'll drop it and we can just go with yours. I was trying to create a more modular class and separate the Redux listeners out into their own module (state-handlers.js), rather than bundling coordinates, state, and color parsing into a single GradientModel. Basically - class for the overlay, small focused modules for everything else. But honestly either approach works, so whatever you prefer!

Joeclinton1:
@griffpatch I'll take a look at the feature branch tomorrow.

Joeclinton1:
So sorry for the delay. I got hugely busy. In the interest of moving this along I have chosen to replace my code with yours. Which is what the [Adopt griffpatch paint-gradient-editor refactor](https://github.com/ScratchAddons/ScratchAddons/pull/8918/commits/8209f83572656429c93bedf3ac4b49d103cbd900) commit does. Yours still leaves a "god file" userscript and continues to pass around state but how best to do this was subjective, and I trust your judgment so feel comfortable just using yours. Yours handles the GradientOverlay split better also.

On top of your code I added the changes to use the tinycolor dependency. Because we load that anyway we might aswell use it.

I've tested everything and am happy to approve it, if you're happy with it. 

(also if you're not happy with the choice to copy the code ontop of the changes leaving the history of the 5 now redundant commits, we can also do a force push clearing all the history as if I never did those commits, this will make the blame correctly attribute it to you)

griffpatch:
Yes, I'm all good with that - I agree that I'm still not 100% happy with
the modularization, but there reaches a point where I don't have time to
invest in working out that perfect split :) It's win some lose some.
If you are happy, then great. I'm happy enough! Thanks a lot for your help.

addons/paint-gradient-editor/stop-color-picker.js
```diff
@@ -0,0 +1,445 @@
...
+    track.style.cssText = `position:absolute;top:3px;left:${trackInset}px;right:${trackInset}px;height:12px;border-radius:6px;pointer-events:none;`;
+    const thumb = document.createElement("div");
+    thumb.className = "sa-stop-picker-thumb";
+    outer.append(track, thumb);
+    return { outer, track, thumb };
+  }
+
+  // Close any open panel immediately (e.g. when the addon is disabled).
+  close() {
+    document.querySelector(".sa-extra-stop-picker")?._close?.();
```
pufferfish101007: If the element exists, is `_close` ever going to be undefined?

addons/paint-gradient-editor/stop-color-picker.js
```diff
@@ -0,0 +1,445 @@
...
+  }
+
+  // ── Open / reuse the picker panel ────────────────────────────────────────
+  // color           — starting colour as any CSS string
+  // onCommit(color) — called live on every change with the new CSS colour string
+  open(color, onCommit, clientX, clientY) {
+    const existing = document.querySelector(".sa-extra-stop-picker");
+    if (existing?._setColor) {
```
pufferfish101007: As far as I can tell, `_setColor` will be defined if the element exists, assuming that no other code deletes that property (which I think is a reasonable assumption). So it seems to me that the condition can be changed to `!!existing`, and the `existing?.remove()` is redundant due to the `return` in the if block

Joeclinton1 (dismissed):
In interest of pushing this along, since it passes all the tests, and the structure looks good enough, I am approving this!

To the next reviewer: I was focused on the macroscopic effects and overall structure's maintainability in this review pass, but you may want to look at the details. I didn't see anything that jumped out at me, but honestly i'm a bit rusty now on this whole js extension stuff.

pufferfish101007 (commented):
This was quite difficult to test, as I ended up discovering a lot of bugs in the vanilla editor! Luckily, there's only two things that I think are more broken than they should be:

Clicking 'swap' does not swap the two colours that are shown on either side of the swap button, instead it sets one of the end stops to the same colour as the one next to it

https://github.com/user-attachments/assets/1d5beeba-0ce9-492c-b176-e9a54b646abf

<br>

It is possible to lie all stops right on top of each other such that the paint editor preview is different to what's on the stage. Is this intentional / a paper.js bug / a scratch-svg-renderer bug?  Also, once I'm in this state, I can't move the stops at all. Image

<br>

Other general comments:
- Is it intentional that the opacity slider is shown even when the opacity slider addon is disabled?
- I found the transparency of the color picker when using the eyedropper a bit confusing, because the colour preview is behind the transparent colour picker window, but the transparency is so great that I thought there was some strange visual artifact going on with the eyedropper preview. I personally would prefer for the colour picker to be slightly less transparent, or for the eyedropper preview circle to appear in front of the colour picker. Not a big deal at all though image

Seems to work well otherwise, modulo vanilla scratch bugs. My few comments on the code are very very minor. I'm a little confused by the complexity of the userscript initialisation - i.e. why it is `"runAtComplete": false` but then proceeds to wait for lots of things (redux, scratchClass) to load - but I trust that the complexity is necessary.

addons/paint-gradient-editor/color-utils.js
```diff
@@ -0,0 +1,25 @@
+// Pure colour-conversion helpers shared by the gradient editor modules.
+// No side-effects, no DOM access — safe to import from any context.
```
CST1229, line 2: em dash detected!! ai code?

griffpatch:
> This was quite difficult to test, as I ended up discovering a lot of bugs in the vanilla editor! Luckily, there's only two things that I think are more broken than they should be:
> 
> Clicking 'swap' does not swap the two colours that are shown on either side of the swap button, instead it sets one of the end stops to the same colour as the one next to it

Arrghh!!! - Regression, I must have broken that. It's now fixed in the next commit I make.

> It is possible to lie all stops right on top of each other such that the paint editor preview is different to what's on the stage. Is this intentional / a paper.js bug / a scratch-svg-renderer bug? Also, once I'm in this state, I can't move the stops at all. 

Yeah, this is intentionally done. I don't like it when you can't use the stops because they are overlapping. So I made the gradient stop handles have a min distance from each other where they actually are considered touching. This does mean the actual gradient is slightly out of sync with the stop handles, but I think it's acceptable for the functional benefit. You can see exactly what you're getting anyhow.

> * Is it intentional that the opacity slider is shown even when the opacity slider addon is disabled?

Yes.

> * I found the transparency of the color picker when using the eyedropper a bit confusing
I agree - it seemed a good idea at the time, but I've reverted that and it just works as you would otherwise except.

> I'm a little confused by the complexity of the userscript initialisation - i.e. why it is `"runAtComplete": false` but then proceeds to wait for lots of things (redux, scratchClass) to load - but I trust that the complexity is necessary.

So, "runAtComplete: false" is needed so our Redux listeners are registered before React mounts the paint editor - if we waited until page load we'd miss the initial CHANGE_SELECTED_ITEMS dispatch and start out of sync. But, running that early means Redux, paper.js, and Scratch's hashed CSS class names aren't available yet, so we have to defer the parts that need those until they show up. The async waits are just bridging that gap between "early enough to catch events" and "late enough for dependencies to exist".

I'm going to go through your other code notes before I push any of my fixes to this pull request

addons/paint-gradient-editor/stop-color-picker.js
griffpatch: Quite right... I'll fix

addons/paint-gradient-editor/stop-color-picker.js
griffpatch: Yeah, that was just defensive coding, but it shouldn't ever happen, so I'll remove it.

griffpatch:
Found a new issue while testing - the new gradient editor did not appear when in shaping mode. I've fixed this now so you can easily target individual shapes within a group to change their gradient fills without ungrouping first (using the shaping tool)

image

pufferfish101007 (approved):
Looks good to me. I don't think my approval counts toward anything but you can have it anyway.

addons/paint-gradient-editor/addon.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+      "name": "griffpatch",
+      "link": "https://scratch.mit.edu/users/griffpatch/"
+    },
+    {
+      "name": "Chrome_Cat",
+      "link": "https://scratch.mit.edu/users/Chrome_Cat/"
+    }
+  ],
+  "tags": ["editor", "costumeEditor"],
```
DNin01: Would it be a bad idea to add the Beta tag?
```suggestion
  "tags": ["editor", "costumeEditor", "beta"],
```

addons/paint-gradient-editor/addon.json
```diff
@@ -0,0 +1,50 @@
...
+  "versionAdded": "1.45.0",
```
DNin01:
```suggestion
  "versionAdded": "1.46.0",
```

DNin01 (commented):
It feels pretty robust until you start using the paint bucket.
- The selected color sometimes resets or goes out of sync.
- Using the paint bucket inconsistently sets the colors of objects just by hovering over them.
- Editing the colors of a selection of multiple different gradients is weird.

https://github.com/user-attachments/assets/a8ff4f5a-b1ce-4ec1-bd5d-5e06042863c4

DNin01:
Note that it is possible to attribute Claude Code, just remember to add this to the commit message when merging, filling in the model used (e.g., Sonnet 4.6):
```
Co-authored-by: Claude [model] <noreply@anthropic.com>
```
Just thinking, it's not a bad thing to disclose.

griffpatch:
Pushed a follow-up commit with fixes/additions to the Fill tool workflow:

- **Fix:** switching to the Fill tool now always copies the exact fill of whatever shape was last selected (solid colour or gradient, including its radial/linear axis and any extra stops) instead of Scratch's own sticky "last gradient type used in Fill tool" behaviour.
- **New: Shift-key eyedropper for the Fill tool** - hold Shift while the Fill tool is active to preview a colour picker: the cursor changes, and hovering any shape's fill or outline temporarily shows that colour (gradient stops included) in the fill palette. Click to pick it; release Shift without clicking to cancel and restore the previous fill.
  - Tried Alt for this first, but a lone Alt press/release turned out to be swallowed by Chrome/Windows as the menu-accelerator key with no way to detect it from page JS, so it's Shift instead.
- Deduplicated some shared gradient-angle math and canvas/DOM lookups between userscript.js and state-handlers.js into gradient-coords.js while working on this.
- Updated addon.json with a new info bullet describing the eyedropper.

Commit: e19a9571

Niki182:
The idea is awesome

PR #5729, opened 2023-02-15
New addon: Scroll while dragging blocks

CST1229:
Resolves [💬︎ drag blocks while scrolling in code area](https://discord.com/channels/806602307750985799/1024627431526645810), ["make the block following your mouse while scrolling"](https://discord.com/channels/806602307750985799/806609030935740477/920655752887169055) (7th most upvoted suggestion)

### Still to do

- [ ] Fix zoom

https://user-images.githubusercontent.com/68464103/219133644-e6713602-6690-4571-9ee7-46d08b7e82b6.mp4

- [ ] Fix browser scrolling maybe? (probably a vanilla bug but could belong in this addon)

https://user-images.githubusercontent.com/68464103/219133928-070d3e34-7baf-4cc0-97bf-55115b11d698.mp4

### Changes

Adds an addon that lets you scroll while dragging blocks.

https://user-images.githubusercontent.com/68464103/219133564-af62c62b-0acc-42fa-9d56-3b50a3dd2c8e.mp4

### Reason for changes

Fixes a _really_ annoying bug.

### Tests

_still a draft PR_

PR #8919, opened 2026-03-27
custom-zoom: add expanded workspace scrolling option

griffpatch:
### Changes

Adds a new **"Extended scrolling limit for code area"** setting (boolean, enabled by default) to the Customizable code area zoom addon.

When enabled, overrides Blockly's `workspace.getMetrics()` to expand the scrollable content boundary by one screen-width on each side horizontally and one screen-height on each side vertically. The override is applied once and restored cleanly if the setting is toggled off. Scrollbars are refreshed after each change.

Also:
- Refactors the `enabledLate` path to store the workspace in a variable rather than calling `getWorkspace()` twice
- Adds griffpatch to credits

### Reason for changes

By default, Scratch's code area prevents scrolling beyond the bounding box of your blocks. This makes it difficult to place new blocks in empty space far from existing ones, or to work near the edges of a large project. The extended scrolling setting adds a comfortable margin of empty space in all four directions so you can freely scroll and position blocks without hitting an invisible wall.

### Tests

Tested in Chrome. Verified:
- Extra scroll space is available in all four directions when the setting is enabled
- The setting can be toggled off at runtime and the original scroll boundaries are restored
- Existing zoom controls (min/max/speed/start/autohide) are unaffected

### DIFF
--- a/addons/custom-zoom/addon.json
+++ b/addons/custom-zoom/addon.json
@@ -6,6 +6,10 @@
     {
       "name": "ErrorGamer2000",
       "link": "https://scratch.mit.edu/users/ErrorGamer2000/"
+    },
+    {
+      "name": "griffpatch",
+      "link": "https://scratch.mit.edu/users/griffpatch/"
     }
   ],
   "versionAdded": "1.13.0",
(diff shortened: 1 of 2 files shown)


Samq64:
Related pull requests:
- #5729
- #7690

Samq64 (approved):

Samq64:
The old Blockly code is no longer necessary now that Scratch has shipped the update.

griffpatch:
> The old Blockly code is no longer necessary now that Scratch has shipped the update.

@Samq64 Are we stripping the old blockly code out now then?

Samq64:
No, not yet but it will be at some point and new code doesn't need that compatibly but because this is part of an existing addon it doesn't hurt to keep it for now.

PR #8924, opened 2026-03-30
Show Easter Eggs Category on April 1st

Samq64:
April 1st is coming very soon and this is an easy way to do something and doesn't require any new strings. Unlike previous years this is based on the browser's date instead of a timestamp. The position in the sidebar doesn't matter it just can't be the last because of `marginBottom: true`.

### DIFF
--- a/webpages/settings/components/category-selector.html
+++ b/webpages/settings/components/category-selector.html
@@ -5,7 +5,6 @@
         sel: category.id === selectedCategory && !this.$root.relatedAddonsOpen,
         hasParent: category.parent,
       }"
-    v-if="!category.hidden"
     v-show="shouldShow"
     transition="expand"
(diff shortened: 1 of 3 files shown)


Samq64:
@WorldLanguages Are you able to publish a release soon or should April Fools be skipped this year?

WorldLanguages:
We're skipping it this year, I guess it's time to acknowledge that April Fools is fun but shouldn't take our attention away from more important matters

Samq64:
This was meant as a way to still provide something to users while not wasting development time by keeping the check year round. Then again addons staying enabled after April fools and disappearing after being toggled off from the popup is not good UX.

WorldLanguages:
> by keeping the check year round

Ah, I see

Issue #8912, opened 2026-03-26
Scratch will move Set Thumbnail button to the editor


mxmou:
There are open PRs in scratch-www and scratch-editor that will move the Set Thumbnail button. The Thumbnail uploader addon and the stage header shared space need to be updated.

<img width="648" height="243" alt="Image" src="https://github.com/user-attachments/assets/c91d9d96-c28e-4b53-aa6f-11b28a12c0b4" />

<img width="320" height="233" alt="Image" src="https://github.com/user-attachments/assets/17400231-3871-4fb8-ad92-cf2431fcf3e4" />

IncrediboxDevToolsCrew:
This Set Thumbnail feature got famous.

The message of the thumbnail feature will be as this:

### Hi, there! 👋

The Set Thumbnail button now has a new spot! You can use it on editor or on the project page.

---
The another message has the "yes" and "no" buttons got switched.

mxmou:
@IncrediboxDevToolsCrew The message will be exactly as shown in the picture. If you don't have any new and useful information to add, then please don't make comment - use the reaction buttons.

IncrediboxDevToolsCrew:
Ok, sorry.

TSFBCE24RhythmHeaveners:
https://github.com/ScratchAddons/ScratchAddons/discussions/8922

IncrediboxDevToolsCrew:
We have 1 change dairy streaks, when we lost 1, we will lose us

WorldLanguages:
Can someone spell out the pull request number? (avoid linking directly to scratch-www pull requests because GitHub will add a reference to this issue)

Probably `scratchfoundation/scratch-www/pull/10113`

mxmou:
The scratch-editor PR is 471.

DNin01:
Is this complete?

mxmou:
There are a few small changes I intend to make after the Scratch update is released: adding styles for the confirmation popup for setting a thumbnail to editor dark mode and updating the design of the "Upload Custom Thumbnail" button to make it look better in the editor.

Samq64:
On the master branch with all addons enabled there are still come graphical glitches:

<img width="216" height="170" alt="Image" src="https://github.com/user-attachments/assets/b22051c3-16d1-411c-8ed5-d143b2e88439" />

DNin01:
Scratch released the thumbnail button update!

DNin01:
And the new Blockly?

[...]

PR #8925, opened 2026-04-01
animated-thumb: add dropdown in editor

mxmou:
### Changes

Updates the thumbnail uploader to work in the editor as well, which makes it compatible with the upcoming Scratch update (#8912). Also updates the shared space API.

image

### Tests

Tested on the current Scratch website and the manual thumbnail update branch.

### DIFF
--- a/addons/animated-thumb/userscript.css
+++ b/addons/animated-thumb/userscript.css
@@ -112,6 +112,9 @@ p.sa-animated-thumb-uploaded-thumb {
   width: 10px;
   height: 10px;
 }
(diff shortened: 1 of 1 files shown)


Samq64:
The dropdown item is only blue because it still appears on the project page and will be changed to white after it is editor-only, right?

mxmou:
> The dropdown item is only blue because it still appears on the project page and will be changed to white after it is editor-only, right?

Yes, it should probably be changed to white after the update is released. It could also be purple like Scratch's confirmation popup for setting the thumbnail:

Image

mxmou:
The Scratch PRs have been merged.

PR #8928, opened 2026-04-02
Settings page: fix `.main` height

Samq64:
Fixes #8927

#8896 caused the sidebar to increase the height of `.main` when overflowed instead of scrolling.

PR #8934, opened 2026-04-09
Add arrow buttons to table rows

mxmou:
### Changes

* Adds buttons for moving a table row up or down. They're next to the drag handle as suggested by @DNin01. The delete button is moved to the right side.
* The drag handle now has the grab cursor and can no longer receive keyboard focus.
* Reduces the amount of space between setting labels and inputs in a table row. Since all buttons are now inside the row's card, it looked weird when the labels were much closer to the buttons on the left than the inputs they're describing.
* Changes the "Add row" button from  image to image - I think it makes sense for it to be more prominent that the buttons that control individual rows.
* Moves button CSS from `addon-body.html` to `buttons.css` because it's used in both `<addon-body>` and `<addon-setting>`.

Screenshot of a table row on the settings page:

image

In the popup:

image

### Reason for changes

This makes it possible to reorder rows using the keyboard.

### Tests

Tested on Edge and Firefox.


DNin01 (approved):
I like it!

Samq64:
Could the dropdown list get a wider minimum width?

Image

Samq64:
Should the X be in the corner instead of centred vertically? Also is it worth adding a transition animation when using the buttons?

DNin01:
Also, when moving rows, the arrow buttons lose focus due to how the elements are repositioned (which was also an issue in my PR, #8689). It's better than nothing, but could it be done while retaining focus?

mxmou:
> Should the X be in the corner instead of centred vertically?

I prefer the center position because everything else is centered:

image

It would look like this in the top corner:

image

image

webpages/settings/components/addon-setting.js
```diff
@@ -90,6 +90,18 @@ export default async function ({ template }) {
...
         return this.setting.row.find((setting) => setting.id === id);
       },
+      moveTableRow(oldIndex, newIndex) {
+        let list = this.addonSettings[this.setting.id];
+        list.splice(newIndex, 0, list.splice(oldIndex, 1)[0]);
+        this.updateSettings();
+        this.sortable.captureAnimationState();
+        const focusedElement = document.activeElement;
+        setTimeout(() => {
+          this.sortable.animateAll();
```
mxmou, line 100: This function and `captureAnimationState()` are undocumented. I think it's fine to use them as long as we're careful when updating the library, but if it's an issue, I can try to implement the animation in a different way.

Samq64 (dismissed):

webpages/settings/components/addon-setting.html
```diff
@@ -293,29 +309,41 @@
...
     justify-content: center;
   }
+  .setting-table-options .handle {
+    cursor: grab;
+  }
   .setting-table-dropdown {
     align-self: flex-start;
   }
+  .setting-table-dropdown .dropdown-list {
+    min-width: 150px;
```
DNin01: Maybe not that wide, though.

DNin01 (dismissed):

webpages/settings/components/addon-setting.html
mxmou: If the menu isn't at least as wide as the button, the rounded corners look wrong:

image

I don't think there's a CSS-only solution to that, other than increasing the width, which I did, or changing the design.

DNin01: Could it be aligned to the right of the dropdown arrow instead of the left of the button row?

mxmou: Then there would be the same issue with the left corner. I was thinking about making all four corners of the menu rounded and adding a few pixels of space above it, like the dropdown menus on Windows 11.

DNin01: > Then there would be the same issue with the left corner.

That's true, but it doesn't have to be attached to both buttons; I think it would look fine like this:

Split button with dropdown rounded on one corner

...for the short ones.

mxmou: I don't like that because of this corner (it would look better if it wasn't rounded):

image

What do you think about making everything rounded and adding some space between the button and the menu? I'd add a shadow as well, so that it's clear the menu is above the other content:

image

DNin01: > I don't like that because of this corner (it would look better if it wasn't rounded)

Yeah, okay, maybe like this?

Split button with left aligned dropdown rounded on one corner

webpages/settings/components/addon-setting.html
DNin01, line 1: The menu should still be aligned to the side of the dropdown button, like the other menus.

Split button with floating right aligned dropdown menu

DNin01 (approved):
If the floating menu reduces the complexity of rounded corners, that's cool.

webpages/settings/components/addon-setting.html
DNin01: Never mind.

Samq64:
Should this include a fix for https://github.com/ScratchAddons/ScratchAddons/pull/8928#discussion_r3068340271 for should I open a separate PR for that?

mxmou:
I can add a fix for that, since it's related to dropdown menus.

DNin01:
@Samq64 Final review?

Samq64 (approved):

Samq64:
Shouldn't this have waited for a non-patch release?

DNin01:
> Shouldn't this have waited for a non-patch release?

I just wanted to be able to continue working on some updates that I may have had to resolve merge conflicts with. We can cherry-pick... all but one commit for v1.45.1 :P (or we could release it early as an accessibility "fix").

@WorldLanguages Maybe we should consider having a branch for upcoming features (like other development projects) so we don't have to do this again? Lol

DNin01:
At least it includes some nice internal updates, like simplifying whatever this was...
https://github.com/ScratchAddons/ScratchAddons/blob/ab5f2e8f0c9d293feb39ebc5f30b828a4a1af743/webpages/settings/components/addon-body.html#L238-L242
...to just this:
https://github.com/ScratchAddons/ScratchAddons/blob/e6de62310533ca32c5da4c6b6bbf364e6ade0b8e/webpages/styles/components/buttons.css#L45
👏

mxmou:
> (or we could release it early as an accessibility "fix")

It adds new strings - it should probably wait so they can be translated.

> Maybe we should consider having a branch for upcoming features (like other development projects) so we don't have to do this again? Lol

In the past, patch releases were developed on a separate branch which was then merged (if someone remembered to merge it - see #1612) into master. But I prefer working on a single branch because dealing with merge conflicts after branches diverge significantly can be difficult. The git history also becomes more complicated when merging things from different branches.

WorldLanguages:
> Shouldn't this have waited for a non-patch release?

I think we can be flexible, we're all volunteers here so no one can blame you for not wanting to deal with with merge conflicts, which are annoying

PR #8938, opened 2026-04-10
Merge sidebar buttons and add darkened background overlay

### LINKED ISSUE #8937: Better sidebar toggle placement
> Now that the name of the selected category is shown in the toolbar, it would make sense to put the button that opens the sidebar in small mode next to it. I can do that in a future PR.

_Originally posted by @mxmou in https://github.com/ScratchAddons/ScratchAddons/pull/6104#pullrequestreview-2742065295_

Replacing the `<` in wide mode could also be considered.

Samq64:
Resolves #8937

### Changes

Combined both sidebar buttons into a single one next the the category name and darkens the background when the sidebar is open in small mode.

### Reason for changes

It's more consistent and the darkening makes it much clearer how to close the sidebar when no button is visible.

### Tests

Tested on Helium.

Right after opening more settings in small mode there are two overlays until the sidebar transition finishes.

### DIFF
--- a/webpages/settings/index.html
+++ b/webpages/settings/index.html
@@ -228,7 +228,7 @@
       </div>
 
       <!-- This is the main menu, where the searchbar and the addon items are located -->
-      <div class="addons-block">
+      <div class="addons-block" :class="{darken: smallMode && categoryOpen}">
         <div class="addons-block-header">
           <div v-if="!isIframe" class="category-header-title" v-cloak>
             <button
--- a/webpages/settings/index.html
+++ b/webpages/settings/index.html
@@ -232,10 +232,9 @@
         <div class="addons-block-header">
           <div v-if="!isIframe" class="category-header-title" v-cloak>
             <button
-              id="sidebar-toggle"
               class="arrow-button"
               :title="msg('toggleSidebar')"
-              @click="sidebarToggle()"
+              @click="categoryOpen = !categoryOpen"
               v-cloak
             >
               <img src="../../images/icons/menu.svg" draggable="false" />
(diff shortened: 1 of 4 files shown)


webpages/styles/components/categories.css
```diff
@@ -8,7 +8,7 @@
   overflow-y: auto;
   scrollbar-width: thin;
   scrollbar-color: var(--gray-text) transparent;
-  padding: 20px 0;
+  padding-top: 20px;
```
DNin01, line 11: The top and bottom paddings should probably match. With padding on only one side, the sidebar looks uneven.

webpages/settings/index.html
```diff
@@ -232,28 +221,20 @@ <h1 v-cloak>{{ msg("settings") }}</h1>
...
-      </button>
 
       <!-- This is the main menu, where the searchbar and the addon items are located -->
-      <div class="addons-block">
+      <div class="addons-block" :class="{darken: smallMode && categoryOpen}">
         <div class="addons-block-header">
           <div v-if="!isIframe" class="category-header-title" v-cloak>
+            <button class="arrow-button" :title="msg('toggleSidebar')" @click="categoryOpen = !categoryOpen" v-cloak>
+              <img src="../../images/icons/menu.svg" draggable="false" />
+            </button>
```
DNin01, line 236: In small mode, this button is covered by the sidebar when it's open, so there needs to be another way to close it, such as by clicking on an already-selected category. Clicking outside works, but only on wide enough screens.

DNin01 (changes requested):
I was thinking about the addition of a persistent hamburger button, but on the settings page header, not the category strip. There are a few problems with the new position of the sidebar toggle button:
- Being part of the addon list header, the button is expected to control a menu related to the list, but the sidebar also contains external links and the More Settings button, which are not related to filtering addons.
- Because of this, it's not clear that the menu button closes the main sidebar.

PR #8939, opened 2026-04-11
Bump version to 1.46.0-prerelease

WorldLanguages:
Bumps version


Samq64:
There are only bug fixes on the master branch so 1.45 patch releases may still be based directly on it if needed.

PR #8940, opened 2026-04-11
Improve navigation bars

DNin01:
### Changes

Small changes have been made to the sidebar in the settings page and the tab bar in the popup.

- Design:
  - Clicking on a selected category in the sidebar in small mode now closes the sidebar, making navigation easier.[^1]
  - The sidebar now has a shadow in small mode and is positioned behind the navbar.
- Fixes:
  - A bug which made the sidebar open under the More Settings dialog when visiting `(...)#moresettings` has been fixed.
  - Animations are reduced if that is the user's motion preference.
- Code clarity:
  - Uses a single style rule to set the length of the orange indicators on selected categories/tabs, reducing the number of lines of code. Previously, left/right offsets were used in the popup and a separate length value was used for child categories in the settings page.

### Tests

Tested on Edge 146.

[^1]: Partially solves #9001

### DIFF
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -104,9 +104,8 @@ button {
   content: "";
   display: block;
   position: absolute;
-  left: 10px;
-  right: 10px;
   bottom: 0;
+  width: calc(100% - 20px);
   height: 4px;
   background-color: var(--orange);
   border-radius: 4px 4px 0 0;
--- a/webpages/popup/style.css
+++ b/webpages/popup/style.css
@@ -105,7 +105,7 @@ button {
   display: block;
   position: absolute;
   bottom: 0;
-  width: calc(100% - 20px);
+  inset-inline: 10px;
   height: 4px;
   background-color: var(--orange);
   border-radius: 4px 4px 0 0;
--- a/webpages/settings/components/category-selector.html
+++ b/webpages/settings/components/category-selector.html
@@ -75,7 +75,7 @@
     display: block;
     width: 4px;
     border-radius: 0 4px 4px 0;
-    height: 40px;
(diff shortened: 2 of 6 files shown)


webpages/settings/components/category-selector.html
```diff
@@ -75,7 +75,7 @@
     display: block;
     width: 4px;
     border-radius: 0 4px 4px 0;
-    height: 40px;
+    height: calc(100% - 8px);
```
mxmou: This matches the old height of top-level categories, but not child categories - those currently have more space above and below the indicator, which I think looks better:

```suggestion
    height: calc(100% - 12px);
```

webpages/settings/style.css
```diff
@@ -32,7 +32,7 @@ button {
   box-shadow: var(--large-shadow);
   user-select: none;
   position: relative;
-  z-index: 10;
+  z-index: 100;
```
mxmou, line 35: It would be nice to have all z-index values in one place so that it's easier to keep track of them - something like Scratch's [z-index.css](https://github.com/scratchfoundation/scratch-editor/blob/bbb26f8/packages/scratch-gui/src/css/z-index.css) file.

webpages/popup/style.css
```diff
@@ -104,9 +104,8 @@ button {
   content: "";
   display: block;
   position: absolute;
-  left: 10px;
-  right: 10px;
   bottom: 0;
+  width: calc(100% - 20px);
```
mxmou: I'm not sure if this is an improvement - when reading or editing code like this, I find it more useful to know the spacing on each side than the total width.

mxmou (approved):

Samq64 (approved):

PR #8945, opened 2026-04-12
feat(forum-copy-code): support copying scratchblocks code

### LINKED ISSUE #8944: Support copying scratchblocks code with `forum-copy-code` addon
### Why this would be helpful

Easier to copy scratchblocks code from forum posts.

### My suggestion

Make forum-copy-code addon also show the copy button on scratchblocks and copy the raw code.

### Possible alternatives

_No response_

### Additional context

_No response_

LuYifei2011:
Resolves #8944

### Changes

image

### Reason for changes

Easier to copy scratchblocks code from forum posts.

### Tests

On Chrome 147.

### DIFF
--- a/addons-l10n/en/forum-copy-code.json
+++ b/addons-l10n/en/forum-copy-code.json
@@ -1,4 +1,4 @@
 {
   "forum-copy-code/copy-code": "Copy Code",
-  "forum-copy-code/copy-scratchblocks": "Copy Scratchblocks Code"
+  "forum-copy-code/copy-scratchblocks": "Copy Scratchblocks"
 }
--- a/addons-l10n/en/forum-copy-code.json
+++ b/addons-l10n/en/forum-copy-code.json
@@ -1,3 +1,4 @@
 {
-  "forum-copy-code/copy-code": "Copy Code"
+  "forum-copy-code/copy-code": "Copy Code",
+  "forum-copy-code/copy-scratchblocks": "Copy Scratchblocks Code"
 }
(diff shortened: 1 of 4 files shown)


Samq64:
Could the addon description be updated and a second label be used for blocks? Also it's worth considering if the two buttons should be individually toggleable but still be part of the same addon.

CST1229:
> Could the addon description be updated and a second label be used for blocks? Also it's worth considering if the two buttons should be individually toggleable but still be part of the same addon.

imo i dont think they need to be individually togglable

addons-l10n/en/forum-copy-code.json
```diff
@@ -1,3 +1,4 @@
 {
-  "forum-copy-code/copy-code": "Copy Code"
+  "forum-copy-code/copy-code": "Copy Code",
+  "forum-copy-code/copy-scratchblocks": "Copy Scratchblocks Code"
```
Samq64: This label is pretty long, maybe remove "Code"?
```suggestion
  "forum-copy-code/copy-scratchblocks": "Copy Scratchblocks"
```

addons/forum-copy-code/userstyle.css
```diff
@@ -1,6 +1,8 @@
 .sa-copyCodeDiv {
   text-align: right;
   height: 0px;
+  position: relative;
+  z-index: 1;
```
Samq64, line 6: Could you explain why these are needed, maybe as a code comment?

Samq64 (approved):

PR #8946, opened 2026-04-12
stage-monitors: Increase CSS specificity

Samq64:
Fixes https://github.com/ScratchAddons/feedback-internal/issues/570

### Changes

Increases the specificity of some `stage-monitor` selectors.

### Reason for changes

#8850 decreased it which caused some styles to not apply.

### Tests

Tested on Helium

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -182,7 +182,6 @@
   "horizontal-mystuff-tabs",
   "// For editor:",
   "editor-dark-mode",
-  "stage-monitor",
   "editor-colored-context-menus",
   "hide-stage",
   "editor-stage-left",
@@ -190,5 +189,6 @@
   "fullscreen",
   "columns",
   "editor-compact",
-  "editor-theme3"
+  "editor-theme3",
+  "stage-monitor"
 ]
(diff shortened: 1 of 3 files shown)


DNin01 (approved):
The bug only happened with `stage-monitor` and `editor-theme3` enabled together. Shows the importance of testing with multiple users/configurations.

mxmou:
> The bug only happened with `stage-monitor` and `editor-theme3` enabled together.

Would moving stage-monitor after theme3 in `addons.json` also fix the issue? If so, I'd perfer that solution, but increasing the specificity is fine too.

DNin01:
> > The bug only happened with `stage-monitor` and `editor-theme3` enabled together.
> 
> Would moving stage-monitor after theme3 in `addons.json` also fix the issue? If so, I'd perfer that solution, but increasing the specificity is fine too.

Seems like it does.

Samq64:
> The bug only happened with stage-monitor and editor-theme3 enabled together. 

This also fixes the outer border radius which is unrelated to `editor-theme3`.

DNin01 (approved):

mxmou (approved):

PR #8953, opened 2026-04-16
Fix no-sprite-confirm and confirmation positions with editor-stage-left

### LINKED ISSUE #8952: Delete sprites without confirmation broke
```
The setting "Delete sprites without confirmation" doesn't let you delete a
sprite. I am using Firefox 149.0.2 on Windows 11, if that helps.
```
Scratch Addons version: 1.45.0

### LINKED ISSUE #8963: Deletion confirmation goes off-screen if using Display stage on left side addon.
### Description

If I try to delete a sprite, the deletion confirmation goes offscreen, which does not let me delete a sprite.

### Steps to reproduce

1. Turn on the Display stage on left side addon.
2. Create a new project.
3. Try to delete the Scratch Cat.

Image

### Environment

Browser: Chrome 132
Scratch Addons version: 1.45.0
Operating system: Windows 10

Samq64:
Resolves https://github.com/ScratchAddons/feedback-internal/issues/589
Resolves #8952
Resolves #8963

This does not fix the positioning of the thumbnail one.

----
Bugs fixed by this PR:
1. no-sprite-confirm does not delete the sprite without confirmation, instead the modal simply is hidden.
2. Display stage on left side addon, causes the delete sprite confirmation modal to be off screen.

### DIFF
--- a/addons/editor-stage-left/stageleft.css
+++ b/addons/editor-stage-left/stageleft.css
@@ -65,6 +65,7 @@
 }
 
 /* Move delete button confirmation to the right side */
+/* TODO: override the props of the ModalWithArrow component instead */
 .ReactModal__Content[style*="width: 290px"] {
   /* 290px (popup width)
    + 25px (padding)
@@ -72,11 +73,11 @@
    + 25px (padding) */
   margin-left: 410px !important;
 }
-/* The arrow is outside the modal's DOM and its class is shared with the set thumbnail callout */
-[class*="modal-with-arrow_arrow_"]:not([class*="feature-callout-popover_arrow_"]) {
+[class*="modal-with-arrow_arrow_"][style*="width: 14px"] {
   /* 70px (sprite selector item width)
    + 50px (padding)
-   - 10px (arrow width) */
+   + 4px (popup borders)
+   - 14px (arrow width) */
   margin-left: 110px;
   transform: scaleX(-1);
 }
--- a/addons/editor-stage-left/stageleft.css
+++ b/addons/editor-stage-left/stageleft.css
@@ -65,15 +65,19 @@
 }
 
 /* Move delete button confirmation to the right side */
-.ReactModal__Content[style*="width: 300px"] {
-  /* 300px (popup width)
+.ReactModal__Content[style*="width: 290px"] {
+  /* 290px (popup width)
    + 25px (padding)
    + 70px (sprite selector item width)
    + 25px (padding) */
-  margin-left: 420px !important;
+  margin-left: 410px !important;
 }
-[class*="delete-confirmation-prompt_arrow-container_"] {
-  order: -1;
+/* The arrow is outside the modal's DOM and its class is shared with the set thumbnail callout */
+[class*="modal-with-arrow_arrow_"]:not([class*="feature-callout-popover_arrow_"]) {
+  /* 70px (sprite selector item width)
+   + 50px (padding)
+   - 10px (arrow width) */
+  margin-left: 110px;
   transform: scaleX(-1);
 }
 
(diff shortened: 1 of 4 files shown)


addons/no-sprite-confirm/userscript.js
```diff
@@ -1,7 +1,7 @@
 export default async ({ addon, console }) => {
   document.body.addEventListener("pointerup", () => {
     setTimeout(() => {
-      const confirmButton = document.querySelector("[class*='delete-confirmation-prompt_ok-button_']");
+      const confirmButton = document.querySelector("[class*='confirmation-prompt_confirm-button_']");
```
mxmou: This also matches the "yes" button in the Set Thumbnail confirmation popup.

addons/editor-stage-left/stageleft.css
```diff
@@ -65,15 +65,19 @@
...
    + 25px (padding)
    + 70px (sprite selector item width)
    + 25px (padding) */
-  margin-left: 420px !important;
-}
-[class*="delete-confirmation-prompt_arrow-container_"] {
-  order: -1;
+  margin-left: 410px !important;
+}
+/* The arrow is outside the modal's DOM and its class is shared with the set thumbnail callout */
```
mxmou: ... and with the Set Thumbnail confirmation popup. Which is unfortunate because it looks like the only way to differentiate these two arrows with CSS selectors is using their inline styles, which could change in the future (like the `width: 300px` above did).

Maybe it would be better to remove these styles and try to override the props of the [ModalWithArrow](https://github.com/scratchfoundation/scratch-editor/blob/364af13/packages/scratch-gui/src/components/modal-with-arrow/modal-with-arrow.jsx) component instead.

mxmou (changes requested):

addons/editor-stage-left/stageleft.css
mxmou: Or we can release a temporary CSS-only solution in v1.45.1 and implement something more reliable later.

addons/no-sprite-confirm/userscript.js
```diff
@@ -1,7 +1,9 @@
 export default async ({ addon, console }) => {
   document.body.addEventListener("pointerup", () => {
     setTimeout(() => {
-      const confirmButton = document.querySelector("[class*='delete-confirmation-prompt_ok-button_']");
+      const confirmButton = document.querySelector(
+        "[class*='confirmation-prompt_modal-container_'] [class*='confirmation-prompt_confirm-button_']"
```
mxmou: This still matches both confirmation popups.
```suggestion
        "[class*='delete-confirmation-prompt_body_'] [class*='confirmation-prompt_confirm-button_']"
```

mxmou (changes requested):

mxmou (approved):
Ideally, editor dark mode would style both popups, but that can be added in a future PR.

addons/no-sprite-confirm/userscript.js
Samq64: I forgot about the thumbnail confirmation, I was focused on the callout tooltip. It would be nice for it to also support editor-stage-left but manual positioning is too tricky in the stage header.

mxmou: I don't think we have to worry about that tooltip too much - users will only see it once.

mxmou:
I've just remembered that I added the necessary dark mode styles a while ago but forgot to make a PR -  I'll do that now. My implementation doesn't use inline styles to select elements and also removes some selectors that are now unused, so I suggest reverting the dark mode changes in this PR.

mxmou (approved):

Joeclinton1:
Guys how was a hotfix pushed without this, this is like the 4th most talked about bug in feedback.

Joeclinton1 (approved):
LGTM

WorldLanguages:
> Guys how was a hotfix pushed without this, this is like the 4th most talked about bug in feedback.

My bad, I didn't notice it in the PR list.

PR #8956, opened 2026-04-16
Fix debugger block preview and heatmap for new Blockly

### LINKED ISSUE #8955: Debugger block preview and timing heatmap is broken
### Description
Profiling is not showing the correct block preview and locations
Image

Heatmap is entirely broken erroring with:

Image

### Steps to reproduce

1. Turn on profiling and heatmap from debugger timing addon
2. Run any code

_No response_

### Environment

Browser: Chrome
Scratch Addons version: v1.45
Operating system: Windows

Joeclinton1:
<!-- Which issue(s) does this pull request fix or resolve? If there aren't any, please submit one first unless this is a hotfix or minor string update. -->

Resolves #8955

### Changes                                                                                                                                                                                
                                                                                                                                                                                             
  Updated the debugger for the new Blockly renderer in two places:                                                                                                                           
                                                                                                                                                                                             
  - Fixed block previews in the profiler/timing table by deriving preview categories from Scratch block extensions such as colours_control, instead of relying on the removed                
    jsonData.category field.                                                                                                                                                                 
  - Fixed debugger heatmap rendering on new Blockly by replacing old internal block lookups with workspace.getBlockById(...), using current block renderer APIs, and restoring block/input   
    styling correctly when heatmap is disabled or cleared.

<img width="700" alt="image" src="https://github.com/user-attachments/assets/3c589df0-3299-40e9-8eac-5cc0f050d190" />
                                                                                                                            
                                                                                                                                                                                             
  ### Reason for changes                                                                                                                                                                     

  Scratch’s move away from scratch-blocks broke debugger assumptions in both preview rendering and heatmap coloring. The profiler preview path was still expecting old block JSON category   
  metadata, and the heatmap path was still reaching into old Blockly internals and renderer fields. These changes align the debugger with the current Scratch/Blockly behavior so previews   
  and heatmap work again.                                                                                                                                                                    
                                                                                                                                                                                             
  ### Tests                                                                                                                                                                                  
                                                                                                                                                                                             
  Tested manually in the editor with the debugger timing tab on https://scratch.mit.edu/projects/1030987979:                                                                                                                                
                                                                                                                                                                                             
  - Verified profiler block previews render again for line-by-line timing rows.                                                                                                              
  - Verified heatmap can be enabled without throwing errors.                                                                                                                                 
  - Verified heatmap can be disabled and cleared without leaving text input fields stuck in the wrong visual state.

### DIFF
--- a/addons/debugger/timing/HeatmapManager.js
+++ b/addons/debugger/timing/HeatmapManager.js
@@ -13,11 +13,13 @@ function recursiveFillBlock(block, fill = null) {
     else style.fill = fill;
   };
   setFill(block.pathObject.svgPath.style);
+  // Mark heatmapped blocks so CSS can keep Blockly's selected overlay unfilled.
+  block.getSvgRoot().classList.toggle("sa-debugger-heatmap", !isReset);
 
   // Set text color for blocks with heatmap applied
   const textElements = block
     .getSvgRoot()
-    .querySelectorAll(":scope > :not(.blocklyDraggable):not([data-shapes='stack']) > text, :scope > text");
+    .querySelectorAll(":scope > :not(.blocklyDraggable) > text, :scope > text");
 
   if (!isReset) {
     // Heatmap is being applied - force white text
--- a/addons/debugger/timing/HeatmapManager.js
+++ b/addons/debugger/timing/HeatmapManager.js
@@ -15,9 +15,9 @@ function recursiveFillBlock(block, fill = null) {
   setFill(block.pathObject.svgPath.style);
 
   // Set text color for blocks with heatmap applied
-  const textElements = block.getSvgRoot().querySelectorAll(
-    ":scope > :not(.blocklyDraggable):not([data-shapes='stack']) > text, :scope > text"
-  );
+  const textElements = block
+    .getSvgRoot()
+    .querySelectorAll(":scope > :not(.blocklyDraggable):not([data-shapes='stack']) > text, :scope > text");
 
   if (!isReset) {
     // Heatmap is being applied - force white text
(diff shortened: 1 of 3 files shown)


addons/debugger/timing/HeatmapManager.js
```diff
@@ -7,15 +7,19 @@ function valueToHeatmapColor(value, min = 0, max = 1) {
...
+  };
+  setFill(block.pathObject.svgPath.style);
 
   // Set text color for blocks with heatmap applied
-  const textElements = block.svgGroup_.querySelectorAll(
-    ":scope > :not(.blocklyDraggable):not([data-shapes='stack']) > text, :scope > text"
-  );
+  const textElements = block
+    .getSvgRoot()
+    .querySelectorAll(":scope > :not(.blocklyDraggable):not([data-shapes='stack']) > text, :scope > text");
```
mxmou: The `data-shapes` attribute no longer exists. That part of the selector can probably be removed.

Joeclinton1: Thanks will make this change. Please approve after I do, so this can be part of the hotfix.

Joeclinton1: Image

I was skeptical but indeed it doesn't seem to have the old bug of colouring in the procedure block label (which is why that selector existed in the first place)

Image

if you drag the block though you can get a failure though. Idk if worth fixing though, as it goes away as soon as you click something.

Joeclinton1: Ok I decided to put in the time to fix it, despite it being a minimal issue. (it's possible this isn't even due to new blockly and has always been a bug)

mxmou (approved):

DNin01 (approved):
Seems OK.

PR #8958, opened 2026-04-17
Remove animated-thumb styles from scratch3to2

mybearworld:
### Changes

Before:
image
After:
image

### Reason for changes

With #8925, I don't think this is necessary anymore

### Tests

Tested in Chrome

### DIFF
--- a/addons/scratch3to2/project.css
+++ b/addons/scratch3to2/project.css
@@ -631,17 +631,6 @@
   height: 46px; /* two lines */
 }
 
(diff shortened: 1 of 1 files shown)


mxmou (commented):
The styles for the modal (`.sa-animated-thumb-popup-content` and `.sa-animated-thumb-popup-actions`) can also be removed.

mxmou (approved):

Samq64 (approved):

PR #8978, opened 2026-04-19
Remove fix-pasted-scripts

mxmou:
Removes the fix-pasted-scripts addon - the bug it fixed no longer exists.


DNin01 (approved):
LGTM. (Fun fact, the bug does still happen to me if I start dragging a block, open the Windows Start menu, release the mouse button, close the Start menu, and then click the block into place, but who's going to do that by accident?)

It's possible the bug could still be triggered by `middle-click-popup` (when you click to place a block you picked—if that happens, we may want to change that addon so that blocks don't get dragged automatically, preventing it from happening).

DNin01:
> It's possible the bug could still be triggered by `middle-click-popup`

Actually, it seems like it doesn't. 👍

PR #8982, opened 2026-04-20
Grab single block with Ctrl key -> Grab single block by default

Samq64:
### Changes

`block-cherry-picking` now only flips the controls and the name and description have been updated to reflect that.

### Reason for changes

Cherry picking is now included in the default editor.

### Tests

Tested on Helium.

### DIFF
--- a/addons/block-cherry-picking/addon.json
+++ b/addons/block-cherry-picking/addon.json
@@ -3,6 +3,10 @@
   "description": "Inverts the block dragging controls. Drag a single block out of the middle of a script by default and hold the Ctrl key to drag the entire stack attached below it.",
   "info": [
     {
+      "text": "To drag a single block without this addon hold Ctrl.",
+      "id": "vanilla"
+    },
+    {
       "text": "On macOS, use the Cmd key instead of the Ctrl key.",
       "id": "macContextDisabled"
     }
--- a/addons/block-cherry-picking/addon.json
+++ b/addons/block-cherry-picking/addon.json
@@ -3,7 +3,7 @@
   "description": "Flips the block dragging controls. Drag a single block out of the middle of a script by default, or hold the Ctrl key to drag the entire stack attached below it.",
   "info": [
     {
-      "text": "To drag a single block without this addon hold Ctrl.",
+      "text": "To drag a single block without this addon, hold Ctrl.",
       "id": "vanilla"
     },
     {
--- a/addons/block-cherry-picking/addon.json
+++ b/addons/block-cherry-picking/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Grab single block by default",
-  "description": "Inverts the block dragging controls. Drag a single block out of the middle of a script by default and hold the Ctrl key to drag the entire stack attached below it.",
+  "description": "Flips the block dragging controls. Drag a single block out of the middle of a script by default, or hold the Ctrl key to drag the entire stack attached below it.",
   "info": [
     {
       "text": "To drag a single block without this addon hold Ctrl.",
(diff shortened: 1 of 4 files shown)


DNin01:
> The feature is now included in the editor

No way!

DNin01 (commented):
> I don't see a reason to keep the option to flip the controls around.

We _could_ keep the addon but...
- Disable it for existing users that don't have the "flip controls" setting enabled
- Remove the "flip controls" setting but force the flipped controls
- Rename the addon to "Always cherry pick blocks" or something like that

The advantages of keeping it are that the addon doesn't inexplicably vanish and we can use it to give users a pro tip. However, this probably won't land until v1.46.0.

Samq64:
By default without any addons cherry picking is bound to both to Alt and Ctrl.

DNin01:
> By default without any addons cherry picking is bound to both to Alt and Ctrl.

`block-duplicate` could override that if it's enabled. We don't have to mention the Alt key in this addon but could still support it if `block-duplicate` is disabled.

Anyhow, it would be nice if it was possible to change the shortcut keys... someday.

mxmou:
> The advantages of keeping it are that the addon doesn't inexplicably vanish and we can use it to give users a pro tip.

There's also the much more important advantage that if someone currently has the setting enabled and is used to the inverted controls, they won't have to learn to use the modifier key in the exact opposite way. I think we should keep the addon and rename it, maybe to "Grab single block by default" (not "always" because the modifier can still be used to grab a stack).

mxmou:
The flip controls setting already supports the Alt key if block-duplicate is disabled: I added that in #8430.

addons/block-cherry-picking/addon.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "Grab single block with Ctrl key",
-  "description": "Adds the ability to drag a single block out of the middle of a script (instead of the entire stack attached below it) while holding the Ctrl key.",
+  "name": "Grab single block by default",
+  "description": "Inverts the block dragging controls. Drag a single block out of the middle of a script by default and hold the Ctrl key to drag the entire stack attached below it.",
```
mxmou:
```suggestion
  "description": "When dragging blocks out of the middle of a script, only moves a single block by default. Hold the Ctrl key to drag the entire stack attached below it.",
```

I also suggest adding a note (similar to the one in disable-paste-offset) explaining how cherry picking works in vanilla, because it was only added recently and many people probably don't know about it.

Samq64: I was trying to keep the words invert and controls for searchability.

mxmou: That makes sense. Then it should probably be "flips", not "inverts" (in the current version, searching for "flip controls" works, but "invert controls" only shows the gamepad support addon).

In the second sentence, "or" might make more sense than "and" since you can't drag both a single block and the whole stack at the same time:

```suggestion
  "description": "Flips the block dragging controls. Drag a single block out of the middle of a script by default, or hold the Ctrl key to drag the entire stack attached below it.",
```

addons/block-cherry-picking/addon.json
```diff
@@ -1,7 +1,11 @@
 {
-  "name": "Grab single block with Ctrl key",
-  "description": "Adds the ability to drag a single block out of the middle of a script (instead of the entire stack attached below it) while holding the Ctrl key.",
+  "name": "Grab single block by default",
+  "description": "Flips the block dragging controls. Drag a single block out of the middle of a script by default, or hold the Ctrl key to drag the entire stack attached below it.",
   "info": [
+    {
+      "text": "To drag a single block without this addon hold Ctrl.",
```
mxmou:
```suggestion
      "text": "To drag a single block without this addon, hold Ctrl.",
```

mxmou (approved):

PR #8984, opened 2026-04-21
animated-thumb: change dropdown color, remove unused code

mxmou:
### Changes

* Removes unused code from before the Set Thumbnail button was moved to the editor.
* Changes the color of the dropdown to white.

image

### Tests

Tested on Edge.


addons/animated-thumb/userscript.css
```diff
@@ -137,7 +117,7 @@ p.sa-animated-thumb-uploaded-thumb {
 }
 
 .sa-set-thumbnail-upload-button:hover {
-  background-color: #4280d7;
+  background-color: color-mix(in srgb, var(--editorDarkMode-input, white), var(--editorDarkMode-primary, #855cd6) 15%);
```
mxmou, line 120: I used color-mix because I didn't want to add a new customCssVariable specifically for this. If someone is using an older browser that doesn't support color-mix, the color won't change on hover, but the cursor will still change which I think is good enough.

DNin01: It could also just become purple like hovered context menu items.

mxmou: I tried that and didn't like how it looked. There aren't any other purple elements in the stage header, so it really stands out if it becomes purple when hovered.

PR #8987, opened 2026-04-21
Remove new table buttons (reverted)

DNin01:
Removes the arrow buttons from table items, added in #8934. Other design changes are kept, leaving it to look like this:

Redesigned setting table interface

This is temporary; the buttons can be added again in a feature update or after v1.45.1 once the new "move up" and "move down" strings are translated.

@Samq64 @mxmou What do you think of this plan?

### Follow-up

- [ ] Re-add in v1.46.0

### DIFF
--- a/webpages/settings/components/addon-setting.html
+++ b/webpages/settings/components/addon-setting.html
@@ -47,16 +47,12 @@
           </div>
         </div>
         <div class="split-button setting-table-dropdown dropdown-parent">
-          <button
-            :disabled="!addon._enabled"
-            class="icon-button split-button-button"
-            :title="msg('addRow')"
-            @click="addTableRow()"
-          >
+          <button :disabled="!addon._enabled" class="large-button split-button-button" @click="addTableRow()">
             <img class="icon-type" src="../../images/icons/plus.svg" draggable="false" />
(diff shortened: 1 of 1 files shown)


mxmou:
> @Samq64 @mxmou What do you think of this plan?

I like it. I think the label on the Add row button can also stay since that string isn't new.

mxmou:
> Re-add in v1.46.0

Instead of merging this into master now and then reverting it after 1.45.1, this could be merged into a separate v1.45.x branch.

DNin01:
> I think the label on the Add row button can also stay since that string isn't new.

Yeah, I probably should have kept it because I kept the other design updates, too.

...Or we could try another design with only icons on the row add button, but with a background that matches that of the rest of the table items. It's supposed to show that it's related to the table interface:
New setting table design

mxmou:
I think the add row button should stand out somewhat, since it's one of the most important elements in the whole table setting interface.

DNin01:
I'll just un-revert the row add button.

[Updated.](https://github.com/ScratchAddons/ScratchAddons/pull/8987/commits/1615f2d67f0c3f628728c6849a6f255d8b5aee1c)

mxmou (approved):

PR #8995, opened 2026-04-23
middle-click-popup: more spork fixes

### LINKED ISSUE #8991: middle-click-popup can create "cursed inputs"
### Description

Blocks created using the middle click popup can have "cursed inputs". Other addons that create or duplicate blocks might have the same issue, but I haven't tested them yet.

### Steps to reproduce

1. Use the middle click popup to add two "move (my variable) steps" blocks
2. Delete one of them
3. Switch to another sprite and back
4. Remove the variable reporter from the input

Image

### Environment

Browser: Edge 149 dev
Scratch Addons version: cd27f3e7280f06fdc4ae001c8f41bd1a36f2d3e7
Operating system: Windows

mxmou:
Resolves #8991

### Changes

* Fixes cursed inputs by using Scratch's `serializeBlock()` method, which creates a representation of the block that doesn't include shadow IDs, forcing new ones to be generated when the block is added to a workspace.
* Fixes the popup closing when Shift+dragging blocks out of it. The Shift modifier is meant to keep the popup open so that multiple blocks can be dragged out without having to type the query each time.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/middle-click-popup/BlockTypeInfo.js
+++ b/addons/middle-click-popup/BlockTypeInfo.js
@@ -275,13 +275,10 @@ export class BlockInstance {
    * @returns {*} A 'workspace form' block.
    */
   createWorkspaceForm() {
-    if (this.typeInfo.id === "control_stop") {
-      this.typeInfo.domForm
-        .querySelector("mutation")
-        .setAttribute("hasnext", "" + (this.inputs[0].value === "other scripts in sprite"));
-    }
-
-    const block = this.typeInfo.Blockly.Xml.domToBlock(this.typeInfo.domForm, this.typeInfo.workspace);
+    const block = this.typeInfo.Blockly.serialization.blocks.append(
+      this.typeInfo.serializedForm,
+      this.typeInfo.workspace
+    );
     for (let i = 0; i < this.typeInfo.inputs.length; i++) {
       const inputValue = this.inputs[i];
       if (inputValue !== null) this.typeInfo.inputs[i].setValue(block, inputValue);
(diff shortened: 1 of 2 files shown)


Joeclinton1 (approved):
The bigger bug this fixes is that it can literally cause your blocks to be deleted from the workspace. I have been unable to repro the cursed input bug even without this fix so I suspect that was a scratch problem not us.

I also tested the bug about popup closing with shift and that works.

I don't understand the code well enough to know if it's a "good" solution, but it seems fine and it works.

WorldLanguages:
Change of plans: after this is merged, I will delete the `v1.45.1` github release tag and recreate it.

v1.45.1 will be released with this fix included.

(Reasoning: in the store version, this addon does not work, so by releasing without this fix, the addon works but with bugs, so we would be making data loss likely to happen for our users)

mxmou:
@WorldLanguages I'm working on a similar fix for block-duplicate. That one should probably be included too - since the bugs are closely related, I suspect they can both cause the code deletion issue.

mxmou:
> I have been unable to repro the cursed input bug even without this fix so I suspect that was a scratch problem not us.

I can still repro on master by following the steps in the issue. Switching to another sprite and back before removing the reporter is a necessary step.

Joeclinton1:
> > I have been unable to repro the cursed input bug even without this fix so I suspect that was a scratch problem not us.
> 
> I can still repro on master by following the steps in the issue. Switching to another sprite and back before removing the reporter is a necessary step.

Ah ok, when you say delete one of them, I thought you meant delete the variable. inside the block not both blocks. Have reproduced it now. And fortunately can confirm your pr fixes it.

mxmou:
I've opened the block-duplicate PR: #8997

PR #8996, opened 2026-04-23
v1.45.2-pre

mxmou:


WorldLanguages:
Merge this if you want, but I also have no problem with cherrypicking into the v1.45.x branch if needed

PR #8997, opened 2026-04-23
block-duplicate: spork fixes

mxmou:
### Changes

Fixes two bugs in the block-duplicate addon.

#### Duplicating custom block prototypes

To reproduce:
* Try to Alt+drag a prototype out of a define block. It looks like nothing has happened, but this actually creates a duplicate of the prototype.
* Move the define block out of the way to reveal the duplicate.

#### Cursed inputs

To reproduce:
* Put a variable into a text or number input
* Duplicate the block using Alt+drag
* Delete the original block
* Switch to another sprite and back
* Remove the variable from the input

Fixed by using the `stripIds()` utility function that Scratch added for this purpose.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/block-duplicate/module.js
+++ b/addons/block-duplicate/module.js
@@ -47,7 +47,11 @@ export async function load(addon) {
       }
 
       const isDuplicating =
-        enableDuplication && e.altKey && !this.flyout && this.targetBlock.type !== "procedures_definition";
+        enableDuplication &&
+        e.altKey &&
+        !this.flyout &&
+        this.targetBlock.type !== "procedures_definition" &&
+        this.targetBlock.type !== "procedures_prototype";
 
       if (isDuplicating) {
         this.startWorkspace_.setResizesEnabled(false);
--- a/addons/block-duplicate/module.js
+++ b/addons/block-duplicate/module.js
@@ -54,8 +54,9 @@ export async function load(addon) {
         ScratchBlocks.Events.disable();
         let newBlock;
         try {
-          const xmlBlock = ScratchBlocks.Xml.blockToDom(this.targetBlock);
-          newBlock = ScratchBlocks.Xml.domToBlock(xmlBlock, this.startWorkspace_);
+          const serializedBlock = ScratchBlocks.serialization.blocks.save(this.targetBlock);
+          ScratchBlocks.scratchBlocksUtils.stripIds(serializedBlock);
+          newBlock = ScratchBlocks.serialization.blocks.appendInternal(serializedBlock, this.startWorkspace_);
           const xy = this.targetBlock.getRelativeToSurfaceXY();
           newBlock.moveBy(xy.x, xy.y);
         } catch (e) {


Joeclinton1:
Almost certainly btw it's about any block in an input not just variables.

Joeclinton1:
Can confirm you get the same block deletion bug. if you remove the variable from both the original and duplicated block.

https://github.com/user-attachments/assets/acb029e8-a7af-4faa-9b1b-1b844969e575

(this is without the fix to be clear)

Joeclinton1 (dismissed):
https://github.com/user-attachments/assets/d7a9f13e-23f4-4c53-8ea5-edce306e4abd

Not fixed, in fact your blocks now disappear faster.

Joeclinton1:
Problem is upstream not us (happens also with regular duplicate and no extension), will report to scratch on their spork issue lisst

WorldLanguages:
You discard your review then?

Joeclinton1:
> You discard your review then?

done

Joeclinton1 (dismissed):
The duplicate for define prototypes bug is not fixed. First time I drag them out fine. Second time though I move the define block and the two inputs appear in top left of workspace.

the cursed input for variables is fixed.

WorldLanguages:
Should we release v1.45.1 with or without this?

Samq64 (approved):
I think v1.45.1 should be pulled if it hasn't already and v1.45.2 should be released with this change.

Edit: I did not notice Joe's review until now but I just tested and can't reproduce the bug.

Joeclinton1:
> I think v1.45.1 should be pulled if it hasn't already and v1.45.2 should be released with this change.
> 
> Edit: I did not notice Joe's review until now but I just tested and can't reproduce the bug.

Let me make a video then. Because there was a bug for sure.

mxmou:
> I move the define block and the two inputs appear in top left of workspace

That doesn't sound as serious as the bug that happens on master and is fixed in this PR:

https://github.com/user-attachments/assets/429674ad-d51a-4bc3-9da4-eeecdd84ea4b

The duplicated prototype block this creates can't be dragged or even deleted. So I think this PR definitely improves things, even if there are still some bugs.

DNin01:
> Should we release v1.45.1 with or without this?

https://github.com/ScratchAddons/ScratchAddons/commit/0e9232432c4a03d1cbee54349ba2c2df1501c9b2 should be included in the next release, if anything.

Joeclinton1:
After some difficulty trying to reproduce the bug I figured it out. It happens when you drag the custom block over a connection point. And is maybe a different bug?

oh. It's not even us, it's again: upstream



https://github.com/user-attachments/assets/d8ff52fc-3298-4c64-bf25-07731fe956da

Joeclinton1 (approved):
The PR works well, and removes the two serious bugs.

mxmou:
@WorldLanguages There's an open PR in scratch-blocks that will break the cursed inputs fix from this PR. See #9003

WorldLanguages:
Okay, releasing another patch soon then.

PR #9003, opened 2026-04-25
block-duplicate: use return value of stripIds

mxmou:
### Changes

There's an open PR in scratch-blocks that changes `stripIds()` to create a new object instead of mutating the argument, which will break the cursed inputs fix introduced in #8997. This PR updates the code that calls `stripIds()` to use the returned value if there's one.

### Tests

Verified that the cursed inputs bug is still fixed on scratch.mit.edu and on the [fix/stripIds-mutation](https://github.com/scratchfoundation/scratch-blocks/tree/fix/stripIds-mutation) branch of scratch-blocks.

### DIFF
--- a/addons/block-duplicate/module.js
+++ b/addons/block-duplicate/module.js
@@ -58,8 +58,9 @@ export async function load(addon) {
         ScratchBlocks.Events.disable();
         let newBlock;
         try {
-          const serializedBlock = ScratchBlocks.serialization.blocks.save(this.targetBlock);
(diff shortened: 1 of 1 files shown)


Joeclinton1:
I saw the stripids thing yesterday and was like huh that sounds relevant. But didn't piece together it would break your fix.

Joeclinton1:
You can request me to review you know if you want something done faster

DNin01 (approved):
Untested but looks right

PR #9008, opened 2026-04-26
Add landmarks to sidebar and main settings

DNin01:
This pull request adds landmarks to the Scratch Addons settings page, helping screen reader users navigate and understand the general layout of the settings page (many screen readers allow users to jump between landmarks and other important elements for faster navigation).

Image showing landmark layout of settings page

### How I made this

First, I split up the settings page sidebar into sections. Previously, the sidebar was a flat list containing every addon category and other element on one level, but now each type of item lives inside a different parent element. There are two of these on the sidebar, one for the addon categories and one for the other items like More Settings and Send Feedback.

This also allowed me to transfer the margin of 12px below the last addon category list item to the parent `div`, making those components more flexible to develop.

After that, it was easy to convert the two new sidebar sections to landmark elements: `<aside>` for the addon category list (which is labeled "Addon filters") and `<nav>` for the rest, and to put both of them in a group called "Sidebar". The addon category list is `<aside>` because it is complementary to the main content, the addon settings.

I also added the landmark elements `<header>`, `<main>`, and `<search>`.

### Testing

For testing, activate a screen reader and ensure that the landmarks are announced properly when navigated to (e.g., `<main>` should be read as something like "main landmark").

I used Windows Narrator with Microsoft Edge.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -149,8 +149,8 @@
   "close": {
     "message": "Close"
   },
-  "filters": {
-    "message": "Addon filters"
+  "categoryFilter": {
+    "message": "Addon category filter"
   },
   "resetToDefault": {
     "message": "Reset to default settings"
@@ -265,6 +265,9 @@
   },
   "search": {
     "message": "Search"
+  },
+  "addonsSearch": {
+    "message": "Addons Search"
   },
   "permissionsTitle": {
     "message": "Permissions - Scratch Addons"
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -149,8 +149,8 @@
   "close": {
     "message": "Close"
   },
-  "filter": {
-    "message": "Filter:"
+  "filters": {
+    "message": "Addon filters"
   },
   "resetToDefault": {
     "message": "Reset to default settings"
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -143,6 +143,9 @@
   "moreSettings": {
     "message": "More Settings"
   },
+  "sidebar": {
+    "message": "Sidebar"
+  },
   "toggleSidebar": {
     "message": "Toggle sidebar"
   },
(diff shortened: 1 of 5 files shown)


Samq64 (approved):
Not tested with a screen reader

DNin01:
I don't know if anyone knows more than I do about landmarks, but I keep going back and forth on the `<aside>` element—a [complementary landmark](https://developer.mozilla.org/docs/Web/Accessibility/ARIA/Reference/Roles/complementary_role)—that I used in this PR. I believed it was the best match for the addon category selection in the sidebar because `<nav>`—a [navigation landmark](https://developer.mozilla.org/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role)—is typically used for sets of links or tabs and isn't used to control the main content like the addon category selector is.

Complementary landmarks are supposed to contain content that is indirectly related to the main content, for example, a "related articles" section on a blog. In our case, the addon categories make up a UI widget for the addon settings panel but with standalone content (a list of addon categories instead of addon settings), if that makes sense.

Either way, there's the fact that the "toggle sidebar" button is essentially controlling the visibility of two different menus, `<aside>` and `<nav>`, that _appear_ to be on the same panel.

Until one day we rethink our settings page layout to better accommodate non-visual representations, there are a number of roles to consider assigning the category selector, ~but I think `<aside>` still fits it better than any other landmark role.~

---

**Edit:** No, that's still not right. `<aside>` seems to be for sidebars with tangentially related information, and I'm not finding anything about usage with UI widgets. It's not a tab panel either because there aren't any tabs, there are only filters.

DNin01:
@mxmou This is basically ready if you want to review or ask questions.

DNin01:
I still think the landmarks are good enough, but I've also added a label to the sidebar (without adding or removing any landmarks), mainly to provide context for the "toggle sidebar" button.

So, when you navigate the settings page with a screen reader, you'll hear something like this:
- "Settings - Scratch Addons", document
  - <kbd>→</kbd> banner landmark
    - <kbd>→</kbd> "Logo", image
    - <kbd>→</kbd> "Settings", heading level 1
    - <kbd>→</kbd> "Switch to light theme", button
  - <kbd>→</kbd> "Sidebar", group
    - <kbd>→</kbd> "Addon category filter", complementary landmark
      - <kbd>→</kbd> "All", button...

Issue #8967, opened 2026-04-18
Clean up blocks + sometimes does not respect columns


Joeclinton1:
### Description

Sometimes clean up blocks + will change two columns back to 1.

### Steps to reproduce

1. Try creating 3 columns of blocks
2. Use clean up blocks+
3. Often it fails to clean up into columns correctly.

### Demonstration(s) and additional context

https://github.com/user-attachments/assets/ce60a21e-9278-436a-942a-a82dfd5ee056

### Environment

Browser: 
Scratch Addons version: 
Operating system:

mxmou:
This doesn't seem to be a new bug - I can repro in TurboWarp with the same arrangement of blocks.

Joeclinton1:
Ok, well we should probably this this anyway. But I guess lower priority.

PR #9024, opened 2026-05-10
Bug fixes for editor-cleanup-plus (general failure, gap uniformity and makeSpace Y shift)

griffpatch:
Resolves #

### Changes

- **`editor-devtools` - `makeSpace`**: fixed two bugs where scripts below didn't move when they should, and overlapping scripts weren't pushed far enough out of the way.
- **`editor-cleanup-plus`**: fixed oversized / uneven gaps between scripts when the script-snap addon is not active.
- **`editor-cleanup-plus` / `devtools-utils`**: running "Clean Up Scripts" twice no longer merges narrow columns into one - columns are now always spaced further apart than the grouping threshold

### Reason for changes

**`makeSpace` - scripts below don't move when right-clicking inside a nested (indented) block**

If you right-clicked a block that was indented (e.g. inside a loop), "Make Space" would often do nothing for the scripts below - they'd stay put. The fix was to always measure the target stack's position from its root (hat) block rather than the clicked block. An indented block reports a position further right than the stack's actual left edge, which caused the column-overlap check to miss nearby scripts entirely.

**`makeSpace` - overlapping scripts don't move far enough**

If another script was sitting on top of or partially overlapping the target stack, "Make Space" would only shift it a little - not enough to actually clear the target. The Y shift is now measured from the bottom of the target stack rather than its top, so scripts always land fully below it.

**`editor-cleanup-plus` - large uneven gaps without script-snap**

When the script-snap addon was not enabled, "Clean Up Scripts" still used the grid size as the gap between stacks and applied grid-snap rounding. This made gaps much larger than necessary and visually uneven. A flat 50px gap is now used when script-snap is off, matching what you'd expect from a tidy layout.

### Tests

Tested in Chrome.

- Right-click a block inside a loop → **Make Space** → scripts below now shift ✓
- Right-click a top-level hat block → **Make Space** → behaviour unchanged ✓  
- Script overlapping the target → pushed fully clear after Make Space ✓
- **Clean Up Scripts** with script-snap off → compact, even gaps ✓
- **Clean Up Scripts** with script-snap on → grid-aligned gaps as before ✓
- Ctrl+Z after Make Space → all moves undo in one step ✓
- Three narrow columns (e.g. lone hat blocks) → Clean Up → Clean Up again → columns stay separate ✓

### DIFF
--- a/addons/editor-cleanup-plus/userscript.js
+++ b/addons/editor-cleanup-plus/userscript.js
@@ -5,8 +5,16 @@ import {
   autoPositionComment,
 } from "../../libraries/common/cs/devtools-utils.js";
 
+// Gap between stacks when script-snap is not active (not constrained to grid size)
+const NON_SNAP_GAP = 50;
+
 export default async function ({ addon, console, msg, safeMsg: m }) {
   const blockly = await addon.tab.traps.getBlockly();
+
+  const isScriptSnapEnabled = async () => {
+    const enabledAddons = await addon.self.getEnabledAddons();
+    return enabledAddons.includes("script-snap");
+  };
 
   const originalMsg = blockly.Msg.CLEAN_UP;
   addon.self.addEventListener("disabled", () => (blockly.Msg.CLEAN_UP = originalMsg));
@@ -16,12 +24,13 @@ export default async function ({ addon, console, msg, safeMsg: m }) {
   const oldCleanUpFunc = blockly.WorkspaceSvg.prototype.cleanUp;
   blockly.WorkspaceSvg.prototype.cleanUp = function () {
     if (addon.self.disabled) return oldCleanUpFunc.call(this);
-    doCleanUp();
+    void doCleanUp();
   };
 
-  const doCleanUp = () => {
+  const doCleanUp = async () => {
     const workspace = addon.tab.traps.getWorkspace();
     const promptUnused = addon.settings.get("promptUnused");
+    const scriptSnapEnabled = await isScriptSnapEnabled();
 
     UndoGroup.startUndoGroup(workspace);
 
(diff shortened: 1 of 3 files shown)


Samq64:
Not a bug caused by this PR but cleaning this up twice results in a single column:
image

griffpatch:
Yes, I have seen this too... It's because the column width is counted as
too narrow, and the cleanup overzealously stuffs it into a single column...
I think I could improve this by applying a minimum distance to space out
columns... one that is larger than the minimum considered distance for
bringing them back into a single column

Joeclinton1:
https://github.com/ScratchAddons/ScratchAddons/issues/8967 is it related to this?

Also I'm the one that made the make space and clean up block plus changes, so I probably need to review this.

griffpatch:
**Reply to Samq64's comment** (the screenshot of three columns merging into one):

Good spot! This is now fixed in the latest commit. The root cause was that the column-grouping code in devtools-utils.js uses a 256px tolerance — any two stacks within 256px get merged into the same column on the next cleanup pass. Narrow blocks like a lone "when green flag clicked" only placed columns ~150px apart, well within that threshold.

The fix has two parts:
- devtools-utils.js — reduced the merge threshold from 256px to 128px, exported as a shared `COLUMN_GROUP_TOLERANCE` constant
- `doCleanUp` — columns now always advance by at least `COLUMN_GROUP_TOLERANCE`, so a tidied workspace stays stable no matter how many times you clean it up

addons/editor-cleanup-plus/userscript.js
```diff
@@ -3,11 +3,20 @@ import {
   getVariableUsesById,
   getOrderedTopBlockColumns,
   autoPositionComment,
+  COLUMN_GROUP_TOLERANCE,
 } from "../../libraries/common/cs/devtools-utils.js";
 
+// Gap between stacks when script-snap is not active (not constrained to grid size)
+const NON_SNAP_GAP = 50;
```
mxmou, line 10: Vanilla Scratch uses `workspace.renderer.getConstants().MIN_BLOCK_HEIGHT` as the gap, which is 48 by default but can change if the custom block shape addon is enabled (if blocks are bigger, the gaps are also bigger). I think it would be nice if the addon did the same.

mxmou (commented):
Works perfectly on Firefox as well. I just have one small comment.

PR #9029, opened 2026-05-16
Fix for Bug: "Insert blocks by name" creates two undo points when placing a block

griffpatch:
## Bug: "Insert blocks by name" creates two undo points when placing a block

**Summary**  
When using the **Insert blocks by name** addon to place a block, pressing Ctrl+Z once doesn't remove it - instead the block jumps back to where your cursor was when you opened the search popup. A second Ctrl+Z is needed to actually remove the block. Expected behaviour is one Ctrl+Z removes the block entirely.

## Bug: "Insert blocks by name" creates two undo points when placing a block

**Summary**  
When using the **Insert blocks by name** addon to place a block, pressing Ctrl+Z once doesn't remove it - instead the block jumps back to where your cursor was when you opened the search popup. A second Ctrl+Z is needed to actually remove the block. Expected behaviour is one Ctrl+Z removes the block entirely.

## Steps to reproduce

1. Make sure the **Insert blocks by name** addon is enabled
2. Open a Scratch project and go to the code editor
3. Middle-click (or Shift-click) in the workspace to open the block search
4. Search for any block (e.g. "move"), select it, and drag it onto the workspace
5. Press **Ctrl+Z** - the block snaps back to the popup cursor position instead of disappearing
6. Press **Ctrl+Z** again - the block is finally removed

## Expected behaviour

A single Ctrl+Z removes the block entirely, as it does when dragging a block from the palette.

## Root cause

The addon fires a `BlockCreate` event before starting the drag. Blockly's drag start creates a new event group for the drag, leaving `BlockCreate` in a separate group - so Scratch sees two distinct undo steps: one for the creation (at the popup cursor position), and one for the drop (where you released the block).

## Fix

Fire `BlockCreate` after `onDragStart` so it joins the drag's event group. `BlockCreate` and `BlockMove` then undo together as a single step, matching the behaviour of blocks dragged from the palette.

### DIFF
--- a/addons/middle-click-popup/userscript.js
+++ b/addons/middle-click-popup/userscript.js
@@ -122,7 +122,7 @@ export default async function ({ addon, msg, console }) {
       // 4. everything else
       const prio = (block) =>
         ["operators", "data"].indexOf(block.category.name) -
-        (block.id.startsWith("data_") && block.id !== "data_variable");
+        (block.id.startsWith("data_") && block.id !== "data_variable" ? 1 : 0);
       return prio(b) - prio(a);
     });
 
@@ -138,10 +138,13 @@ export default async function ({ addon, msg, console }) {
     popupInput.focus();
     updateInput();
 
+    // @ts-ignore - registry exists in scratch-blocks but not in standard Blockly types
     if (Blockly.registry) {
-      // new Blockly: register delete area
+      // new Blockly: register delete area (remove first in case a previous open didn't clean up)
+      workspace.getComponentManager().removeComponent("saMiddleClickPopup");
       const component = new Blockly.DeleteArea();
       component.id = "saMiddleClickPopup";
+      // @ts-ignore - Blockly type mismatch between versions
       component.getClientRect = () => {
         const rect = popupContainer.getBoundingClientRect();
         return new Blockly.utils.Rect(rect.top, rect.bottom, rect.left, rect.right);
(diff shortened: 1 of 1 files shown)


addons/middle-click-popup/userscript.js
```diff
@@ -138,10 +138,13 @@ export default async function ({ addon, msg, console }) {
     popupInput.focus();
     updateInput();
 
+    // @ts-ignore - registry exists in scratch-blocks but not in standard Blockly types
```
mxmou: Which Blockly types is this referring to? I don't think we've ever used any Blockly type definitions in this project.

addons/middle-click-popup/userscript.js
```diff
@@ -138,10 +138,13 @@ export default async function ({ addon, msg, console }) {
...
     if (Blockly.registry) {
-      // new Blockly: register delete area
+      // new Blockly: register delete area (remove first in case a previous open didn't clean up)
+      workspace.getComponentManager().removeComponent("saMiddleClickPopup");
       const component = new Blockly.DeleteArea();
       component.id = "saMiddleClickPopup";
+      // @ts-ignore - Blockly type mismatch between versions
```
mxmou: Same here.

mxmou (commented):
The main changes look good and the fix works, but I don't understand some of the added ts-ignore comments.

addons/middle-click-popup/userscript.js
griffpatch: Good catch. These were stale comments carried over from an unmerged experiment that added Scratch Blocks type declarations. This PR does not use those declarations, so I removed these and the other newly added ts-ignore comments. Thanks for spotting this.

addons/middle-click-popup/userscript.js
griffpatch: Removed this one as well. It came from the same stale type experiment.

addons/middle-click-popup/userscript.js
griffpatch: You are right - this is my bad - I have a much more complex branch where I used typescript and backported this fix from... I didn't intend the typescript comments to leak in.

mxmou (approved):

PR #8816, opened 2026-01-31
Customize member decorations

DNin01:
Resolves #8729

### Changes

This addon lets you remove the cat ears from members' avatars and change how the "Member" label looks.

This works in several areas of the Scratch website, including browse pages, project and studio pages, and forums.

Thumbnail of a project by member codubee Member and non-member variants of title and creator side by side
A forum topic by member ScratchCat

### Reason for changes

Some people may find these new elements distracting or unhelpful, and to remove them might improve the experience.

### Tests

Tested in Edge 144.

PR #8744, opened 2025-12-24
Hide membership

random-beta-tube:
Resolves #8729.

### Changes

Adds the extension "hide membership" which hides the following (all of them have settings to show/hide seperately):
* Membership badges on user profiles who are Scratch members
* "Member" flairs next to Scratch members' usernames
* The cat avatars that some Scratch members may have enabled.

Note that the cat avatar setting _probably_ doesn't hide your own cat avatar in the header (I can't test this because I don't have Scratch membership), but Scratch members can disable their own cat avatars and badges on their account settings so this is not a problem.

image
image

> Update after inital PR:
> * I figured out how to use Actions so the code is now properly formatted.

### Reason for changes

Some users might find the new membership decorations annoying.

### Tests

The badges and flairs is a fairly straightforward userstyle which I tested on Chrome 143. The cat avatars are a bit more complicated  as different parts of the website have different avatar settings but I have implemented and the following parts of the website (let me know if I missed one):
* Main page ("What's happening")
* Messages
* Project pages (the icon in the top left, remix credits and comments)
* Studio pages (both curators and comments)
* User profiles (following/followers and comments)
* The explore page
* Forums

> I'm new to contributing so sorry if I made any mistakes!

PR #9030, opened 2026-05-16
add remove-membership addon

### LINKED ISSUE #8729: New addon: Hide membership badges
### Why this would be helpful

The membership labels are annoying and might unintentionally lead users into bullying.

### How this addon works

Flip the toggle and all members badges will instantly disappear.

### Possible alternatives

_No response_

### Additional context

_No response_

rosics-code:
### Resolves
People hate membership

### Changes

It uses JS to remove membership from the site.

### Reason for changes

People hate membership.

### Tests

Tested locally:
Screenshot 2026-05-16 7 41 42 PM

### DIFF
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -1,4 +1,4 @@
-f[
+[
   "editor-searchable-dropdowns",
   "editor-devtools",
   "forum-search",
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -1,4 +1,4 @@
-[
+f[
   "editor-searchable-dropdowns",
   "editor-devtools",
   "forum-search",
@@ -82,6 +82,7 @@
   "no-script-bumping",
   "user-id",
   "hide-delete-button",
+  "hide-membership",
   "name-backpack-items",
   "forum-copy-code",
   "paint-by-default",
--- a/addons/hide-membership/addon.json
+++ b/addons/hide-membership/addon.json
@@ -1,6 +1,6 @@
 {
   "name": "Hide Membership",
-  "description": "Hides Scratch Membership (e.g. \"Member\" Badge, Membership Images, etc.).",
+  "description": "Hides Scratch Membership badges.",
   "tags": ["community", "theme"],
   "credits": [
     {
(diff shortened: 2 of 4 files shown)


CST1229:
i think the title and description need further clarification (like, "Hide membership badges")

also i feel like this could just be implemented as a userstyle instead of a userscript

rosics-code:
> i think the title and description need further clarification (like, "Hide membership badges")
> 
> also i feel like this could just be implemented as a userstyle instead of a userscript

@CST1229 Does user styles have it so when you click load more in a comment, it auto hides? I forgot.

rosics-code:
@CST1229 Is it fine if its a userscript though?

DNin01:
> @CST1229 Is it fine if its a userscript though?

It is _okay_ to use a userscript if things other than style properties are changed, for example, class lists, which are HTML attributes, but if the script is used only for setting CSS values, a userstyle would probably be able to do the same thing more efficiently if the style rules are prioritized to override others using [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity) and/or the [`!important` keyword](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/important).

Userstyles also have the advantage of affecting content immediately on page load if the addon is assigned the property [`injectAsStyleElt: true`](https://scratchaddons.com/docs/reference/addon-manifest/#injectasstyleelt). This is important for addons like dark mode if you don't want screen flashes.

rosics-code:
@DNin01 Should be fixed. The JS is still there because some things needs to use classes removal to not leave gaps like the membership ears.

rosics-code:
@mxmou What do you think?

Samq64:
There are already a couple of alternative implementations open:
- #8816
- #8744

rosics-code:
> There are already a couple of alternative implementations open:
> 
> * [Customize member decorations #8816](https://github.com/ScratchAddons/ScratchAddons/pull/8816)
> * [Hide membership #8744](https://github.com/ScratchAddons/ScratchAddons/pull/8744)

@Samq64 The second one doesn't close the gap on profile between the profile picture and username so this one fixes that. First one customizes it.

rosics-code:
Now a draft since membership is being paused.

mxmou:
Closing because the membership features have been removed by Scratch.

PR #9037, opened 2026-05-23
Settings page - Load addons lazily

DNin01:
Finally. Should help with #2531, something @WorldLanguages knows is pretty annoying.

https://github.com/user-attachments/assets/dc41d646-29ac-4a5e-aebf-6795e525bf5a

### Description

This pull request changes two things which significantly contributed to the **perceived** loading time of the list of addon settings:
1. Rendering every item in the list, including those offscreen, wasted time. Now, all contents of addon list items and loaded lazily, which means they render only after they are scrolled into view, using IntersectionObserver.
2. Populating the entire list in one pass also adds a delay. Now, a set number of items near the top of the page are rendered and the rest of the list continues to render after the screen refreshes, allowing the first page of content to be presented quicker.

Now the addon settings UI responds much faster after loading the page or returning to the main categories after opening a Related Addons list. I challenge you to go back to how it was before after testing this out!

Here are the most notable side effects caused by each change:
1. Addons may briefly appear blank before they render. You might see this if you scroll quickly or search, but it's a small trade-off for improved response time.
2. Numbers may appear dynamically and the scroll track may shift as items load.

Some code comments have also been clarified.

### Notes

Tested on Edge 148.

Some changes to `addon-body.js` were based on carefully reviewed code generated by Claude Haiku 4.5 via GitHub Copilot.

### DIFF
--- a/webpages/settings/components/addon-body.html
+++ b/webpages/settings/components/addon-body.html
@@ -1,6 +1,6 @@
 <template>
   <div class="addon-body" v-show="shouldShow" :id="'addon-' + addon._addonId">
-    <div class="addon-topbar">
+    <div class="addon-topbar" v-if="everWithinView">
       <div class="clickable-area" @click="expanded = !expanded">
         <button class="arrow-button" :title="msg(expanded ? 'collapse' : 'expand')">
           <img src="../../images/icons/expand.svg" :class="{'reverted': expanded}" draggable="false" />
@@ -28,7 +28,7 @@
         <input type="checkbox" class="switch" v-model="addon._enabled" @click="toggleAddonRequest" />
       </div>
     </div>
-    <div class="addon-settings" v-if="everExpanded" v-show="expanded">
+    <div class="addon-settings" v-if="everWithinView && everExpanded" v-show="expanded">
       <div class="addon-description-full">{{ addon.description }}</div>
       <div class="addon-message addon-update" v-if="showUpdateNotice">
         <addon-tag tag="new"></addon-tag>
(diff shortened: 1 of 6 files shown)


Samq64:
In my testing with the CPU throttled the initial load is roughly on par with Vue 3!

I do have a couple suggestions though:
1. Could the group headers wait until the final count is known before displaying it?
2. On slow systems there may be many placeholders when scrolling. Could those be replaced with a single loading message?

DNin01:
> 1. Could the group headers wait until the final count is known before displaying it?

Good idea for slower devices. Done.

> 2. On slow systems there may be many placeholders when scrolling. Could those be replaced with a single loading message?

All the placeholders trying to load while you're scrolling fast can be a hindrance, but I don't know of a simple way to fix this.

Ending the list after a few unloaded items with "Loading more...", if that's what you meant by "a single loading message", would have to account for collapsed groups and the variable order of list items among other things. I played around with throttling (limiting how many can be loading at the same time), but the way I did it adds the items loading to a permanent queue, even if they go offscreen before they load.

Let me know if you have any other ideas. At least a user can search, collapse groups, or filter to reduce the number of lazy items they have to scroll through.

Samq64:
> Ending the list after a few unloaded items with "Loading more...", if that's what you meant by "a single loading message"

Or a spinner. Either way I see now that it is not that easy to implement well.

DNin01:
For those who haven't tried it out yet, here's what scrolling through unrendered items looks like with CPU throttling active:

https://github.com/user-attachments/assets/da62c20c-677d-4546-bfcc-24b0d178b214

mxmou (approved):

Issue #9042, opened 2026-05-24
Adding a permission prompt for scratch.org


DNin01:
### Scratch website changes

Scratch Team member [codubee](https://scratch.mit.edu/users/codubee/) has recently answered some questions regarding the move from scratch.mit.edu to scratch.org.

> It's been the plan for a very long time. Kind of by definition, if we aren't a part of MIT anymore, then we normally wouldn't have been allowed to keep the subdomain of scratch.mit.edu. It's amazing that MIT has been open to us keeping it for this long in the first place.[^1]

[^1]: https://scratch.mit.edu/users/codubee/#comments-408160933

> ngp ( . . . ) will be reachable via scratch.mit.edu when we have finished launching it. It will redirect to scratch.org though.[^2]

[^2]: https://scratch.mit.edu/users/codubee/#comments-408066174

> ( . . . ) we hope that we'll be able to have scratch.mit.edu redirect to preserve old links in perpetuity.[^3]

[^3]: https://scratch.mit.edu/users/codubee/#comments-408893226

Along with official announcements from Scratch, these answers suggest that the website will redirect to https://scratch.org/ when the Scratch NGP rolls out, which Scratch has said is expected to happen by the end of this year[^4].

[^4]: https://scratchfoundation.org/announcements/ngp#:~:text=When%20is%20Scratch%20moving%20to%20the%20NGP%3F

### Updating permissions

To adapt, we'll need to plan a change to the list of permissions that the Scratch Addons extension has used for years.

Adding new entries to `host_permissions` [triggers a warning that disables the extension until resolved](https://developer.chrome.com/docs/extensions/develop/concepts/permission-warnings), which prevents us from communicating why we need the permissions. Instead, I suggest we add, perhaps even switch to, `optional_host_permissions`, which can be granted at runtime. Then, using a permission management screen that we'd nudge the user towards using banner notifications, we could make an effective prompt to grant the permissions that are necessary without obstructing other user flows.

https://github.com/user-attachments/assets/137ad2dc-92fc-4276-9f61-e6f9aa5dba5a

This system also has other opportunities:
- It could be integrated into a first-time setup or "welcome" experience, with a different prompt for new users.
- Currently, users who deny access are interrupted by an automatic permission prompt if they try to use any part of the extension. This intrusive prompt could be changed to a banner notification with a polite explanation and a link to the permission prompt. The hope is that a user will find their way there if the addons stop working after the move to scratch.org.
[...]

PR #9046, opened 2026-05-27
Move the Permissions prompt to the settings page

DNin01:
Related conversation: https://github.com/ScratchAddons/ScratchAddons/issues/9042#issuecomment-4547109061

### Changes

This changes the Permissions page to a dialog inside the settings page. Here's how it works:

When the popup detects missing permissions, the **settings page** is opened. The settings page will then open a popup prompting the user for those permissions, just like the original `permissions.html`.

Based on the full Permissions page, the popup is an identical prompt but can be dismissed easily.

There's a new feature too: accepting the permission request now displays the message, "Site access was enabled successfully. Scratch Addons will reload," confirming the action.

### Reason for changes

It should be easy to access the settings page for things that don't require permission, such as exporting settings.

---

| All changes | |
|---:|:---|
| 03179d2 | Remove permissions.html and move the content to a dialog in the settings page |
| 03a8973..0f3aa5a | Further updates to the handling of opening the dialog |
| 719450c, cff13b6..642f9eb | Design updates |

### Tests

- [x] Chromium test
- [x] Firefox test

Tested behaviors on Edge 149:
- [x] Settings page opens after deactivating permissions.
- [x] Settings page opens after opening the popup when the extension's required permissions are insufficient.
- [x] Dialog appears on the settings page when the extension's required permissions are insufficient.
- [x] Button action responds to both granted and rejected permissions.
- [x] Dialog looks correct in both large and small windows.

### Images

**\*\* See https://github.com/ScratchAddons/ScratchAddons/pull/9046#issuecomment-4813928963 for current design \*\***

Permissions modal on settings page asking the user to enable site access


Samq64 (approved):

Samq64:
Actually on Firefox the popup does not close after opening the settings page permission dialog.

DNin01:
> Actually on Firefox the popup does not close after opening the settings page permission dialog.

Should work now.

Samq64 (approved):

Samq64:
Something to consider is when the popup icon is clicked multiple times the existing settings tab could be re-focused using `chrome.runtime.openOptionsPage()` then `chrome.runtime.sendMessage()` could re-open the prompt.

webpages/settings/index.js
```diff
@@ -46,6 +46,14 @@ let fuse;
     "webpages/settings/components/previews/workspace-dots",
   ]);
 
+  if (!isIframe) {
+    chrome.runtime.sendMessage("checkPermissions", (hasPermissions) => {
+      if (!hasPermissions) {
```
mxmou, line 51: This isn't really necessary, but I think it would be nice if a specific `location.hash` would force the modal to appear, so that it could be tested without having to disable the permissions or modify the code.

mxmou (changes requested):
Depending on the window size, there can be a lot of empty space at the bottom of the popup. This could be fixed by setting the height to `auto` (with a `max-height` to avoid overflow) instead of a fixed value.

A few other improvements could be made to the popup's layout as well:

Image

Image

webpages/settings/index.js
DNin01: I thought of that too. I'm on it.

DNin01: Though, I feel that, opening the options page directly (not through the popup), it should still be able to prompt for permission without needing another page to open `#permissions`. But if we have both, one would be redundant. I could create a Vue method so that developers can open the modal by running `vue.openPermissionPrompt()` in the console, though.

DNin01: But at least I got to clean up some code related to URL hashes. (With [these 3 commits](https://github.com/ScratchAddons/ScratchAddons/pull/9046/changes/03a8973cea02e4b3e595b9b434cfa3b726113fee..0f3aa5aa24670b2bb4316162b74ab84365cfb0a5). Just extra changes that were easy to do at the same time.)

webpages/settings/style.css
DNin01: Should I make the button have an orange border? It might pop out more that way.

```suggestion
  font-size: 2.5rem;
  border-color: var(--orange);
```

Image

DNin01, line 1: Should I add step numbers, like this?

Image

DNin01:
@mxmou, I [updated the design](https://github.com/ScratchAddons/ScratchAddons/pull/9046/changes/0f3aa5aa24670b2bb4316162b74ab84365cfb0a5..642f9eb75287443e125722e8b4329f212997b4fd). How does it look?

Please enable site access modal

The code for showing the More Settings dialog when visiting `#moresettings` has also been cleaned up.

webpages/settings/style.css
DNin01: That was just a rough idea, and they don't have to take the form of blue floating labels.

mxmou: Maybe make it blue, like the Post button in Scratch Messaging? I think orange makes it look like something bad will happen when you click the button.

image

mxmou:
I think it would look even better if these two labels were aligned vertically:

image

The easiest way to do that is probably to use CSS Grid.

webpages/settings/style.css
DNin01: A blue background would look like this:

Blue enable button

I think it's a little bright but not too bad for a prompt you won't see that often.

DNin01:
Design update:

image

I was able to use margins to make the captions go to the bottom and newly added step numbers go to the top. There's also some space on either side and a gap in the middle. It looks more streamlined.

WorldLanguages:
I was going to suggest adding the numbers. Great idea.

Samq64:
When I approved I assumed design changes would be in a followup PR but doing it now is a good idea.

webpages/settings/style.css
mxmou: This button is probably the most important thing in the popup, so I wouldn't worry about it being too bright.

webpages/styles/components/modal.css
```diff
@@ -45,3 +45,65 @@
...
+  background-color: var(--blue);
+  color: #fff;
+  border-radius: 50%;
+  line-height: 2em;
+  width: 2em;
+  margin: 0 auto 1em auto;
+}
+@media not (max-width: 1000px) {
+  .permissions-steps p {
+    height: 2rem;
```
mxmou: I don't like this fixed height - the step descriptions might be shorter or longer depending on the language, and they might also change in future versions. It would be better to use CSS Grid for the steps layout.

Since this is a very minor issue, I'll approve anyway. I can open a new PR to improve the CSS after this is merged.

webpages/settings/index.js
```diff
@@ -11,7 +11,6 @@ import globalTheme from "../../libraries/common/global-theme.js";
 import { deserializeSettings, serializeSettings } from "./settings-utils.js";
 import { isFirefox } from "../../libraries/common/cs/detect-browser.js";
 
-const MORE_SETTINGS_HASH = "#moresettings";
```
mxmou, line 14: I'd keep this constant so that all the hashes are defined in one place.

mxmou (approved):

webpages/settings/style.css
DNin01: I could also make it dark blue and brighten when hovered:

https://github.com/user-attachments/assets/c9cc4ba8-a28e-4c53-927c-74324c22ec85

But that might take a little more work for light mode. Or I could make it gray just to add some depth:

https://github.com/user-attachments/assets/09f90d6c-3524-4c93-bbee-0932a57428cb

@mxmou, think I should go with blue, dark blue, or gray?

DNin01:
I gave the button a blue background, and thanks for giving me the opportunity to learn CSS grid lol.

I think this final design is pretty nice (and responsive):

https://github.com/user-attachments/assets/140fdec5-981a-4fd1-b0f5-dd34ad17e1a4

mxmou (approved):

Issue #9052, opened 2026-05-31
localhost webpages can get access to data from Scratch Addons


DNin01:
This obvious security shortcoming only affects developers, but with developer attacks becoming more common, it shouldn't be overlooked.

### Overview

Since developments addressing [Official way to test SA with local scratch-www instance](https://github.com/ScratchAddons/ScratchAddons/issues/5281), when you install the source code of Scratch Addons, it gets access to `http://localhost/*` by default. This is then used to run all of Scratch Addons' main functions on two specific localhost ports. The purpose of this is to allow developers to test addons on locally-hosted instances of [Scratch](https://github.com/scratchfoundation/scratch-www), such as pre-releases, before they are shipped to scratch.mit.edu.

The problem is that localhost **might not serve an instance of Scratch**—it could be a malicious page, or one with compromised dependencies.

In order to share data with addons and other scripts, the extension's content script hands out some data, **critically, including Scratch CSRF tokens and X-Tokens**, to the `window` object. Since it assumes that it is running on a trusted Scratch webpage, any malware that resides in it also gets access to this data. For this to happen realistically, a webpage that contains malware would have to be served to either `http://localhost:8333/` or `http://localhost:8601/` and opened in a web browser with Scratch Addons loaded unpacked and enabled. That's why it only affects developers.

### How we got here

Scratch Addons hasn't always run on local servers. As such, and because Scratch already has all user account data and is trusted not to tamper with extensions anyway, we didn't design our content script to be resistant to attacks at a webpage level. But, when we expanded our runtime scope to localhost, these issues weren't considered perhaps as much as they should have been. We didn't even warn developers or use opt-ins to control damage.

### Mitigation

An obvious fix is to make `http://localhost/*` an optional permission, so that access can be granted only to trusted environments. Developers could opt in when they need to test Scratch Addons together with Scratch on localhost, significantly limiting the attack surface.

I want to look for other possible improvements from a security-research standpoint, but permission management and documentation are alone an effective preventative measure.

### Smaller Issues

- The Comlink iframes also expose some exploitable functions, although they aren't that damaging. These include `getEnabledAddons()` and `copyImage()`. For all versions of Chrome and Firefox version 127 and up, `copyImage()` is unnecessary since the Web Clipboard API supports copying images (see #7625), so it might be worth removing entirely. The other functions of the Comlink iframes are more complex.
[...]

PR #9056, opened 2026-06-03
[Developer security] Don't send account data to untrusted tabs

### LINKED ISSUE #9054: Tokens should not be shared with localhost
Scratch Addons currently Scratch username, CSRF token, and X-Token into `window.scratchAddons`, even on localhost, which is not trustworthy like scratch.mit.edu is. See #9052 for more info.

DNin01:
Resolves #9054

### Changes

Previously, Scratch Addons could hand out your CSRF token and X-Token to tabs of sites outside scratch.mit.edu, including localhost, where the content script exposed them via the `window` object, potentially putting developers' Scratch accounts at risk.

Now, account info and credentials are replaced with a placeholder before the data is sent anywhere except scratch.mit.edu, preventing unauthorized webpages from seeing your account data.

This change applies to two background modules, `get-userscripts.js` and `global-state.js`.

The different data received by two tabs on different URLs is shown below:

globalState auth on scratch
globalState auth on localhost

The two objects would have been the same before this change.

### Reason for changes

These changes introduce a valuable security measure for developers. While Scratch accounts aren't as important as others, they can still be used for impersonation if compromised. They're still meaningful assets that deserve protection.

These fixes shouldn't affect the developer experience. Logged-out user status is already handled gracefully by the extension, so nothing will fail. Production account credentials were never meant to be used by local instances of `scratch-www` anyway.

See #9052 for more info.

### Security overview

I would rate this security issue a "moderate" due to the specific conditions an attacker would need to exploit it.

I would say a full security advisory isn't necessary for this vulnerability due to the limited impact (likely a few hundred developers who actively use recent versions of the source code) and low probability of exploitation.

For all Scratch Addons developers, I would recommend **disabling `http://localhost/*` in the Manage Extension menu** when you're not actively testing `scratch-www` locally. This prevents exploits like the one described above.

### Notes

Tested in Edge 148.

Changes inspired by a conversation with GitHub Copilot (Claude Haiku 4.5).

### DIFF
--- a/background/get-userscripts.js
+++ b/background/get-userscripts.js
@@ -174,7 +174,7 @@ async function getAddonData({ addonId, manifest, url }) {
   return { userscripts, userstyles, cssVariables: manifest.customCssVariables || [] };
 }
 
-async function getContentScriptInfo(url) {
+async function getContentScriptInfo(url, trust) {
   const data = {
     url,
     httpStatusCode: null, // Set by webRequest onResponseStarted listener
@@ -208,7 +208,20 @@ async function getContentScriptInfo(url) {
   await Promise.all(promises);
   data.globalState = scratchAddons.globalState._target;
 
-  return data;
+  if (trust) {
+    return data;
+  } else {
+    // Request came from an untrusted location, so remove sensitive account data
+    const defaultAuth = {
+      isLoggedIn: false,
+      username: null,
+      userId: null,
+      xToken: null,
+      csrfToken: null,
+      scratchLang: navigator.language,
+    };
+    return { ...data, globalState: { ...data.globalState, auth: defaultAuth } };
+  }
 }
 
 function createCsIdentity({ tabId, frameId, url }) {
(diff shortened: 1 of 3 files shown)


WorldLanguages:
> Production account credentials were never meant to be used by local instances of scratch-www anyway.

I'm not sure about this. Can anyone confirm? 

I thought the usual way of launching scratch-www was by using api.scratch.mit.edu which would require a real X-Token to work

The CSRF token is useless in scratch-www so no problem there

WorldLanguages:
Also, worth checking if the content script(s) ever actually read from globalState.XToken. I'm not sure about that, since it depends on how we implemented support for incognito and Firefox containers, so I don't remember right now. (remember that there's a single globalState but the user might be logged in into different accounts based on the specific tab due to incognito or FF containers) 

If that property is never read, no need for complicated fixes, we can just remove it from globalState

DNin01:
> Also, worth checking if the content script(s) ever actually read from globalState.XToken.

I also don't see why a content script wouldn't be capable of getting any token by itself, without a background context or secure message passing headaches. Could look into that in the future.

mxmou:
> I also don't see why a content script wouldn't be capable of getting any token by itself, without a background context or secure message passing headaches. Could look into that in the future.

The x-token could be read from the Redux state. In an addon userscript, the following would work: `addon.tab.redux.state.session.session.user.token`

animated-thumb already reads the CSRF token from `document.cookie` instead of using `addon.auth`: https://github.com/ScratchAddons/ScratchAddons/blob/f683626981976c71cc1fed0a077a36f43d57d41a/libraries/common/cs/thumb-setter.js#L60-L63

PR #9062, opened 2026-06-06
editor-unshare-button: update selector

### LINKED ISSUE #9061: "Unshare button in editor" addon not working at all.
### Description

The "Unshare" button does not appear when the `Unshare button in editor` addon is enabled. It just says `Shared`.

### Steps to reproduce

1. Open any project.
2. Enable the `Unshare button in editor` addon if it isn't yet.
3. Share the project if it isn't.
4. Look at the top bar. 

Image
As we can see here the addon is not working because it says `Shared` instead of `Unshare`.
Related to #9055 as we can se that error here too.

### Environment

**Browser:** Google Chrome 148
**Scratch Addons version:** 1.45.2
**Operating system:** Windows 10

mxmou:
Resolves #9061

### Changes

Fixes the "Unshare button in editor" addon.

### Tests

Tested on Edge.

### DIFF
--- a/addons/editor-unshare-button/userscript.js
+++ b/addons/editor-unshare-button/userscript.js
@@ -42,6 +42,7 @@ export default async function ({ addon, msg, console }) {
         });
 
         button.classList.remove("sa-unshare-button");
+        button.querySelector("span").innerText = addon.tab.scratchMessage("gui.menuBar.share");
         button.removeEventListener("click", thisFunction);
 
         redux.dispatch({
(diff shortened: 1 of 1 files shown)


Samq64:
Unsharing does not revert the label to "Share". Was that the case before the recent accessibility update?

mxmou:
> Unsharing does not revert the label to "Share". Was that the case before the recent accessibility update?

I don't know - I've never used this addon. But I can try to fix that.

Samq64 (approved):

PR #9070, opened 2026-06-11
Convert enable confirmation to web dialog

### LINKED ISSUE #1861: Give in-the-face warnings before turning on an addon
Alright, hear me out.

We got quite some people who didn't read the warnings on some addons, which causes unintended behavior on their side.

What if we have some kind of modal/window/notification that is shown if there are some kind of warning/notice, so we can make sure that they aware of the implications. The notices needs one click to be seen, so an in-the-face modal would help people notice the warnings.

DNin01:
Resolves #1861

### Changes

The confirmation alert for enabling an addon that is marked as dangerous has been converted to a `<dialog>` with more details about the addon being enabled.

https://github.com/user-attachments/assets/e9cd1bc4-5a8b-4d95-997e-262d3e6dd2f2

### Reason for changes

Web dialogs don't block the browser window or the main thread and can have styled content.

### Technical overview of added features

- Dialog before enabling dangerous addons
  - The addon's description is displayed in the new dialog, as well as its warnings and other information.
  - The addon to confirm enabling is set with the `addonToEnable` property, which is already used by a different dialog.
- Small dialogs
  - `modal`s are small by default but can be made full-size using the `modal-full-width` and `modal-full-height` classes. The small size is perfect for alerts like the enable confirmation.
  - justify-content activates center alignment if the modal is taller than the content (for small, portrait screens).

### Tests

Tested in Edge 149.

#### Known Issues

- The addon tags in the confirmation dialog have tooltips but extend above the visible area and are unreadable. This could be fixed in a future PR by changing the alignment of these tooltips.
- If you attempt to enable an addon with optional permissions in the Addons popup, which opens the half-modal "enablepopup", and then a dangerous addon to also open the dangerous addon modal, "enablepopup" will update to refer to the dangerous addon. This is unintentional and is caused by the property `addonToEnable` being shared between both of these dialogs, but it is not a realistic problem. After this PR, it could easily be converted to a modal.


DNin01:
The dialog was updated. It now looks like this:

Enable this addon? dialog

webpages/settings/index.html
```diff
@@ -435,7 +437,56 @@ <h1 v-cloak>{{ msg("settings") }}</h1>
...
       </div>
     </modal>
-    <modal v-cloak v-el:permissions>
+    <modal :title="msg('enableConfirmation')" v-cloak v-el:enableaddon>
+      <div class="addon-body" v-if="addonToEnable">
+        <div class="addon-topbar">
+          <div class="clickable-area">
+            <button class="arrow-button" disabled>
+              <img src="../../images/icons/expand.svg" class="reverted" draggable="false" />
+            </button>
```
mxmou: Since the addon can't be collapsed anyway, I think it's better to remove this button entirely.

webpages/settings/index.js
```diff
@@ -199,6 +199,30 @@ let fuse;
     },
 
     methods: {
+      /* Also in addon-body */
+      addonIconSrc(addon) {
```
mxmou: Could these methods be moved into a module instead of duplicating them?

DNin01: I have replaced these methods entirely by adding new `_iconSrc` and `_iconAlt` properties to addon objects.

webpages/settings/components/modal.html
mxmou, line 75: This `height: 100%` should probably only apply to `.modal-full-height`. That way, the modal would look like this in small windows and in the popup:

Image<br />

And not like this:

Image

mxmou (approved):

PR #9075, opened 2026-06-17
swap-local-global: allow conversion between local and global across multiple sprites

griffpatch:
### Overview

The **swap-local-global** addon now allows converting a variable between local and global even when it is used by multiple sprites. Previously this was blocked with an error message.

### Changes

**Global → Local:** When converting a global variable to local on the current sprite, any *other* sprite that also uses that variable now automatically gets its own independent local copy with the same name and current value. All of those sprites' block references are remapped to their new local ID so nothing breaks.

**Local → Global:** When converting a local variable to global, any other sprite that happens to have a same-named local variable is automatically merged into the new global — its blocks are remapped to point at the global ID and its now-redundant local is deleted.

A small shared helper `remapVariableReferencesInTarget` was added to walk a target's block fields and update any field whose ID matches the old variable, used by both conversion paths.

### Reason for changes

The previous behaviour blocked very common real-world cases. A game's `score` or `lives` variable is often shared between a player sprite and the stage. A student clicking "make local" on such a variable was met with an unhelpful error rather than a useful outcome. The new behaviour does what users intuitively expect: each sprite gets its own copy (or they all share the one global), with no manual cleanup needed.

### Tests

Tested in Chrome:
- Global → local on a variable used by 3 sprites: each sprite received an independent local copy; all scripts continued to work correctly
- Local → global where 2 other sprites had a same-named local: both were merged into the global; block references updated correctly in all sprites
- Single-sprite conversions (the original behaviour) are unchanged

### DIFF
--- a/addons/swap-local-global/userscript.js
+++ b/addons/swap-local-global/userscript.js
@@ -66,6 +66,18 @@ export default async function ({ addon, msg, console }) {
     }
   };
 
+  // Rewrites all block field references from oldId to newId in a target's blocks.
+  const remapVariableReferencesInTarget = (target, oldId, newId, newName) => {
+    for (const block of Object.values(target.blocks._blocks)) {
+      for (const field of Object.values(block.fields)) {
+        if (field.id === oldId) {
+          field.id = newId;
+          field.value = newName;
+        }
+      }
+    }
+  };
+
   const syncBlockVariableNameWithActualVariableName = (workspace, id) => {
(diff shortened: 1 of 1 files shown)


apple502j (commented):
Question, should the code change `sensing_of` (for local->global?)

griffpatch:
Yeah, I wondered about that... I am not sure, maybe even if it should have
a warning when that is the case because it can be a breaking change as in
the other direction it will not know which sprite you were referencing

PR #9080, opened 2026-06-19
Revise warning message in disable-auto-save addon

DNin01:
For a while, "Disable auto-save" has displayed this justifiable warning message:

> This will disable all types of project autosaving. Make sure to manually save your changes frequently to avoid losing them.

This proposes a revised warning message for the addon:

> This will prevent projects from saving automatically; manually save your changes whenever possible to avoid data loss.

This revised message was inspired by potential guilt felt from ignoring the original warning for real reasons. In Scratch, autosaves usually happen at 10-minute intervals, so disabling the feature would imply that you want to edit without saving for longer. While that's already not ideal for avoiding data loss, saving changes directly to a shared project does publish them immediately (which is one of the main reasons people have asked for the option to disable autosaving). Also, autosaves overwrite anything you've modified or deleted, so if you want to revert them, you'll need to have a backup.

The original warning message may be stronger, but the cost is that users could feel criticized for avoiding saving very often, despite Scratch Addons still providing the option and being all about customizing to suit your needs. "Make sure to manually save your changes frequently" is just a little overly commanding for a recommendation, considering the point of the addon is to not be forced to have such frequent saves.

The new warning message points out the risks while trusting users to make informed decisions. It was written with an understanding that users know that saving is important but could have reasons not to save at an ideal frequency. Some of its phrase changes include:
- "Frequently" has been changed to "whenever possible," which is meant to be interpreted based on what I said earlier. That might be, "when I've finished drawing and am ready to release my changes."
- "All types" has been omitted because Scratch only auto-saves on a single timer currently.

Both messages have advantages. In my opinion, the original message might be enough of a helpful push for someone to not lose work the hard way while the revised message is less likely to be off-putting. Ultimately, I think both wordings do a good job at concisely indicating the risks, but I think the updated warning message is a better fit for our extension.

What do you think, and why do you prefer having or not having autosaves in Scratch (or any other program)?

### DIFF
--- a/addons/disable-auto-save/addon.json
+++ b/addons/disable-auto-save/addon.json
@@ -4,7 +4,7 @@
   "info": [
     {
       "type": "warning",
-      "text": "This will disable all types of project autosaving. Make sure to manually save your changes frequently to avoid losing them.",
+      "text": "This will prevent projects from saving automatically; manually save your changes whenever possible to avoid data loss.",
       "id": "manuallysave"
     }
   ],
--- a/addons/disable-auto-save/addon.json
+++ b/addons/disable-auto-save/addon.json
@@ -4,7 +4,7 @@
   "info": [
     {
       "type": "warning",
-      "text": "This will prevent projects from saving automatically; manually save your changes whenever possible to avoid data loss.",
+      "text": "Manually save your changes whenever possible to prevent data loss.",
       "id": "manuallysave"
     }
   ],


DNin01:
### Copilot says...

> **Strengths of your revision:**
> 1. **More actionable** — "whenever possible" is more realistic and empowering than "frequently," which can feel vague or guilt-inducing. It acknowledges that users have legitimate reasons for disabling autosave and respects their agency.
> 2. **Clearer consequence** — "avoid data loss" is more direct and specific than "avoid losing them," making the risk feel more concrete.

<details><summary>Full conversation with Claude Haiku 4.5 (AI model):</summary>

```
# Copilot Chat Conversation Export: Revised warning message analysis

**User:** @DNin01
**Thread URL:** https://github.com/copilot/c/7ad2bd36-52f0-448b-be87-2dbb1acea71a

## @DNin01

This file is for an addon that contains a warning message, "This will disable all types of project autosaving. Make sure to manually save your changes frequently to avoid losing them."
Now, analyze this revised warning message I wrote: "This will disable all types of project autosaving. Manually save your changes whenever possible to avoid data loss."

I should mention that autosaves usually happen at 10-minute intervals, so disabling the autosave feature would imply that you would want to edit a project without saving it for over 10 minutes. While this is already not ideal for protecting your data from loss, saving changes directly to a shared project does publish them immediately, a common reason people have asked for the option to disable autosaving. The new wording accounts for the fact that a user might not be ready to save their changes at an ideal frequency, but "whenever possible" is still an encouragement to do so with a different connotation.

Do you think this is an improvement or is the previous wording still better?

## Copilot

I think your revised message is a solid improvement for these reasons:

**Strengths of your revision:**
1. **More actionable** — "whenever possible" is more realistic and empowering than "frequently," which can feel vague or guilt-inducing. It acknowledges that users have legitimate reasons for disabling autosave and respects their agency.
2. **Clearer consequence** — "avoid data loss" is more direct and specific than "avoid losing them," making the risk feel more concrete.
3. **Shorter and punchier** — It removes redundancy ("all types of project autosaving" → "autosaving") and gets to the point faster, which is better for a warning that needs immediate comprehension.

**Where the original has a slight edge:**
- The original's emphasis on *frequency* ("frequently") + *manual action* ("manually save") creates a two-part behavioral prompt that some users might find more helpful. Your version assumes users already know they should be proactive.

**My recommendation:**
Use your revision. The original warning reads like it's scolding users for their choice; yours reads like it's giving them practical guidance while trusting their judgment. That tone shift is appropriate for an addon that users have explicitly opted into, especially when the use case (immediate publishing to shared projects) is legitimate.

If you wanted to split the difference, you could try: *"This will disable autosaving. Manually save your changes whenever possible to prevent data loss."* But honestly, yours is clearer as-is.
```

</details>

<sub>AI was not used to write the pull request description or the revised message—only for inspiration.</sub>

mxmou:
I suggest removing the first sentence entirely, since it essentially repeats the addon description. If the warning is shorter, users will be more likely to read it.

DNin01:
Done. It's very concise.

Disable auto-save addon with shorter info

mxmou (approved):

WorldLanguages (approved):

PR #9088, opened 2026-06-27
Split sidebar content into two flexboxes

DNin01:
As shown in the side-by-side comparison image below, the flex container consisting of the category has been split in two. Padding and margin overrides that were previously set on individual items have been moved or changed to row gaps.

Sidebar layout comparison

The visual layout and the design of the sidebar remain the same*—it's just simpler for developers to modify, which will make accessibility features and layout changes easier to implement in the future. For example, the 12px space between addon categories and external links is no longer the margin of any particular addon category, so the [`marginBottom` flag](https://github.com/ScratchAddons/ScratchAddons/blob/ee6157c11597db7c732e120841becf5cce96eed4/webpages/settings/data/categories.js#L91) is no longer needed. This would've helped with changes like:
- #8924
- #9008

<sub>*The 14-px space below the More Settings button in small mode have been transferred to the parent element and increased to 20px to match the padding on the top of the sidebar.</sub>

Tested on Edge 149.

### DIFF
--- a/webpages/settings/components/category-selector.html
+++ b/webpages/settings/components/category-selector.html
@@ -8,7 +8,6 @@
     v-if="!category.hidden"
     v-show="shouldShow"
     transition="expand"
-    :style="{ marginBottom: category.marginBottom ? '12px' : 0 }"
     @click="onClick($event)"
   >
     <img :src="'../../images/icons/' + category.icon + '.svg'" draggable="false" />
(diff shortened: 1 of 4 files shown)


WorldLanguages:
Good idea!

DNin01:
I'm going to merge this without a review because it does not seem to have any negative effects and will make accessibility updates easier.

Issue #6311, opened 2023-06-29
dark-www: custom banner colour


Secret-chest:
Limiting users to 3 choices makes no sense, especially since you can't have the exact same colour.

mxmou:
Is this specifically about purple banners? I think #5971 is a better solution for those.

If you want to be able to customize all banners (there are also green, blue and pink ones), that should probably be a separate addon.

DNin01:
There are only three choices because there are multiple different banner colors, and I would assume making all of them fully customizable could become overwhelming.

WorldLanguages:
What three choices are we talking about?

DNin01:
> What three choices are we talking about?

Probably these:
https://github.com/ScratchAddons/ScratchAddons/blob/4151aa83859acf91fc87b91f5faff9dc7e078113/addons/dark-www/addon.json#L2128-L2141

Secret-chest:
> There are only three choices because there are multiple different banner colors, and I would assume making all of them fully customizable could become overwhelming.

No .. they only affect the purple ones

mxmou:
> > There are only three choices because there are multiple different banner colors, and I would assume making all of them fully customizable could become overwhelming.
> 
> No .. they only affect the purple ones

They affect all banners.
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/255dfbb8-1a3d-4d47-a6c9-4cb9fa06498d)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/618c9c6b-fe94-473d-9413-9376ac8c93e5)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/a483bb77-4da1-4359-b82a-7d55037497bb)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/6f426852-ce5e-42f2-9d18-131158757fa1)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/911db04e-7dba-49ac-9686-b9aa7236652f)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/aa6379ad-0efa-4807-a790-c8c1b6b490e2)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/eb7508c0-45a5-439b-9fc6-4166c18399d2)
![image](https://github.com/ScratchAddons/ScratchAddons/assets/51849865/b9621e1a-0b8f-4197-b91d-d43155929bea)
[...]

Issue #5971, opened 2023-04-29
Option to make the banner colour exactly the same as the navbar colour (dark-www)


Secret-chest:
### Why this would be helpful

_fancy issue forms_

### My suggestion

_fancy issue forms_

### Possible alternatives

We could also remove the _fancy issue forms_ instead.

### Additional context

_fancy issue forms_

WorldLanguages:
The issue forms are fancy but I don't know what banner you're referring to.

Secret-chest:
The banners like the ones for My Stuff, Discuss Scratch etc.

WorldLanguages:
I think this was discussed somewhere but I can't find it right now.

Btw, keep in mind that My Stuff only has a banner if "Scratch 2.0→3.0" is enabled.

Secret-chest:
Yes, but the one in the Explore page is always orange. I'd also like to change it, but that's a separate issue.

Issue #3938, opened 2021-12-17
Design bug?


DogCatPuppyLover:
### Description

Recently, the design seems to have updated for worse, and it is not consistent across the website.

- On some pages, the scrollbar is a dark blue.
- On the Messages and My Stuff page, the banner is a different dark blue.
- The reply button on project and studio comments is a different color than the "reply" text.
- The "Remix tree" button is misplaced on project pages.
- Some other aspects of the studio design seem to have changed, as well, such as the line around "See more replies."

### Screenshots

Messages page:
![](https://user-images.githubusercontent.com/79656307/146585558-b8b62a2c-f8b4-4812-85cc-0878df799a69.png)

Reply button:
![](https://user-images.githubusercontent.com/79656307/146587852-3c52ea01-70c5-405f-b31e-f6a7e7c6137c.png)

Remix tree button and borders:
![](https://user-images.githubusercontent.com/79656307/146588100-27fe91c9-d0f4-47d4-b5fe-9e856316dfd9.png)

### Environment

- Browser: Firefox 95.0.1
- Scratch Addons version: 1.23.0
- Operating system: Windows 10

RedGuy12:
Also

***@***.***

***@***.***

CST1229:
Also, `forum-quote-code-beautifier` doesn't color properly anymore when `scratchr2` is also enabled (the border is always gray and the background is the same gray), which might be because of their order in `addons.json`.

DogCatPuppyLover:
> Also https://user-images.githubusercontent.com/61329810/146589112-5adfb6ab-1339-4f1d-91fe-16896c3bcda1.png https://user-images.githubusercontent.com/61329810/146589136-804f107a-e80f-4f1e-9c8d-0a019282278b.png

It looks like I can fix the color problems by setting that to "Default."

DogCatPuppyLover:
The "Cancel" button on studio comments also seems to have become darker.

CST1229:
> The "Cancel" button on studio comments also seems to have become darker.

Might be because of this option:
![](https://user-images.githubusercontent.com/68464103/146590084-919f8168-d6fb-49b5-a680-3026d2682a9d.png)

Samq64:
> On some pages, the scrollbar is a dark blue.
[...]

PR #9096, opened 2026-06-29
New color picker for My stuff and Messages tab banners in dark-www. Only works with scratchr2.

skima-a:
Self explanatory. Changes the addon file for dark-www, adding a new color picker that lets you change the banner color for the "My stuff" tab, the "Messages" tab, and the main scratch page (pre login). These banners (with the exception of the main scratch page) only appear with the scratchr2 extension. 

https://github.com/user-attachments/assets/970c8348-bce7-454c-94a4-15e6a53eb60a

**### Reason for changes:**

As of now, the mentioned banners can only be less saturated versions of the navbar color. For some navbar colors this looks fine, but with colors like orange it can look ugly. This new color picker allows the user to choose what color they want instead of only having 3 options.
      This doesn't have to be in dark-www and can be moved to scratchr2 if requested.

This change was tested and low and behold, the 15 lines of code I changed didn't break anything. This change also changes the color of the "Next" button on the sign up page.


Samq64:
Related:
- #3938
- #5971
- #6311

PR #9098, opened 2026-06-30
Compact remix list

### LINKED ISSUE #9097: Redesign remix credits list on projects
### Why this would be helpful

The section crediting up to two projects with serving as bases of a remix can take too much space. If you are already familiar with how remixing works, an icon sized note (with no or only limited amount of text) would be enough for letting you know that the project is not an original idea, only a remix.
And there you go, you'll have a little bit of more place for the instructions section for the project, which I find much more important.

### My suggestion

No1: list the previous versions of the project in one line (clicking on them would lead you to the project itself).
Image
No2: if possible, complete the list of all the projects in the remixing history

### Possible alternatives

Button for previous version may also open the profile of the project's creator, but I think, that makes less sense.
Second idea can be dropped.

TBH, I just got upset of the remix badges taking up too much space, so I decided to make fast a userscript, but if I wrote it, why not be an addon?

### Additional context

Related Discord suggestions and discussion from long ago: 
* https://discord.com/channels/806602307750985799/1178377891445215252
* https://discord.com/channels/806602307750985799/1178218611144409138

IDK about the Copyright issue, perhaps it's required to express credits in text too, then the addon can't be added 😿

OregSamSas:
Resolves #9097

### Changes

Image

- [x] Make remix history list more compact
- [ ] Complete remixing history

### Reason for changes

To obtain more space for the Instructions + Notes & Credits sections.

### Tests

Platforms:
- [x] Chromium based (Brave)
- [x] FireFox

Features:
- [x] having effect in correct time (after loaded needed stuff)
- [x] clicking on the buttons (if addon is enabled) brings up the correct pages
- [x] disabling works properly
- [x] reenabling works properly


OregSamSas:
If features are too little for making it to an addon, it can perhaps be expanded by developing it to a "compact website" (not editor) mode (as if Sratch Tools had a similar addon, maybe their code, or ideas can be used for that).
There could also fit a feature that collapses, or at least reduces the size of the Instructions or the Notes & Credits tabs, when empty.

PR #9111, opened 2026-07-16
Added addon cloud-to-turbowarp.

### LINKED ISSUE #8771: Alternate Cloud Servers
### Why this would be helpful

The cloud variables have been down for ages and i really want to make a cloud game, and there are many people out there who want to make and play them too. It's really annoying to make cloud games right now, because you have to package them every time you want to test. (Currently my only known way of having alternate cloud servers is that.) Is there any chance that a new addon called "Use Turbowarp Cloud Servers" or something like that could be created so you could use the turbowarp cloud servers if the scratch ones are down for development and development only? Thanks ^_^

### How this addon works

Basically, you enable it, and if you are editing a project of yours, the cloud variables hook up to turbowarp's servers.

### Possible alternatives

_No response_

### Additional context

_No response_

ash-luigi:
This PR adds a addon called cloud-to-turbowarp.
When enabled, online multiplayer projects will link to turbowarp servers instead!

### Reason for changes

This addon allows people who prefer turbowarp's fast, generally more reliable servers to use in the scratch GUI! Also adds banner to games tab when enabled as servers still are scratch on games tab since turbowarp does not store logs.

### Tests

Tested on Edge (based off chromium and edge extension support chrome so will also work on chrome)
Firefox didn't work...but that's a ScratchAddon or testing bug because the addon didn't load at all (it said it did but the background process died after a bit and nothing popped changed).

### DIFF
--- a/addons-l10n/en/cloud-games.json
+++ b/addons-l10n/en/cloud-games.json
@@ -14,5 +14,6 @@
   "cloud-games/add-studio": "Add this studio",
   "cloud-games/add-studio-desc": "Click to add the studio from the current tab to the list.",
   "cloud-games/added": "Added!",
-  "cloud-games/fetch-error": "Could not load cloud data for this project:"
+  "cloud-games/fetch-error": "Could not load cloud data for this project:",
+  "cloud-games/turbo-warp-banner": "Games data is from Scratch servers (Cloud variables → TurboWarp is active)"
 }
--- a/addons/addons.json
+++ b/addons/addons.json
@@ -17,6 +17,7 @@
   "search-profile",
   "better-emojis",
   "cloud-games",
+  "cloud-to-turbowarp",
   "studio-tools",
   "last-edit-tooltip",
   "semicolon",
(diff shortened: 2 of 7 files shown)


pufferfish101007:
Some thoughts:
- I'm not sure it makes sense for this addon to be dynamically enabl-/disablable, seeing as presumably only one websocket connection will ever be initiatied per page load? i.e. you'd need to reload the page to connect to the new server anyway (please correct me if I'm wrong)
- If you would want it to be properly dynamic, it would likely be better to override `WebSocket` methods rather than the constructor, and test for `addon.self.disabled` inside of those methods, rather than using event listeners. You'd probably need to check if the enabled-ness had changed since the last method call, and reinitiate the connection if needed
- Have you considered the comments given in https://github.com/ScratchAddons/ScratchAddons/issues/8771#issuecomment-3751183046 and the comment below that?

Looks like a reasonable implementation though at a glance. I haven't tested it.

PR #9118, opened 2026-07-22
A new addon for block comment alignment

DNin01:
### Better Editor Comments addon removal

All of the addon patches in `fix-editor-comments` were fixed in the current version of Scratch (hooray! 🎉) so those settings are no longer necessary. With only the "Keep connections straight" feature left, I created a dedicated addon, "Horizontally aligned comments," to take its place.

### The Shift key update

I also updated how the alignment feature works. The alignment feature is nice if you like comments to be directly to the right of the block they're attached to. But what if you only want to do this for particular comments?

So, I made it require holding a key by default, and it now aligns the comment as you drag it, not just when you let go. This was inspired by snapping features in image/graphic editors.

https://github.com/user-attachments/assets/5a848a7c-4fd9-4f90-b936-8526fc30fb61

For users who had the "Keep connections straight" setting enabled, this addon and its "Flip controls" setting will be enabled automatically to preserve user preferences. Otherwise, both must be enabled by the user.

### "New" tag behavior

This addon has both a `versionAdded` and `latestUpdate` in order to display an update-styled notice while still being displayed as new. Normally, this would cause a "new features" or "new settings" tag to appear next to the "new" tag, but the settings page code has been modified so that doesn't happen.

---

I tested these changes and the setting migration works.


addons/editor-straight-comments/addon.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "Horizontally aligned comments",
+  "description": "While moving a comment attached to a block, hold the Shift key to align the comment so that its connection is perpendicular to the block.",
```
mxmou: "Perpendicular to the block" sounds weird because a block isn't a line - it should probably be "perpendicular to the edge of the block". Or just "horizontal":

```suggestion
  "description": "While moving a comment attached to a block, hold the Shift key to align the comment so that the line connecting it to the block is horizontal.",
```

addons/editor-straight-comments/addon.json
```diff
@@ -0,0 +1,44 @@
...
+  "tags": ["editor", "codeEditor", "featured"],
+  "info": [
+    {
+      "id": "fix-comments-update",
+      "type": "notice",
+      "text": "If you had the \"Keep connections straight\" setting enabled in the former addon \"Better editor comments\", this addon will be enabled automatically. For those who had, you now have the option to only straighten connections while holding Shift. If you prefer that, just disable the setting below."
```
mxmou: This doesn't seem important enough for a notice to me. cleanup-plus, which was also moved from another addon, has a regular info message.

I also don't think that the new setting needs to be described here - most people who used the old addon will probably want to keep the old behavior anyway. Instead, I suggest adding a `latestUpdate` so that the setting gets a "New" badge.

DNin01: It doesn't feel correct for `editor-straight-comments` to be marked as an update to an existing addon because it is the result of separating a secondary feature from a different addon, even though that addon is being removed. For comparison, `editor-cleanup-plus` has `"versionAdded": "1.43.0"` and its setting isn't labeled "New".

I could set `versionAdded` and `latestUpdate.version` to the same value for this addon, but that looks like a paradox:

Horizontally aligned comments, New! New settings

If an update banner works best, the settings page's logic could be changed to only allow the addon name to be labeled with one "New" badge.

What should I do?

mxmou: > I could set `versionAdded` and `latestUpdate.version` to the same value for this addon, but that looks like a paradox:

I don't think that's necessary wrong - the addon is new and the setting is also new.

DNin01:
Addon details updated.

mxmou:
While testing, I noticed that there's currently a bug that makes block comments revert to their original position after switching sprites. That makes an addon like this not very useful - maybe we should wait until it's fixed?

DNin01:
> While testing, I noticed that there's currently a bug that makes block comments revert to their original position after switching sprites. That makes an addon like this not very useful - maybe we should wait until it's fixed?

How do you trigger that bug? What project(s) is it happening in? I'm not spotting any problems myself.

It _might_ have to do with the fact that the new addon's userscript overrides `onDrag` but not `onDragEnd`.

<sub>Edge 151</sub>

mxmou:
> It _might_ have to do with the fact that the new addon's userscript overrides `onDrag` but not `onDragEnd`.

I thought it was a vanilla bug because I could reproduce it without your addon enabled, but disabling SA fixes it. I'll try to find the addon that is causing it.

Edit: see #9151.

addons/editor-straight-comments/addon.json
```diff
@@ -0,0 +1,42 @@
...
+      "default": false
+    }
+  ],
+  "dynamicEnable": true,
+  "dynamicDisable": true,
+  "versionAdded": "1.46.0",
+  "latestUpdate": {
+    "version": "1.46.0",
+    "newSettings": ["invert"],
+    "temporaryNotice": "This feature was separated from the deprecated \"Better editor comments\" addon and updated so you can use the Shift key to control comment alignment."
```
mxmou:
```suggestion
    "temporaryNotice": "This feature was separated from the removed \"Better editor comments\" addon and updated so you can use the Shift key to control comment alignment."
```

background/get-addon-settings.js
```diff
@@ -188,6 +188,18 @@ chrome.storage.sync.get([...ADDON_SETTINGS_KEYS, "addonsEnabled"], (storageItems
...
+    if (
+      addonsEnabled["editor-straight-comments"] === undefined &&
+      addonsEnabled["fix-editor-comments"] &&
+      addonSettings["fix-editor-comments"].straighten === true
+    ) {
+      // fix-editor-comments replaced with editor-straight-comments in v1.46
+      madeAnyChanges = true;
+      addonsEnabled["editor-straight-comments"] = true;
+      // Invert the activation so that the behavior is consistent with the old addon as per user settings
+      addonSettings["editor-straight-comments"].invert = true;
```
mxmou: This throws an error for me:
```
Uncaught TypeError: Cannot set properties of undefined (setting 'invert')
```

I used `chrome.storage.sync.clear()` to reset all settings before testing the migration.

(Somewhat off-topic: we really should get #7632 merged to make writing migration code less complicated.)

webpages/settings/index.js
```diff
@@ -551,6 +551,8 @@ let fuse;
       manifest._addonId = addonId;
       manifest._groups = [];
 
+      // In production, all addons should have versionAdded.
+      // However, it shouldn't have to be set in stone during development.
```
mxmou: I don't agree with this comment. This particular bit of code checks if versionAdded exists, but addon authors shouldn't rely on that - the property is marked as required in the schema and the docs.

mxmou (changes requested):

background/get-addon-settings.js
DNin01: I thought that was guarded by the condition, but it looks like `addonSettings["editor-straight-comments"]` can still be a null value. Thanks.

webpages/settings/index.js
DNin01: I've removed the comment.

mxmou (approved):

Issue #8382, opened 2025-06-09
Optimization: Locale packing


DNin01:
This proposal is a solution to #7964.

In the current locale system, all English locales for addons are created manually and written to `addons-l10n/en/<addonid>.json`. Addon locales for other languages are generated automatically by CI and written to `addons-l10n/<language>/<addonid>.json`. This structure means every addon that contains localized strings has a separate locale file, which increases the complexity of Scratch Addons and probably makes it take longer to download.

Each addon-specific string key is also prefixed with the addon's ID, creating repetition that inflates the file size.

> #### 📂 `addons-l10n` / 📂 `en` / 📄 `<addonid>.json`
> ```json
> {
>   "<addonid>/confirmation": "Are you sure?",
>   "<addonid>/yes": "Yes",
>   "<addonid>/no": "No"
> }
> ```

---

Here's a better way: instead of writing locales for addons in an irrelative directory, English locales would be saved in `addons/<addonid>` and look like this:

> #### 📂 `addons` / 📂 `<addonid>` / 📄 `locale.json`
> ```json
> {
>   "confirmation": "Are you sure?",
>   "yes": "Yes",
>   "no": "No"
> }
> ```

But when building a release of Scratch Addons, we would combine the contents of all these locale files into one, write it to a single file, and then remove the individual `<addonid>/locale.json` files from the build. The result...

> #### 📂 `addons-l10n` / 📄 `en.json`
> ```json
> {
>   "addon1": {},
>   "addon2": {},
>   "addon3": {},
>   "...": {}
> }
> ```

...one locale file per language, consisting of each string for each addon. It's a cleaner, more manageable locale structure.

Using a build script for this would be ideal but it wouldn't be strictly necessary — if we don't use one, Scratch Addons would just need to support loading strings from both source data (during development) and compiled data (for users).

---

I invite @Hans5958 and @WorldLanguages to this conversation.

WorldLanguages:
Do we have an estimate for the size of, let's say, all Spanish locales merged together on a single file?

Samq64:
For RAM of disk usage?

The size of the first 5 addons alphabetically is 3.48KB without file metadata. As a single file without the addon prefixes it's 3.03KB (about a 15% improvement)

~~So using those numbers the full Spanish locale file would be roughly 555KB.~~

My sample was bad because it included 2 addons with a lot of strings. It would likely be 95-105KB considering the `es` folder has 113KB of json.

[...]

PR #9120, opened 2026-07-24
Split off debugger category string

Samq64:
In preparation for #8382

Splits the debugger category label into its own `_general` string so the addon API (and userscripts) no longer rely on metadata strings. As far as I know that was the only case where it did.

### DIFF
--- a/addons-l10n/en/_general.json
+++ b/addons-l10n/en/_general.json
@@ -5,6 +5,7 @@
   "_general/blocks/green-flag": "flag",
   "_general/blocks/clockwise": "clockwise",
   "_general/blocks/anticlockwise": "anti-clockwise",
+  "_general/blocks/sa-category": "Debugger",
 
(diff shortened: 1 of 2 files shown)


addons-l10n/en/_general.json
```diff
@@ -5,6 +5,7 @@
   "_general/blocks/green-flag": "flag",
   "_general/blocks/clockwise": "clockwise",
   "_general/blocks/anticlockwise": "anti-clockwise",
+  "_general/blocks/sa-category": "Debugger",
```
mxmou, line 8: Do you think a developer comment would be helpful here (to explain to translators how this string is different from the addon name)?

Samq64: Sure, maybe renaming the key to `debugger-category` would also help?

DNin01: Or `debugger-addon-category`.

PR #9127, opened 2026-07-28
Bump deps since nobody has bothered to do it

apple502j:
Bumps GHA deps and `text-field-edit`. Other deps untouched for now.

### DIFF
--- a/addons/workspace-dots/userscript.js
+++ b/addons/workspace-dots/userscript.js
@@ -32,7 +32,7 @@ export default async function ({ addon, console }) {
       const CROSSHAIR_LENGTH = this.getSpacing() / 2.5;
       const NO_LENGTH = 0;
 
-      // We are mulitplying line1's (vertical line) and line2's (horizontal line)'s length based on the user's setting.
+      // We are mulitiplying line1's (vertical line) and line2's (horizontal line)'s length based on the user's setting.
       // Scratch calls it "width", so don't get confused.
       width *= (() => {
         switch (addon.settings.get("theme")) {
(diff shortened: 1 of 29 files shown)


Hans5958:
ESLint was locked to 9.x on #8845 because 10.x has rule changes. Uness anyone wants to fix the errors while the development is slow, or there are some critical security issues, I think this can be deferred.

Everything else is good.

cc @DNin01, creator of #8845

DNin01:
> ESLint was locked to 9.x on #8845 because 10.x has rule changes. Uness anyone wants to fix the errors while the development is slow, or there are some critical security issues, I think this can be deferred.

~I will open an issue so we remember.~

Edit: I didn't realize that this PR unlocks the ESLint version.

addons/workspace-dots/userscript.js
```diff
@@ -32,7 +32,7 @@ export default async function ({ addon, console }) {
       const CROSSHAIR_LENGTH = this.getSpacing() / 2.5;
       const NO_LENGTH = 0;
 
-      // We are mulitplying line1's (vertical line) and line2's (horizontal line)'s length based on the user's setting.
+      // We are mulitiplying line1's (vertical line) and line2's (horizontal line)'s length based on the user's setting.
```
DNin01: Still has a typo.
```suggestion
      // We are multiplying line1's (vertical line) and line2's (horizontal line)'s length based on the user's setting.
```

apple502j:
@Hans5958 what are the actual errors encountered on ESLint? this PR is obviously passing checks and no error is logged when manually running, either.

Samq64:
I am not sure why 9.x was chosen specifically because #8836 merged the changes needed to pass ESLint 10.

PR #9134, opened 2026-07-31
Multiitem Managment for Folders

### LINKED ISSUE #9133: Multiitem Managment for Folders
### Why this would be helpful

THIS BRINGS US indirectly A LOT CLOSER TO MULTISELECTING SPRITES & ASSETS!

Consider this example: when you have 100+ costumes for a sprite (e. g. Text engines), and you realise that half of the costumes at the end of the list are useless (e.g. random special characters, or letters with diacritic marks, and you only need basic English), or they fall into the same category, and it'd good to have them in the same folder, currently you have to go through them one by one and apply the desired action. 😠 

If only you could be able to achieve the same result with a few clicks... 😮‍💨

### My suggestion

With some refinement for the current context menu actions in the **folders addon**, THIS COULD BE POSSIBLE! 😲

* add a way to add to or remove from an existing folder starting from a certain item (backward or forward) ➕/➖  ⏫ ⏬ 

  * **with this** the useless rest of costumes can be **easily deleted** just in **4 steps**:
  (requested for example here: [Discord Ref 1](https://discord.com/channels/806602307750985799/1055783003156127765))
  1. add the last costume to a new folder
  2. select the one before the first useless
  3. add to folder all the following ones
  4. delete folder

* add a way to delete all the items outside of folders ⛔📁 ➡️ 🗑️
 
  * with this feature, you can easily select a few needed costume out of many

### Possible alternatives

I have currently in my PR a "preceeding" or "following" implementation. Maybe we should have "add to ... *starting from* this on" instead of "*following* this"

### Additional context

Related issues&requests&discussions:
* #7369
* [Discord Ref 2](https://discord.com/channels/806602307750985799/1057604742160863252)
* [Discord Ref 3](https://discord.com/channels/806602307750985799/1198337126937403502)
* #5183

This feature would also smoothly complement the addon `delete-others` #4692, since it also provides a way to delete only a part of the other items

OregSamSas:
Resolves #9133

### Changes

image
image

### Reason for changes

Read the issue for reasoning.

### Tests

Tested only in Brave.
Tested the keyboard navigation in the nested menu. (Only navigation in repeatedly nested menus don't work 100% clean, but we don't need such nested menus so far (only 2lvl deep ones).

#### Issues

<details><summary>Outdated</summary>Probably we should also crop too long folder names in the context menu... 🤔 🤣 
image
</details>

### DIFF
--- a/addon-api/content-script/Tab.js
+++ b/addon-api/content-script/Tab.js
@@ -904,6 +904,7 @@ export default class Tab extends Listenable {
    * @param {boolean=} opts.border - whether to add a border at the top or not.
    * @param {boolean=} opts.dangerous - whether to indicate the item as dangerous or not.
    * @param {Tab~EditorContextMenuItemCondition} opts.condition - a function to check if the item should be shown.
+  * @param {object[]=} opts.children - nested submenu items shown when hovering over this item.
    */
   createEditorContextMenu(...args) {
     addContextMenu(this, ...args);
--- a/addon-api/content-script/contextmenu.js
+++ b/addon-api/content-script/contextmenu.js
@@ -132,6 +132,7 @@ const onReactContextMenu = async function (e) {
       });
       itemElem.role = "menuitem";
       itemElem.tabIndex = "-1";
+
       if (hasChildren) {
         itemElem.setAttribute("aria-haspopup", "menu");
         itemElem.setAttribute("aria-expanded", "false");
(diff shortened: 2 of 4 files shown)


OregSamSas:
Also, you can help me with naming the features properly... 😃

OregSamSas:
Disclaimer: the "delete-others" option is not part of this addon, it's only on the image to show "compatibility", how the "delete ..." actions group together in the context menu.

OregSamSas:
The folder name overflow is fixed: 
image

The max width of the right click menu may be reduced if desired.

PR #9146, opened 2026-08-03
Add Ctrl/Cmd key support to editor-number-arrow-keys

kalyzu:
**Resolves**: No prior issues; I decided not to be redundant. This minimal change can be disputed here.

### Changes

Allows for another option for arrow key incrementation, the Ctrl key.
Logic has been implemented to use `e.metaKey` if the platform is MacOS.

### Reason for changes

I have found that I need arrow key incrementation for 10, 1, 0.1, _and_ 0.01 at the same time, so I (the lvl 0 experience dev) took it upon myself to add it, not just open an issue.

### Tests
All platforms worked exceptionally well with no bugs to mention.

Linux (Arch)
 * [x] Chromium
 * [x] Firefox

(Fix was extremely minimal, cross-platform testing was skipped)

### DIFF
--- a/addons/editor-number-arrow-keys/addon.json
+++ b/addons/editor-number-arrow-keys/addon.json
@@ -127,7 +127,44 @@
       "max": 8,
       "if": { "settings": { "useCustom": true } }
     },
+    {
+      "name": "Change on Ctrl+Key (Cmd on Mac)",
+      "id": "ctrl",
+      "type": "select",
+      "default": "hundredth",
+      "potentialValues": [
+        {
+          "id": "none",
+          "name": "None"
+        },
+        {
+          "id": "hundredth",
+          "name": "0.01"
+        },
+        {
+          "id": "tenth",
+          "name": "0.1"
+        },
+        {
+          "id": "one",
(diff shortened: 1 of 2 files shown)


Samq64:
See https://github.com/ScratchAddons/ScratchAddons/pull/6020#issuecomment-1646697293 and the surrounding discussion for why ctrl wasn't included. As far as I know it wasn't rejected, just unnecessary.

mxmou:
Since there are four possible increments by default, I think it makes sense to have a way to use all of them at once.

kalyzu:
> See [#6020 (comment)](https://github.com/ScratchAddons/ScratchAddons/pull/6020#issuecomment-1646697293) and the surrounding discussion for why ctrl wasn't included. As far as I know it wasn't rejected, just unnecessary.

I don't see how the referenced issue explains how it's unnecessary. I get the thought of not cluttering UI menus, but it's not that big of a deal if it gives users value. I like @mxmou's thought that if there are 4 input options, then it makes sense to be able to use them all at once.

For precise timing inputs, being able to use all 4 at once would be very handy.

PR #9149, opened 2026-08-05
Preserve stroke styles after paint boolean operations

griffpatch:
Follow-up to #8917 & the request in #8914

### Changes

Boolean operations now preserve stroke joins, caps, and miter limits when creating replacement paths. Text converted to paths uses the same shared style helper.

### Reason for changes

Combine and Expand previously restored only colours and width, causing round corners and endpoints to revert to Paper.js defaults.

### DIFF
--- a/addons/paint-boolean-ops/paper-helpers.js
+++ b/addons/paint-boolean-ops/paper-helpers.js
@@ -35,13 +35,15 @@ export const cleanResult = (r) => {
 };
 
 // ── Style helpers ──────────────────────────────────────────────────────
-// Snapshots an item's fill, stroke colour, and stroke width. Boolean ops
-// in paper.js reliably lose style information, so we save before and
-// restore after every operation.
+// Snapshots an item's fill and stroke style. Boolean ops in paper.js reliably
+// lose style information, so we save before and restore after every operation.
 export const cloneStyle = (src) => ({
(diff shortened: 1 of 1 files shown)


pufferfish101007:
When uniting a shape with round stroke joins, and a shape with miter stoke joins, the united shape ends up taking the stroke join style of the topmost shape. Is this consistent with behaviour of other programs providing similar functionality?

It makes sense why it happens, to me as someone who has some limited understanding of how SVGs are represented, but I'm not sure that it would be obvious to the average user.

pufferfish101007:
For anyone else looking to review this: you need to merge #8914 into this branch locally beforehand.

griffpatch:
You mentioned that uniting shapes with different corner styles was taking the top shape’s style. It should use the bottom shape’s outline, including its corner style.

I found that when groups are involved, it can get the order wrong and take the top shape’s style instead. I’ve fixed that locally, and my checks now keep the bottom shape’s style with grouped shapes too.

Were either of your shapes inside a group?

pufferfish101007:
I don't think either shape was inside of a group, no. I also might have incorrectly recorded which shape's style was being preserved.

I still don't find this behaviour particularly intuitive, but I also understand that doing anything more to try to preserve different stroke styles wouldn't really be feasible :)

pufferfish101007 (approved):

PR #9150, opened 2026-08-06
Join flow fixes

mxmou:
The bad username page [was "temporarily" removed in 2023](https://github.com/scratchfoundation/scratch-www/commit/d121de84e9a02884ae161db7dcd5bcffc58611e5). This PR deletes it from the `scratchWWWNoProject` match pattern and removes the CSS for it, which also fixes the join flow not having rounded corners - it looks like this in v1.45.2:

image<br />

After this change:

image

### DIFF
--- a/addons/dark-www/banners/darker.css
+++ b/addons/dark-www/banners/darker.css
@@ -99,13 +99,8 @@
 .sa-annual-report-2020 .reach-scratch-jr {
   background-color: #1c465a;
 }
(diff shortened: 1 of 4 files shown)


Samq64:
The main join flow looks good but with `scratch3to2` the editor one (which was already styled inconsistently) now has an unreadable header.

Samq64 (approved):

PR #9152, opened 2026-08-07
Use VM's serialization methods in updateAllBlocks

### LINKED ISSUE #9151: `updateAllBlocks()` makes block comments behave weirdly
### Description

If an addon that calls `updateAllBlocks()` is enabled, block comments move back to their default position and empty comments disappear after switching sprites.

### Steps to reproduce

1. Enable recolor-custom-blocks or any other addon that uses `updateAllBlocks()`
2. Add a comment to a block
4. Move the comment
5. Switch to another sprite and back

If the comment is not empty, it will move back to its default position next to the block. If it's empty, it will disappear.

### Environment

Browser: Edge 152 dev
Scratch Addons version: 2c4389744825217a2348a8813fffe27a7d03808d
Operating system: Windows

mxmou:
Resolves #9151

### Changes

Changes `updateAllBlocks()` to use the VM's methods to serialize blocks and comments instead of `blockly.Xml.workspaceToDom()`.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/libraries/common/cs/update-all-blocks.js
+++ b/libraries/common/cs/update-all-blocks.js
@@ -28,11 +28,11 @@ export async function updateAllBlocks(
       }
     }
     if (updateMainWorkspace) {
-      const xml = blockly.Xml.workspaceToDom(workspace);
-      if (xml.querySelector("variables")) {
-        xml.querySelector("variables").remove();
-      }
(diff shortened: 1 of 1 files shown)


Samq64 (approved):
When switching sprites comments do not update their positions instantly but that is also the case without Scratch Addons to an extent.

PR #9153, opened 2026-08-07
Popup & settings page: partial support for forced colors mode (high contrast)

mxmou:
### Changes

Fixes some issues in the extension UI that made it difficult to use when the forced colors mode was enabled. Side-by-side comparison on Edge:

| | 1.45.2 | This PR |
| --- | --- | --- |
| Switch (on) | image | image |
| Switch (off) | image | image |
| Category selector | image | image |
| Color picker | image | image |

The full settings page looks like this:

image<br />

And with a light color scheme (the default one on Firefox):

image<br />

### Known issue

A remaining issue is that icons don't have the right color (they might even be invisible, as seen in the screenshot above). This could be solved by using an icon font, as suggested in #1220, or inline SVGs, as suggested in #4989.

### Testing

There are several ways to turn on forced colors when testing these changes:
* System-wide setting on Windows: several preset color schemes are available in Windows settings → Accessibility → Contrast themes. It's also possible to choose custom colors.
* Developer setting on Chromium browsers: enable the forced-colors media feature on the Rendering tab in devtools.
* Browser setting on Firefox: Settings → Accessibility → Website contrast. A custom color scheme can be set.

### DIFF
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -364,7 +364,9 @@ button.setting-input.color {
 }
 
 @media (forced-colors: active) {
-  .header-button {
+  .header-button,
+  .categories-shrink,
+  .arrow-button {
     background-color: ButtonFace;
     outline: 1px solid ButtonBorder;
     outline-offset: -1px;
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -1,5 +1,7 @@
 html {
   background: var(--content-background);
+  scrollbar-width: thin;
+  scrollbar-color: var(--gray-text) transparent;
 }
 
 body {
@@ -228,26 +230,6 @@ body.iframe #search-not-found {
   text-align: center;
 }
 
-/*scrollbar*/
-
-/* width */
-::-webkit-scrollbar {
-  width: 7px;
-  height: 7px;
-}
-
-/* hide track */
-::-webkit-scrollbar-track,
-::-webkit-scrollbar-corner {
-  background: none;
-}
-
-/* Handle */
-::-webkit-scrollbar-thumb {
-  background: var(--gray-text);
-  border-radius: 4px;
-}
-
 /* Small screens & popup */
 @media (max-width: 700px) {
   .addons-block-header {
@@ -379,4 +361,20 @@ button.setting-input.color {
   align-items: center;
   justify-content: center;
   user-select: none;
+}
+
+@media (forced-colors: active) {
+  .header-button {
+    background-color: ButtonFace;
+    outline: 1px solid ButtonBorder;
+    outline-offset: -1px;
+  }
+
+  .color-container {
+    forced-color-adjust: none;
+  }
+
+  .color-container .setting-input.color {
+    border-color: ButtonBorder;
+  }
 }
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -371,6 +371,11 @@ button.setting-input.color {
     outline: 1px solid ButtonBorder;
     outline-offset: -1px;
   }
+  .header-button:focus,
+  .categories-shrink:focus,
+  .arrow-button:focus {
+    outline: revert;
+  }
 
   .addons-block-header {
     border-color: Canvas;
--- a/webpages/settings/style.css
+++ b/webpages/settings/style.css
@@ -372,6 +372,10 @@ button.setting-input.color {
     outline-offset: -1px;
   }
 
+  .addons-block-header {
+    border-color: Canvas;
+  }
+
   .color-container {
     forced-color-adjust: none;
   }
--- a/webpages/styles/components/buttons.css
+++ b/webpages/styles/components/buttons.css
@@ -93,4 +93,9 @@
     outline: 1px solid ButtonBorder;
     outline-offset: -1px;
   }
+
+  .large-button:focus-visible,
+  .icon-button:focus-visible {
+    outline: revert;
+  }
 }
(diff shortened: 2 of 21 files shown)


webpages/settings/style.css
```diff
@@ -380,3 +422,19 @@ button.setting-input.color {
   justify-content: center;
   user-select: none;
 }
+
+@media (forced-colors: active) {
+  .header-button {
+    background-color: ButtonFace;
+    outline: 1px solid ButtonBorder;
+    outline-offset: -1px;
+  }
```
DNin01: I think more buttons should have a border like this one.

Theme switch button

High contrast borders are a common theme for buttons on this page, but none of the close or arrow buttons have one.

Close button in modal

Currently, this is a problem because the icons don't adjust to forced colors (they can be the same color as the background).

webpages/settings/style.css
```diff
@@ -380,3 +422,19 @@ button.setting-input.color {
...
+
+  .color-container {
+    forced-color-adjust: none;
+  }
```
DNin01, line 381: This style rule applies to all elements inside the color input, including the color picker, preventing it from being fully accessible. This should be applied only to the color input, color well, and sliders.

webpages/settings/style.css
DNin01, line 1: `.addons-block-header` has a border but `.categories-block` and `.navbar` don't. I don't know what's conventional, but I would either remove the border from the bottom of `.addons-block-header` or add a border to the rest of these elements. Compare using these screenshots:

no borders borders

DNin01, line 1: The step numbers on the permission prompt have no background.

Permission prompt

(Updated from `upstream/master`.)

DNin01 (changes requested):
Tested on Edge 151 using the page colors accessibility setting.

webpages/settings/style.css
mxmou: I don't think it's possible to do that - those elements are in the picker's shadow DOM and are [only affected by style elements inside the shadow root](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#encapsulation_from_css).

webpages/settings/style.css
mxmou: I'll remove the border from `.addons-block-header` because it's mostly decorative.

webpages/settings/style.css
DNin01: I meant by modifying the color picker component's own [stylesheet](https://github.com/ScratchAddons/ScratchAddons/blob/2c4389744825217a2348a8813fffe27a7d03808d/libraries/thirdparty/color-picker.js#L178) to make the color well and color sliders appear as normal while all of its other elements are adjusted by the browser.

If you were talking about the `forced-color-adjust` property, it does affect the shadow root's elements, strangely. It's probably being inherited, because when I remove the property, everything in the color picker reverts to your "before" color picker screenshot. Knowing this, the effect could be counteracted with this style rule:
```css
color-picker {
  forced-color-adjust: auto;
}
```

mxmou: The file you linked is a minified [library](https://www.jsdelivr.com/package/npm/color-picker-web-component) that isn't meant to be edited directly.

DNin01: We could still do one of these things:
- The JavaScript is minified, but I don't see a problem with modifying the template literals that encode the stylesheets used by this library's components.
- Since the components are open shadow roots, their shadow DOM can be modified from the outside after insertion, allowing non-library code to patch in additional CSS.
- We can switch to a different color picker library.
- We can save this for a different PR.

mxmou: > * We can switch to a different color picker library.
> * We can save this for a different PR.

I think we need to switch a different color picker anyway (the current one has quite a few bugs), but that can be done in a future PR.

DNin01 (approved):
Other than forced colors not being applied to the color picker, I think this is an improvement.

I kind of wish, though, that fewer of these forced-colors media queries had to be used. Perhaps setting different values for the CSS variables in the `colors.css` file when forced colors are active would help save some lines of code now and in the future? Just an idea. Kinda difficult without a refactor, though.

DNin01:
@Samq64, you might be interested in testing and reviewing this.

webpages/styles/components/buttons.css
```diff
@@ -86,3 +86,11 @@
 [dir="rtl"] .split-button-dropdown {
   border-radius: 4px 0 0 4px;
 }
+
+@media (forced-colors: active) {
+  .icon-button {
+    background-color: ButtonFace;
+    outline: 1px solid ButtonBorder;
```
DNin01, line 93: Button outlines don't change color when focused.

webpages/styles/components/filter.css
DNin01, line 1: The focus ring is difficult to see over highlighted backgrounds like selected option rows and categories.

Option row with selected and focused item

Maybe it could be changed to `HighlightText` and inset a bit? I'm not sure what's standard for these.

DNin01 (commented):
I found a couple more issues.

webpages/styles/components/filter.css
mxmou: I'll probably change the outline to be outside the element and slightly offset. That's what Windows itself does for most focused elements, including those which have the highlight color as the background color:

image

image

mxmou: I wasn't able to do that because the parent element has `overflow: hidden`, so I changed the color to `HighlightText` as you suggested.

DNin01 (approved):

mxmou:
> Perhaps setting different values for the CSS variables in the `colors.css` file when forced colors are active would help save some lines of code now and in the future? Just an idea. Kinda difficult without a refactor, though.

One thing that would definitely help is reusing more CSS between the popup and the setting page.

PR #9154, opened 2026-08-07
color-picker: set color when clicking outside

### LINKED ISSUE #5993: color-picker bug
> Are you aware of any reproduceable bugs mentioned above (in this issue) that weren't fixed by your PR? If so, we'll need to create issues for them.

Clicking away from the input without pressing Enter updates the color of the input, but not the brush color.

_Originally posted by @CST1229 in https://github.com/ScratchAddons/ScratchAddons/issues/5296#issuecomment-1529147021_

mxmou:
Resolves #5993

### Changes

Fixes the color not updating when clicking outside the hex color input after editing it.

To update the color, the addon activates the eyedropper, then deactivates it after a timeout. But if the update was the result of a click, the click event would deactivate the eyedropper before the addon could do it. Reducing the timeout from 50ms to a single tick fixes the bug.

### Tests

Tested on Edge and Firefox.

### DIFF
--- a/addons/color-picker/paint-editor.js
+++ b/addons/color-picker/paint-editor.js
@@ -39,7 +39,7 @@ export default async ({ addon, console, msg }) => {
         addon.tab.redux.dispatch({
           type: "scratch-paint/eye-dropper/DEACTIVATE_COLOR_PICKER",
         });
-      }, 50);
+      }, 0);
     };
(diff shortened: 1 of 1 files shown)


mxmou:
I've decided to refactor the code so that it won't use the eyedropper hack at all.

PR #9157, opened 2026-08-10
Fix paint editor color picker z-index

### LINKED ISSUE #5986: 2d-color-picker causes color picker modal to be in wrong position
### Description

Here's a screenshot of the bug:

image

### Steps to reproduce

> it just moves off screen up, this has only happened with scratch adons

> idk how it happens, it just does
and it only happens with 2 colors

> \- When this happens, does it continue to happen during the session? Or does it get fixed if you exit the color picker and immediately reopen it?
> \- It continues to happen

> the specific thing is that i select the fading options, but i dont know if any addon causes his
its caused by the 2d color picker
i tried it with, and without

_No response_

### Environment

Browser: probably Chromium
Scratch Addons version: probably latest

mxmou:
Resolves #5986

### Changes

The color picker in the costume editor automatically adjusts its position if there isn't enough vertical space for it, but because its z-index is too low, this may cause other elements to appear in front of it.

This PR adds a CSS rule that increases the z-index to all addons that make the color picker taller.

image

### Tests

Tested on Edge.

### DIFF
--- a/addons/2d-color-picker/style.css
+++ b/addons/2d-color-picker/style.css
@@ -1,3 +1,7 @@
+.Popover {
+  z-index: 492; /* above menu bar */
+}
+
 .sa-2dcolor-picker {
   width: 150px;
   height: 150px;
--- a/addons/color-picker/style.css
+++ b/addons/color-picker/style.css
@@ -1,3 +1,7 @@
+.Popover {
+  z-index: 492; /* above menu bar */
+}
+
 .sa-color-picker {
   display: flex;
 }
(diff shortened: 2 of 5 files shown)


Samq64 (changes requested):
The z-index is overridden again by editor-stage-left:
https://github.com/ScratchAddons/ScratchAddons/blob/86b4c0dc04a68f07db0cf31e9de207d9ce4fd51c/addons/editor-stage-left/stageleft.css#L85-L88

Samq64 (approved):

Issue #7933, opened 2024-11-11
Make `redux.initialize()` async


Samq64:
### Why this would be helpful

Running `addon.tab.redux.initialize()` doesn't grantee that Redux is finished initializing because it is not asynchronous.``

### My suggestion

Make `addon.tab.redux.initialize()` asynchronous so calling it grantees Redux is fully initialized by the next line.

### Possible alternatives

Only using Redux after a `waitForElement` or checking if it is initialized in a loop.

mxmou:
> Only using Redux after a `waitForElement` or checking if it is initialized in a loop.

`waitForElement` is actually affected by this issue: if it's called with a reduxEvents option in a userscript with `"runAtComplete": false`, it doesn't work consistently.

pufferfish101007:
As a note, the addons that I could find that currently use `"runAtComplete": false` along with aforementioned workarounds are:
- `default-project`
- `player-thumb`
- `folders`
(plus `paint-boolean-ops` but #9164 changes that)

I didn't check for `waitForElement` with `reduxEvents`.

PR #9164, opened 2026-08-13
paint-boolean-ops: fix toolbar and Redux initialization timing

### LINKED ISSUE #9159: Open/Close Shape button doesn't appear until More menu is opened
### Description

The Open Shape or Close Shape button added to the reshape tool by paint-boolean-ops doesn't appear until the More menu is opened, even though it's not in that menu. It also doesn't disappear when switching from the Reshape tool to the Select or Text tool.

### Steps to reproduce

1. Enable paint-boolean-ops and open the costume editor
2. Resize the window so that a "More" menu appears
3. Select a shape
4. Select the Reshape tool - the button should now appear, but it's missing
5. Open the More menu - the button now appears, even though it's not in the menu
6. Switch to the Select or Text tool - the button should now disappear, but it's still there

_No response_

### Environment

Browser: Edge 152 dev
Scratch Addons version: 2c4389744825217a2348a8813fffe27a7d03808d
Operating system: Windows

griffpatch:
**Blocked by #9180. Please merge #9180 first.** This PR currently includes the same Redux changes so the paint fix can use them. Once #9180 is merged, I will update this branch so only the paint changes remain.

The Open/Close button belongs to Scratch Paint's Reshape toolbar, but it was being updated along with the separate fixed toolbar and More menu.

On a narrow editor, this meant the button did not appear until More was opened. It could also stay behind after switching to Select or Text.

The button now follows the Reshape toolbar directly. It is removed when Reshape is no longer active and added again if React replaces that toolbar.

While checking this, I found that Open/Close was disabled for paths selected inside a Group or CompoundPath, even though Scratch can open and close those paths. The selection check now includes them. Releasing a compound path also keeps its children in the same parent.

Disabled Boolean operations in More now use the same disabled styling as Scratch's own items.

### Follow-up after review

After the original review, I checked this alongside the Redux initialization issue in #7933.

Open/Close needs the current paint mode when the addon starts, so it now waits for Redux before checking that mode. This uses the Redux initialization change in #9180.

Waiting for Redux revealed that Scratch can create the fixed toolbar row before adding its button groups. The toolbar helper now refreshes the native classes when those groups appear, so the Boolean buttons stay on one row.

Fixes #9159

### DIFF
--- a/addons/paint-boolean-ops/userscript.js
+++ b/addons/paint-boolean-ops/userscript.js
@@ -16,8 +16,10 @@ import {
 } from "./paper-helpers.js";
 
 export default async function ({ addon, msg }) {
-  // Redux must be initialized before waitForElement can use reduxCondition.
+  // Watch Redux so the button follows the Costume editor's active paint tool.
   addon.tab.redux.initialize();
+  // Wait until Scratch's generated More-menu disabled class is available.
+  await addon.tab.scratchClassReady();
 
   // Mangled disabled-state class, populated after first DOM injection.
   let modDisabledClass = "";
@@ -76,7 +78,8 @@ export default async function ({ addon, msg }) {
     const btn = e.target.closest("[data-sa-op]");
     if (!btn) return;
     const op = btn.dataset.saOp;
-    if (modDisabledClass && btn.classList.contains(modDisabledClass)) return;
+    // Inline and More buttons use different disabled classes, but both contain this marker.
+    if ([...btn.classList].some((className) => className.includes("mod-disabled"))) return;
     if (op === "combine") performCombine();
     else if (op === "release") performRelease();
     else if (op === "expand") performOffset();
@@ -89,8 +92,10 @@ export default async function ({ addon, msg }) {
 
   // ── More-popover item factory ──────────────────────────────────────────
   // Builds a button matching the native More popover item structure.
-  // Disabled-state classes are copied from native items when the popover opens.
-  let moreItemDisabledClasses = [];
+  // Disabled More items need both the menu class and the shared button class
+  // for correct behavior and opacity. Resolve the menu class directly because
+  // the menu may contain no disabled native item to copy.
+  const moreMenuDisabledClass = addon.tab.scratchClass("fixed-tools_mod-disabled");
 
   const makeMoreItem = (iconFile, label, op) => {
     const btn = document.createElement("span");
@@ -123,6 +128,10 @@ export default async function ({ addon, msg }) {
   // Schedules updateButtonStates once the round-trip above has settled.
   const deferUpdateButtonStates = () => afterReduxRoundTrip(updateButtonStates);
 
+  // Include paths inside Groups and CompoundPaths, not only top-level shapes.
+  const getOpenClosePaths = (paper) =>
+    paper.project.selectedItems.filter((item) => item.layer?.data?.isPaintingLayer && item instanceof paper.Path);
+
   // Reads the current paper.js selection and enables or disables each button
   // accordingly. Also morphs the Combine/Release and Open/Close buttons.
   const updateButtonStates = async () => {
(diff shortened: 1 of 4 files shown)


addons/paint-boolean-ops/userscript.js
```diff
@@ -682,22 +680,22 @@ export default async function ({ addon, msg }) {
...
-  // bar. Removes it in all other modes. Called on every CHANGE_MODE event
-  // because React rebuilds mode-tools children when the active tool changes.
-  const injectModeToolsBtn = () => {
-    const mode = addon.tab.redux.state?.scratchPaint?.mode;
-    if (mode !== "RESHAPE") {
-      modeToolsOCBtn.remove();
-      return;
-    }
+  // Reshape's mode-tools bar is separate from the fixed toolbar/More menu.
+  // Watching the fixed toolbar caused Open/Close to appear only after More opened.
```
mxmou: The comment should explain why the current code is correct (if that's necessary), not why the previous code was wrong.

addons/paint-boolean-ops/userscript.js
```diff
@@ -682,22 +680,22 @@ export default async function ({ addon, msg }) {
...
+  const modeToolsGroupSelector = "[class*='mode-tools_mode-tools_'] [class*='mode-tools_mod-dashed-border_']";
+  const isReshapeState = (state) =>
+    state?.scratchGui?.editorTab?.activeTabIndex === 1 &&
+    !state.scratchGui?.mode?.isPlayerOnly &&
+    state.scratchPaint?.mode === "RESHAPE";
+  const isReshapeActive = () => !addon.self.disabled && isReshapeState(addon.tab.redux.state);
```
mxmou:
```suggestion
  const showOpenCloseButton = () => !addon.self.disabled && isReshapeState(addon.tab.redux.state);
```

mxmou (commented):
Looks good; the only things I'm not sure about are a comment and a function name.

griffpatch:
Thanks @mxmou, both points make sense. I've updated the comment to explain why the Reshape sub-toolbar is watched separately, and renamed the helper to `showOpenCloseButton`. No behavior changes. Lint and syntax checks pass.

addons/paint-boolean-ops/userscript.js
```diff
@@ -725,9 +723,39 @@ export default async function ({ addon, msg }) {
...
+      injectModeToolsBtn(dashedGroup);
+    }
+  };
+  watchModeTools();
+
+  // React can also reuse the same group, so recheck it after React updates.
+  const syncModeToolsBtn = () =>
+    afterReduxRoundTrip(() => injectModeToolsBtn(document.querySelector(modeToolsGroupSelector)));
+
+  addon.tab.redux.addEventListener("statechanged", ({ detail }) => {
```
mxmou, line 744: This doesn't always work because of a bug in the addon API (#7933). As a workaround, `"runAtComplete": false` could be removed from the addon's userscript.

mxmou (approved):

griffpatch:
I have added a small follow-up for #9180.

Open/Close now waits until the current paint mode is available. The toolbar helper also refreshes its native classes if Scratch adds the button groups after the toolbar row, which stops the Boolean buttons stacking onto separate rows.

addons/paint-boolean-ops/userscript.js
griffpatch: Thanks, removing "runAtComplete": false would avoid this race by delaying the whole userscript, but it would only work around #7933. I have fixed the addon API issue in #9180 and included that change here, so this addon can keep its existing early startup and wait specifically for Redux before checking the paint mode. If we want a quick fix then I don't mind reverting and putting in the "runAtComplete": false option... but see what you guys think first, this could be useful going forward.

pufferfish101007:
I'm happy with the fixes for `paint-boolean-ops` but I'm not confident enough with internals to approve the changes to `redux.initialize`.

If the changes here entirely subsume the changes in #9180 (which it looks like they do?) then I think it should either be marked as blocking so that it is merged first, or closed in favour of this.

griffpatch:
Thanks @pufferfish101007, yes, #9164 currently includes the same Redux changes as #9180. I'd prefer to keep the shared API fix separate for review, so I've marked this PR as blocked by #9180 in the description. Once that is reviewed and merged, I'll update this branch so only the paint changes remain.

pufferfish101007 (approved):
approving with the expectation of #9180 being merged first

PR #9167, opened 2026-08-15
New addon: clear backpack

### LINKED ISSUE #7642: Clear backpack
### Why this would be helpful

I have too much old junk in my backpack and/or I backpacked something a long time ago that I want to access again.

### How this addon works

This addon adds a right click item in the backpack to delete all backpacked code and assets. It should display a confirmation dialog as well.

### Possible alternatives

Alternatively, items in the backpack could be made multi-select-able with shift-click or ctrl-click so you can also delete multiple items at once.

### Additional context

_No response_

mohammedhamdankhangeminiuser-crypto:
Resolves #7642


Samq64:
Could you provide some information and maybe a screenshot?

PR #9172, opened 2026-08-21
fix: fix dark mode styles for confirmation prompts

### LINKED ISSUE #9171: Set thumbnail popup 'yes' button is dark on dark editor addon
### Description

Set thumbnail popup 'yes' button is dark on dark editor addon.

### Steps to reproduce

1. get scratch addons
2. turn on dark mode for editor addon
3. go to scratch and then the editor and click the set thumbnail button

Image

### Environment

Browser: Chrome 151
Scratch Addons version: 1.45.2.
Operating system: Windows

rosics-code:
Resolves https://github.com/ScratchAddons/ScratchAddons/issues/9171

### Changes

Moves a class selector to the bottom.

### Reason for changes

Fixes an issue where the yes button is dark.

### DIFF
--- a/addons/editor-dark-mode/experimental_editor.css
+++ b/addons/editor-dark-mode/experimental_editor.css
@@ -184,7 +184,6 @@
   color: var(--editorDarkMode-accent-text);
   color-scheme: var(--editorDarkMode-accent-colorScheme);
 }
-[class*="confirmation-prompt_confirm-button_"],
 [class*="confirmation-prompt_cancel-button_"][class*="delete-confirmation-prompt_button-row-button_"] {
   background-color: var(--editorDarkMode-accent);
 }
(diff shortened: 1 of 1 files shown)


Samq64:
If I remember correctly the dark button backgrounds are intentional but I do think white would look better in this case. Right now though the yes button is invisible with most colour combinations. Also keep in mind the styles are shared with the sprite deletion prompt.

rosics-code:
> If I remember correctly the dark button backgrounds are intentional but I do think white would look better in this case. Right now though the yes button is invisible with most colour combinations. Also keep in mind the styles are shared with the sprite deletion prompt.

Okay but the sprite deletion prompt is also purple so it's pretty useless to be black.

mxmou:
This makes both Yes buttons very hard to see:

image<br />

image<br />

They should either be dark, as in the current version of the addon, or white, as in light mode:

image<br />

image

PR #9180, opened 2026-09-02
Make redux.initialize() wait until Redux is ready

### LINKED ISSUE #7933: Make `redux.initialize()` async

griffpatch:
Some userscripts start before Scratch has finished setting up Redux. If they call `addon.tab.redux.initialize()` too early, it currently returns straight away and their listener is never connected.

This creates the Redux event channel earlier and lets addons wait until the first real Redux state is available:

```js
await addon.tab.redux.initialize();
```

Addons that only listen for later changes can continue calling it without `await`.

`waitForState()` now waits for Redux to be ready before checking the state.

The readiness promise exists even while Redux is still loading. If it is missing, `initialize()` now throws an error instead of carrying on as though Redux were ready.

Fixes #7933

### DIFF
--- a/addon-api/content-script/ReduxHandler.js
+++ b/addon-api/content-script/ReduxHandler.js
@@ -13,21 +13,25 @@ export default class ReduxHandler extends Listenable {
   }
 
   /**
-   * Initialize the handler. Must be called before adding events.
+   * Initialize the handler and wait until Redux has its first real state.
+   * Existing callers do not need to await this unless they use Redux immediately.
+   * @returns {Promise<void>}
    */
   initialize() {
-    if (!__scratchAddonsRedux.target || this.initialized) return;
-    this.initialized = true;
-    __scratchAddonsRedux.target.addEventListener("statechanged", ({ detail }) => {
-      const newEvent = new CustomEvent("statechanged", {
-        detail: {
-          action: detail.action,
-          prev: detail.prev,
-          next: detail.next,
-        },
+    if (__scratchAddonsRedux.target && !this.initialized) {
+      this.initialized = true;
+      __scratchAddonsRedux.target.addEventListener("statechanged", ({ detail }) => {
+        const newEvent = new CustomEvent("statechanged", {
+          detail: {
+            action: detail.action,
+            prev: detail.prev,
+            next: detail.next,
+          },
+        });
+        this.dispatchEvent(newEvent);
       });
-      this.dispatchEvent(newEvent);
-    });
+    }
+    return __scratchAddonsRedux.ready ?? Promise.resolve();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


addon-api/content-script/ReduxHandler.js
```diff
@@ -13,21 +13,25 @@ export default class ReduxHandler extends Listenable {
   }
 
   /**
-   * Initialize the handler. Must be called before adding events.
+   * Initialize the handler and wait until Redux has its first real state.
+   * Existing callers do not need to await this unless they use Redux immediately.
```
mxmou:
```suggestion
   * Callers do not need to await this unless they use Redux immediately.
```
People who read this comment in the future won't know which callers were existing when it was written.

addon-api/content-script/ReduxHandler.js
```diff
@@ -13,21 +13,25 @@ export default class ReduxHandler extends Listenable {
...
+            prev: detail.prev,
+            next: detail.next,
+          },
+        });
+        this.dispatchEvent(newEvent);
       });
-      this.dispatchEvent(newEvent);
-    });
+    }
+    return __scratchAddonsRedux.ready ?? Promise.resolve();
```
mxmou, line 34: If I'm reading the code correctly, the `ready` promise is supposed to always exist. So I'd expect this to throw an error instead of simply returning if `ready` is somehow undefined.

addon-api/content-script/ReduxHandler.js
mxmou, line 17: Assuming this applies to future callers as well, it should probably just say "callers".

addon-api/content-script/ReduxHandler.js
DNin01: Wouldn't it safely result in undefined, assuming `__scratchAddonsRedux` itself is defined as an object? Accessing object keys that aren't defined just result in undefined. It's only when you try to access a property within the undefined value without optional chaining that you get an error.

griffpatch: Thanks, agreed. The promise should always exist, even while Redux is still loading, so I've removed the fallback and added an explicit error if it is missing.

@DNin01 yes, reading a missing property gives `undefined`. The problem here was that the fallback then returned an already resolved promise, so `await initialize()` could carry on as though Redux were ready. Simply removing the fallback would still allow `await undefined` to continue, which is why I've added the explicit check. Normal loading still waits on the original promise.

addon-api/content-script/ReduxHandler.js
griffpatch: Good point, this applies to future callers too. Changed it to "Callers".

mxmou (approved):

PR #9188, opened 2026-09-18
Added ScratchAddons' Customizable FPS blocks addon

cloudberrypitech:
#### Usage
To use customizable fps, enable the addon in ScratchAddons, reload the Scratch page and then click on the Custom FPS tab


Samq64:
Sorry but generally new blocks are not allowed because they won't work for those without Scratch Addons. The debugger blocks are an exception because they are not intended to be used in shared projects.

